Issue #2, opened 2018-02-04
Decide on a unicode symbol for Cred


dandelionmane:
It's important that Cred have a unicode symbol. This will save us from needing to write "SourceCredCred" and other such ungainly constructs.

I've been provisionally using ¤ as the symbol; I like that it is reminiscent of a network, something that is both a circle and has flows in/out... the circle represents egalitarian community and the edges represent new contributions? :smiley: 

Here are some other options:
¢￠ ₡ ¢ ₢ ₵

Here are all the options in context:

SourceCred¤
SourceCred¢
SourceCred￠
SourceCred₡
SourceCred¢
SourceCred₢
SourceCred₵

![image](https://user-images.githubusercontent.com/1400023/35776257-cffa59d2-094d-11e8-8c7b-ccac52fe59a9.png)

My vote is for ¤. Please feel free to suggest characters I haven't thought of yet. There are *a lot* of them.

Also, yes this is the bikeshed.

000drax:
I like the symbol idea!

Chiming in on the bikeshed... I think the symbol should be simple and familiar.
SourceCredϾ
SourceCred¿

I'm rather partial to the Lunate Sigma (Ͼ).

dandelionmane:
I like Lunate Sigma too, although I'm still quite fond of the [currency sign ¤](https://en.wikipedia.org/wiki/Currency_sign_(typography)). Now that I know that ¤ semantically means "an unknown currency" it seems pretty appropriate. 

Some more examples to look at:
Python¤
TensorFlow¤
Git¤

PythonϾ
TensorFlowϾ
GitϾ

Hmm, the fact that Lunate Sigma looks like a "C" does make it easier for my brain to parse/pronounce PythonϾ as "PythonCred"

dandelionmane:
Ooh, another benefit to using Ͼ: Everywhere that I was writing "Cred", I can now write "Ͼred". That makes it extremely transparent that Ͼ stands for Ͼred. 

I'm going to switch to Lunate Sigma as the default choice for unicode symbol :D

dandelionmane:
Another benefit to Ͼ: if someone wants to write 1000Ͼ but doesn't know how to grab that unicode symbol, they can just write 1000C and it will still be pretty clear what they mean.

wchargin:
Two notes on the lunate sigma:
[...]

PR #3, opened 2018-02-05
s/Cred/Ͼred/

dandelionmane:
@000drax suggested in #2 using Ͼ as our unicode symbol.
This PR is an experiment in spelling Cred as Ͼred, so that it's really clear from context what Ͼ means.

But maybe it is too clever / kinda ugly / a pain in the ass to follow as possible since keyboards don't ship with Ͼ keys. Also, depending on the font my brain wants to read it as "Gred" :( 

Also, it's bikeshedding. Don't take this too seriously, but tell me what you think :)


dandelionmane:
Actually, when describing the concept of cred in general, I think it's better to use lowercase c rather than 'Cred' or 'Ͼred'. Closing this PR.

PR #4, opened 2018-02-05
Add overview of cred to the README

dandelionmane:
Originally, I planned to keep the README very succinct, and put an overview of SourceCred into a separate overview document.

This PR contains a first draft of the overview, but it seemed short enough to include in the README directly. I welcome feedback on whether this is a good fit for the README, or if keeping a separate overview is a good idea.


dandelionmane:
Actually, going to keep the README simple and move the overview into its own document.

PR #8, opened 2018-02-11
Check in experimental commit graph traversal

wchargin:
Summary:
This is a minimal script that will crawl the file-reachable commits in a
repository and dump statistics to a JSON file that can be further
analyzed interactively on a frontend.

Requires PyPI `gitpython` (and `matplotlib`, but this can be excised).
Requires `jupyter` to run.

(Mostly paired with @dandelionmane on 2018-02-08.)

wchargin-branch: experiment-commit-traversal

### DIFF
--- a/experiments/01_commit_graph_dump.ipynb
+++ b/experiments/01_commit_graph_dump.ipynb
@@ -1,0 +1,368 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Exploring the commit graph\n",
+    "\n",
+    "(2018-02-08)\n",
+    "\n",
+    "This script crawls a repository's commit graph by inspecting files\n",
+    "currently in the repository and looking at all commits that have changed\n",
+    "that file (tracking over renames). The data is lightly processed and is\n",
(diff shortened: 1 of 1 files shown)


wchargin:
The rendered notebook apparently doesn't appear in the diff, but does appear here:

https://github.com/sourcecred/sourcecred/blob/5f3db643b172c963557e3cbf2c313dc95ae0f3d3/experiments/01_commit_graph_dump.ipynb

dandelionmane:
x-ref from https://github.com/sourcecred/sourcecred/pull/10, it looks like the script is incorrectly recording rename paths as file paths.

wchargin:
That makes sense, given that we didn't put any logic to follow the chain of renames. I'll revise and resubmit.

dandelionmane:
I'm going to merge this, then fix issues myself.

dandelionmane:
Looks like the underlying GitPython library is doing something dumb with the renames - rather than having an explicit field in the FileStats object, it just makes the file name into "(?prefix/){old => new}(?/suffix)".
Since this notebook is marked as an experimental tool, let's check it in for now, and compensate by skipping renames in the demo frontend. We'll worry about doing the right thing in v0 of the production code.

PR #9, opened 2018-02-12
[wip] Develop the explorer with Typescript

dandelionmane:
This is a work in progress. Currently trying to persuade Typescript to compile.


wchargin:
Cool; ping me when you want me to take a look!

dandelionmane:
Meh, I'm starting to feel discouraged re: using Typescript. 

Currently I'm blocked on importing json, our `import data from './commits.json'` oneliner works with react-scripts but not react-scripts-ts. I'm not sure what's wrong (or even where the error is being thrown from?) - webpack >=2.0 has default support for json so maybe it's the Typescript compiler complaining. One thing I'm noticing is that react-scripts-ts just has a much lower level of polish than react-scripts (which makes sense as its 3rd party to fb). The default config is also really, really noisy and it has some bad usability e.g. blocks compilation on lint errors.


## Pros/Cons of Typescript:

### Pros: 
The typing in commits.ts is pretty nice. Files like that are where I like static typing. On the other hand, ts didn't let me define a recursive data type :(

Writing explicit types for every Props and State is pretty nice. 

### Cons:
Forces us to pay maintenance cost upfront, rather than once we know what code we're gonna keep.

Forces us to deal with typescript in areas where it doesn't actually provide much lift (its awful when you are just touching html stuff)

Significant build and dependency complexity, also "off the beaten path" of great tooling provided by fb

### Alternatives:
you mentioned react proptypes in email, these actually look better for component boundary typing, as they do runtime checking and therefore can even do things like verify enum values

we could use Flow for files like commit.ts where static typing is a big help

### conclusion
Let's ditch Typescript like you suggested. If you wanna do the honors, feel free to take a lock via comment on this PR

dandelionmane:
Deprecated in favor of https://github.com/sourcecred/sourcecred/pull/10

PR #13, opened 2018-02-16
Frontend setup

dandelionmane:
Initial infrastructure for sourcecred-explorer. 
Setup npm, yarn, flow, jest using scaffolding from
create-react-app.

- used create-react-app to get the basic pieces
- added support for flow via instructions in the
create-react-app user guide
- added flow-typed so that flow+jest will work well


explorer/.flowconfig
```diff
@@ -0,0 +1,12 @@
+[ignore]
+
+[include]
+
+[libs]
+flow-typed
+
+[lints]
```
wchargin: Okay for now. I really want to get ESLint set up, and preferably also as a blocking CI. But that can come later.

explorer/src/index.css
```diff
@@ -0,0 +1,5 @@
+body {
```
wchargin: Also okay for now. In the future, I'd like to at least consider using [aphrodite] for styling. It's worked very well for me in the past. TL;DR: It combines the benefits of inline styling (locality of definitions, single-file components) with the full power of CSS—things like `:hover`, `::after`, `@keyframes`, and `@media` that you can't get with just `style` attributes.

(If you care about things like server rendering and minimizing bandwidth, it fares well there, too. But that's a discussion for another day.)

[aphrodite]: https://github.com/Khan/aphrodite

explorer/src/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import './index.css';
+import App from './App';
```
wchargin: Missing `App.js`?

Missing tests?

explorer/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+/node_modules
+
+# testing
+/coverage
+
+# production
+/build
+
+# misc
+.DS_Store
```
wchargin: I know that this is just the output of `create-react-app`, but I don't think that `.DS_Store` should be here. Instead, it should be in a global gitignore:
```sh
$ echo '.DS_Store' >> ~/.gitignore_global
$ git config --global core.excludesFile ~.gitignore_global
```
There are dual advantages:
  * your global gitignore, with your platform/editor/IDE-specific bindings, persists across all projects;
  * when ten different developers on ten different platforms/editor/IDEs start contributing to the project, the project's `.gitignore` doesn't get polluted.

For instance, here's mine: https://gist.github.com/wchargin/868e0147911c0486a751ad99a71133fb

explorer/src/index.js
wchargin: Per discussion—we'll keep this PR as is; #14 will make everything work.

PR #14, opened 2018-02-16
Initial prototype of the cred explorer

dandelionmane:
Sets up an initial cred explorer that looks at Git commit data

- uses git commit data as generated by `experiments/01_commit_graph_dump.ipynb`
- displays a file explorer, which lets you navigate the file tree by expanding/collapsing dirs, and select a path via double clicking (changes background color)
- displays a user explorer, which shows a sorted list of authors and cred amounts for the selected file path

The commit-wrangling logic is in commitUtils.js and has typechecking (via flow) and some basic testing.
The frontend is built with react; setup was bootstrapped via create-react-app.

Test plan:

```
cd explorer;
yarn install; # get dependencies
yarn flow; # check that types check out
yarn test; # check that tests pass, then `q`
yarn start; # go to localhost:3000 and click around
```

expected behavior: there's a file explorer that shows TensorBoard file structure on the left. you can click the buttons to explore directory, and you can click files to select that path. on the right, it shows you the users that contributed to that directory and their cred for those commits.

note: the ui is very crude.


wchargin:
(I've diffbased this against `frontend-setup`. Before this is merged, PR #13 should be merged into master, and this should be diffbased against master again. If you have a better workflow, please let me know!)

explorer/src/commitUtils.js
```diff
@@ -0,0 +1,93 @@
...
+  added: number;
+  deleted: number;
+}
+
+export function commitWeight(commit: Commit, filepath: string): number {
+  // hack - GitPython encodes renames in the filepath. ignore for now.
+  if (filepath.indexOf('=>') !== -1) {
+    return 0;
+  }
+  return Math.sqrt(commit.stats[filepath].lines);
```
wchargin: Fine for now, but—per our discussions, I thought that *this* was going to be log, and it would be the voting-power-as-a-function-of-cred that was radical?

explorer/src/commitUtils.js
```diff
@@ -0,0 +1,93 @@
...
+        w = 0;
+      } else {
+        w = weightFn(commit, file);
+      }
+      yield [commit.author, w];
+    }
+  }
+}
+
+type WeightFn = (commit: Commit, filepath: string) => number;
```
wchargin: I think this merits a docstring, to make it clear that weights are aggregated for a file, and that this does not compute the "weight of a commit". What do you think of something like

> A weight function determines how much commit contributes to the importance of a particular file. For instance, a weight function might be defined as, “the number of lines changed in `filepath` in commit `commit`”, or “`1` if `filepath` was changed in `commit`, else `0`”.

?

explorer/src/commitUtils.test.js
```diff
@@ -0,0 +1,58 @@
...
+      'stats': {
+        'foo.txt': {'lines': 5, 'added': 3, 'deleted': 2},
+        'bar.txt': {'lines': 100, 'added': 100, 'deleted': 0}
+        }
+      },
+    '2': {
+      'author': 'wchargin',
+      'stats': {
+      'bar.txt': {'lines': 100, 'added': 50, 'deleted': 50},
+        }
```
wchargin: nit: could you reindent this object literal, please? (e.g., in vim: press `=ap`)

wchargin:
Forgot to mention:

> ```sh
> yarn flow;  # check that types check out, then ctrl+c
> ```
I don’t have to Ctrl+C; my `yarn flow` doesn’t watch automatically. Does yours, or is this a typo?

> ```sh
> yarn test;  # check that tests pass, then ctrl+c
> ```
Alternately: `CI=true yarn test` will run tests just once. (Kind of bad UI, in my opinion; facebook/create-react-app#784 has discussion.)

And thanks for the test plan! We can add frontend component tests soon.

explorer/src/commitUtils.js
```diff
@@ -0,0 +1,93 @@
+// @flow
+
+type CommitData = {
+  // TODO improve variable names
+  fileToCommits: {[filename: string]: string[]};
+  commits: {[commithash: string]: Commit};
```
wchargin: Do you know why we get `no-use-before-define` lints here and in similar places (lines 12, 34)? Pasting these three paragraphs into https://flow.org/try yields no errors ([direct link here](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDCcAtiQJYYAiAhhjWALxgDeqYYwwYAKgPJU+YciRwAnOADcCkmmPI0ARjAIA7GiTwBndmCjkVPOMTKUtALlYBtfSvWbLWjPNUBzALqPn5N1fcBfAG5dAGNSCgwLazDTDAALGi04rxcPSxMIoN0aAFd4uDEop1S-YP90bHwicMomVmy8uIKUn1dgjic6KJYbAzx7PBa3TzAAMT6AZXpIrPLMXAJxlSmuurYOGB9tS1UckkU8MXawGgATU7xTnb2Do90LlQxL6-3DstQgA)).

I'd love to not get into the habit of ignoring lint at this stage. :-)

explorer/src/commitUtils.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+  return Math.sqrt(commit.stats[filepath].lines);
+}
+
+function allSelectedFiles(filepath: string, data: CommitData): string[] {
+  const fnames = Object.keys(data.fileToCommits);
+  return fnames.filter(x => x.startsWith(filepath));
+}
+
+function *userWeights(files: string[], data: CommitData, weightFn: WeightFn): Iterable<[string, number]> {
```
wchargin: This also merits a docstring, in my opinion; it's not clear from the interface there may be multiple tuples for a given user. (I.e., the result is not just an associative array.)

explorer/src/FileExplorer.js
```diff
@@ -0,0 +1,95 @@
...
+    return <div 
+        className={isSelected ? 'selected-path' : ''}
+        style={{marginLeft: this.props.path === "." ? 0 : 25}}
+      >
+
+      <p>
+        {isFolder && <button
+          style={{marginRight: 3}}
+          onClick={toggleExpand}>»</button>}
+        <span // TODO should be a button or <a> for accessibility
```
wchargin: Why not just make this an `<a>` or `<button>` now?

explorer/src/commitUtils.js
dandelionmane: Yeah, log is probably a better choice. We'll play around with a few different weightFns (note that it's an argument, easy to swap)

explorer/src/commitUtils.js
dandelionmane: Done

PR #18, opened 2018-02-17
Set up prettier as an autoformatter

dandelionmane:
I followed the setup instructions in the [create-react-app user guide](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#formatting-code-automatically). The formatter runs as a git commit hook.

 I made one change, which is to prefer double rather than single quotes.

I also ran the formatter for all our existing source files, except for src/data.json.

### DIFF
--- a/explorer/package.json
+++ b/explorer/package.json
@@ -50,14 +50,14 @@
   "scripts": {
     "precommit": "npm run lint-staged",
     "lint-staged": "lint-staged",
-    "pretty": "prettier --write 'src/**/*.js'",
+    "prettify": "prettier --write '**/*.js'",
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
     "test": "node scripts/test.js --env=jsdom",
     "flow": "flow"
   },
   "lint-staged": {
-    "src/**/*.js": "prettier --list-different"
+    "**/*.js": "prettier --list-different"
   },
   "jest": {
     "collectCoverageFrom": [
--- a/explorer/package.json
+++ b/explorer/package.json
@@ -27,10 +27,13 @@
     "flow-bin": "^0.65.0",
     "fs-extra": "3.0.1",
     "html-webpack-plugin": "2.29.0",
+    "husky": "^0.14.3",
     "jest": "20.0.4",
+    "lint-staged": "^6.1.1",
     "object-assign": "4.1.1",
     "postcss-flexbugs-fixes": "3.2.0",
     "postcss-loader": "2.0.8",
+    "prettier": "^1.10.2",
     "promise": "8.0.1",
     "raf": "3.4.0",
     "react": "^16.2.0",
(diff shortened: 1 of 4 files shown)


wchargin:
Heh, cool. I had done the same thing, but waited to PR it until we had CI.

I'm quite wary of the pre-commit hook, because it does the wrong thing when only part of a file is staged—it will silently stage the whole file. Not good! The `precise-commits` package fixes this. Can we either

  * use `precise-commits` instead, or
  * use only a pre-commit hook that aborts the commit when any file does _not_ match Prettier's format (`prettier --list-different`), and then expect developers to either fix errors manually or install editor hooks?

One thing that I noticed is that Prettier doesn't normalize import order (probably because it can't). Should we?

My thoughts on the settings: the default settings are fine with me *except* that
  * I would really like `trailingComma: es5`. I'm convinced that trailing commas are objectively superior (two main reasons: they prevent pollution of the git diff, and they allow easily `:sort`ing and otherwise rearranging blocks). I spent a while explicitly searching for arguments _against_ them and didn't find anything substantive.
  * I have a very weak preference toward `bracketSpacing: false`, but I really don't mind either way.
  * Everything else I don't care about, though I wondered whether you preferred `arrowParens: always` to get `(x) => x` instead of `x => x`.

I think that setting changes should go into `.prettierrc`, not directly into the package.json script, so that different invocations of prettier (e.g., `prettier --list-different` in a CI) can use the same settings.

So, my concrete suggestion for this diff is `echo 'trailingComma: es5' >.prettierrc`.

(I don't mind bikeshedding a bit here, because we shouldn't have to bikeshed on these ever again.)

dandelionmane:
Ok, @wchargin PTAL. I changed it to just be a precommit error, and added `npm run pretty` as a command to prettify all the js.

notes:
- I tried out precise-commit, but it interacts badly with prettier. For example, I found that if I change one line inside a long Flow type definition, it tries to reformat while only modifying one line, resulting in reflowing the entire multiline type definition into one mega-line that breaks the line width requirements
- I force pushed and blew away the history because it was difficult to reconcile with ejecting
- I tested doing an interactive add that stages only part of a file, and it seems not to change what is staged now
- I tested that it does, indeed, error on commit when the lint is wrong
- I would love to customize the error message so that it advises that "npm run pretty" is the right solution
- For the moment, I'm only having it format .js. I figured we are planning to get rid of css, and having it format .json means it will keep trying to format data.json and I want to avoid polluting our git history with changes to a large file. (I think we should get rid of the tensorboard data anyway, and swap it to SourceCred git history once feasible)
- To make the PR easy to review, I didn't run pretty in this PR. we can do that as a followup
- I set up a .prettierrc file (.prettierrc.json, the file format is optional and I figured it adds clarity)

explorer/package.json
```diff
@@ -45,11 +48,17 @@
     "whatwg-fetch": "2.0.3"
   },
   "scripts": {
+    "precommit": "npm run lint-staged",
+    "lint-staged": "lint-staged",
+    "pretty": "prettier --write 'src/**/*.js'",
```
wchargin: It should be safe to write `'**/*.js'` here. By default, `prettier` excludes `node_modules`. You can verify this by running `yarn prettier --list-different '**/*.js'` at the command line.

wchargin: I'd slightly prefer that this be called `prettify`.

wchargin:
In the future, I hope that we can:
  - set up a CI (I think that only you can actually flip this switch?)
  - add ESLint as a commit hook
  - add both ESLint and `prettier --list-different` as blocking CI steps

:-)

explorer/package.json
wchargin: (So that it's a verb. `yarn test`, `yarn start`, `yarn prettify`.)

dandelionmane:
addressed your comments - PTAL.

it modifies flow_typed now, which is a little weird, but I think it's OK - and it will only happen (as often as we update flow_typed, which will be rarely)

dandelionmane:
Added .prettierignore

Issue #17, opened 2018-02-17
Consider using babel-plugin-flow-react-proptypes


wchargin:
Per discussion earlier: Flow and PropTypes are both great, but neither is perfect by itself. Using only Flow, we’d miss runtime errors that could arise when we read from JSON and the schema doesn’t match what we thought it was. (This has happened already, albeit benignly!) Using only PropTypes, we have less expressive types (e.g., there’s only one function type, `PropTypes.func`), and we have less integration with the rest of our Flow-typed logic.

There exists a plugin, `babel-plugin-flow-react-proptypes`, that enables you to only _write_ Flow types, and have the PropTypes generated from the Flow types at compilation time. We can’t use it directly before ejecting from `create-react-app`, but I tried it out anyway and it seems to do the job perfectly on our existing code. (See below to repro yourself.)

I think that this plugin is a good candidate, and that we should strongly consider using it after ejecting (and maybe consider ejecting in order to use it). What do you think?

cc: @dandelionmane

To try it out:

 1. Check out c1b37fa729a8d3802fe2c3a4ee4923e958ac4f8d (master, at the time of writing).
 2. Apply the patch listed below, then `yarn install`.
 3. Manually edit the file `node_modules/babel-preset-react-app/index.js`, to add the line

    ```js
    require.resolve('babel-plugin-flow-react-proptypes'),  // XXX
    ```

    at the top of the `const plugins` array.
 4. Manually edit the file `node_modules/react-scripts/config/webpack.config.dev.js`, changing the unique occurrence of `cacheDirectory: true` to `cacheDirectory: false` (may not be strictly necessary, depending on your state).
 5. Run `yarn flow`, and note that everything checks out. Run `yarn start`, and note that the app functions without error.
 6. Introduce a type error that would not cause a hard JS error in render: say, in `UserExplorer.js`, change `<UserEntry userId={author} ... />` to `<UserEntry userId={55} />`.
 7. Run `yarn flow`, and note the helpful static errors. Run `yarn start`, and note the helpful runtime errors.
 8. Note that `CI=true yarn test` is still green.

The above workflow isn’t production-ready because of the step in which you manually edit the contents of some node modules. After ejecting, these steps would be replaced by editing some config files within our repo, which is totally fine.

The patch mentioned in step 2 appears below:
<details>
<summary><strong>Contents of patch</strong> to automatically generate PropTypes from Flow types</summary>

[...]

PR #20, opened 2018-02-17
Autogenerate PropTypes from Flow types

wchargin:
Summary:
See discussion on #17.

This commit uses the `babel-plugin-flow-react-proptypes` package to
automatically create PropType definitions from components that are typed
with Flow. It simultaneously updates all of our existing components to
be typed with Flow. As a result, we have both static and dynamic type
checking.

Test Plan:
Note that `yarn test` and `yarn flow` report no errors, and that there
are no prop validation errors at runtime with `yarn start`.

Then, apply the following patch:
```diff
diff --git a/explorer/src/UserExplorer.js b/explorer/src/UserExplorer.js
index bb574cd..636a10d 100644
--- a/explorer/src/UserExplorer.js
+++ b/explorer/src/UserExplorer.js
@@ -18,7 +18,7 @@ export class UserExplorer extends Component<{
         .sort((a,b) => b[1] - a[1]);
     const entries = sortedUserWeightTuples.map(authorWeight => {
       const [author, weight] = authorWeight;
-      return <UserEntry userId={author} weight={weight} key={author}/>
+      return <UserEntry userId={55} weight={weight} key={author}/>
     });
     return <div className="user-explorer">
       <h3> User Explorer </h3>
```
Note that `yarn test` fails (the `App.test.js` E2E rendering test),
`yarn flow` fails, and there is a runtime prop validation error.

wchargin-branch: autogenerate-proptypes

### DIFF
--- a/explorer/package.json
+++ b/explorer/package.json
@@ -85,11 +85,17 @@
     ]
   },
   "babel": {
+    "plugins": [
+      "flow-react-proptypes"
+    ],
     "presets": [
       "react-app"
     ]
   },
   "eslintConfig": {
     "extends": "react-app"
+  },
+  "devDependencies": {
+    "babel-plugin-flow-react-proptypes": "^17.1.2"
   }
 }
--- a/explorer/src/App.js
+++ b/explorer/src/App.js
@@ -1,10 +1,12 @@
+// @flow
 import React, { Component } from 'react';
 import data from './data.json';
 import './App.css';
 import { FileExplorer } from './FileExplorer.js';
 import { UserExplorer } from './UserExplorer.js';
 
-class App extends Component {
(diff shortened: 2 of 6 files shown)


wchargin:
Unfortunately, this didn't catch as much as I thought it would, because of issues with the underlying babel plugin. Basically, props of the form `{[key: string]: ValueType}` weren't being typechecked at all. I've PRed those against the babel plugin and the author pushed a new version; #27 pulls those changes in, and #28 appeases the babel plugin in other ways.

It's certainly better than nothing, but I think that we shouldn't put full trust in the plugin—if we suspect that something isn't being typechecked correctly, we should verify the output in the generated bundle (it's easy to read).

PR #28, opened 2018-02-18
Avoid forward references in `commitUtils` types

wchargin:
Summary:
The resulting types less logically structured (as discussed), but while
Flow considers them equivalent, the linter does not—and, more
importantly, the runtime `PropTypes` are not generated correctly when
forward-uses are in play, due to the following issue:

https://github.com/brigand/babel-plugin-flow-react-proptypes/issues/185

I _think_ this is the lesser of two evils—but am happy to be convinced
otherwise.

Test Plan:
Note that `yarn flow` and `yarn test` are both fine, and `yarn start`
has no lint errors or runtime errors. Note further that when reverting
`insertions → added` and `deletions → deleted`, there are prop-type
errors in both `yarn test` and `yarn start`. (Without the permutation of
the types, this does not occur.)

wchargin-branch: avoid-forward-references

### DIFF
--- a/explorer/src/commitUtils.js
+++ b/explorer/src/commitUtils.js
@@ -2,21 +2,19 @@
 
 import PropTypes from "prop-types";
 
-export type CommitData = {
-  // TODO improve variable names
-  fileToCommits: {[filename: string]: string[]},
-  commits: {[commithash: string]: Commit},
+type FileStats = {
+  lines: number,
+  insertions: number,
+  deletions: number,
 };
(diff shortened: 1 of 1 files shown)


wchargin:
Interesting.

This PR was mistakenly merged into a non-master branch, and as such the commit appears nowhere in the repository history.

Neither of us noticed until now. We now have a dangling `MERGED_AS` edge in the SourceCred graph. :-)

PR #35, opened 2018-02-27
Add types for GitHub plugin.

dandelionmane:
Paired with @wchargin

### DIFF
--- a/backend/githubPlugin.js
+++ b/backend/githubPlugin.js
@@ -1,8 +1,6 @@
 // @flow
 
-import type {ID, GraphNode, GraphEdge} from "./graph";
-
-export const GITHUB_PLUGIN_NANE = "sourcecred/github-beta";
+export const GITHUB_PLUGIN_NAME = "sourcecred/github-beta";
 
 export type NodeType = "PULL_REQUEST" | "ISSUE" | "COMMENT" | "USER";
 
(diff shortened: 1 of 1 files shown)


backend/githubPlugin.js
```diff
@@ -0,0 +1,45 @@
...
+export type UserPayload = {
+  githubUserId: number,
+};
+
+export type GithubNodePayload = {
+  type: NodeType,
+  subpayload: PullRequestPayload | IssuePayload | CommentPayload | UserPayload,
+};
+
+export type EdgeType = "AUTHOR" | "REFERENCE";
```
wchargin: As discussed: inclusion and co-inclusion will both fall under the `REFERENCE` type. That is, an issue REFERENCEs its comments, and the comments also REFERENCE their enclosing issues (resp. PR). We don’t see any reason why this should cause ambiguity, because an issue cannot otherwise reference a comment: it has no body in which to do so.

We’re punting on close/reopen actions for now.

backend/githubPlugin.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+
+import type {ID, GraphNode, GraphEdge} from "./graph";
```
wchargin: Unused imports.

backend/githubPlugin.js
```diff
@@ -0,0 +1,45 @@
...
+
+export const GITHUB_PLUGIN_NANE = "sourcecred/github-beta";
```
wchargin: s/nane/name

PR #37, opened 2018-02-27
Move package json to root

dandelionmane:
Reorganize the code so that we have a single package.json file, which is at the root.
All source code now lives under `src`, separated into `src/backend` and `src/explorer`.

Test plan:

- run `yarn start` - it works
- run `yarn test` - it finds the tests (all in src/explorer) and they pass
- run `yarn flow` - it works. (tested with an error, that works too)
- run `yarn prettify` - it finds all the js files and writes to them


wchargin:
Commit hooks are worth testing! LGTM otherwise.

dandelionmane:
Tested the commit hooks, they're fine.

PR #38, opened 2018-02-27
Add conversion functions for id

wchargin:
Test Plan:
Run `yarn test` and note that tests pass.

wchargin-branch: id-conversion


wchargin:
It sure would! It sure is.

PR #43, opened 2018-03-01
Small graph improvements

dandelionmane:
Paired with @wchargin

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -1,47 +1,47 @@
 // @flow
 
-export type ID = {
-  pluginName: string,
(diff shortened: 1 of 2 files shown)


src/backend/graph.js
```diff
@@ -1,47 +1,47 @@
 // @flow
 
-export type ID = {
-  pluginName: string,
+export type Address = {
   repositoryName: string,
-  name: string,
+  pluginName: string,
```
wchargin: As discussed, we're reordering these because we think that the repository name is the broader concept. Let's also change the serialized form.

PR #44, opened 2018-03-01
Add tiny-example-repository example data

wchargin:
Summary:
The sourcecred/tiny-example-repository repository stores some example
data that we can use to generate test cases. As of now, the repository
has been archived so that its state is stable. This commit checks in the
result of our scraper on the repository.

wchargin-branch: example-data


wchargin:
Note: we use the non-pretty-printed (`python -m json.tool`) output here because we want to have the exact output of the script. We didn't see any downsides to this. (In particular, subsequent changes to this file will result in a messy diff, but won't take up any more space in git because git stores the entirety of each changed file in a commit object—not just deltas.)

PR #46, opened 2018-03-01
Rename `sourceId`/`destId` to `src`/`dst`

wchargin:
Summary:
The “ID” parts were left-over from the Great Address Migration, and we
think that abbreviations are fine here, anyway.

Test Plan:
`yarn flow && yarn test`

wchargin-branch: src-dst-rename


dandelionmane:
Great Address Migration (#43)

PR #47, opened 2018-03-01
Remove weights from the weighted graph

wchargin:
Summary:
This is an experiment. There are a couple diffferent meanings of
“weight” in play: most prominently, weights assigned by plugins versus
those suitable for comparison among other arbitrary weights. We’re not
sure what the right thing is to put in the actual graph object, so we’re
going to think about this a bit more before adding the field back in.

wchargin-branch: remove-weights

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -17,7 +17,6 @@ export type Edge<T> = {
   address: Address,
   src: Address,
   dst: Address,
-  weight: number,
   payload: T,
 };
 


wchargin:
Here is an argument that removing the `weight` field entirely is the correct decision.

Suppose that we have a “Mighty Fruit and Seafood Mix” that’s made by cooking a razorclaw crab and some mighty bananas. The dish heals 3 hearts, and provides a medium-level attack boost for 2:30. The bananas alone heal 1 heart and provide a low-level attack boost for 1:00, and the crab heals 2 hearts and provides a low-level attack boost for 1:30. The edge payloads, without weights, might look like this:

```js
const crabIngredientEdge = {
  address: ...,
  src: crab.address,
  dst: meal.address,
  payload: {
    hearts: 2,
    effect: {
      type: 'attack',
      strength: 'low',
      duration: 90,
    },
  },
};
const bananasIngredientEdge = {
  address: ...,
  src: bananas.address,
  dst: meal.address,
  payload: {
    hearts: 1,
    effect: {
      type: 'attack',
      strength: 'low',
      duration: 60,
    },
  },
};
```

If we want to ascribe weights, then we need to effectively collapse the `strength` and `duration` and `hearts` fields into one scalar value. **Claim:** It is reasonable that these parameters should be configurable at the plugin level—generically, say, in the sense that “I value hearts more than effect across all dishes”—without modifying the graph. In this case, it would make sense for plugins to provide a valuation function, with signature something like `PluginConfig × Graph × Edge → Weight`. The whole graph might be necessary if, for instance, we cooked mighty bananas together with stamella shrooms—regardless of how much we value effects, the effects cancel out because they are of different classes.

@dandelionmane: Thoughts? Is this configurability along the lines of your previous thinking?

dandelionmane:
Yep, this is exactly the sort of case I had in mind. 

Abstractly: the graph should only contain objective data, which incorporates "facts" about the world that are external to perception of the graph itself. Thus, including the number of hearts provided by a food, or a record of the reactions a user had on GitHub, or even the project significance as assigned on GitHub, are all external facts.*

However, determining how to relatively weight different types of contributions should be subjective - you and I should be able to take the same graph and arrive at different weightings, because we value different things. For example, maybe I value pragmatism relatively more and you value purism more, and so we arrive at different valuations for contributions to Haskell vs Javascript. That's a fine disagreement to have.

Keeping the objective and subjective facets separate makes our disagreement easier to explore and resolve - if we disagree, we can first check that we're both operating from the same objective fact graph, and then compare our value vectors to see where our disagreement originates. In contrast, if the weight is encoded directly in the graph, the subjectivity of our disagreement is smeared all over the graph, and it's harder to tell exactly why we have different results, or how to resolve them in a principled way. Merging the two into one data structure is a bad separation of concerns.

---

* The boundary between 'objectivity' and 'subjectivity' here is quite interesting, and the concept of 'intersubjectivity' (hat tip to Yuval Noah Harari via [Sapiens: A Brief History of Humankind](https://en.wikipedia.org/wiki/Sapiens:_A_Brief_History_of_Humankind)) is quite useful. Basically:

- Objective -> You can measure it with dumb tools. In principle, we can reach agreement about these facts, even with aliens. 

- Subjective -> Perceptually, as much a property of the viewer as the object. In general, a personal experience and you can't expect anyone else to see it the same.

- Inter-subjective -> Not a objectively real or measurable; however, it is embedded in the shared subjectivity across many minds, e.g. within a community. Therefore, attains a much greater degree of "reality" and "significance" than the subjective, while still being more flexible and mutable than the objective

Examples of the intersubjective include the United States of America, and Google - neither entity "exists" in a material way that we can measure; however, because those ideas are embedded in the beliefs and behavior patterns of so many humans, they are durably a part of the world.

There is another interesting intersection between objectivity and subjectivity, when you have an objective  record of a subjective judgment. For example, if you gave a :heart: to this comment, your subjectivity becomes objectively verifiable. 

And if a project comes together and through some consensus mechanism, they decide to value a bunch of major pull requests as being particularly important, and the results are posted in their git repository, that becomes an objective measure of an intersubjective fact.

I have a subjective hunch that continuing to keep the objective/subjective/intersubjective distinction in mind as we design the system will be quite helpful. We might imagine the process of producing cred allocations as starting with the "objective" and then moving through increasing degrees of inter-subjectivity towards final subjectivity. For example:

To start, we generate a maximally objective graph (what we're working on now). This is the shared basis for different ¤-valuations.

SourceCred tries to take a point of most-objective-intersubjectivity, which is to say we use reported intersubjectivity from projects, and otherwise try to stay neutral and put in as few value judgments as possible. This creates SourceCred¤.

However, other communities will want to make different intersubjective judgments. For example, suppose the ML community wants to have a cred flavor. They want to put a much higher weighting on ML research papers than the entire SourceCred¤ universe, and put a lower weighting on (say) the Node ecosystem. So they fork SourceCred¤ to ML¤. Fortunately, they can continue to use the same root fact graph, they just apply a different value function.

Then there might be further forks - e.g. within the ML community, there may be a group of people who are focused on AI safety, and they may want a cred of their own that puts a much higher weight on safety-related projects. So they create an "AISafety" project that records their intersubjective judgments on which contributions are most valuable to AI Safety. They can include that project in the root objective graphs, so that SourceCred¤ and ML¤ will both be awarded to AI safety contributors; however, they then fork ML¤ to AISafety¤, so that it uses the overall judgments of the ML community, except that the AI Safety subproject has much higher weight.

cc @achildatplay as this idea of forking as a mechanism for expressing different subjective values was based on conversations with him

PR #48, opened 2018-03-01
Pull in-edges and out-edges up to top-level graph

wchargin:
Summary:
The main problem with having these fields on the node is that this
presents the illusion that the API surface area is larger than it
actually is. Clients with reference to a node object could
somewhat-reasonably expect that mutating these fields would be
sufficient to update the structure of the graph, but this isn’t the case
(as the edge objects would need to be updated, too). It’s a nice
semantic bonus, too, as edges aren’t conceptually “part of” nodes.

wchargin-branch: top-level-edges

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -8,8 +8,6 @@ export type Address = {
 
 export type Node<T> = {
   address: Address,
-  inEdges: Address[],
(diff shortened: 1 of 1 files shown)


wchargin:
(Note: these will soon be pulled out to fields on a class.)

PR #54, opened 2018-03-02
Create an 'advancedMealGraph' test case

dandelionmane:
The advancedMealGraph will be a grab-all that holds all advanced and
edge behaviors, e.g. the crab-self-referential loop, and the case
where there are multiple directed edges between the same two nodes.

Aggregating them into one test case will make it easier to test more
complex behaviors, like graph merging and serialization, on the
edge case graphs. However, it's still nice to have the simple graph
so that we can test simple things too. The specific tests for edge
case behavior are left mostly unchanged, in that they start from the
simple graph and add just the advanced feature that they want to test.

### DIFF
--- a/src/backend/graph.test.js
+++ b/src/backend/graph.test.js
@@ -74,7 +74,7 @@ describe("graph", () => {
       dst: mealNode().address,
       payload: {},
     });
-    const mealGraph = () =>
+    const simpleMealGraph = () =>
       new Graph()
         .addNode(heroNode())
         .addNode(bananasNode())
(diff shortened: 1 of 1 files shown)


src/backend/graph.test.js
```diff
@@ -87,14 +87,36 @@ describe("graph", () => {
         .addEdge(crabIngredientEdge())
         .addEdge(eatEdge());
 
+    const crabLoop = () => ({
```
wchargin: `crabLoopEdge`?

src/backend/graph.test.js
```diff
@@ -87,14 +87,36 @@ describe("graph", () => {
...
+      address: makeAddress("crab-self-assessment"),
+      src: crabNode().address,
+      dst: crabNode().address,
+      payload: {evaluation: "not effective at avoiding hero"},
+    });
+
+    const critCookEdge = () => ({
```
wchargin: I'd like to see a comment/name here indicating what value this provides: namely, it tests the case where we have non-singleton hom-sets (i.e., there are multiple arrows from `a` to `b`).

src/backend/graph.test.js
```diff
@@ -87,14 +87,36 @@ describe("graph", () => {
...
+
+    const advancedMealGraph = () =>
+      simpleMealGraph()
+        .addEdge(crabLoop())
+        .addEdge(critCookEdge());
+
     describe("construction", () => {
       it("works for a simple graph", () => {
-        mealGraph();
+        simpleMealGraph();
```
wchargin: Let's check `"works on a graph with self-loops and multiple edges"`, too.

src/backend/graph.test.js
```diff
@@ -87,14 +87,36 @@ describe("graph", () => {
...
       });
 
       it("forbids adding an edge with dangling `dst`", () => {
         expect(() => {
-          mealGraph().addEdge({
+          simpleMealGraph().addEdge({
```
wchargin: How about making the "normal test cases" default to working on the advanced graph, unless there's a reason not to?

PR #61, opened 2018-03-02
Add an equality function for `Graph`

wchargin:
Summary:
We need this for testing graph equality: deep-equality is not sufficient
because two graphs can be logically equal even if, say, two nodes are
added in different orders.

This commit adds a dependency on `lodash.isequal` for deep equality.

Test Plan:
New unit tests added. Run `yarn flow && yarn test`.

wchargin-branch: graph-equals

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -1,5 +1,7 @@
 // @flow
 
+import deepEqual from "lodash.isequal";
+
 export type Address = {|
   +repositoryName: string,
   +pluginName: string,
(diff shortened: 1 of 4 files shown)


src/backend/graph.js
```diff
@@ -37,6 +39,31 @@ export class Graph {
     this._inEdges = {};
   }
 
+  equals(that: Graph): boolean {
+    function nodesSubset(a: Graph, b: Graph): boolean {
```
dandelionmane: why is `deepEqual(this._nodes, that._nodes) && deepEqual(this._edges, that._edges)` not sufficient?

wchargin: Good call: it is. Deep-diffing the full `Graph` objects is not sufficient, because `_inEdges` and `_outEdges` have undefined order. But the `_nodes` and `_edges` objects should be fine.

Issue #65, opened 2018-03-03
Choose a serialized form for graphs


wchargin:
Graphs should have a serialized form so that we can use Jest’s snapshot testing. Two options seem reasonable to me, so I implemented them.

On branch `wchargin-graph-json-arrays`, we have
```js
export type GraphJSON = {|
  +nodes: Node<mixed>[],
  +edges: Edge<mixed>[],
|};
```
which is a completely non-redundant format, but for which it is possible to have multiple JSON representations of logically-equal graphs, because the arrays are really sets and so are invariant under permutation. If we choose this method, we’ll provide an additional `toCanonicalJSON` method that sorts the `nodes` and `edges` by their addresses.

On branch `wchargin-graph-json-objects`, we have
```js
export type GraphJSON = {|
  +nodes: {[nodeAddress: string]: Node<mixed>},
  +edges: {[edgeAddress: string]: Edge<mixed>},
|};
```
which satisfies that if `a.equals(b)` then `deepEqual(a.toJSON(), b.toJSON())` at the cost of a bit of redundancy and potential for mismatch: i.e., we should require that `addressToString(json.nodes[key].address) === key` and likewise for edges, and need to guard against bad inputs. On the other hand, this form is somewhat more useful to external clients that want to directly consume the JSON.

On our `advancedMealGraph`, the latter is about 15% bigger in terms of file size:
```shell
$ filename=src/backend/__snapshots__/graph.test.js.snap
$ git show origin/wchargin-graph-json-arrays:"${filename}" | wc -c
5602
$ git show origin/wchargin-graph-json-objects:"${filename}" | wc -c
6476
```
This factor will increase as addresses get longer in proportion to the payloads, and indeed for our `advancedMealGraph` the addresses are reasonably short.

It’s worth reading the diffs at 574f916d0ecb1f167a6bd013de6f58d22625e2da (arrays) and 60771c56e257a398d23632ed64391f3df05439e0 (objects). In particular, note that using the objects form gets us free interop with Jest and `JSON.stringify`.

I’m inclined to go with the objects solution due to its simplicity in not requiring an additional canonical form. This has already paid off in the Jest and stringify interops, and it seems totally plausible that similar dividends will continue to arise down the road.

Once we come to a consensus, I can PR one of the two branches.

cc @dandelionmane

wchargin:
> This factor will increase as addresses get longer in proportion to the payloads, and indeed for our `advancedMealGraph` the addresses are reasonably short.

[...]

PR #67, opened 2018-03-05
Create an `AddressMap` abstraction

wchargin:
Summary:
This commit reifies the concept of an `Addressable`, which is any object
that has a covariant `address: Address` attribute, and implements a
simple data structure for storing addressable items keyed against their
addresses. Instances of `AddressMap` can replace the four fields of
`Graph`:
```js
_nodes: AddressMap<Node<mixed>>;
_edges: AddressMap<Edge<mixed>>;
_outEdges: AddressMap<{|+address: Address, +edges: Address[]|}>;
_inEdges: AddressMap<{|+address: Address, +edges: Address[]|}>;
```

See discussion at #65.

Test Plan:
New unit tests included, with 100% coverage: `yarn flow && yarn test`.

wchargin-branch: address-map

### DIFF
--- a/src/backend/address.js
+++ b/src/backend/address.js
@@ -1,0 +1,112 @@
+// @flow
+
+import deepEqual from "lodash.isequal";
+
+export type Address = {|
+  +repositoryName: string,
+  +pluginName: string,
+  +id: string,
+|};
+
+export interface Addressable {
+  +address: Address;
+}
+
+export type SansAddress<T: Addressable> = $Exact<$Diff<T, {+address: Address}>>;
+
+export type AddressMapJSON<T: Addressable> = {
+  [serializedAddress: string]: SansAddress<T>,
+};
+
+/**
+ * A data structure for storing addressable objects, keyed by their
+ * addresses.
+ */
+export class AddressMap<T: Addressable> {
+  // TODO(@wchargin): Evaluate performance gains from using a triple-map
+  // here. Cf. https://jsperf.com/address-string-302039074.
+  _data: {[serializedAddress: string]: T};
+
+  /**
+   * Create an empty `AddressMap`.
+   */
+  constructor() {
+    this._data = {};
+  }
+
+  /**
+   * Test whether this map logically equals another map. Two maps are
+   * logically equal if they contain the same keys and the values at
+   * each key are deep-equal.
+   */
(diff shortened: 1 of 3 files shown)


src/backend/address.js
```diff
@@ -0,0 +1,112 @@
...
+  +repositoryName: string,
+  +pluginName: string,
+  +id: string,
+|};
+
+export interface Addressable {
+  +address: Address;
+}
+
+export type SansAddress<T: Addressable> = $Exact<$Diff<T, {+address: Address}>>;
```
dandelionmane: I assume it's not possible to re-use the Addressable interface in this expression.

src/backend/address.test.js
```diff
@@ -0,0 +1,154 @@
...
+    baths: 5,
+  });
+  const fakeMansion = (): House => ({
+    // Same address, different content.
+    address: makeAddress("mansion"),
+    beds: 33,
+    baths: 88,
+  });
+  const mattressStore = (): House => ({
+    address: makeAddress("mattressStore"),
```
dandelionmane: hehe

src/backend/address.js
```diff
@@ -0,0 +1,112 @@
...
+  getAll(): T[] {
+    return Object.keys(this._data).map((k) => this._data[k]);
+  }
+}
+
+/**
+ * Create a copy of the given array and sort its elements by their
+ * addresses. The original array and its elements are not modified.
+ */
+export function sortedByAddress<T: Addressable>(xs: T[]) {
```
dandelionmane: Not entirely clear to me why this code is part of the module. Used for test code by other consumers?

dandelionmane: ok, i see how its used in the followon pr. seems a little bit like leakage of test code into main scope but not a big deal

src/backend/address.js
wchargin: Correct.

src/backend/address.js
wchargin: It is for tests, yes, but I also think that this is a reasonable place for it. It defines how to sort addresses, and lives in `address.js`. (Compare: in Java/etc. you'd have `implements Comparable<Address>` on the class; in Haskell you'd have `instance Ord` in the module.)

PR #70, opened 2018-03-06
Make App.js into skeleton for GraphExplorer

dandelionmane:
We make a very basic skeleton for the Graph Explorer as a basis
for future development.

This commit also removes the UserExplorer and FileExplorer from
App.js. Since we have changed the underlying data model, we are
unlikely to use the UserExplorer or FileExplorer in anything like
their current state, so they are effectively deprecated. I am deferring
removing them entirely because it is nice to have some examples of working
React code to copy from, before the Graph Explorer is ready.

Test plan: run `yarn start`, and observe that the App displays the
words "Graph Explorer" underneath the "SourceCred Explorer" title bar.

### DIFF
--- a/src/explorer/GraphExplorer.js
+++ b/src/explorer/GraphExplorer.js
@@ -1,0 +1,17 @@
+// @flow
+// A frontend for visualizing Contribution Graphs
+
+import React, {Component} from "react";
+
+type Props = {};
+type State = {};
+
+export class GraphExplorer extends Component<Props, State> {
+  render() {
+    return (
+      <div>
+        <h1> Graph Explorer </h1>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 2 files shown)


src/explorer/GraphExplorer.js
```diff
@@ -0,0 +1,17 @@
...
+import React, {Component} from "react";
+
+type Props = {};
+type State = {};
+
+export class GraphExplorer extends Component<Props, State> {
+  render() {
+    return (
+      <div>
+        <h1> Graph Explorer </h1>
```
wchargin: By universal convention, this should be `<h1>Graph Explorer</h1>`: no spaces inside the angle brackets. (Prettier won't change this because it's technically different HTML output, though it's rendered the same way.)

PR #71, opened 2018-03-06
Factor eventide graph demo data to a new module

dandelionmane:
It would be helpful to make our standard tiny graph available to other
test and demo instances, outside of just graph.test.js. This way we can
use it as a test case for the Graph Explorer.

### DIFF
--- a/src/backend/graph.test.js
+++ b/src/backend/graph.test.js
@@ -3,6 +3,23 @@
 import type {Address, Addressable} from "./address";
 import {sortedByAddress} from "./address";
 import {Graph} from "./graph";
+import {
+  makeAddress,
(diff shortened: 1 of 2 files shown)


src/backend/graph.test.js
```diff
@@ -3,6 +3,23 @@
 import type {Address, Addressable} from "./address";
 import {sortedByAddress} from "./address";
 import {Graph} from "./graph";
+import {
```
wchargin: Personally, I'd write `import * as demoData from "./graphDemoData";` instead of this giant list, and then use `demoData.makeAddress` (etc.) below. But it's up to you.

Issue #76, opened 2018-03-15
Create a structured form for GraphQL queries


wchargin:
We will need to be able to structurally manipulate GraphQL queries so
that we can implement general-purpose pagination against the GitHub API.
Amazingly, I wasn’t able to find such a library after some searching. In
particular, the official `graphql` npm package has structured schemata
(of course), but only stringly-typed queries, as far as I can tell.

Happily, GraphQL has a proper spec:
https://facebook.github.io/graphql/October2016/

This makes it easy to correctly define the relevant API. Our needs are
very simple: we just need an ADT for a GraphQL request (queries and
fragment definitions). The Flow-level types should express all the
complexity of the system, and there will be a projection function to
take a query to a GraphQL-formatted string. Some helper functions will
assist creation of the types for a fluent API, but this should basically
be it.

cc @dandelionmane

dandelionmane:
This looks nice to me, although I don't fully understand the motivation behind it. I assume this gets us DRY-er code and, more importantly, a basis to handle pagination and retries in a principled fashion.

A few thoughts:

- Would it be an unreasonable metaprogramming ask to overload this structure so that it can both construct the appropriate graphQL query, and the flow type signature that the query will return? Then we could commit the generated typesignature in its own file, but have a test that verifies it is identical to what would be generated if we re-ran the generation function. It seems this could be accomplished by having your code be aware of the type expected for each field.

- Is this sufficiently generic that we should think about getting it adopted back into the reference implementation, or similar? (Maybe would consume a lot of rabbit hole budget.)

- We should factor all of the (first N) type arguments into constants, since we'll want to make those easy to tune.

wchargin:
> (1)

It wouldn't be _unreasonable_ metaprogramming, in the sense that I could implement that in zero time I would absolutely do so. But I definitely think that it's out of the rabbit hole budget for right now. You say "having the code be aware of the type _expected_ for each field"—emphasis mine; I think we probably want something stronger than expectation. We should do something 

> (2)

Yes, I think so. The only major thing that I left out of the spec was directives, because we don't plan to use them, but they could be trivially added back in. I don't want to do this right now, though, not least because I want flexibility and control over the API. (There are already things that I want to change.)

> (3)

(I think that you are talking about our particular query for sourcecred rather than a generic API.) I am happy to consider doing that. Note that, now that we have a DSL, we can factor them into GraphQL query parameters or into JavaScript constants, as desired.

dandelionmane:
[...]

PR #77, opened 2018-03-15
Add a GraphQL structured query format

wchargin:
Summary:
See motivation in #76. Feel free to look at the new snapshot file to
inspect the structured representation and also the stringified output.

This implementation is sufficient to encode our query against the
GitHub v4 API; see the test plan below.

Test Plan:
Unit tests added; run `yarn flow && yarn test`.

This code has full coverage except for lines 247, 302, and 363 of
`queries.js`; these lines check invariants that should never be
violated.

You can also use the following steps to verify that the sample query is
valid GraphQL that produces the same results as our hand-written query:

 1. Apply the following hacky patch:

    ```diff
    diff --git a/src/backend/graphql/queries.test.js b/src/backend/graphql/queries.test.js
    index 52bdec7..c04a636 100644
    --- a/src/backend/graphql/queries.test.js
    +++ b/src/backend/graphql/queries.test.js
    @@ -3,6 +3,18 @@
     import type {Body} from "./queries";
     import {build, stringify, multilineLayout, inlineLayout} from "./queries";

    +function emitGitHubQuery(layout, filename) {
    +  const fs = require("fs");
    +  const path = require("path");
    +  const result = stringify.body(usefulQuery(), layout);
    +  const outputFilepath = path.join(__dirname, "..", filename);
    +  const outputText = `module.exports = ${JSON.stringify(result)};\n`;
    +  fs.writeFileSync(outputFilepath, outputText);
    +  console.log(`Wrote output to ${outputFilepath}.`);
    +}
    +emitGitHubQuery(multilineLayout("  "), "githubQueryMultiline.js");
    +emitGitHubQuery(inlineLayout(), "githubQueryInline.js");
    +
     describe("queries", () => {
       describe("end-to-end-test cases", () => {
         const testCases = {
    ```

 2. Run `CI=true yarn test`, and verify that the following two files
    written to `src/backend/` contain appropriate contents. You can just
    eyeball them, or check that they match my results:
    https://gist.github.com/wchargin/f37b99fd4ec345c9d2541c2dc53ceda9

 3. In `fetchGitHubRepo.js`, change the definition of `const query` to

    ```js
    const query = require("./githubQueryMultiline.js");
    ```

    Run

    ```shell
    GITHUB_TOKEN="<your_token_here>" src/backend/fetchGitHubRepoTest.sh
    ```

    and verify that it exits successfully.

 4. Repeat for `require("./githubQueryInline.js")`.

wchargin-branch: graphql-structured-queries

### DIFF
--- a/src/backend/graphql/queries.js
+++ b/src/backend/graphql/queries.js
@@ -1,0 +1,398 @@
+// @flow
+
+/**
+ * GraphQL structured query data format.
+ *
+ * Main module exports:
+ *   - lots of types for various GraphQL language constructs
+ *   - the `build` object, providing a fluent builder API
+ *   - the `stringify` object, and particularly `stringify.body`
+ *   - the two layout strategies `multilineLayout` and `inlineLayout`
+ */
+
+export type Body = Definition[];
+
+export type Definition = QueryDefinition | FragmentDefinition;
+
+// We only need opaque type handles; no need to embed the GraphQL type
+// system into Flow.
+export type GraphQLType = string;
+
+export type QueryDefinition = {|
+  +type: "QUERY",
+  +name: string,
+  +params: Parameter[],
+  +selections: Selection[],
+|};
+export type Parameter = {|+name: string, +type: GraphQLType|};
+
+export type FragmentDefinition = {|
+  +type: "FRAGMENT",
+  +name: string,
+  +typeCondition: GraphQLType,
+  +selections: Selection[],
+|};
+
+export type Selection = Field | FragmentSpread | InlineFragment;
+export type Field = {|
+  +type: "FIELD",
+  +name: string,
+  +args: Arguments,
+  +selections: Selection[],
+|};
+export type FragmentSpread = {|
+  +type: "FRAGMENT_SPREAD",
+  +fragmentName: string,
+|};
+export type InlineFragment = {|
+  +type: "INLINE_FRAGMENT",
+  +typeCondition: ?GraphQLType,
+  +selections: Selection[],
+|};
+
+export type Arguments = {[string]: Value};
+export type Value =
+  | VariableValue
+  | LiteralValue
+  | EnumValue
+  | ListValue
+  | ObjectValue;
+export type VariableValue = {|+type: "VARIABLE", +data: string|};
+export type LiteralValue = {|
+  +type: "LITERAL",
+  +data: number | string | boolean | null,
+|};
+export type EnumValue = {|+type: "ENUM", +data: string|};
+export type ListValue = {|+type: "LIST", +data: Value[]|};
+export type ObjectValue = {|+type: "OBJECT", +data: {[string]: Value}|};
+
+export const build = {
+  query(
+    name: string,
+    params: Parameter[],
+    selections: Selection[]
+  ): QueryDefinition {
+    return {
+      type: "QUERY",
+      name,
+      params,
+      selections,
+    };
+  },
+
+  param(name: string, type: GraphQLType): Parameter {
+    return {name, type};
+  },
+
+  fragment(
+    name: string,
+    typeCondition: GraphQLType,
+    selections: Selection[]
+  ): FragmentDefinition {
+    return {
+      type: "FRAGMENT",
+      name,
+      typeCondition,
+      selections,
+    };
+  },
+
+  field(name: string, args: ?Arguments, selections: ?(Selection[])): Field {
+    return {
+      type: "FIELD",
+      name,
+      args: args || {},
+      selections: selections || [],
+    };
+  },
+
+  fragmentSpread(fragmentName: string): FragmentSpread {
+    return {
+      type: "FRAGMENT_SPREAD",
+      fragmentName,
+    };
+  },
+
+  inlineFragment(
+    typeCondition: ?GraphQLType,
+    selections: Selection[]
+  ): InlineFragment {
+    return {
+      type: "INLINE_FRAGMENT",
+      typeCondition,
+      selections,
+    };
+  },
+
+  variable(name: string): VariableValue {
+    return {
+      type: "VARIABLE",
+      data: name,
+    };
+  },
+
+  literal(value: number | string | boolean | null): LiteralValue {
+    return {
+      type: "LITERAL",
+      data: value,
+    };
+  },
+
+  enumLiteral(value: string): EnumValue {
+    return {
+      type: "ENUM",
+      data: value,
+    };
+  },
+
+  list(data: Value[]): ListValue {
+    return {
+      type: "LIST",
+      data: data,
+    };
+  },
+
+  object(data: {[string]: Value}): ObjectValue {
+    return {
+      type: "OBJECT",
+      data: data,
+    };
+  },
+};
+
+/**
+ * A strategy for stringifying a sequence of GraphQL language tokens.
+ */
+export interface LayoutStrategy {
+  // Lay out a group of tokens that should be treated atomically.
+  atom(line: string): string;
+
+  // Join groups of tokens. The elements should be the results of calls
+  // to `atom` (or other `join`s).
+  join(lines: string[]): string;
+
+  // Get a strategy for the next level of nesting. For instance, if this
+  // object lays out its result over multiple lines, then this method
+  // might produce a strategy with one extra level of indentation.
+  next(): LayoutStrategy;
+}
+
+/**
+ * Create a layout strategy that lays out text over multiple lines,
+ * indenting with the given tab string (such as "\t" or "  ").
+ */
+export function multilineLayout(tab: string): LayoutStrategy {
+  function strategy(indentLevel: number) {
+    return {
+      atom: (line: string) => {
+        const indentation = Array(indentLevel)
+          .fill(tab)
+          .join("");
+        return indentation + line;
+      },
+      join: (xs: string[]) => xs.join("\n"),
+      next: () => strategy(indentLevel + 1),
+    };
+  }
+  return strategy(0);
+}
+
+/**
+ * Create a layout strategy that lays out all text on one line.
+ */
+export function inlineLayout(): LayoutStrategy {
+  const result = {
+    atom: (line) => line,
+    join: (xs) => xs.join(" "),
+    next: () => result,
+  };
+  return result;
+}
+
+/*
+ * Map a stringification function across a list, and join the results
+ * with a formatter.
+ */
+function formatList<T>(
+  values: T[],
+  subformatter: (value: T, ls: LayoutStrategy) => string,
+  ls: LayoutStrategy
+): string {
+  return ls.join(values.map((x) => subformatter(x, ls)));
+}
+
+/*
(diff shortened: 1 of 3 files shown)


wchargin:
Note: The above auxiliary test plan was written when the test script still canonicalized output, but this was removed in #75. Without canonicalization, it trivially fails. I vote to reinstate the canonicalization in the test script, even though it's redundant with `json-stable-stringify` for one particular data source. What do you think?

src/backend/graphql/queries.js
```diff
@@ -0,0 +1,411 @@
...
+  fmt: LineFormatter
+): string {
+  function formatKey(k: string, fmt: LineFormatter): string {
+    return fmt.join([fmt.format(`${k}:`), subformatter(object[k], fmt.next())]);
+  }
+  return formatList(Object.keys(object), formatKey, fmt);
+}
+
+export const stringify = {
+  body(body: Body, fmt: LineFormatter): string {
```
dandelionmane: There's inconsistent usage of "stringify" vs "format" across this file. Maybe just standardize on "format", where "format" means "turn this thing into a string, using a provided LineFormatter". Then this const can be called "formats".

Possibly it would be relevant to define the type `Format<T>: function(T, LineFormatter) => string`.
Then `formats` is literally a collection of `Format` implementations

dandelionmane: Alternatively, I think using "stringify" as the consistent verb and "stringifiers" as the object containing many stringify functions may be even clearer, as it removes ambiguity with the two LineFormatters.

wchargin:
> Regarding the canonicalization issue - I don't understand why the test script will require canonicalization going forward. The responsibility of the test script is to test that one piece we won't put into standard CI, which is hitting Github's live API with an API key.

On reflection, that makes sense. My auxiliary test plan is auxiliary anyway. I can deal with it. :-)

> There's inconsistent usage of "stringify" vs "format" across this file. Maybe just standardize on "format", where "format" means "turn this thing into a string, using a provided LineFormatter". Then this const can be called "formats".

As I intended to structure it, `stringify` and `format` have different meanings. A stringifier is specific to a type, and knows how to convert instances of that type to strings. So, a `Stringifier<T>` is a function `(T, LineFormatter) => string`—and I’m happy to reify this.

A formatter, on the other hand, describes how to lay out arbitrary strings; it does not care whether those strings came from a structured form or anywhere else. It is essentially what’s described in the `LineFormatter` interface, but I agree that the names are not great, especially when the reader is forced to contrast them with the stringifier. To discover the right names, perhaps we should make the abstraction more precise. I took some liberties with the typing, but the underlying abstraction is a mechanism for combining _token groups_, with the following signature:

  - an opaque type `TokenRepresentation`
  - an injection `atom : self -> string -> TokenRepresentation`
  - a reduction `join : self -> TokenRepresentation * TokenRepresentation -> TokenRepresentation`
  - a projection `print : self -> TokenRepresentation -> string`
  - (an endomorphism `next : self -> self`, but this is not relevant right now)

In this implementation I’ve let `TokenRepresentation` be literally the type `string`. Then, `atom` is `format`, and `print` is the identity function. However, this means that it is well-typed to `join` strings that are not the result of `format`, or to `format` something twice; neither of these should be possible. Given that this interface is internal, I opted to take the simpler version with fewer coercions, because I can verify that I myself use it correctly throughout this file.

So… all this to say that I agree that the names on `LineFormatter` are somewhat confusing, especially when `stringify` is also in the mix. Maybe it’s worth making the notion of token grouping more explicit, and using the names above. I think that that might be less confusing. I’ll give it a shot and see what I think.

If we’re bikeshedding on `stringify` vs. `format` for the `T -> string` thing, then I prefer `stringify` because it’s what `JSON.stringify` uses, but if you want to use `format` then I am happy to change it.

> Alternatively, I think using "stringify" as the consistent verb and "stringifiers" as the object containing many stringify functions may be even clearer, as it removes ambiguity with the two LineFormatters.

Dropping the part of this that we’ve already addressed, and regarding only `s/const stringify/const stringifiers`: I like this suggestion because then `builders` and `stringifiers`, which are structurally the same (a dictionary of actions) are grammatically consistent. The other nice option is `export const build = { ... }` and `export const stringify = { ... }`; personally, I like this even more, because they read as verbs, and so `build.field(...)` and `stringify.field(...)` seem very clear to me.

wchargin:
While I’m at it, things I want to change (future PRs):
  - The time system should enforce that lists are homogeneously typed. (This is possible.)
  - The type system should separate out the `Literal` value types into the individual value types, and in particular should separate `Int` and `Float`, even though JS doesn’t.
  - There should be validation, possibly optional. Things to be validated include: names must match `/[A-Za-z_][A-Za-z0-9_]*/`, `int` literals must have zero fractional part; variables must be declared; fragments must be declared.
  - The type system should be more consistent between `{[string]: T}` and `{name: string, value: T}`. Examples: `Arguments` and `ObjectValue.data` for the former; `Parameter[]` for the latter. I’m not sure which one I prefer; I’ll use the API a bit more first.

dandelionmane:
A lot of the "future PR work" items make more sense if we factor the GraphQL structural query generator out of sourcecred/sourcecred and into its own project. The level of polish you describe is probably over-engineering if it's only scratching our own itch, but is worth it if we make it generally available.

It seems very node-ish of us to factor this out as a separate npm dependency relatively early, I'm fine with doing that if you're enthused.

wchargin:
> Maybe it’s worth making the notion of token grouping more explicit, and using the names above. I think that that might be less confusing. I’ll give it a shot and see what I think.

So… I looked into this. At length. It’s possible, but typing it correctly requires that `inlineFormatter` and `multilineFormatter` and `LineFormatter` each be declared in separate files. (Why? I had foolishly assumed that Flow’s “existential type” represented an existential type, but it does not; instead, it just means “please automatically infer this type and expose it as if I had written it myself”, which is worlds different. Flow does offer opaque types, but only allows scoping the opacity at a module level. Way better than nothing, but a lot of unnecessary cruft.) You can see the result here: https://gist.github.com/wchargin/e890f0cbae10a2d1d32eefd54f5011a5

The types are stronger… but at what cost? I like the implementation code more (especially with the explicit projections in `queryDefinition` and `field`). But the user interface is less fluent (see test code). Really, `Stringifier<E>` should be an opaque type, and users should only see a function `stringify: <E, T>(stringifier: Stringifier<E>, element: E, fmt: LineFormatter<T>): string` whose implementation is `fmt.project(stringifier(element, fmt))`. Then users would call `stringify(stringifiers.body, myBody, multilineFormatter("  "))`. Is this the right way to go? The precision of the types makes it easier for a user to understand what’s _actually_ going on, but on the other hand it would be nice if there were nothing _to_ understand, and the former API (as originally in this PR) does a better job of that.

Honestly, I don’t really want to spend any more time on the fine details of the stringification API right now. As long as it’s internal to our repo, it’s okay if it’s imperfect. There are better things for me to be working on.

I think that I will keep the API as originally in this PR, but make the following name changes:
  - `builders` becomes `build`;
  - `LineFormatter` becomes `LayoutStrategy`;
  - `LineFormatter.format` becomes `LayoutStrategy.atom`;
  - `multilineFormatter` becomes `multilineLayout`;
  - `inlineFormatter` becomes `inlineLayout`.
In my opinion, these changes are sufficient to avoid API confusion. Then I can work on more valuable things. :-)

> A lot of the "future PR work" items make more sense if we factor the GraphQL structural query generator out of sourcecred/sourcecred and into its own project. The level of polish you describe is probably over-engineering if it's only scratching our own itch, but is worth it if we make it generally available.

Hmm; I think I mostly disagree. I agree for the point about validation, because that requires structuring the API such that validation can be disabled for performance or enabled with good error messages, etc. But the first two changes should be trivial changes to the type system that give us a fair amount of safety for just a few minutes’ worth of effort, and the last change is a trivial change to the type system that avoids having two standards for a simple concept. Perhaps I can convince you by sending you tiny PRs!

> It seems very node-ish of us to factor this out as a separate npm dependency relatively early, I'm fine with doing that if you're enthused.

I am not enthused about releasing in any form an API with one client (a single test case) and a total lifespan of two days—even though doing so would be very node-ish! But we can certainly keep it separated in our codebase: say, in the `graphql` directory, and without any `import`s from the rest of our code. Then, actually publishing it at any point will be easy.

wchargin:
Updated PR; please take a look at your convenience.

dandelionmane:
>> A lot of the "future PR work" items make more sense if we factor the GraphQL structural query generator out of sourcecred/sourcecred and into its own project. The level of polish you describe is probably over-engineering if it's only scratching our own itch, but is worth it if we make it generally available.

> Hmm; I think I mostly disagree. I agree for the point about validation, because that requires structuring the API such that validation can be disabled for performance or enabled with good error messages, etc. But the first two changes should be trivial changes to the type system that give us a fair amount of safety for just a few minutes’ worth of effort, and the last change is a trivial change to the type system that avoids having two standards for a simple concept. Perhaps I can convince you by sending you tiny PRs!

If you send me tiny PRs, I will be convinced.

>> It seems very node-ish of us to factor this out as a separate npm dependency relatively early, I'm fine with doing that if you're enthused.

> I am not enthused about releasing in any form an API with one client (a single test case) and a total lifespan of two days—even though doing so would be very node-ish! But we can certainly keep it separated in our codebase: say, in the graphql directory, and without any imports from the rest of our code. Then, actually publishing it at any point will be easy.

In general, there's merit to releasing early, since the longer something goes unreleased, the more time it has for natural software entropy to entangle it with the rest of the system, so that releasing it becomes difficult. However, as you say, keeping it explicitly separated in our codebase provides a good defense. Maybe when we implement pagination in a general purpose way, that will be a good time to release it (we can also then link to it from the github issues we've found where other people complain about the lack of an easy solution for pagination).

Overall: Code LGTM to me, thanks for this work! As usual, I'm impressed with how thorough and principled an approach you take.

wchargin:
Regarding those auxiliary features: I’ve decided not to implement them for now, for the following reasons. Items (1) and (2) involve strengthening the type system to be more akin to GraphQL’s type system, but this is currently a non-goal for me: I’m trying to represent the query language only. Item (3), as discussed above, is overengineering at this point. Item (4) requires more thought/experience with the API and isn’t a big deal right now.

> In general, there's merit to releasing early, since the longer something goes unreleased, the more time it has for natural software entropy to entangle it with the rest of the system, so that releasing it becomes difficult. However, as you say, keeping it explicitly separated in our codebase provides a good defense.

Agreed with the first sentence, and I do think that keeping it separated will be sufficient to make it _easy_ to extricate. Even if it does become a little tangled, it shouldn’t be hard to untangle. And, if it is hard to untangle without compromising SourceCred functionality, then it will have been a good thing that it remained in SourceCred. :-)

> Maybe when we implement pagination in a general purpose way, that will be a good time to release it (we can also then link to it from the github issues we've found where other people complain about the lack of an easy solution for pagination).

This sounds reasonable.

> Overall: Code LGTM to me, thanks for this work! As usual, I'm impressed with how thorough and principled an approach you take.

You’re welcome, and thanks!

PR #79, opened 2018-03-16
Allow redundant adds to the Graph

dandelionmane:
Graph.addNode and Graph.addEdge now allow adding the same node or edge
multiple times, provided that the duplicate adds are trying to insert
identical content.

This came up while prototyping the GitHub plugin; rather than create
myriad subgraphs and merge them, I found it convenient to construct a
single graph and iteratively add nodes. Since the same node may be
discovered multiple times (most notably user identities), there was a
need for a "conservative add" abstraction that adds a node if it doesn't
exist yet, but errors only if multiple adds conflict.

Since this behavior is generic and highly conservative, it seemed
appropriate to include in the graph class itself.

Test Plan:
The unit tests have been updated to include the new behavior.

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -70,10 +70,17 @@ export class Graph {
     if (node == null) {
       throw new Error(`node is ${String(node)}`);
     }
-    if (this.getNode(node.address) !== undefined) {
-      throw new Error(
-        `node at address ${JSON.stringify(node.address)} already exists`
-      );
+    const existingNode = this.getNode(node.address);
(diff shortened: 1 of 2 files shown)


wchargin:
I am fine with including this functionality, and I am also fine with making it the default behavior of `addNode` and `addEdge`, as you have done. However, I am surprised that you prefer to imperatively construct a megagraph instead of constructing and merging many small graphs. (I did the latter, and thought that it was very elegant.) I suppose that I will understand better when I see the relevant plugin code!

PR #80, opened 2018-03-16
Allow adding explicitly typed nodes/edges

dandelionmane:
Summary:
Flow doesn’t allow us to specify variance annotations in generic
function parameters, and doesn’t allow coercing `Node<T>` to
`Node<mixed>`. This forces us to put `any`s in our code, which…works.

Paired with @wchargin

Test Plan:
New unit tests trivially pass dynamically, and now pass statically
(failing before the changes to `graph.js`).

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -17,13 +17,13 @@ export type Edge<T> = {|
 |};
 
 export type GraphJSON = {|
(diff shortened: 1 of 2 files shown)


wchargin:
Note to self: if you accidentally restore a branch, wait until the newly created Travis build completes before subsequently deleting it again. Otherwise, you get a "failed to clone from GitHub" build failure.

Issue #66, opened 2018-03-03
Consider removing Address↔string conversion functions


wchargin:
Consider removing address↔string conversion functions

**understandabad** (*adj.*): A seemingly confusing decision that makes sense only once you realize the implementation details behind a leaky abstraction.

*ex.* Our `Address` type requires that none of its components contain a literal dollar sign. Why? Because the `addressToString` and `stringToAddress` conversions use the dollar sign as a delimiter for the three components, and string forms are used as keys in the `Graph`’s maps.

We could relatively easily allow arbitrary addresses by using nested maps:

```js
type AddressMap<T> = {[repositoryName: string]: {[pluginName: string]: {[id: string]: T}}};
class Graph {
  nodes: AddressMap<Node<mixed>>;
  edges: AddressMap<Edge<mixed>>;
  /* ... */
}
/* private */ function addressMapGet<T>(map: AddressMap<T>, key: Address): ?T {
  const repositoryBucket = map[key.repositoryName];
  if (repositoryBucket === undefined)
    return undefined;
  const pluginNameBucket = repositoryBucket[key.pluginName];
  if (pluginNameBucket === undefined)
    return undefined;
  return pluginNameBucket[key.id];
}
/* private */ function addressMapSet<T>(map: AddressMap<T>, key: Address, value: T) {
  let repositoryBucket = map[key.repositoryName];
  if (repositoryBucket === undefined)
    repositoryBucket = map[key.repositoryName] = {};
  let pluginNameBucket = repositoryBucket[key.pluginNameName];
  if (pluginNameBucket === undefined)
    pluginNameBucket = repositoryBucket[key.pluginNameName] = {};
  pluginNameBucket[key.id] = value;
}
```
This would be a completely under-the-hood implementation change, and would remove the `$`-restriction. It may well be a faster implementation, too.

dandelionmane:
What if we replace `addressToString(address)` with `JSON.stringify(address)`?
We get the same benefits, and perhaps more (going to add a comment on #65 going into more details). And it's simpler than both your suggestion, and the current implementation.

dandelionmane:
Since it's just an implementation detail, favoring the more-complex approach over JSON.stringify seems like premature optimization; we can revive the triplemap when/if JSON.stringify shows up on our profilers.

dandelionmane:
Also, understandabad: 👍

dandelionmane:
I decided to test it.

Test 1:
```js
function addressMapGet(map, key) {
[...]

Issue #40, opened 2018-02-28
Plugins: GitHub plugin


decentralion:
The GitHub plugin will be responsible for sourcing the following types of information:

Nodes:
- Pull requests
- Issues
- Issue comments
  n.b. the body of an issue (or PR) is treated as its first "issue comment"
- Commit comments
- Review comments
- GitHub user identities

Edges:
- authorship 
  users author the other nodes, e.g. user authors commit
- containership 
  issues and pull requests contain comments
  *(might merge with reference below)*
- reference 
  comments can reference other nodes, e.g. the line below where this issue comment references PR-35

This is based off the overall plugin architecture design (#26); implementation has started in 'Add types for GitHub plugin' (#35)

decentralion:
The GitHub plugin is pretty well established! We have some open feature issues (e..g #131, #132, #148, #171) but the plugin as a whole is working. :)

PR #81, opened 2018-03-16
[WIP] GitHub Plugin: Parse issues, comments, and users

dandelionmane:
[WIP] GitHub Plugin: Parse issues, comments, users

This commit adds a GithubParser class to the GitHub plugin.
The GithubParser currently can parse issues, creating nodes for issues,
comments, and users, and authorship and containment edges between them.

It does not parse pull requests, and it does not create reference edges.

Test plan:
Carefully verify that the snapshots encode appropriate behavior
(Not yet done)

Partially paired with @wchargin

### DIFF
--- a/src/backend/githubPlugin.test.js
+++ b/src/backend/githubPlugin.test.js
@@ -1,0 +1,84 @@
+// @flow
+
+import {GithubParser, getNodeType, getEdgeType} from "./githubPlugin";
+import exampleRepoData from "./githubDemoData/example-repo.json";
+
+describe("GithubParser", () => {
+  describe("whole repo parsing", () => {
+    const parser = new GithubParser("sourcecred/example-repo");
+    parser.addData(exampleRepoData.data);
+    const graph = parser.graph;
+
+    it("parses the entire example-repo as expected", () => {
+      expect(graph).toMatchSnapshot();
+    });
+
+    it("every comment has an author and container", () => {
+      const comments = graph
+        .getAllNodes()
+        .filter((n) => getNodeType(n) === "COMMENT");
+      expect(comments).not.toHaveLength(0);
+      comments.forEach((c) => {
+        const authorEdges = graph
+          .getOutEdges(c.address)
+          .filter((e) => getEdgeType(e) === "AUTHORSHIP");
+        expect(authorEdges.length).toBe(1);
+        const containerEdges = graph
+          .getInEdges(c.address)
+          .filter((e) => getEdgeType(e) === "CONTAINMENT");
+        expect(containerEdges.length).toBe(1);
+      });
+    });
+
+    it("every pull request and issue has an author", () => {
+      const issuesAndPRs = graph
+        .getAllNodes()
+        .filter(
+          (n) => ["ISSUE", "PULL_REQUEST"].indexOf(getNodeType(n)) !== -1
+        );
+      expect(issuesAndPRs).not.toHaveLength(0);
+      issuesAndPRs.forEach((x) => {
+        const outEdges = graph.getOutEdges(x.address);
+        const authorEdges = outEdges.filter(
+          (e) => getEdgeType(e) === "AUTHORSHIP"
+        );
+        expect(authorEdges.length).toBe(1);
+      });
+    });
+  });
+
+  describe("issue parsing", () => {
+    it("parses a simple issue (https://github.com/sourcecred/example-repo/issues/1)", () => {
+      const issue1 = exampleRepoData.data.repository.issues.nodes[0];
+      expect(issue1.number).toBe(1);
+      const parser = new GithubParser("sourcecred/example-repo");
+      parser.addIssue(issue1);
+      expect(parser.graph).toMatchSnapshot();
+    });
+
+    it("parses an issue with comments (https://github.com/sourcecred/example-repo/issues/6)", () => {
+      const issue6 = exampleRepoData.data.repository.issues.nodes[3];
+      expect(issue6.number).toBe(6);
+      const parser = new GithubParser("sourcecred/example-repo");
+      parser.addIssue(issue6);
+      expect(parser.graph).toMatchSnapshot();
+    });
+  });
+
+  describe("pull request parsing", () => {
+    it("parses a simple pull request (https://github.com/sourcecred/example-repo/pull/3)", () => {
+      const pr3 = exampleRepoData.data.repository.pullRequests.nodes[0];
+      expect(pr3.number).toBe(3);
+      const parser = new GithubParser("sourcecred/example-repo");
+      parser.addPullRequest(pr3);
+      expect(parser.graph).toMatchSnapshot();
+    });
+    it("parses a pr with review comments (https://github.com/sourcecred/example-repo/pull/3)", () => {
+      const pr5 = exampleRepoData.data.repository.pullRequests.nodes[1];
+      expect(pr5.number).toBe(5);
+      const parser = new GithubParser("sourcecred/example-repo");
+      parser.addPullRequest(pr5);
+      expect(parser.graph).toMatchSnapshot();
+    });
+  });
+});
--- a/src/backend/__snapshots__/githubPlugin.test.js.snap
+++ b/src/backend/__snapshots__/githubPlugin.test.js.snap
@@ -1,0 +1,536 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`GithubParser issue parsing parses a simple issue (https://github.com/sourcecred/example-repo/issues/1) 1`] = `
+Object {
+  "edges": Object {
+    "{\\"pluginName\\":\\"sourcecred/github-beta\\",\\"repositoryName\\":\\"sourcecred/example-repo\\",\\"id\\":\\"{\\\\\\"authorID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDQ6VXNlcjE0MDAwMjM=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"USER\\\\\\"},\\\\\\"contributionID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDU6SXNzdWUzMDA5MzQ4MTg=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"ISSUE\\\\\\"},\\\\\\"type\\\\\\":\\\\\\"AUTHORSHIP\\\\\\"}\\"}": Object {
+      "dst": Object {
+        "id": "{\\"id\\":\\"MDQ6VXNlcjE0MDAwMjM=\\",\\"type\\":\\"USER\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+      "payload": Object {},
+      "src": Object {
+        "id": "{\\"id\\":\\"MDU6SXNzdWUzMDA5MzQ4MTg=\\",\\"type\\":\\"ISSUE\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+    },
+  },
+  "nodes": Object {
+    "{\\"pluginName\\":\\"sourcecred/github-beta\\",\\"repositoryName\\":\\"sourcecred/example-repo\\",\\"id\\":\\"{\\\\\\"id\\\\\\":\\\\\\"MDQ6VXNlcjE0MDAwMjM=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"USER\\\\\\"}\\"}": Object {
+      "payload": Object {
+        "login": "dandelionmane",
+      },
+    },
+    "{\\"pluginName\\":\\"sourcecred/github-beta\\",\\"repositoryName\\":\\"sourcecred/example-repo\\",\\"id\\":\\"{\\\\\\"id\\\\\\":\\\\\\"MDU6SXNzdWUzMDA5MzQ4MTg=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"ISSUE\\\\\\"}\\"}": Object {
+      "payload": Object {
+        "body": "This is just an example issue.",
+        "number": 1,
+        "title": "An example issue.",
+      },
+    },
+  },
+}
+`;
+
+exports[`GithubParser issue parsing parses an issue with comments (https://github.com/sourcecred/example-repo/issues/6) 1`] = `
+Object {
+  "edges": Object {
+    "{\\"pluginName\\":\\"sourcecred/github-beta\\",\\"repositoryName\\":\\"sourcecred/example-repo\\",\\"id\\":\\"{\\\\\\"authorID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDQ6VXNlcjE0MDAwMjM=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"USER\\\\\\"},\\\\\\"contributionID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDEyOklzc3VlQ29tbWVudDM3Mzc2ODQ0Mg==\\\\\\",\\\\\\"type\\\\\\":\\\\\\"COMMENT\\\\\\"},\\\\\\"type\\\\\\":\\\\\\"AUTHORSHIP\\\\\\"}\\"}": Object {
+      "dst": Object {
+        "id": "{\\"id\\":\\"MDQ6VXNlcjE0MDAwMjM=\\",\\"type\\":\\"USER\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+      "payload": Object {},
+      "src": Object {
+        "id": "{\\"id\\":\\"MDEyOklzc3VlQ29tbWVudDM3Mzc2ODQ0Mg==\\",\\"type\\":\\"COMMENT\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+    },
+    "{\\"pluginName\\":\\"sourcecred/github-beta\\",\\"repositoryName\\":\\"sourcecred/example-repo\\",\\"id\\":\\"{\\\\\\"authorID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDQ6VXNlcjE0MDAwMjM=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"USER\\\\\\"},\\\\\\"contributionID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDEyOklzc3VlQ29tbWVudDM3Mzc2ODUzOA==\\\\\\",\\\\\\"type\\\\\\":\\\\\\"COMMENT\\\\\\"},\\\\\\"type\\\\\\":\\\\\\"AUTHORSHIP\\\\\\"}\\"}": Object {
+      "dst": Object {
+        "id": "{\\"id\\":\\"MDQ6VXNlcjE0MDAwMjM=\\",\\"type\\":\\"USER\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+      "payload": Object {},
+      "src": Object {
+        "id": "{\\"id\\":\\"MDEyOklzc3VlQ29tbWVudDM3Mzc2ODUzOA==\\",\\"type\\":\\"COMMENT\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+    },
+    "{\\"pluginName\\":\\"sourcecred/github-beta\\",\\"repositoryName\\":\\"sourcecred/example-repo\\",\\"id\\":\\"{\\\\\\"authorID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDQ6VXNlcjE0MDAwMjM=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"USER\\\\\\"},\\\\\\"contributionID\\\\\\":{\\\\\\"id\\\\\\":\\\\\\"MDU6SXNzdWUzMDU5OTM3NzM=\\\\\\",\\\\\\"type\\\\\\":\\\\\\"ISSUE\\\\\\"},\\\\\\"type\\\\\\":\\\\\\"AUTHORSHIP\\\\\\"}\\"}": Object {
+      "dst": Object {
+        "id": "{\\"id\\":\\"MDQ6VXNlcjE0MDAwMjM=\\",\\"type\\":\\"USER\\"}",
+        "pluginName": "sourcecred/github-beta",
+        "repositoryName": "sourcecred/example-repo",
+      },
+      "payload": Object {},
+      "src": Object {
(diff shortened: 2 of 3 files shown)


wchargin:
It looks like the merge-base is wrong here? Three of the commits (at least) seem like they shouldn't be here.

wchargin:
For now, I'll look at the contents of `githubPlugin.js` at your tip commit:  
https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js

wchargin:
First thoughts.

> https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js#L85-L86

I’m not so sure! Note that, for a fixed repository, there is an isomorphism between a node’s `Address` (which is a property of the node) and the `NodeID` (passed separately). I would suspect an abstraction leak if we _needed_ to pass the `NodeID` separately from the node, but we don’t: we could reconstruct it from `JSON.parse(authoredNode.address.id)`. Passing the `NodeID` separately is nothing more than a performance optimization. In the same vein, I would not balk at a function that accepted `commitSha: string` and `commitMessage: string` with the understanding that `git show --pretty=%B "${commitSha}"` should equal `commitMessage` exactly.

> https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js#L92-L94

Perhaps not `Organization`s, but what about `Bot`s? They comment on PRs all the time.

> https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js#L81

This should probably be `json-stable-stringify`.

> https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js#L59-L66

How do you feel about exposing these individually as `IssueNodeID`, etc., adjacent to their respective payload types?

Related: I note that there is no pull request node ID type, though there is a pull request node type. Presumably you’ll add the former in a later revision to this PR.

> https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js#L65

Can we rename `contributionID` to `childID`? Also, will every contribution have at most one out-containment edge, or should we include `parentID` as well (à la `REFERENCE`)?

> https://github.com/sourcecred/sourcecred/blob/6ca5f6082918e73c1ee9ffd7d9f6be937de0f498/src/backend/githubPlugin.js#L9-L10

In my opinion, the node payloads should not include the `type`. We do not include any information in the payload that is also contained in the address, and the address includes the node type. This has the advantage of being DRYer, and forces a decision (always good): we should have `export type NodeType = $ElementType<NodeID, "type">`.

> general

When you add nodes and edges to the graph, nothing verifies that they are type-correct unless you add type annotations (on either the payload or the node/edge; it doesn’t matter). You do so for `userPayload` and `authorshipEdge`, but not the others, so a typo in one of the strings would go unnoticed. For now, please add type annotations—but we should also consider ways to make this more mandatory. For instance: do we want to parameterize `Graph<NodePayload, EdgePayload>`?

> in your test case

You don’t need `parser.graph.toJSON()`; you can just use `parser.graph`. Reason: Jest calls `JSON.stringify` on the thing that you pass to it, and `JSON.stringify` knows about the special function name `toJSON`. We have tests for this.

---

Aside: GitHub’s GraphQL API doesn’t let you query for `user(id: $someId)`. But you can still look up users by ID, as shown in the following sample query:
```graphql
{
  u1: user(login: "wchargin") {
    login
    id
  }
  u2: node(id: "MDQ6VXNlcjQzMTc4MDY=") {
    ... on User {
      login
      id
    }
  }
  # `u1` and `u2` have identical data
}
```

wchargin:
> > general
> 
> When you add nodes and edges to the graph, nothing verifies that they are type-correct unless you add type annotations (on either the payload or the node/edge; it doesn’t matter). You do so for `userPayload` and `authorshipEdge`, but not the others, so a typo in one of the strings would go unnoticed. For now, please add type annotations—but we should also consider ways to make this more mandatory. For instance: do we want to parameterize `Graph<NodePayload, EdgePayload>`?

As of #83, this comment can be replaced by: “Simply change `graph: Graph` to `graph: Graph<NodePayload, EdgePayload>` and you’ll get all your typechecking for free.”

dandelionmane:
> Isomorphism between Address and ID

I felt a little uncomfortable depending on the Address for data that changes the interpretation of the node, e.g. putting the node type there. It felt safer to have the address be opaque. On reflection, however, I think that since the GitHub plugin owns these nodes, for the GitHub plugin to view the addresses as transparent is reasonable.

I don't think we need the performance optimization of passing around the nodeIds separately from the addresses, so I just parsed them from the address where necessary.

> Include bots?

Sure, it's trivial so I added support for bots and organizations.

> `json-stable-stringify`

Done.

> Pull request node ID

Added, although I haven't put in parsing for PRs yet. (But it will be straightforward.)

> Exposing ID types individually as IssueNodeID, etc.

I've done that, although I think it is of limited utility unless we can express to the type system that there's a relationship between (FooNodePayload, FooNodeID). Otherwise it will typecheck to bind a FooNodePayload with a BarNodeID. I guess there still is a value because sometimes we will explicitly know that we are dealing with e.g. a ReferenceEdgeID and can then depend on its properties.

> Can we rename `contributionID` to `childID`? Also, will every contribution have at most one out-containment edge, or should we include `parentID` as well (à la `REFERENCE`)?

Did the rename. We can add a `parentID` too, why not, it doesn't cost much and is defensive api design.

> in my opinion, the node payloads should not include the `type`

Done

> type safety

I'm having problems here - your insights would be appreciated (note the current commit doesn't typecheck). 

Given that `parseAddress => NodeID | EdgeID`, how do I decisively assign the result of parseAddress to some particular subtype, e.g. `UserID`?

This is a place where it would be really great if we could represent mappings between the Payload and ID types - ie. if `parseID<T>(n: Node<T>): ID<t>`

> No need for Graph.toJSON

Ack, changed.

wchargin:
I will answer first your most technical question.

> Given that `parseAddress => NodeID | EdgeID`, how do I decisively assign the result of `parseAddress` to some particular subtype, e.g. `UserID`?

The answer to the question that you have asked is that you can’t, without using `any`. As written, the result of `parseAddress` is a union type, so you only know that it is _either_ a node ID or an edge ID. Therefore, you can only perform operations on the result that are valid on _both_ node and edge IDs. In particular, you cannot downcast it to anything.

Of course, this is all a bit of a sham. You could implement
```js
parseNodeAddress(address: Address): NodeID {
  return JSON.parse(address.id);
}
parseEdgeAddress(address: Address): EdgeID {
  return JSON.parse(address.id);
}
```
which is obviously alarming; the only reason that either of these works, let alone both of them, is that `JSON.stringify` returns `any`:
https://github.com/facebook/flow/blob/469518a07adc8ae2662070f52a1e1ba7b01b5e33/lib/core.js#L481-L482

So you are already implicitly, unsafely downcasting from `any` to `NodeID | EdgeID` in the return type of your `parseAddress`.

Now, I see three potential solutions to the underlying problem. The simplest, cleanest option is to pass around richly typed node IDs wherever you pass nodes. If a function accepts `authorNode: UserNode`, it should also accept `authorNodeID: UserNodeID`. You had well-typed node IDs and you want well-typed node IDs; the problem is the retraction through the untyped string layer, so don’t do that.

The next idea is to switch the whole system over to `Address<T: JSONSerializable>` (as previously considered), and parameterize `Graph<NodeID, NodePayload, EdgeID, EdgePayload>`. This would work for the cases in this file, where you only need `NodeID` and not its subtypes, but it is not sufficient in general, and it doesn’t answer your original question of how to get a specific subtype of `NodeID`. The problem is, as you say, that we need to somehow “link” the corresponding address and node types. In other words, nodes’ address-IDs and payloads are not independent; that is, the set of valid types for `(node.address.id, node.payload)` is not the product of the sets of valid types for `node.address.id` and `node.payload`. Our state space is, in a precise sense, _entangled_.

With this in mind, the last option is to explicitly express the entangled state space in the type system, by parameterizing over a single type that is a carefully chosen subtype of the ID/payload product:
```js
type Address<+I> = {
  +group: string,
  +id: I,
};
type Node<+Descriptor> = {
  +address: Address<$ElementType<Descriptor, "id">>,
  +payload: $ElementType<Descriptor, "payload">,
};

type UserNodeDescriptor = {
  id: {
    type: "USER",
    id: string,
  },
  payload: {
    login: string,
  },
};
type IssueNodeDescriptor = {
  id: {
    type: "ISSUE",
    id: string,
  },
  payload: {
    title: string,
    number: number,
    body: string,
  },
};
type GitHubNodeDescriptor = UserNodeDescriptor | IssueNodeDescriptor;
type GitHubNode = Node<GitHubNodeDescriptor>;

const ex1: GitHubNodeDescriptor = {
  id: {type: "USER", id: "01010101"},
  payload: {login: "wchargin"},
};
const ex2: GitHubNodeDescriptor = {
  id: {type: "ISSUE", id: "23232323"},
  payload: {title: "hello", body: "world", number: 1},
};

const exBad = {
  id: {type: "USER", id: "01010101"},
  payload: {title: "hello", body: "world", number: 1},
}
// $ExpectError
(exBad: GitHubNodeDescriptor);
```
This should theoretically be strong enough to get us everything that we need. We can see that with the above definitions, a function like
```js
function processNode(node: GitHubNode) {
  if (node.address.id.type === "USER") {
    console.log(node.payload.login);
  }
  if (node.address.id.type === "ISSUE") {
    console.log(node.payload.number);
  }
}
```
_must_ be well-typed: each `if`-statement refines the type of `node`’s descriptor to one of the two branches of `GitHubNodeDescriptor`, and this information is used within the branch. Sadly, Flow is not smart enough to realize that this code is well-typed. A bit of exploration indicates that the root cause is that Flow does not propagate constraints learned from inference across sibling boundaries in an object type ([here is a minimal example][mwe]). Note that making everything exact, read-only, and covariant does not help. (See also: appendix below.)

[mwe]: https://flow.org/try/#0GYVwdgxgLglg9mABAcznAJgCgJSIN4BQiiUAngA4CmiAGogLz5HGIC2lrARpQE4BcTFizJUBAIgDKAFQBKASQByAcTEAaZkNakJUHjDDIBAZ137k6oQF8LiSwG5mI6gE0Ggluy68BhIcSfiCgCqALIAQgCiMmoaHqQKIF78iGCJ3Dw2xNbM9gTMoJCwCIjAmABeAnQAPojOuL4sMMCI5QB0numtTgz0jJKyiir1scRl7RydWjp6BnbEAPTzKGjoAISxlrFNLWMdvF0U1L19weFRYsN+iLsT+1oJSXOIi8sY61Y5BJsEBdDwSJwAIZYS4kQ60NwNYh7HgARh8IzBokQ-XkyhiVkybFuPAATAirlNTAZjMTzBsbLl-ODXIwodikvD3EIAijTpFoljspocfjmXEHukBKkklzKQ58uA-sVShUITU6vztm0YbCDlQen1pGihvzRuMkrj2toyU8XrweHAeIg+ABaTAbLbNFU4tXdY5s0Ici5664G9JG+5pXhmpYWq02+2OrJfAhAA

If the third “clever” solution had worked perfectly with Flow, I might have recommended it. After all, it does identify and attack the fundamental problem in a principled and extensible way, and the runtime elision is exactly the same as it was before (i.e., we didn’t have to change the object structure to capitulate to the type system). Its main downside is probably the performance discussed in #66, et al., and it could be argued that it makes the code harder for a newcomer to understand. But if Flow can’t handle these trivial examples, then the whole thing is moot.

So—my recommendation for now remains the same as my previous advice: pass around the IDs without ever dropping the types. It’s possible that this will not scale sufficiently, but it requires minimal effort right now, so let’s give it a shot.

---

Appendix: to justify my claim that the function `processNode` is provably correct, I’ve ported it to `typed/racket`, which uses a proper occurrence typing system and has no problem inferring the type:
```racket
#lang typed/racket

(define-type UserId (Pair 'user String))
(define-type UserPayload String)
(define-type UserDesc (Pair UserId UserPayload))

(define-type IssueId (Pair 'issue String))
(define-type IssuePayload Integer)
(define-type IssueDesc (Pair IssueId IssuePayload))

(define-type GitHubDesc (U UserDesc IssueDesc))

;; this type-checks just fine
(: process-desc (-> GitHubDesc Void))
(define (process-desc d)
  (match d
    [(cons (cons 'user id) payload)
     (void (string-length payload))]
    [(cons (cons 'issue id) payload)
     (void (* 2 payload))]))
```
Note that the Racket code is a bit awkward, using cons cells instead of structs everywhere, because Racket naturally lends itself to code that doesn’t need these shenanigans; I had to place down some hoops so that I could jump through them. As a sanity check, note that swapping the two continuations of the `match` form yields the appropriate type errors: if I write `(string-length payload)` in the second clause, I get “expected `UserPayload`, given `IssuePayload`”.

wchargin:
> On reflection, however, I think that since the GitHub plugin owns these nodes, for the GitHub plugin to view the addresses as transparent is reasonable.

This makes sense to me.

> I don't think we need the performance optimization of passing around the nodeIds separately from the addresses, so I just parsed them from the address where necessary.

I am okay with this, but only because it is an implementation detail localized to this module.

Other comments: acknowledged.

dandelionmane:
There is a fourth option, which is to make explicit all of the type mappings in function signatures, e.g.:

```
_getNodeIDWithAssertion(n: Node<NodePayload>, t: NodeType): any {
  const id: any = this.parseFromAddress(n.address);
  if (id.type !== t) { throw new Error("Type assertion violated") };
  return id;
}

getCommentID(n: Node<CommentNodePayload>): CommentNodeID {
  return this._getNodeIDWithAssertion(n, "COMMENT");
}

getIssueID(n: Node<IssueNodePayload>): IssueNodeID {
  return this._getNodeIDWithAssertion(n, "ISSUE");
}
// and so forth for every Node and Edge type
```

I think this is not a bad pattern. It's not super DRY, but it is very easy to read and to consume as an API, and it propagates type information we care about. For the parser code (which is all implementation, not API) there's no need for it, and you'll see in this commit I have adopted your option 1. However, I think we can use this pattern in public APIs we develop to access the GitHub data in a type-aware fashion.

Speaking of APIs: this PR does not add any convenient APIs for the GitHub data. Once we start using this data, we'll want those to exist; if you take a look at the test code, you'll see that doing simple tasks like acquiring every comment is cumbersome (and type-ignorant). We're almost certainly going to want an api somewhere that exposes `getAllComments(): GitHubCommentAPIEntity[]`. And it would be really nice if that `GitHubCommentAPIEntity` can have `.author` and `.container` properties so that people who want to find the author don't need to do edge-twiddling like I do in the test code. I considered creating a `GitHubGraph` which extends `Graph<GitHubNode, GitHubEdge>` and adds sugar methods for working with GitHub info in a more convenient fashion, but decided there was a lot of extra API design work that should go into that (how do we make a cohesive experience for inter-graph connections?).

Also: the `src` and `dst` pointers on our edges feel like a force fit to me. In the case of the `"COMMENT"` relationship, trying to decide which node is the src and which is the dst feels arbitrary. Each edge "type" is potentially a different kind of relationship; some of those relationships lend themselves to directional thinking, and some don't. Let's consider replacing `src`, `dst` with names that acknowledge that we can't assume directedness in general. 

Also: I took your suggestion of not putting `type` in the payload (because it is redundant with the address), but encoding the type in the ID doesn't seem right to me either. It's too opaque; it makes the type inaccessible in general to other plugins, but if I am another plugin connecting into or built on-top of the GH plugin, I want to find out the types of things. I think we should consider making `type` a first-class field on the `Address` object, which means that all `Node`s and `Edge`s get types by default, which should improve interpretability across the graph. The types should be namespaced by plugins, so this can actually replace the pluginName field: e.g. for comments here, the type is `GitHubPlugin-beta.Comment`. Going back to the issue of edge semantics, each Edge type can explicitly define its own semantics on how its `src` and `dst` relate.

& the code changed, so PTAL. It supports PRs now (but doesn't handle review or commit comments)

wchargin:
I've only read the first paragraph of your comment (will read the rest later), but if you're willing to do runtime checks then you can improve the statics, too:
```js
_getNodeIDWithAssertion<T: NodeID>(
    n: Node<NodePayload>,
    t: $ElementType<T, "type">
): T {
  const id: NodeID = this.parseFromAddress(n.address);
  if (id.type !== t) {
      throw new Error("Type assertion violated");
  }
  return id;
}

getCommentID(n: Node<CommentNodePayload>): CommentNodeID {
  return this._getNodeIDWithAssertion(n, "COMMENT");
}

getIssueNodeID(n: Node<IssueNodePayload>): IssueNodeID {
  return this._getNodeIDWithAssertion(n, "ISSUE");
}
```
([here is a `flow.org/try` for the above example](https://flow.org/try/#0KYDwDg9gTgLgBDAnmYcCCATDVgGddwC8cA3gD4BQccA1DpLgJYzSIByAhgLbABccuGFEYA7AOYAaKrTAAbAK5jRnHv0HDxU6jUYY1Q0ZIpkAvgG4KFUJFhxRMYFABmHAMapM2PLg4AjWagk0jQcWDj4-J7huBYmltbQ8EgocGwQGMAAPDQAKgB8RKSU2qFeEehh3loyHIiyEKH8OVKmFlbgiQjIqACiGGJZuQXE5MGl0ZGV+NU0uFCuk2W4MxiCi9EzYLX1jXDNxubxHbbJqACS+PLAaRkACtsNGIVB1DDMAfoaRtQi8ly+jn4v3+jmqvnSiE+hikhwSJ26cAuuCuN2AZwAIs9pKd+AAiM4AZQJAFUerjqrooZoKId2jYkgjbvJZLIAErAACOV0EqPudUez2KCHefAEBmpPz+AKgQKloOk4IwkLFXxasOODJSTJZ7K5eBgqIxWNe3Txt2JABkLQB9Vk9ACKpIJOXJ0kpKuhNLacM1qAAwhAuDwRAb0sA+TsniMhfIoLIqd84IrlepPa06Z1TnAA0HgCHDZiRtjTXBcX6APIAWUrPTYLopeg91NpPq6KWJuEcvIeoUF0nqShECbVFlbWY7XbDRqLJpQeOJBJ6rNd1Hdqeb3o1bdQACEIKG7j2o6R+xBB8OvRn4Sk9we0YWT7PRbid+X627G+ujC2t1ny1AxA4ERGAALw4N4IBEbt+V7Gc4AHUQL3Veltzgf9AOAsCIKgqcHxebc8XLVkAHE0DYM4AC00ByM5yzYFc7E-cVv03FCszQeQYAAC2gaDIyIaQyDgCcoD4x5BLgW8xNCCT0KA0DwMYSDpIwUdfwRDjuN43DChEgs4CEqSdKEuTMMU5TcNYzMERUgTqCEpEUTDCNxPsuBtTZTluTvFyZLcnNgx8o8JM0njROco81LYmydMICTHOuXCJI83VvILCSArzO8MRCziwoLKLrJSVEcgRYgABIegCQLSpQTICwkUtTlxPIrOvDw8ugXAuMYMA+gGXzjxIZCio6rSoG63r+vvY0CNLNBiRyAAJIiCSWs5bgY1xIIMXxOKUkQMX4BrpA4TqoCO1JcJhQr2uzHaOFEQLpsG54RrugMQ0ekRnv6Ga4JxUsKzrNAzjYGs6y2nrZAwS6TuoLYcBDOHrsvMcEXZJxHDzdwXqPN7bt9OBMexkRcb+6dHzm3E7QAMSXWs-TJaocCxqAcBlK6MgxFngDZrLjtRn9opSPGYKjXLxsmvq-sGjKHqerKxcjCSSaR8mBsiq8iemynQq6nqZYGI0hM+mBvt+43MSEtWceAXX0UJ1Dptq1AKqq4Aau6TIHca3FmtaywMlcWQOBwOAnHkMnsLgRHO1pqBAyibwAApxm8dZvAASkF7nHcsSPo4OuBrQGbL0QAdWYLi0HwRxsMyHJc-vPIU+kH5m-qiLxbyapXn4Srqqy13G79gOKBzvYqe2kRBEY5vKbj4AE6TqZcBTkQADp0-wLOLFXJw4BT3RN6zABCQhiBgLOqeoV4uMTgB3OARGAZ+enZ6AU9xV24A4OvYDFwAG5KVDg4DAuI97SDiNQHAMBYwiEYrEAuUdXAxzLplZG6IN6d0wUFHuk88H6XwnAhBJcy4FirtxWunZAGQQ3n7Cs1ZawuigXECghc0HFzLglAsOCuZZF4d3SMeRJ5CLzlTUhUBEGl2AOXKhNcAHYQYaWQkJIyRsKAA))

wchargin:
[draft comment]

> Speaking of APIs: this PR does not add any convenient APIs for the GitHub data. Once we start using this data, we'll want those to exist

Yes, agreed.

> I considered creating a `GitHubGraph` which extends `Graph<GitHubNode, GitHubEdge>` and adds sugar methods for working with GitHub info in a more convenient fashion, but decided there was a lot of extra API design work that should go into that

Definitely the right call to not do that right now, IMO.

> Let's consider replacing `src`, `dst` with names that acknowledge that we can't assume directedness in general.

Happy to consider this in more detail, but my immediate reactions are (a) you can embed an “undirected edge” in a directed graph by creating two edges, but you cannot embed in the other direction, so we should favor a directed structure, and (b) we’ve talked about this before and always wound up coming back to directed graphs; we can still revisit, but maybe we should have a canonical place to put arguments in favor of directed/undirected structure to use as documentation/reference.

> Also: I took your suggestion of not putting `type` in the payload (because it is redundant with the address), but encoding the type in the ID doesn't seem right to me either. It's too opaque; it makes the type inaccessible in general to other plugins, but if I am another plugin connecting into or built on-top of the GH plugin, I want to find out the types of things.

There’s nothing surprising here. The ID should definitely be opaque to the plugin, and the plugin can provide APIs for working with the IDs.

> I think we should consider making `type` a first-class field on the Address object

Of what type? `type: string`? Or are these types going to become rich enough that we want `type: JSONSerializable`, and then we want to expose subfields of `type` on the plugin?

> & the code changed, so PTAL. It supports PRs now (but doesn't handle review or commit comments)

will do.

dandelionmane:
Discussed offline: Most of my comments above are actually design considerations for the Graph/Address systems in general, and are not specific to the GitHub plugin, so we are going to punt on them until we've designed and implemented at least one additional plugin :)

src/backend/githubPlugin.test.js
```diff
@@ -0,0 +1,86 @@
...
+      issuesAndPRs.forEach((x) => {
+        const outEdges = graph.getOutEdges(x.address);
+        const authorEdges = outEdges.filter(
+          (e) => getEdgeType(e) === "AUTHORSHIP"
+        );
+        expect(authorEdges.length).toBe(1);
+        const commentEdges = graph
+          .getInEdges(x.address)
+          .filter((e) => getEdgeType(e) === "CONTAINMENT");
+        expect(commentEdges.length).toBeGreaterThanOrEqual(0);
```
wchargin: This assertion trivially holds.

src/backend/githubPlugin.test.js
```diff
@@ -0,0 +1,86 @@
...
+
+    it("parses the entire example-repo as expected", () => {
+      expect(graph).toMatchSnapshot();
+    });
+
+    it("every comment has an author and container", () => {
+      const comments = graph
+        .getAllNodes()
+        .filter((n) => getNodeType(n) === "COMMENT");
+      comments.forEach((c) => {
```
wchargin: Let's assert that there are not zero comments (and likewise for issues and PRs).

dandelionmane:
BTW, this PR is a good start on https://github.com/sourcecred/sourcecred/issues/40

PR #83, opened 2018-03-17
Parameterize `Graph` over node and edge payloads

### LINKED ISSUE #82: Consider parameterizing `Graph<N, E>`
A typical `Foo` plugin will have the `FooNode` and `FooEdge` types, and
construct a `Graph<FooNode, FooEdge>`—this is already true, but our type
system can't express it. Because `Graph` currently accepts `Node<any>`
and `Edge<any>` for additions, clients adding nodes and edges must
type-annotate them manually, or risk adding ill-typed data.

Merging becomes:
```js
merge: (g1: Graph<N1, E1>, g2: Graph<N2, E2>) => Graph<N1 | N2, E1 | E2>`
```

General graph algorithms will be able to operate parametrically,
yielding free theorems.

This sounds like a good idea. I hope that implementing it, and
expressing that these parameter types are covariant, is tractable within
Flow.

cc @dandelionmane; if you think that this is a good idea then I will be
happy to implement it.

wchargin:
Summary:
Closes #82. This affords clients type-safety without needing to
verbosely annotate every node or edge passed into the graph functions.
It also enables graph algorithms to be more expressive in their types:
for instance, the merge function now clearly indicates from its type
that the first graph’s nodes are passed as the first argument to the
node reducer, and the second graph’s nodes to the second. Clients can
upgrade immediately by using `Graph<*, *>`.

Thankfully, Flow supports variance well enough for this all to be
possible without too much trouble.

Test Plan:
Existing unit tests pass statically and at runtime. I added a test case
to demonstrate that merging works covariantly.

To see some failures, change `string` to an incompatible type, like
`number`, in the definitions of `makeGraph` in test functions for
conservatively rejecting graphs with conflicting nodes/edges
(ll. 446, 462).

wchargin-branch: parameterize-graph

### DIFF
--- a/src/backend/graph.js
+++ b/src/backend/graph.js
@@ -4,26 +4,26 @@ import deepEqual from "lodash.isequal";
 import type {Address, Addressable, AddressMapJSON} from "./address";
 import {AddressMap} from "./address";
 
-export type Node<T> = {|
+export type Node<+T> = {|
   +address: Address,
   +payload: T,
 |};
 
-export type Edge<T> = {|
+export type Edge<+T> = {|
   +address: Address,
   +src: Address,
   +dst: Address,
   +payload: T,
 |};
 
-export type GraphJSON = {|
-  +nodes: AddressMapJSON<Node<any>>,
-  +edges: AddressMapJSON<Edge<any>>,
+export type GraphJSON<NP, EP> = {|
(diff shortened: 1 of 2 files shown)


wchargin:
This also obsoletes #80 and removes all `any`s from our non-test codebase, which is surely a promising sign.

PR #85, opened 2018-03-19
Updated npm by running `npm i npm`.

dandelionmane:
Running said command generated a package-lock.json file, and instructed
us to commit it. However, I believe this is redundant with yarn.lock, so
following
https://stackoverflow.com/questions/44552348/should-i-commit-yarn-lock-and-package-lock-json-files
I decided to gitignore it instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# redundant with yarn.lock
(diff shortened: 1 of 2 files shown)


wchargin:
This isn't quite right: the dependency on `npm` expresses that `sourcecred`’s code depends on `npm` (like, we shell out to it or something). Instead, you just want to upgrade your global npm.

PR #90, opened 2018-03-20
Artifact editor work

wchargin:
Mostly paired with @dandelionmane.


src/plugins/artifact/editor/LocalStore.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+    const cacheKey = this.cacheKey(key);
+    if (cacheKey in window.localStorage) {
+      // (IE throws when deleting a non-existent entry.)
+      delete window.localStorage[cacheKey];
+    }
+  },
+
+  /*
+     * Local storage might be disabled in old browsers
```
dandelionmane: Indentation (here & throughout)

src/plugins/artifact/editor/LocalStore.js
```diff
@@ -0,0 +1,84 @@
...
+     * or in Safari's private browsing mode.
+     * Don't die.
+     */
+  isEnabled(): boolean {
+    const uid = String(+new Date());
+    try {
+      window.sessionStorage[uid] = uid;
```
dandelionmane: Why sessionStorage vs localStorage?

wchargin: I don't remember from when I originally wrote the code. I'd prefer not to change it right now because it's for Safari compatibility, and I'm pretty sure that the code here works. :shrug:

wchargin: (Discussed offline, and we agreed to keep it as is.)

PR #94, opened 2018-03-20
Use structured GraphQL query API in GitHub fetcher

wchargin:
Summary:
In addition to being nicer on the eyes, this enables the query to be
statically analyzed (e.g., by an auto-pagination API) and used by other
modules.

Test Plan:
Manually running
```shell
$ yarn backend
$ GITHUB_TOKEN="<your_token>" src/plugins/github/fetchGitHubRepoTest.sh
```
succeeds.

wchargin-branch: use-structured-graphql-queries

### DIFF
--- a/src/plugins/github/fetchGitHubRepo.js
+++ b/src/plugins/github/fetchGitHubRepo.js
@@ -6,6 +6,9 @@
 
 import fetch from "isomorphic-fetch";
 
+import type {Body} from "../../graphql/queries";
+import * as GraphQLQueries from "../../graphql/queries";
+
 /**
  * Scrape data from a GitHub repo using the GitHub API.
  *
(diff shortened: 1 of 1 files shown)


src/plugins/github/fetchGitHubRepo.js
```diff
@@ -42,111 +45,100 @@ export default function fetchGitHubRepo(
...
+                  makePageInfo(),
+                  b.field("nodes", {}, [
+                    b.field("id"),
+                    b.field("body"),
+                    makeAuthor(),
+                    b.field("state"),
+                    b.field("comments", {first: b.literal(10)}, [
+                      makePageInfo(),
+                      b.field("nodes", {}, [
+                        b.field("id"),
```
dandelionmane: Please get the comment url too

src/plugins/github/fetchGitHubRepo.js
```diff
@@ -42,111 +45,100 @@ export default function fetchGitHubRepo(
...
+  const body: Body = [
+    b.query(
+      "FetchData",
+      [b.param("repoOwner", "String!"), b.param("repoName", "String!")],
+      [
+        b.field(
+          "repository",
+          {owner: b.variable("repoOwner"), name: b.variable("repoName")},
+          [
+            b.field("issues", {first: b.literal(100)}, [
```
dandelionmane: Maybe factor all the first (N) arguments into constants?

wchargin: I'd like to keep the query identical in this commit. We can do this after, if you like.

src/plugins/github/fetchGitHubRepo.js
wchargin: Will do. (The query was correct when written; more stuff has been merged onto master since then. The tests will catch this when I rebase, and then I'll fix it.)

Issue #25, opened 2018-02-18
develop frontend testing practices


dandelionmane:
It's tempting to charge ahead with the frontend and wait until after the UI is more established to begin testing it. However, based on past experience, it's easy to get to a state where the system is so complex that adding tests is a daunting undertaking. It will be much better to test as we go. 

I'd like to set up functional tests for the following behaviors:
- expand/collapse FileExplorer UI 
- change selected filepath in FileExplorer
  - .. and UserExplorer receives it
     - ... and UserExplorer displays new sums

I'm not sure what the best way to structure these tests. Should it be very tightly scoped unit tests of each individual primitive of the behavior, or should it be end-to-end, going straight from the user input on the FileExplorer to the updated sums in the UserExplorer? 

I think we can have both, in a fairly labor efficient way, if we mix unit testing and [snapshot testing](https://facebook.github.io/jest/docs/en/snapshot-testing.html). For each individual component-specific interaction, we can test the logic via unit tests. For end-to-end integration, we can describe a sequence of user inputs and then take a snapshot in the terminal state. We should be able to write these easily and quickly, and maintain them without too much effort.

cc @wchargin

wchargin:
> For each individual component-specific interaction, we can test the logic via unit tests

Certainly.

> For end-to-end integration, we can describe a sequence of user inputs and then take a snapshot in the terminal state

This suggests to me that

wchargin:
:+1:

> For each individual component-specific interaction, we can test the logic via unit tests

Agreed. 

> For end-to-end integration, we can describe a sequence of user inputs and then take a snapshot in the terminal state

Perhaps. This suggests to me that the E2E test’s snapshot will change frequently: the test could be invalidated by style/copytext changes even at leaf nodes, changes to default sorting/allocation settings, etc. It would be unfortunate if this led to people (read: future-me) regenerating the snapshots blithely enough that actual regressions are not only unnoticed but promoted to the “correct” behavior. But I’m still willing to try it out!

I _do_ agree that we should have “E2E things-still-work” tests for each major feature. Such tests are probably easy enough to write without snapshot testing—DOM inspection is made pretty easy by helper functions like `findManyWithTag`, `findUniqueWithId`, etc. For instance, the test that you describe with “`UserExplorer` displays new sums” is easy to write in this way.

Heck, we could write both versions, and see which ones give better ROC curves. :-)

wchargin:
Closed by #102.

PR #102, opened 2018-03-21
Use Enzyme to test ContributionList dynamics

wchargin:
Summary:
This is our first dynamic test of a React component! Enzyme looks pretty
easy to use to me, for both snapshot tests and interaction simulation.

In doing so, we catch a minor bug in the edge case where a contribution
is not owned by any plugin (`colSpan`, not `colspan`). This edge case
does not appear in the sample data, but it does appear in the test data,
even prior to this commit. The previous renderer, `react-test-renderer`,
appears not to surface this error. Furthermore, this bug did not cause
any user-visible errors except a `console.error`.

Test Plan:
Inspect the snapshot file to make sure that it is reasonable. (The
existing test case has its snapshot regenerated due to formatting
differences between the two renderers.)

To test that the browser error is fixed, render a contribution list on a
GitHub graph but with an empty adapter set. One way to do this is to comment out line 7 of
`standardAdapterSet.js`; alternately, you can use the React Dev Tools to
select the `ContributionList` node, then run
```js
$r.props.adapters.adapters = {};
$r.forceUpdate();
```
Note subsequently that there is no console error and that the `<td>`s in
question span three columns.

wchargin-branch: contributionlist-dynamic-test

### DIFF
--- a/src/plugins/artifact/editor/ContributionList.test.js
+++ b/src/plugins/artifact/editor/ContributionList.test.js
@@ -1,8 +1,9 @@
 // @flow
 
+import type {ReactWrapper} from "enzyme";
 import React from "react";
-import ReactDOM from "react-dom";
-import reactTestRenderer from "react-test-renderer";
+import {mount} from "enzyme";
+import enzymeToJSON from "enzyme-to-json";
 
 import type {Address} from "../../../core/address";
 import type {Node} from "../../../core/graph";
@@ -12,6 +13,7 @@ import {ContributionList} from "./ContributionList";
 import {Graph} from "../../../core/graph";
 
 require("./testUtil").configureAphrodite();
+require("./testUtil").configureEnzyme();
 
 function createTestData(): * {
   type PayloadA = number;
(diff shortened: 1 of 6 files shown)


src/plugins/artifact/editor/ContributionList.test.js
```diff
@@ -152,11 +154,47 @@ function createTestData(): * {
...
+      .filterWhere((x) => pattern.test(x.text()));
+    expect(targetOption).toHaveLength(1);
+    container
+      .find("select")
+      .simulate("change", {target: {value: targetOption.prop("value")}});
+  }
+
+  it("renders some test data in the default state", () => {
+    const result = render();
+    expect(result.find("tbody tr")).toHaveLength(5);
```
dandelionmane: why 5? could this be rewritten as testNodes.length or such?

wchargin: Could be `createTestData().graph().getAllNodes().length`, although the other case of this assertion would have to be either still `(2)` or something weird that recreates the filter function (i.e., not useful).

Really, though, I might as well just remove these assertions: they're redundant with the snapshot test.

wchargin:
> Can we yarn remove react-test-renderer?

It's still used in `githubPluginAdapter.test.js`. We could potentially replace that with Enzyme, too, though [Jest recommends using `react-test-renderer` for basic snapshot tests](https://facebook.github.io/jest/docs/en/tutorial-react.html#snapshot-testing) and [enzyme for interactions.](https://facebook.github.io/jest/docs/en/tutorial-react.html#dom-testing). I switched the static snapshot test in this file to `enzyme` because I thought it was clearer to have the two snapshots in this file be of the same format.

Any preferences—in general, and for this PR specifically?

src/plugins/artifact/editor/ContributionList.test.js
wchargin: > Really, though, I might as well just remove these assertions: they're redundant with the snapshot test.

Although, after removing them, we should keep in mind that if the snapshots undergo a ton of churn, then we could easily replace them with more direct tests like this.

dandelionmane:
Overall, my bias is to have fewer tools, and having two tools that essentially do the same thing (but one is more general-purpose than the other) is unappetizing. Looking at the docs you linked, it sounds like there might be an issue around mocking if we use react-test-renderer.

I think my vote would be: remove react-test-renderer, with a strong willingness to add it back if/when it turns out we really need it. But that leaves us with the bias towards simpler tooling, and adding complexity only at the point where it is clearly providing additional value. Fine with doing that in a separate PR.

dandelionmane:
Also relevant (arguably fixes): https://github.com/sourcecred/sourcecred/issues/25

wchargin:
> I think my vote would be: remove react-test-renderer, with a strong willingness to add it back if/when it turns out we really need it. But that leaves us with the bias towards simpler tooling, and adding complexity only at the point where it is clearly providing additional value. Fine with doing that in a separate PR.

Sounds good to me.

PR #106, opened 2018-03-26
Add a toString and fromString method on addresses

dandelionmane:
The toString and fromString methods use json-stable-stringify,
and we've modified other address code to use these methods.
As such, a number of the snapshots have changed (ordering).

Test plan:
Verify the new included unit tests are comprehensive. Inspect the
snapshots.

### DIFF
--- a/src/core/address.test.js
+++ b/src/core/address.test.js
@@ -1,7 +1,7 @@
 // @flow
 
 import type {Address} from "./address";
-import {AddressMap, sortedByAddress} from "./address";
+import {AddressMap, fromString, sortedByAddress, toString} from "./address";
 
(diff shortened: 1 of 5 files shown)


src/core/address.test.js
```diff
@@ -152,4 +152,32 @@ describe("address", () => {
       expect(x).toEqual(input());
     });
   });
+
+  describe("toString and fromString", () => {
+    const examples = [mansion(), fakeMansion(), mattressStore()];
```
wchargin: Let's make this a function that generates the examples to avoid shared state across test cases:
```js
const examples = () => [mansion(), fakeMansion(), mattressStore()];
```

Issue #117, opened 2018-04-03
Write a pagination API for GitHub GraphQL queries


wchargin:
When we request data from the GitHub API, we are limited to a maximum of
100 items per connection at a time: 100 issues per repository, 100
comments per issue, etc. When we want to fetch _all_ data for a
repository, we’ll need to take advantage of GitHub’s pagination API to
fetch subsequent pages. Currently, our code assumes (and asserts) that
all data always fits on a single page, but clearly this is not a
scalable long-term solution: SourceCred itself will soon have more than
100 pull requests, for instance.

Eventually, we would like to implement an “auto-pagination” function
that takes an arbitrary query against the GitHub API, executes it, and
keeps fetching subsequent pages until all data has been exhausted. I’ve
been working on this for a while (order of a week), and it is
fascinatingly difficult. Without going into too much detail: for a fully
general solution, we need to map each field in the response to the field
in the query and the associated GraphQL type, which gets complicated
when dealing with fragments, and further complicated when fragments have
overlapping support. Our existing GitHub query falls into this last
category, so the functionality, at least, is required.

The GraphQL spec is of excellent quality, and is very clear about the
semantics of what happens in the query. This makes it easy enough to
formulate a top-level API and contracts for the functions that we want.
But the implementation of these functions should clearly be decomposed,
and exactly how to decompose them—what types to use, where to cut the
recursion, which base cases to pick—is surprisingly challenging. I’m
confident that I could implement it, given enough time, but it take a
lot of careful thought. (I’d currently estimate one programmer-week’s
worth of design and implementation work, non-parallelizable.)

As our tolerance for ~~hacky~~ ad hoc solutions increases, the problem
becomes easier. In the short term, we should consider implementing a
pagination function that works just for our particular GitHub query.
This removes the greatest difficulty, type tracking across fragments,
because we can design the code around a particular query structure. This
will unblock us indefinitely, at the cost of increasing the friction
involved in changing the GitHub query. (Estimated work: one or two days,
non-parallelizable, including good test coverage; maybe one extra day if
we want really rigorous tests.)

In the long term, it would be nice to implement a more general-purpose,
but still GitHub-specific, pagination API. A good metric for generality
is that it should be releasable as a standalone npm package, with
dependencies only on the structured query language. We can keep this in
mind as it becomes more cumbersome to write queries with manual
pagination—and also as just a “nice-to-have” on the calendar.

Absent objections, I’ll start implementing this; I have the most context
[...]

PR #119, opened 2018-04-05
Extract GitHub GraphQL code to a module

wchargin:
Summary:
Per #117, we want to develop an ad hoc pagination API written
specifically against the query that we use to interact with GitHub.
The pagination logic should be separate from the logic to actually fetch
the repo, but should be colocated with the query itself, so this commit
extricates the query from `fetchGithubRepo.js` into a new module.

Test Plan:
Existing tests pass, including `fetchGithubRepoTest.sh`.

wchargin-branch: extract-github-graphql

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -6,8 +6,8 @@
 
 import fetch from "isomorphic-fetch";
 
-import type {Body} from "../../graphql/queries";
-import * as GraphQLQueries from "../../graphql/queries";
+import {stringify, inlineLayout} from "../../graphql/queries";
+import {createQuery, createVariables} from "./graphql";
 
 /**
  * Scrape data from a GitHub repo using the GitHub API.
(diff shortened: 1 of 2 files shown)


src/plugins/github/fetchGithubRepo.js
```diff
@@ -45,101 +45,14 @@ export default function fetchGithubRepo(
     throw new Error(`Invalid token: ${token}`);
   }
 
-  const query = GraphQLQueries.stringify.body(
-    createQuery(),
-    GraphQLQueries.inlineLayout()
-  );
-  const variables = {repoOwner, repoName};
+  const query = stringify.body(createQuery(), inlineLayout());
+  const variables = createVariables(repoOwner, repoName);
```
decentralion: why do we want this createVariables function given that it's a pretty trivial one line expression?

wchargin: The variable names are now encapsulated by this module. Code outside the module doesn't know the variable names, doesn't need to, and shouldn't.

decentralion: Got it.

PR #121, opened 2018-04-05
Create pagination continuations for GitHub query

wchargin:
Summary:
Per #117, this is a first step toward at writing a pagination API that
specifically targets our current GitHub query. For design details, see
new module docs on `src/plugins/github/graphql.js`.

This commit modifies the core GitHub query and thus the
`example-repo.json` snapshot: we now request `endCursor` fields for all
pagination info, and we request the `id` of the root `repository` field.
The former is obviously necessary. The latter is necessary for the
repository to be consistent with other nodes that offer connections as
fields: we require an ID on the node containing the connection so that
we can have random access to it in a continuation selector.

Test Plan:
Unit tests added. You can also try out the generated continuation
queries for yourself: apply the patch below, run `yarn backend`, and
then run the `fetchGithubRepo.js` script on `sourcecred/sourcecred`.
This will output a nicely formatted query that you can paste directly
into GitHub’s API explorer and execute. (Note that, because this patch
is not fully polished, the query must be run against a repository that
has a continuation for every node type: more pages of issues, PRs,
comments, reviews, and review comments. This is due to an
easy-but-annoying-to-fix bug in the patch, not in the code included in
this commit.)

<details>
<summary>Patch for generating a continuations query</summary>

```diff
diff --git a/src/plugins/github/fetchGithubRepo.js b/src/plugins/github/fetchGithubRepo.js
index 789a20e..418c736 100644
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -6,8 +6,13 @@

 import fetch from "isomorphic-fetch";

-import {stringify, inlineLayout} from "../../graphql/queries";
-import {createQuery, createVariables} from "./graphql";
+import {stringify, inlineLayout, multilineLayout} from "../../graphql/queries";
+import {
+  continuationsFromQuery,
+  continuationQuery,
+  createQuery,
+  createVariables,
+} from "./graphql";

 /**
  * Scrape data from a GitHub repo using the GitHub API.
@@ -66,8 +71,13 @@ function postQuery(payload, token) {
       if (x.errors) {
         return Promise.reject(x);
       }
-      ensureNoMorePages(x);
-      return Promise.resolve(x);
+      console.log(
+        stringify.body(
+          continuationQuery(Array.from(continuationsFromQuery(x.data))),
+          multilineLayout("  ")
+        )
+      );
+      throw new Error("STOPSHIP");
     });
 }

diff --git a/src/plugins/github/graphql.js b/src/plugins/github/graphql.js
index 9ea2592..9ead42b 100644
--- a/src/plugins/github/graphql.js
+++ b/src/plugins/github/graphql.js
@@ -39,11 +39,11 @@ import {build} from "../../graphql/queries";
  *
  * [1]: https://developer.github.com/v4/guides/resource-limitations/#node-limit
  */
-export const PAGE_LIMIT = 100;
-const PAGE_SIZE_ISSUES = 100;
-const PAGE_SIZE_PRS = 100;
-const PAGE_SIZE_COMMENTS = 20;
-const PAGE_SIZE_REVIEWS = 10;
+export const PAGE_LIMIT = 10;
+const PAGE_SIZE_ISSUES = 10;
+const PAGE_SIZE_PRS = 10;
+const PAGE_SIZE_COMMENTS = 3;
+const PAGE_SIZE_REVIEWS = 1;
 const PAGE_SIZE_REVIEW_COMMENTS = 10;

 /**
@@ -340,6 +340,36 @@ function* continuationsFromReview(
   }
 }

+/**
+ * Combine continuations into a query.
+ */
+export function continuationQuery(
+  continuations: $ReadOnlyArray<Continuation>
+): Body {
+  const nonces: string[] = continuations.map((_, i) => `_n${String(i)}`);
+  const nonceToIndex = {};
+  nonces.forEach((n, i) => {
+    nonceToIndex[n] = i;
+  });
+  const b = build;
+  const query = b.query(
+    "Continuations",
+    [],
+    continuations.map((continuation, i) =>
+      b.alias(
+        nonces[i],
+        b.field(
+          "node",
+          {id: b.literal(continuation.enclosingNodeId)},
+          continuation.selections.slice()
+        )
+      )
+    )
+  );
+  const body = [query, ...createFragments()];
+  return body;
+}
+
 /**
  * These fragments are used to construct the root query, and also to
  * fetch more pages of specific entity types.
```
</details>

wchargin-branch: ad-hoc-pagination-continuations


src/plugins/github/graphql.js
```diff
@@ -12,14 +34,64 @@ import {build} from "../../graphql/queries";
...
+ * The nonce (`_n0`) is deliberately not included in the continuation
+ * type, because the nonce is a property of a particular embedding of
+ * the continuation into a query, and not of the continuation itself.
+ */
+export type Continuation = {|
+  +enclosingNodeType:
+    | "REPOSITORY"
+    | "ISSUE"
+    | "PULL_REQUEST"
+    | "PULL_REQUEST_REVIEW",
```
decentralion: It feels like there's some repetition of information between the EnclosingNodeType and other sequences, e.g. maybe the types of fragments? In principle, it would be nice if we could consolidate this all in one place / enforce the mapping at the type level. In practice, I think our attempts to do that within the GitHub plugin suggest it's not worth the effort right now. But wanted to mention it all-the-same.

src/plugins/github/graphql.js
```diff
@@ -46,9 +119,235 @@ export function createQuery(): Body {
...
+  if (result.comments.pageInfo.hasNextPage) {
+    yield {
+      enclosingNodeType: "ISSUE",
+      enclosingNodeId: nodeId,
+      selections: [
+        b.inlineFragment("Issue", [
+          b.field(
+            "comments",
+            {
+              first: b.literal(PAGE_LIMIT),
```
decentralion: I appreciate that once we hit a continuation, we use the full PAGE_LIMIT rather than the per-type page limit.

(Although, is it hypothetically possible that we we would over-use our total pages limit? what's the maximum number of query fragments we could be putting into a single continuation? is it 100^(max enclosing depth) if we got more-than-full pages on every level of the nesting?)

if we did hit this issue, would it fail clearly? i'm curious, but not too worried in practice.

src/plugins/github/graphql.js
```diff
@@ -46,9 +119,235 @@ export function createQuery(): Body {
...
+        ]),
+      ],
+      destinationPath: path,
+    };
+  }
+  if (result.reviews) {
+    for (let i = 0; i < result.reviews.nodes.length; i++) {
+      const issue = result.reviews.nodes[i];
+      const subpath = [...path, "reviews", "nodes", i];
+      yield* continuationsFromReview(issue, issue.id, subpath);
```
decentralion: it's nice that we can use nested iterators this way.

src/plugins/github/graphql.test.js
```diff
@@ -0,0 +1,381 @@
...
+                comments: {
+                  pageInfo: {
+                    hasNextPage: hasNextPageFor.issueComments,
+                    endCursor: "opaque-cursor-issue1comments",
+                  },
+                  nodes: [
+                    {
+                      id: "opaque-issue1comment1",
+                      author: makeAuthor("wchargin"),
+                      body: "I wish pancakes were still in vogue.",
```
decentralion: :nods contemplatively:

src/plugins/github/graphql.test.js
```diff
@@ -0,0 +1,381 @@
...
+                comments: {
+                  pageInfo: {
+                    hasNextPage: hasNextPageFor.prComments,
+                    endCursor: "opaque-cursor-pr2comments",
+                  },
+                  nodes: [
+                    {
+                      id: "opaque-pr2comment1",
+                      author: makeAuthor("decentralion"),
+                      body: "Why is there air?",
```
decentralion: someday i hope we can get some traction on these questions

src/plugins/github/graphql.test.js
```diff
@@ -0,0 +1,381 @@
...
+                  pageInfo: {
+                    hasNextPage: hasNextPages,
+                    endCursor: "opaque-cursor-issue3comments",
+                  },
+                  nodes: [
+                    {
+                      id: "opaque-issue3comment1",
+                      author: makeAuthor("decentralion"),
+                      body:
+                        "if it means everything, does it really mean anything?",
```
decentralion: mindblown.gif

src/plugins/github/graphql.js
wchargin: It’s not 100^(depth), because we only use the full `PAGE_LIMIT` on the outermost continuation. Within the continuation selections, we use the normal fragments, which have the shorter limits.

I think it is not too likely that we will over-use the total entities-per-page limit of 500K. Note that we can statically determine whether this might happen. If we do hit this issue, it will fail clearly: the GitHub API will return a clear message, we’ll catch it in the `errors` of the result, and reject in the `postQuery` handler in `fetchGithubQuery.js`.

src/plugins/github/graphql.js
wchargin: Yep! As far as I can tell, the iterator and generator APIs are pretty well designed.

src/plugins/github/graphql.js
wchargin: Without getting into too much detail (but happy to elaborate):
  - I agree.
  - These would be improved by a GitHub- or Relay-general API.
  - I am in general open to coalescing the various points of repetition, but think that the marginal value is very low. There are a few places where I can imagine that the types are not necessarily forced to be equal, and in those cases the cost of introducing a bad abstraction would be significant.
  - I think (agree) that doing any kind of type-level validation is not the right idea. Here, it might technically work, because everything is hard-coded—but in general, we want all the types to be dynamic, anyway, so we'd be hamstringing ourselves for much pain and no gain.

PR #122, opened 2018-04-05
Implement a function to merge query results

wchargin:
Summary:
Once we execute the root query, find continuations, embed the
continuations into queries, and execute the continuation query, we will
need to merge the continuations’ results back into the root results.
This commit adds a function `merge` that will be suitable for doing just
that.

Test Plan:
New unit tests added, with 100% coverage. Run `yarn test`.

wchargin-branch: merge-query-results

### DIFF
--- a/src/plugins/github/graphql.js
+++ b/src/plugins/github/graphql.js
@@ -341,6 +341,118 @@ function* continuationsFromReview(
 }
 
 /**
+ * Merge structured data from the given `source` into a given subpath of
+ * the `destination`. The original inputs are not modified.
+ *
+ * Arrays are merged by concatenation. Objects are merged by recursively
+ * merging each key. Primitives are merged by replacement (the
+ * destination is simply overwritten with the source).
+ *
+ * See test cases for examples.
+ *
+ * NOTE: The type of `source` should be the same as the type of
+ *
+ *     destination[path[0]][path[1]]...[path[path.length - 1]],
+ *
+ * but this constraint cannot be expressed in Flow so we just use `any`.
+ */
+export function merge<T>(
+  destination: T,
+  source: any,
+  path: $ReadOnlyArray<string | number>
+): T {
+  if (path.length === 0) {
+    return mergeDirect(destination, source);
+  }
+
+  function isObject(x) {
+    return !Array.isArray(x) && typeof x === "object" && x != null;
+  }
+  function checkKey(key: string | number, destination: Object | Array<any>) {
+    if (!(key in destination)) {
+      const keyText = JSON.stringify(key);
+      const destinationText = JSON.stringify(destination);
+      throw new Error(
+        `Key ${keyText} not found in destination: ${destinationText}`
+      );
+    }
+  }
+
+  const key = path[0];
+  if (typeof key === "number") {
+    if (!Array.isArray(destination)) {
+      throw new Error(
+        "Found index key for non-array destination: " +
+          JSON.stringify(destination)
+      );
+    }
+    checkKey(key, destination);
+    const newValue = merge(destination[key], source, path.slice(1));
+    const result = destination.slice();
+    result.splice(key, 1, newValue);
+    return result;
+  } else if (typeof key === "string") {
+    if (!isObject(destination)) {
+      throw new Error(
+        "Found string key for non-object destination: " +
+          JSON.stringify(destination)
+      );
+    }
+    const destinationObject: Object = (destination: any);
+    checkKey(key, destinationObject);
+    const newValue = merge(destinationObject[key], source, path.slice(1));
+    return {
+      ...destination,
+      [key]: newValue,
+    };
+  } else {
+    throw new Error(`Unexpected key: ${JSON.stringify(key)}`);
+  }
+}
+
+// Merge, without the path traversal.
+function mergeDirect<T>(destination: T, source: any): T {
+  function isObject(x) {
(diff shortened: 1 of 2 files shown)


src/plugins/github/graphql.js
```diff
@@ -340,6 +340,108 @@ function* continuationsFromReview(
...
+    function checkKey(key: string | number, destination: Object | Array<any>) {
+      if (!(key in destination)) {
+        const keyText = JSON.stringify(key);
+        const destinationText = JSON.stringify(destination);
+        throw new Error(
+          `Key ${keyText} not found in destination: ${destinationText}`
+        );
+      }
+    }
+    if (typeof key === "number") {
```
decentralion: shouldn't this block be a part of checkKey?

src/plugins/github/graphql.js
```diff
@@ -340,6 +340,108 @@ function* continuationsFromReview(
...
+ *
+ * See test cases for examples.
+ *
+ * NOTE: The type of `source` should be the same as the type of
+ *
+ *     destination[path[0]][path[1]]...[path[path.length - 1]],
+ *
+ * but this constraint cannot be expressed in Flow so we just use `any`.
+ */
+export function merge<T>(
```
decentralion: complaint: i find this function a little difficult to parse; it's almost 100 lines of code with inline functions but without any paragraph breaks. breaking it up visually (e.g. on the boundaries of inline functions) or breaking it up semantically (into mergeArray or mergeObject as appropriate) could be helpful.

src/plugins/github/graphql.js
decentralion: Ah, I see, that's because this is actually opening the "handle number key block" - I misread it as just being associated with another check. Another point for my "can this be separated a bit more clearly for readability" request, though.

wchargin: Noted. I'll split them out locally and see what I think.

wchargin:
So, I can offer the following refactoring, wherein we split out the path
traversal from the direct merging process. In my opinion, this is a
valuable distinction.

Note that the majority of this code is error handling, including
pretty-printing for error handling. Once we’ve refined the type, the
part that would be inlined is one line or a few lines, and I think that
the extracted versions are harder to read due to the indirection and
lack of context.

Finally, Prettier does not like any line breaks around `if`/`else if`
blocks, so this is as much vertical whitespace as I can reasonably add.

Your call: which do you prefer? (I lean toward this version.)

```js
export function merge<T>(
  destination: T,
  source: any,
  path: $ReadOnlyArray<string | number>
): T {
  if (path.length === 0) {
    return mergeDirect(destination, source);
  }

  function isObject(x) {
    return !Array.isArray(x) && typeof x === "object" && x != null;
  }
  function checkKey(key: string | number, destination: Object | Array<any>) {
    if (!(key in destination)) {
      const keyText = JSON.stringify(key);
      const destinationText = JSON.stringify(destination);
      throw new Error(
        `Key ${keyText} not found in destination: ${destinationText}`
      );
    }
  }

  const key = path[0];
  if (typeof key === "number") {
    if (!Array.isArray(destination)) {
      throw new Error(
        "Found index key for non-array destination: " +
          JSON.stringify(destination)
      );
    }
    checkKey(key, destination);
    const newValue = merge(destination[key], source, path.slice(1));
    const result = destination.slice();
    result.splice(key, 1, newValue);
    return result;
  } else if (typeof key === "string") {
    if (!isObject(destination)) {
      throw new Error(
        "Found string key for non-object destination: " +
          JSON.stringify(destination)
      );
    }
    const destinationObject: Object = (destination: any);
    checkKey(key, destinationObject);
    const newValue = merge(destinationObject[key], source, path.slice(1));
    return {
      ...destination,
      [key]: newValue,
    };
  } else {
    throw new Error(`Unexpected key: ${JSON.stringify(key)}`);
  }
}

// Merge, without the path traversal.
function mergeDirect<T>(destination: T, source: any): T {
  function isObject(x) {
    return !Array.isArray(x) && typeof x === "object" && x != null;
  }
  if (Array.isArray(source)) {
    if (!Array.isArray(destination)) {
      const destinationText = JSON.stringify(destination);
      const sourceText = JSON.stringify(source);
      throw new Error(
        "Tried to merge array into non-array: " +
          `(source: ${sourceText}; destination: ${destinationText})`
      );
    }
    return destination.concat(source);
  } else if (isObject(source)) {
    if (!isObject(destination)) {
      const destinationText = JSON.stringify(destination);
      const sourceText = JSON.stringify(source);
      throw new Error(
        "Tried to merge object into non-object: " +
          `(source: ${sourceText}; destination: ${destinationText})`
      );
    }
    const result = {...destination};
    Object.keys(source).forEach((k) => {
      result[k] = mergeDirect(result[k], source[k]);
    });
    return result;
  } else {
    if (Array.isArray(destination) || isObject(destination)) {
      const destinationText = JSON.stringify(destination);
      const sourceText = JSON.stringify(source);
      throw new Error(
        "Tried to merge primitive into non-primitive: " +
          `(source: ${sourceText}; destination: ${destinationText})`
      );
    }
    return source;
  }
}
```

src/plugins/github/graphql.js
wchargin: Addressed in https://github.com/sourcecred/sourcecred/pull/122#issuecomment-378859005.

decentralion:
I definitely like your split up version with merge and mergeDirect. Let's merge that.

wchargin:
Let’s, indeed.

PR #123, opened 2018-04-05
Implement exhaustive fetching for our GitHub query

wchargin:
Summary:
This commit completes the ad hoc pagination solution described in #117:
we implement pagination specifically for our current query against the
GitHub API. This is done in such a way that reasonable additions to the
query will not be hard to implement—for instance, if we want to fetch
a new kind of field, the marginal cost is at most a bit of extra
copy-and-paste and some modifications to tests. However, we should
certainly still plan to implement the fully automatic pagination system
described in #117.

Running on TensorBoard with the default page limits takes 30–33 seconds
over 7 queries, uses 103 GitHub API points (out of 5000 for any given
hour), and produces a JSON file that is 8.7MB (when pretty-printed).
This all seems quite reasonable to me.

Test Plan:
Extensive unit tests added. The snapshots are quite readable, by design.

For a real-world test, you can `yarn start` the artifact viewer and use
the GUI to fetch the data for tensorflow/tensorboard.

To demonstrate that the fetching process gives the same results
regardless of the page size, follow these steps:

 1. In `fetchGithubRepo.js`’s `postQuery` function, insert a new
    statement `console.error("Posting query...")` at the beginning. (It
    is important to print to stderr instead of stdout.)
 2. Run `yarn backend` and then invoke `fetchGithubRepo.js` on a repo
    large enough to require pagination, like SourceCred or TensorBoard.
    Pipe the result to `shasum -a 256` and note the SHA.
 3. In `github/graphql.js`, change the page size constants near the top
    of the file. Re-run step 2. The number of queries that are posted
    will vary significantly as the page size constants vary, but the
    checksum should remain the same.
 4. Repeat until satisfied. (I tried three sets of values: the standard
    values, the standard values but all divided by 10, and all 5s.)

wchargin-branch: ad-hoc-pagination


src/plugins/github/fetchGithubRepo.js
```diff
@@ -45,15 +45,28 @@ export default function fetchGithubRepo(
...
+    return {data: x};
+  });
 }
 
 const GITHUB_GRAPHQL_SERVER = "https://api.github.com/graphql";
 
-function postQuery(payload, token) {
+function postQuery({query, variables}, token) {
+  const payload = {
+    query: stringify.body(query, inlineLayout()),
```
decentralion: I find the re-use of the name "query" for (i assume) the programmatic query template and the laid-out string query slightly confusing, and would appreciate disambiguation

src/plugins/github/fetchGithubRepo.js
```diff
@@ -45,15 +45,28 @@ export default function fetchGithubRepo(
...
-  const query = stringify.body(createQuery(), inlineLayout());
+  const query = createQuery();
   const variables = createVariables(repoOwner, repoName);
   const payload = {query, variables};
-  return postQuery(payload, token);
+  return postQueryExhaustive(
+    (somePayload) => postQuery(somePayload, token),
+    payload
+  ).then((x) => {
+    ensureNoMorePages(x);
```
decentralion: if we were writing this from scratch today (i.e. after writing the query pagination solution), do you think we would have written an ensureNoMorePages function here? i'm happy to keep it either way (yay defensive coding), but it's maybe an interestingly path-dependent artifact

src/plugins/github/graphql.js
```diff
@@ -340,6 +345,129 @@ function* continuationsFromReview(
   }
 }
 
+function* usedFragmentNames(selection: Selection): Iterator<string> {
```
decentralion: What is this function for? would appreciate a docstring

src/plugins/github/fetchGithubRepo.js
wchargin: I would certainly include it. The cost—computationally, conceptually, in terms of codebase size—is trivial, and the value is high. I would consider moving it somewhere else, though: perhaps folded in as an assertion at the end of the `postQueryExhaustive` routine.

src/plugins/github/fetchGithubRepo.js
wchargin: Fair. Note that the laid-out string query name is forced by the GitHub API. I considered calling the structured one `body` everywhere, and am still happy to do so. (If there is a reason that I decided not to do that, then I forget the reason.)

src/plugins/github/graphql.js
wchargin: Sure; happy to add one. TL;DR: It is an error to include fragment definitions that are not uesd, so this function and `requiredFragments` perform a BFS of the query graph to find the required fragments. This one in particular gets the fragments used *non-transitively* in a given selection; `requiredFragments` does the search.

decentralion: Got it. Yeah, that context should definitely be documented - nonobvious to figure out what problem they solve absent the documentation.

wchargin:
PTAL.

I have rebased the original commit and added two more:

  * 5ae291a - Use `body` for the structured query format name
  * ca76756 - Restructure internals of `postQueryExhaustive`
  * 30d4958 - Implement exhaustive fetching for our GitHub query

Note to any mergers: please squash and merge, and use only the root
commit message (equivalently, the PR description).

> I find the re-use of the name "query" for (i assume) the programmatic
> query template and the laid-out string query slightly confusing, and
> would appreciate disambiguation

This is largely mitigated by the fact that Flow enforces the
distinction. Nevertheless, I’m happy to rename the structured variant to
`body`, primarily because it’s not worth bikeshedding over. :-)
 
> It would be nice to have a docstring that explains how the pieces of
> postQueryExhaustive fit together.

Done. Let me know what you think.

src/plugins/github/graphql.test.js
```diff
@@ -631,4 +637,331 @@ describe("graphql", () => {
...
+                },
+                reviews: {
+                  pageInfo: {
+                    hasNextPage: true,
+                    endCursor: "opaque-cursor-pr2reviews-v0",
+                  },
+                  nodes: [
+                    {
+                      id: "opaque-pr2review1",
+                      body: "You actually introduced a new typo instead.",
```
decentralion: i appreciate the rhyme (cred/instead). i now wish the whole test case were in verse

src/plugins/github/__snapshots__/graphql.test.js.snap
```diff
@@ -0,0 +1,494 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`graphql #postQueryExhaustive resolves a representative query 1`] = `
```
decentralion: so, this test case is very powerful, in that it tests the query construction, the fragment resolution, etc.
however, it also inlines the entire real query in the snapshot, which means that any time we modify the real query, we will generate a snapshot failure here. since we don't actually use the query, this is just noise/false positives for our snapshot detection.

is it feasible to re-write the test without causing spurious failures whenever we tweak the query? perhaps by constructing a small example query to use instead?

decentralion: Note: if we change the query, there is a fairly high probability we also need to update other snapshots, so having spurious failures on that change is not as bad as in the general case

wchargin: > however, it also inlines the entire real query in the snapshot, which means that any time we modify the real query, we will generate a snapshot failure here.

Correct.

> is it feasible to re-write the test without causing spurious failures whenever we tweak the query? perhaps by constructing a small example query to use instead?

I don’t think that constructing a small query is the right way to go. The pagination API is only intended to work on the standard query, so we would be violating the contract in the test—in an important way.

Here are two alternative options:
  - Remove the snapshot assertion entirely. This does not significantly affect the predictive power of the test: the `result` assertion is most important, and the `toHaveBeenCalledTimes(3)` makes sure that we aren’t making spurious requests.
  - Remove the first element from the snapshot array, and instead separately assert that the first argument was `{body: createQuery(), variables}`. This retains the full predictive power of the test, and shrinks the snapshot (but makes the code less uniform).
The reason that I included the snapshots was primarily for readers’ benefit, to make it easy to see what the continuation queries looked like. But, in retrospect, perhaps that’s not necessary: a reader who cares about this probably has a working copy of the code and enough understanding of the API to easily add a `console.log` to the right spot in the test. So I am fine with either of these solutions or the status quo, leaning toward removing the snapshot entirely. Thoughts?

src/plugins/github/graphql.test.js
wchargin: Unfortunately, that would violate the “no fun allowed” rule. ;-)

src/plugins/github/__snapshots__/graphql.test.js.snap
decentralion: I think removing that snapshot assertion is fine. Making it convenient to read what the generated query looks like is nice, but not worth the cause of introducing spurious snapshot failures - to do so privileges an edge case (wanting to see the exact generated query) over the common case (modifying something and not wanting spurious snapshot failures).

wchargin:
> Excellent work with this sequence of PRs!

Thank you.

> I think removing that snapshot assertion is fine. Making it convenient to read what the generated query looks like is nice, but not worth the cause of introducing spurious snapshot failures - to do so privileges an edge case (wanting to see the exact generated query) over the common case (modifying something and not wanting spurious snapshot failures).

Makes sense to me. Merging.

PR #125, opened 2018-04-15
GitHub: parse simple numeric references

decentralion:
This commit adds support for parsing numeric references to the GitHub
plugin. The implementation is mildly complicated, because the parser
now tracks two additional pieces of state:

1. A map from issue/pr # to node, necessary to avoid an O(nodes) scan
every time we encounter a numeric reference
2. A map containing every dangling numeric reference, necessary to avoid
missing references when we parse the reference before the referred node.

I also added a `danglingReferences` method to the parser which
enumerates every dangling reference, in a readable format. In a followon
commit, I intend to log the presence of dangling references in the
`parseGraph` method of the Artifact editor.

Test plan:
Observe that the unit tests are fairly comprehensive. The whole-repo
snapshot changes are as expected.

### DIFF
--- a/src/plugins/github/__snapshots__/githubPlugin.test.js.snap
+++ b/src/plugins/github/__snapshots__/githubPlugin.test.js.snap
@@ -369,6 +369,152 @@ Object {
 }
 `;
 
+exports[`GithubParser reference detection discovers a simple reference 1`] = `
+Object {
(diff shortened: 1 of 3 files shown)


decentralion:
Going to re-write with an enrichment pass, as suggested.

Issue #127, opened 2018-04-15
GitHub plugin: Use urls as ids


decentralion:
Currently, the GitHub plugin uses opaque GitHub id strings as the edge/node ids. I believe we should change to using canonical GitHub urls as ids.

My reasoning is that it will make it possible to reference a GitHub asset without needing to introspect on GitHub's opaque, internal-only id structure. A simple example is if repo A wants to reference an issue in repo B. The user can do that, canonically, via a url. However, at present it isn't possible to parse that url into a graph reference without executing an additional query against GitHub to retrieve the id of the referenced issue. 

Relevant: this issue in which I call for changing addresses to be real addresses in general, rather than opaque ids. https://github.com/sourcecred/sourcecred/issues/73. (Ie: the address type should be able to specify "object A in repository B at hash C" or "url X" or "object in IPFS with hash X"

cc: @wchargin

decentralion:
Closed by #144

PR #128, opened 2018-04-18
Get urls for all GitHub objects

decentralion:
This commit modifies our GitHub graphql query so that we request urls
for all objects (e.g. users, pull requests, pull request review
comments). Some change along these lines is necessary so that we can
correctly represent URL reference edges to e.g. issue comments. (It
might be possible to do without by reverse-enginering from the ids, but
we are resolved to treat ids as opaque).

Strictly speaking, we don't need to collect urls for users, issues, and
pull requests - they are generated via simple schema. However, for
consistency, I think it's better to just take URLs on everything.

Test plan: The example-repo.json has been regenerated. The diffs are as
expected.

### DIFF
--- a/src/plugins/artifact/editor/adapters/__snapshots__/githubPluginAdapter.test.js.snap
+++ b/src/plugins/artifact/editor/adapters/__snapshots__/githubPluginAdapter.test.js.snap
@@ -7,6 +7,7 @@ Array [
     "payload": Object {
       "body": "I'm sold",
       "state": "APPROVED",
+      "url": "https://github.com/sourcecred/example-repo/pull/5#pullrequestreview-100314038",
     },
     "rendered": <div>
       type: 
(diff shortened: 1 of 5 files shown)


decentralion:
Also: this is a prereq for https://github.com/sourcecred/sourcecred/issues/127

decentralion:
Note: In a future commit, I plan to explicitly differentiate between "posts", which have authors and bodies, and "authors", which do not, such that every GitHub node is either a "post" or an "author".

It's important that all posts have URLs, but not very important that every author have a url, since authors are almost always referenced by @-name. I'm defaulting to keeping the author urls because why not, but I'm very open to removing them and only collecting post urls.

wchargin:
> Note: In a future commit, I plan to explicitly differentiate between "posts", which have authors and bodies, and "authors", which do not, such that every GitHub node is either a "post" or an "author".

Fine with me, but: is this something that we want to rely on in the future? Another way to ask that is—what inferences does this assumption allow you to make?

> It's important that all posts have URLs, but not very important that every author have a url, since authors are almost always referenced by @-name. I'm defaulting to keeping the author urls because why not, but I'm very open to removing them and only collecting post urls.

No preference; consistency sounds like a good tiebreaker.

decentralion:
Re posts/non-posts: Will punt on discussing it until / if I produce said pull request. Merging! :)

PR #129, opened 2018-04-18
Add README and regenerate.sh for GitHub demo data

decentralion:
Keeping the GitHub demo data up-to-date is important, and there isn't
good documentation for how to do that.

This commit adds a short README.md for the demo data, and a
regenerate.sh script that can be invoked to easily update it.

regenerate.sh was forked from fetchGithubRepoTest.sh. I made one change,
which is to manually run `yarn backend` rather than trusting the user to
hvae done it; this ensures correctness, as the user may have changed the
query since the last rebuild, which would result in outdated results.
For consistency and correctness, I also updated fetchGithubRepoTest.sh.
(To avoid noise in the console, the stdout of the build command is
redirected to /dev/null. The error message when build fails [1] is still
reasonably interpretable as a build failure, and the user can run `yarn
backend` on their own to see the exact failure.)

Test plan:
Modify example-repo.json (e.g. by deleting it entirely). Run
regenerate.sh and confirm that the data was regenerated. Run
fetchGithubRepoTest.sh and confirm the test passes.

Note: One of the end cursors minutely changed. I do not know why, but
I verified that it is unrelated to this commit (repros on master).

[1]:
```
$ ./src/plugins/github/fetchGithubRepoTest.sh                                                                                                                                      sourcecred/git/demo-data-readme !
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c node scripts/backend.js
Directory: /home/dandelion/git/sourcecred
Output:
".
```


src/plugins/github/demoData/README.md
```diff
@@ -0,0 +1,18 @@
+# DemoData Instructions
+
+This folder contains example data for testing the GitHub plugin. The example
+data is generated by running our GitHub API fetch against the SourceCred
+example repo at https://github.com/sourcecred/example-repo/.
+
+The demo data should be regenerated any time either the example repository or
+the GitHub api query change. To regenerate the demo data, run the following
```
wchargin: nit: s/change/changes

src/plugins/github/demoData/README.md
```diff
@@ -0,0 +1,18 @@
...
+command:
+
+`
+$ GITHUB_TOKEN="your_token_here"; src/plugins/github/demoData/regenerate.sh;
```
wchargin: The first semicolon must be removed for the command to be correct. (The trailing semicolon should be removed for good style.)

wchargin: Can we use code fences here (and below)?
``````markdown
```shell
$ GITHUB_TOKEN="your_token_here" src/plugins/github/demoData/regenerate.sh
```
``````
This has the correct semantic meaning (block of code, not inline code), and also renders appropriately.

File: src/plugins/github/demoData/regenerate.sh [hunk 3]
```diff
@@ -0,0 +1,23 @@
...
+
+set -eu
+
+main() {
+  if [ -z "${GITHUB_TOKEN:-}" ]; then
+    printf >&2 'Please set the GITHUB_TOKEN environment variable\n'
+    printf >&2 'to a 40-character hex string API token from GitHub.\n'
+    return 1
+  fi
+  yarn backend >/dev/null;
```
wchargin: Trailing semicolon is unnecessary.

wchargin: (If only `chronic(1)` were part of `coreutils` and not [`moreutils`]… but this is superseded by general comment about this choice, anyway.)

[`moreutils`]: https://joeyh.name/code/moreutils/

src/plugins/github/demoData/README.md
decentralion: Done.

src/plugins/github/demoData/README.md
decentralion: Done

decentralion: Done

src/plugins/github/demoData/regenerate.sh
decentralion: Done (removed the >/dev/null pipe)

decentralion:
> Can we instead let `fetchGithubRepoTest.sh` take a `-u|--updateSnapshot`
argument? This way, (a) we reduce code duplication, and (b) we present
an interface consistent with Jest (which we shouldn’t necessarily strive
for in general, but is a nice benefit).

Sure, happy to do that. Could you recommend a pattern for how to parse those flags, or point me to a good example?

>> manually run yarn backend rather than trusting the user to hvae done
>> it; this ensures correctness, as the user may have changed the query
>> since the last rebuild, which would result in outdated results. For
>> consistency and correctness, I also updated fetchGithubRepoTest.sh.

>This is fine with me. It’s not ideal: best would be a Makefile-like
>resolution that avoids spurious invocations of `yarn backend`, which
>takes an annoyingly long time. Second-best would be an “override”
>argument, but fine to keep as is for now.

Best would be to make the build fast, fixing anything here is second-best.

> Two things. First: this is a sneaky way to say, “by the way, we now
> silence all warnings”. I disagree with this change. I believe that the
> benefit of seeing new actual warnings outweighs the loss of a few lines
> of terminal space. Second (more minor): if we ever want to run this on
> any kind of CI system, the benefit of having the exact errors available
> without re-running `yarn backend` is increased.

Reverted.

And yes, I was in Israel when I generated this change :)

src/plugins/github/fetchGithubRepoTest.sh
```diff
@@ -2,27 +2,58 @@
...
+  diff -uw "${data_file}" "${output}"
   rm "${output}"
 }
 
+update() {
+  fetch >"${data_file}"
+}
+
+main() {
+	cd "$(git rev-parse --show-toplevel)"
```
wchargin: Tab spotted! :o

PR #130, opened 2018-04-19
Simplify GH reference handling

decentralion:
Currently, every type of reference has its own type signature: numeric
references are returned as numbers, url references are a complicated
object containing url parts, and so forth.

Since ultimately the references are just strings, it makes more sense to
treat references as plain strings. This allows a much simpler
implementation of reference edge creation in the GitHub plugin.

This commit also adds username reference detection.

Test plan:
Verify that the updated tests encode appropriate behavior.

### DIFF
--- a/src/plugins/github/parseReferences.js
+++ b/src/plugins/github/parseReferences.js
@@ -13,26 +13,18 @@ function findAllMatches(re: RegExp, s: string): any[] {
   return matches;
 }
 
-export function findNumericReferences(body: string): number[] {
-  return findAllMatches(/(?:\W|^)#(\d+)(?:\W|$)/g, body).map((x) => +x[1]);
+export function findReferences(body: string): string[] {
+  // Note to maintainer: If it becomes necessary to encode references in a
+  // richer format, consider implementing the type signature described in
+  // https://github.com/sourcecred/sourcecred/pull/130#pullrequestreview-113849998
+  return [...findNumericReferences(body), ...findGithubUrlReferences(body)];
 }
 
-export type GithubUrlMatch = {|
-  +repoName: string,
-  +repoOwner: string,
-  +parentType: "pull" | "issues",
-  +number: number,
-  +commentFragment: ?{|
-    +fragmentType:
-      | "issue" // a directly linked issue or pull request
-      | "issuecomment" // a directly linked regular comment on issue or pull request
-      | "pullrequestreview" // a pull request review
-      | "discussion_r", // a review comment as part of a pull request review
-    +fragmentNumber: number,
-  |},
-|};
+function findNumericReferences(body: string): string[] {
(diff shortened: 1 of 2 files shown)


src/plugins/github/parseReferences.js
```diff
@@ -46,20 +41,5 @@ export function findGithubUrlReferences(body: string): GithubUrlMatch[] {
       /(?:\W|$)/.source,
     "gm"
```
wchargin: Unrelated to this change, but I think that you do not actually need the `m` (multiline) flag, because `/\n|\r/` is a subset of `/\W/`. Tests pass without the flag. If you think that the flag should be there, then I would like to see a test to this effect (in a different PR).

src/plugins/github/parseReferences.js
```diff
@@ -13,26 +13,21 @@ function findAllMatches(re: RegExp, s: string): any[] {
   return matches;
 }
 
-export function findNumericReferences(body: string): number[] {
-  return findAllMatches(/(?:\W|^)#(\d+)(?:\W|$)/g, body).map((x) => +x[1]);
+export function findAllReferences(body: string): string[] {
+  return findNumericReferences(body)
```
wchargin: `concat` is variadic. Prefer either of the following:
```js
return [].concat(
  findNumericReferences(body),
  findUsernameReferences(body),
  findGithubUrlReferences(body)
);
return [
  ...findNumericReferences(body),
  ...findUsernameReferences(body),
  ...findGithubUrlReferences(body),
];
```

src/plugins/github/parseReferences.js
```diff
@@ -46,20 +41,5 @@ export function findGithubUrlReferences(body: string): GithubUrlMatch[] {
...
-    }
-    return {
-      repoOwner: match[1],
-      repoName: match[2],
-      parentType: match[3],
-      number: +match[4],
-      commentFragment,
-    };
-  });
+  return findAllMatches(urlRegex, body).map((match) => match[0].trim());
```
wchargin: This is totally fine with me. You could also use
```js
return body.match(urlRegex).map((x) => x.trim());
```
because you do not care about any particular capturing groups.

src/plugins/github/parseReferences.js
decentralion: Ack

src/plugins/github/parseReferences.js
decentralion: Done

src/plugins/github/parseReferences.js
decentralion: I tried that, but body.match(urlRegex) returns null and so .map throws a type error. Leaving unchanged.

wchargin: Apparently `match` returns an array of matches, or `null` if there are none—a cardinal sin. Your way is better. :-)

PR #133, opened 2018-04-23
Switch typed dispatch table to `empty` assertion

wchargin:
Summary:
This replaces the implementation of a static check from a somewhat
complicated use of higher-order types to a more simple empty-union
assertion, as suggested by jez in “Case Exhaustiveness in Flow”:
https://blog.jez.io/flow-exhaustiveness/

(I know; we’re not using Reason. One step at a time. :-) )

I adapted the implementation a bit because I prefer explicitly disabling
an ESLint warning over a no-op function call; it is not clear from the
latter that the purpose is to suppress a lint warning.

Test Plan:
In `githubPlugin.js`, add `| "ANOTHER"` to the `NodeType` type, and note
a compile-time Flow error on the appropriate line, with a very readable
error message. Note that all unit tests pass, and running the UI on
`sourcecred/sourcecred` yields correct titles for each node type present
(namely, all node types except for `ORGANIZATION` and `BOT`).

wchargin-branch: empty-union-assertion

### DIFF
--- a/src/plugins/artifact/editor/adapters/githubPluginAdapter.js
+++ b/src/plugins/artifact/editor/adapters/githubPluginAdapter.js
@@ -7,7 +7,6 @@ import type {Node} from "../../../../core/graph";
 import type {
   NodePayload,
   NodeType,
-  NodeTypes,
   IssueNodePayload,
   PullRequestNodePayload,
   CommentNodePayload,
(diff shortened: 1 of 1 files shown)


src/plugins/artifact/editor/adapters/githubPluginAdapter.js
```diff
@@ -79,26 +78,27 @@ const adapter: PluginAdapter<NodePayload> = {
...
-        extractCommentTitle("review comment", node),
-      PULL_REQUEST_REVIEW: extractPRReviewTitle,
-      USER: extractAuthorTitle,
-      ORGANIZATION: extractAuthorTitle,
-      BOT: extractAuthorTitle,
-    };
-    function fallbackAccessor(node: Node<NodePayload>) {
-      throw new Error(`unknown node type: ${node.address.type}`);
+    const anyNode: Node<any> = node;
+    const type: NodeType = (node.address.type: any);
```
decentralion: why the any-cast in `(node.address.type: any)`?

would it be cleaner to write `anyNode.address.type`?

PR #138, opened 2018-04-23
Split github/githubPlugin.js into three files

decentralion:
github/githubPlugin.js was growing ungainly - it contained two major
pieces: all of the node and edge types, and the GitHub parser. As I
contemplated adding a third major new section of logic (an easy-to-use
api for traversing the GitHub graph, with first class support for
comments, authorship, etc), I found the prospect of adding even more
into that file quite unappealing. So, I have instead split it into three
files:

* github/pluginName.js: Exports the plugin name.
* github/types.js: Exports types of nodes, edges, and their payloads
* github/parser.js: Exports `GithubParser`

No logic has been changed whatsoever - this is purely a rename-refactor.

Test plan:
CI still passes. I manually verified that the Artifact Editor can still
load and display GitHub data.


src/plugins/github/githubPlugin.test.js
```diff
@@ -1,10 +1,8 @@
...
-  GithubParser,
-  AUTHORS_EDGE_TYPE,
-  CONTAINS_EDGE_TYPE,
-} from "./githubPlugin";
+import {AUTHORS_EDGE_TYPE, CONTAINS_EDGE_TYPE} from "./types";
+
+import {GithubParser} from "./parser";
 import exampleRepoData from "./demoData/example-repo.json";
 
 describe("GithubParser", () => {
```
wchargin: Is this parser actually` parser.test.js` now?

PR #139, opened 2018-04-24
Load Git repositories into memory

wchargin:
Summary:
In this newly added module, we load the structural state of a git
repository into memory. We do not load into memory the contents of any
blobs, so this is not enough information to perform any analysis
requiring file diffing. However, it is sufficient to develop a notion of
“this file was changed in this commit”, by simply diffing the trees.

Test Plan:
Unit tests added; `yarn test` suffices. Reading these snapshots is
pretty easy, even though they’re filled with hashes:
  - First, read over the commit specifications on lines 69–83 of
    `loadRepository.test.js`, so you know what to expect.
  - In the snapshot file, keep handy the time-ordered list of commit
    SHAs at the bottom of the file, so that you know which commit SHA is
    which.
  - To verify that the large snapshot is correct: for each commit, read
    the corresponding tree object and make sure that the structure is
    correct.
  - To verify the small snapshot, just check that it’s the correct
    subset of the large snapshot.
  - If you want to verify that the SHA for a blob is correct, open a
    terminal and run `git hash-object -t blob --stdin`; then, enter the
    content of the blob and press `<C-d>`. The result is the blob SHA.

To run a sanity-check on a large repository: apply the following patch:

<details>
<summary>Patch to print out statistics about loaded repository</summary>

```diff
diff --git a/config/paths.js b/config/paths.js
index d2f25fb..8fa2023 100644
--- a/config/paths.js
+++ b/config/paths.js
@@ -62,5 +62,6 @@ module.exports = {
     fetchAndPrintGithubRepo: resolveApp(
       "src/plugins/github/bin/fetchAndPrintGithubRepo.js"
     ),
+    loadRepository: resolveApp("src/plugins/git/loadRepository.js"),
   },
 };
diff --git a/src/plugins/git/loadRepository.js b/src/plugins/git/loadRepository.js
index a76b66c..9380941 100644
--- a/src/plugins/git/loadRepository.js
+++ b/src/plugins/git/loadRepository.js
@@ -106,3 +106,7 @@ function findTrees(git: GitDriver, rootTrees: Set<Hash>): Tree[] {
   }
   return result;
 }
+
+const result = loadRepository(...process.argv.slice(2));
+console.log("commits", result.commits.size);
+console.log("trees", result.trees.size);
```
</details>

Then, run `yarn backend` and put the following script in `test.sh`:

<details>
<summary>Contents for `test.sh`</summary>

```shell
#!/bin/bash
set -eu

repo="$1"
ref="$2"

via_node() {
    node bin/loadRepository.js "${repo}" "${ref}"
}

via_git() (
    cd "${repo}"
    printf 'commits '
    git rev-list "${ref}" | wc -l
    printf 'trees '
    git rev-list "${ref}" |
        while read -r commit; do
            git rev-parse "${commit}^{tree}"
            git ls-tree -rt "${commit}" \
                | grep ' tree ' \
                | cut -f 1 | cut -d ' ' -f 3
        done | sort | uniq | wc -l
)

echo
printf 'Running directly via git...\n'
time a="$(via_git)"

echo
printf 'Running Node script...\n'
time b="$(via_node)"

diff -u <(cat <<<"${a}") <(cat <<<"${b}")
```
</details>

Finally, run `./test.sh /path/to/some/repo origin/master`, and verify
that it exits successfully (zero diff). Here are some timing results on
SourceCred and TensorBoard:

  - SourceCred: 0.973s via Node, 0.327s via git.
  - TensorBoard: 30.836s via Node, 6.895s via git.

For TensorFlow, running via git takes 7m33.995s. Running via Node fails
with an out-of-memory error after 39 minutes, with 10GB RAM and 4GB
swap. See details below.

<details>
<summary>
Full timing details, commit SHAs, and OOM error message
</summary>

```
+ ./test.sh /home/wchargin/git/sourcecred 01634aabcca3756b38e13aaf2f451cfbda2ad5ea

Running directly via git...

real	0m0.327s
user	0m0.016s
sys	0m0.052s

Running Node script...

real	0m0.973s
user	0m0.268s
sys	0m0.176s
+ ./test.sh /home/wchargin/git/tensorboard 7aa1ab9d60671056b8811b7099eec08650f2e4fd

Running directly via git...

real	0m6.895s
user	0m0.600s
sys	0m0.832s

Running Node script...

real	0m30.836s
user	0m3.216s
sys	0m10.588s
+ ./test.sh /home/wchargin/git/tensorflow 968addadfd4e4f5688eedc31f92a9066329ff6a7

Running directly via git...

real	7m33.995s
user	5m21.124s
sys	1m5.476s

Running Node script...
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x121a2cc [node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewFixedArray(int, v8::internal::PretenureFlag) [node]
 6: v8::internal::DeoptimizationInputData::New(v8::internal::Isolate*, int, v8::internal::PretenureFlag) [node]
 7: v8::internal::compiler::CodeGenerator::PopulateDeoptimizationData(v8::internal::Handle<v8::internal::Code>) [node]
 8: v8::internal::compiler::CodeGenerator::FinalizeCode() [node]
 9: v8::internal::compiler::PipelineImpl::FinalizeCode() [node]
10: v8::internal::compiler::PipelineCompilationJob::FinalizeJobImpl() [node]
11: v8::internal::Compiler::FinalizeCompilationJob(v8::internal::CompilationJob*) [node]
12: v8::internal::OptimizingCompileDispatcher::InstallOptimizedFunctions() [node]
13: v8::internal::Runtime_TryInstallOptimizedCode(int, v8::internal::Object**, v8::internal::Isolate*) [node]
14: 0x12dc8b08463d
```
</details>

wchargin-branch: load-git-repositories

### DIFF
--- a/src/plugins/git/loadRepository.js
+++ b/src/plugins/git/loadRepository.js
@@ -1,0 +1,121 @@
+/*
+ * Load a git repository into memory. This dumps the commit and tree
+ * data into a structured form. Contents of blobs are not loaded.
+ *
+ * If the repository contains file names that are not valid UTF-8
+ * strings, the result is undefined.
+ *
+ * Note: git(1) is a runtime dependency of this module.
+ */
+// @flow
+
+import {execFileSync} from "child_process";
+
+export type GitDriver = (args: string[], options?: ExecOptions) => string;
+type ExecOptions = Object; // close enough
+export function localGit(repositoryPath: string): GitDriver {
+  return function git(args: string[], options?: ExecOptions): string {
+    // Throws an Error on shell failure.
+    return execFileSync(
+      "git",
+      ["-C", repositoryPath, ...args],
+      options
+    ).toString();
+  };
+}
+
+export type Repository = {|
+  +commits: Map<Hash, Commit>,
+  +trees: Map<Hash, Tree>,
+|};
+export type Hash = string;
+export type Commit = {|
+  +hash: Hash,
+  +treeHash: Hash,
+|};
+export type Tree = {|
+  +hash: Hash,
+  +entries: Map<string, TreeEntry>, // map from name
+|};
+export type TreeEntry = {|
+  +type: "blob" | "commit" | "tree",
+  +name: string,
+  +hash: Hash,
+|};
+
+/**
+ * Load a Git repository from disk into memory. The `rootRef` should be
+ * a revision reference as accepted by `git rev-parse`: "HEAD" and
+ * "origin/master" will be common, while a specific SHA or tag might be
+ * used to fix a particular state of a repository.
+ */
+export function loadRepository(
+  repositoryPath: string,
+  rootRef: string
+): Repository {
+  const git = localGit(repositoryPath);
+  const commits = findCommits(git, rootRef);
+  const trees = findTrees(git, new Set(commits.map((x) => x.treeHash)));
+  return {commits: hashMap(commits), trees: hashMap(trees)};
+}
+
+function hashMap<T: {+hash: Hash}>(ts: $ReadOnlyArray<T>): Map<Hash, T> {
+  const result = new Map();
+  ts.forEach((t) => {
+    result.set(t.hash, t);
+  });
+  return result;
+}
+
+function findCommits(git: GitDriver, rootRef: string): Commit[] {
+  return git(["log", "--oneline", "--pretty=%H %T", rootRef])
+    .split("\n")
+    .filter((line) => line.length > 0)
+    .map((line) => {
+      const [hash, treeHash] = line.split(" ");
+      return {hash, treeHash};
+    });
+}
+
+function findTrees(git: GitDriver, rootTrees: Set<Hash>): Tree[] {
+  const result: Tree[] = [];
+  const visited: Set<Hash> = new Set();
+  const frontier: Set<Hash> = new Set(rootTrees);
+  while (frontier.size > 0) {
+    const next = frontier.values().next();
+    if (next.done) {
+      // Flow doesn't know that this is impossible, but it is.
+      throw new Error("Impossible! `frontier` had positive size.");
+    }
+    const treeHash: Hash = next.value;
+    visited.add(treeHash);
+    frontier.delete(treeHash);
+    const tree = loadTree(git, treeHash);
+    result.push(tree);
+    for (const entry of tree.entries.values()) {
+      if (entry.type === "tree" && !visited.has(entry.hash)) {
+        frontier.add(entry.hash);
+      }
+    }
+  }
+  return result;
+}
+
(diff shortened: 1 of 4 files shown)


src/plugins/git/loadRepository.test.js
```diff
@@ -0,0 +1,156 @@
...
+import mkdirp from "mkdirp";
+import path from "path";
+import tmp from "tmp";
+
+import type {GitDriver} from "./loadRepository";
+import {localGit, loadRepository} from "./loadRepository";
+
+const cleanups: (() => void)[] = [];
+afterAll(() => {
+  cleanups.forEach((f) => {
```
decentralion: (nit): why not `cleanups.forEach((f) => f())`?

src/plugins/git/loadRepository.test.js
```diff
@@ -0,0 +1,156 @@
...
+    f();
+  });
+});
+
+function mkdtemp() {
+  const result = tmp.dirSync();
+  cleanups.push(() => result.removeCallback());
```
decentralion: looks like this would be a really nice case to use python context managers (if we were in python) :)

src/plugins/git/loadRepository.test.js
wchargin: That lambda is equivalent to `(f) => { return f(); }`, which is not what I mean. In this case, the effect is the same (because JavaScript), but I prefer to follow Google style and only use the expression form when you actually want an expression.

src/plugins/git/loadRepository.test.js
wchargin: Yep: `with tempfile.TemporaryDirectory as tmpdir():` is great.

src/plugins/git/loadRepository.js
```diff
@@ -0,0 +1,109 @@
...
+    .split("\n")
+    .filter((line) => line.length > 0)
+    .map((line) => {
+      const [hash, treeHash] = line.split(" ");
+      return {hash, treeHash};
+    });
+}
+
+function findTrees(git: GitDriver, rootTrees: Set<Hash>): Tree[] {
+  const result: Tree[] = [];
```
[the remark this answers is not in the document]
wchargin: Noted. There are a number of potential reasonable formulations, which I’m happy to explore.

wchargin:
Pushing a no-op change to get an up-to-date Travis merge check.

PR #140, opened 2018-04-24
Add flow types for GitHub graphql query response

decentralion:
This commit adds flow typing for the JSON result from hitting the GitHub
graphql api. We can't prove that the flow typing is correct, but since
the type definition is colocated with the corresponding fragment
definitions, we can hope that maintainers will maintain both together.

We update the parser to consume the new flow types. There are no flow
errors.

Test plan:
Inspect the flowtypes, verify that they correspond to the data in
example-repo.json, and that there are no flow errors.

### DIFF
--- a/src/plugins/github/graphql.js
+++ b/src/plugins/github/graphql.js
@@ -93,6 +93,22 @@ export type Continuation = {|
   +destinationPath: $ReadOnlyArray<string | number>,
 |};
 
+export type ConnectionJSON<+T> = {|
+  +nodes: $ReadOnlyArray<T>,
+  +pageInfo: {|
+    +endCursor: ?string,
+    +hasNextPage: boolean,
+  |},
+|};
+
+export type RepositoryJSON = {|
+  +repository: {
+    +id: string,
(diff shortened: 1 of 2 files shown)


src/plugins/github/graphql.js
```diff
@@ -93,6 +93,22 @@ export type Continuation = {|
   +destinationPath: $ReadOnlyArray<string | number>,
 |};
 
+export type Connection<T> = {|
```
wchargin: `ConnectionJSON`?

wchargin: This should be covariant in `T`: namely, `ConnectionJSON<+T>`.

src/plugins/github/graphql.js
```diff
@@ -607,88 +623,179 @@ function mergeDirect<T>(destination: T, source: any): T {
   }
 }
 
-/**
- * These fragments are used to construct the root query, and also to
- * fetch more pages of specific entity types.
- */
-export function createFragments(): FragmentDefinition[] {
+export type AuthorJSON = {
```
wchargin: `@q` :-)

PR #141, opened 2018-04-24
Extract repository types to a separate module

wchargin:
Summary:
We should be able to get the types without depending on the function to
load a Git repo from disk, and in particular without depending on
`child_process`.

Test Plan:
Flow and tests are sufficient.

wchargin-branch: extract-repository-types

### DIFF
--- a/src/plugins/git/loadRepository.js
+++ b/src/plugins/git/loadRepository.js
@@ -11,6 +11,8 @@
 
 import {execFileSync} from "child_process";
 
+import type {Repository, Hash, Commit, Tree, TreeEntry} from "./types";
+
 export type GitDriver = (args: string[], options?: ExecOptions) => string;
 type ExecOptions = Object; // close enough
 export function localGit(repositoryPath: string): GitDriver {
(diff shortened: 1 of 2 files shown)


wchargin:
Some minutes from an offline discussion.

Point (in favor): If an external plugin wants to interface with the GitHub graph—say, create an edge from a GitHub user node to some FooPlugin node—then it will need to construct the appropriate ID, and it would be nice if it didn’t have to query a GitHub API to do so. Transparent IDs are helpful in this case.

Counterpoint: The most salient point (in my opinion) of the preceding statement is that external consumers need to have a way to construct IDs. It seems clear that this should be functionality provided by the GitHub plugin—something like `declare function userNodeAddress(login: string): Address`—and not just expecting clients to do string interpolation or something. This gives us more flexibility: the format is still opaque to clients, but clients have ways to work with these opaque formats.

Point (in favor): URLs are already part of GitHub’s contract. If GitHub changes the issue URLs to something else, the web breaks, and so GitHub is incentivized not to do that.

Counterpoint: True, but they could (and would) just add 301s and make the change anyway, so this doesn’t actually stop them from changing the URL format.

Related observation: It looks like GitHub’s GraphQL IDs are guaranteed to be stable; the [`ID` documentation][id] mentions that an ID is “often used to refetch an object or as key for a cache”; neither of these operations makes sense unless the ID is stable.

[id]: https://developer.github.com/v4/scalar/id/

Point (against): Making fundamental API decisions based on snapshot test legibility is clearly questionable.

Counterpoint: True, but legibility in general _is_ an important criterion, and snapshot test legibility is perhaps a good indicator of overall legibility.

Point (in favor): Our whole addressing API will almost certainly be overhauled due to things that we’ve learned since its creation. For instance, the address–payload distinction is very clunky, and we are looking into ways to improve it. So there’s not too much harm in making a decision now that we might reverse later.

Conclusions: We agree that the opaque-IDs-with-exposed-operations is a good paradigm. Given the last point about the early-alpha status of the addressing API, we feel comfortable making this change now with the understanding that we can feel free to change it later.

PR #145, opened 2018-04-26
Refactor GH parser to expose a functional api

decentralion:
Our GitHub parser is implemented via a `GithubParser` class which builds
the GitHub graph. This is a convenient implementation, but an awkward
API. This commit refactors the module so that it exposes a clean `parse`
function, which ingests the GitHub JSON data and returns as completed
graph.

Test plan:
The unit tests have been re-written to use the new public API. All the
snapshots are unchanged, and flow passes. Additionally, I ran `yarn
start` and verified that the GithubGraphFetcher for the Artifact plugin
is still working.

### DIFF
--- a/src/plugins/github/parser.test.js
+++ b/src/plugins/github/parser.test.js
@@ -1,15 +1,15 @@
 // @flow
 
 import {AUTHORS_EDGE_TYPE, CONTAINS_EDGE_TYPE} from "./types";
-
-import {GithubParser} from "./parser";
+import type {NodePayload, EdgePayload} from "./types";
+import {parse} from "./parser";
+import type {RepositoryJSON, PullRequestJSON, IssueJSON} from "./graphql";
+import {Graph} from "../../core/graph";
 import exampleRepoData from "./demoData/example-repo.json";
 
 describe("GithubParser", () => {
   describe("whole repo parsing", () => {
-    const parser = new GithubParser("sourcecred/example-repo");
-    parser.addData(exampleRepoData);
-    const graph = parser.graph;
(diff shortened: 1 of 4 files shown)


src/plugins/github/parser.test.js
```diff
@@ -66,38 +66,66 @@ describe("GithubParser", () => {
...
+    }
+    return selected[0];
+  }
+  function parseExample(
+    issueNums: number[],
+    prNums: number[]
+  ): Graph<NodePayload, EdgePayload> {
+    const issues = issueNums.map(getIssue);
+    const pullRequests = prNums.map(getPR);
+    const exampleData: RepositoryJSON = {
```
wchargin: Having this type annotation gives me a lot of confidence. :+1:

src/plugins/github/parser.test.js
```diff
@@ -66,38 +66,66 @@ describe("GithubParser", () => {
...
+  }
+  function getPR(n: number): PullRequestJSON {
+    const pulls = exampleRepoData.repository.pullRequests.nodes;
+    const selected = pulls.filter((x) => x.number === n);
+    if (selected.length !== 1) {
+      throw new Error(`Failure finding PR #${n}`);
+    }
+    return selected[0];
+  }
+  function parseExample(
```
wchargin: Two comments:
  - It's clear enough after reading the function implementation that this parses a specified _subset_ of the example repo, but this wasn't immediately obvious when just reading the signature.
  - I think that the clients are more readable with named arguments: `parseExample({issues: [1], prs: []})`. What do you think?

PR #147, opened 2018-04-27
Enable type filtering on Graph.get{In,Out}Edges

decentralion:
This commit adds an optional `typeOptions` argument to Graph.getInEdges
and Graph.getOutEdges. The `typeOptions` allow filtering the returned
edges by the type of the edge, and the type of the node that the edge is
connected to. This makes it much easier to use these methods to find
connections that have a certain relationship, e.g. finding the author of
a commit or the comments on an issue.

Test plan:
A new test suite was written that comprehensively tests this behavior,
both for getInEdges and getOutEdges.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -172,26 +172,52 @@ export class Graph<NP, EP> {
    * Gets the array of all out-edges from the node at the given address.
    * The order of the resulting array is unspecified.
    */
-  getOutEdges(nodeAddress: Address): Edge<EP>[] {
+  getOutEdges(
+    nodeAddress: Address,
+    typeOptions?: {+nodeType?: string, +edgeType?: string}
+  ): Edge<EP>[] {
     if (nodeAddress == null) {
       throw new Error(`address is ${String(nodeAddress)}`);
     }
-    return this._lookupEdges(this._outEdges, nodeAddress).map((e) =>
+    let result = this._lookupEdges(this._outEdges, nodeAddress).map((e) =>
       this.getEdge(e)
     );
+
+    if (typeOptions != null && typeOptions.edgeType != null) {
+      const edgeType = typeOptions.edgeType;
+      result = result.filter((e) => e.address.type === edgeType);
+    }
+    if (typeOptions != null && typeOptions.nodeType != null) {
+      const nodeType = typeOptions.nodeType;
+      result = result.filter((e) => e.dst.type === nodeType);
+    }
+    return result;
   }
 
   /**
    * Gets the array of all in-edges to the node at the given address.
    * The order of the resulting array is unspecified.
    */
-  getInEdges(nodeAddress: Address): Edge<EP>[] {
+  getInEdges(
+    nodeAddress: Address,
+    typeOptions?: {+nodeType?: string, +edgeType?: string}
+  ): Edge<EP>[] {
     if (nodeAddress == null) {
       throw new Error(`address is ${String(nodeAddress)}`);
     }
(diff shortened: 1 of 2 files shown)


src/core/graph.js
```diff
@@ -172,26 +172,52 @@ export class Graph<NP, EP> {
...
   }
 
   /**
    * Gets the array of all in-edges to the node at the given address.
    * The order of the resulting array is unspecified.
    */
-  getInEdges(nodeAddress: Address): Edge<EP>[] {
+  getInEdges(
+    nodeAddress: Address,
+    typeOptions?: {nodeType?: string, edgeType?: string}
```
wchargin: Can make these read-only, too.

src/core/graph.test.js
```diff
@@ -316,7 +316,93 @@ describe("graph", () => {
...
+              address: this.makeAddress(edgeType),
+              src: isInEdge ? node.address : this.root,
+              dst: isInEdge ? this.root : node.address,
+              payload: {},
+            };
+            this.graph.addEdge(edge);
+            return edge;
+          }
+        }
+        ["inEdges", "outEdges"].forEach((choice) => {
```
wchargin: Can get around the string-based dispatch, and avoid the `: any` casts, with something like:
```js
const branches = [{
  choice: "in-edges",
  edges: (egc) => egc.inEdges,
  getEdges: (egc) => (node, options) => egc.graph.getInEdges(node, options);
}, /* the other one */];
branches.forEach(() => { … });
```

src/core/graph.test.js
```diff
@@ -316,7 +316,93 @@ describe("graph", () => {
...
+          const egc = new ExampleGraphConstructor();
+          const A1 = (egc: any)[choice]["A1"];
```
wchargin: nit: mind lowercasing these, as standard variable names?

src/core/graph.test.js
```diff
@@ -316,7 +316,93 @@ describe("graph", () => {
       });
     });
 
-    describe("in- and out-edges", () => {
+    describe("getInEdges and getOutEdges", () => {
+      describe("type filtering", () => {
+        class ExampleGraphConstructor {
```
wchargin: This name is confusing to me. Isn't it more like an `ExampleGraph`? I expected to see `new ExampleGraphConstructor().createGraph()` or something.

src/core/graph.test.js
```diff
@@ -316,7 +316,93 @@ describe("graph", () => {
...
+            this.graph.addEdge(edge);
+            return edge;
+          }
+        }
+        ["inEdges", "outEdges"].forEach((choice) => {
+          const egc = new ExampleGraphConstructor();
+          const A1 = (egc: any)[choice]["A1"];
+          const A2 = (egc: any)[choice]["A2"];
+          const B1 = (egc: any)[choice]["B1"];
+          const B2 = (egc: any)[choice]["B2"];
```
wchargin: These can all be
```js
const {A1, A2, B1, B2} = (egc: any)[choice];
```

src/core/graph.js
decentralion: Done

src/core/graph.test.js
decentralion: Done

src/core/graph.test.js
decentralion: Ok

src/core/graph.test.js
decentralion: Done

PR #156, opened 2018-04-27
Include a submodule in the main example repository

wchargin:
Summary:
The main example repository now covers the currently desired features:
it has blobs, subtrees, and submodules, and commits that change each of
these. (We don’t have merge commits yet—we can add those once we start
to care about them.)

Once this is merged, I will push the two repositories to GitHub.

Test Plan:
Verifying and understanding is easier than ever before. You can run the
following commands to create the repositories in question on your disk:
```shell
$ yarn backend
$ node bin/createExampleRepo.js /tmp/repo
$ node bin/createExampleRepo.js --submodule /tmp/subrepo
```

You can then explore these repositories at your leisure. For instance,
to check that the `loadRepository` snapshot has the right set of
commits, inspect the output of the following command:
```shell
$ git -C /tmp/repo log --format='%H %T'
```
Or, to check that a particular tree has the right contents, just run:
```shell
$ git -C /tmp/repo ls-tree TREE_SHA
```
Verifying the `exampleRepo` snapshot is similarly easy: just check that
the lists of commit SHAs in `/tmp/repo` and `/tmp/subrepo` are correct.

wchargin-branch: include-submodule


src/plugins/git/demoData/exampleRepo.js
```diff
@@ -18,8 +37,40 @@ export function createExampleRepo(intoDirectory: string): RepositoryInfo {
...
+      "example-git\n",
+      "-----------\n\n",
+      "This repository provides example data for the SourceCred Git plugin.\n",
+      "Pay no attention to the contents behind the curtain.\n",
+    ].join()
+  );
   git.deterministicCommit("Initial commit");
+
+  git.writeAndStage("science.txt", "Amazing physics going on...\n");
+  commits.push(git.head());
```
decentralion: instead of manually managing the commits array (which allows desync, as seen here where we seem to be missing a commit?), lets have one function that commits and updates the array

wchargin: Good idea. Done.

PR #162, opened 2018-04-27
Replace `Graph.{getAdjacentEdges,getNeighborhood}`

decentralion:
`Graph.getAdjacentEdges` had a serious defect: for the adjacent edges,
it's hard to tell which of the {src,dst} is the neighboring node address
and which is the node we called `getAdjacentEdges` on.

This commit fixes that limitation by replacing `getAdjacentEdges` with
`getNeighborhood`, with a return signature of
`{edge: Edge<EP>, neighborAddress: Address}[]`

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -220,16 +220,27 @@ export class Graph<NP, EP> {
     return result;
   }
 
-  getAdjacentEdges(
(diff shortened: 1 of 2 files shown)


wchargin:
okay

PR #164, opened 2018-04-27
Collapse the 3 author types into 1

decentralion:
Currently, we store GitHub Users, Organizations, and Bots as separate
nodetypes in the graph. This is inconvenient, as we don't care very much
what type of entity authored a node.

This commit collapses those three categories into one nodetype. The
extra information has migrated to the node payload, so it is still
possible to discover this information if it's important.

Test plan: There is some amount of snapshot churn because the author
node types and payloads have changed. Expected snapshot changes: 
every edge pointing to a "USER" points to an "AUTHOR instead.
The two author nodes in the graph now have type "AUTHOR" and 
authorSubtype "USER".

Verify that the snapshot changes are appropriate, and that CI passes.

@wchargin: Red diffs :)

### DIFF
--- a/src/plugins/github/parser.js
+++ b/src/plugins/github/parser.js
@@ -13,6 +13,7 @@ import type {
   CommentNodePayload,
   PullRequestNodePayload,
   IssueNodePayload,
+  AuthorSubtype,
 } from "./types";
 
 import type {
@@ -76,11 +77,7 @@ class GithubParser {
     >,
     authorJson: AuthorJSON
   ) {
-    let authorPayload: AuthorNodePayload = {
-      login: authorJson.login,
-      url: authorJson.url,
-    };
-    let authorType: NodeType;
+    let authorType: AuthorSubtype;
     switch (authorJson.__typename) {
       case "User":
         authorType = "USER";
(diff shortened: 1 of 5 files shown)


src/plugins/github/types.js
```diff
@@ -47,33 +47,14 @@ export type PullRequestReviewCommentNodePayload = {|
...
 |};
 
-export const USER_NODE_TYPE: "USER" = "USER";
-export type UserNodePayload = {|
+export const AUTHOR_NODE_TYPE: "AUTHOR" = "AUTHOR";
+export type AuthorSubtype = "USER" | "BOT" | "ORGANIZATION";
+export type AuthorNodePayload = {|
   +login: string,
   +url: string,
+  +authorSubtype: AuthorSubtype,
```
wchargin: Can we just call this attribute `subtype`?

src/plugins/github/parser.js
```diff
@@ -98,9 +95,14 @@ class GithubParser {
           )}`
         );
     }
+    let authorPayload: AuthorNodePayload = {
```
wchargin: This can be `const` now. :-)

PR #167, opened 2018-04-28
Reorganize graph.test.js

decentralion:
This commit removes an unnecessary level of nesting in graph.test.js
(the "#graph" describe block inside the "graph" describe block), and
moves all the typefiltering logic into its own section.

Test plan: Absolutely no changes to code were made, except for moving
blocks of code around. The fact that tests still pass should be
sufficient.


decentralion:
Decided that this PR is unnecessary.

PR #169, opened 2018-04-28
Convert in-memory Git repos to cred graphs

wchargin:
Test Plan:
This snapshot test is too unwieldy to actually read—it’s 1000 lines of
opaque SHAs and thrice-stringified JSON objects—so it should be
interpreted as a regression test only. The programmatic tests should
suffice.

wchargin-branch: wip-git-create-graph


src/plugins/git/createGraph.js
```diff
@@ -0,0 +1,132 @@
...
+    const treeNode = {
+      address: this.makeAddress("TREE", tree.hash),
+      payload: {},
+    };
+    const result = new Graph().addNode(treeNode);
+    Object.keys(tree.entries).forEach((name) => {
+      const entry = tree.entries[name];
+      const entryNode = {
+        address: this.makeAddress(
+          "TREE_ENTRY",
```
decentralion: Here and elsewhere: use the variable rather than string constant

src/plugins/git/createGraph.test.js
```diff
@@ -0,0 +1,109 @@
...
+        nodeType: "TREE_ENTRY",
+        edgeType: "HAS_CONTENTS",
+      });
+      expect(graph.inEdges(address)).toHaveLength(
+        owningCommits.length + parentTreeEntries.length
+      );
+    });
+  });
+
+  it("has all tree entries, each with correct edges", () => {
```
decentralion: We don't have any testing for the fact that tree entries point to blobs, just that they point to something. Could we be a bit more specific?

wchargin:
> Would it be feasible to have a test case that examines a particular known path thru the repo?

I like this idea, and will implement it.

> Here and elsewhere: use the variable rather than string constant

Sure; will do.

> We don't have any testing for the fact that tree entries point to blobs, just that they point to something. Could we be a bit more specific?

This is a fair criticism. A solution is not immediately obvious to me. I will think about it.

src/plugins/git/createGraph.js
wchargin: Done.

src/plugins/git/createGraph.test.js
wchargin: Resolved with the test for `HEAD^{tree}:src/quantum_gravity.py`, I think.

PR #170, opened 2018-04-28
Add a porcelain api for Github data

decentralion:
Interacting with raw contribution graphs is cumbersome. We'll need
more fluent and convenient ways to retrieve data from them; we can do
this by creating porcelain APIs that wrap the underlying graph.

This commit adds a simple porcelain API for the GitHub data. It creates
the following classes:

* `api.Repository`
* `api.Issue`
* `api.PullRequest`
* `api.Comment`
* `api.Author`

The classes all wrap a graph and a nodeAddress. They provide read-only
functions for retreiving data from the graph; that data might be a part
of the node payload, or it might do some graph traversal under the hood.

The choice to have the wrapper hold onto the Address rather than the
node itself was deliberate; in the future, the graph may contain nodes
that are not synchronously reachable, so this approach allows us to
create wrappers for nodes we can't synchronously reach. When this comes
up in practice, we can then add async methods to the wrapper.

Note that some data already included in our graph, such as
PullRequestReviews and PullRequestReviewComments, were deliberately
excluded, so as to allow the core ideas to be reviewed without
unnecessary clutter.

Test plan:
Check that the unit tests appropriately test the behavior, and that the
API seems pleasant to use.

### DIFF
--- a/src/plugins/github/api.js
+++ b/src/plugins/github/api.js
@@ -1,0 +1,153 @@
+// @flow
+
+import {Graph} from "../../core/graph";
+import type {Node} from "../../core/graph";
+import type {Address} from "../../core/address";
+import type {
+  NodePayload,
+  EdgePayload,
+  NodeType,
+  IssueNodePayload,
+  PullRequestNodePayload,
+  CommentNodePayload,
+  AuthorNodePayload,
+  AuthorSubtype,
+} from "./types";
+
+import {
+  CONTAINS_EDGE_TYPE,
+  COMMENT_NODE_TYPE,
+  AUTHORS_EDGE_TYPE,
+  AUTHOR_NODE_TYPE,
+  ISSUE_NODE_TYPE,
+  PULL_REQUEST_NODE_TYPE,
+} from "./types";
+
+export class Repository {
+  repositoryName: string;
+  graph: Graph<NodePayload, EdgePayload>;
+
+  constructor(repositoryName: string, graph: Graph<NodePayload, EdgePayload>) {
+    this.repositoryName = repositoryName;
+    this.graph = graph;
+  }
+
+  issueOrPRByNumber(number: number): ?(Issue | PullRequest) {
+    let result: Issue | PullRequest;
+    this.graph.nodes({type: ISSUE_NODE_TYPE}).forEach((n) => {
+      if (n.payload.number === number) {
+        result = new Issue(this.graph, n.address);
+      }
+    });
+    this.graph.nodes({type: PULL_REQUEST_NODE_TYPE}).forEach((n) => {
+      if (n.payload.number === number) {
+        result = new PullRequest(this.graph, n.address);
+      }
+    });
+    return result;
+  }
+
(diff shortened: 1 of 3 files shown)


src/plugins/github/api.test.js
```diff
@@ -0,0 +1,76 @@
...
+      throw new Error("Issue reaching issue!");
+    }
+    expect(issue.title()).toBe("An example issue.");
+    expect(issue.body()).toBe("This is just an example issue.");
+    expect(issue.number()).toBe(1);
+    expect(issue.type()).toBe(ISSUE_NODE_TYPE);
+    expect(issue.url()).toBe(
+      "https://github.com/sourcecred/example-repo/issues/1"
+    );
+    expect(issue.type()).toMatchSnapshot();
```
wchargin: Why is this not necessarily `"ISSUE"`?

src/plugins/github/api.test.js
```diff
@@ -0,0 +1,76 @@
...
+    expect(issue.authors()[0].login()).toBe("decentralion");
```
wchargin: Stronger with `expect(issue.authors().map((x) => x.login())).toEqual(["decentralion"])`; up to you whether to change.

src/plugins/github/api.test.js
```diff
@@ -0,0 +1,76 @@
...
+  });
```
wchargin: Did you want to test the `issue.address()`, as with the other object types?

src/plugins/github/api.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  authors(): Author[] {
+    return this.graph
+      .nodes({type: AUTHOR_NODE_TYPE})
+      .map((n) => new Author(this.graph, n.address));
+  }
+}
+
+class GithubEntity<T: NodePayload> {
```
wchargin: It is somewhat strange that a repository is not a `GithubEntity`. I suppose that this is because it is not actually a node in the graph… should it be? (No change requested.)

src/plugins/github/api.test.js
decentralion: Meant to test address - fixed.

src/plugins/github/api.test.js
decentralion: Done

src/plugins/github/api.test.js
decentralion: Fixed.

src/plugins/github/api.js
decentralion: Having a repository node makes a lot of sense; we're almost sure to need them. Made an issue to track this suggestion. https://github.com/sourcecred/sourcecred/issues/171

decentralion:
Addressed your comments; also, per offline discussion, we now snapshot test the node and programatically verify that `Entity.address()` returns the node address.

Re your suggestion for composing the classes via higher-order functions: It's interesting, I think you'll agree we don't need the complexity at this time, but I'll keep it in mind for later.

PR #174, opened 2018-04-30
Remove Graph.{in,out}Edges

decentralion:
This method removes `Graph.inEdges` and `Graph.outEdges`. As a
replacement to these two functions, `graph.neighborhood` now takes an
optional `direction` flag, which can be set to `"IN" | "OUT" | "ANY"`.

This reduces the surface area of the Graph API, and means that the same
pattern can be used when requesting in or out neighbors as is used when
requesting all neighbors.

This change generates significant churn in the test files, and in some
cases the tests are less elegant / show historicity, as they were
written for the type signature of `{in,out}Edges`, which just returns an
array of edges, and now receive an array of neighbors. I think this is
acceptable, and it's not worth re-writing the test.

In many cases, replacing existing calls to `{in,out}Edges` in our actual
codebase resulted in cleaner code, as `neighborhood` successfully
abstracts over the common patterns that users of `{in,out}Edges` were
implementing.

As a fly-by refactor, I also changed the `neighborAddress` part of the
`neighborhood` return value to `neighbor`. It's a little less
descriptive, but it's more concise, and flow is there to help ensure
it's used correctly.

Test plan: Note that CI passes. Inspect the test changes, and verify
that they are appropriate transformations for consuming the new API.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -169,78 +169,70 @@ export class Graph<NP, EP> {
   }
 
   /**
-   * Gets the array of all out-edges from the node at the given address.
-   * The order of the resulting array is unspecified.
-   */
-  outEdges(
-    nodeAddress: Address,
-    typeOptions?: {+nodeType?: string, +edgeType?: string}
-  ): Edge<EP>[] {
-    if (nodeAddress == null) {
-      throw new Error(`address is ${String(nodeAddress)}`);
-    }
-    let result = this._lookupEdges(this._outEdges, nodeAddress).map((e) =>
-      this.edge(e)
-    );
-
-    if (typeOptions != null && typeOptions.edgeType != null) {
-      const edgeType = typeOptions.edgeType;
-      result = result.filter((e) => e.address.type === edgeType);
-    }
-    if (typeOptions != null && typeOptions.nodeType != null) {
-      const nodeType = typeOptions.nodeType;
-      result = result.filter((e) => e.dst.type === nodeType);
-    }
-    return result;
-  }
-
-  /**
-   * Gets the array of all in-edges to the node at the given address.
-   * The order of the resulting array is unspecified.
-   */
-  inEdges(
-    nodeAddress: Address,
-    typeOptions?: {+nodeType?: string, +edgeType?: string}
-  ): Edge<EP>[] {
-    if (nodeAddress == null) {
-      throw new Error(`address is ${String(nodeAddress)}`);
-    }
-    let result = this._lookupEdges(this._inEdges, nodeAddress).map((e) =>
-      this.edge(e)
-    );
-
-    if (typeOptions != null && typeOptions.edgeType != null) {
-      const edgeType = typeOptions.edgeType;
-      result = result.filter((e) => e.address.type === edgeType);
-    }
-    if (typeOptions != null && typeOptions.nodeType != null) {
-      const nodeType = typeOptions.nodeType;
-      result = result.filter((e) => e.src.type === nodeType);
-    }
-    return result;
-  }
-
-  /**
-   * Find the neighborhood of a given nodeAddress
+   * Find the neighborhood of the node at the given address.
    *
-   * By neighborhood, we mean every `{edge, neighborAddress}` such that edge
-   * has `nodeAddress` as either `src` or `dst`, and `neighborAddress` is the
+   * By neighborhood, we mean every `{edge, neighbor}` such that edge
+   * has `nodeAddress` as either `src` or `dst`, and `neighbor` is the
    * address at the other end of the edge.
+   *
+   * The returned neighbors are filtered according to the `options`. Callers
+   * can filter by nodeType, edgeType, and whether it should be an "IN" edge
+   * (i.e. the provided node is the dst), an "OUT" edge (i.e. provided node is
+   * the src), or "ANY".
    */
   neighborhood(
     nodeAddress: Address,
-    typeOptions?: {+nodeType?: string, +edgeType?: string}
-  ): {+edge: Edge<EP>, +neighborAddress: Address}[] {
-    const inNeighbors = this.inEdges(nodeAddress, typeOptions).map((e) => {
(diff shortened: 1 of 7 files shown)


src/core/graph.js
```diff
@@ -169,80 +169,76 @@ export class Graph<NP, EP> {
   }
 
   /**
-   * Gets the array of all out-edges from the node at the given address.
-   * The order of the resulting array is unspecified.
+   * Find the neighborhood of a given nodeAddress
+   *
+   * By neighborhood, we mean every `{edge, neighbor}` such that edge
+   * has `nodeAddress` as either `src` or `dst`, and `neighbor` is the
+   * address at the other end of the edge.
```
wchargin: …and such that the filter is satisfied.

src/core/graph.js
```diff
@@ -169,80 +169,76 @@ export class Graph<NP, EP> {
   }
 
   /**
-   * Gets the array of all out-edges from the node at the given address.
-   * The order of the resulting array is unspecified.
+   * Find the neighborhood of a given nodeAddress
```
wchargin: revenge of https://github.com/sourcecred/sourcecred/pull/162#discussion_r184813987

src/core/graph.js
```diff
@@ -169,80 +169,76 @@ export class Graph<NP, EP> {
...
-      throw new Error(`address is ${String(nodeAddress)}`);
+    if (direction === "ANY" || direction === "IN") {
+      let inNeighbors = this._lookupEdges(this._inEdges, nodeAddress).map(
+        (e) => {
+          const edge = this.edge(e);
+          return {edge, neighbor: edge.src};
+        }
+      );
+
+      if (typeOptions != null && typeOptions.nodeType != null) {
```
wchargin: It should be possible to factor this out to
```js
if (typeOptions != null && typeOptions.nodeType != null) {
  const nodeType = typeOptions.nodeType;
  result = result.filter(({neighbor}) => neighbor.type === nodeType);
}
```
and include the logic just once, above the `edgeType` filter.

Namely, when the following patch is applied, tests still pass:
<details>
<summary>Patch to deduplicate node type filtering logic</summary>

```diff
diff --git a/src/core/graph.js b/src/core/graph.js
index a1f8b89..f90bc31 100644
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -197,13 +197,6 @@ export class Graph<NP, EP> {
           return {edge, neighbor: edge.src};
         }
       );
-
-      if (typeOptions != null && typeOptions.nodeType != null) {
-        const nodeType = typeOptions.nodeType;
-        inNeighbors = inNeighbors.filter(
-          ({edge}) => edge.src.type === nodeType
-        );
-      }
       result = result.concat(inNeighbors);
     }
 
@@ -214,14 +207,6 @@ export class Graph<NP, EP> {
           return {edge, neighbor: edge.dst};
         }
       );
-
-      if (typeOptions != null && typeOptions.nodeType != null) {
-        const nodeType = typeOptions.nodeType;
-        outNeighbors = outNeighbors.filter(
-          ({edge}) => edge.dst.type === nodeType
-        );
-      }
-
       if (direction === "ANY") {
         // If direction is ANY, we already counted self-referencing edges as
         // an inNeighbor
@@ -229,9 +214,13 @@ export class Graph<NP, EP> {
           ({edge}) => !deepEqual(edge.src, edge.dst)
         );
       }
-
       result = result.concat(outNeighbors);
     }
+
+    if (typeOptions != null && typeOptions.nodeType != null) {
+      const nodeType = typeOptions.nodeType;
+      result = result.filter(({neighbor}) => neighbor.type === nodeType);
+    }
     if (typeOptions != null && typeOptions.edgeType != null) {
       const edgeType = typeOptions.edgeType;
       result = result.filter(({edge}) => edge.address.type === edgeType);
```
</details>

src/core/graph.test.js
```diff
@@ -441,14 +444,30 @@ describe("graph", () => {
...
+            "neighborhood OUT",
             exampleGraph.outEdges,
-            (opts) => exampleGraph.graph.outEdges(exampleGraph.root, opts),
+            (opts: any) => {
+              if (opts == null) {
+                opts = {};
+              }
+              opts["direction"] = "OUT";
+              return exampleGraph.graph
+                .neighborhood(exampleGraph.root, opts)
```
wchargin: Can replace `opts` with `{...(opts || {}), direction: "OUT"}` to save some lines, reduce mutation, and remove an `any`.

src/core/graph.test.js
```diff
@@ -692,7 +721,11 @@ describe("graph", () => {
             .addEdge(fullyDanglingEdge())
             .removeNode(danglingSrc().address)
             .removeNode(danglingDst().address);
-          const inEdges = g.inEdges(fullyDanglingEdge().dst);
+          const inEdges = g
+            .neighborhood(fullyDanglingEdge().dst, {
```
wchargin: A bunch of these objects can be collapsed onto the same line without Prettier complaining; it becomes more readable that way, IMO, but up to you.

src/core/graph.js
decentralion: Added a description of how the options work.

src/core/graph.js
decentralion: :)

src/core/graph.js
```diff
@@ -169,80 +169,72 @@ export class Graph<NP, EP> {
...
-   * The order of the resulting array is unspecified.
+   * Find the neighborhood of the node at the given address.
+   *
+   * By neighborhood, we mean every `{edge, neighbor}` such that edge
+   * has `nodeAddress` as either `src` or `dst`, and `neighbor` is the
+   * address at the other end of the edge.
+   *
+   * The returned neighbors are filtered according to the `options`. Callers
+   * can filter by nodeType, edgeType, and whether it should be an "IN" edge
+   * (i.e. the provided node is the dst), and "OUT" edge (i.e. provided node is
```
wchargin: `s/and/an/`

PR #178, opened 2018-05-01
Add commit parent edges in the Git graph

wchargin:
Test Plan:
To verify the snapshot change, either believe the programmatic tests, or
use the following script to verify that the right edges were added:
```bash
#!/bin/bash
set -eu
example_repo="$(mktemp -d)"
yarn backend >/dev/null 2>/dev/null
node bin/createExampleRepo.js "${example_repo}"
expected() {
    git -C "${example_repo}" log --format='%H %P' \
        | awk '{ if (NF > 1) { print $2 " " $1 } }' \
        | sort
}
actual() {
    git diff HEAD^..HEAD | grep -A 1 -F -e src -e dst \
        | sed -n 's/^+.*"id": "\(.\+\)".*/\1 /p' \
        | tr -d $'\n' | cat - <(echo) \
        | fold -s -w 82 | sed 's/ *$//' \
        | sort
}
diff -u <(expected) <(actual)
```

wchargin-branch: graph-commit-parents

### DIFF
--- a/src/plugins/git/types.js
+++ b/src/plugins/git/types.js
@@ -55,6 +55,28 @@ export type NodeType =
 
 // Edges
 
+// CommitNode -> CommitNode
+export const HAS_PARENT_EDGE_TYPE: "HAS_PARENT" = "HAS_PARENT";
+export type HasParentEdgePayload = {|
+  +parentIndex: number, // one-based
+|};
+export function hasParentEdgeId(
+  childCommitHash: Hash,
+  oneBasedParentIndex: number
+) {
+  if (
+    !isFinite(oneBasedParentIndex) ||
+    oneBasedParentIndex !== Math.floor(oneBasedParentIndex) ||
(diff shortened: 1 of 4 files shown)


src/plugins/git/types.js
```diff
@@ -55,6 +55,27 @@ export type NodeType =
...
+export const HAS_PARENT_EDGE_TYPE: "HAS_PARENT" = "HAS_PARENT";
+export type HasParentEdgePayload = {|
+  +parentIndex: number, // one-based
+|};
+export function hasParentEdgeId(
+  childCommitHash: Hash,
+  oneBasedParentIndex: number
+) {
+  if (
+    oneBasedParentIndex !== (oneBasedParentIndex | 0) ||
```
decentralion: Math.floor

wchargin: Sure; done, and added the necessary `isFinite` check.

PR #184, opened 2018-05-02
Store `TreeEntry` metadata in non-string form

wchargin:
Summary:
Prior to this commit, given a `Tree` node with an edge to a `TreeEntry`
node, there was no way to tell what the entry name was other than
parsing the ID (which should never be required). This adds appropriate
data to the payload of a `TreeEntry`, and also to the inclusion edge (so
that if you only have the edge, you don’t have to fetch the entry).

Test Plan:
Snapshot changes are readable.

wchargin-branch: treeentry-metadata

### DIFF
--- a/src/plugins/git/createGraph.js
+++ b/src/plugins/git/createGraph.js
@@ -1,14 +1,17 @@
 // @flow
 
 import type {Address} from "../../core/address";
+import type {Edge, Node} from "../../core/graph";
 import type {
-  Repository,
   Commit,
-  Tree,
-  NodePayload,
   EdgePayload,
-  NodeType,
   EdgeType,
+  IncludesEdgePayload,
+  NodePayload,
+  NodeType,
+  Repository,
+  Tree,
+  TreeEntryNodePayload,
 } from "./types";
 import {Graph, edgeID} from "../../core/graph";
 import {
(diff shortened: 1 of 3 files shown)


src/plugins/git/createGraph.js
```diff
@@ -103,21 +106,21 @@ class GitGraphCreator {
...
+      const entryNode: Node<TreeEntryNodePayload> = {
         address: this.makeAddress(
           TREE_ENTRY_NODE_TYPE,
           treeEntryId(tree.hash, entry.name)
         ),
-        payload: {},
+        payload: {tree: tree.hash, name},
       };
-      const entryEdge = {
+      const entryEdge: Edge<IncludesEdgePayload> = {
```
decentralion: Why is the tree hash in the entryNode payload? It should be in the treeNode payload. Then if you need the tree hash for an entry, you get it via porcelain over:
`graph.neighborhood(entryNode, {nodeType: "TREE", edgeType: "INCLUDES_EDGE", direction: "IN"})[0].hash`

wchargin: I included this mostly for consistency (putting both relevant fields in). You're right that it is easily available from the tree, and I didn't end up using it. I'd be happy to remove it if you like.

wchargin:
> Outside of the plugin that generated the ID, parsing the id is
> inappropriate.

Absolutely.

> However, for the Git plugin to know how to parse Git plugin ids is
> reasonable.

Fair. Let me clarify my claims, then:
 1. A plugin should theoretically be allowed to parse its own IDs,
    because it created them, controls the format, etc.
 2. A plugin may choose to make it a design goal that strings should
    never be parsed. In general, parsing information out of strings like
    this is considered a Bad Idea (or at least a Last Resort).
 3. The Git plugin indeed makes this design goal.

> Avoids a 2x redundancy in memory usage

I thought about this. I think that this will go away when we refactor
out the address/payload distinction, so I am not too worried about it.
If it truly does prove to be a big problem, we can revisit this.

Do you mind giving it a shot and keeping open the option to revert?
(This seems to be our MO for contentious changes, and it’s worked well
in the past. :-) )

src/plugins/git/createGraph.js
decentralion: Actually, it isn't available in the tree ([see here]). Could you remove it from the entryNodePayload and add it to the treeNodePayload?

[see here]: https://github.com/sourcecred/sourcecred/pull/184/files#diff-080791a1f5043a49563a9556411954e2L101

wchargin: It is not available in the payload, but it is available as the (literal) address ID. You raise a fair criticism: "if you're against parsing IDs in general, this shouldn't change just because the parsing function is the identity function in some cases." I agree. But, instead of putting in the work to expose everything on payloads, I think that I'd rather wait until we have some more direction on #190.

wchargin:
I will remove the `treeHash: Hash` from the tree entry node, per your request. Given that the tree ID is not currently used, do you still want me to add it to the tree node? I am happy to include it or not, as you prefer. (If you have no preference, I'd bias toward not including it, consistent with our "add things lazily" policy.)

decentralion:
"add things lazily" is ok with me.

PR #185, opened 2018-05-02
Load submodule URLs at each commit

wchargin:
Summary:
See new explanatory comment in `src/plugins/git/types.js`.

Shout-out to `git` for having an excellent command-line API:
the `--blob` argument to `git-config` is perfect for this situation.

Test Plan:
Snapshot updates are readable and sufficient.

wchargin-branch: load-submodule-urls

### DIFF
--- a/src/plugins/git/demoData/example-git.json
+++ b/src/plugins/git/demoData/example-git.json
@@ -5,6 +5,9 @@
             "parentHashes": [
                 "69c5aad50eec8f2a0a07c988c3b283a6490eb45b"
             ],
+            "submoduleUrls": {
+                "pygravitydefier": "https://github.com/sourcecred/example-git-submodule.git"
+            },
             "treeHash": "7be3ecfee5314ffa9b2d93fc4377792b2d6d70ed"
         },
         "69c5aad50eec8f2a0a07c988c3b283a6490eb45b": {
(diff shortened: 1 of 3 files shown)


wchargin:
> The commit description mentions an explanatory comment that doesn't
> seem to exist.

Ooh, you're right; the comment is in a later commit. Here it is (I'll
copy it into the commit description):

```
// In Git, a tree may point to a commit directly; in our graph, we have
// an explicit notion of "submodule commit", because, a priori, we do
// not know the repository to which the commit belongs. A submodule
// commit node stores the hash of the referent commit, as well as the
// URL to the subproject as listed in the .gitmodules file.
```

> I don't feel I have great context on this commit, and the testing is
> pretty light

Cool; thanks for noting. I actually think that the snapshot tests are
totally sufficient, for exactly the reasons that you describe. Feeling
that you lack context makes sense—most of my day yesterday was spent
playing around with Git's semantics and source code, exploring edgy edge
cases, discovering what kinds of approaches fundamentally cannot work,
and the like. (Interestingly, with Git I continue to find that it is
often the edgiest edge cases that inform the nicest general solutions.)
This commit is the product of that exploration, and so is necessarily
build on the resulting context. I don’t think that it makes sense to
write down everything that I learned/all conclusions that I came to in
code comments, but if you ever want to dive into this stuff together,
I’d be happy to do so (or, of course, to answer any specific questions).

PR #186, opened 2018-05-02
Create nodes for submodule commits

wchargin:
Summary:
Previously, a tree entry had exactly one `HAS_CONTENTS` edge, unless the
tree entry corresponded to a submodule commit, in which case we had no
information at all. Now, submodule commit tree entries point to zero or
more `SUBMODULE_COMMIT` nodes. In the vast majority of cases, there will
be exactly one such node—however, it is possible that a particular tree
entry could correspond to multiple submodules (clone two identical
submodules with different URLs) or none at all (some manually edited
`.gitmodules` or other corruption).

Test Plan:
The snapshot changes are easy enough to read and verify (two new nodes
and five new edges). Additionally, the path-to-submodule `createGraph`
test has been updated.

wchargin-branch: graph-submodule-urls

### DIFF
--- a/src/plugins/git/createGraph.js
+++ b/src/plugins/git/createGraph.js
@@ -3,29 +3,36 @@
 import type {Address} from "../../core/address";
 import type {Edge, Node} from "../../core/graph";
 import type {
+  BlobNodePayload,
   Commit,
   EdgePayload,
   EdgeType,
+  HasContentsEdgePayload,
+  Hash,
   IncludesEdgePayload,
   NodePayload,
   NodeType,
   Repository,
+  SubmoduleCommitPayload,
   Tree,
   TreeEntryNodePayload,
+  TreeNodePayload,
 } from "./types";
 import {Graph, edgeID} from "../../core/graph";
 import {
   BLOB_NODE_TYPE,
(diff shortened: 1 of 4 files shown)


src/plugins/git/createGraph.js
```diff
@@ -123,34 +165,50 @@ class GitGraphCreator {
...
+      contentsNodes.forEach((contentsNode) => {
+        const contentsEdge: Edge<HasContentsEdgePayload> = {
           address: this.makeAddress(
             HAS_CONTENTS_EDGE_TYPE,
             edgeID(entryNode.address, contentsNode.address)
           ),
           src: entryNode.address,
           dst: contentsNode.address,
-          payload: {},
+          payload: (({}: any): {||}),
```
decentralion: What does this mean?

wchargin: Yeah (\*sigh\*). This is to work around a long-standing bug in Flow: https://github.com/facebook/flow/issues/2977

wchargin: (The change is necessary in this commit because I added an explicit type to `contentsEdge` on the preceding line 201.)

PR #187, opened 2018-05-02
Add BECOMES edges in the Git graph

wchargin:
Summary:
If a commit causes a tree entry to change hash while keeping the same
name, we now add a BECOMES edge between the corresponding entries.

Details with submodules introduce some interesting edge cases, but
otherwise this is reasonably straightforward.

Test Plan:
Snapshot changes are readable enough to manually verify. Programmatic
tests also added.

wchargin-branch: becomes-edges


wchargin:
(For now, just rebasing this PR onto master.)

PR #188, opened 2018-05-02
Add options to not rebuild on shell script tests

wchargin:
Summary:
This can be useful for speed, but it can also be important for
correctness (at least theoretically): if we run both these scripts
concurrently, then we don’t want one of them to squash the `bin`
directory while the other is about to invoke an executable therein.

One might note that the diffs to the two files in this commit are
virtually identical, and indeed the files themselves are quite similar.
I’d prefer to keep the duplication for now; if we really need a Bash
snapshot testing framework, we can factor one out.

Test Plan:
Run each script with `--help`, with `--build` and `--no-build`, and with
and without `-u`.

wchargin-branch: optional-rebuild

### DIFF
--- a/src/plugins/git/loadRepositoryTest.sh
+++ b/src/plugins/git/loadRepositoryTest.sh
@@ -9,12 +9,14 @@ usage() {
   printf 'Flags:\n'
   printf '  -u|--updateSnapshot\n'
   printf '      Update the stored file instead of checking its contents\n'
+  printf ' --[no-]build\n'
+  printf '      Whether to run "yarn backend" before the test.\n'
+  printf '      Default is --build.\n'
   printf '  --help\n'
   printf '      Show this message\n'
 }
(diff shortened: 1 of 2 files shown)


wchargin:
The idiom is that boolean flags are true whenever nonempty, so `0` is
true. Why this instead of `0`/`1` or `true`/`false`? I don’t know,
really. I suppose that a benefit is that you don’t have to worry about
whether the falsy value is `0` or `false`? The two implementations are
```
[ -n "${thing:-}" ]  # -n = nonempty?
```
and
```
[ "${thing:-0}" != 0 ]
```
respectively, neither of which seems particularly offensive. Perhaps it
just comes down to “we had to pick one”. :-)

wchargin:
(Also, in this particular case it is an implementation detail only, so I really don't have a strong preference.)

PR #189, opened 2018-05-02
Implement a custom CI script

wchargin:
Summary:
This CI script accomplishes two tasks:
 1. It speeds up our build by parallelizing where possible.
 2. It opens the possibility for running Travis cron jobs.

Currently, this script by default does the same amount of work as our
current CI script. However, I’d like to move `yarn backend` into the
list of basic actions: a backend build failure should fail CI.

Note: this script is written to be executable directly by Node, so we
can’t use Flow types with the standard syntax. Instead, we use the
comment syntax: https://flow.org/en/docs/types/comments/

Test Plan:
The following should pass with useful output:
  - `npm run travis`
  - `GITHUB_TOKEN="your_github_token" npm run travis -- --full`

The following should fail with useful output:
  - `npm run travis -- --full` (fail)

To test different failure modes, it can be helpful to add
```js
    {id: "doomed", cmd: ["false"], deps: []},
    {id: "orphan", cmd: ["whoami"], deps: ["who", "are", "you"]},
```
to the list of `basicTasks` in `travis.js`.

To test performance:
```shell
$ time node ./config/travis.js >/dev/null 2>/dev/null

real    0m8.306s
user    0m20.336s
sys     0m1.364s

$ time bash -c \
>     'npm run check-pretty && npm run lint && npm run flow && CI=1 npm run test' \
>     >/dev/null 2>/dev/null

real    0m12.427s
user    0m13.752s
sys     0m0.804s
```
A 50% savings is not bad at all—and the raw time saved should only
improve from here on, as the individual steps start taking more time.

wchargin-branch: custom-ci

### DIFF
--- a/config/travis.js
+++ b/config/travis.js
@@ -1,0 +1,247 @@
+// @flow
+
+const chalk = require("chalk");
+const child_process = require("child_process");
+
+/*::
+type TaskId = string;
+type Task = {|
+  +id: TaskId,
+  +cmd: $ReadOnlyArray<string>,
+  +deps: $ReadOnlyArray<TaskId>,
+|};
+
+type TaskResult = {|
+  +id: TaskId,
+  +success: boolean,
+  +status: number,
+  +stdout: string,
+  +stderr: string,
+|};
+*/
+
+function main() {
+  const mode =
+    process.env["TRAVIS_EVENT_TYPE"] === "cron" ||
+    process.argv.includes("--full")
+      ? "FULL"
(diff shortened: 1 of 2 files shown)


wchargin:
Here is the output of a successful build:
https://travis-ci.org/sourcecred/sourcecred/builds/374155945

wchargin:
Here is a screenshot of the build output on my terminal:
Screenshot

PR #192, opened 2018-05-03
Run `flow check` instead of `flow` in CI

wchargin:
Summary:
This means that Flow will check from scratch, instead of merging in
incremental stages and talking to an existing server.

This improves test isolation. If your existing Flow server is in a
different state than a new Flow server would be, then without this patch
you might not get correct results. (This can happen if, for instance,
you add additional `flow-typed` declarations.)

A side benefit is that it gets rid of the dozens of lines of log spam of
the form, “Please wait. Server is initializing (parsed files 12345)”.

A minor benefit is that this fixes a memory leak on CI—previously, the
Travis instance would have a useless Flow server running in the
background after checks complete. But I’m sure that Travis doesn’t
actually care, so this isn’t very important.

A downside of this change is that it will often make `yarn travis`
slower when run locally. Build times on a Travis instance should not be
affected, because Travis has to start a server either way, and that’s
the slow part. For me, a `yarn flow check` takes about 11s, so the build
must take at least that long.

[1]: https://stackoverflow.com/questions/38902752/

Test Plan:
Run `yarn travis` and note that it still passes. Then, introduce a Flow
error and note that the build properly fails.

wchargin-branch: ci-flow-check

### DIFF
--- a/config/travis.js
+++ b/config/travis.js
@@ -206,8 +206,8 @@ function makeTasks(mode /*: "BASIC" | "FULL" */) {
       deps: [],
     },
     {
-      id: "flow",
-      cmd: ["npm", "run", "--silent", "flow"],
+      id: "flow-check",
+      cmd: ["npm", "run", "--silent", "flow", "check"],
       deps: [],
     },
     {


wchargin:
Given the downside mentioned in the summary, I am happy to consider not merging this.

decentralion:
My testing. Given that a flow server was already running, running `yarn travis` took:
10s on master
15s with this change

This is a pretty small difference, but this PR is also a pretty small improvement. Let's think about this in terms of expected time savings.

Suppose that a contributor is in the habit of running `yarn travis` manually before submitting to travis, and they do this, on average, three times per push. Then they lose 15s per push.

If they do get in a case where their local build incorrectly passes due to stale flow data, then they lose, say, 600 seconds.

In which case, this bug would need to happen >2% of the time to have temporal breakeven. In practice, I think it's never happened, and based on having 160 PRs I think we can safely estimate that the failure prevented by this PR comes up <1% of the time.

Also, I think that the cost of increasing local test time by 50% is worse-than-linear - it makes it more likely that contributors will opt for some other ad hoc workflow, thus potentially introducing new (and more probable) spurious passes.

So, let's not merge this until we discover that the problem it would solve comes up often in practice.

wchargin:
Reasoning makes sense to me. Thanks.

Issue #190, opened 2018-05-02
Consider refactoring away the address–payload distinction


wchargin:
Currently, a node has both an address and a payload. The address ID
needs to uniquely identify the node, and the payload holds anything else
that we want to store.

This can be confusing: if you have a node and want to access a field,
you have to think about where the data lies. It’s a leaky abstraction.

Also, as the address ID must be a string, this often leads to “stringify
the whole shebang and put it in the payload”. If you want to get the
data back out of the node, you have to either additionally store it in
the payload (wasting space and causing extra confusion) or parse the ID
(slow, breaks typing, and generally considered a Questionable Idea).

We should consider eliminating this distinction by doing something more
reasonable.

(This all applies to edges, too.)

wchargin:
Here is one idea: nodes are arbitrary JSON blobs with a `type`
identifier; plugins register handlers that take nodes of a fixed `type`
to a unique address.

In more detail, suppose that we define nodes as follows (names up for
discussion):
```js
interface Node {
  __type: {|
    +repositoryName: string,
    +pluginName: string,
    +type: string,
  |};
}
```
Then, a plugin can register a handler:
```
type Id = string;
interface AddressHandler<T> {
  (node: Node<T>): Id;
}
type AddressHandlers = {[type: string]: AddressHandler};
```

For simple nodes, this can just be a projection:
```js
type Commit: Node = {|
  __type: {|
    +repositoryName: string,
    +pluginName: "sourcecred/git-beta",
    +type: "COMMIT",
  |},
  hash: string,
  summary: string,
|};
function handleCommit(commit: Commit) {
  // Commits can be uniquely identified by hash.
  return commit.hash;
}
```
Something more complicated, like `JSON.stringify(someFieldsOfNode)`,
might be necessary for more complicated nodes.

In some sense, this addressing function can be considered an
optimization, because `JSON.stringify(theWholeNode)` should always be a
safe (i.e., injective) implementation. This implementation has a few
downsides, though. First, memory performance: using a commit’s full
summary, or a blob’s full contents, as an address will lead to a lot of
wasted space. This is especially true for nodes with lots of incident
edges. Second, decreased lookup potential: using `handleCommit` above,
one can write
```js
graph.node(gitPlugin.commitNodeAddress(someHash)).summary
```
to find the summary of a commit for which only the hash is known. (This
pattern obviously generalizes.) If the content-addressing requires the
whole object, then this is no longer feasible. It could be argued that
[...]

PR #195, opened 2018-05-03
Remove the `repositoryName` field from Addresses

decentralion:
See [#190] for context.

The change is almost entirely straightforward; the only "interesting"
decision I made was to move the repo owner and repo name into the string
id for the Artifact Plugin addresses, as the id would otherwise not be
unique.

[#190]: https://github.com/sourcecred/sourcecred/issues/190#issuecomment-386362870


wchargin:
Nice that we have, in the artifact plugin, an example where it is important that the `repositoryName` be considered, and that it works out as planned.

PR #199, opened 2018-05-03
Expose `commitAddress` from the Git plugin

decentralion:
For the GitHub plugin to create edges pointing to commits from the Git
plugin, it needs a way to create the appropriate address given the
commit's hash. This commit exposes that functionality by moving
`makeAddress` out of the "createGraph" module and into a new "address"
module, and using it to implement `commitAddress`.

Test plan: The code is so trivial that I don't think it merits testing.

### DIFF
--- a/src/plugins/git/address.js
+++ b/src/plugins/git/address.js
@@ -1,0 +1,17 @@
+// @flow
+
+import type {Address} from "../../core/address";
+import type {EdgeType, NodeType} from "./types";
+import {COMMIT_NODE_TYPE, GIT_PLUGIN_NAME} from "./types";
+
+export function _makeAddress(type: NodeType | EdgeType, id: string): Address {
+  return {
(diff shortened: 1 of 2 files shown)


src/plugins/git/address.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import type {Address} from "../../core/address";
+import type {EdgeType, NodeType} from "./types";
+import {COMMIT_NODE_TYPE, GIT_PLUGIN_NAME} from "./types";
+
+export function makeAddress(type: NodeType | EdgeType, id: string): Address {
```
wchargin: As discussed, this should be package-private: it enables clients who only call the public API to get invalid addresses (by supplying invalid IDs). But JavaScript doesn't have a package-private visibility option. Please either (a) underscore-name this, or (b) export all the individual `fooAddress` functions instead of this.

PR #201, opened 2018-05-03
Find BECOMES edges for high-level commits

wchargin:
Test Plan:
Unit tests included. I verified that the hashes in the snapshot are
correct.

wchargin-branch: high-level-becomes-commits

### DIFF
--- a/src/plugins/git/createGraph.js
+++ b/src/plugins/git/createGraph.js
@@ -196,6 +196,59 @@ class GitGraphCreator {
   }
 }
 
+export type BecomesEdge = {|
+  +from: {|
+    +tree: Hash,
+    +name: string,
+  |},
+  +to: {|
+    +tree: Hash,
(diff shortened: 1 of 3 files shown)


src/plugins/git/createGraph.js
```diff
@@ -196,6 +196,59 @@ class GitGraphCreator {
   }
 }
 
+export type BecomesEdge = {|
+  +from: {|
```
decentralion: the type signature implies it's possible that `from.name !== to.name`. My understanding is that with the current algorithm this is impossible, however in the future (rename detection) that would be a valuable relationship to express. is this correct?

decentralion: confirmed offline that this is the case :)

wchargin: (Can confirm confirmation.)

PR #203, opened 2018-05-03
Add BECOMES edges in the Git graph

wchargin:
Summary:
If a commit causes a tree entry to change hash while keeping the same
name, we now add a BECOMES edge between the corresponding entries.

Test Plan:
Snapshot changes are readable enough to manually verify. Programmatic
tests also added.

wchargin-branch: graph-becomes-edges

### DIFF
--- a/src/plugins/git/createGraph.test.js
+++ b/src/plugins/git/createGraph.test.js
@@ -2,14 +2,18 @@
 
 import cloneDeep from "lodash.clonedeep";
 
+import type {Address} from "../../core/address";
+import type {Edge} from "../../core/graph";
 import type {BecomesEdge} from "./createGraph";
-import type {Hash, Tree} from "./types";
+import type {BecomesEdgePayload, Hash, Tree} from "./types";
+import {_makeAddress} from "./address";
 import {
   createGraph,
   findBecomesEdges,
   findBecomesEdgesForCommits,
 } from "./createGraph";
 import {
+  BECOMES_EDGE_TYPE,
   BLOB_NODE_TYPE,
   COMMIT_NODE_TYPE,
   GIT_PLUGIN_NAME,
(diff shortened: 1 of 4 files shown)


src/plugins/git/createGraph.test.js
```diff
@@ -143,11 +146,67 @@ describe("createGraph", () => {
...
+      .map((edge) => ((edge: Edge<any>): Edge<BecomesEdgePayload>));
+    expect(becomings).not.toHaveLength(0);
+    becomings.forEach((edge) => {
+      expect(edge.dst).not.toEqual(edge.src);
+      const {payload: {childCommit, parentCommit, path}} = edge;
+      expect(path).not.toHaveLength(0);
+      expect(data.commits[childCommit].parentHashes).toEqual(
+        expect.arrayContaining([parentCommit])
+      );
+      function expectedTreeEntryAddress(commit: Hash): Address {
```
decentralion: Let's test that edge.src and edge.dst are valid addresses to nodes in the graph, and those nodes have expected properties.

src/plugins/git/createGraph.test.js
```diff
@@ -143,11 +146,67 @@ describe("createGraph", () => {
...
+              );
+            }
+            return {
+              tree: data.trees[tree].entries[name].hash,
+              name: newName,
+            };
+          },
+          {tree: data.commits[commit].treeHash, name: path[0]}
+        );
+        return {
```
decentralion: use _makeAddress

src/plugins/git/createGraph.test.js
wchargin: Sure; done.

src/plugins/git/createGraph.test.js
wchargin: Yep; done.

PR #208, opened 2018-05-04
Extract `execDependencyGraph` core from CI script

wchargin:
Summary:
We’d like to use the same abstraction for creating multiple cred graphs
and then combining them together. This will enable us to do that.

Test Plan:
Run `yarn travis` to test the success case, and `yarn travis --full`
(without setting a `GITHUB_TOKEN`) to test the failure case.

wchargin-branch: execdepgraph


wchargin:
Recommend reviewing by locally checking out this commit and then running `git diff HEAD^:config/travis.js HEAD:src/tools/execDepgraph.js`.

wchargin:
Or `git difftool …` if you want vimdiff. :-)

PR #217, opened 2018-05-05
Implement a command-line interface

wchargin:
Summary:
This commit implements the `sourcecred` command-line utility, which has
three subcommands:
  - `plugin-graph` creates one plugin’s graph;
  - `combine` combines multiple on-disk graphs; and
  - `graph` creates all plugins’ graphs and combines them.

As an implementation detail, the `into.sh` script is very convenient,
avoiding needing to do any pipe management in Node (which is Not Fun).
When we build for release, we may want to factor that differently.

Test Plan:
To see it all in action, run `yarn backend`, and then try:
```
$ export SOURCECRED_GITHUB_TOKEN="your_token_here"
$ node ./bin/sourcecred.js graph sourcecred sourcecred
Using output directory: /tmp/sourcecred/sourcecred

Starting tasks
  GO   create-git
  GO   create-github
 PASS  create-github
 PASS  create-git
  GO   combine
 PASS  combine

Full results
 PASS  create-git
 PASS  create-github
 PASS  combine

Overview
Final result:  SUCCESS

$ ls /tmp/sourcecred/sourcecred/
graph-github.json  graph-git.json  graph.json

$ jq '.nodes | length' /tmp/sourcecred/sourcecred/*.json
1000
7302
8302
```
The `node sourcecred.js graph` command takes 9.8s for me.

(The salient point of the last command is that the two small graphs have
node count adding up to the node count of the big graph. Incidentally,
we are [almost][1] at a nice round number of nodes in the GitHub graph.)

[1]: https://xkcd.com/1000/

wchargin-branch: cli


src/cli/commands/combine.js
```diff
@@ -0,0 +1,41 @@
...
+import {Graph} from "../../core/graph";
+
+export default class CombineCommand extends Command {
+  static description = "combine multiple contribution graphs into one big graph";
+
+  // We disable strict-mode so that we can accept a variable number of
+  // arguments. Although this is what the docs [1] suggest, it doesn't
+  // seem like the best way to do it. TODO: Investigate.
+  //
+  // [1]: https://oclif.io/docs/commands.html
```
decentralion: We could file an issue on oclif asking for a way to do this without disabling strict.

src/cli/commands/graph.js
```diff
@@ -0,0 +1,106 @@
...
+      name: "repo_name",
+      required: true,
+      description: "name of the GitHub repository for which to fetch data",
+    },
+  ];
+
+  static flags = {
+    "output-directory": flags.string({
+      short: "o",
+      description: "directory into which to store graphs",
```
decentralion: In our previous example,we computed an appropriate temporary output dir and then set it as the default here, which would make the command usage more helpful. is there a reason not to do that here?

src/cli/commands/graph.js
```diff
@@ -0,0 +1,106 @@
...
+) {
+  mkdirp.sync(outputDirectory);
+  const tasks = makeTasks(outputDirectory, {repoOwner, repoName, token});
+  execDependencyGraph(tasks).then(({success}) => {
+    process.exitCode = success ? 0 : 1;
+  });
+}
+
+function makeTasks(outputDirectory, {repoOwner, repoName, token}) {
+  const plugins = ["git", "github"];
```
decentralion: Let's factor this out somewhere to mandate consistency between this and `PluginGraphCommand`?

src/cli/commands/pluginGraph.js
```diff
@@ -0,0 +1,85 @@
...
+      required: true,
+      description: "name of the GitHub repository for which to fetch data",
+    },
+  ];
+
+  static flags = {
+    plugin: flags.string({
+      description: "plugin whose graph to generate",
+      required: true,
+      options: ["git", "github"],
```
decentralion: also factor to a constant somewhere

decentralion: also, I'm sure you loved seeing that "multiple" [is an option that defaults to false](https://oclif.io/docs/flags.html)

src/cli/commands/graph.js
decentralion: Also: `"env": SOURCECRED_OUTPUT_DIRECTORY`? :smiley:

src/cli/commands/combine.js
wchargin: Agreed; I was thinking the same thing. This seems okay to me for now, though.

src/cli/commands/graph.js
wchargin: > In our previous example,we computed an appropriate temporary output dir and then set it as the default here, which would make the command usage more helpful. is there a reason not to do that here?
 
On line 49 (`const outputDirectory = …`) below, we do do that. We do not set an explicit default value in the flags configuration because the default value depends on the repository name.

wchargin: > Also: `"env": SOURCECRED_OUTPUT_DIRECTORY`? :smiley: 

Sure, can do.

src/cli/commands/graph.js
wchargin: Sure.

src/cli/commands/pluginGraph.js
wchargin: I did enjoy seeing that. I considered using that instead of `strict: false` for `combine`, but I believe that that would require writing `-f graph1 -f graph2 -f graph3 …`, instead of the more acceptable `-f graph1 graph2 graph3`, which is in turn inferior to just `graph1 graph2 graph3`.

decentralion: :+1:

src/cli/commands/graph.js
wchargin: > > In our previous example,we computed an appropriate temporary output dir and then set it as the default here, which would make the command usage more helpful. is there a reason not to do that here?
>
> On line 49 (`const outputDirectory = …`) below, we do do that. We do not set an explicit default value in the flags configuration because the default value depends on the repository name.

I'll change this to have the following semantics. The output directory defaults to `path.join(os.tmpdir(), "sourcecred")`, and graphs are always put in `path.join(outputDirectory, repoOwner, repoName)`, regardless of whether `outputDirectory` is manually set or the default value. This seems like better behavior.

decentralion: :+1:

src/cli/commands/graph.js
```diff
@@ -0,0 +1,103 @@
...
+import mkdirp from "mkdirp";
+import os from "os";
+import path from "path";
+
+import {pluginNames} from "./common";
+
+const execDependencyGraph = require("../../tools/execDependencyGraph").default;
+
+export default class GraphCommand extends Command {
+  static description = "create the contribution graph for a repository";
```
decentralion: [optional] I'd like to see more documentation here. Specifically, it should document the output format, that the actual resultant graph is `outputdir/repoOwner/repoName/graph.json` and that the other json files are intermediary artifacts. Note that multiline descriptions are supported.

I mark this [optional] since it's reasonably easy to infer by inspecting the output directory, and this is likely to be a power user feature (regular users will prefer the method that starts sourcecred and calls this method under-the-hood)

src/cli/commands/graph.js
```diff
@@ -0,0 +1,103 @@
...
+        "https://github.com/settings/tokens/new",
+      required: true,
+      env: "SOURCECRED_GITHUB_TOKEN",
+    }),
+  };
+
+  async run() {
+    const {
+      args: {repo_owner: repoOwner, repo_name: repoName},
+      flags: {"github-token": token, "output-directory": outputDirectory},
```
decentralion: this is cool btw, I didn't realize destructuring assignment could be used this way. I'm really a fan of destructuring assignment now, it can make for some really clean code.

src/cli/commands/common.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+
+export type PluginName = "git" | "github";
```
decentralion: I don't think this file is in the right spot.
- the set of standard plugins isn't a CLI-specific concept, it is generic to the project as a whole
- even if it were CLI specific, I think the "commands" subdir loses its meaning if we put files there that don't correspond to commands. (note: historically, the reason I created a commands/ dir was to satisfy oclif's requirement that there be a 1:1 mapping between commands in a command dir and actual commands; however, since we have arbitrary path redirection, that isn't a constraint on our source tree. but, I think it's still nice to maintain that organization.)

decentralion: [offline discussion]: until we have a need for this elsewhere, I'm happy to have it under `cli/` (just not `cli/commands`)

Issue #171, opened 2018-04-28
Create a node in the GitHub graph for the whole repository


decentralion:
As proposed by @wchargin in https://github.com/sourcecred/sourcecred/pull/170#discussion_r184839308

decentralion:
I do believe this is now implemented! :)

PR #219, opened 2018-05-07
Split `GithubResponseJSON` and `RepositoryJSON`

decentralion:
Currently, we generate a `RepositoryJSON` object via querying GitHub.
That `RepositoryJSON` object has a `repository` field... which is weird,
and suggests we got the names slightly wrong.

This commit renames the top-level response to `GithubResponseJSON`, and
factors the `repository` field out as `RepositoryJSON`. Correspondingly,
the `addData` and `addRepository` methods on the parser are now
distinct.

This is a precursor for #171.

Test plan: This is a simple refactor; the fact that yarn travis passes
should be sufficient.


decentralion:
Verified.

PR #224, opened 2018-05-07
CLI commands error on unhandled promise rejection

decentralion:
Previously, if a CLI command had an unhandled promise rejection, this
would result in a spurious success and zero exit value.

This commit causes all of our CLI commands to instead fail if they have
an unhandled promise rejection.

Test plan: Previously, `sourcecred graph src-d go-git` would claim to
succeed, although it actually fails due to an unrelated bug. After this
change is applied, it correctly fails to retrieve the GitHub graph (and
hte combine step is never run).

### DIFF
--- a/src/cli/commands/combine.js
+++ b/src/cli/commands/combine.js
@@ -7,6 +7,13 @@ import {promisify} from "util";
 
 import {Graph} from "../../core/graph";
 
+// Makes the script crash on unhandled rejections instead of silently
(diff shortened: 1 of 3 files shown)


wchargin:
An easier test plan is `SOURCECRED_GITHUB_TOKEN=notokenforyou node bin/sourcecred.js sourcecred example-github`, right?

Looks good, though.

PR #227, opened 2018-05-07
Extract a common SOURCECRED_DIRECTORY flag

wchargin:
Summary:
This solves two problems:

 1. The “output directory” argument to `sourcecred graph` is also the
    input directory to other commands, like `sourcecred analyze`
    (hypothetically). In such cases, it would be nice for the flag to
    have the same name, but clearly `--output-directory` does not always
    make sense.

 2. In addition to storing graphs, we’ll need to store other kinds of
    data: settings, intermediate data for plugins to cache results, etc.
    We should store these under a single umbrella.

With both of these problems in mind, it makes sense to create a
`SOURCECRED_DIRECTORY` flag under which we store all relevant data.

Test Plan:
Run:
```shell
$ yarn backend
$ node bin/sourcecred.js help graph
$ node bin/sourcecred.js graph sourcecred example-github
$ node bin/sourcecred.js graph sourcecred example-github -d /tmp/sorcecrod
$ SOURCECRED_DIRECTORY=/tmp/srccrd node bin/sourcecred.js graph sourcecred example-github
$ for dir in /tmp/{sourcecred,sorcecrod,srccrd}; do find "${dir}"; done
```

wchargin-branch: graph-directory

### DIFF
--- a/src/cli/commands/graph.js
+++ b/src/cli/commands/graph.js
@@ -4,7 +4,7 @@ import {Command, flags} from "@oclif/command";
 import mkdirp from "mkdirp";
 import path from "path";
 
-import {defaultStorageDirectory, pluginNames} from "../common";
+import {pluginNames, sourcecredDirectoryFlag} from "../common";
 
 const execDependencyGraph = require("../../tools/execDependencyGraph").default;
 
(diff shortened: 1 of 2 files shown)


decentralion:
Let's make it "SOURCECRED_DIRECTORY" instead. That future proofs us for putting other stuff there too.

wchargin:
Substantially rewritten. Please take a look.

PR #228, opened 2018-05-07
Deleted users no longer break GitHub parser

decentralion:
When a GitHub user delete their account, all of their comments remain,
but with a `null` author. Previously, our code did not account for this
possibility. Now it does (by simply not adding an AUTHORSHIP edge).
Conveniently, our porcelain API already represents authors as a list, so
this doesn't require any change in porcelain API usage.

Test plan:
I did not add any unit tests, simply because
creating-and-deleting a GitHub user to make a repro seemed like a bit of
a pain. However, it is very unlikely that this bug will re-occur,
because the nullability of AuthorJSON is now enforced at the type level
inside graphql.js.

Also, running `node bin/sourcecred.js src-d go-git` now succeeds instead
of failing.

### DIFF
--- a/src/plugins/github/graphql.js
+++ b/src/plugins/github/graphql.js
@@ -631,6 +631,11 @@ function mergeDirect<T>(destination: T, source: any): T {
   }
 }
 
+// If a user deletes their account, then the author is null
+// (the UI will show that the message was written by @ghost).
+// Therefore, NullableAuthorJSON is preferred to AuthorJSON
+// for most actual usage.
+export type NullableAuthorJSON = AuthorJSON | null;
 export type AuthorJSON = {|
   +__typename: "User" | "Bot" | "Organization",
   +id: string,
@@ -667,7 +672,7 @@ export type IssueJSON = {|
   +title: string,
   +body: string,
   +number: number,
-  +author: AuthorJSON,
+  +author: NullableAuthorJSON,
   +comments: ConnectionJSON<CommentJSON>,
 |};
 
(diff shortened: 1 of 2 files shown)


wchargin:
I believe that you checked all the places in the schema where an `Actor` is returned, and verified that each has the correct nullability. Is this correct? (If so—thanks!)

src/plugins/github/parser.js
```diff
@@ -76,8 +76,11 @@ class GithubParser {
       | PullRequestReviewCommentNodePayload
       | PullRequestReviewNodePayload
     >,
-    authorJson: AuthorJSON
+    authorJson: NullableAuthorJSON
   ) {
+    if (authorJson == null) {
```
wchargin: Interesting that this is the only place where a check is required. I guess that that's a good thing!

src/plugins/github/graphql.js
```diff
@@ -631,6 +631,11 @@ function mergeDirect<T>(destination: T, source: any): T {
   }
 }
 
+// If a user deletes their account, then the author is null
+// (the UI will show that the message was written by @ghost).
+// Therefore, NullableAuthorJSON is preferred to AuthorJSON
+// for most actual usage.
+export type NullableAuthorJSON = AuthorJSON | null;
```
wchargin: An alternate way to do this would be to export only `NullableAuthorJSON`
(possibly even renamed to `AuthorJSON`), and ask clients to write
`$NonMaybeType<NullableAuthorJSON>` when they want the non-null version.
I suspect that this would incur fewer accidental uses of `AuthorJSON`,
but I don’t have a strong preference.

PR #229, opened 2018-05-08
Create "REPOSITORY" nodes in GitHub plugin graph

decentralion:
This commit creates a new node type in the GitHub graph: the REPOSITORY
node. The REPOSITORY node has the following payload properties:
- url (string)
- name (string)
- owner (string)

Things this commit does:
- Add new node type and payload type (RepositoryNodePayload)
- Update parser to instantiate the new node type
- Update api.js to have Repository wrap the new node type (thus
Repository is a GitHub entity)
- Update snapshots
- Update users of GitHub node types to ensure they are exhaustive

Things that will come in a followon commit:
- Add CONTAINS edges from the repository to all its PRs and Issues
- Update the Repository porcelain to use those edges, rather than
scanning the graph for every possible Issue/PR (eventually those might
belong to other Repositories)
- Create a GitHubGraph abstraction in the porcelain, which makes it easy
to find all of the Repositories in a graph

Note that retrieving the repository owner technically involved fetching
the whole owner representation (as a GitHub user). I could have chosen
to add that user to the graph, with a "OWNS" edge pointing to the
repository. For simplicity's sake, I've declined to do that, and instead
just parse the owner's name directly.

Test plan:
Added tests to verify that the Repository porcelain entity has the right
properties. Combined with the snapshot tests, that should be sufficient.

This mostly completes #171

### DIFF
--- a/src/plugins/github/parser.js
+++ b/src/plugins/github/parser.js
@@ -9,6 +9,7 @@ import type {
   NodePayload,
   EdgePayload,
   PullRequestReviewNodePayload,
+  RepositoryNodePayload,
   AuthorNodePayload,
   AuthorsEdgePayload,
   PullRequestReviewCommentNodePayload,
@@ -273,6 +274,8 @@ class GithubParser {
       const anyNode: Node<any> = node;
       const type: NodeType = (node.address.type: any);
       switch (type) {
+        case "REPOSITORY":
+          break;
         case "ISSUE":
         case "PULL_REQUEST":
           const thisPayload: IssueNodePayload | PullRequestNodePayload =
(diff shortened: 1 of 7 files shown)


src/plugins/github/parser.js
```diff
@@ -162,12 +163,17 @@ class GithubParser {
...
+      | IssueNodePayload
+      | PullRequestNodePayload
+      | PullRequestReviewNodePayload
+      | RepositoryNodePayload
     >,
     childNode: Node<
       | CommentNodePayload
       | PullRequestReviewCommentNodePayload
       | PullRequestReviewNodePayload
+      | IssueNodePayload
```
wchargin: I hear that this should not be in this PR.

PR #230, opened 2018-05-08
Create `github.Porcelain`: whole-graph porcelain

decentralion:
Now that we have repository nodes (#171), it makes sense that the Github
porcelain should provide a way to wrap the entire graph, and provide
easy access for the various repositories. This adds a `Porcelain` class
to fulfill that need.

The `Porcelain` is very straightforward: it takes in the whole graph,
and gives a way to get all the Repositories, or to request a particular
Repository by owner/name. In the odd case wherein a graph contains
multiple repository nodes with the same owner and name, an error is
thrown. Per standard JS map semantics (bleh), it can return undefined if
there is no matching repository.

Test plan:
See that the unit tests now use the standard behavior, and a test
verifies behavior for non-existant repositories. I don't have a test
case where there are multiple repo nodes, but that itself would be an
error, so throwing an error in that case is just defensive programming.

### DIFF
--- a/src/plugins/github/api.test.js
+++ b/src/plugins/github/api.test.js
@@ -2,7 +2,7 @@
 
 import {parse} from "./parser";
 import exampleRepoData from "./demoData/example-github.json";
-import {Repository, Issue, PullRequest, Comment, Author} from "./api";
+import {Porcelain, Issue, PullRequest, Comment, Author} from "./api";
 import {
   AUTHOR_NODE_TYPE,
   COMMENT_NODE_TYPE,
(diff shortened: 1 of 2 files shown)


src/plugins/github/api.test.js
```diff
@@ -11,18 +11,30 @@ import {
...
     if (result == null) {
       throw new Error(`Expected Issue/PR ${n} to exist`);
     }
     return result;
   }
+  describe("has repository finding", () => {
+    it("which works for an existing repository", () => {
+      expect(
+        porcelain.repository("sourcecred", "example-github")
+      ).toBeDefined();
```
decentralion: This might look inadequate as a test case, but remember that the actual properties on this object are tested under the Wrappers header

wchargin: I have no problem with the breadth of the test case. Note, however, that `expect(null).toBeDefined()` will pass; you might have meant `expect(…).toEqual(expect.anything())`, which passes for everything except `null` and `undefined`. (Not that it matters in this case, but in general this keeps better in line with our philosophy of "try to treat `null` and `undefined` equivalently whenever possible.)

PR #232, opened 2018-05-08
Ensure that flow is used in all js files

decentralion:
This script ensures that `@flow` or `@no-flow` is present in
every js file. Every existing js file that would fail this check has
been given `@no-flow`, we should work to remove all of these in the
future.

Test plan:
I verified that `yarn travis` fails before fixing the other js files,
and passes afterwards.

### DIFF
--- a/config/env.js
+++ b/config/env.js
@@ -1,3 +1,4 @@
+// @no-flow
 const fs = require("fs");
 const path = require("path");
 const paths = require("./paths");
--- a/scripts/ensure-flow.sh
+++ b/scripts/ensure-flow.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -eu
+comm -13 -z \
+  <(git grep -z --name-only -e @flow -e @no-flow | sort -z) \
+  <(find src config scripts -name '*.js' -print0 | sort -z) \
+  | tr '\0' '\n' \
+  | diff -u /dev/null -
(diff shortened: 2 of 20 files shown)


config/env.js
```diff
@@ -1,3 +1,4 @@
+//@no-flow
```
wchargin: Nit: can you add these with `// @no-flow` (with a space), per our convention?

scripts/ensure-flow.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -eu
+comm -13 -z \
+  <(git grep -z --name-only -e @flow -e @no-flow | sort -z) \
+  <(find src config scripts -name '*.js' -print0 | sort -z) \
```
wchargin: I note that there are a number of places where we implicitly depend on all "non-`node_modules`"/"relevant" code being in `{src,config,scripts}`. No change requested, but (a) moving forward, we might want to prefer explicitly excluding `node_modules`, because (b) if we change this invariant then we will have to remember to fixup a bunch of places in the code.

scripts/ensure-flow.sh
```diff
@@ -0,0 +1,7 @@
...
+  | tr '\0' '\n' \
+  | diff -au /dev/null -
```
wchargin: We don't need the `-a` flag to `diff`, because of the `tr` on the previous line.

scripts/ensure-flow.sh
decentralion: A good point.

scripts/ensure-flow.sh
decentralion: Fixed.

PR #236, opened 2018-05-08
Refactor artifact app

decentralion:
Please review the commits individually; they all have their own test plans. I verified that `yarn travis` passes for each one.


src/app/App.js
```diff
@@ -1,10 +1,38 @@
 // @flow
 
 import React from "react";
+import {BrowserRouter as Router, Route, NavLink} from "react-router-dom";
```
wchargin: This is fine with me for now. I suspect that eventually, instead of this import, we will want to DI `<App router={BrowserRouter} />` so that we can use a `MemoryRouter` (or similar) for testing.

src/app/App.js
```diff
@@ -1,10 +1,38 @@
...
+          <Route path={ARTIFACT_EDITOR_ROUTE} component={ArtifactEditor} />
+        </div>
+      </Router>
+    );
   }
 }
+
+const Home = () => (
+  <div>
+    <h2>Welcome to SourceCred! Please enjoy your stay.</h2>
```
wchargin: `<h1>`, please.

PR #246, opened 2018-05-08
Factor out `github.porcelain.asEntity`

decentralion:
@wchargin suggested that the entity-wrapping logic in porcelain
reference handling should be factored out as its own method. This was a
great suggestion; it will be very useful for plugin consumers, and it
also results in better test coverage.

Test plan: The new unit tests are nice. For your own safety, do not
question or quibble with the magnificent foo plugin.

### DIFF
--- a/src/plugins/github/porcelain.test.js
+++ b/src/plugins/github/porcelain.test.js
@@ -1,8 +1,16 @@
 // @flow
 
+import type {Address} from "../../core/address";
 import {parse} from "./parser";
 import exampleRepoData from "./demoData/example-github.json";
-import {Porcelain, Issue, PullRequest, Comment, Author} from "./porcelain";
+import {
+  asEntity,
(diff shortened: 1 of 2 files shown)


src/plugins/github/porcelain.test.js
```diff
@@ -40,6 +72,7 @@ describe("GitHub porcelain", () => {
       expect(repo.url()).toBe("https://github.com/sourcecred/example-github");
       expect(repo.owner()).toBe("sourcecred");
       expect(repo.name()).toBe("example-github");
+      expect(deepEqual(repo, asEntity(graph, repo.address()))).toBe(true);
```
wchargin: Could you check whether `expect(x).toEqual(y)` works (in both the positive and negative cases)? If it does, that'd be better—we get a diff in case of failure, rather than just `false`.

PR #252, opened 2018-05-09
Implement basic PageRank analysis

wchargin:
Summary:
We don’t expect the results to be of good quality right now. Rather,
this gives us a starting point from which to iterate the algorithm.

The convergence criterion also needs to be adjusted. (In particular, it
should almost certainly not be a constant.)

Test Plan:
Run `yarn start`. Select a graph, like `sourcecred/example-github`. Open
the JS console and click “Run basic PageRank”. Watch the console.

wchargin-branch: basic-pagerank


src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+
+import stringify from "json-stable-stringify";
+import * as tf from "@tensorflow/tfjs-core";
+
+import {Graph} from "../../core/graph";
+
+/**
+ * Run eigenvector centrality, and print results to console.
+ */
+export default function basicPagerank(graph: Graph<any, any>): void {
```
decentralion: This function should return the stationary distribution (I'm fine for it to also print to console as a side effect)

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+  tf.tidy(() => {
+    const {nodes, addressToIndex, markovChain} = graphToMarkovChain(graph);
+    const stationaryDistribution = findStationaryDistribution(markovChain);
```
decentralion: is there a reason not to have `const stationaryDistribution = findstationaryDistribution(markovChain).dataSync()`?

src/app/credExplorer/App.js
```diff
@@ -62,6 +63,18 @@ export default class App extends React.Component<Props, State> {
           ) : (
             <p>Graph not loaded.</p>
           )}
+          <button
+            disabled={graph == null}
+            onClick={() => {
+              setTimeout(() => {
+                if (graph != null) {
+                  basicPagerank(graph);
```
decentralion: This should set the stationary distribution in the page state.

decentralion:
Note: as of this PR, `yarn test` now prints the following warning:

```
 PASS  src/app/credExplorer/App.test.js
  ● Console

    console.warn node_modules/@tensorflow/tfjs-core/dist/environment.js:248
      WebGL is not supported on this device
```

src/app/credExplorer/basicPagerank.js
wchargin: I disagree. In general, a function that computes and returns a value should not have effectful output. (Some people call this "command–query separation.") If it did, then callers who just want the value would be forced to put up with the noise. Choosing not to return the value avoids this problem by making there not be any such callers.

I intended for this module to be an exit point: you feed it data and get nothing out. I could later refactor it to export two functions—one query to find the stationary distribution from a graph, and one command to describe the output to the console. This might be helpful for your UI work—would you like me to do that in the next PR?

src/app/credExplorer/basicPagerank.js
wchargin: I don't see any substantive reason one way or the other, but I find the way that I wrote it more readable (and (side-benefit) it takes fewer lines of code).

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+function findStationaryDistribution(markovChain) {
+  const n = markovChain.shape[0];
+  if (markovChain.shape.length !== 2 || markovChain.shape[1] !== n) {
+    throw new Error(`Expected square matrix; got: ${markovChain.shape}`);
+  }
+  let r0 = tf.tidy(() => tf.ones([n, 1]).div(tf.scalar(n)));
+  function computeDelta(pi0, pi1) {
+    const delta = tf.tidy(() => tf.max(tf.abs(pi0.sub(pi1))));
+    const result = delta.dataSync()[0];
+    delta.dispose();
```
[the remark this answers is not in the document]
wchargin: That makes sense; I'm happy to use `tidy` here. (I had no particular reason for not doing so.)

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+    return result;
+  }
+  let iteration = 0;
+  while (true) {
+    iteration++;
+    const r1 = tf.matMul(markovChain, r0);
+    const delta = computeDelta(r0, r1);
+    console.log(`[${iteration}] delta = ${delta}`);
+    if (delta < 1e-7) {
+      break;
```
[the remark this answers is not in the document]
wchargin: Agreed; will do.

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+    if (iteration >= 255) {
+      console.log(`[${iteration}] FAILED to converge`);
+      return r0;
+    }
+  }
+  console.log(`[${iteration}] CONVERGED`);
+  return r0;
+}
+
+function analyze(nodes, addressToIndex, pi) {
```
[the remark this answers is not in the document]
wchargin: Okay. In this case, I can write `pi: $ReadOnlyArray<number>`. But note that neither `tfjs-core` nor `deeplearn.js` has flow-typed bindings, so I'm not sure how we would type the actual tensors.

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+  const addressKey = ({pluginName, type}) => stringify({pluginName, type});
+  const nodesByKey = {};
+  nodes.forEach((node) => {
+    const {address} = node;
```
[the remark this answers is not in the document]
wchargin: What I find weird, and a bit sad, is that there is no support for alias patterns: I should be able to write `forEach((node as {address}) => …`. I hit this reasonably commonly. It should be forward-compatible, but it looks like there's no TC39 proposal to add it. :-(

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,108 @@
...
+    if (nodesByKey[addressKey(address)] === undefined) {
+      nodesByKey[addressKey(address)] = [];
```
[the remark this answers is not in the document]
wchargin: Yeah. Any of a couple of things would be nice: a Python `defaultdict`, a Java `Map.computeIfAbsent`, or even just `asObject: <T>($ReadOnlyArray<[string, T]>) => {[string]: T}`. I'm sure that each of these exists (in particular, the last is `lodash.fromPairs`), but not being in the standard library is a real bummer.

src/app/credExplorer/App.js
wchargin: See my comment on https://github.com/sourcecred/sourcecred/pull/252#discussion_r187194747.

src/app/credExplorer/basicPagerank.js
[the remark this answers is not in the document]
decentralion: You could write a proposal :)
If you want to go down this route, I can connect you to a friend who used to on TC39 standards, he might have some useful pointers.

src/app/credExplorer/basicPagerank.js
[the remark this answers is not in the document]
decentralion: I want the module surface area to be well typed, so even if we can't get flow typings for the tfjs stuff, we should implement a cleanly typed wrapper for the export points.

src/app/credExplorer/basicPagerank.js
decentralion: Yep, I'd like the module to export the data for usage in the UI.
I view the print-to-console as a short-term hack UI, so IMO that should not live in basicPageRank - it should be implemented in the cred explorer app, as it is stand in for a real UI that would be generated in that file.

decentralion:
Re the yarn test warning: Not too worried about it for now.

Re the flow typing: see https://github.com/sourcecred/sourcecred/pull/252#discussion_r187207663

wchargin:
PTAL. One thing to note: I use `$Call<tf.tensor2d>` for the type of a Tensor2D, even though this actually resolves to `any`.

src/app/credExplorer/App.js
```diff
@@ -88,6 +104,34 @@ export default class App extends React.Component<Props, State> {
         console.error("Error while fetching:", e);
       });
   }
+
+  analyzePagerankResult(pagerankResult: PagerankResult) {
```
decentralion: I was about to nit about `PageRank` vs `Pagerank`... but realized we already decided to group on semantic capitalization rather than brand capitalization w/ `Github`

src/app/credExplorer/basicPagerank.js
```diff
@@ -0,0 +1,99 @@
...
+import {AddressMap} from "../../core/address";
+import {Graph} from "../../core/graph";
+
+export type PagerankResult = AddressMap<{|
+  +address: Address,
+  +probability: number,
+|}>;
+
+/**
+ * Run eigenvector centrality, and print results to console.
```
decentralion: update docstring

src/app/credExplorer/App.js
wchargin: I was wondering this as I wrote it. My guiding factor was that the identifier `pagerankResult` seems (much) more natural than `pageRankResult`, which suggests that the type name should be `PagerankResult`. Also, consistency, as you note.

src/app/credExplorer/basicPagerank.js
wchargin: Of course; thanks.

wchargin:
Note for posterity: an earlier version of this code was paired with @decentralion . We're not sure what the best way is to describe this kind of "partial paired-with" relationship in the graph.

PR #261, opened 2018-05-10
Add `nodeDescription` for GitHub nodes

decentralion:
`nodeDescription` gives a short, readable description of the content at
a given node.

Test plan: View the included snapshot test.

### DIFF
--- a/src/plugins/github/render.js
+++ b/src/plugins/github/render.js
@@ -1,0 +1,67 @@
+// @flow
+
+/*
+ * Methods for rendering and displaying GitHub nodes.
+ */
+import stringify from "json-stable-stringify";
+import {Graph} from "../../core/graph";
+import type {Address} from "../../core/address";
+import {
+  asEntity,
+  Issue,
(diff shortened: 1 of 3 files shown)


src/plugins/github/render.js
```diff
@@ -0,0 +1,68 @@
+// @flow
+
+/*
+ * Methods for rendering and dispalying GitHub nodes.
```
wchargin: sp.: *displaying*

src/plugins/github/render.js
```diff
@@ -0,0 +1,68 @@
...
+  Author,
+  Repository,
+} from "./porcelain";
+
+/* Give a short description for the GitHub node at given address.
+ * Useful for e.g. displaying a title.
+ */
+export function nodeDescription(graph: Graph<any, any>, addr: Address) {
+  const entity = asEntity(graph, addr);
+  let author: Author;
```
wchargin: Let's fold this into `const`s where it belongs.

PR #271, opened 2018-05-11
Create a `markovChain.js` module

wchargin:
Summary:
For now, this module has just two types: `Distribution` and
`SparseMarkovChain`. We’ll gradually pull code from `basicPagerank` into
this module, adding unit tests along the way.

Test Plan:
None required.

wchargin-branch: markov-chain-module

### DIFF
--- a/src/core/attribution/markovChain.js
+++ b/src/core/attribution/markovChain.js
@@ -1,0 +1,28 @@
+// @flow
+
+/**
+ * A distribution over the integers `0` through `n - 1`, where `n` is
+ * the length of the array. The value at index `i` is the probability of
+ * `i` in the distribution. The values should sum to 1.
+ */
+export type Distribution = Float64Array;
+
+/**
+ * A representation of a sparse transition matrix that is convenient for
+ * computations on Markov chains.
+ *
+ * A Markov chain has nodes indexed from `0` to `n - 1`, where `n` is
+ * the length of the chain. The elements of the chain represent the
+ * incoming edges to each node. Specifically, for each node `v`, the
+ * in-degree of `v` equals the length of both `chain[v].neighbor` and
+ * `chain[v].weight`. For each `i` from `0` to the degree of `v`
+ * (exclusive), there is an edge to `v` from `chain[v].neighbor[i]` with
+ * weight `chain[v].weight[i]`.
+ *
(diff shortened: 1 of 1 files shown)


decentralion:
btw: `src/core/algorithms` would be a fine place.
i also want to keep (at least module level separation) between code that deals with "pure" markov chains and pagerank, and code that has any knowledge of graphs and how to map them into markov chains.

decentralion:
actually, I think you mentioned not liking the term `algorithms` as a directory name because it is over-broad. Perhaps `attribution` or `calculation`? Either way, I think this code should live in core, not app.

wchargin:
I am happy to move this and downstream PRs. I also agree that `src/core/` is better than `src/app/`—though if attribution algorithms are to be pluggable, then perhaps later a `src/attribution/` would be a better home. (Or perhaps split `src/plugins/data/*` and `src/plugins/attribution/*`.)

I am fine with putting it in `src/core/attribution/markovChain.js` for now, and will do so unless you voice a different preference.

I agree with the principle of doing the move upfront, but I do want to note that (a) with verbatim moves, `git blame` follows the diff properly (see the `-f` option to `git-blame`), and (b) if you're talking about SourceCred handling moves, then this is certainly something that we want to handle well anyway. :-)

wchargin:
I have modified all commits in this stack for the move, as follows:
```shell
git checkout origin/wchargin-extract-findStationaryDistribution
filter='
mkdir -p src/core/attribution/;
for filename in \
    src/app/credExplorer/markovChain.js \
    src/app/credExplorer/markovChain.test.js \
; do
    if [ -f "${filename}" ]; then
        mv "${filename}" src/core/attribution/
    fi
done
sed -i "s:\\./markovChain:../../core/attribution/markovChain:g" \
    src/app/credExplorer/basicPagerank.js
node /home/wchargin/git/sourcecred/node_modules/prettier/bin-prettier.js \
    --config /home/wchargin/git/sourcecred/.prettierrc.json \
    --write src/app/credExplorer/basicPagerank.js
'
git filter-branch --tree-filter "${filter}" HEAD ^origin/master
git rebase origin/master --exec 'yarn travis'
git rebase origin/master --exec 'git push-to-target'
```
All (actual) Travis builds pass, so I will now proceed to merge the PRs.

PR #276, opened 2018-05-11
Make `findStationaryDistribution` configurable

wchargin:
Summary:
There are substantive options for `convergenceThreshold` and
`maxIterations`, as well as the output option `verbose`. This change is
made in preparation for extracting this function into `markovChain`,
where we will add unit tests for it.

Test Plan:
Behavior of `yarn start` is unchanged.

wchargin-branch: configurable-findstationarydistribution

### DIFF
--- a/src/app/credExplorer/basicPagerank.js
+++ b/src/app/credExplorer/basicPagerank.js
@@ -34,7 +34,7 @@ type OrderedSparseMarkovChain = {|
 
 export default function basicPagerank(graph: Graph<any, any>): PagerankResult {
   const {nodeOrder, chain} = graphToOrderedSparseMarkovChain(graph);
-  const pi = findStationaryDistribution(chain);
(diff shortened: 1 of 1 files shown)


src/app/credExplorer/basicPagerank.js
```diff
@@ -113,7 +113,20 @@ export function graphToOrderedSparseMarkovChain(
...
+    +verbose?: boolean,
+    +convergenceThreshold?: number,
+    +maxIterations?: number,
+  |}
+): Distribution {
+  const fullOptions = {
+    verbose: false,
+    convergenceThreshold: 1e-7,
+    maxIterations: 255,
+    ...(options || {}),
```
decentralion: Nice pattern. I keep being impressed with how much can be accomplished via spreads

wchargin: :+1:

PR #279, opened 2018-05-12
Print timing info in `execDependencyGraph`

### LINKED ISSUE #270: execDependencyGraph should print task times
E.g. when running `sourcecred graph foo bar` it would be great to see how long each step took. cc @wchargin

wchargin:
Summary:
Resolves #270.

These timings should be considered approximate. If one process finishes
while another process’s termination callback is still being handled,
then the second process may report a time that is too high. This might
happen if, for instance, calling `stdout.toString()` takes a really long
time.

Also, these timings should not be considered indicative of the
individual performance of any particular task. As tasks compete for CPU
and other resources, the elapsed time of one task depends greatly on the
other tasks running at the same time. For instance, using the set of
tasks given by
```js
const n = 10;
return Array(n)
  .fill(null)
  .map((_, i) => ({
    id: "check-pretty-" + String(i),
    cmd: ["npm", "run", "--silent", "check-pretty"],
    deps: [],
  }));
```
we see that all tasks are reported to take about the same amount of
time, but that time varies dramatically with `n`: for `n=1` it is
approximately 4s, while for `n=10` it is approximately 24s. This can be
partially mitigated by inspecting the output of `time(1)`, but not
entirely: even the “user” time varies by around a factor of 2 depending
on what other trasks are running.

In practice, this means that we have inaccurate reportings like the
following:

  - `backend` reportedly takes around 10.5s, but when run individually
    takes only around 4.6s;
  - `ci-test` is reported at 9.4s but actually takes only 4.4s;
  - `lint` is reported at 8.8s but actually only takes 3.5s.

We _could_ try to mitigate this by including with each task a “hint” of
how many cores we expect it to peg (`ci-test` gets “all”, `backend` gets
“just one”) and trying not to spawn more processes than there are cores.
But this sounds pretty fragile, so I haven’t made any efforts to do so.

Test Plan:
Replace the Travis config with the following, then run `yarn travis`:
```diff
diff --git a/config/travis.js b/config/travis.js
index be308d4..ba00a04 100644
--- a/config/travis.js
+++ b/config/travis.js
@@ -16,6 +16,14 @@ function main() {
 }

 function makeTasks(mode /*: "BASIC" | "FULL" */) {
+  return [
+    {id: "true", cmd: ["true"], deps: []},
+    {id: "false", cmd: ["false"], deps: []},
+    {id: "sleep-a", cmd: ["sleep", "2"], deps: []},
+    {id: "sleep-b", cmd: ["sleep", "2"], deps: []},
+    {id: "sleep-c", cmd: ["sleep", "2"], deps: ["sleep-b"]},
+    {id: "never", cmd: ["echo", "wat"], deps: ["false"]},
+  ];
   const basicTasks = [
     {
       id: "ensure-flow-typing",
```

You can see that both passing and failing tests have times printed, but
tasks that never ran do not. You can also see that time is measured from
task start, not from dependency graph start: the times for all three
occurrences of `sleep` are approximately 2s, even though one depends on
another.

wchargin-branch: execDependencyGraph-timing

### DIFF
--- a/src/tools/execDependencyGraph.js
+++ b/src/tools/execDependencyGraph.js
@@ -19,6 +19,10 @@ type Task = {|
 type TaskResult = {|
   +id: TaskId,
   +success: boolean,
+  +timeElapsed: {|
+    +secondsPart: number,
+    +nanosecondsPart: number,
+  |},
   +status: number,
   +stdout: string,
   +stderr: string,
@@ -116,7 +120,17 @@ exports.default = async function execDependencyGraph(
     const badge = success
       ? chalk.bgGreen.bold.white(spacedLabel(fullOptions.taskPassLabel))
       : chalk.bgRed.bold.white(spacedLabel(fullOptions.taskFailLabel));
-    console.log(`${badge} ${id}`);
+    const seconds = (() => {
+      if (result == null) {
+        return "";
+      }
+      const secondsNumber = (
+        result.timeElapsed.secondsPart +
+        result.timeElapsed.nanosecondsPart * 1e-9
+      ).toFixed(3);
+      return " " + chalk.blue(`(${secondsNumber}s)`);
+    })();
+    console.log(`${badge} ${id}${seconds}`);
 
     if (mode === "OVERVIEW" && success) {
       return;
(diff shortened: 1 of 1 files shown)


wchargin:
I think that I would prefer that this pull request not be merged. Let me
know what you think.

decentralion:
Discussed offline: We don't think this should merge, as prominently displaying info that is quite unreliable will just be a recipe for confusion. We might want to add a sequential mode that prints accurate timing info (at the cost of being slower overall)

PR #281, opened 2018-05-13
Add porcelain for the Git plugin

decentralion:
This commit adds "Node Porcelain" for the Git plugin. Node porcelain is
a wrapper over a graph and node address, which makes it possible to
access payload data and adjacent nodes via a familiar object-oriented
API.

I believe this porcelain provides substantially better legibility and
usability for the Git plugin. Consider that it is now easy to see what
relationships each Git node type can have by reading the method
signatures in the porcelain, rather than needing to inspect all of the
Edge types in types.js.

This porcelain has slightly different conventions from the porcelain in
the GitHub plugin, although the APIs are very similar. I intend to
follow this commit with two more: one that switches clients of the Git
plugin to use the porcelain, and another that refactors the GitHub and
Git porcelains to use a base Porcelain implementation in src/core.

Test plan:
Examine the public API of the Git porcelain (this is unlikely to change
much), and its corresponding test code.

`yarn travis --full` passes.

### DIFF
--- a/src/plugins/git/porcelain.js
+++ b/src/plugins/git/porcelain.js
@@ -1,0 +1,243 @@
+// @flow
+
+/*
+ * This module contains "Porcelain" for working with Git graphs. By
+ * "Porcelain", we mean it is a much more convenient and polished API. It
+ * allows accessing Git graph data via a familiar object-oriented API,
+ * rather than needing to use the specific graph-based methods in the
+ * underlying graph.
+ *
+ * In general, the porcelain module provides wrapper objects that contain the
+ * entire Git graph, and a pointer to a particular entity in that graph.
+ * Creating the wrappers is extremely cheap; all actual computation (e.g.
+ * finding the body or author of a post) is done lazily when that information
+ * is requested.
+ *
+ * The porcelain system is under active development.
+ * I expect that we will soon refactor the base porcelain abstraction out of
+ * this module into core/, and the implementation may shift to be more conserved
+ * with the GitHub porcelain. The APIs should remain unchanged.
+ * */
+import stringify from "json-stable-stringify";
+
+import {Graph} from "../../core/graph";
+import type {Node} from "../../core/graph";
+import type {Address} from "../../core/address";
+
+import type {
+  TreeEntryNodePayload,
+  SubmoduleCommitPayload,
+  BlobNodePayload,
+  TreeNodePayload,
+  CommitNodePayload,
+  NodePayload,
+  NodeType,
+  Hash,
+} from "./types";
+import {GIT_PLUGIN_NAME} from "./types";
+import {commitAddress} from "./address";
+
+export class PorcelainGraph {
+  graph: Graph<any, any>;
+  constructor(graph: Graph<any, any>) {
+    this.graph = graph;
+  }
+
+  // Note that this method is presently unsafe, as the hash may not exist.
+  // In the future, we will come up with a general case solution to have
+  // the type system verify that returned porcelains must be existence-tested
+  // before their properties are usable.
+  commitByHash(h: Hash): Commit {
+    const addr = commitAddress(h);
+    return new Commit(this.graph, addr);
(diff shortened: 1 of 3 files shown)


src/plugins/git/porcelain.js
```diff
@@ -0,0 +1,263 @@
...
+
+  neighbors(options?: {|
+    +nodeType?: string,
+    +edgeType?: string,
+    +direction?: "IN" | "OUT",
+  |}): BaseNodePorcelain[] {
+    return this.graph
+      .neighborhood(this.nodeAddress, options)
+      .map(({neighbor}) => new BaseNodePorcelain(this.graph, neighbor))
+      .map(asGitPorcelain);
```
wchargin: This shouldn't be here—it will crash when a Git node has incident edges to/from other plugins' nodes (e.g., a pull-request-to-merged-commit edge).

src/plugins/git/porcelain.test.js
```diff
@@ -0,0 +1,116 @@
...
+  const commit = graph.commitByHash(commitHash);
+  if (commit.hash() !== commitHash) {
+    throw new Error(
+      `Expected commit hash ${commitHash}, got hash ${commit.hash()}`
+    );
+  }
+  return commit;
+};
+describe("Git porcelain", () => {
+  it("commits have hashes", () => {
```
wchargin: Why doesn't this function just call `getCommit()`?

src/plugins/git/porcelain.js
```diff
@@ -0,0 +1,264 @@
...
+export class SubmoduleCommit extends BaseGitPorcelain {
+  static from(n: BaseGitPorcelain): SubmoduleCommit {
+    if (n.type() !== "SUBMODULE_COMMIT") {
+      throw new Error(`Unable to cast ${n.type()} to SubmoduleCommit`);
+    }
+    return new SubmoduleCommit(n.graph, n.nodeAddress);
+  }
+
+  url(): string {
+    const node: Node<SubmoduleCommitPayload> = (this.node(): any);
```
wchargin: All these `any`s are disturbing. Let's parameterize `BaseNodePorcelain` over a node payload, as in the GitHub plugin.

src/plugins/git/porcelain.js
```diff
@@ -0,0 +1,264 @@
...
+      })
+      .map(({neighbor}) => neighbor);
+    if (neighbors.length !== 1) {
+      throw new Error(
+        `Wrong number of contents for ${stringify(this.address())}`
+      );
+    }
+    const n = neighbors[0];
+    const type = n.type;
+    switch (type) {
```
wchargin: IMO, this is easier to read as something like "if the type is BLOB, TREE, or SUBMODULE_COMMIT, then return `asGitPorcelain(n)`, else error." Or, "let `result` be `asGitPorcelain(n)`, and assert that the type of `result` is one of the three valid types."

src/plugins/git/porcelain.js
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  contents(): Tree | Blob | SubmoduleCommit {
+    const neighbors = this.graph
+      .neighborhood(this.nodeAddress, {
+        edgeType: "HAS_CONTENTS",
+        direction: "OUT",
+      })
+      .map(({neighbor}) => neighbor);
+    if (neighbors.length !== 1) {
```
wchargin: I believe that this invariant does not hold for submodule commit nodes. See:
https://github.com/sourcecred/sourcecred/blob/115d7f3921c5145180b472f1678edaf804d0973f/src/plugins/git/createGraph.js#L153-L198

We might want to add such a case to our example-git repository. (Later.)

src/plugins/git/porcelain.js
wchargin: Not a problem now that `asGitPorcelain` has been removed (there's no confusing repetition).

PR #282, opened 2018-05-13
Absolute import

### LINKED ISSUE #256: Use absolute, not relative paths
Our use of relative paths throughout the codebase is quite cumbersome. Let's fix this. 

On the webpack side, it looks like the solution [is simple].

[is simple]: https://spin.atomicobject.com/2017/10/07/absolute-paths-javascript/

For flow, it seems the solution [is possible] albeit not phenomenally well supported. Here's [a bit more info].

[is possible]: https://github.com/facebook/flow/issues/4186#issuecomment-336084041
[a bit more info]: https://github.com/facebook/flow/issues/1826#issuecomment-220792807

decentralion:
Node imports are now allowed from absolute paths, starting with `src/`.

For example, imports from "../../core/graph" can now be replaced with
"core/graph".

Alternatives considered: Consider that we want to import the file
`src/core/graph`. I considered the following four approaches for
this import:

1. `import {Stuff} from "core/graph"`
2. `import {Stuff} from "Core/graph"`
3. `import {Stuff} from "@core/graph"`
4. `import {Stuff} from "@/core/graph"`

I chose option 1, because it is both concise and easy to implement.

Because we don't use a separate namespace for our src/ directories and
node modules, there is a possibility of name collisions.  (If so,
node_modules will win.) However, I believe this issue is unlikely to
happen in practice, as we have only a few directories in `src`, we would
need a conflict on the full (pathed) name.

In the event that such a conflict does arise, and is not easily resolved
by disambiguating the directory name, it should not be disruptive to
switch to import aliasing at that time.

This pull request modifies many existing imports with convoluted relative
paths to use absolute paths instead. The changes were generated via
the following two shell script commands.

```
find src -name '*.js' | xargs sed -ri 's%(from ")(../)+(plugins/)?(git|github|artifact)/%\1plugins/\4/%g'
find src -name '*.js' | xargs sed -ri 's%(from ")(../)+(core|graphql|tools|app)?/%\1\3/%g'
```

Note that the changes are segregated by commit, so that all config
changes are in one commit, all path changes in another.

Test plan: 
- `yarn travis --full` still passes
- `yarn start` still provides a working development server
- `yarn build && yarn backend && node bin/sourcecred.js start` 
still provides a working production instance

Fixes #256

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -8,5 +8,7 @@ flow-typed
 [lints]
 
 [options]
+module.system.node.resolve_dirname=node_modules
+module.system.node.resolve_dirname=src
 
 [strict]
--- a/config/webpack.config.backend.js
+++ b/config/webpack.config.backend.js
@@ -2,6 +2,7 @@
 const webpack = require("webpack");
 const eslintFormatter = require("react-dev-utils/eslintFormatter");
 const ModuleScopePlugin = require("react-dev-utils/ModuleScopePlugin");
+const path = require("path");
 const paths = require("./paths");
 const nodeExternals = require("webpack-node-externals");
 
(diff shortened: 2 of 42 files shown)


decentralion:
Offline, you said you preferred to see `"@/core"` style imports. I've implemented that over at #283.

PR #283, opened 2018-05-14
Enable unambiguous absolute path imports for node

### LINKED ISSUE #256: Use absolute, not relative paths

decentralion:
This commit configures webpack, jest, and flow so that we can use the @
symbol in module imports as a reference to the project's `src/`
directory.

For example, instead of:
`import {Stuff} from "../../../core/graph"`
You can now write:
`import {Stuff} from "@/core/graph"`

The files `.flowconfig`, `config/webpack.config/{backend,dev,prod}.js`
and `package.json` were manually changed to enable module name aliasing.
All other changes in the commit were generated automatically by running
the following two commands:

```
find src -name '*.js' | xargs sed -ri 's%(from ")(../)+(plugins/)?(git|github|artifact)/%\1@/plugins/\4/%g'
find src -name '*.js' | xargs sed -ri 's%(from ")(../)+(core|graphql|tools|app)?/%\1@/\3/%g'
```

This fixes #256.

To see an alternative that is slightly more concise (no `"@/"` at the
beginning of imports) but has the disadvantage of creating potential
name collisions with node modules, see #282.

Test plan:
* `yarn travis --full` passes.
* `yarn start` produces a working dev server.
* `yarn build && yarn backend && node bin/sourcecred.js start` produces
a working web application.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -8,5 +8,6 @@ flow-typed
 [lints]
 
 [options]
+module.name_mapper='^@/' -> '<PROJECT_ROOT>/src/'
 
 [strict]
--- a/config/webpack.config.backend.js
+++ b/config/webpack.config.backend.js
@@ -2,6 +2,7 @@
 const webpack = require("webpack");
 const eslintFormatter = require("react-dev-utils/eslintFormatter");
 const ModuleScopePlugin = require("react-dev-utils/ModuleScopePlugin");
+const path = require("path");
 const paths = require("./paths");
 const nodeExternals = require("webpack-node-externals");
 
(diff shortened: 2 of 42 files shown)


decentralion:
Closing as stale.

PR #289, opened 2018-05-16
Remove Node/Edge polymorphism on Graph

decentralion:
As previously implemented, the Graph was polymorphic in its NodePayload
and EdgePayload. This was a lot of bookkeeping for very
little apparent benefit. In general, graphs may be constructed with
foreign plugins, so it's hard to do anything with this information.

In my experience, having the Graph polymorphism has never caught a bug,
and has led to lots of boilerplate code, especially typing `Graph<any,
any>`.  I view the fact that in #286 we added a new core `NodeReference`
concept, which always types its Graph as `<any, any>`, as strongly
suggestive that this was not going to provide any lasting value.

In this commit, I've removed the Graph polymorphism. Note how in many
cases where we were typing the graph, it provided no value, as evidenced
by the fact that the imported Node and Edge types were used no-where
else in the file other than in the Graph declaration.

Test plan:
Removing extra typing information is very unlikely to cause regressions.
`yarn flow` and `yarn lint` both pass.

### DIFF
--- a/src/core/graph.test.js
+++ b/src/core/graph.test.js
@@ -388,7 +388,7 @@ describe("graph", () => {
     describe("neighborhood detection", () => {
       describe("type filtering", () => {
         class ExampleGraph {
-          graph: Graph<{}, {}>;
+          graph: Graph;
           root: Address;
           idIncrement: number;
           inEdges: {[string]: Edge<{}>};
@@ -891,9 +891,7 @@ describe("graph", () => {
        * node `u`, create a graph with just that node, its neighbors,
        * and its incident edges (in both directions).
        */
-      function neighborhoodDecomposition<NP, EP>(
-        originalGraph: Graph<NP, EP>
-      ): Graph<NP, EP>[] {
+      function neighborhoodDecomposition(originalGraph: Graph): Graph[] {
         return originalGraph.nodes().map((node) => {
           const miniGraph = new Graph();
           miniGraph.addNode(node);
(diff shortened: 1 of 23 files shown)


wchargin:
I’m not sure whether I agree with this change, but I definitely disagree
with the assertion that “having the Graph polymorphism has never caught
a bug”. When we had `type: string` on the payload objects, having a
parameterized `Graph<GithubNodePayload, GithubEdgePayload>` was valuable
indeed: calling `addNode({/* some object literal */})` would throw an
error in the case that we forgot a field or specified the wrong node
type. This certainly did catch bugs for me.

Let’s postpone this until after we address the address–payload
distinction, because that change might make fields like `type: string`
more tractable again, meaning that the polymorphism would provide more
safety benefits.

wchargin:
I’m now okay with merging this given that we’re working on implementing
the plan set out in #296, which will mean that we have type information
at the plugin level (where it is useful), and not at the graph level
(where it is less useful).

src/core/graph.test.js
```diff
@@ -1001,15 +997,15 @@ describe("graph", () => {
...
           .addNode(data.a())
           .addNode(data.b())
           .addEdge(data.u());
-        const g2: Graph<boolean, null> = new Graph()
+        const g2: Graph = new Graph()
           .addNode(data.c())
           .addNode(data.d())
           .addEdge(data.v());
-        type ResultGraph = Graph<string | boolean, number | null>;
+        type ResultGraph = Graph;
```
wchargin: If you were feeling particularly motivated, you could inline this one. :-)

Issue #280, opened 2018-05-13
Implement versioning for all serializables


decentralion:
Currently, SourceCred has no consistent way to approach versioning. This hasn't been a problem thus far as we have no users, and it's reasonable to regenerate all graphs from scratch every session. Once we start persisting graphs (/ have users), we will need versioning and approach for backcompat on old data structures.

We should implement a consistent versioning system. Ideally it will have at least the following two properties:
- Allow fine-grained versioning of composite objects, e.g. at present a Graph serializes using Address Maps, and the address maps should be independently versionable (h/t @wchargin )
- Allow plugins to have separate data versioning, e.g. the GitHub plugin can change its graph representation without this leaking over to other plugins

Beanow:
Nowadays we use `Compatible<T>` for this.

https://github.com/sourcecred/sourcecred/blob/f260569605f2c97f52cafd226b172070de3fb351/src/util/compat.js#L3

PR #292, opened 2018-05-17
Add data versioning utilities

decentralion:
This commit adds the `src/util/compat` module, which is responsible for
data compatibility. It provides two methods: `toCompat` and
`fromCompat`.

`toCompat` allows an object to be versioned with a type and version
string. Doing so wraps the object in an array whose first element has
`version` and `type` fields.

`fromCompat` allows loading compatibilized data. It takes a `type` and
`version` expectation, as well as optional processors for various data
versions. It throws an error if the data was not compatibilized, or if
it is the wrong type, or if it has an unsupported version.

The versioning is designed to be composable; see the final test block
for an example of loading a structure with nested versioning.

Test plan: Carefully inspect the included unit tests, and feel free to
propose more.

For context, see #280.

### DIFF
--- a/src/util/compat.js
+++ b/src/util/compat.js
@@ -1,0 +1,45 @@
+// @flow
+
+export opaque type Compatible<T> = [CompatInfo, T];
+type CompatInfo = {|
+  +type: string,
+  +version: string,
+|};
+
+export function toCompat<T>(compatInfo: CompatInfo, obj: T): Compatible<T> {
+  return [compatInfo, obj];
+}
+
+/**
+ * Load an object from compatibilized state created by `toCompat`.
+ * The object has an expected type and version, and may optionally have
+ * handler functions for transforming previous versions into a canonical state.
+ * If a handler is present for the current version, it will be applied.
+ * Throws an error if the compatibilized object is the wrong type, or if its version
+ * is not current and there was no handler for its version.
+ */
+export function fromCompat<T>(
+  expectedCompatInfo: CompatInfo,
+  obj: Compatible<any>,
+  handlers: ?{[version: string]: (any) => T}
+): T {
+  if (!Array.isArray(obj) || obj.length !== 2) {
+    throw new Error(
+      "Tried to load object that didn't have compatibility defined"
+    );
+  }
+  const {type, version} = obj[0];
+  let result = obj[1];
+
+  const {type: expectedType, version: expectedVersion} = expectedCompatInfo;
+  if (type !== expectedType) {
+    throw new Error(`Expected type to be ${expectedType} but got ${type}`);
+  }
+
+  if (handlers != null && handlers[version] != null) {
+    result = handlers[version](result);
+  } else if (version !== expectedVersion) {
+    throw new Error(`${type}: tried to load unsupported version ${version}`);
+  }
+  return result;
+}
--- a/src/util/compat.test.js
+++ b/src/util/compat.test.js
@@ -1,0 +1,168 @@
+// @flow
+
+import {toCompat, fromCompat} from "./compat";
+import type {Compatible} from "./compat";
+
+describe("compat utilities", () => {
+  type FooV1 = {foo: number};
+  type FooV2 = {bar: {foo: number}};
+  const type = "Foo Plugin's Experiment";
+  const v1 = "v1";
+  const v2 = "v2";
+  function fooV1ToV2(x: FooV1): FooV2 {
+    return {bar: x};
+  }
+  const example = () => {
+    const dataV1: FooV1 = {foo: 1};
+    const dataV2: FooV2 = {bar: {foo: 1}};
+    const compatV1 = toCompat({type, version: v1}, dataV1);
+    const compatV2 = toCompat({type, version: v2}, dataV2);
+    return {dataV1, dataV2, compatV1, compatV2};
+  };
+
+  it("toCompat doesn't fails on primitives", () => {
+    expect(
+      fromCompat({type, version: v1}, toCompat({type, version: v1}, 3))
+    ).toBe(3);
+  });
+
+  it("toCompat -> fromCompat is identity", () => {
+    const {dataV1, compatV1} = example();
+    expect(fromCompat({type, version: v1}, compatV1)).toEqual(dataV1);
+  });
+
+  it("fromCompat fails if compatibility undefined", () => {
+    const dataV1: any = example().dataV1;
+    expect(() => fromCompat({type, version: v1}, dataV1)).toThrowError(
+      "didn't have compatibility defined"
+    );
+  });
+
+  it("fromCompat fails if type is wrong", () => {
+    const {compatV1} = example();
+    expect(() =>
+      fromCompat({type: "who is Foo?", version: v1}, compatV1)
+    ).toThrowError("Expected type");
+  });
+
+  it("fromCompat fails if version is wrong", () => {
+    const {compatV1} = example();
+    expect(() => fromCompat({type, version: v2}, compatV1)).toThrowError(
+      "unsupported version"
+    );
+  });
+
+  it("handlers can load older versions", () => {
+    const {compatV1, dataV2} = example();
+    const handlers = {[v1]: fooV1ToV2};
+    expect(fromCompat({type: type, version: v2}, compatV1, handlers)).toEqual(
+      dataV2
+    );
+  });
+
+  it("handlers activate even on current version", () => {
+    const {compatV1} = example();
+    const handlers = {
+      [v1]: () => ({
+        hello: "world",
+      }),
+    };
+    expect(fromCompat({type, version: v1}, compatV1, handlers)).toEqual({
+      hello: "world",
+    });
+  });
+
+  describe("composable versioning", () => {
+    class InnerV1 {
+      x: number;
+      constructor(x: number) {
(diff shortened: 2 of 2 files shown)


decentralion:
See #293 for an example.

I wrote this code with the expectation that `__compat__` info only exists on serialized data, e.g. the result of calling `toJSON()`. It's removed when the data is loaded with `fromCompat`, and attempting to attach compat info to something that already has it is an error.

This works well for classes like `Graph` and `Address` that define matching `toJSON` and `fromJSON` methods, as these methods make it easy to have canonical entry/exit points for compat info. For cases like the `repositoryJSON`, where we may directly serialize the data to disk without going through explicit `{to,from}JSON` methods, it's harder to use `compat`, as it's not clear where the entry/exit points should live. I think a reasonable resolution is to have `toJSON/fromJSON` methods for all the data that we are explicitly supporting compatibility on, and to treat any data dumped without compatibility as ephemeral. Thus, if we do make the `repositoryJSON` part of the officially supported data layer, we should define a class to hold it with to/from methods.

src/util/compat.js
```diff
@@ -0,0 +1,66 @@
...
+): T & Compatible {
+  if (typeof obj !== "object") {
+    throw new Error(
+      `Tried to compatibilize ${typeof obj}, but only objects are supported`
+    );
+  }
+  if (obj == null) {
+    throw new Error(`Tried to compatibilize null value`);
+  }
+  if (obj.__compat__ != null) {
```
wchargin: Would prefer `"__compat__" in obj` here: an explicitly set `null` or `undefined` anywhere on the proto should be sufficient to fail.

src/util/compat.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+
+export type Compatible = {
```
wchargin: This seems like a perfect occasion for `export opaque type`.

decentralion:
Thanks for the suggestion. I find your argument and proposed solution persuasive, and have updated the PR accordingly. 

This change has two additional consequences, both of which I think are good:
- As previously implemented, the `__compat__` field might show up anywhere in the JSON blob, meaning you need to parse the whole object to find the compat. Now, for any compatible object looking at the first ~100 bytes of the json is sufficient.

- As previously implemented, all compatibilized objects were compatible with compatibility-ignorant consumers, i.e. they could just ignore the `__compat__` field and proceed as usual. This is "convenient", in the type-coercion mold of "conveniently allows you to do something potentially dangerous without realizing it". Now, if you don't know that objects have been compatibilized, you probably won't know how to read them, which is a good inducement to use `fromCompat` when interpreting the object.

decentralion:
Something I think we should still discuss (perhaps offline) before merging: I'm not sure the `fromCompat` API is quite right. Specifically, it doesn't give any mechanism to parse version strings, just to exhaustively enumerate them. Suppose, for example, that we got in the habit of making backwards-compatible changes to serialization and recording this as proper [semver] by incrementing the minor or patch values in the version string. Then, we would need to repetitively enumerate every historical version, even when no new handler is required.

However, the API as written is very convenient for our current use case, so we could punt on improving the API and refactor it when it becomes inconvenient. (Actually, for my current intentions we don't need any handlers at all; I just want to cleanly break on old data versions. But I wanted to demonstrate that we had a composable versioning strategy, so that necessitated putting some backcompat support in the API :slightly_smiling_face: )

[semver]: https://semver.org/

src/util/compat.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+
+export opaque type Compatible<T> = [Compat, T];
+type Compat = {|
```
wchargin: Could we rename this to `CompatibilityInfo`/`CompatInfo`, so that we don't have `Compat` and `Compatible` meaning different things, with `toCompat` actually returning a `Compatible`, etc.?

src/util/compat.js
```diff
@@ -0,0 +1,49 @@
...
+  +type: string,
+  +version: string,
+|};
+
+export function toCompat<T>(
+  type: string,
```
wchargin: A priori, I have a weak preference for taking the actual `compat: Compat` here instead of two unnamed arguments, but if you have a reason for doing it this way then I'm probably fine with it.

src/util/compat.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Load an object from compatibilized state created by `toCompat`.
+ * The object has an expected type and version, and may optionally have
+ * handler functions for transforming previous versions into a canonical state.
+ * If a handler is present for the current version, it will be applied.
+ * Throws an error if the compatibilized object is the wrong type, or if its version
+ * is not current and there was no handler for its version.
+ */
+export function fromCompat<T>(
+  expectedType: string,
```
wchargin: Likewise here could take `expected: Compat`.

src/util/compat.js
```diff
@@ -0,0 +1,49 @@
...
+  expectedVersion: string,
+  obj: Compatible<T>,
```
wchargin: This really is not a `Compatible<T>`; it is a `Compatible<mixed>`.

If we wanted to be more precise, we could write something using `$ObjMap` on `handlers` below, but using `any`/`mixed` here I think is acceptable, because appropriate types would just be unified away, anyway.

src/util/compat.test.js
```diff
@@ -0,0 +1,169 @@
...
+import {toCompat, fromCompat} from "./compat";
+import type {Compatible} from "./compat";
+
+describe("compat utilities", () => {
+  type FooV1 = {foo: number};
+  type FooV2 = {bar: {foo: number}};
+  const type = "Foo Plugin's Experiment";
+  const v1 = "v1";
+  const v2 = "v2";
+  function fooHandler(x: FooV1): FooV2 {
```
wchargin: Can we call this `convertV1ToV2` or something similar?

src/util/compat.test.js
```diff
@@ -0,0 +1,169 @@
...
+    const dataV1: FooV1 = {foo: 1};
+    const dataV2: FooV2 = {bar: {foo: 1}};
+    const compatV1 = toCompat(type, v1, dataV1);
+    const compatV2 = toCompat(type, v2, dataV2);
+    return {dataV1, dataV2, compatV1, compatV2};
+  };
+  it("toCompat adds a compat field", () => {
+    const {dataV1, compatV1} = example();
+    expect(dataV1).toEqual({foo: 1}); // calling compat caused no mutation
+    const {type: actualType, version: actualVersion} = (compatV1: any)[0];
```
wchargin: Doesn't look like this test provides much value—the others are sufficient—and it deliberately breaks the encapsulation, so let's get rid of it.

src/util/compat.test.js
```diff
@@ -0,0 +1,169 @@
...
+  it("fromCompat fails if version is wrong", () => {
+    const {compatV1} = example();
+    expect(() => fromCompat(type, v2, compatV1)).toThrowError(
+      "unsupported version"
+    );
+  });
+
+  it("handlers can load older versions", () => {
+    const {compatV1, dataV2} = example();
+    const handlers = {v1: fooHandler};
```
wchargin: You probably meant `[v1]: fooHandler` and not `v1: fooHandler`. (The former depends on the fact that the value of `v1` is literally `"v1"`.) Likewise lots of places below.

src/util/compat.test.js
```diff
@@ -0,0 +1,169 @@
...
+
+    class InnerV2 {
+      x: number;
+      constructor(x: number) {
+        this.x = x;
+      }
+      toJSON(): Compatible<FooV2> {
+        return toCompat("inner", v2, {bar: {foo: this.x}});
+      }
+
```
wchargin: nit: blank line inconsistency between `InnerV1` and `InnerV2`

src/util/compat.test.js
```diff
@@ -0,0 +1,169 @@
...
+        });
+      }
+    }
+
+    const canonical = () => new OuterV2(new InnerV2(1));
+    it("loads OuterV1<InnerV1>", () => {
+      const json = new OuterV1(new InnerV1(1)).toJSON();
+      expect(OuterV2.fromJSON(json)).toEqual(canonical());
+    });
+
```
wchargin: nit: wherefore this blank line?

src/util/compat.js
decentralion: Done

src/util/compat.js
decentralion: Done

src/util/compat.js
decentralion: Done

src/util/compat.test.js
decentralion: Done

src/util/compat.js
decentralion: Changed to any

src/util/compat.test.js
decentralion: Fixed.

src/util/compat.test.js
decentralion: Done

src/util/compat.test.js
decentralion: Fixed

src/util/compat.test.js
decentralion: Fixed

PR #293, opened 2018-05-17
Use versioning for `Graph` and `AddressMap`

decentralion:
This commit adds explicit versioning to the `Graph` and `AddressMap`
JSON representations, using the new `compat` module. This will make it
safer to change the serialization format for these classes.

(Note: I don't expect we'll add backcompat handlers for these classes
soon, but having versioning means that we can change the serialization
format in a way that breaks old data cleanly and explicitly, rather than
introducing undefined behavior.)

Test plan: The changes are slight, and well-captured by the snapshot
tests. Note that after this commit, the SourceCred commands will fail on
old data, so old data will need to be regenerated.

### DIFF
--- a/src/core/address.test.js
+++ b/src/core/address.test.js
@@ -3,8 +3,15 @@
 import sortBy from "lodash.sortby";
 import stringify from "json-stable-stringify";
 
+import {fromCompat} from "../util/compat";
 import type {Address} from "./address";
-import {AddressMap, fromString, toString} from "./address";
+import {
+  AddressMap,
+  fromString,
+  toString,
+  COMPAT_TYPE,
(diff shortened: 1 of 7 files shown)


src/core/address.test.js
```diff
@@ -70,7 +70,8 @@ describe("address", () => {
     });
 
     it("stringifies elements sans addresses", () => {
-      const json = makeMap().toJSON();
+      const compatJson = makeMap().toJSON();
+      const json = (compatJson: any)[1]; // Elide the compat field
```
wchargin: Can we call `fromCompat` here instead of breaking the encapsulation?

PR #295, opened 2018-05-21
Use indexed edges in graph internals

wchargin:
Summary:
This is an implementation-only, API-preserving change to the `Graph`
class. Edges’ `src` and `dst` attributes are now internally represented
as integer indices into a fixed ordering of nodes, which may depend on
non-logical properties such as insertion order. The graph’s serialized
form also now stores edges with integer `src`/`dst` keys, but the node
ordering is canonicalized so that two graphs are logically equal if and
only if their serialized forms are equal. This change substantially
reduces the rest storage space for graphs: the `sourcecred/sourcecred`
graph drops from 39MB to 30MB.

Currently, the graph will have to translate between integer indices and
full addresses at each client operation. This is not actually a big
performance regression, because it is just one more integer-index
dereference over the previous behavior, but it does indicate that the
optimization is not living up to its full potential. In subsequent
changes, the `NodeReference` class will be outfitted with facilities to
take advantage of the internal indexing; a long-term goal is that
roughly all operations should be able to be performed within the indexed
layer, and that translating between integers and addresses should only
happen at non-hot-path API boundaries.

This diff is considerably smaller and easier to read with `-w`.

Paired with @decentralion.

Test Plan:
I inspected the snapshots for general form, and manually verified that
the indices for one edge were correct (the MERGED_AS edge for the head
commit of the example repo). Other than that, existing unit tests mostly
suffice; one new unit test added.

wchargin-branch: graph-indexed-edges

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -48,6 +48,10 @@ export class Graph<NP, EP> {
   _nodeIndices: AddressMap<{|+address: Address, +index: Integer|}>;
   _nodes: MaybeNode<NP>[];
   _edges: AddressMap<IndexedEdge<EP>>;
+
+  // If `idx` is the index of a node `v`, then `_outEdges[idx]` is the
+  // list of `e.address` for all edges `e` whose source is `v`.
+  // Likewise, `_inEdges[idx]` has the addresses of all in-edges to `v`.
   _outEdges: Address[][];
   _inEdges: Address[][];
 
@@ -69,23 +73,23 @@ export class Graph<NP, EP> {
     if (theseNodes.length !== thoseNodes.length) {
       return false;
     }
-    for (const node of theseNodes) {
-      if (!deepEqual(node, that.node(node.address))) {
-        return false;
-      }
-    }
 
     const theseEdges = this.edges();
     const thoseEdges = that.edges();
     if (theseEdges.length !== thoseEdges.length) {
       return false;
     }
+
+    for (const node of theseNodes) {
+      if (!deepEqual(node, that.node(node.address))) {
+        return false;
+      }
+    }
     for (const edge of theseEdges) {
       if (!deepEqual(edge, that.edge(edge.address))) {
         return false;
       }
     }
-
     return true;
   }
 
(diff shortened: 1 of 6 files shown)


src/core/graph.js
```diff
@@ -19,88 +21,175 @@ export type Edge<+T> = {|
...
+        +address: Address,
+        +payload?: NP,
+      |},
+    |}[] = this._nodes
+      .map((maybeNode, oldIndex) => {
+        const key = stringify(maybeNode.address);
+        const data = maybeNode.node || {address: maybeNode.address};
+        return {key, oldIndex, data};
+      })
+      .filter(({oldIndex: idx}) => {
```
decentralion: Let's add a unit test that verifies that this filtering occurs

src/core/graph.js
```diff
@@ -19,88 +21,175 @@ export type Edge<+T> = {|
...
+        // Eliminate dangling nodes with neither in- nor out-edges:
+        // e.g., nodes that exist as a result of an `addNode` followed
+        // by a `removeNode` and no additional incident edges.
+        return (
+          this._nodes[idx].node !== undefined ||
+          this._outEdges[idx].length > 0 ||
+          this._inEdges[idx].length > 0
+        );
+      });
+    partialNodes.sort((a, b) => {
```
decentralion: Do we have a unit test verifying that the JSON representation is independent of node insertion order?
I know we test this for equality; we should check for JSON equality too.

src/core/graph.js
```diff
@@ -19,88 +21,175 @@ export type Edge<+T> = {|
...
+      const ka = a.key;
+      const kb = b.key;
+      return ka < kb ? -1 : ka > kb ? +1 : 0;
+    });
+    const oldIndexToNewIndex: Integer[] = Array(this._nodes.length).fill(
+      undefined
```
decentralion: nit: request comment explaining why

wchargin:
I've verified that the snapshot changes "look reasonable", and I hand-verified one of the `srcIndex`/`dstIndex` pairs (the MERGED_AS edge for the head commit of the example repo).

src/core/graph.js
```diff
@@ -256,7 +362,11 @@ export class Graph<NP, EP> {
    * If filter is provided, it will return only nodes with the requested type.
    */
   nodes(filter?: {type?: string}): Node<NP>[] {
-    let nodes = this._nodes.getAll();
+    /*:: declare function nonNulls<T>(x: (T | void)[]): T[]; */
+    let nodes = this._nodes.map((x) => x.node).filter((x) => {
```
decentralion: nit: filter((x) => Boolean(x))

src/core/graph.test.js
```diff
@@ -1088,6 +1088,15 @@ describe("graph", () => {
           JSON.stringify(demoData.advancedMealGraph().toJSON())
         );
       });
+      it("should canonicalize away phantom nodes", () => {
+        const g1 = new Graph().addNode(demoData.heroNode());
+        const g2 = new Graph()
+          .addNode(demoData.heroNode())
+          .addNode(demoData.mealNode())
+          .removeNode(demoData.mealNode().address);
+        expect(g1.equals(g2)).toBe(true);
```
decentralion: Thanks for checking that equality and serialization both ignore phantom nodes!

wchargin:
I’ve pushed one more commit, which adds a phantom node and a node with
`null` payload into the canonical graph demo data, so that all our test
cases benefit from them. I trust that this change will not be
contentious, so absent further feedback I’ll merge once it passes CI.

PR #296, opened 2018-05-23
Add design doc for address–payload unification

wchargin:
We can use this pull request for discussion.

### DIFF
--- a/src/core/address_payload_unification_design.md
+++ b/src/core/address_payload_unification_design.md
@@ -106,9 +106,9 @@ interface NodeReference {
   address(): Address;
   get(): ?Node<any, any>;
 
-  neighbors(options?: NeighborhoodOptions): Iterator<NodeReference>;
+  neighbors(options?: NeighborsOptions): Iterator<NodeReference>;
 }
-type NeighborhoodOptions = {|
+type NeighborsOptions = {|
   +nodeType?: string,
   +edgeType?: string,
   +direction?: "IN" | "OUT" | "ANY",
@@ -152,7 +152,7 @@ class IndexedNodeReference implements NodeReference {
     }
   }
 
-  neighbors(options?: NeighborhoodOptions) {
+  neighbors(options?: NeighborsOptions) {
     this._ensureUpToDate();
     if (this._index == null) {
       return [].values();
@@ -194,7 +194,7 @@ class NodeDelegateReference implements NodeReference {
   graph() { return this._base.graph(); }
   address() { return this._base.address(); }
   get() { return this._base.get(); }
-  neighbors(options?: NeighborhoodOptions) {
+  neighbors(options?: NeighborsOptions) {
     return this._base.neighbors(options);
   }
 }
@@ -364,7 +364,7 @@ declare class Graph /* no type parameters! */ {
   _indexVersion: number;
   _internalNeighbors(
     number: number,
-    options?: NeighborhoodOptions
+    options?: NeighborsOptions
   ): Iterator<NodeReference>;
   _addressToIndex(address: Address): ?number;
 }
@@ -521,14 +521,14 @@ const allFirstPrs = nodesV2({
 });
 
 type FilterV2 = {|+pluginName?: string, +id?: Object|};
-type NeighborhoodOptionsV2 = {|
+type NeighborsOptionsV2 = {|
   +nodeFilter?: FilterV2,
   +edgeFilter?: FilterV2,
   +direction?: "IN" | "OUT" | "ANY",
 |};
 
 declare function neighborsV2(
-  options?: NeighborhoodOptionsV2
+  options?: NeighborsOptionsV2
 ): Iterator<NodeReference>;
 declare function nodesV2(filter?: FilterV2): Iterator<Node<any, any>>;
 ```
--- a/src/core/address_payload_unification_design.md
+++ b/src/core/address_payload_unification_design.md
@@ -42,10 +42,10 @@ porcelain. For instance, the GitHub plugin might have:
 ```javascript
 type GithubNodeType = "PULL_REQUEST" | "PULL_REQUEST_COMMENT";  // e.g.
 class GithubNodePayload<+T: {+url: string}> implements NodePayload {
-  +record: T;
+  +_record: T;
   +_type: GithubNodeType;
   constructor(record: T, type: GithubNodeType) {
-    this.record = record;
+    this._record = record;
     this._type = type;
   }
   address() {
@@ -56,7 +56,7 @@ class GithubNodePayload<+T: {+url: string}> implements NodePayload {
     };
   }
   url() {
-    return this.record.url;
+    return this._record.url;
   }
   toJSON() { throw new Error("Must be implemented by subclass."); }
   static fromJSON() { throw new Error("Must be implemented by subclass."); }
@@ -72,10 +72,10 @@ class PullRequestPayload extends GithubNodePayload<PullRequestJson> {
   constructor(record: PullRequestJson) {
     super(record, "PULL_REQUEST");
   }
-  number(): number { return this.record.number; }
-  title(): string { return this.record.title; }
-  body(): string { return this.record.body; }
-  toJSON() { return toCompat(PULL_REQUEST_COMPAT_INFO, this.record); }
+  number(): number { return this._record.number; }
+  title(): string { return this._record.title; }
+  body(): string { return this._record.body; }
+  toJSON() { return toCompat(PULL_REQUEST_COMPAT_INFO, this._record); }
   static fromJSON(json) { return fromCompat(PULL_REQUEST_COMPAT_INFO, json); }
 }
 const PULL_REQUEST_COMPAT_INFO: CompatInfo = {
--- a/src/core/address_payload_unification_design.md
+++ b/src/core/address_payload_unification_design.md
@@ -427,9 +427,9 @@ references.
 This API encompasses an improvement to how nodes are handled. Clients
 will usually work with either a `Node<NR, NP>` or a `NodeReference`
 alone. There is still a bit of friction in that some methods are
-inherently on the reference while others are on the porcelain, but this
-is not actually a problem when the `Node<NR, NP>` is treated as a
-package (not destructured): consider
+inherently on the reference while others are on the payload, but this is
+not actually a problem when the `Node<NR, NP>` is treated as a package
+(not destructured): consider
 
 ```javascript
 function someActions(
--- a/src/core/address_payload_unification_design.md
+++ b/src/core/address_payload_unification_design.md
@@ -1,0 +1,556 @@
+# Design: Address–payload unification
+
+The purpose of this document is to describe a system by which we can
+remove the hard separation between node addresses and payloads, as
+described in [sourcecred/sourcecred#190][issue-190].
+
+[issue-190]: https://github.com/sourcecred/sourcecred/issues/190
+
+This document typechecks—run its source through [`print_code_fences`]
+and copy the result into https://flow.org/try.
+
+[`print_code_fences`]: https://gist.github.com/wchargin/e371d4f3047cf5d89e7ff9bb47224243
+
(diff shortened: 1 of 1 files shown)


src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+one for each logical type. These classes can provide domain-specific
+porcelain. For instance, the GitHub plugin might have:
+
+```javascript
+type GithubNodeType = "PULL_REQUEST" | "PULL_REQUEST_COMMENT";  // e.g.
+class GithubNodePayload<+T: {+url: string}> implements NodePayload {
+  +record: T;
+  +_type: GithubNodeType;
+  constructor(record: T, type: GithubNodeType) {
+    this.record = record;
```
decentralion: I'm curious why the record is a public field, as opposed to having a public method that returns the record. Everything else is accessed functionally, is there a reason to be inconsistent on this point?

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+    super(record, "PULL_REQUEST");
+  }
+  number(): number { return this.record.number; }
+  title(): string { return this.record.title; }
+  body(): string { return this.record.body; }
+  toJSON() { return toCompat(PULL_REQUEST_COMPAT_INFO, this.record); }
+  static fromJSON(json) { return fromCompat(PULL_REQUEST_COMPAT_INFO, json); }
+}
+const PULL_REQUEST_COMPAT_INFO: CompatInfo = {
+  type: "sourcecred/sourcecred/github/PullRequest",
```
decentralion: Hmm, when we serialize the Graph JSON this will be considerable bloat, and is wholly redundant with the `pluginName` and `type` fields. Can we think of a way to deduplicate this with the pluginName/type, or to use index references rather than including the full string every time?

If we internally store nodes in a `pluginName/type` hierarchy, then potentially we could apply the compat_info to an array of all the payloads rather than on each one; this would also make it easy to deduplicate the address info.

Also, I know you want to de-privilege the "type" field, but notice that this is an orthogonal case where we have a first-class `type` construct. More to say on this in the part of the doc where you propose de-privileging type.

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+
+```javascript
+interface NodeReference {
+  graph(): Graph;
+  address(): Address;
+  get(): ?Node<any, any>;
+
+  neighbors(options?: NeighborhoodOptions): Iterator<NodeReference>;
+}
+type NeighborhoodOptions = {|
```
decentralion: `s/Neighborhood/Neighbors/`?

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+    super(base);
+  }
+  *comments(): Iterator<PullRequestCommentReference> {
+    for (const neighbor of this.neighbors({
+      direction: "OUT",
+      nodeType: "PULL_REQUEST_COMMENT",
+      edgeType: "CONTAINS",
+    })) {
+      if (!(neighbor instanceof PullRequestCommentReference)) {
+        throw new Error(neighbor.constructor.name);
```
decentralion: Is there any reason not to move this pattern down into the `neighbors` method? I don't look forward to re-implementing this for every `.*Reference` accessor

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+  ): Iterator<NodeReference>;
+  _addressToIndex(address: Address): ?number;
+}
+type GraphJson = {/* elided */};
+```
+
+I’ll draw attention to a few things in this proposal, which I am open to
+changing.
+
+  - APIs provide iterators by default. Conversion with `Array.from` is
```
decentralion: Seems reasonable, I don't 100% love it because it adds some extra boilerplate, but I do like defaulting to giving us more expressive / flexible semantics.

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+    easy. Retrofitting clients to stream data when possible only becomes
+    harder as we go on.
+  - I’ve removed `neighborhood`. Clients of `Graph` will certainly want
```
decentralion: :+1:

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+
+Some of these latter questions apply in case (2), too.
+
+One resolution to, I think, all of these questions is to change the
+`PluginHandler.createReference` method so that it does not accept a
+`NodePayload`, or at least to make the payload optional. This has the
+nice property that `PluginHandler<NR, NP>` has two simple functions: one
+to enrich a base reference into an `NR`, and one to enrich a JSON blob
+into an `NP`. I suspect that this is likely a good solution, but I am
+interested to hear other opinions.
```
decentralion: I believe that requesting a reference to an address should always return a valid NodeReference (albeit one which might have no payload and no neighbors). Since we are able to provide valid implementations of all the NR functions even if it is not defined. Implementing inconsistent behavior where we don't return those perfectly valid NRs seems needlessly confusing to me. Your proposal on how to do this in a manner that is consistent with plugins makes sense to me.

src/core/address_payload_unification_design.md
```diff
@@ -0,0 +1,556 @@
...
+content-addressing might make more sense. But that is not an entirely
+straightforward change from our current model.
+
+I (@wchargin) still think that we should plan to eventually remove
+`type` from being privileged on the address. The `type` and `id`
+properties together actually represent a plugin-specific set of values
+that are used to uniquely determine the identity of a node, and which we
+therefore require to always be available—even when the full contents of
+a node are not loaded. This grants us more flexibility: perhaps it makes
+more sense to have
```
decentralion: In the examples below, you make a decent case for allowing other properties to be always included in the address-layer for a node (ie. required to get a reference). However, none of these examples make a case for removing type from its privileged position. Without type, it would be difficult to use those extra properties - if you can't filter by type, how do you know which properties will be available to you? Duck typing?

Our `compat` framework encodes the idea of type, and I think for very good reason. I think we should consistently enshrine type across the system, where type is not a string but a [pluginName: string, subtype: string] 2-tuple, or perhaps a [pluginName, subtypes....] k-tuple. It's actually the `id` that should be de-privileged. Git plugin types should have the option of using the `hash` as the id, or in the case of e.g. TreeEntrys using the union of `treeHash` and `entryName`. GitHub author-like nodes should be able to use `login` as id. And so forth.

wchargin: > However, none of these examples make a case for removing type from its
> privileged position. Without type, it would be difficult to use those
> extra properties - if you can't filter by type, how do you know which
> properties will be available to you? Duck typing?

I don’t follow the logic here (and I’m certainly not advocating for duck
typing). Who says “you can’t filter by type”? You’d be allowed to filter
by `type` just like any other property. If your filter includes a `type`
constraint and a node doesn’t have a `type` property, then the node is
not included in the results.

To put it another way: in a world where plugins can make their own
choices about what information is used to identify a node, there’s no
longer any _need_ to privilege `type`. I’m not trying to knock `type`
down a peg; I’m trying to raise everything else up. :-)

src/core/address_payload_unification_design.md
decentralion: Makes sense. Weak preference for `_record` then.

src/core/address_payload_unification_design.md
decentralion: Interesting point with the splitting-author case. I'm doubtful that we'd want a single `CompatInfo` for each of those node types; at the time when we split into three distinct types, they should have three `CompatInfos`, which reflects the fat that they can version independently.

It's not quite clear what to do with the legacy `Author` nodes; it seems to suggest that maybe the `CompatInfo` should be at the level of the plugin, not the individual node. Then the plugin can define compat handlers so that pre-split plugin data can be properly reconciled when it's loaded by post-split handlers.

src/core/address_payload_unification_design.md
decentralion: We could add e.g. a `ReferenceConstructorName` option to `NeighborOptions`, which either filters out references that don't have the right constructor, or requires that every neighbor have that constructor and throw an error if not (ie: the same semantics here).

My point is, your including this in the generic example for using a reference suggests that we expect basically every caller to neighbors to apply this logic, in which case we not be expecting clients to all re-write this boilerplate over and over.

src/core/address_payload_unification_design.md
decentralion: 1. seems totally reasonable to me.

I don't think 2. and 3. are necessary implications. When the graph gets a request for a NodeReference that does not exist in the graph, it sets its index to null. Whenever `_ensureUpToDate()` is called on that `NodeReference`, it checks its address in the graph to see if it has an index. If it does, then it can delegate to the graph to find its neighbors. If not, then it knows it has no empty and it does not have a corresponding payload. I believe this behavior is correct, and also these "phantom references" can be GC'd like any other object, are not pointed to by the graph, and generating them does not require polluting the graph's internal state.

decentralion:
Re types - yep, doesn't block this change; let's discuss more offline.
Re concise naming: I think the value of having a concise name for the `payload` is slightly greater than you appreciate in this case; it's not just about saving keystrokes. From the perspective of a `Node<NR,NP>`, the `ref` and `payload` are symmetric/parallel concepts, and that symmetry is a little more apparent if the names are of similar length. Thus, I think the value of consistency is both cognitive and orthographic :)

I would be exceedingly pleased if the names were precisely the same number of characters - as in `ref` and `dat` - but I understand if you prefer `data` :)

src/core/address_payload_unification_design.md
wchargin: I've adopted this and written the implications in the doc.

wchargin: (Didn’t see your last two comments at time of last push.)

It sounds like what you’re suggesting requires abandoning the goal that
we only create a `NodeReference` once per address. This is totally
reasonable, in my opinion.

It sounds like a consequence of what you’re suggesting is that we need
to bump the `_indexVersion` whenever a node is added to the graph.
Consider the following scenario:

```javascript
const g = new Graph();
const ref = g.nodeReference(anAddress);
// now, `ref`'s internal `index` is `null`
assert(g.get() == null);
g.addNode({address: anAddress, payload: {}});
assert(g.get() != null);
```

For the second assertion to succeed, the cache must have been
invalidated by the `addNode`. Or am I misunderstanding something?

decentralion:
>> It sounds like a consequence of what you’re suggesting is that we need
>> to bump the _indexVersion whenever a node is added to the graph.
>> Consider the following scenario:
>> 
>> ```
>> const g = new Graph();
>> const ref = g.nodeReference(anAddress);
>> // now, `ref`'s internal `index` is `null`
>> assert(g.get() == null);
>> g.addNode({address: anAddress, payload: {}});
>> assert(g.get() != null);
>> ```
>> 
>> For the second assertion to succeed, the cache must have been
>> invalidated by the addNode. Or am I misunderstanding something?

When the ref has a null internal index, we attempt to correct the index every time `_ensureUpToDate` is called on that reference, which is to say, every time a graph property is requested.
Thus, calling neighbors on a phantom reference is relatively expensive, as it always goes through stringification. However, if the reference ever stops being a phantom, on the first time it is successfully used, it will adopt the correct index, and future calls will be efficient.
Thus, all phantom-reference related work is handled not when the graph is updated, but when the phantom reference is actually used.

I believe having slow lookups on phantom references is fine, since you will never encounter them "organically" - i.e. through graph traversal or calls to `nodes`. You only get them when you explicitly construct them.

Make sense?

wchargin:
I went through `:%S/payload/data/gc`, and as I was approving each of the
individual changes I came to like “data” less and less… it just doesn’t
seem like the right name. Per discussion, we’ll stick with “payload” for
now.

PR #297, opened 2018-05-25
Merge `merge`s

wchargin:
Summary:
We had three graph merging functions: `merge`, `mergeConservative`, and
`mergeManyConservative`. Of these, `merge` was never used outside of
code directly testing its behavior, and `mergeConservative` is a
strictly inferior version of `mergeManyConservative`. This commit
removes `merge` and `mergeConservative`, and renames
`mergeManyConservative` to `mergeConservative`.

Paired with @decentralion.

Test Plan:
Existing unit tests suffice; some useless tests pruned.

wchargin-branch: mmeerrggee

### DIFF
--- a/src/core/graph.test.js
+++ b/src/core/graph.test.js
@@ -885,7 +885,7 @@ describe("graph", () => {
       });
     });
 
-    describe("merging", () => {
+    describe("mergeConservative", () => {
       /**
        * Decompose the given graph into neighborhood graphs: for each
        * node `u`, create a graph with just that node, its neighbors,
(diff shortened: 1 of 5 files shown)


src/core/graph.test.js
```diff
@@ -943,80 +943,30 @@ describe("graph", () => {
         return [].concat(edgeGraphs, nodeGraphs);
       }
 
-      it("conservatively recomposes a neighborhood decomposition", () => {
+      it("recomposes a neighborhood decomposition", () => {
         const result = neighborhoodDecomposition(
           demoData.advancedMealGraph()
-        ).reduce((g1, g2) => Graph.mergeConservative(g1, g2), new Graph());
+        ).reduce((g1, g2) => Graph.mergeConservative([g1, g2]), new Graph());
```
decentralion: I believe using reduce here is inefficient / a bad pattern, and only exists because we didn't have the current mergeConservative api when written. should be `Graph.mergeConservative(neighborhoodDecomposition(`

src/core/graph.test.js
```diff
@@ -1045,40 +995,30 @@ describe("graph", () => {
...
-          assertNotCalled,
-          assertNotCalled
-        );
+        ]);
         expect(merged.equals(demoData.advancedMealGraph())).toBe(true);
       });
-      it("has the empty graph as a right identity", () => {
-        const merged = Graph.merge(
+
+      it("has the empty graph as a left identity", () => {
```
decentralion: "has the empty graph as a left identity" - two tests of this name now?

src/core/graph.test.js
wchargin: Absolutely. Fixed.

src/core/graph.test.js
wchargin: Fixed.

PR #299, opened 2018-05-25
Fix tests for edge-induced graph disequality

wchargin:
Summary:
These tests have apparently been borked since the beginning. They were
testing properties about nodes instead of edges, had the wrong test
names, and had the LHS/RHS backward… this should make them more useful.

Test Plan:
Unit tests suffice, I suppose.

wchargin-branch: tests-for-missing-edges


wchargin:
Yarn’s registry is down, so Travis will fail. I’ve verified that `yarn
travis` passes locally, but I am happy to wait until our normal CI is
back up to merge.

PR #300, opened 2018-05-26
Remove `lint-staged` pre-commit hook

wchargin:
Summary:
This just slows down commits by a few seconds. We `check-pretty` in
Travis, so this doesn’t actually catch anything—and, anecdotally, it has
never caught anything for me because I automatically run `prettier` on
save and also (almost) always run Travis before pushing.

Test Plan:
Run `git commit --amend --no-edit` and note that it is now fast!

wchargin-branch: no-lint-on-commit


wchargin:
Again, Travis will fail because the Yarn registry is down.

PR #310, opened 2018-05-28
Create an explicit `PluginType`

decentralion:
Currently, when filtering by type (e.g. in neighbors), we require only a
type string. This is a design flaw, as if two plugins both define an
"ISSUE" type, either plugin may unexpectedly receive the other plugin's
nodes or edges.

We fix the flaw by explicitly binding the plugin name and type field
together as `PluginType`.

Test plan: Flow passes, and so does the following invocation to check
our design doc:
```
cat <(printf '// @flow\n') \
  <(awk -v RS='```[a-z]*' '(NR+1)%2' \
      src/core2/address_payload_unification_design.md) \
  | yarn -s flow check-contents
```

### DIFF
--- a/src/core2/address_payload_unification_design.md
+++ b/src/core2/address_payload_unification_design.md
@@ -17,15 +17,24 @@ point.
 
 ## Core structures
 
-The `Address` type is, for now, unchanged:
+The `Address` type is modified to have an explicit `PluginType`.
 
 ```javascript
-type Address = {|
-  +pluginName: string,
+type PluginType = {|
+  +plugin: string,
   +type: string,
+|};
+
+type Address = {|
+  +owner: PluginType,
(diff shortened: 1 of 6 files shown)


src/core2/address_payload_unification_design.md
```diff
@@ -219,8 +227,8 @@ class GithubNodeReference extends DelegateNodeReference {
   constructor(base: NodeReference) {
     super(base);
     const address = base.address();
-    if (address.pluginName !== "sourcecred/github-beta") {
-      throw new Error("pluginName: " + address.pluginName);
+    if (address.owner.plugin!== "sourcecred/github-beta") {
```
wchargin: nit: `s/!==/ !==` (missing whitespace)

PR #312, opened 2018-05-28
Graphv2: enable adding and retrieving nodes

decentralion:
This commit adds the following methods to the `Graph`:

* `addNode`
* `removeNode`
* `ref`
* `node`
* `nodes`
* `equals`

The graph now supports adding nodes, with thorough testing. Other
methods were implemented as necessary to test that `addNode` was
implemented properly.

Also, we've made a slight change to spec: `nodes` (and other filter
options) accept a `PluginFilter` object, which, if present, must
specify a plugin and may specify a type.

I've taken the opportunity to re-write the graph test code. Instead of
having a complicated `graphDemoData` file that creates a graph with many
different nodes, I've created an `examplePlugin` which makes it trivial
to instantiate new simple Foo and Bar nodes on the fly. Then, test cases
can construct a small graph that is clearly appropriate for whatever
functionality they are testing.

Test plan: Unit tests were added, travis passes.


src/core2/graph.js
```diff
@@ -145,8 +199,18 @@ export class Graph {
   }
 
   equals(that: Graph): boolean {
```
wchargin: We should document that this tests for structural equality, and does not check the `plugins`.

src/core2/graph.test.js
```diff
@@ -1,27 +1,217 @@
...
+      expect(fail).toThrow("exists with distinct contents");
+    });
+    it("errors for null or undefined payload", () => {
+      expect(() => newGraph().addNode((null: any))).toThrow("null");
+      expect(() => newGraph().addNode((undefined: any))).toThrow("undefined");
+    });
+  });
+
+  describe("removeNode", () => {
+    it("removing a nonexistant node is not an error", () => {
```
wchargin: sp.: *existent*

PR #320, opened 2018-05-30
Implement `mergeConservative` for GraphV2

wchargin:
Summary:
I dropped the `edgeDecomposition` and `neighborhoodDecomposition` test
cases, because (a) we don’t have a canonical “interesting graph” on
which to run them, compounding that (b) it’s not clear how much value
they add.

Test Plan:
New unit tests suffice.

wchargin-branch: mergeConservative-v2

### DIFF
--- a/src/core2/graph.js
+++ b/src/core2/graph.js
@@ -313,8 +313,16 @@ export class Graph {
     plugins: Plugins,
     graphs: $ReadOnlyArray<Graph>
   ): Graph {
-    const _ = {plugins, graphs};
-    throw new Error("Graphv2 is not yet implemented");
+    const result = new Graph(plugins);
+    graphs.forEach((graph) => {
+      for (const node of graph.nodes()) {
+        result.addNode(node.payload);
+      }
+      for (const edge of graph.edges()) {
+        result.addEdge(edge);
+      }
+    });
+    return result;
   }
 
   /**
--- a/src/core2/graph.test.js
+++ b/src/core2/graph.test.js
@@ -625,6 +625,104 @@ describe("graph", () => {
     });
   });
 
+  describe("mergeConservative", () => {
+    const merge = (graphs) => Graph.mergeConservative([new Handler()], graphs);
+
+    it("yields the empty graph on empty input", () => {
+      expect(merge([]).equals(newGraph())).toBe(true);
+    });
+
+    it("is the identity on a singleton input", () => {
+      const g = () =>
+        newGraph()
+          .addNode(new FooPayload())
+          .addEdge({
+            address: {
+              owner: {plugin: EXAMPLE_PLUGIN_NAME, type: "EDGE"},
+              id: "edge",
+            },
+            src: new FooPayload().address(),
+            dst: new BarPayload(1, "hello").address(),
+            payload: "nothing",
+          });
+      expect(merge([g()]).equals(g())).toBe(true);
+    });
+
+    it("merges two graphs with nontrivial intersection", () => {
+      const nodes = {
+        a: () => new BarPayload(1, "alpha"),
+        b: () => new BarPayload(2, "bravo"),
+        absent: () => new FooPayload(),
+      };
+      const edge = (srcPayload, dstPayload, id): Edge<string> => ({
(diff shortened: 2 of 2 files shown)


src/core2/graph.js
```diff
@@ -313,8 +313,16 @@ export class Graph {
     plugins: Plugins,
     graphs: $ReadOnlyArray<Graph>
   ): Graph {
-    const _ = {plugins, graphs};
```
decentralion: Hmm, I notice that we could instead have an API that doesn't require passing plugins, but instead takes the set union of plugins in the merged graphs. I posit that would basically always be "the right thing", the current api basically forces the user to separately track which plugins are in the merged graph, with the penalty that the call will fail if any necessary plugins were excluded. What do you think of making that change?

src/core2/graph.test.js
```diff
@@ -625,6 +625,104 @@ describe("graph", () => {
     });
   });
 
+  describe("mergeConservative", () => {
```
decentralion: It doesn't test the plugin behavior except implicitly (via not throwing an error from the plugin handler not being registered). Regardless of what the precise plugin behavior is, it would be good to test.

src/core2/graph.js
wchargin: It’s not clear to me that there is a meaningful definition of set union
for plugin lists, because it’s not clear that there’s a meaningful
definition for equality of plugin handlers. Suppose that `h1` and `h2`
are two plugin handlers with the same `.pluginName()`. It is clear that
in some cases we must throw a “duplicate plugin handler” error, and in
others we must recognize that `h1` and `h2` are “the same” and pick one
of them arbitrarily. But how do we make this choice?

We can’t rely on reference equality: if we have
```js
const g1 = new Graph([new GitPluginHandler()]);
const g2 = new Graph([new GitPluginHandler(), new GithubPluginHandler()]);
```
then `g1.plugins()[0] !== g2.plugins()[0]`.

We can’t rely on class/`.constructor` equality, for two reasons. First,
it would be totally reasonable to define
```js
class FooPluginHandler {
  constructor(someImportantOptions) {
    // …
  }
  // …
}
```
such that two instances of `FooPluginHandler` could really be
meaningfully different. Second, it is also totally reasonable to define
```js
export function barPluginHandler() {
  return {
    createRef: /* … */,
    createPayload: /* … */,
  };
}
```
such that a plugin handler is just an `Object`.

One attempt at a resolution would be to soft-mandate that plugins expose
not a class implementing `PluginHandler` but instead singleton instance
of such a class, with the understanding that this exact instance should
be used to represent the plugin.

What do you think?

src/core2/graph.test.js
wchargin: Are you suggesting that I add a test like
```js
it("installs appropriate plugin handlers", () => {
  const plugins = [new Handler()];
  expect(Graph.mergeConservative(plugins, []).plugins()).toEqual(plugins);
});
```
? If so, this seems reasonable to me, and I’m happy to add it. If not,
I don’t understand the comment.

decentralion: Yep, that was what I had in mind.

src/core2/graph.js
decentralion: I like the "singleton instance" proposal. We can then assume (and document) that all plugins with the same name are assumed equivalent, and if multiple distinct instances are in fact provided, we choose between them arbitrarily.

decentralion: Thanks for coming up with a thoughtful and interesting example of when we'd want to pass options from the plugin object. When we do add options like you described, I think we'd need to serialize them into the graph, otherwise graph semantics are not portable / reproducible. So we'd probably come up with some way of storing pluginOptions in the graph, and graph equality (and thus merge conservative's conservativism) would explicitly respect those pluginOptions (which would, of course, also have compat handlers). Thus, I think that even in light of your points, mergeConservative should eventually not require a plugins argument.

However, you've persuaded me that this merits careful attention, and there's no need to figure this all out now. Given that we have been quite good about refactoring APIs, I'm fine approving this with its current API and we can take another pass on it as we develop the plugin serialization story.

PR #322, opened 2018-05-31
Use payload-JSON equality for `Node`s

wchargin:
Summary:
A `Node` includes a ref and a payload. We should say that two nodes at
the same address are equal iff their payloads are equal, and two
payloads at the same address are equal iff their JSON contents are
equal. Importantly, `Node` equality does not depend on the `ref`’s
internal structure: this structure includes a reference to the enclosing
graph, and so using this notion of node equality caused `equals` to
incorrectly return `false` when the inputs were two logically equal
graphs with different internal structure. (It also caused `equals` to be
very slow, performing a deep equality check on the graphs for every
node.)

Test Plan:
A unit test has been strengthened. It fails before this patch, and
passes afterward.

wchargin-branch: v2-node-payload-json-equality

### DIFF
--- a/src/core2/graph.js
+++ b/src/core2/graph.js
@@ -343,8 +343,12 @@ export class Graph {
       return false;
     }
 
-    for (const node of theseNodes) {
-      if (!deepEqual(node, that.node(node.address))) {
+    for (const thisNode of theseNodes) {
+      const thatNode = that.node(thisNode.address);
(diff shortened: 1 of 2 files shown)


wchargin:
Yep. This was caught while testing `fromJSON` on a graph with phantom
nodes, which are removed through a serialization roundtrip… so at least
some test cases caught it! :-)

PR #323, opened 2018-05-31
Implement `toJSON` and `fromJSON` for GraphV2

wchargin:
Summary:
The storage format is similar to the V1 storage format. Nodes are sorted
by address and stored in an ordered list. Edges have `srcIndex` and
`dstIndex`, keyed against the node order. Each node stores the name of
its plugin and the result of invoking `toJSON` on its payload.

Test Plan:
Unit tests added; run `yarn travis`. The single added snapshot is easy
to read and verify.

wchargin-branch: v2-json


src/core2/graph.js
```diff
@@ -79,6 +80,18 @@ export interface PluginHandler<NR: NodeReference, NP: NodePayload> {
...
+type NodesSortedByStringifiedAddress = (
+  | {|+address: Address|}
+  | {|+payload: any, +pluginName: string|}
+)[];
+export type GraphJSON = {|
+  +nodes: NodesSortedByStringifiedAddress,
+  +edges: AddressMapJSON<IndexedEdge>,
+|};
+
+const COMPAT_TYPE = "sourcecred/sourcecred/Graph";
```
decentralion: Wondering what you think of just `sourcecred/Graph`? As in, the organization SourceCred will only ever publish one "Graph". Then it's actually portable if we refactor our repository structure.

src/core2/graph.js
```diff
@@ -369,17 +382,115 @@ export class Graph {
...
+    const partialNodes: {|
+      key: string,
+      oldIndex: Integer,
+      data:
+        | {|+address: Address|}
+        | {|+payload: NodePayload, +pluginName: string|},
+    |}[] = this._nodes
+      .map((maybeNode, oldIndex) => {
+        const key = stringify(maybeNode.address);
+
```
decentralion: nit: why this blankline?

src/core2/graph.test.js
```diff
@@ -808,6 +808,128 @@ describe("graph", () => {
...
+    };
+
+    it("serializes a sample graph", () => {
+      expect(sampleGraph().toJSON()).toMatchSnapshot();
+    });
+
+    it("installs appropriate plugin handlers", () => {
+      const g = newGraph();
+      const plugins = g.plugins();
+      expect(g.plugins()).not.toHaveLength(0);
```
decentralion: nit: s/g.plugins()/plugins/g

src/core2/graph.test.js
```diff
@@ -808,6 +808,128 @@ describe("graph", () => {
...
+      const g2 = newGraph().addEdge(edges.a_absent());
+      const g3 = newGraph()
+        .addNode(nodes.a())
+        .addEdge(edges.a_absent());
+      expect(g1.toJSON()).not.toEqual(g2.toJSON());
+      expect(g2.toJSON()).not.toEqual(g3.toJSON());
+      expect(g3.toJSON()).not.toEqual(g1.toJSON());
+    });
+
+    it("is not affected by phantom nodes", () => {
```
decentralion: Can we sandwich the phantom node in between two existing nodes? This forces a re-order.

src/core2/graph.js
wchargin: The principle of organizations owning compat types is fine with me. In
this particular case, it might be okay: “graph” is a pretty big class of
thing to promise to only ever create one of, but (a) maybe that’s true
for SourceCred, and (b) we’re well before the point where we provide any
actual compat guarantees. :-)

I’m happy to make this change if you like.

src/core2/graph.js
wchargin: It does look a bit out of place, doesn’t it? I’ll remove it.

src/core2/graph.test.js
wchargin: Sure; that makes sense.

src/core2/graph.test.js
wchargin: Good idea. Will do.

PR #324, opened 2018-06-01
Make Edges into Payload-esque objects

decentralion:
This commit changes `Edges` from being plain JS objects into
plugin-constructed class instances, like with NodePayload and
NodeReference. The interface is as follows:

```js
export interface Edge {
  /**
   * Convert this object to its serialized form. This must be a plain
   * old JSON value: i.e., a value `v` such that
   * `JSON.parse(JSON.stringify(v))` is deep-equal to `v`.
   * The `src` and `dst` should not be included in the JSON representation;
   * rather, the PluginHandler.createEdge method will be offered the src
   * and dst when regenerating the edge from JSON.
   */
  toJSON(): any;
  address(): Address;
  src(): NodeReference;
  dst(): NodeReference;
}
```

Plugin handlers must now provide a `createEdge` function:
```js
  /**
   * Deserialize a JSON edge payload, which is guaranteed to be the
   * seriaization of a previous `EP`.
   */
  createEdge(src: NodeReference, dst: NodeReference, json: any): E;
```

This has some nice consequences:
- We no longer need to serialize edge addresses, which are often very
long concatenations of two other addresses. Instead, we can serialize
the following `EdgeJSON`:

```js
type EdgeJSON = {|
  payload: any,
  plugin: string,
  srcIndex: Integer,
  dstIndex: Integer,
|};
type EdgesSortedByStringifiedAddress = EdgeJSON[];
export type GraphJSON = {|
  +nodes: NodesSortedByStringifiedAddress,
  +edges: EdgesSortedByStringifiedAddress,
|};
```

- Plugins can now provide Edge classes with refined type signatures, as
in the following example:
```js
export class MergedAsEdge {
  static TYPE: "MERGED_AS" = "MERGED_AS";
  _src: PullRequestReference;
  _dst: CommitReference;

  constructor(src: PullRequestReference, dst: CommitReference) {
    // ...
  }

  src(): PullRequestReference {
    return this._src;
  }
  // and so forth
}
```

- It makes the API overall feel more consistent. In the pre-refactor
world, all nodes and edges were stored as plain JS objects. Presently,
nodes are plugin-provided classes with accessor methods, but edges are
still plain JS objects. This commit unifies them.

However, it's a bit weird in this design that `Edges` (which are
nominally graph-agnostic) give `NodeReferences`, which are tied to
specific graphs. As implemented, graphs are permissive in that they will
accept edges from other graphs, but careful in what they return, in that
the edges they return always contain references to their own nodes.

Thus, it is not the case in general that adding an edge to a graph and
requesting it back gives a reference-identical edge; if the initial edge
contained references to a different graph, a new one is returned.

Also, the performance of this commit is not great, since internally the
edges go through address translation a lot: in `neighbors`, `addEdge`,
`removeEdge`, `toJSON`, `fromJSON`, etc. Most of these address
translations are unnecessary; if we refactor to expose an module-private
way to retrieve the index from an indexed `NodeReference`, we can solve
all of these performance issues without breaking the API.

Test plan: Many unit tests were added.

### DIFF
--- a/src/core2/examplePlugin.js
+++ b/src/core2/examplePlugin.js
@@ -1,10 +1,13 @@
 // @flow
 
+import stringify from "json-stable-stringify";
+
 import type {Address} from "./address";
 import {DelegateNodeReference} from "./graph";
-import type {NodeReference, NodePayload, PluginHandler} from "./graph";
+import type {NodeReference, NodePayload, Edge, PluginHandler} from "./graph";
 
 export type NodeType = "FOO" | "BAR";
+export type EdgeType = "SIMPLE" | "STRANGE";
 export const EXAMPLE_PLUGIN_NAME = "sourcecred/graph-demo-plugin";
 
 export class FooPayload implements NodePayload {
@@ -63,7 +66,77 @@ export class BarReference extends DelegateNodeReference {
   }
 }
 
-export class Handler implements PluginHandler<NodeReference, NodePayload> {
+export class SimpleEdge implements Edge {
+  _src: NodeReference;
+  _dst: NodeReference;
+
+  constructor(src: NodeReference, dst: NodeReference) {
+    this._src = src;
+    this._dst = dst;
+  }
+  src(): NodeReference {
+    return this._src;
+  }
+  dst(): NodeReference {
+    return this._dst;
+  }
+  address(): Address {
+    const srcFragment = stringify(this.src().address());
+    const dstFragment = stringify(this.dst().address());
+    const id = `SIMPLE:${srcFragment}-${dstFragment}`;
+    return {owner: {plugin: EXAMPLE_PLUGIN_NAME, type: "SIMPLE"}, id};
+  }
+  toJSON() {
+    return {type: "SIMPLE"};
+  }
+}
+
+export class StrangeEdge implements Edge {
+  _src: NodeReference;
+  _dst: NodeReference;
+  _strangeness: number;
+  _name: string;
+
+  constructor(
+    src: NodeReference,
+    dst: NodeReference,
+    name: string,
+    strangeness: number
+  ) {
+    this._src = src;
+    this._dst = dst;
+    this._name = name;
+    this._strangeness = strangeness;
+  }
+  src(): NodeReference {
+    return this._src;
+  }
+  dst(): NodeReference {
+    return this._dst;
+  }
+  strangeness(): number {
+    return this._strangeness;
+  }
+  name(): string {
+    return this._name;
+  }
+  address(): Address {
+    return {
+      owner: {plugin: EXAMPLE_PLUGIN_NAME, type: "STRANGE"},
+      id: this._name,
+    };
+  }
+  toJSON() {
+    return {type: "STRANGE", name: this._name, strangeness: this._strangeness};
+  }
+}
+
+export class Handler
+  implements PluginHandler<
+      NodeReference,
+      NodePayload,
+      SimpleEdge | StrangeEdge
+    > {
   createReference(ref: NodeReference) {
     const type: NodeType = (ref.address().owner.type: any);
     switch (type) {
@@ -89,6 +162,20 @@ export class Handler implements PluginHandler<NodeReference, NodePayload> {
         // eslint-disable-next-line no-unused-expressions
         (type: empty);
         throw new Error(`Unexpected NodeType: ${type}`);
+    }
+  }
+
+  createEdge(src: NodeReference, dst: NodeReference, json: any) {
+    const type: EdgeType = json.type;
+    switch (type) {
+      case "SIMPLE":
+        return new SimpleEdge(src, dst);
+      case "STRANGE":
+        return new StrangeEdge(src, dst, json.name, json.strangeness);
+      default:
+        // eslint-disable-next-line no-unused-expressions
+        (type: empty);
+        throw new Error(`Unexpected EdgeType: ${type}`);
     }
   }
 
--- a/src/core2/graph.js
+++ b/src/core2/graph.js
@@ -2,7 +2,7 @@
 
 import type {Compatible} from "../util/compat";
 import {toCompat, fromCompat} from "../util/compat";
-import type {Address, AddressMapJSON} from "./address";
+import type {Address} from "./address";
 import {AddressMap} from "./address";
 import stringify from "json-stable-stringify";
 import deepEqual from "lodash.isequal";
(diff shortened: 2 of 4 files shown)


src/core2/graph.js
```diff
@@ -250,60 +253,83 @@ export class Graph {
...
-    this._edges.add(indexedEdge);
-    this._outEdges[indexedEdge.srcIndex].push(indexedEdge.address);
-    this._inEdges[indexedEdge.dstIndex].push(indexedEdge.address);
+    // We make a redundant trip through the addressing layer; since
+    // src and dst are both refs, we've likely already done this.
+    // Potential perf optimization to internally expose the index directly.
+    const srcIndex = this._addNodeAddress(edge.src().address());
+    const dstIndex = this._addNodeAddress(edge.dst().address());
+
+    this._edges.add({edge, address: edge.address()});
```
wchargin: You have `address` in scope.

src/core2/graph.js
```diff
@@ -250,60 +253,83 @@ export class Graph {
...
-      [
-        this._outEdges[indexedEdge.srcIndex],
-        this._inEdges[indexedEdge.dstIndex],
-      ].forEach((edges) => {
-        const index = edges.findIndex((ea) => deepEqual(ea, address));
+      // As above: Potential perf gain if get index w/o address translation
+      const srcIndex = this._addNodeAddress(edge.src().address());
+      const dstIndex = this._addNodeAddress(edge.dst().address());
+      [this._outEdges[srcIndex], this._inEdges[dstIndex]].forEach((edges) => {
+        const index = edges.findIndex((edge_) =>
```
wchargin: Can we call `edge_` something less confusing, like `candidate` or `x`?

src/core2/graph.test.js
```diff
@@ -74,42 +83,42 @@ describe("graph", () => {
...
-        src: src.address(),
-        dst: dst.address(),
-        payload: {},
-      });
-      // A cute little diagram:
-      // A>B = edge from A to B
-      // Nodes in the graph: bar, foo, isolated
-      // (Parens) = Repeated or absent node
-      //
+      // Here's a diagram of the example graph:
```
wchargin: Aww, no longer cute.

src/core2/graph.test.js
```diff
@@ -118,11 +127,12 @@ describe("graph", () => {
       };
 
       function expectNeighborsEqual(
-        actual: Iterable<{|+edge: Edge<any>, +ref: NodeReference|}>,
-        expected: {|+edge: Edge<any>, +ref: NodeReference|}[]
+        actualIterator: Iterable<{|+edge: Edge, +ref: NodeReference|}>,
```
wchargin: If you call this an `actualIterator`, maybe make its type `Iterator<>`?

But why the name change at all?

src/core2/graph.test.js
```diff
@@ -162,8 +172,8 @@ describe("graph", () => {
       });
       describe("filters edges by type:", () => {
         [
-          ["EDGE", [fooNeighbor.bar, fooNeighbor.absent]],
-          ["SELF", [fooNeighbor.foo]],
+          ["SIMPLE", [fooNeighbor.bar, fooNeighbor.absent]],
```
wchargin: Why not use `SimpleEdge.TYPE` here (and create that accordingly), as we
expect actual clients to do?

src/core2/graph.test.js
```diff
@@ -577,7 +585,11 @@ describe("graph", () => {
     });
 
     it("throws for null or undefined address", () => {
-      const e = (address: any) => ({...edge(), address});
+      const e = (address: any) => {
+        const edge = new SimpleEdge(srcRef(), dstRef());
+        (edge: any).address = () => address;
```
wchargin: Eek. How about instead:

```js
const e = (address: null | void) =>
  new class extends SimpleEdge {
    address() {
      return (address: any);
    }
  }(srcRef(), dstRef());
```

As you use similar things a bunch below, maybe

```js
class ShiftyEdge extends SimpleEdge {
  _overrides: {|+src?: any, +dst?: any, +address?: any|};
  constructor(overrides, ...args) {
    super(...args);
    this._overrides = overrides;
  }
  src() {
    return "src" in this._overrides ? this._overrides.src : super.src();
  }
  dst() {
    return "dst" in this._overrides ? this._overrides.dst : super.dst();
  }
  address() {
    return "address" in this._overrides
      ? this._overrides.address
      : super.address();
  }
}
```

?

src/core2/graph.test.js
```diff
@@ -597,14 +613,31 @@ describe("graph", () => {
...
+      const graph = newGraph();
+      const otherGraph = newGraph();
+      const srcRef = otherGraph.ref(new FooPayload().address());
+      const edge = new SimpleEdge(srcRef, srcRef);
+      graph.addEdge(edge);
+      const edgeFromGraph = graph.edge(edge.address());
+      if (edgeFromGraph == null) {
+        throw new Error("Shocking turn of events");
+      }
+      expect(edgeFromGraph.src().graph()).toEqual(graph);
```
wchargin: This should be `toBe` instead of `toEqual`; we don’t want a deep
equality check.

src/core2/examplePlugin.js
```diff
@@ -63,7 +66,77 @@ export class BarReference extends DelegateNodeReference {
...
+  }
+  toJSON() {
+    return {type: "SIMPLE"};
+  }
+}
+
+export class StrangeEdge implements Edge {
+  _src: NodeReference;
+  _dst: NodeReference;
+  _strangeness: number;
```
wchargin: Why do we need both a strangeness and a name?

src/core2/graph.js
```diff
@@ -250,60 +253,83 @@ export class Graph {
...
+        this.ref(dst.address()),
+        edge.toJSON()
+      );
     }
-    this._edges.add(indexedEdge);
-    this._outEdges[indexedEdge.srcIndex].push(indexedEdge.address);
-    this._inEdges[indexedEdge.dstIndex].push(indexedEdge.address);
+    // We make a redundant trip through the addressing layer; since
+    // src and dst are both refs, we've likely already done this.
+    // Potential perf optimization to internally expose the index directly.
```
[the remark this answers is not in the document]
wchargin: After discussion—I was misunderstanding the problem here. The above code
is sound, but wouldn’t do what we want, because almost always the
reference that we get is a delegate reference, so it won’t actually be
an instance of `IndexedReference`.

Furthermore, it is not unreasonable for a delegate reference to be
wrapping another delegate reference, so it’s not quite sufficient to
just say “if it’s a `DelegateReference`, unpack the internal base
reference, which should probably be an `IndexedReference`”—we’d have to
do this recursively, and even then this would only work with the
`DelegateReference` implementation that we know about, as opposed to any
other kind of implementation that users are totally free to use.

This is kind of a big problem, in my opinion: it changes the perf TODOs
from “not going to bother doing this right now” to “it’s not clear that
there’s actually a good way to do this, and we are incurring away what
might reasonably be a significant performance hit”.

decentralion:
Now that we're dropping v2 in favor of v3, I think it's safe to close this PR.

PR #329, opened 2018-06-02
Implement {to,from}{Node,Edge}Address

decentralion:
This commit implements Node/Edge addresses, and helper functions for
generating and manipulating them.

Paired with @wchargin

### DIFF
--- a/src/v3/core/graph.js
+++ b/src/v3/core/graph.js
@@ -1,5 +1,6 @@
 // @flow
 
+import stringify from "json-stable-stringify";
 export opaque type NodeAddress = string;
 export opaque type EdgeAddress = string;
 export type Edge = {|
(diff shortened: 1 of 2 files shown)


decentralion:
BTW, when we paired this commit, we forgot to consider the case where `toNodeAddress` is passed an empty array. This resulted in a node address without any null separators, which `fromNodeAddress` interpreted as a  malformed node address. I've changed the behavior so `toNodeAddress` errors on an empty array input, and tested this.

wchargin:
Oh, I see your comment now:

> we forgot to consider the case where toNodeAddress is passed an empty
> array. This resulted in a node address without any null separators,
> which fromNodeAddress interpreted as a malformed node address.

That’s just an implementation error, then. We should fix the
implementation, not disallow the empty array.

I see two reasonable options:

  - We could change the representation from “NUL between components” to
    “NUL after each component”, so that `["foo"]` becomes `"N\0foo\0"`
    and `[]` becomes `"N\0"`.

  - We could keep the current representation and fix the `assert`
    functions, to check that the string’s first NUL-separated component
    is `N`: i.e., that the string either is literally `"N"` or starts
    with `"N\0"`. This should be equivalent to checking that
    `theString.split("\0")[0] === "N"`, which is the intended semantics
    anyway—the fact that this had the wrong behavior on the empty
    array is an error.

Either of these is fine with me.

decentralion:
Makes sense. I changed the implementation to always include a NUL separator after every path component; this involved re-writing the methods. I also refactored the test code. PTAL.

src/v3/core/graph.test.js
```diff
@@ -0,0 +1,94 @@
...
+    describe("toNodeAddress & fromNodeAddress", () => {
+      throwOnNullOrUndefined(toNodeAddress);
+      throwOnNullOrUndefined(fromNodeAddress);
+      it("toNodeAddress errors on path containing NUL char", () => {
+        expect(() => toNodeAddress(["foo", "bar\0", "zoink"])).toThrow(
+          "NUL char"
+        );
+      });
+      describe("compose to identity", () => {
+        function checkIdentity(name, example) {
```
wchargin: +1

src/v3/core/graph.js
```diff
@@ -20,22 +21,72 @@ export type NeighborsOptions = {|
...
+      throw new Error(`NUL char: ${stringify(arr)}`);
+    }
+  });
+}
+
 export function toNodeAddress(arr: $ReadOnlyArray<string>): NodeAddress {
-  const _ = arr;
-  throw new Error("toNodeAddress");
+  assertAddressArray(arr);
+  if (arr.length > 0) {
```
wchargin: Okay. This implementation isn’t quite as neat as I’d like (something
like `["N", ...arr, ""].join(SEPARATOR)` would be better, in my
opinion), but not a big deal.

src/v3/core/graph.js
```diff
@@ -20,22 +21,72 @@ export type NeighborsOptions = {|
...
+function assertNodeAddress(x: NodeAddress) {
+  if (x == null) {
+    throw new Error(String(x));
+  }
+  if (!x.startsWith(BASE_NODE_ADDRESS)) {
+    if (x.startsWith(BASE_EDGE_ADDRESS)) {
+      throw new Error(`Expected NodeAddress, got EdgeAddress: ${x}`);
+    }
+    throw new Error(`Malformed NodeAddress: ${x}`);
+  }
```
wchargin: Should we check that the input ends with a `SEPARATOR`?

src/v3/core/graph.test.js
```diff
@@ -0,0 +1,94 @@
...
+          it(name, () => {
+            expect(fromNodeAddress(toNodeAddress(example))).toEqual(example);
+          });
+        }
+        checkIdentity("on a simple example", ["an", "example"]);
+        describe("with an empty component", () => {
```
wchargin: Nice, thanks.

decentralion:
Thanks for the review comments. I made the following changes, based on your suggestions:
- Switched back to having `X_PREFIX` rather than `BASE_X_ADDRESS`, and now using the cleaner `toXAddress` implementation you suggested
- We now check that valid addresses end with a separator
- Factored out non-exported function `isNodeAddress`, which was convenient once checking involved multiple properties, and will be helpful for the extra assertions in https://github.com/sourcecred/sourcecred/pull/332

src/v3/core/graph.js
decentralion: Done.

src/v3/core/graph.js
```diff
@@ -20,22 +21,71 @@ export type NeighborsOptions = {|
...
+    throw new Error(`Malformed address: ${x}`);
+  }
+}
+
+function assertAddressArray(arr: $ReadOnlyArray<string>) {
+  if (arr == null) {
+    throw new Error(String(arr));
+  }
+  arr.forEach((s: string) => {
+    if (s.indexOf(SEPARATOR) !== -1) {
```
wchargin: It would be nice to first check that `s != null`.

PR #334, opened 2018-06-04
Reorganize Graph address-related code

decentralion:
Reorganize Graph address-related code

Our Graph uses specialized strings as addresses. These addresses are
exported as an opaque type; they can't be manipulated via standard
string methods, but instead require special address-aware methods.
Addresses have path-like semantics (they are hierarchically orgnaized by
string components), and we expect to have methods for appending
components, finding prefixes, concatenting addresses, mapping them to
and from string arrays, etc.

It would be nice to attach these methods to the addresses, so it
would be possible to call `address.append(component)`,
`address.prefix()`, and so forth. However, as addresses are strings,
this would require modifying the string prototype. We could make
addresses classes instead, that are backed by or convertible to strings,
but this would break the ability to test address equality using `===`.

After some offline discussion with @wchargin, we decided the best
compromise is to create a "psuedo-module" which contains all address
related methods, and can be imported from the `core/graph` module. This
commit creates that module, called `Address`, which contains the
following methods:

- `nodeAddress`: The factory function for node addresses. Returns a base
`NodeAddress` if called without arguments, or constructs a componentized
address if called with an array of parts.

- `edgeAddress`: The factory function for edge addresses. Returns a base
`EdgeAddress` if called without arguments, or constructs a componentized
address if called with an array of parts.

- `toParts`: Takes a node or edge address and returns a string array of
parts.

The module is implemented in `address.js`. Clients are not intended to
import `address.js` directly; rather, `graph.js` re-exports the public
pieces. (This is just for clients' convenience; being able to import all
Graph-required functionality from one file is pleasant.)

A note about `NodeAddress` and `EdgeAddress`: There are actually two
types of Addresses (Node and Edge), which have different
representations. It is a type and runtime error to pass a `NodeAddress` to
a function expecting an `EdgeAddress` and vice versa. Offline, @wchargin
and I discussed exporting two psuedomodules, one for node and one for
edge addresses. On further consideration, I believe this introduces
needless API complexity. It will be easy to write the single `Address`
module with functions that are apprporiately generic over Node/Edge
addresses, so we will still get type (and runtime) errors when passing
(for example) an `EdgeAddress` into `Graph.hasNode`.

`address.js` also exports `assertNodeAddress` and `assertEdgeAddress`.
These functions aren't yet used, but will be needed for appropriate
runtime assertions in `Graph` methods.

Test plan: Observe unit tests, Travis.


decentralion:
> Why is `address.js` this kind of pseudo-private module that shouldn’t be
> directly imported by clients? It could just expose:
> 
> ```js
> declare module "address" {
>   declare opaque type NodeAddress;
>   declare opaque type EdgeAddress;
>   declare function nodeAddress(parts: $ReadOnlyArray<string>): NodeAddress;
>   declare function edgeAddress(parts: $ReadOnlyArray<string>): EdgeAddress;
>   declare function toParts(address: NodeAddress | EdgeAddress): string[];
> }
> ```

In offline discussions, you've noted that it's annoying importing the Graph from one file and the Address type from another. The psuedo-private module means it can be canonically imported `import {Address} from "../../core/graph"`, and come with all address methods attached, similar to how if we had used a class to back address, we wouldn't need to import every method individually. 

I'm fine having `Address` be an actual, regular module though.

> Why export an object `_Address` containing functions? 

Convenience for the caller, to have them all packaged together. (I guess this could be replicated via `import * as Address from "../../core/address"`)

> This is unsafe (callers may freely and typesafely modify the attributes on the object)

Didn't consider this.

> and annoying (callers can no longer import individual values). If you
export values individually, callers can either import them individually
or `import * as Address from "./address"` (which _is_ safe).

Yep, maybe that's better.

> It seems desirable that `NodeAddress` and `EdgeAddress` should be
[declared as opaque subtypes][1] of `string`:

> ```js
> export opaque type NodeAddress: string = string;
> export opaque type EdgeAddress: string = string;
> ```

Sounds good, happy to do that.

> I’m not dead-set against using union types in public APIs for things
like `toParts` and `concat`. However, I thoroughly disagree that this is
“simpler”—one function that has a bounded wildcard with an upper bound
of a union type is clearly more complicated than two monomorphic
functions with simple domain and codomain. Is there an advantage here
besides implementation laziness (no offense)? 

I think we are thinking of "simplicity" in different contexts. I'm thinking about simplicity for the caller. The caller can have one set of functions for modifying addresses rather than needing to have functions specialized for node addresses and functions specialized for edge addresses.

The simplicity you're describing here - type implementation simplicity - is, imo, less important than API simplicity.

> That is, under what circumstances is there a value whose type is known to be “either
`NodeAddress` or `EdgeAddress`, but we’re not sure which one”? This
can’t be a very useful type for callers, because it is incompatible with
every method on `Graph` (you can’t `addNode` or `hasEdge` or anything,
and with good reason). I’m feeling very “when in doubt, leave it out”.

I'm not sure how you imagine getting that type. Given that the user started with a `NodeAddress`, the polymorphic function will always return a `NodeAddress`, and likewise for `EdgeAddress`. 

Consider the following two snippets of code:

v1:
```js
const base = Address.nodeAddress(); // type NodeAddress
const scoped = Address.append(base, "github"); // type NodeAddress
const thisAddress = Address.append(scoped, this.id); // type NodeAddress
this.graph.addNode(thisAddress); // no flow error
```

v2:
```js
const base = NodeAddress.address(); // type NodeAddress
const scoped = NodeAddress.append(base, "github"); // type NodeAddress
const thisAddress = NodeAddress.append(scoped, this.id); // type NodeAddress
this.graph.addNode(thisAddress); // no flow error
```
In neither case do we ever instantiate something with type `NodeAddress | EdgeAddress`. The difference is the first does it with fewer API elements (since we don't need a separate `EdgeAddress` module).

Given that it doesn't produce any type-unsafety, I don't understand opposition to making the Address APIs generic. If we were implementing Python, would the same reasoning suggest that it's better to have `lenOfList` and `lenOfSet` as separate methods?

> Why allow `nodeAddress` and `edgeAddress` to be called with `undefined`?
This seems like a great way to mask bugs: if a client passes `undefined`
instead of an array, it seems at least as likely that the client has a
bug causing their intended-array to be `undefined` than that they
actually want the address given by `[]`. If a client does want this
semantics, they can just write `myArrayOrUndefined || []`.

I wanted to write the factory method so that it can be called without arguments to generate a base address. Having looked into this matter a little more, I see that `arguments.length` is a better way to achieve this.

> The following two tests would have passed in the previous API, but fail
> under the code in this pull request:

That's a bug in the implementation, thanks for catching.

> Note that `typeof NODE | typeof EDGE` is just `Symbol`; there is no
additional safety granted by using that definition.

Interesting, I guess using strings is better for that purpose.

decentralion:
Will wait on in-person discussion to resolve the contentious bits. :)

decentralion:
Discussed offline and resolved:
- To have a single `Address` module in favor of separate `NodeAddress` / `EdgeAddress`
- To have the `nodeAddress` and `edgeAddress` functions always require an explicit parts argument

Otherwise, things were resolved per William's suggestions.

PR #335, opened 2018-06-04
Upgrade prettier to 1.13

decentralion:
Release notes: https://prettier.io/blog/2018/05/27/1.13.0.html


src/v1/cli/commands/combine.js
```diff
@@ -24,7 +25,8 @@ export default class CombineCommand extends Command {
   // [1]: https://oclif.io/docs/commands.html
   static strict = false;
 
-  static usage = "" +
+  static usage =
+    "" +
```
wchargin: Not asking for a change (rather, I like that this commit is reproducible
by “run prettier” only), but this `"" + ` now seems redundant (yay).

Issue #269, opened 2018-05-11
Add tests for Cred Explorer


decentralion:
In https://github.com/sourcecred/sourcecred/pull/265, we created a new UI for the Cred Explorer and merged it without tests. This was reasonable, as we were pushing to build out a lot of new features, and indeed succeeded in prototyping a much faster and more configurable PageRank implementation!

However, we should not forget to add tests, e.g. for the following properties:

- [x] Expand/collapse recursive tables
- [x] Nested tables have a darker color
- [x] Top-level filtering is possible via the selector
- [x] Scores (and log scores) are displayed 
- [x] Top level table is sorted by score
- [x] And recursive ones are too
- [x] Dangling edges do not cause a crash
- [ ] Changing the graph does not cause a crash
- [x] nodes without plugin adapter don't cause a crash

PR #342, opened 2018-06-05
Limit Cred Explorer to display 100 entries

decentralion:
Right now, the cred explorer attempts to display every node in the
graph. As graphs easily grow to O(100k) nodes, this is not kind to the
browser.

This commit limits display to the first 100 entries. Since they are
sorted by percieved importance, and it's easy to filter by type (e.g. to
find all users), this limitation is fine in practice.

Test plan:
Run the cred explorer on a larger repo and observe that the performance
is enormously improved. No unit tests added, as the cred explorer is a
prototype which is basically untested (#269)

### DIFF
--- a/src/v1/app/credExplorer/pagerankTable.js
+++ b/src/v1/app/credExplorer/pagerankTable.js
@@ -13,6 +13,8 @@ import {nodeDescription as githubNodeDescription} from "../../plugins/github/ren
 import {nodeDescription as gitNodeDescription} from "../../plugins/git/render";
 import type {PagerankResult} from "./basicPagerank";
 
+const MAX_TABLE_ENTRIES = 100;
+
 type Props = {
   pagerankResult: ?PagerankResult,
   graph: ?Graph,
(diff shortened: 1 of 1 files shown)


decentralion:
> If I’m understanding the code correctly, this limitation only applies to
the top level. As soon as you expand the tree view, you’re back to the
full contents. In particular, if you expand the root repository node, or
the node for a user with lots of edges, you’re in just as much of a
pickle as before this commit. Is this correct?

That's correct, and an oversight on my part. I've fixed that.

> This also, of course, doesn’t have pagination.

Yep!

> I would prefer a solution that applied at each layer of the recursive
table a pair of “show more” and “show less” buttons, or something else.
What do you think?

That's reasonable, and not a priority for me right now. I just want the demo to visibly "work" on larger examples; enabling much deeper dives doesn't add much value until we are getting the results themselves to a higher quality. (Punt!)

wchargin:
Mind rebasing on #343?

PR #348, opened 2018-06-05
Graph: implement edge methods

decentralion:
This commit implements the following edge related methods on graph:

- `Graph.addEdge(edge)`
- `Graph.hasEdge(address)`
- `Graph.edge(address)`
- `Graph.edges()`

We've decided to enforce an invariant that for every edge, its `src` and
`dst` nodes must be present in the graph. As such, `Graph.addEdge` may
error if this condition is not true for the proposed edge, and
`Graph.removeNode` may error if any edges depend on the node being
removed. This invariant is documented via comments, and checked by the
test code.

Test plan:
Extensive unit tests have been added. Run `yarn travis`.

Paired with @wchargin

### DIFF
--- a/src/v3/core/graph.js
+++ b/src/v3/core/graph.js
@@ -26,11 +26,14 @@ export type NeighborsOptions = {|
 export opaque type GraphJSON = any; // TODO
 
 export class Graph {
+  // A node `n` is in the graph if `_nodes.has(n)`.
+  //
+  // An edge `e` is in the graph if `_edges.get(e.address)`
+  // is deep-equal to `e`.
+  //
+  // Invariant: If an edge `e` is in the graph, then `e.src` and `e.dst`
+  // are both in the graph.
   _nodes: Set<NodeAddress>;
-  // If `e` is an Edge in the graph, then:
-  // * _edges.get(e.address) `deepEquals` e
-  // * _inEdges.get(e.dst) `contains` e
-  // * _outEdges.get(e.src) `contains` e
   _edges: Map<EdgeAddress, Edge>;
   _inEdges: Map<NodeAddress, Edge[]>;
   _outEdges: Map<NodeAddress, Edge[]>;
@@ -50,6 +53,14 @@ export class Graph {
 
   removeNode(a: NodeAddress): this {
     Address.assertNodeAddress(a);
+    for (const e of this.edges()) {
+      if (e.src === a || e.dst === a) {
+        const srcOrDst = e.src === a ? "src" : "dst";
+        throw new Error(
+          `Attempted to remove ${srcOrDst} of ${edgeToString(e)}`
+        );
+      }
+    }
     this._nodes.delete(a);
     return this;
   }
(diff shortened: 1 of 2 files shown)


src/v3/core/graph.js
```diff
@@ -26,14 +27,23 @@ export type NeighborsOptions = {|
 export opaque type GraphJSON = any; // TODO
 
 export class Graph {
+  // A node `n` is in the graph if `_nodes.has(n)`.
+  //
+  // An edge `e` is in the graph if `_edges.get(e.address)`
+  // is deep-equal to `e`.
+  //
+  // Invariant: For an edge `e`, the following are equivalent:
```
wchargin: Note that this invariant is knowingly violated in this commit (we do not
use `_inEdges`/`_outEdges` because we do not yet have `neighbors`).

We could remove this invariant, add a note that it’s intentionally
violated in this commit, or leave it as is (my preferences in that
order). Thoughts?

src/v3/core/graph.js
```diff
@@ -26,14 +27,23 @@ export type NeighborsOptions = {|
...
+  // are both in the graph.
   _nodes: Set<NodeAddress>;
-  // If `e` is an Edge in the graph, then:
-  // * _edges.get(e.address) `deepEquals` e
-  // * _inEdges.get(e.dst) `contains` e
-  // * _outEdges.get(e.src) `contains` e
   _edges: Map<EdgeAddress, Edge>;
   _inEdges: Map<NodeAddress, Edge[]>;
   _outEdges: Map<NodeAddress, Edge[]>;
+  _nodes: Set<NodeAddress>;
```
wchargin: Duplicate?

src/v3/core/graph.js
```diff
@@ -63,28 +81,51 @@ export class Graph {
...
+    }
+    const existingEdge = this._edges.get(edge.address);
+    if (existingEdge != null) {
+      if (
+        existingEdge.src !== edge.src ||
+        existingEdge.dst !== edge.dst ||
+        existingEdge.address !== edge.address
+      ) {
+        throw new Error(
+          `conflict between ${stringify(edge)} and ${stringify(existingEdge)}`
```
wchargin: Should be `edgeToString(…)`.

src/v3/core/graph.test.js
```diff
@@ -96,14 +122,207 @@ describe("core/graph", () => {
...
+        const dst = nodeAddress(["dst"]);
+        const address = edgeAddress(["hi"]);
+        it("`addEdge`", () => {
+          const g = new Graph()
+            .addNode(src)
+            .addNode(dst)
+            .addEdge({src, dst, address})
+            .addEdge({src, dst, address});
+          expect(edgeArray(g)).toEqual([{src, dst, address}]);
+        });
```
wchargin: One thing that we don’t test here is that if you add a node twice, then
remove it once, the node is actually removed (not “buffered”). That’s
probably okay to omit. Thoughts?

src/v3/core/graph.js
decentralion: Removed.

src/v3/core/graph.js
decentralion: Fixed.

src/v3/core/graph.js
decentralion: Fixed.

src/v3/core/graph.test.js
decentralion: Yeah, I think it's fine.

PR #354, opened 2018-06-06
Make all `neighbors` options required

wchargin:
Summary:
In actual code, we almost always call `neighbors` with a specified
direction. Usually, you want to inspect some relation like “parents of
this commit” or “GitHub nodes referenced by this comment”, and so the
edge direction matters. In each of the above cases, forgetting to
include the direction would introduce a bug: you’d get parents _and
children_ of a commit, or GitHub nodes referenced by _or that refer to_
a comment. It’s easy to forget this when writing the code, so we prefer
to make an explicit direction required, and allow people to specify
`Direction.ANY` in the case that that is what they want.

(In other words: we want to make the common thing easy, the uncommon
thing possible, and the wrong thing impossible.)

A similar situation holds for filters. By forcing clients to think about
what kinds of edges they want to follow to what kinds of nodes, we
encourage them to write more robust code. As before, if clients do want
to consider all nodes or all edges, they can pass the appropriate empty
address (`nodeAddress([])` or `edgeAddress([]`), which is a prefix of
every address.

Therefore, we require that clients pass an `options` object to
`neighbors`, and we furthermore require that each of the three options
be present.

Paired with @decentralion, in spirit.

Test Plan:
None; this changes the API for a function that has no implementation or
clients.

wchargin-branch: neighbors-options-required

### DIFF
--- a/src/v3/core/graph.js
+++ b/src/v3/core/graph.js
@@ -16,15 +16,20 @@ export type Edge = {|
 export type Neighbor = {|+node: NodeAddress, +edge: Edge|};
 
 export opaque type DirectionT = Symbol;
-export const Direction: {|+IN: DirectionT, +OUT: DirectionT|} = Object.freeze({
+export const Direction: {|
+  +IN: DirectionT,
+  +OUT: DirectionT,
+  +ANY: DirectionT,
+|} = Object.freeze({
   IN: Symbol("IN"),
   OUT: Symbol("OUT"),
+  ANY: Symbol("ANY"),
 });
 
 export type NeighborsOptions = {|
-  +direction: ?DirectionT,
-  +nodePrefix: ?NodeAddress,
-  +edgePrefix: ?EdgeAddress,
+  +direction: DirectionT,
+  +nodePrefix: NodeAddress,
+  +edgePrefix: EdgeAddress,
 |};
 
 export opaque type GraphJSON = any; // TODO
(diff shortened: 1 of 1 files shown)


decentralion:
In the case of direction, this change is a clear win. Setting direction ANY is rarely the right thing to do, so defaulting to it invites bugs.

In the case of making Node and Edge prefix mandatory, this change could seem frustrating. Sometimes it forces redundant specification; as with specifying that the `parent` of a GitHub issue is always a repo. However, in the redundant specification is redundant thanks to an invariant about how node type and edge type interrelate, and this invariant cannot be enforced locally (in code or in time). Therefore, we should not encourage users of the API to depend on unreliable invariants, given that they only afford the ability to save some keystrokes.

(I am sure we will want more sophisticated querying on addresses in the future; much more sophisticated than is possible using address prefixing. The key insight of the v3 API is that the data structure backing those queries does not need to be the graph itself. We expect it will be better to use a relational database which use the addresses as string ids. Then, we are able to use dedicated tools for managing database complexity rather than being forced to maintain our homegrown, ever more cumbersome system. And our graph can just be really good at being a graph. :slightly_smiling_face: )

wchargin:
> However, in the redundant specification is redundant thanks to an
> invariant about how node type and edge type interrelate, and this
> invariant cannot be enforced locally (in code or in time).

I’m not sure exactly how to parse this, but I’m going to assume that it
means “the specification _seems_ redundant, and in fact that may
currently be the case, but new node or edge types added later could
invalidate the relevant invariant, so we are encouraging users to create
robust code”.

If that is the case, then I agree.

> I am sure we will want more sophisticated querying on addresses in the
> future; much more sophisticated than is possible using address
> prefixing. The key insight of the v3 API is that the data structure
> backing those queries does not need to be the graph itself.

Agreed on both counts.

PR #355, opened 2018-06-07
Add foundations to unify address implementations

wchargin:
Summary:
We have `NodeAddress` and `EdgeAddress`, which are opaque aliases of
`string` each with separate associated functions. We really want to keep
this setup: having the address types be structurally distinct is very
nice. But currently the implementation is extremely repetitive. Core
functionality is implemented twice, once for nodes and once for edges.
The test code is even worse, as it is forced to include ugly, hacky,
parametric generalizations to test both implementations—which are really
the same code, anyway!

In this commit, we introduce a system to unify the _implementations_
while keeping the _APIs_ separate. That is, users still see separate
opaque types `NodeAddressT` and `EdgeAddressT`. Users now also see
separate modules `NodeAddress` and `EdgeAddress`, each of which
implements the same interface for its appropriate type. These modules
are each implemented by the same address module factory.

To get this to work, we clearly need to parameterize the module type
over the address type. The problem is getting this to work in a way that
interacts nicely with the opaque types. The trick is to let the factory
always return a transparent module at type `string`, but to then
specialize the type of the resulting module in the module in which the
underlying type of the opaque type is known.

This commit includes specifications for all functions that are in the
current version of the API, but includes only as much implementation
code as is needed to convince me that tests and Flow are actually
running (i.e., very little). I’ll send implementations in separate PRs
for easier review.

The preliminary modules created in this commit _are_ exported from the
graph module, even though they are incomplete. This is so that we can be
sure that nothing will catch fire in Flow-land when we try to export
them (which is plausible, given that we have nontrivial interactions
between opaque types and parametric polymorphism).

Test Plan:
Unit tests included. Run `yarn travis`.

wchargin-branch: address-unified-foundations

### DIFF
--- a/src/v3/core/address.js
+++ b/src/v3/core/address.js
@@ -1,0 +1,142 @@
+// @flow
+
+export interface AddressModule<Address> {
+  /**
+   * Assert at runtime that the provided address is actually a valid
+   * address of this kind, throwing an error if it is not. If `what` is
+   * provided, it will be included in the error message.
+   */
+  assertValid(address: Address, what?: string): void;
+
+  /**
+   * Assert at runtime that the provided array is a valid array of
+   * address parts (i.e., a valid input to `fromParts`), throwing an
+   * error if it is not. If `what` is provided, it will be included in
+   * the error message.
+   */
+  assertValidParts(parts: $ReadOnlyArray<string>, what?: string): void;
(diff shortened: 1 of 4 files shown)


wchargin:
Good idea; will do.

PR #356, opened 2018-06-07
Add top-level address module error handling

wchargin:
Summary:
This commit implements all the code needed for the top-level
`makeAddressModule` function, without implementing any of the address
functions. This mostly comprises handling of errors in the module
options.

Test Plan:
Unit tests added. Run `yarn travis`.

wchargin-branch: address-error-handling

### DIFF
--- a/src/v3/core/address.js
+++ b/src/v3/core/address.js
@@ -1,5 +1,7 @@
 // @flow
 
+import stringify from "json-stable-stringify";
+
 export interface AddressModule<Address> {
   /**
    * Assert at runtime that the provided address is actually a valid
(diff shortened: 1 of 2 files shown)


wchargin:
I’m not sure that I understand. I can inline `separator` if you want,
but as it stands it _is_ factored out to a constant. The only place in
which the sequence `\0` appears in `address.js`, through this whole
sequence of PRs, is in `const separator = "\0"`. The constant is roughly
at the top of the module. Maybe you’re suggesting that it be called
`SEPARATOR` instead and pulled to the top of the JavaScript module, but
I don’t see a reason to do that.

I’ll merge this as is; if I’m misunderstanding something, we can fix it.

PR #357, opened 2018-06-07
Add unified address assertions, fromParts, toParts

wchargin:
Summary:
This implements the following functions for the unified addresses:

  - assertions: `assertValid`, `assertValidParts`
  - injection: `fromParts`
  - projection: `toParts`

(These functions depend on each other for testing, so we implement them
together.)

Test Plan:
Unit tests included. Run `yarn travis`.

wchargin-branch: address-assertion-injection-projection

### DIFF
--- a/src/v3/core/address.test.js
+++ b/src/v3/core/address.test.js
@@ -58,5 +58,163 @@ describe("core/address", () => {
         FooAddress.assertValid = FooAddress.assertValid;
       }).toThrow(/read.only property/);
     });
+
+    it("has a stable internal representation", () => {
+      const input = ["", "hello", "", "", "sweet", "world", "", "", ""];
+      const address = makeModules().FooAddress.fromParts(input);
+      // We stringify the address here because otherwise literal NUL
+      // characters will appear in the snapshot file.
+      expect(JSON.stringify(address)).toMatchSnapshot();
+    });
+
+    describe("assertValid", () => {
+      const {FooAddress, BarAddress, WatAddress} = makeModules();
+      it("rejects `undefined`", () => {
+        expect(() => {
+          // $ExpectFlowError
+          FooAddress.assertValid(undefined, "widget");
+        }).toThrow("widget: expected FooAddress, got: undefined");
+      });
+      it("rejects `null`", () => {
+        expect(() => {
+          // $ExpectFlowError
+          FooAddress.assertValid(null, "widget");
+        }).toThrow("widget: expected FooAddress, got: null");
+      });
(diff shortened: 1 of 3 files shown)


src/v3/core/address.test.js
```diff
@@ -58,5 +58,142 @@ describe("core/address", () => {
...
+        expect(() => {
+          FooAddress.assertValidParts(["hello", "n\0o", "world"], "widget");
+        }).toThrow(
+          "widget: part contains NUL character: " +
+            '"n\\u0000o" in ["hello","n\\u0000o","world"]'
+        );
+      });
+    });
+
+    describe("fromParts", () => {
```
decentralion: I note that a `fromParts` implementation that neglected to call `assertValidParts` would pass tests. please fix

src/v3/core/address.test.js
```diff
@@ -58,5 +58,142 @@ describe("core/address", () => {
...
+        const foo = FooAddress.fromParts(["hello", "world"]);
+        const bar = BarAddress.fromParts(["hello", "world"]);
+        expect(foo).not.toEqual(bar);
+      });
+    });
+
+    describe("toParts", () => {
+      const {FooAddress, BarAddress} = makeModules();
+
+      // We use this next test as a proxy for fully correct validation,
```
decentralion: Is explicit mocking feasible?

src/v3/core/address.test.js
wchargin: Good point; will fix.

src/v3/core/address.test.js
wchargin: Did you want to mock `assertValid` and `assertValidParts` to check that
they’re called? That’s reasonable, but definitely increases the
fragility of the tests. For instance, suppose that the implementation
were changed to

```js
function _assertValid(…) { … }
function assertValid(...args) {
  if (NODE_ENV !== "production") {
    return assertValid(...args);
  }
  // or, do some other preprocessing, like pre-compute a default value
  // of `what`, or catch and rethrow errors adding some information to
  // the message, or…
}
function toParts(address) {
  _assertValid(address);
  // …
}
```

Then our mock that `assertValid` is called would fail, because only the
internal function `_assertValid` is called.

I definitely prefer not to mock when we don’t have to, as it increases
complexity and means that the code that actually runs in production is
not the code that you’re testing. Sometimes it’s obviously the right
thing to do, but I don’t think that that’s the case here.

I’ll merge as-is; if I misunderstood or you think that mocking here is a
really great idea, then I’m happy to change this.

PR #361, opened 2018-06-07
Use new unified addresses in Graph V3

wchargin:
Summary:
This is basically a textual substitution.

Test Plan:
Existing unit tests suffice. Note that `_address.js` (with underscore)
is no longer imported except from its own tests.

wchargin-branch: unified-addresses-in-graph

### DIFF
--- a/src/v3/core/graph.test.js
+++ b/src/v3/core/graph.test.js
@@ -1,48 +1,21 @@
 // @flow
 
 import {
-  type EdgeAddress,
-  type NodeAddress,
-  type _EdgeAddressT,
-  type _NodeAddressT,
-  Address,
+  type EdgeAddressT,
+  type NodeAddressT,
   Direction,
+  EdgeAddress,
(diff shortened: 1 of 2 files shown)


src/v3/core/graph.test.js
```diff
@@ -1,35 +1,14 @@
...
+  edgeToString,
+} from "./graph";
 
 describe("core/graph", () => {
-  const {nodeAddress, edgeAddress} = Address;
-  describe("Address re-exports", () => {
-    it("exist", () => {
-      expect(Address).toEqual(expect.anything());
-    });
-    it("include distinct NodeAddress and EdgeAddress types", () => {
```
decentralion: (Also referenced on the first PR) - I like this test and think it should continue in some form.

wchargin: Reinstated in `graph.test.js` in #355.

PR #368, opened 2018-06-08
Implement `Graph.neighbors`

decentralion:
This is the change that puts the Graph into `Graph` :) We add `_inEdges`
and `_outEdges`, and use them to identify the neighbors of a given
`node`.

The API is implemented pretty uncontroversially. (We've done this a few
times before: see #319, #162). As with other iterators, we check for
comodification and error if this has occurred.

The tests cover some interesting cases like absent nodes, loops, and
multiple edges with the same src and dst.

Test plan:
Unit tests have been added. Run `yarn travis`.

Paired with: @wchargin

### DIFF
--- a/src/v3/core/graph.js
+++ b/src/v3/core/graph.js
@@ -52,6 +52,11 @@ export class Graph {
   //
   // Invariant: If an edge `e` is in the graph, then `e.src` and `e.dst`
   // are both in the graph.
+  //
+  // Invariant: For an edge `e`, the following are equivalent:
+  //   - `e` is in the graph;
+  //   - `_inEdges.get(e.dst)` contains `e` exactly once;
+  //   - `_outEdges.get(e.src)` contains `e` exactly once.
   _nodes: Set<NodeAddressT>;
   _edges: Map<EdgeAddressT, Edge>;
   _inEdges: Map<NodeAddressT, Edge[]>;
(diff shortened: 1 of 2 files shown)


src/v3/core/graph.js
```diff
@@ -97,21 +102,31 @@ export class Graph {
 
   addNode(a: NodeAddressT): this {
     NodeAddress.assertValid(a);
-    this._nodes.add(a);
     this._markModification();
```
wchargin: Slight preference for this going right before the `return`.

src/v3/core/graph.js
```diff
@@ -97,21 +102,31 @@ export class Graph {
...
+    const existingInEdges = this._inEdges.get(a) || [];
+    const existingOutEdges = this._outEdges.get(a) || [];
+    const existingEdges = existingInEdges.concat(existingOutEdges);
+    if (existingEdges.length > 0) {
+      const strAddress = NodeAddress.toString(a);
+      const strExampleEdge = edgeToString(existingEdges[0]);
+      throw new Error(
+        `Attempted to remove ${strAddress}, which is incident to ${
+          existingEdges.length
+        } edges, eg: ${strExampleEdge}`
```
wchargin: nit: can we write `edge(s)` and `e.g.,`, please?

PR #369, opened 2018-06-08
Add and implement `Graph.equals`

decentralion:
It turns out we forgot to add this to the API, so I added it. I also
implemented it. The tests are pretty thorough; as an added innovation
over our previous tests (e.g. in #312 and #61), we now consistently test
that equality is commutative.

Paired with @wchargin

Test plan:
Unit tests were added. Run `yarn travis`

### DIFF
--- a/src/v3/core/graph.test.js
+++ b/src/v3/core/graph.test.js
@@ -703,6 +703,100 @@ describe("core/graph", () => {
         });
       });
     });
+
+    describe("equals", () => {
+      const src = NodeAddress.fromParts(["src"]);
+      const dst = NodeAddress.fromParts(["dst"]);
+      const edge1 = () => ({
+        src,
+        dst,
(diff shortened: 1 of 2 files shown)


src/v3/core/graph.test.js
```diff
@@ -703,6 +703,100 @@ describe("core/graph", () => {
...
+      });
+      function expectEquality(g1, g2, isEqual: boolean) {
+        expect(g1.equals(g2)).toBe(isEqual);
+        expect(g2.equals(g1)).toBe(isEqual);
+      }
+
+      it("empty graph equals itself", () => {
+        expectEquality(new Graph(), new Graph(), true);
+      });
+      it("empty graph doesnt equal nonempty graph", () => {
```
wchargin: sp.: `doesn't`

PR #374, opened 2018-06-09
Implement `Graph.toJSON` and `Graph.fromJSON`

decentralion:
The serialization scheme uses `IndexedEdge`s:

```js
type Integer = number;
type IndexedEdge = {|
  Address: EdgeAddressT,
  srcIndex: Integer,
  dstIndex: Integer,
|}
```

The nodes are first sorted. Then, we generate indexed edges from the
regular edges by replacing each node address with its index in the
sorted order. This encoding reduces the number of addresses serialized
from `n + 3e` to `n + e` (where `n` is the number of nodes and `e` is
the number of edges).

This is based on work in #295, but in contrast to that PR, we do not
index the in-memory representations of graphs. Only the JSON
representation is indexed.

Test plan:
Unit tests added. A snapshot test is also included, both to make it easy
to inspect an example of a JSON-serialized graph, and to ensure
backwards-compatibility. (The snapshot likely should not change
independent of the VERSION string.)

### DIFF
--- a/src/v3/core/graph.test.js
+++ b/src/v3/core/graph.test.js
@@ -947,6 +947,96 @@ describe("core/graph", () => {
     });
   });
 
+  describe("toJSON / fromJSON", () => {
+    const src = NodeAddress.fromParts(["src"]);
+    const dst = NodeAddress.fromParts(["dst"]);
+    const edge1 = () => ({
+      src,
+      dst,
+      address: EdgeAddress.fromParts(["edge"]),
+    });
+    const edge2 = () => ({
+      src: dst,
+      dst: src,
+      address: EdgeAddress.fromParts(["edge", "2"]),
+    });
+    it("toJSON matches snapshot", () => {
+      const graph = new Graph()
+        .addNode(src)
+        .addNode(dst)
+        .addEdge(edge1())
+        .addEdge(edge2());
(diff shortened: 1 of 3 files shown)


wchargin:
I’ve only glanced very briefly at the code, but: do we really want to
encode addresses on disk with literal NUL characters? This has the
effect, for instance, that snapshots (and graphs in general) are treated
as binary files and are not very human-editable or human-readable.

decentralion:
Good point. I've changed the representation to  store the array-ified addresses. PTAL

src/v3/core/graph.test.js
```diff
@@ -847,6 +847,96 @@ describe("core/graph", () => {
...
+      dst: src,
+      address: EdgeAddress.fromParts(["edge", "2"]),
+    });
+    it("toJSON matches snapshot", () => {
+      const graph = new Graph()
+        .addNode(src)
+        .addNode(dst)
+        .addEdge(edge1())
+        .addEdge(edge2());
+      expect(graph.toJSON()).toMatchSnapshot();
```
wchargin: Hmm. I appreciate that we have a snapshot test, but it would be nice if
that snapshot test could demonstrate some edge cases: multiple edges
with the same hom, loops, multiple loops, isolated nodes? It looks like
these cases aren’t tested at all, in fact.

PR #383, opened 2018-06-13
Rename various aspects of GitHub `nodes` module

wchargin:
Summary:
First, we rename the module itself from `address` to `nodes`: we’d like
to put the edge functions in a parallel `edges` module instead of
cramping it into this one, so it stands to reason that this one should
be called `nodes`.

We also rename the `GithubAddressT` type to `RawAddress`, so that the
module exports `RawAddress` and `StructuredAddress`. The functions then
have much better natural names of `toRaw` and `fromRaw`.

Test Plan:
Existing unit tests suffice.

wchargin-branch: rename-nodes


wchargin:
Coverage of the resulting module remains at 100%.

PR #385, opened 2018-06-13
Add GitHub `edges` module

wchargin:
Summary:
This module includes a raw edge type, a structured edge type, and edge
creation functions that take source and destination and create an edge.

Paired with @decentralion (for the important bits, anyway).

Test Plan:
Unit tests added. These cover all of the successful cases, and none of
the unsuccessful cases. We plan to refactor this code Soon™, and it is
hard to see how to nicely factor the tests without just testing the same
code paths over and over.

wchargin-branch: github-edges

### DIFF
--- a/src/v3/plugins/github/edges.js
+++ b/src/v3/plugins/github/edges.js
@@ -1,0 +1,208 @@
+// @flow
+
+import {
+  type Edge,
+  type EdgeAddressT,
+  type NodeAddressT,
+  EdgeAddress,
+  NodeAddress,
+} from "../../core/graph";
+import * as GithubNode from "./nodes";
+
+export opaque type RawAddress: EdgeAddressT = EdgeAddressT;
+
+export type AuthorsAddress = {|
+  +type: "AUTHORS",
+  +author: GithubNode.UserlikeAddress,
+  +content: GithubNode.AuthorableAddress,
+|};
+export type MergedAsAddress = {|
+  +type: "MERGED_AS",
+  +pull: GithubNode.PullAddress,
+|};
+export type HasParentAddress = {|
+  +type: "HAS_PARENT",
+  +child: GithubNode.ChildAddress,
+|};
+export type ReferencesAddress = {|
+  +type: "REFERENCES",
+  +referrer: GithubNode.TextContentAddress,
+  +referent: GithubNode.ReferentAddress,
+|};
+
+export type StructuredAddress =
+  | AuthorsAddress
+  | MergedAsAddress
+  | HasParentAddress
+  | ReferencesAddress;
+
+export const createEdge = Object.freeze({
+  authors: (
+    author: GithubNode.UserlikeAddress,
+    content: GithubNode.AuthorableAddress
+  ): Edge => ({
+    address: toRaw({type: "AUTHORS", author, content}),
+    src: GithubNode.toRaw(author),
+    dst: GithubNode.toRaw(content),
+  }),
+  mergedAs: (
+    pull: GithubNode.PullAddress,
+    commitAddress: NodeAddressT /* TODO: Make this a Git commit node address. */
+  ): Edge => ({
+    address: toRaw({type: "MERGED_AS", pull}),
+    src: GithubNode.toRaw(pull),
+    dst: commitAddress,
+  }),
+  hasParent: (
+    child: GithubNode.ChildAddress,
+    parent: GithubNode.ParentAddress
+  ): Edge => ({
+    address: toRaw({type: "HAS_PARENT", child}),
+    src: GithubNode.toRaw(child),
+    dst: GithubNode.toRaw(parent),
+  }),
+  references: (
+    referrer: GithubNode.TextContentAddress,
+    referent: GithubNode.ReferentAddress
+  ): Edge => ({
+    address: toRaw({type: "REFERENCES", referrer, referent}),
+    src: GithubNode.toRaw(referrer),
+    dst: GithubNode.toRaw(referent),
+  }),
+});
+
+const NODE_PREFIX_LENGTH = NodeAddress.toParts(GithubNode._githubAddress())
+  .length;
+
+const GITHUB_PREFIX = EdgeAddress.fromParts(["sourcecred", "github"]);
+function githubEdgeAddress(...parts: string[]): RawAddress {
+  return EdgeAddress.append(GITHUB_PREFIX, ...parts);
+}
+function lengthEncode(x: GithubNode.RawAddress): $ReadOnlyArray<string> {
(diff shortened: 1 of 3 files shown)


src/v3/plugins/github/edges.js
```diff
@@ -0,0 +1,216 @@
...
+import * as GithubNode from "./nodes";
+
+export opaque type RawAddress: EdgeAddressT = EdgeAddressT;
+
+// This particular union of types is used frequently. It seems
+// reasonable that it semantically represents some kind of "post" that
+// has a parent, was authored, and can be referred to. But if one of
+// those axes of symmetry breaks down, then we shouldn't feel hesitant
+// to inline this type in the appropriate places.
+export type PostAddress =
```
decentralion: This should be exported from nodes, not edges.

src/v3/plugins/github/edges.js
```diff
@@ -0,0 +1,216 @@
...
+    pull: GithubNode.PullAddress,
+    commitAddress: NodeAddressT /* TODO: Make this a Git commit node address. */
+  ): Edge => ({
+    address: toRaw({type: "MERGED_AS", pull}),
+    src: GithubNode.toRaw(pull),
+    dst: commitAddress,
+  }),
+  hasParent: (
+    child: PostAddress,
+    parentAddress: GithubNode.RawAddress
```
decentralion: Why does this take a RawAddress?
Let's define a ParentAddress union alongside PostAddress, and it is a union of RepoAddress | IssueAddress | PullAddress | ReviewAddress.

I believe this will be useful because I already defined that union in my createGraph branch because I also needed it (in addition to using the PostAddress union).

In any case, breaking consistency and requiring a RawAddress for this arg only seems clearly wrong.

src/v3/plugins/github/edges.js
```diff
@@ -0,0 +1,216 @@
...
+  }),
+  references: (referrer: PostAddress, referent: PostAddress): Edge => ({
+    address: toRaw({type: "REFERENCES", referrer, referent}),
+    src: GithubNode.toRaw(referrer),
+    dst: GithubNode.toRaw(referent),
+  }),
+});
+
+const NODE_PREFIX_LENGTH = NodeAddress.toParts(GithubNode._githubAddress())
+  .length;
```
decentralion: le wrap :/

src/v3/plugins/github/edges.js
```diff
@@ -0,0 +1,216 @@
...
+
+const GITHUB_PREFIX = EdgeAddress.fromParts(["sourcecred", "github"]);
```
decentralion: i feel like the parts array should be factored out somewhere, and shared between the node and edge. it would be weird if one could change without the other. doesn't necessarily need to happen in this pr.

src/v3/plugins/github/edges.js
wchargin: Well… as noted in the comment, it’s really not clear what, if anything,
this union type represents. It’s quite plausible that it is a false
abstraction born out of coincidence.

What do you think about splitting this up into a larger set of unions,
each of which actually does have reasonable semantics? It seems that we
might want `ChildAddress`, `ParentAddress`, `AuthorableAddress`,
`ReferentAddress`, and `TextContentAddress`. Currently, four of those
types happen to be a union of the same four members (comment, issue,
pull request, review). If there is evidence that some of these
relationships are fundamental—e.g., “everything that is authorable has
text content” (dubious)—then we can unify the appropriate unions, but if
there isn’t then we probably shouldn’t just say “eh, they’re all some
kind of ‘post’ ” and leave it at that.

With these types, we would have:

```js
export type AuthorsAddress = {|
  +type: "AUTHORS",
  +author: GithubNode.UserlikeAddress,
  +content: GithubNode.AuthorableAddress,
|};
export type MergedAsAddress = {|
  +type: "MERGED_AS",
  +pull: GithubNode.PullAddress,
|};
export type HasParentAddress = {|
  +type: "HAS_PARENT",
  +child: GithubNode.ChildAddress,
|};
export type ReferencesAddress = {|
  +type: "REFERENCES",
  +referrer: GithubNode.TextContentAddress,
  +referent: GithubNode.ReferentAddress,
|};
```

If we effect this change, then it does make sense to me to export these
types from `nodes.js` instead of `edges.js`. Otherwise, I’m not so sure.

src/v3/plugins/github/edges.js
wchargin: Okay; it’s fine with me for this to take a structured address. I’m
willing to define and export the `ParentAddress` union, while noting
that it seems pretty weak: while we effectively have `ParentAddress` and
`ChildAddress` types, almost everything is a parent and almost
everything is a child, and we are failing to express (at the type level)
the relationship that (e.g.) a repository cannot be the parent of a
comment. But that seems like a reasonable concession to me.

src/v3/plugins/github/edges.js
wchargin: Sad.

src/v3/plugins/github/edges.js
wchargin: Eh; neutral. It would be weird, but it wouldn’t be wrong (in that it
wouldn’t break anything). I’ll keep things as they are in this PR,
though I am open to factoring this out.

src/v3/plugins/github/edges.js
wchargin: Discussed more: https://discordapp.com/channels/453243919774253079/454007907663740939?jump=456564101183832064

I will wait for #387 to land.

decentralion:
Let's rebase this on 387, and I think it will be an easy approval :)

src/v3/plugins/github/edges.js
```diff
@@ -0,0 +1,208 @@
...
+    pull: GithubNode.PullAddress,
+    commitAddress: NodeAddressT /* TODO: Make this a Git commit node address. */
+  ): Edge => ({
+    address: toRaw({type: "MERGED_AS", pull}),
+    src: GithubNode.toRaw(pull),
+    dst: commitAddress,
+  }),
+  hasParent: (
+    child: GithubNode.ChildAddress,
+    parentAddress: GithubNode.RawAddress
```
decentralion: Still sad that this is a RawAddress

wchargin:
> Please fix the rawAddress and change to ParentAddress

Thanks for the reminder; I had just forgotten.

> Thanks for writing this, I know it wasn't fun.

Indeed. Hopefully, we can move on to doing interesting things, as
opposed to hacking together awful data formats all because the only
two value types in JavaScript are “IEEE 754 double-precision
floating-point number” and “array of UTF-16 code units”.

PR #98, opened 2018-03-20
Implement GitHub reference detection

dandelionmane:
This commit only adds logic for finding references in GitHub posts,
either by #-numeric reference, or explicit urls. Adding the reference
edges to the graph will occur in a followon commit. Progress on #40 

Test plan: New unit tests are included

PR #388, opened 2018-06-13
Add methods for parsing Ids from GitHub urls

decentralion:
Test plan:
Unit tests added. Run `yarn travis`.


src/v3/plugins/github/urlIdParse.test.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+    it("for reviewComment", () => {
+      expect(reviewCommentUrlToId(reviewComment)).toEqual("194816899");
+    });
+    it("for review", () => {
+      expect(reviewUrlToId(review)).toEqual("128199239");
+    });
+  });
+  describe("errors on", () => {
+    const bad = [
```
wchargin: Nice.

src/v3/plugins/github/urlIdParse.test.js
```diff
@@ -0,0 +1,47 @@
...
+      {f: commentUrlToId, instances: [reviewComment, review]},
+      {f: reviewCommentUrlToId, instances: [pullComment, issueComment, review]},
+      {f: reviewUrlToId, instances: [pullComment, issueComment, reviewComment]},
+    ];
+    bad.forEach(({f, instances}) => {
+      it(`${f.name} receiving the wrong url type`, () => {
+        instances.forEach((i) => {
```
wchargin: It would be nice if this could indicate which example failed (e.g., by
using separate `it` blocks). If that’s not too hard to implement, I’d
prefer it.

src/v3/plugins/github/urlIdParse.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
+export function commentUrlToId(url: string): string {
```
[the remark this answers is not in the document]
decentralion: I decided to just explicitly create 3 separate functions for issues, pulls, and reviews.
It also simplifies the regex - issue comments and pull comments do actually have separate urls.

src/v3/plugins/github/urlIdParse.test.js
decentralion: Done. I also added checks on null, undefined, and a garbage string.

decentralion:
Here's what the test structure now looks like:
```
  plugins/github/urlToId
    works correctly
      ✓ for issueComment (3ms)
      ✓ for pullComment
      ✓ for reviewComment (1ms)
      ✓ for review
    errors on
      issueCommentUrlToId
        ✓ recieving pull comment (1ms)
        ✓ recieving review comment (1ms)
        ✓ recieving review
        ✓ recieving null
        ✓ recieving undefined
        ✓ recieving nonsense (1ms)
      pullCommentUrlToId
        ✓ recieving issue comment
        ✓ recieving review comment
        ✓ recieving review
        ✓ recieving null (1ms)
        ✓ recieving undefined
        ✓ recieving nonsense
      reviewCommentUrlToId
        ✓ recieving pull comment
        ✓ recieving issue comment (1ms)
        ✓ recieving review
        ✓ recieving null
        ✓ recieving undefined
        ✓ recieving nonsense (1ms)
      reviewUrlToId
        ✓ recieving pull comment
        ✓ recieving issue comment
        ✓ recieving review comment
        ✓ recieving null
        ✓ recieving undefined (1ms)
        ✓ recieving nonsense
```

src/v3/plugins/github/urlIdParse.js
```diff
@@ -0,0 +1,41 @@
+// @flow
+
+export function reviewUrlToId(url: string): string {
+  const result = new RegExp(
+    "^https://github.com/\\w+/\\w+/pull/\\d+#pullrequestreview-(\\d+)$"
```
wchargin: The `\w` character class does not include `-`, which is a valid
character in GitHub repositories (e.g., sourcecred/example-github).
Consider instead `[a-zA-Z0-9_-]` throughout this module, and consider
adding an `example-github` test.

src/v3/plugins/github/urlIdParse.js
```diff
@@ -0,0 +1,41 @@
...
+  ).exec(url);
+  if (result == null) {
+    throw new Error(`Error parsing review url ${url}`);
```
wchargin: I’m surprised that exceptions are the right semantics. Have you
considered returning `null` in the case where there is no match? Having
a URL that does not refer to a comment is not really an exceptional
condition.

decentralion: My intention is to use this function when generating GitHub addresses from the GitHub graphql response. In this case, you have a JSON that you know definitely corresponds to (say) a review comment, so you call `reviewCommentUrlToId` on that json's url. This api is appropriate for that case - if anything goes wrong in getting the id out of the review comment's url, it's cause for an error. So the contract  for the function is that it should take in review comment urls and nothing else.

If we start to need the function for other things we can upgrade its scope, but if we start using it e.g. to sniff arbitrary urls and figure out what they are, at that point we need a fairly different return signature anyway.

wchargin: Understood. I had the `findReferences` context in mind, but you’re right
that the signature (and implementation) would need to be different
anyway.

src/v3/plugins/github/urlIdParse.js
decentralion: Great catch, thanks. I re-wrote the regex generation to be cleaner, and to handle this case. I used #98 as inspiration. Also, I changed the test case to include underscores, numbers, and dashes in the owner and repo name.

src/v3/plugins/github/urlIdParse.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+
+// Given a sequence of regex parts that appear after the repo name and owner
+// in a GitHub url, create a regex to match them.
+function githubRegex(...regexParts) {
```
wchargin: Cool; this seems like a useful function.

The implementation, however, is not correct. The problem is that the
`RegExp` constructor is not homomorphic with respect to concatenation.
That is, there is a well-defined operation of regular expression
concatenation, which is what you’re trying to express, but this
operation is not equivalent to “return the `RegExp` whose source is the
string concatenation of the sources of the two `RegExp`s”. That is: if
we let `∙` denote regular expression concatenation and `++` denote
string concatenation, then it is not the case that

    RegExp (s1 ++ s2) == (RegExp s1) ∙ (RegExp s2),

but this is the assumption that your code is making.

As a concrete example, the result of

```js
// Find an issue or pull request!
const re = githubRegex(/issues|pull/, /(\d+)/);
```

is the expression `/https:…\/issues|pull\/(\d+)$/`, which is not
correct, because it matches a URL ending in `"/issues/foo/bar/baz"`.

Basically, the thing is vulnerable to “regular expression injection”.

We can fix this by surrounding the individual parts with non-capturing
groups, which I believe (but have not proven) to be sufficient. This
also suggests what I think is a significantly cleaner implementation:

```js
const parts = [
  /https:\/\/github.com/,
  githubNamePart,
  githubNamePart,
  ...regexParts,
];
const mainSource = parts.map((re) => `(?:${re.source})`).join("/");
return new RegExp(`^${mainSource}$`);
```

Accordingly, as you may expect, I would like to see this function
tested. :-)

src/v3/plugins/github/urlIdParse.test.js
```diff
@@ -0,0 +1,77 @@
...
+import {
+  issueCommentUrlToId,
+  pullCommentUrlToId,
+  reviewCommentUrlToId,
+  reviewUrlToId,
+} from "./urlIdParse";
+
+describe("plugins/github/urlToId", () => {
+  const issueComment =
+    "https://github.com/example_owner/example-repo0/issues/350#issuecomment-394939349";
```
wchargin: A GitHub username may not actually include an underscore. It’s okay with
me if the RegExp is overly lenient in allowing hyphens (I don’t remember
why we decided to do this back in March), but the test data should
probably be actually valid.

If you want to make the RegExp more strict, that’s okay with me, too.
I am not even opposed to pulling in the `github-username-regex` package
from npm:
https://www.npmjs.com/package/github-username-regex

src/v3/plugins/github/urlIdParse.js
decentralion: Good find. I removed this function as it's incorrect and I don't want to test it, given that I only needed it for 4 specific instances. PTAL.

PR #395, opened 2018-06-14
Add Github prefixes and const types

decentralion:
- Switch string constant node and edge types (e.g. "REPO") to exported
consts (eg `export const REPO_TYPE`).
- Add (and internally use) a `_Prefix` psuedomodule which contains
per-type address prefixes
- Test that constructing a StructuredAddress with the wrong type is an
error.

Test plan:
Unit tests pass, snapshots unchanged.

Paired with @wchargin

### DIFF
--- a/src/v3/plugins/github/nodes.js
+++ b/src/v3/plugins/github/nodes.js
@@ -9,33 +9,53 @@ export function _githubAddress(...parts: string[]): RawAddress {
   return NodeAddress.append(GITHUB_PREFIX, ...parts);
 }
 
+export const REPO_TYPE: "REPO" = "REPO";
+export const ISSUE_TYPE: "ISSUE" = "ISSUE";
+export const PULL_TYPE: "PULL" = "PULL";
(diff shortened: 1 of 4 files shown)


src/v3/plugins/github/nodes.js
```diff
@@ -99,82 +122,82 @@ export function fromRaw(x: RawAddress): StructuredAddress {
...
       if (rest.length !== 4) {
         throw fail();
       }
       const [owner, name, pullNumber, id] = rest;
-      const repo = {type: "REPO", owner, name};
-      const pull = {type: "PULL", repo, number: pullNumber};
-      return {type: "REVIEW", pull, id};
+      const repo = {type: REPO_TYPE, owner, name};
+      const pull = {type: PULL_TYPE, repo, number: pullNumber};
+      return {type: REVIEW_TYPE, pull, id};
```
wchargin: Type assertions on others, but not here. Make consistent?

decentralion: Done

PR #396, opened 2018-06-14
Add `CommentableAddress`

decentralion:
Test plan: Not needed.

Paired with @wchargin


src/v3/plugins/github/nodes.js
```diff
@@ -90,6 +90,9 @@ export type ParentAddress =
   | PullAddress
   | ReviewAddress;
 
+// Each of these types may have Comments as children
+export type CommentableAddress = IssueAddress | PullAddress | ReviewAddress;
```
wchargin: Should we check statically that `CommentableAddress <: ParentAddress`?

(I vote yes!)

PR #403, opened 2018-06-19
Add node type definitions for V3 Git plugin

wchargin:
Summary:
This is modeled after the GitHub node module format, with the obvious
alterations plus a bit more type safety in the implementation of `toRaw`
(namely, we check `type` exhaustively).

Test Plan:

wchargin-branch: git-v3-nodes

### DIFF
--- a/src/v3/plugins/git/nodes.js
+++ b/src/v3/plugins/git/nodes.js
@@ -1,0 +1,121 @@
+// @flow
+
+import {NodeAddress, type NodeAddressT} from "../../core/graph";
+import type {Hash} from "./types";
+
+export opaque type RawAddress: NodeAddressT = NodeAddressT;
+
+const GIT_PREFIX = NodeAddress.fromParts(["sourcecred", "git"]);
+export function _gitAddress(...parts: string[]): RawAddress {
+  return NodeAddress.append(GIT_PREFIX, ...parts);
+}
+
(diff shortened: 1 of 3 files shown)


src/v3/plugins/git/nodes.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export const BLOB_TYPE: "BLOB" = "BLOB";
+export const COMMIT_TYPE: "COMMIT" = "COMMIT";
+export const SUBMODULE_COMMIT_TYPE: "SUBMODULE_COMMIT" = "SUBMODULE_COMMIT";
+export const TREE_TYPE: "TREE" = "TREE";
+export const TREE_ENTRY_TYPE: "TREE_ENTRY" = "TREE_ENTRY";
+
+export const _Prefix = Object.freeze({
+  base: GIT_PREFIX,
```
decentralion: Nit: `GIT_PREFIX` is not technically a RawAddress, and so `_Prefix` is not an object whose every value is a RawAddress. IMO, this is a bug; easily fixed by setting `base: _gitAddress()`.

I'm fine not fixing in this commit, since the same bug exists on the GitHub side, we can fix both plugins at once after this is merged.

wchargin: Hmm, interesting. Agreed on all counts.

PR #405, opened 2018-06-19
create the GitHub graph

decentralion:
This commit:
- adds `github/createGraph.js`
  - which ingests GitHub GraphQL response
  - and creates a GitHub graph
- adds `github/graphView.js`,
  - which takes a Graph
  - and validates that all GitHub specific node and edge invariants hold
    - every github node may be parsed by `github/node/fromRaw`
      - with the right node type
    - every github edge may be parsed by `github/edge/fromRaw`
      - with the right edge type
      - with the right src address prefix
      - with the right dst address prefix
    - every child node has exactly one parent
      - of the right type
  - and provides convenient porcelain methods for
    - finding repos in the graph
    - finding issues of a repo
    - finding pulls of a repo
    - finding reviews of a pull
    - finding comments of a Commentable
    - finding authors of Authorables
    - finding parent of a ChildAddress
- tests `createGraph`
  - via snapshot testing
  - by checking the GraphView invariants hold
- tests `graphView`
  - by checking individual entities in the example-git repository have
  the proper relationships

Test plan:
- Extensive unit and snapshot tests added. `yarn travis` passes.


src/v3/plugins/github/createGraph.js
```diff
@@ -0,0 +1,167 @@
...
+    const id = (function() {
+      switch (parent.type) {
+        case "ISSUE":
+          return issueCommentUrlToId(commentJSON.url);
+        case "PULL":
+          return pullCommentUrlToId(commentJSON.url);
+        case "REVIEW":
+          return reviewCommentUrlToId(commentJSON.url);
+        default:
+          throw new Error(`Unexpected comment parent type: ${parent.type}`);
```
wchargin: This can be `(parent.type: empty)`, right?

src/v3/plugins/github/createGraph.js
```diff
@@ -0,0 +1,167 @@
...
+  addPull(repo: RepoAddress, pullJSON: PullJSON) {
+    const pull: PullAddress = {
+      type: "PULL",
+      repo,
+      number: String(pullJSON.number),
+    };
+    this.addNode(pull);
+    this.addAuthors(pull, pullJSON.author);
+    this.addHasParent(pull, repo);
+    pullJSON.comments.nodes.forEach((comment) =>
```
wchargin: If you change `comment` to `c`, this fits on a line. :-)

(Same with `reviewJSON.comments` below.)

src/v3/plugins/github/createGraph.test.js
```diff
@@ -0,0 +1,36 @@
...
+  const data = cloneDeep(require("./demoData/example-github"));
+  return createGraph(data);
+}
+
+describe("plugins/github/createGraph", () => {
+  it("example graph matches snapshot", () => {
+    expect(exampleGraph()).toMatchSnapshot();
+  });
+
+  it("passes all GithubGraphView invariants", () => {
```
wchargin: nit: just `GraphView` (and `GV` below).

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+      return;
+    }
+    if (process.env.NODE_ENV === "test") {
+      // TODO(perf): If this method becomes really slow, we can disable
+      // it on specific tests wherein we construct large graphs.
+      this.checkInvariants();
+    }
+  }
+
+  checkInvariants() {
```
wchargin: I would like to see a test that this is properly reentrant, perhaps like:

```js
const g = exampleGraph();
const view = new GraphView(g);
// Deliberately corrupt the invariants: add an extra `hasParent` edge or
// something.
try {
  view.graph();
} catch (_) {}
expect(() => {
  view.graph();
}).toThrow("Invariant violation");
```

(such that the test fails without the `finally`).

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+
+    // Check that the edge type <-> prefix mapping is maintained.
+    // Check that each edge's src has an expected prefix
+    // Check that each edge's dst has an expected prefix
+    type EdgeInvariant = {|
+      +prefix: GE.RawAddress,
+      +srcPrefixes: NodeAddressT[],
+      +dstPrefixes: NodeAddressT[],
+    |};
+    const edgeTypeToInvariants: {[type: string]: EdgeInvariant} = {
```
wchargin: I note that the actual contract of `HAS_PARENT` is stronger than what is
checked in this function. For instance, a review comment may not have a
repository as its parent.

What do you think about changing `EdgeInvariant` to

```js
type EdgeInvariant = {|
  +prefix: GE.RawAddress,
  +homs: $ReadOnlyArray<Hom>,
|};
type Hom = {|+src: NodeAddressT, +dst: NodeAddressT|};
```

with the semantics that any edge must match at least one of the
specified homs? Then, you could write

```js
const edgeTypeToInvariants: {[type: string]: EdgeInvariant} = {
  [GE.HAS_PARENT_TYPE]: {
    prefix: GE._Prefix.hasParent,
    homs: [
      {src: GN._Prefix.issue, dst: GN._Prefix.repo},
      {src: GN._Prefix.pull, dst: GN._Prefix.repo},
      {src: GN._Prefix.review, dst: GN._Prefix.pull},
      {src: GN._Prefix.issueComment, dst: GN._Prefix.issue},
      {src: GN._Prefix.pullComment, dst: GN._Prefix.pull},
      {src: GN._Prefix.reviewComment, dst: GN._Prefix.review},
    ],
  },
  ...rest,
};
```

for a much more expressive test. In the case where you really do want
the Cartesian product (e.g., `REFERENCES` edges), you can define

```js
declare function homProduct(
  srcs: $ReadOnlyArray<NodeAddressT>,
  dsts: $ReadOnlyArray<NodeAddressT>
): Hom[];
```

with the obvious implementation, enabling

```js
const edgeTypeToInvariants: {[type: string]: EdgeInvariant} = {
  [GE.REFERENCES_TYPE]: {
    prefix: GE._Prefix.references,
    homs: homProduct(
      [
        GN._Prefix.issue,
        GN._Prefix.pull,
        GN._Prefix.review,
        GN._Prefix.comment,
      ],
      [
        GN._Prefix.repo,
        GN._Prefix.issue,
        GN._Prefix.pull,
        GN._Prefix.review,
        GN._Prefix.comment,
        GN._Prefix.userlike,
      ]
    ),
  },
  ...rest,
};
```

for the same expressive power as the current code and just one more line
of code.

This also presents, I posit, a much more useful and reusable
abstraction.

wchargin: If you’re inclined, you should be able to specify the type more
precisely using `$ElementType<StructuredAddress, "type">` instead of
`string`. No strong preference.

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+        );
+      }
+      if (!dstPrefixes.some((p) => NodeAddress.hasPrefix(edge.dst, p))) {
+        throw new Error(
+          `Invariant: Edge ${stringify(structuredEdge)} has bad dst prefix`
+        );
+      }
+    }
+
+    // Check that the special constraints of the has_parent edge are maintained:
```
wchargin: If `Hom`s is implemented as suggested above, then this becomes a simple
uniqueness check.

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+    }
+    this._maybeCheckInvariants();
+  }
+
+  *_neighbors<T: GN.StructuredAddress>(
+    node: GN.StructuredAddress,
+    options: NeighborsOptions
+  ): Iterator<T> {
+    if (!NodeAddress.hasPrefix(options.nodePrefix, GN._Prefix.base)) {
+      throw new Error(`_neighbors must filter to Github nodes`);
```
wchargin: nit: `GitHub` (caps.)

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+
+  comments(commentable: GN.CommentableAddress): Iterator<GN.CommentAddress> {
+    return this._children(commentable, GN._Prefix.comment);
+  }
+
+  reviews(pull: GN.PullAddress): Iterator<GN.ReviewAddress> {
+    return this._children(pull, GN._Prefix.review);
+  }
+
+  // TODO figure out how to overload this fn signature
```
wchargin: Could you `TODO(@someone)` here, please?

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+  parent(child: GN.ChildAddress): GN.ParentAddress {
+    const options = {
+      direction: Direction.OUT,
+      edgePrefix: GE._Prefix.hasParent,
+      nodePrefix: GN._Prefix.base,
+    };
+    const parents: GN.ParentAddress[] = Array.from(
+      this._neighbors(child, options)
+    );
+    if (parents.length !== 1) {
```
wchargin: What do you think about omitting this check? (This is a question, not a
suggestion.)

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,274 @@
...
+  checkInvariants() {
+    this._isCheckingInvariants = true;
+    try {
+      this._checkInvariants();
+    } finally {
+      this._isCheckingInvariants = false;
+    }
+  }
+
+  _checkInvariants() {
```
wchargin: With great predictive power comes great responsibility to ensure that
your tests are actually testing anything at all. That is: I would
strongly request that we verify that each invariant failure actually
triggers the appropriate error. The `_maybeCheckInvariants` method of
the `Graph` class, for comparison, has full test coverage on all
failure cases.

src/v3/plugins/github/createGraph.js
```diff
@@ -0,0 +1,167 @@
+// @flow
+
+import {Graph, NodeAddress} from "../../core/graph";
+import type {
```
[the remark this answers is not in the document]
decentralion: Ack, didn't feel like re-writing it.

src/v3/plugins/github/createGraph.js
decentralion: Yep, added.

src/v3/plugins/github/createGraph.js
decentralion: Done.

src/v3/plugins/github/graphView.js
decentralion: Done.

src/v3/plugins/github/graphView.js
decentralion: Done. It is a nicer abstraction indeed.

decentralion: Ack. No action taken.

src/v3/plugins/github/graphView.js
decentralion: Fixed.

src/v3/plugins/github/graphView.js
decentralion: Yep, rewritten.

src/v3/plugins/github/graphView.js
decentralion: You get a TODO! :)

src/v3/plugins/github/graphView.js
decentralion: Well, it encodes a true invariant. I don't see any reason to omit it.

src/v3/plugins/github/graphView.js
decentralion: Done.

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,298 @@
...
+      const invariant = nodeTypeToInvariant[type];
+      if (invariant == null) {
+        throw new Error(`Invariant: Unexpected node type ${type}`);
+      }
+      const {hasParent, parentAccessor} = invariant;
+      if (hasParent) {
+        // Will throw invariant error if there is not exactly 1 parent
+        const parent = this.parent((structuredNode: any));
+        if (parentAccessor == null) {
+          throw new Error(`Impossible; make flow happy`);
```
wchargin: Minor complaint (but not, I think, a nit): this is impossible in a way
that could be changed in the future such that this error is thrown at
runtime\*. I suggest either of the following alternatives:

  - Make the message clearer about what has gone wrong: this is an error
    in the invariant-checking code itself. For instance, something like
    `bad invariant for ${type}: missing accessor` would make me happier.

  - Better yet, you can encode this check statically. The problem is
    that your invariants are encoded as

    ```js
    type Invariant = {|
      +hasParent: boolean,
      +parentAccessor: ?((*) => GN.ParentAddress),
    |};
    ```

    where there is a (meta-)invariant between the value of the first
    field and the nullity of the second that is not enforced by the
    type. Instead, you could use either a union of more specific object
    types:

    ```js
    type Invariant =
      | {|+hasParent: true, +parentAccessor: (*) => GN.ParentAddress|}
      | {|+hasParent: false, +parentAccessor: null|};
    ```

    or simply functions:

    ```js
    type Invariant = ?((*) => GN.ParentAddress);
    ```

    I’ve verified that each of these typechecks without the
    `null`-check, and also that the first option statically errors if
    the meta-invariant is violated (in the second option, it’s not
    possible to write something that violates this meta-invariant).

\* What I mean is that the impossibility is [admissible but not
derivable][1], which I think is an important didstinction.

[1]: https://en.wikipedia.org/wiki/Rule_of_inference#Admissibility_and_derivability

src/v3/plugins/github/createGraph.js
```diff
@@ -0,0 +1,165 @@
...
+      number: String(pullJSON.number),
+    };
+    this.addNode(pull);
+    this.addAuthors(pull, pullJSON.author);
+    this.addHasParent(pull, repo);
+    pullJSON.comments.nodes.forEach((c) => this.addComment(pull, c));
+    pullJSON.reviews.nodes.forEach((review) => this.addReview(pull, review));
+    if (pullJSON.mergeCommit != null) {
+      const commitHash = pullJSON.mergeCommit.oid;
+      // TODO: Replace with an API from the v3 Git plugin
```
wchargin: This API now exists; you can use it.

src/v3/plugins/github/graphView.js
```diff
@@ -0,0 +1,298 @@
...
+          {srcPrefix: GN._Prefix.issueComment, dstPrefix: GN._Prefix.issue},
+          {srcPrefix: GN._Prefix.pullComment, dstPrefix: GN._Prefix.pull},
+        ],
+        srcAccessor: (x) => GN.toRaw((x: any).child),
+      },
+      [GE.MERGED_AS_TYPE]: {
+        homs: [
+          {
+            srcPrefix: GN._Prefix.pull,
+            dstPrefix: NodeAddress.fromParts(["sourcecred", "git", "COMMIT"]),
```
wchargin: Interestingly, this API exists, but is not public. What should we do
here? Should we make `_Prefix` public?

src/v3/plugins/github/graphView.test.js
```diff
@@ -0,0 +1,383 @@
...
+          const g = exampleWithParents().addEdge(references);
+          expect(() => new GraphView(g)).toThrow("Invariant: Expected dst");
+        });
+      });
+      describe("has parent edges", () => {
+        it("must satisfy specific hom relationships", () => {
+          const g = new Graph()
+            .addNode(GN.toRaw(repo))
+            // $ExpectFlowError
+            .addEdge(GE.createEdge.hasParent(repo, repo));
```
wchargin: A repository can’t have a parent at all. Can we instead provide a valid
child and a valid parent, like `hasParent(pullComment, issue)`? (In
other words, this test currently checks “general” hom relationships, not
“specific” ones.)

src/v3/plugins/github/graphView.test.js
```diff
@@ -0,0 +1,383 @@
...
+          expect(() => new GraphView(g)).toThrow("Invariant: Edge");
+        });
+        it("must be unique", () => {
+          const otherRepo = {type: "REPO", owner: "foo", name: "bar"};
+          const g = exampleWithParents();
+          g.addNode(GN.toRaw(otherRepo));
+          const otherParent = {
+            src: GN.toRaw(issue),
+            dst: GN.toRaw(otherRepo),
+            address: EdgeAddress.append(GE._Prefix.hasParent, "foobar"),
```
wchargin: This is mostly okay with me, but note that really the behavior is
undefined: the implementation is free to throw either “Parent invariant”
or “Bad address”, because `foobar` is not a valid GitHub node.

The consequence of writing the test like this is that it may fail
spuriously (and require patching) if and when the invariant-checking
implementation changes. But it’s plausible that the expected work of
that occurrence is lower than the actual work of robustifying this now,
so no change requested.

wchargin: In fact, now that I think about this a little more: it seems to me that
uniqueness of parent edges is derivable from the invariants (a) every
child has at least one outgoing parent edge and (b) all parent edges
match the parent and child specified in the edge address. Maybe this is
what you were saying originally on Discord, and I didn’t understand. If
this is correct, then I’m fine removing the test case altogether if you
like.

src/v3/plugins/github/graphView.test.js
```diff
@@ -0,0 +1,383 @@
...
+      const g = new Graph();
+      const view = new GraphView(g);
+      // no error, empty graph is fine
+      view.graph();
+      // introduce an invariant violation (no parent)
+      g.addNode(GN.toRaw(issue));
+      try {
+        view.graph();
+      } catch (_) {}
+      expect(() => view.graph()).toThrow("invariant violated");
```
wchargin: Cool; confirmed that this fails if the `finally` is squashed. Thanks for
writing this.

src/v3/plugins/github/graphView.test.js
```diff
@@ -0,0 +1,383 @@
...
+          });
+          it("has the right parent", () => {
+            expect(view.parent(reviewComment)).toEqual(review);
+          });
+        });
+      });
+    });
+  });
+  describe("invariants", () => {
+    const userlike: GN.UserlikeAddress = {
```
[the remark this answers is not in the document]
decentralion: I'm open to it, although not persuaded it would add much value.

If we wanted to change the internal format without affecting clients, we could also hold this structured representation as the API and make an internal-only representation that we translate into. I'm not sure what benefits taking the constructor approach has over the approach I described (other than being cleaner).

src/v3/plugins/github/graphView.test.js
decentralion: That is harder than it sounds, because such cases are already checked by the node-level parent invariant. Suppose that I try the following example:

```js
          const g = new Graph()
            .addNode(GN.toRaw(repo))
            .addNode(GN.toRaw(issue))
            .addEdge(GE.createEdge.hasParent(issue, repo))
            .addNode(GN.toRaw(pullComment))
            .addEdge(GE.createEdge.hasParent(pullComment, issue));
          expect(() => new GraphView(g)).toThrow("Invariant: Edge");
```

This triggers the node-wise invariant checker to throw, because `pullComment` _is_ a pull comment by virtue of having a pull as its address-encoded parent, and the `hasParent` edge is a mismatch on its address-encoded parent.

I can get away with using `repo,repo` because the node-wise invariant checker doesn't expect repo to have a parent at all.

It's technically possible to surgically construct addresses that are valid except with mutilated prefixes, but the effort-to-power ratio is unfavorable. Making no change.

src/v3/plugins/github/graphView.test.js
decentralion: It's impossible to create a valid GitHub has parent edge to test this condition, because the GitHub has parent edge is uniquely specified by the child address, and adding duplicate edges with conflicting contents would trigger a different violation. So this solution is as bad as any other that could feasibly test this codepath.

decentralion: +1, except (b) all parent edges match the child address specified in the edge address

decentralion:
FYI, I changed the base so as to cleanly use the new git node types (which required changes to pre-existing code).

I put [change-base] in the title so I wouldn't accidentally merge it into non-master. Feel free to review and approve as usual.

src/v3/plugins/github/graphView.test.js
wchargin: Noted, and reasonable. Thank you for the explanation.

decentralion:
Thanks!

PR #406, opened 2018-06-21
Create Git graph

wchargin:
Summary:
This commit adds logic to create the Git graph, modeled after the GitHub
graph creator in #405. In this commit, we do not include the
corresponding porcelain; a Git `GraphView` will be added subsequently.

Kudos to @decentralion for suggesting in #187 that I write the logic to
detect BECOMES edges against the high-level data structures. Due to that
decision, the logic and tests are copied directly from the V1 code
without change, because the high-level data structures are the same. The
new code is exactly the body of the `GraphCreator` class.

Test Plan:
Verify that the new snapshot is likely equivalent to the V1 snapshot,
using the heuristic that the two graphs have the same numbers of nodes
(59) and edges (84). (I have performed this check.)

wchargin-branch: git-v3-create-graph


wchargin:
@decentralion: Curious what you think about this approach of committing
the graph creation before the porcelain has been written. If you object,
I’m happy to wait until the porcelain is written and then resubmit this.

wchargin:
> I don't mind having the graph creator before the porcelain; your
> checksumming approach (and the comparative simpleness of the git
> plugin) leaves me reasonably confident it's correct.

Cool. After having written the checks, this code still looks correct.
:-)

> I sort of skimmed this PR for plausibility, but didn't reload all the
> git data structure stuff into my mental state. It seems like a
> straightforward port of existing logic so I'm happy to approve it, but
> if you want a more thorough review I can provide that too.

Nope, that’s fine with me. It is a mostly straightforward port.

PR #409, opened 2018-06-22
Only check graph invariants after modification

wchargin:
Summary:
Given that invariants are only invalidated when the modification count
is changed, it suffices to only call `_maybeCheckInvariants` from within
`_markModification` (and the constructor). This eliminates a function
call and cache key check for many public accessor functions, but does
not present a substantial improvement: the best-of-three timing only
improved from 3.866s to 3.799s, which is not statistically significant.
The patch makes it less obvious that the invariants are always in fact
valid, so I recommend that it not be merged.

Test Plan:
Existing unit tests suffice.

wchargin-branch: graph-check-invariants-only-indirectly

### DIFF
--- a/src/v3/core/graph.js
+++ b/src/v3/core/graph.js
@@ -266,7 +266,6 @@ export class Graph {
     // TODO(perf): Consider eliding this in production.
     const now = this._modificationCount;
     if (now === since) {
-      this._maybeCheckInvariants();
       return;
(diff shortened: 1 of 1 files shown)


wchargin:
As it says in the message, I recommend that this not be merged, but I
still wanted to put the code out there to document the timing
difference.

PR #412, opened 2018-06-23
[r] Refactor `RelationalView`

decentralion:
This pull merges several improvements to `RelationalView`:
- Modify the `Entry` types so that they generally point directly to entries, rather than holding addresses as pointers
- Add retrieval methods for every entry type
- Add `type` fields to entries to enable type refinement

See individual commits for details.

This PR should be merged via rebase and merge, not squash.


wchargin:
The second and third commits are fine with me, but I am skeptical of the
change to denormalize everything. This implies that the `RelationalView`
is no longer a view of a relational database but rather a view of a
[document-oriented database][dodb] (like MongoDB).

[dodb]: https://en.wikipedia.org/wiki/Document-oriented_database

Here is one practical problem presented by this example: in our example
repo, there are only two users, but the URL for `decentralion` appears
12 times and the URL for `wchargin` appears 6 times. It is easy to see
that this will lead to a lot of data duplication. This is worse than
just “wasted space”: we now have to ensure that these are all kept in
sync (namely, that `login` should be a super key).

This approach is the approach taken by the GitHub v3 API, which—you’ll
recall—yields [results like this][ghv3], wherein requesting the comments
on an issue requires serving 40 copies of URLs to various aspects of
@decentralion’s login page.

[ghv3]: https://api.github.com/repos/sourcecred/sourcecred/issues/350/comments

This approach also has the consequence that objects explode in size: a
reference to a repository now includes the full contents of all pull
request review comments under the repository.

As I understand it, you’re denormalizing to avoid null checks in the
client code. In a relational model, this isn’t a problem. If you want
all the pull request review comments on all the reviews of a given pull
request, you can

```sql
SELECT *
FROM pulls JOIN reviews JOIN review_comments
ON pulls.id = reviews.pull AND review_comments.review = reviews.id
WHERE pulls.id = ?
```

which is all type-checked and everything by the server.
([Example fiddle.][fiddle])

I’d strongly request reconsidering this, and I’m happy to chat about it.

[fiddle]: http://sqlfiddle.com/#!5/a05bf/2

decentralion:
Closing in favor of #418

PR #416, opened 2018-06-23
[cr] Find all references in the `RelationalView`

decentralion:
This pull request renames `findReferences` to `parseReferences`, and then implements `findReferences` as a module that finds every reference in a `RelationalView`. This will make it very easy to add reference edges to the GitHub graph.

See individual commits for details.

### DIFF
--- a/src/v3/plugins/github/findReferences.js
+++ b/src/v3/plugins/github/findReferences.js
@@ -1,0 +1,63 @@
+// @flow
+
+import * as N from "./nodes";
+import * as R from "./relationalView";
+import {parseReferences} from "./parseReferences";
+
+export type GithubReference = {|
+  src: N.TextContentAddress,
+  dst: N.ReferentAddress,
+|};
+
+type TextContentEntry =
+  | R.IssueEntry
+  | R.PullEntry
+  | R.CommentEntry
+  | R.ReviewEntry;
+export function* findReferences(
+  view: R.RelationalView
+): Iterator<GithubReference> {
+  function* allGithubEntities(): Iterator<R.Entry> {
+    yield* view.repos();
+    yield* view.issues();
+    yield* view.pulls();
+    yield* view.reviews();
+    yield* view.comments();
+    yield* view.userlikes();
+  }
+
+  const refToAddress: Map<string, N.StructuredAddress> = new Map();
+  for (const e of allGithubEntities()) {
+    const a = e.address;
+    refToAddress.set(e.url, a);
+    switch (e.type) {
+      case "USERLIKE":
+        refToAddress.set(`@${e.address.login}`, a);
+        break;
+      case "ISSUE":
+        refToAddress.set(`#${e.address.number}`, a);
+        break;
+      case "PULL":
+        refToAddress.set(`#${e.address.number}`, a);
+        break;
+      default:
+        break;
+    }
+  }
+
+  function* allTextContentEntries(): Iterator<TextContentEntry> {
+    yield* view.issues();
+    yield* view.pulls();
+    yield* view.reviews();
+    yield* view.comments();
+  }
+
+  for (const e of allTextContentEntries()) {
+    for (const ref of parseReferences(e.body)) {
+      const refAddress = refToAddress.get(ref);
+      if (refAddress != null) {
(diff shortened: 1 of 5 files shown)


src/v3/plugins/github/findReferences.js
```diff
@@ -1,58 +1,63 @@
...
-}
+export type GithubReference = {|
+  src: N.TextContentAddress,
+  dst: N.ReferentAddress,
+|};
 
-function findNumericReferences(body: string): string[] {
-  return findAllMatches(/(?:\W|^)(#\d+)(?:\W|$)/g, body).map((x) => x[1]);
-}
+type TextContentEntry =
```
wchargin: Should this be defined in the `relationalView.js` module?

src/v3/plugins/github/findReferences.test.js
```diff
@@ -2,130 +2,13 @@
...
-      )
-    ).toEqual([
-      "#125",
-      "https://github.com/sourcecred/sourcecred/pull/125#pullrequestreview-113402856",
-      "@wchargin",
-    ]);
+import * as R from "./relationalView";
+
+describe("plugins/github/findReferences", () => {
+  it("matches snapshot on exmaple-github data", () => {
```
wchargin: nit: _example_ (sp.)

src/v3/plugins/github/findReferences.js
```diff
@@ -0,0 +1,63 @@
...
+  function* allGithubEntities(): Iterator<R.Entry> {
+    yield* view.repos();
+    yield* view.issues();
+    yield* view.pulls();
+    yield* view.reviews();
+    yield* view.comments();
+    yield* view.userlikes();
+  }
+
+  const refToAddress: Map<string, N.StructuredAddress> = new Map();
```
wchargin: I don’t understand why this is called `refToAddress` and not
`urlToAddress`. It doesn’t map from `GithubReference`s. What
am I missing?

wchargin: I don’t understand why this is called `refToAddress` and not
`urlToAddress`. It doesn’t map from `GithubReference`s. What
am I missing?

src/v3/plugins/github/findReferences.test.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import {findReferences} from "./findReferences.js";
+
+import * as R from "./relationalView";
+
+describe("plugins/github/findReferences", () => {
+  it("matches snapshot on exmaple-github data", () => {
+    const data = require("./demoData/example-github");
```
wchargin: Do you want a type assertion here, or do you find it superfluous?

src/v3/plugins/github/findReferences.js
wchargin: I don’t understand why this is called `refToAddress` and not
`urlToAddress`. It doesn’t map from `GithubReference`s. What
am I missing?

src/v3/plugins/github/findReferences.js
```diff
@@ -0,0 +1,63 @@
...
+      case "USERLIKE":
+        refToAddress.set(`@${e.address.login}`, a);
+        break;
+      case "ISSUE":
+        refToAddress.set(`#${e.address.number}`, a);
+        break;
+      case "PULL":
+        refToAddress.set(`#${e.address.number}`, a);
+        break;
+      default:
```
wchargin: Consider adding here explicit cases for each other possible value of
`e.type`, and then using an `empty`-assertion in the `default`-case.
That way, when we add new GitHub node types, we’ll be forced to consider
this block, and either add a new reference type or explicitly decline to
do so.

src/v3/plugins/github/findReferences.js
decentralion: Made a clearer docstring in the new version, but:
"# 34", "@ wchargin" and "https://github.com/ sourcecred/sourcecred/pull/416" are all considered refs.

src/v3/plugins/github/findReferences.js
decentralion: Done, in the new version.

PR #418, opened 2018-06-26
Replace GitHub `relationalView`

decentralion:
In #411 and #412, I implemented `RelationalView`, which maps the GraphQL
GitHub data into a format that has `addresses`. The `RelationalView`
allows clients that have the `addresses` of various GitHub nodes to then
retrieve additional information, such as the body of the text.
`RelationalView` admits a convenient re-implementation of `createGraph`,
and an implementation of Github reference detection.

However, I don't like that when paired with the `graph`, the
`RelationalView` has a redundant representation of the `CONTAINS` type
edges; the `RelationalView`'s repos point to its contained issues, and
the graph does too. I think it would be better to have a single source
of truth for relations between data. As such, after offline discussion
with @wchargin, I decided to implement `EntityStore`.

`EntityStore` is much like the relational view, except with all relation
data (e.g. pointers from an issue to its author or its comments)
stripped out. It provides methods for retrieving an entity by address,
and for iterating over all entities of a given type.

Of course, to create the Graph, we *do* need that relational data. So
this commit creates an additional method, `addressify`.
`addressify` takes the GitHub GraphQL data, and annotates it
with GitHub addresses. Then, the addressed data is used both to
construct the `EntityStore` and to create the GitHub graph. These two
pieces replcae the `RelationalView`.

Test plan:
Observe snapshots, and run `yarn travis`.


src/v3/plugins/github/entityStore.js
```diff
@@ -0,0 +1,180 @@
+// @flow
```
wchargin: Could we include a top-of-module comment like,

```js
/*
 * Provides "entity" data types for GitHub data, which are like the
 * addressed types in `addressify.js` but with addresses referencing
 * other entities removed (an entity still stores its own address).
 */
```

?

src/v3/plugins/github/entityStore.test.js
```diff
@@ -0,0 +1,63 @@
...
+function entityStore() {
+  const data = require("./demoData/example-github");
+  const addressed = A.addressify(data);
+  return new E.EntityStore(addressed);
+}
+
+function checkEntities<T: E.Entity>(
+  name: string,
+  getAll: () => Iterator<T>,
+  get: (a: any) => ?T
```
wchargin: Can replace `a: any` with `a: $ElementType<T, "address">`.

src/v3/plugins/github/entityStore.test.js
```diff
@@ -0,0 +1,63 @@
...
+import * as E from "./entityStore";
+import * as A from "./addressify";
+
+function entityStore() {
+  const data = require("./demoData/example-github");
+  const addressed = A.addressify(data);
+  return new E.EntityStore(addressed);
+}
+
+function checkEntities<T: E.Entity>(
```
wchargin: Nice test automation. Clearly a good abstraction.

wchargin:
I’ve given a low-to-the-ground review; I don’t feel that I have enough
context to give a design review.

decentralion:
Replaced by #424.

PR #419, opened 2018-06-26
[c] Modify `createGraph` to use `DataAddressed`.

decentralion:
`addressify` takes the GitHub data and annotates it with `address`es.
This commit refactors the GitHub `createGraph` method to use this
addressed data, thus removing the need for `createGraph` to construct
addresses manually.

This replaces #413.

Test plan:
`yarn travis`, and graph snapshot unchanged.


decentralion:
Replaced by #428

PR #423, opened 2018-06-26
Represent submodule commits as normal commits

### LINKED ISSUE #417: Reconsider Git plugin graph structure
**TL;DR:** Remove the notion of “submodule commits” from the Git plugin,
observing that this makes for a more consistent “universal monorepo”
point of view.

---

Currently, the Git plugin has the following kinds of relationships
(among others):

  - A tree includes tree entries.
  - A tree entry may have a blob as contents.
  - A tree entry may have a tree as contents.
  - A tree entry may represent a submodule commit, in which case it has
    zero or more “submodule commit” nodes as contents.

This is all easy to understand until the last point. Submodule commits
change the has-contents relationship from “a tree entry has exactly one
node as contents” to “a tree entry has zero or more nodes as contents”.

Furthermore, the contents of a tree entry representing a submodule
commit cannot be determined by looking at the tree in Git. For a fixed
tree entry, the contents may even change over time as new commits are
added to the head of the repository—without force-pushes or anything.

This is all quite confusing, and of questionable value (which, of
course, is my fault).

The reason that I originally implemented the system like this is the
following. In Git, the actual tree entry for a submodule commit as
shown by `git cat-file -p HEAD^{tree}` or `git cat-file tree HEAD | xxd`
includes only the name of the submodule (as with all tree entries) and
the submodule commit hash. Armed with only the hash, one cannot find
more information about the remote commit, because one does not know how
to locate the repository to which it belongs. Including the URL helped
to remedy this; it gave one a place to look up the commit and find its
author, its contents, etc.

But, in the context of a SourceCred graph, I don’t think that this makes
much sense. Suppose that we have a cred graph for Repositories A and B,
and that Repository A has submodule commits from Repositories B and C.
Given just the hash of a submodule commit from Repository B, the
relevant commit is already in the graph, so we can get all the
information that we want. Given the hash of a submodule commit from
Repository C, we can’t get any more information about this commit—but we
don’t need to, either. What were we going to do with the URL—hope that
the repository is still mirrored there, go clone it, and generate a cred
graph on the fly? Doubtful.

Instead, we should consider moving to the following ontology, which is
simpler and a more faithful model of Git:

  - A tree includes tree entries.
  - A tree entry may have a blob as contents.
  - A tree entry may have a tree as contents.
  - A tree entry may have a commit as contents.

Here are some consequences:

  - We get invariants like “every tree entry has exactly one node as
    contents”. Consequently, both APIs and implementations become
    simpler.

  - We don’t have to read the `.gitmodules` file at every commit. This
    saves (a) for every commit: one `git` subprocess call that must read
    and parse a tree object, and (b) for commits that have at least one
    submodule, one additional `git` subprocess call that must read and
    parse a blob. It is probably a marginal performance improvement, but
    it makes the code significantly simpler.

  - We must relax the invariant that every commit have exactly one tree:
    in the case of a submodule commit, we might not know what the
    appropriate tree is. I posit that this is okay. The invariant
    changes from a uniqueness constraint to a subsingleton constraint.

  - It is increasingly apparent that the Git graph is not a graph of one
    Git repository, but of all Git repositories. Nodes are not scoped by
    owner or repository name; they simply exist in the universe,
    identified by their hashes.

cc @decentralion; thoughts appreciated.

wchargin:
Summary:
Closes #417. Submodule commits are dead; long live commits. The ontology
is now:

  - A tree includes tree entries.
  - A tree entry may have a blob as contents.
  - A tree entry may have a tree as contents.
  - A tree entry may have a commit as contents.

Test Plan:
Existing unit tests suffice, especially `#commits yields all commits`.

wchargin-branch: git-remove-submodule-commits

### DIFF
--- a/src/v3/plugins/git/__snapshots__/createGraph.test.js.snap
+++ b/src/v3/plugins/git/__snapshots__/createGraph.test.js.snap
@@ -121,7 +121,7 @@ Array [
           "3dfb84795e07341b05fad3a0d5a55f8304b2d7d8",
           "pygravitydefier",
         ],
-        "dstIndex": 15,
+        "dstIndex": 10,
         "srcIndex": 29,
       },
       Object {
@@ -186,7 +186,7 @@ Array [
           "569e1d383759903134df75230d63c0090196d4cb",
           "pygravitydefier",
         ],
-        "dstIndex": 15,
+        "dstIndex": 10,
         "srcIndex": 34,
       },
       Object {
@@ -303,7 +303,7 @@ Array [
           "7be3ecfee5314ffa9b2d93fc4377792b2d6d70ed",
           "pygravitydefier",
         ],
-        "dstIndex": 14,
+        "dstIndex": 7,
         "srcIndex": 43,
       },
       Object {
@@ -381,7 +381,7 @@ Array [
           "819fc546cea489476ce8dc90785e9ba7753d0a8f",
           "pygravitydefier",
         ],
-        "dstIndex": 14,
+        "dstIndex": 7,
         "srcIndex": 49,
       },
       Object {
@@ -459,7 +459,7 @@ Array [
           "bbf3b8b3d26a4f884b5c022d46851f593d329192",
           "pygravitydefier",
         ],
-        "dstIndex": 14,
+        "dstIndex": 7,
         "srcIndex": 55,
       },
       Object {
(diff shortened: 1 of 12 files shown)


wchargin:
I should also indicate: `yarn travis --full` passes.

PR #424, opened 2018-06-27
Add porcelain-style classes to `RelationalView`

decentralion:
Based on offline design discussion with @wchargin, we've decided to
upgrade the `RelationalView` to be *the* comprehensive source for GitHub
data inside SourceCred. The `RelationalView` will contain the full
dataset, including parsed relational information (such as
cross-references between GitHub entities). Then, we will project our
GitHub graph out of the `RelationalView`.

To that end, the `RelationalView` no longer exports raw data blobs.
Instead, it exports nice classes: `Repo`, `Issue`, `Pull`, `Review`,
and `Userlike`. These classes have convenient methods for accessing both
their own data and related entities, e.g. `repo.issues()` yields all
the issues in that repo.

This is effectively a port of #170 into the v3 API. The main difference
is that in v1, the Graph contained this data store, whereas in v3, we
will use this data store to generate the graph.

This supercedes #418.

Test plan:
The snapshot tests are quite readable.


src/v3/plugins/github/relationalView.js
```diff
@@ -224,3 +219,301 @@ export class RelationalView {
...
+  }
+  login(): string {
+    return this.address().login;
+  }
+  url(): string {
+    return this._entry.url;
+  }
+}
+
+function assertExists<T>(item: ?T, address: N.StructuredAddress): T {
```
wchargin: It would be nice to be able to write

```js
function assertExists<T>(
  item: ?T,
  address: N.StructuredAddress & {+type: $Call<$PropertyType<T, "type">>}
): T {
```

but, unfortunately, [a Flow bug][bug] prevents this from working in the
case of comment parents.

[bug]: https://github.com/facebook/flow/issues/6530

src/v3/plugins/github/relationalView.test.js
```diff
@@ -1,70 +1,162 @@
...
 
-  const repo = () => assertNotNull(repos()[0]);
-  it("repo matches snapshot", () => {
-    expect(repo()).toMatchSnapshot();
-  });
+  function has(name, method) {
+    it(`has ${name}`, () => {
+      const element = method();
+      let snapshot;
+      if (element != null && typeof element.type === "function") {
```
wchargin: This could be `if (element instanceof Entity)`, which is cleaner and
removes the need for the cast through `any`.

src/v3/plugins/github/relationalView.test.js
```diff
@@ -1,70 +1,162 @@
...
+        // element is an Entity. Entities have pointers to the RelationalView,
+        // and it would pollute our snapshot horribly. Just show the url.
```
wchargin: What do you think about having separate `has` and `hasEntity` functions?
That, I think, violates the principle of least astonishment less.

src/v3/plugins/github/relationalView.test.js
```diff
@@ -1,70 +1,162 @@
...
+    });
+  }
 
-  it("repo retrievable by address", () => {
-    expect(view.repo(repo().address)).toEqual(repo());
+  describe("RelationalView", () => {
+    function hasEntityMethods<T: R.Entity>(
+      name,
+      getAll: () => Iterator<T>,
+      get: (x: any) => ?T
```
wchargin: If `Entity` is made a superclass, then you can replace this `any` with:

```js
      get: (x: $Call<$PropertyType<T, "address">>) => ?T
```

(You should be able to do it with a union, too, but Flow complains,
which I think is due to that same bug.)

src/v3/plugins/github/relationalView.test.js
```diff
@@ -1,70 +1,162 @@
...
   });
-  it("userlike retrievable by address", () => {
-    expect(view.userlike(userlike().address)).toEqual(userlike());
+
+  describe("comment parent differentiation", () => {
+    function hasCorrectParent(name, parent) {
+      it(name, () => {
+        const comment = Array.from(parent.comments())[0];
+        expect(comment).toEqual(expect.anything());
+        const _parent = comment.parent();
```
wchargin: nit: Mind calling this `actualParent`?

src/v3/plugins/github/relationalView.js
```diff
@@ -77,28 +30,70 @@ export class RelationalView {
...
+    for (const entry of this._repos.values()) {
+      yield new Repo(this, entry);
+    }
   }
 
-  repo(address: N.RepoAddress): ?RepoEntry {
-    return this._repos.get(N.toRaw(address));
+  repo(address: N.RepoAddress): ?Repo {
+    const entry = this._repos.get(N.toRaw(address));
+    return entry == null ? null : new Repo(this, entry);
```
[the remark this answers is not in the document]
decentralion: No, I just treat them as identical in my head. I can modify it so that it remains undefined.

src/v3/plugins/github/relationalView.js
```diff
@@ -224,3 +219,301 @@ export class RelationalView {
...
+  entry: IssueEntry | PullEntry | ReviewEntry | CommentEntry
+) {
+  const address = entry.nominalAuthor;
+  if (address != null) {
+    const author = view.userlike(address);
+    yield assertExists(author, address);
+  }
+}
+
+export type Entity = Repo | Issue | Pull | Review | Comment | Userlike;
```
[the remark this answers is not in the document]
decentralion: Done.

src/v3/plugins/github/relationalView.test.js
decentralion: Done.

src/v3/plugins/github/relationalView.test.js
decentralion: I like the convenience of `has`, and since it's a local, test-only method I'm not too worried about the API. I can change it if you feel strongly.

src/v3/plugins/github/relationalView.test.js
decentralion: Done

src/v3/plugins/github/relationalView.test.js
decentralion: Done

src/v3/plugins/github/relationalView.test.js
wchargin: Okay, no problem.

src/v3/plugins/github/relationalView.js
```diff
@@ -224,3 +228,250 @@ export class RelationalView {
...
+        parent = this._view.pull(address);
+        break;
+      case "ISSUE":
+        parent = this._view.issue(address);
+        break;
+      case "REVIEW":
+        parent = this._view.review(address);
+        break;
+      default:
+        throw new Error(`Unexpected parent address: ${stringify(address)}`);
```
wchargin: Can add an `empty` assertion here.

PR #425, opened 2018-06-27
Copy Markov chain code from V1 to V3

wchargin:
Summary:
This code is independent of the graph abstraction, and so is mostly
copied. The only change is to the structure of the test code (we now
prefer to wrap everything in a big `describe` block with an absolute
path to the module under test).

Test Plan:
Unit tests included.

wchargin-branch: v3-markov-chain

PR #427, opened 2018-06-27
Convert V3 graphs to Markov chains

wchargin:
Summary:
This is based on the V1 file `basicPagerank.js`. Here, we include only
the code to convert a graph to a Markov chain. We necessarily rewrite
the implementation, as the graph format has changed, and also make the
key function `graphToAddressMapMarkovChain` easier to understand.

Test Plan:
Unit tests included.

wchargin-branch: v3-graph-markov-chain


wchargin:
> Also, please consider linking to the pull requests that implemented
> the v1 versions of this code.

Okay: as in #425, it’s not exactly clear what the target is, even more
so given the fact that this commit is not a copy but an adaptation, with
different implementation, types, and API. But I suppose that I could
link to the original TFJS implementation of PageRank. Would you like
that?

src/v3/core/attribution/graphToMarkovChain.test.js
```diff
@@ -0,0 +1,229 @@
...
+      };
+      expect(normalize(osmc)).toEqual(normalize(expected));
+    });
+
+    it("works on the example graph", () => {
+      const ag = advancedGraph();
+      const g = ag.graph1();
+      const epsilon = 0.5;
+      function edgeWeight() {
+        // This just gives nice round numbers in the snapshot.
```
decentralion: nit: there is no snapshot

PR #428, opened 2018-06-27
Refactor GH `createGraph` to use `RelationalView`

decentralion:
This commit modifies `github/createGraph` to use the `RelationalView`
class created in #424. The code is now much cleaner.

I also fixed some `any`s that were leaking in our test code (due to use
of runtime require for GitHub example data). These anys were discovered
by bumping into uncaught type errors. :)

This commit supercedes #413 and #419.

Test plan:
Observe that the graph snapshot was not changed.

### DIFF
--- a/src/v3/plugins/github/createGraph.js
+++ b/src/v3/plugins/github/createGraph.js
@@ -1,44 +1,14 @@
 // @flow
 
 import {Graph} from "../../core/graph";
-import type {
-  GithubResponseJSON,
-  RepositoryJSON,
-  ReviewJSON,
-  PullJSON,
-  IssueJSON,
-  CommentJSON,
-  ReviewCommentJSON,
-  NullableAuthorJSON,
-} from "./graphql";
-
-import type {
-  RepoAddress,
-  IssueAddress,
-  PullAddress,
-  ReviewAddress,
-  UserlikeAddress,
-  StructuredAddress,
-  AuthorableAddress,
-  ChildAddress,
-  ParentAddress,
-} from "./nodes";
-import {toRaw} from "./nodes";
-
+import * as GitNode from "../git/nodes";
+import * as N from "./nodes";
+import * as R from "./relationalView";
 import {createEdge} from "./edges";
 
-import {COMMIT_TYPE, toRaw as gitToRaw} from "../git/nodes";
-
-import {
-  reviewUrlToId,
-  issueCommentUrlToId,
-  pullCommentUrlToId,
-  reviewCommentUrlToId,
-} from "./urlIdParse";
-
-export function createGraph(data: GithubResponseJSON): Graph {
+export function createGraph(view: R.RelationalView): Graph {
   const creator = new GraphCreator();
-  creator.addData(data);
+  creator.addData(view);
   return creator.graph;
 }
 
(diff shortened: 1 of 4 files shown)


src/v3/plugins/github/createGraph.js
```diff
@@ -1,161 +1,95 @@
...
 
 class GraphCreator {
   graph: Graph;
+  view: R.RelationalView;
 
-  constructor() {
+  constructor(view: R.RelationalView) {
     this.graph = new Graph();
+    this.view = view;
+    for (const r of view.repos()) {
```
wchargin: Is there a reason to do this in the constructor? All else equal,
I prefer the previous pattern of an `addData` method that’s called by
the `createGraph` function. It’s nice when (a) constructors don’t do any
work, and (b) constructors don’t call instance methods. The former is
mostly aesthetic (in JavaScript, at least), but the latter can be
important: instance methods may assume that class invariants hold, but
these class invariants may not be established before the constructor has
finished.

src/v3/plugins/github/createGraph.test.js
```diff
@@ -2,11 +2,16 @@
...
 import {createGraph} from "./createGraph";
 import {GraphView} from "./graphView";
+import {RelationalView} from "./relationalView";
+import type {GithubResponseJSON} from "./graphql";
 import cloneDeep from "lodash.clonedeep";
 
 function exampleGraph() {
-  const data = cloneDeep(require("./demoData/example-github"));
-  return createGraph(data);
+  const data: GithubResponseJSON = cloneDeep(
```
wchargin: We could make this robust by creating a JavaScript module that exports

```js
// @flow

import cloneDeep from "lodash.clonedeep";

export function exampleGithub(): GithubResponseJSON {
  return cloneDeep(require("./example-github"));
}
```

so that clients don’t have to remember to (a) annotate the type and
(b) clone the raw value (which is stored in the `require` cache and
shared across all clients!).

You might also want to expose functions to construct the relational view
and/or graph from this example data, just for convenience.

Would you like to do that in a standalone commit prior to this one? (Or
after it, if you prefer.)

decentralion: I'd like to do it after this commit. Created an issue to track it: https://github.com/sourcecred/sourcecred/issues/430

src/v3/plugins/github/createGraph.js
decentralion: I changed it to suit your preference.

decentralion:
Great catch; the hierarchical implementation was a legacy from pre-relational-view implementation startegy.
I've refactored the code to be non-hierarchical, and it's now cleaner / even shorter. PTAL.

src/v3/plugins/github/createGraph.js
```diff
@@ -49,113 +19,53 @@ class GraphCreator {
...
+      // Authors are added before authored content
+      // Parents are added before children
+      yield* view.userlikes();
+      yield* view.repos();
+      yield* view.issues();
+      yield* view.pulls();
+      yield* view.reviews();
+      yield* view.comments();
+    }
+    for (const entity of entitiesInInsertionOrder()) {
```
wchargin: Hmm. I’m not a huge fan of this structure. You mix all the data types
together into one heterogeneous stream, and then for each element case
on its type to perform some actions, possibly multiple depending on the
type. It seems pretty ad hoc and complicated compared to the following
simple approach:

```js
for (const user of view.userlikes()) {
  this.addNode(user.address());
}
for (const repo of view.repos()) {
  this.addNode(repo.address());
}
for (const issue of view.issues()) {
  this.addNode(issue.address());
  this.addHasParent(issue);
  this.addAuthors(issue);
}
for (const pull of view.pulls()) {
  this.addNode(pull.address());
  this.addHasParent(pull);
  this.addAuthors(pull);
  const commit = entity.mergedAs();
  if (commit != null) {
    this.graph.addNode(GitNode.toRaw(commit));
    this.graph.addEdge(createEdge.mergedAs(entity.address(), commit));
  }
}
for (const review of view.reviews()) {
  this.addNode(review.address());
  this.addHasParent(review);
  this.addAuthors(review);
}
for (const comment of view.comments()) {
  this.addNode(comment.address());
  this.addHasParent(comment);
  this.addAuthors(comment);
}
```

Compared to the approach above, the approach in this pull request has
less duplicated code, but it’s not clear that the deduplication actually
represents a useful abstraction.

For reference, the proposed approach is only two extra lines of code.

What do you think?

wchargin: I would alternately support a “trait-based” approach like:

```js
// Add nodes
for (const node of view.allEntities()) {
  this.addNode(node.address());
}

// Add parents
for (const node of view.childEntities()) {
  this.addHasParent(node);
}

// Add authors
for (const node of view.authorEntities()) {
  this.addAuthors(node);
}

// Add "merged-as" edges
for (const pull of view.pulls()) {
  const commit = entity.mergedAs();
  if (commit != null) {
    this.graph.addNode(GitNode.toRaw(commit));
    this.graph.addEdge(createEdge.mergedAs(entity.address(), commit));
  }
}
```

…with the differences from the approach in this pull request being that
(a) the entities come from named methods under the purview of the
relational view, (b) there’s clear separation among traits (e.g., the
coincidence of the set of child entities and the set of authorable
entities is not relevant), and (c) there’s no need to lose information
by injecting into a heterogeneous stream and then refine out of it
again.

wchargin: A notable difference between the two methods that I’ve proposed is how
they deal with extension. With the first method, it is easy to see how
to add a new entity type, whereas adding a new trait is a cross-cutting
concern. The opposite is true with the second proposal. In either case,
of course, adding a new entity type or trait is not _difficult_.

decentralion: Changed to your trait based approach.

src/v3/plugins/github/relationalView.js
```diff
@@ -127,6 +127,36 @@ export class RelationalView {
...
+  }
+
+  *authoredEntities(): Iterator<AuthoredEntity> {
+    yield* this.issues();
+    yield* this.pulls();
+    yield* this.reviews();
+    yield* this.comments();
+  }
+
+  *entities(): Iterator<Entity<Entry>> {
```
wchargin: Yep. If you want to make `Entity` an implementation detail and expose the union only, that's okay with me. (I see now why it makes sense in parallel with the other concepts.)

PR #447, opened 2018-06-29
Save the GitHub relational store from the CLI

wchargin:
Summary:
This provides a command-line entry point `load-plugin-v3` (which will
become `load-plugin` eventually), which fetches the GitHub data via
GraphQL and saves the resulting `RelationalStore` to disk.

A change to the Babel config is needed to prevent runtime errors of the
form `_callee7` is not defined, where `_callee7` is a gensym that is
appears exactly once in the source (in use position, not definition
position). I’m not sure exactly what is causing the error or why this
config change fixes it. But while this patch may be fragile, I don’t
think that it’s likely to subtly break anything, so I’m okay with
pushing it for now and dealing with any resulting breakage as it arises.

Paired with @decentralion.

Test Plan:
Run `yarn backend`, then run something like:

```
node bin/sourcecredV3.js load-plugin-v3 \
    sourcecred example-github --plugin github
```

Inspect results in `SOURCECRED_DIR/data/OWNER/NAME/github/view.json`,
where `SOURCECRED_DIR` is `/tmp/sourcecred` by default, and `OWNER` and
`NAME` are the repository owner and name.

This example repository takes about 1.1 seconds to run. The SourceCred
repository takes about 45 seconds.

wchargin-branch: cli-load-plugin

### DIFF
--- a/src/v3/cli/commands/loadPlugin.js
+++ b/src/v3/cli/commands/loadPlugin.js
@@ -1,0 +1,87 @@
+// @flow
+
+import {Command, flags} from "@oclif/command";
+import mkdirp from "mkdirp";
+import path from "path";
+
+import {loadGithubData} from "../../plugins/github/loadGithubData";
+import {pluginNames, sourcecredDirectoryFlag} from "../common";
+
+export default class PluginGraphCommand extends Command {
+  static description = "load data required for a single plugin";
+
+  static args = [
+    {
+      name: "repo_owner",
+      required: true,
+      description: "owner of the GitHub repository for which to fetch data",
+    },
+    {
+      name: "repo_name",
+      required: true,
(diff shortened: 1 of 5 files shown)


src/v3/cli/commands/loadPlugin.js
```diff
@@ -0,0 +1,87 @@
...
+    {
+      name: "repo_name",
+      required: true,
+      description: "name of the GitHub repository for which to fetch data",
+    },
+  ];
+
+  static flags = {
+    plugin: flags.string({
+      description: "plugin whose graph to generate",
```
decentralion: plugin whose data to load

wchargin: Done; thanks.

PR #451, opened 2018-06-29
Implement detection of paired authorship

### LINKED ISSUE #218: Implement "Paired with @" parsing
@wchargin and I have pair-programmed a lot of pull requests, e.g. #35. We should add logic to the GitHub plugin such that it creates dual authorship edges in that case.

decentralion:
This commit enables paired authorship on GitHub authored entities.
If the entity has the string "Paired with" in the body, followed by a
username reference, that entity will be recorded as having dual
authorship, with the nominal author and the paired-with author being
treated identically in the relational view and the graph.

If there's a need to pair with more than one author, the "Paired with"
signifier may be repeated. The regex matcher is forgiving of
capitalizing the P or W, and an optional colon may be added immediately
after the word "with".

Note that the code assumes that every `TextContentEntity` is also an
`AuthoredEntity`. If that changes, it will cause a type error and we'll
need to refine the code somewhat.

As implemented, it is impossible for the same user to author a post
multiple time; if this is textually suggested (e.g. by a paired-with
reference to the post's nominal author), the extra paired-with
references are silently ignored. Also, having a paired-with reference
suppresses the basic reference (although it is possible to have a post
that is paired with someone, and additionally references them).

Test plan:
Tests have been updated, and the behavior of the parser is extensively
tested. For an end-to-end demonstration, I've also added a unit test in
the relational view that verifies that sourcecred/example-github#10 has
two authors. You can also see that the graph snapshot has updated to
include additional authorship edges (and that corresponding reference
edges have disappeared).

Closes #218


src/v3/plugins/github/parseReferences.js
```diff
@@ -1,19 +1,12 @@
...
-    if (m) {
-      matches.push(m);
-    }
-  } while (m);
-  return matches;
-}
+export type ParsedReference = {|
+  // "@user" or "#123" or "https://github.com/owner/name/..."
+  +ref: string,
+  +refType: "basic" | "pairedWith",
```
wchargin: Might as well be `"BASIC" | "PAIRED_WITH"` for consistency.

src/v3/plugins/github/parseReferences.js
```diff
@@ -24,17 +17,33 @@ export function parseReferences(body: string): string[] {
...
+  }));
 }
 
-function findUsernameReferences(body: string): string[] {
-  return findAllMatches(/(?:\W|^)(@[a-zA-Z0-9-]+)(?:\W|$)/g, body).map(
-    (x) => x[1]
-  );
+function findUsernameReferences(body: string): ParsedReference[] {
+  const pairedWithRefs = findAllMatches(
+    /(?:\W|^)(?:P|p)aired (?:w|W)ith:? (@[a-zA-Z0-9-]+)(?:\W|$)/g,
```
wchargin: Appreciate the leniency in the regexp. Could you also permit the space
between “paired” and “with” to be a single hyphen, like `Paired-with`?
This makes the format compatible with `git-interpret-trailers`, which
uses keys like `Signed-off-by`.

src/v3/plugins/github/parseReferences.js
```diff
@@ -24,17 +17,33 @@ export function parseReferences(body: string): string[] {
...
+  const basicRefs = findAllMatches(
+    /(?:\W|^)(@[a-zA-Z0-9-]+)(?:\W|$)/g,
+    body
+  ).map((x) => ({ref: x[1], refType: "basic"}));
+  for (const {ref} of pairedWithRefs) {
+    const basicRefIndexToRemove = basicRefs.findIndex((x) => x.ref === ref);
+    if (basicRefIndexToRemove === -1) {
+      throw new Error(`Couldn't find basic ref for paired with ref: ${ref}`);
+    }
+    basicRefs.splice(basicRefIndexToRemove, 1);
```
wchargin: Funky implementation, but okay.

src/v3/plugins/github/relationalView.js
```diff
@@ -346,10 +346,32 @@ export class RelationalView {
...
+            case "pairedWith":
+              if (refAddress.type !== N.USERLIKE_TYPE) {
+                throw new Error(
+                  `paired with non userlike: ${stringify(refAddress)}`
+                );
+              }
+              const userlike = this.userlike(refAddress);
+              if (userlike == null) {
+                throw new Error(
+                  `paired with nonexistent userlike: ${stringify(refAddress)}`
```
wchargin: This should not be an error.

Watch: Paired with @someonewhodoesnotexist.

wchargin: Okay: I see now how the code works. Thanks for improving the error message. :-)

src/v3/plugins/github/parseReferences.js
decentralion: Done.

src/v3/plugins/github/parseReferences.js
decentralion: Done.

src/v3/plugins/github/relationalView.js
decentralion: Done.

PR #454, opened 2018-06-29
fixed minor typos

hamidzr:


decentralion:
Thanks for the PR! I think "themself" actually makes sense in this context, though. The README is using the [singular they], and according to [this table on wikipedia] "themself" is the default reflexive form of the singular they (although "themselves" is also valid). 

I'm going to close this as I don't think it's a typo, but thanks for sending in the PR! Feel free to drop by [our discord] and say hi anytime. :slightly_smiling_face: 

[our discord]: https://discord.gg/tsBTgc9
[singular they]: https://en.wikipedia.org/wiki/Singular_they
[this table on wikipedia]: https://en.wikipedia.org/wiki/Singular_they#Inflected_forms_and_derivative_pronouns

PR #466, opened 2018-06-30
Recreate Pagerank Table

decentralion:
Ports #265 to the v3 branch, along with some tweaks:
- Only display log score, and normalize them by adding 10 (so that most
are non-negative)
- Change colors to a soothing green
- Improve display, e.g. make overflowing node description text wrap
within the row

Implements most of the tests requested in #269

Test plan: Many unit tests added

Paired with @wchargin

### DIFF
--- a/src/v3/app/credExplorer/PagerankTable.js
+++ b/src/v3/app/credExplorer/PagerankTable.js
@@ -1,0 +1,252 @@
+// @flow
+
+import sortBy from "lodash.sortby";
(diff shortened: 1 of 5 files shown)


src/v3/app/credExplorer/App.js
```diff
@@ -1,127 +0,0 @@
-// @flow
```
wchargin: nooooo

PR #470, opened 2018-06-30
Add load CLI command

decentralion:
The `load` command replaces `plugin-load`. By default, it loads data for
all plugins, and does so in parallel using execDependencyGraph. If
passed the optional `--plugin` flag, then it will load data just for
that plugin.

As an implementation detail, when loading all plugins, load calls itself
with the plugin flag set.

Usage:
`node bin/sourcecred.js load repoOwner repoName`

Test plan:
Tested by hand; I blew away my SourceCred directory and then loaded the
example-github repository.

### DIFF
--- a/src/v3/cli/commands/load.js
+++ b/src/v3/cli/commands/load.js
@@ -6,10 +6,17 @@ import path from "path";
 
 import {loadGithubData} from "../../plugins/github/loadGithubData";
 import {loadGitData} from "../../plugins/git/loadGitData";
-import {pluginNames, sourcecredDirectoryFlag} from "../common";
+import {
+  pluginNames,
+  nodeMaxOldSpaceSizeFlag,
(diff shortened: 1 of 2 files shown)


src/v3/cli/commands/load.js
```diff
@@ -46,13 +54,49 @@ export default class PluginGraphCommand extends Command {
...
+        "node",
+        `--max_old_space_size=${maxOldSpaceSize}`,
+        "./bin/sourcecred.js",
+        "load",
+        repoOwner,
+        repoName,
+        "--plugin",
+        pluginName,
+        "--github-token",
+        githubToken,
```
wchargin: This passes the literal string `undefined` to the child process.

decentralion: Fixed.

PR #477, opened 2018-07-03
Properly detect GitHub references at start of line

wchargin:
Summary:
The `^` metacharacter only matches the start of a non-initial line when
the `m` (“multiline”) flag is set on the RegExp. This flag was set only
for the GitHub URL RegExp.

Test Plan:
Regression tests added. They fail before this patch and pass with it.

wchargin-branch: fix-sol-github-references


decentralion:
Thanks for catching this!

PR #489, opened 2018-07-05
Display edges in Cred Explorer

decentralion:
Previously, when expanding a node in the cred explorer, it would display
the neighboring nodes, but not any information about the edges linking
to that node. If the same node was reached by multiple edges, this
information was not communicated to the user.

As of this commit, it now concisely communicates what kind of edge was
connecting the chosen node to its adjacencies. There's a new `edgeVerb`
method that plugin adapters must implement, which gives a
direction-based verb descriptiong of the edge, e.g. "authors" or "is
authored by".

Test plan:
Unit tests added to the PagerankTable tests, and hand inspection.

Paired with @wchargin

### DIFF
--- a/src/app/credExplorer/PagerankTable.js
+++ b/src/app/credExplorer/PagerankTable.js
@@ -7,8 +7,11 @@ import {
   Graph,
   NodeAddress,
   type NodeAddressT,
+  type Neighbor,
   Direction,
+  type Edge,
   EdgeAddress,
+  type EdgeAddressT,
 } from "../../core/graph";
 import type {PagerankResult} from "../../core/attribution/pagerank";
 import type {PluginAdapter} from "../pluginAdapter";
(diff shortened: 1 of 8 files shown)


wchargin:
> Paired with @wchargin

For reference, I only deserve about half of a paired-with edge;
@decentralion wrote most of the code, and I helped test it. :-)

src/app/credExplorer/PagerankTable.js
```diff
@@ -48,6 +51,44 @@ export function nodeDescription(
...
+  }
+}
+
+export function neighborVerb(
+  {node, edge}: Neighbor,
+  adapters: $ReadOnlyArray<PluginAdapter>
+): string {
+  const forwardVerb = edgeVerb(edge.address, "FORWARD", adapters);
+  const backwardVerb = edgeVerb(edge.address, "BACKWARD", adapters);
+  if (edge.src === edge.dst) {
```
wchargin: I’ll leave a TODO on my mental backlog to make this show two badges:
e.g., “**REFERENCES** and **IS REFERENCED BY** something”.

src/app/credExplorer/PagerankTable.js
```diff
@@ -193,23 +238,36 @@ class RecursiveTable extends React.PureComponent<RTProps, RTState> {
...
+          <span>
+            {edgeVerbString != null && (
+              <React.Fragment>
+                <span
+                  style={{
+                    display: "inline-block",
+                    textTransform: "uppercase",
+                    fontWeight: 700,
+                    fontSize: "smaller",
+                    padding: 2,
```
wchargin: Remove padding?

src/app/pluginAdapter.js
```diff
@@ -1,16 +1,18 @@
 // @flow
 
-import type {Graph, NodeAddressT} from "../core/graph";
+import type {Graph, NodeAddressT, EdgeAddressT} from "../core/graph";
 
 export interface Renderer {
   nodeDescription(NodeAddressT): string;
+  edgeVerb(EdgeAddressT, "FORWARD" | "BACKWARD"): string;
```
wchargin: There are a number of alternate factorizations that make sense here.
First, this function could return `{forward: string, backward: string}`
instead of requiring two invocations (and two `try`-`catch`es, etc.).
Second, it would seem reasonable to me to move these onto an
`edgeTypes()` method of the plugin adapter itself, for symmetry with the
node adapter. This seems like a low-difficulty refactor, so I’m okay
with merging the code as is.

src/app/credExplorer/PagerankTable.js
```diff
@@ -132,7 +173,7 @@ export class PagerankTable extends React.PureComponent<Props, State> {
           </tr>
         </thead>
         <tbody>
-          <RecursiveTables
+          <NodesTables
```
wchargin: Inlining this class would be fine with me. No strong preference.

src/app/credExplorer/PagerankTable.js
```diff
@@ -258,3 +316,47 @@ class RecursiveTables extends React.PureComponent<RecursiveTablesProps> {
...
+  +depth: number,
+  +adapters: $ReadOnlyArray<PluginAdapter>,
+|};
+
+class NeighborsTables extends React.PureComponent<NeighborsTablesProps> {
+  render() {
+    const {neighbors, graph, pagerankResult, depth, adapters} = this.props;
+    return neighbors
+      .slice()
+      .sort((a, b) => {
```
wchargin: Can we use `sortBy` here so that
 1. it is simpler, and
 2. we are deterministic in case of ties on weight? (i.e., fall back to name or something)

decentralion: Done

PR #493, opened 2018-07-06
Persist Cred Explorer weights in LocalStore

decentralion:
The implementation is similar to the LocalStore usage in
`credExplorer/app.js`. Had to make a spurious refactor from Map to
Object because ES6 maps don't stringify by default, and I didn't feel
like writing a custom JSON serializer.

Test plan:
Didn't add unit tests, although at some point we should come up with a
nice LocalStore mock and test LocalStore code. I did, however, manually
try it out and verify that it works :)

Paired with @wchargin

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -11,6 +11,7 @@ import {
 
 import {type EdgeEvaluator} from "../../core/attribution/pagerank";
 import {byEdgeType, byNodeType} from "./edgeWeights";
+import LocalStore from "./LocalStore";
 
 // Hacks...
 import * as GithubNode from "../../plugins/github/nodes";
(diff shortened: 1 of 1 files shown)


src/app/credExplorer/WeightConfig.js
```diff
@@ -22,37 +23,45 @@ type Props = {
   onChange: (EdgeEvaluator) => void,
 };
 
+// The key should be an EdgeAddressT, but a Flow bug prevents this.
```
wchargin: Unfortunate. I have filed a bug for this:
<https://github.com/facebook/flow/issues/6569>

I don’t have any good workarounds at this time.

Once these weights become non-hard-coded, I suspect that we will be able
to properly type this map. The problem seems to only occur with literal
keys: if you have `o: {[T]: U}` then you can assign `o[t] = u`. For that
reason, I’m okay with this temporary decrease in precision.

src/app/credExplorer/WeightConfig.js
```diff
@@ -84,55 +104,53 @@ export class WeightConfig extends React.Component<Props, State> {
...
             min={-10}
             max={10}
             step={0.1}
-            value={currentValue.logWeight}
+            value={logWeight}
             onChange={(e) => {
               const value: number = e.target.valueAsNumber;
-              const edgeWeights = new Map(this.props.edgeWeights);
-              const oldValue = edgeWeights.get(key);
+              const edgeWeights = {...this.props.edgeWeights};
```
wchargin: Objects have unchecked property access, so this can be elided to:

```js
const value: number = e.target.valueAsNumber;
const edgeWeights = {
  ...this.props.edgeWeights,
  [key]: {...this.props.edgeWeights[key], logWeight: value},
};
this.props.onChange(edgeWeights);
```

decentralion:
Well, in this case I wrote this on the side while pairing with you on another (more important) pull. If I get paired with cred on that pull, it seems only fair that you get paired with on this one; we were pair programming when both these pulls were produced.

PR #494, opened 2018-07-06
Add edge direction configuration to Cred Explorer

decentralion:
This commit adds another bank of sliders to the cred explorer, for
changing the directionality of edges. The sliders have the range [0,1]
with step size of 0.01.

The layout is pretty ugly and clearly should be refactored. But playing
with these sliders is interesting :)

Test plan: We don't have any unit tests on the WeightConfig, but I did
drive it by hand. An interesting experiment is to set the AUTHORS edge
directionality to 1, so that users get no credit for authoring posts. As
expected, this utterly tanks the users' scores; many users then have a
score of -Infinity.

image

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -126,11 +126,10 @@ class EdgeConfig extends React.Component<{
   edgeWeights: EdgeWeights,
   onChange: (EdgeWeights) => void,
 }> {
-  render() {
-    const controls = [];
-    for (const key of Object.keys(this.props.edgeWeights)) {
+  weightControls() {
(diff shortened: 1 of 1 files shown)


src/app/credExplorer/WeightConfig.js
```diff
@@ -151,11 +150,42 @@ class EdgeConfig extends React.Component<{
...
+      );
+    });
+  }
+  render() {
     return (
       <div>
         <h2>Edge weights (in log space)</h2>
-        {controls}
+        {this.weightControls()}
+        <h2>Edge Directionality</h2>
```
wchargin: nit: downcase `directionality`, please.

PR #495, opened 2018-07-06
Add utilities for working with ES6 Maps

wchargin:
Summary:
We’d like to like ES6 `Map`s, because they provide better type safety
than objects (primarily, `Map.prototype.get` has nullable result type).
However, the vanilla APIs are weak. Prominent problems are that `Map`s
always become `"{}"` under `JSON.stringify`, that there is no easy way
to convert between `Map`s and objects, and that there are no functions
to map over the keys and values of `Map`s.

In this commit, we add versions of those functions to a utility module.
The value-level implementations are straightforward, but these functions
nevertheless deserve a utility module because the types are somewhat
tricky to get right. The implementation requires casts through `any`,
and these should be written, analyzed, and proven correct just once. (In
particular, it would be easy to write an unsound type for `fromObject`.)

In a followup commit, we will amend existing portions of the codebase to
use these functions.

Test Plan:

wchargin-branch: map-util

### DIFF
--- a/src/util/map.js
+++ b/src/util/map.js
@@ -1,0 +1,124 @@
+// @flow
+
+/**
+ * Convert a string-keyed map to an object. Useful for conversion to
+ * JSON. If a map's keys are not strings, consider invoking `mapKeys`
+ * first.
+ */
+export function toObject<K: string, V, InK: K, InV: V>(
+  map: Map<InK, InV>
+): {[K]: V} {
+  const result = {};
+  for (const [k, v] of map.entries()) {
+    result[k] = v;
+  }
+  return result;
+}
+
+/**
+ * Convert an object to a map. The resulting map will have key-value
+ * pairs corresponding to the enumerable own properties of the object in
+ * iteration order, as returned by `Object.keys`.
+ */
+export function fromObject<K, V, InK: K & string, InV: V>(object: {
+  [InK]: InV,
+}): Map<K, V> {
+  const result = new Map();
+  const keys = (((Object.keys(object): string[]): any): InK[]);
+  for (const key of keys) {
+    result.set(key, object[key]);
+  }
+  return result;
+}
+
+/**
+ * Shallow-copy a map, allowing upcasting its type parameters.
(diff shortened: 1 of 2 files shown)


src/util/map.js
```diff
@@ -0,0 +1,124 @@
+// @flow
+
+/**
+ * Convert a string-keyed map to an object. Useful for conversion to
+ * JSON. If a map's keys are not strings, consider invoking `mapKeys`
+ * first.
+ */
+export function toObject<K: string, V, InK: K, InV: V>(
```
decentralion: I find the type signature to this function confusing. Could the docstring explain it, please?

src/util/map.js
```diff
@@ -0,0 +1,124 @@
...
+  }
+  return result;
+}
+
+/**
+ * Convert an object to a map. The resulting map will have key-value
+ * pairs corresponding to the enumerable own properties of the object in
+ * iteration order, as returned by `Object.keys`.
+ */
+export function fromObject<K, V, InK: K & string, InV: V>(object: {
```
decentralion: As above, could the docstring explain the type signature?

wchargin:
Okay. That’s fine with me. I’ll just note the following: the types are
written to be maximally permissive while remaining sound. There are
simpler types that I could have written, but they would have forbidden
legitimate uses. It is not necessary to fully understand the types to
use them properly, just as invoking `Array.from` doesn’t require one to
identify [precisely which of the six complicated declarations of that
function is being used][1]. In general, these functions should behave
just “as one expects” from the names, and Flow will point out if a use
is incorrect. But, all that said, I’m still happy to discuss and/or
document the decisions.

[1]: https://github.com/facebook/flow/blob/fc582b354645b4e6c447d02109583b5baa7fd716/lib/core.js#L279-L284

wchargin:
Okay—before this is merged, (a) do you want to chat in person, and
(b) are there any changes that you want to see?

decentralion:
Nope, feel free to merge it.

PR #497, opened 2018-07-06
Use `Map`s in `WeightConfig`

wchargin:
Summary:
Now that `MapUtil` provides `toObject`/`fromObject`, we can keep storing
weights in localStorage while representing them as ES6 `Map`s in memory.

Here are some advantages:
  - The code is genuinely more typesafe. While writing this,
    I accidentally wrote `edgeWeights.get(key)`, where `edgeWeights`
    should have been `nodeWeights`. This was caught at compile time, and
    would not have been in the previous version.
  - Relatedly, the code now has zero `any`-casts as opposed to four.
  - The initialization of the default values is not abysmally ugly.
  - Whenever we iterate over these maps, (a) we can use `.entries()`,
    and (b) we don’t have to cast between string keys and semantic keys.
    This simplifies some of the control flow.
  - The extra null-checking on `get` forces us to think about ways in
    which the check might fail. For instance, if a new set of plugin
    adapters is loaded while a user drags a slider, we might well hit
    the null case that now has an explanatory comment.
  - A particularly annoying Prettier line-break is avoided. :-)

Here are some disadvantages:
  - The null-pipelining around the `rehydrate` function is a bit
    annoying. As @decentralion pointed out, what we want here is not a
    default value, but a default value and a function to transform a
    present value. This is Haskell’s `maybe : β → (α → β) → Maybe α → β`
    or Java’s `optional.map(fromObject).orElse(defaultValue)`. This
    commit implements one approach; another is to note that `fromObject`
    is invertible, writing `fromObject(LocalStore.get(k, toObject(d)))`.
  - That’s it, I think?

Test Plan:
I’ve tested that the sliders for both edge and node weights correctly
influence the PageRank behavior, that the component is properly
initialized with an empty localStorage, and that the component properly
rehydrates from localStorage.

wchargin-branch: weightconfig-maps


wchargin:
> Will need rebase

Which is not a bad thing—instead of removing four `any`-casts, I get to
remove five! :-)

PR #501, opened 2018-07-06
Add `pagerankNodeDecomposition` to ease rendering

wchargin:
Summary:
When updating `PagerankTable` to work with contributions, we found it
difficult to keep track of everything when we tried to do two things
simultaneously: compute the values to be displayed, and render them
hierarchically. @decentralion suggested computing the relevant data
ahead of time, and then having a straightforward React component to
render this structure. This would incidentally make `PagerankTable`
easier to test.

This commit implements that data structure and the function to create
it from a `PagerankResult`. A subsequent commit will update
`PagerankTable` accordingly.

As evidence that this structure is well-designed, note that the main
contents of a contribution row can be rendered entirely from a
`ScoredContribution` datum (though the component will still of course
require the full `PagerankNodeDecomposition` to pass down to its
children). (At least, I think that it can be!)

Designed with @decentralion.

Test Plan:
Unit tests added. I have checked that the snapshot is structurally
correct: each node has contributions with the correct contributors.
I did not manually compute the stationary distribution and check the
snapshot for correctness. The snapshot is complemented by automated
tests.

wchargin-branch: pagerank-node-decomposition

### DIFF
--- a/src/core/attribution/__snapshots__/pagerankNodeDecomposition.test.js.snap
+++ b/src/core/attribution/__snapshots__/pagerankNodeDecomposition.test.js.snap
@@ -1,0 +1,182 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`core/attribution/contributions decompose has the expected output on a simple asymmetric chain 1`] = `
+Map {
+  "NodeAddress[\\"n1\\"]" => Object {
+    "score": 0.19117656878499834,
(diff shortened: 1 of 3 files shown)


wchargin:
Due to converging histories and the GitHub pull request workflow, this
pull request necessarily has irrelevant commits. Please review only the
most recent commit.

wchargin:
Relatedly: this pull request depends on #495.

PR #506, opened 2018-07-09
Use `NullUtil` functions where appropriate

wchargin:
Summary:
The aesthetically nicest win is in `WeightConfig`. Other changes are
nice to have.

In many cases, we reduce the specificity of error messages thrown. For
instance, if an invariant was violated on an edge `e`, then we might
have thrown an error with message `EdgeAddress.toString(e.address)`. But
we did so not because we thought that this was genuinely worth it, but
only because we were forced to explicitly throw an error at all. These
errors should never be hit, anyway, so we don’t feel bad about replacing
these with errors that are simply the string `"null"` or `"undefined"`,
as appropriate.

Test Plan:
Running `yarn travis --full` passes, and the cred explorer still seems
to work with both populated and empty `localStorage`.

wchargin-branch: use-null-util

### DIFF
--- a/src/app/credExplorer/edgeWeights.js
+++ b/src/app/credExplorer/edgeWeights.js
@@ -8,6 +8,8 @@ import {
 } from "../../core/graph";
 import type {EdgeEvaluator} from "../../core/attribution/pagerank";
 
+import * as NullUtil from "../../util/null";
+
 export function byEdgeType(
   prefixes: $ReadOnlyArray<{|
     +prefix: EdgeAddressT,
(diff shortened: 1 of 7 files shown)


wchargin:
By the way, `RelationalView` has a bunch of code like

```js
    return entry == null ? entry : new Review(this, entry);
```

but I chose not to replace this with `NullUtil.orElse`, because it would
have requiring implementing `orElseGet: ?T => (() => T) => T` to defer
the work of creating the `new Review` in the common case where it
already exists.

Thoughts?

wchargin:
(It also would have been slower and taken longer.)

src/app/credExplorer/edgeWeights.js
```diff
@@ -38,15 +36,14 @@ export function byNodeType(
   |}>
 ): EdgeEvaluator {
   return function evaluator(edge: Edge) {
-    const srcDatum = prefixes.find(({prefix}) =>
-      NodeAddress.hasPrefix(edge.src, prefix)
+    const srcDatum = NullUtil.orThrow(
```
decentralion: In this case, I think it was actually cleaner before. I'm fine either way tho.

wchargin: On second thought, I agree: the old version is clearer because it only
performs one test, and it’s staying in a statement context so we don’t
buy anything from using an expression form. I’ll revert this change.
Thanks!

Issue #476, opened 2018-07-03
Weighted PageRank


decentralion:
# Weighted PageRank Graphs

## Domain introduction

We have a heterogeneous graph `G` which represents contributions/contributors
in SourceCred, and connections between them.

It's heterogeneous in the sense that there are many different "types" of nodes
and edges. In the interest of concreteness, here is a sample of of node types
and edge types we have or expect to add.

### Node types:

  - *GitHub users* are identities we want to assign cred to
  - *GitHub issues* contain documentation, design, bug reports, etc
  - *GitHub pull requests* merge new code into the repository
  - *Git blobs* contain source code
  - *Spotlights* represent cred for contributions that don't otherwise fit
    into the data model, e.g. a spotlight for participtaing in an offline
    design discussion, or for giving a talk on SourceCred

### Edge types:

 - *Authorship edges* connect an issue or pull request with the users that
   created it
 - *MergedAs edges* connect a pull request to the git commit it created
 - *Import edges* connect a Git blob containing source code to the git blobs
   containing code that blob imports
 - *Spotlight edges* connect a Spotlight node to the entities it is flowing cred
   to
 - *Spotlight authorships* flow some cred to the entity that created the spotlight

### The need for weights

As currently implemented, SourceCred does not do a good job of distinguishing
valuable contributions. As an example, when run on the repository in its
current state, it believes that #292 is one of the most valuable pulls in the
project, outshining contributions like #252 and #468 which were actually much
more valuable. Even worse, it massively over-values Git objects; for example,
the blob containing our Prettier configuration is seen as >100x more valuable
than the most valuable pull request.

@wchargin and @decentralion believe that this problem results from the fact
that the graph is currently unweighted. As such, we need to implement a system
for applying weights to different nodes and to the src/dst of edges. There are
several different use cases for weights:

#### Prior over types

Weights can provide prior information on the relative value of different types.
For example, we should have a weight that suggests that GitHub pull requests
are, on average, much more important than Git blobs.

#### Domain-specific evaluation

Weights can use domain-specific signal to differentiate the value of different
nodes of the same type. For example, # of lines of code is a useful signal for
the value of a pull request; a PR that modifies 1000 lines of code is probably
more valuable than one that modifies 3 lines.

#### Distinguish parts of the project

Weights can be used to differentiate different parts of the project for clearer
analysis. For example, suppose a project wants to separately measure cred for
"backend" and "frontend" work, without artificially partioning the project into
[...]

PR #508, opened 2018-07-10
Create `EdgeEvaluator` and `NodeEvaluator` APIs

decentralion:
These APIs are for applying weights to nodes and edges in the graph, so
as to implement weighted PageRank as described in #476. As yet, we
are reasonably confident what the type signatures should be, but do not
know what functions will get the theoretical properties we want.

For now, I've implemented placeholder implementations. These are clearly
deficient in some ways, but at least allow us to start experimenting.

Paired with @wchargin.

Test plan:
No tests were added, as the code modifies prototype parts of our
codebase that are, themselves, not yet tested, and because we already
know these functions have bad semantics.

To ensure that this commit is reasonably correct, I experimented with
changing node/edge weights in the UI and verified that PageRank results
changed as expected.

(for emphasis): #476 has all the context on this PR.

### DIFF
--- a/src/app/credExplorer/App.js
+++ b/src/app/credExplorer/App.js
@@ -10,7 +10,7 @@ import {Graph} from "../../core/graph";
 import {pagerank, type NodeDistribution} from "../../core/attribution/pagerank";
 import {PagerankTable} from "./PagerankTable";
 import type {PluginAdapter} from "../pluginAdapter";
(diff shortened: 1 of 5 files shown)


decentralion:
Closing as this code is going stale, and I think we're likely to go in a different direction, e.g. "WeightRank".

PR #511, opened 2018-07-23
Allow expanding/hiding the WeightConfig

decentralion:
The WeightConfig is a power user feature. Now that we're building a
public-facing demo out of the Cred Explorer, it will be better to hide
the weight configuration by default.

This commit adds a button for showing/hiding the weight configuration.
The weights are still propagated correctly regardless of whether the
weight config is shown.

Test plan:
- Ensure that the site loads with weights hidden by default.
- Ensure that clicking the button causes the weight config to display.
- Ensure that PageRank loads and displays correctly with the weights
hidden.
- Ensure that changes to the weight config still propagate to PageRank
(with weights hidden or not hidden).

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -25,7 +25,6 @@ type Props = {
   onChange: (EdgeEvaluator) => void,
 };
 
-// The key should be an EdgeAddressT, but a Flow bug prevents this.
 type EdgeWeights = Map<EdgeAddressT, UserEdgeWeight>;
 type UserEdgeWeight = {|+logWeight: number, +directionality: number|};
 const EDGE_WEIGHTS_KEY = "edgeWeights";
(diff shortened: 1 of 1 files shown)


src/app/credExplorer/WeightConfig.js
```diff
@@ -90,23 +91,41 @@ export class WeightConfig extends React.Component<Props, State> {
...
-      </div>
+      <React.Fragment>
+        <button
+          onClick={() => {
+            this.setState(({expanded}) => ({
+              expanded: !expanded,
+            }));
+          }}
+        >
+          {expanded ? "Hide Weight Configuration" : "Show Weight Configuration"}
```
wchargin: nit: prefer `Hide weight configuration`, please?

src/app/credExplorer/WeightConfig.js
```diff
@@ -90,23 +91,41 @@ export class WeightConfig extends React.Component<Props, State> {
...
-        />
-        <NodeConfig
-          nodeWeights={this.state.nodeWeights}
-          onChange={(nw) => this.setState({nodeWeights: nw}, () => this.fire())}
-        />
-      </div>
+      <React.Fragment>
+        <button
+          onClick={() => {
+            this.setState(({expanded}) => ({
```
wchargin: No preference from me, but you can (I believe) collapse these three lines if you like.

PR #513, opened 2018-07-23
Split PluginAdapter into Dynamic and Static parts

decentralion:
In some cases (e.g. WeightConfig) we want to have information from the
PluginAdapater before loading any data from the server. In other cases,
we need to combine the PluginAdapater with actual data, e.g. so we can
get the description of a GitHub node.

To support this, we split the PluginAdapter into a Static and Dynamic
component. The Dynamic component has data needed to give node
descriptions, etc. Given a static adapter, you can get a promise to load
the dynamic adapter. Given the dynamic adapter, you can immediately get
the static adapter. (There's a parallel to NodeReference (static) and
NodePorcelain (dynamic)).

Test plan:
Travis passes, as does manual testing of the frontend.

### DIFF
--- a/src/plugins/git/pluginAdapter.js
+++ b/src/plugins/git/pluginAdapter.js
@@ -1,46 +1,22 @@
 // @flow
-import type {PluginAdapter as IPluginAdapter} from "../../app/pluginAdapter";
+import type {
+  StaticPluginAdapter as IStaticPluginAdapter,
+  DynamicPluginAdapter as IDynamicPluginAdapter,
+} from "../../app/pluginAdapter";
 import {Graph} from "../../core/graph";
 import * as N from "./nodes";
 import * as E from "./edges";
 import {description} from "./render";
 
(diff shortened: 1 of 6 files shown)


src/plugins/git/pluginAdapter.js
```diff
@@ -79,4 +55,36 @@ class PluginAdapter implements IPluginAdapter {
       },
     ];
   }
+  async load(
+    repoOwner: string,
+    repoName: string
+  ): Promise<IDynamicPluginAdapter> {
```
wchargin: A client who has specifically a Git plugin static adapter and calls
`load` might hope to receive (a promise for) specifically a Git plugin
dynamic adapter, perhaps to call extra methods on it, like, “who
authored this commit?”. With the current typing, the promise resolves
only to an unspecified plugin dynamic adapter. No change requested; we
can refine this if necessary/useful.

PR #516, opened 2018-07-24
Add a registry of loaded repositories

decentralion:
We want the UI to offer a list of available repositories, rather than
using a text input box. To do this, we first need the backend to include
a registry of all available repositories.

Test plan:
Sadly we don't have CLI testing, so I manually verified this by doing
the following:

```
$ rm -r $SOURCECRED_DIRECTORY
$ node bin/sourcecred.js load sourcecred example-github
$ cat $SOURCECRED_DIRECTORY/repositoryRegistry.json
{"sourcecred/example-github":true}
$ node bin/sourcecred.js load sourcecred example-github
$ cat $SOURCECRED_DIRECTORY/repositoryRegistry.json
{"sourcecred/example-github":true}
$ node bin/sourcecred.js load sourcecred example-git
$ cat $SOURCECRED_DIRECTORY/repositoryRegistry.json
{"sourcecred/example-git":true,"sourcecred/example-github":true}
```

### DIFF
--- a/src/cli/commands/load.js
+++ b/src/cli/commands/load.js
@@ -3,6 +3,8 @@
 import {Command, flags} from "@oclif/command";
 import mkdirp from "mkdirp";
 import path from "path";
+import fs from "fs";
+import stringify from "json-stable-stringify";
 
 import {loadGithubData} from "../../plugins/github/loadGithubData";
 import {loadGitData} from "../../plugins/git/loadGitData";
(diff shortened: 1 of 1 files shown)


src/cli/commands/load.js
```diff
@@ -140,3 +151,22 @@ function loadPlugin({basedir, plugin, repoOwner, repoName, githubToken}) {
       return;
   }
 }
+
+const REPO_REGISTRY_FILE = "repositoryRegistry.json";
```
wchargin: This will need to be used somewhere else, right?

PR #521, opened 2018-07-25
Extract a `LocalStore` interface

wchargin:
Summary:
We’d really like to be able to test components that use `LocalStore`. We
can do this by dependency-injecting the storage backend. This commit
begins that process by extracting `LocalStore` to its interface,
preserving the unique existing implementation.

wchargin-branch: extract-localstore

### DIFF
--- a/src/app/localStore.js
+++ b/src/app/localStore.js
@@ -1,0 +1,26 @@
+// @flow
+
+/**
+ * An abstraction over the browser's `localStorage` API. This enables
+ * persistently storing and retrieving JSON-serializable values by a
+ * string key.
+ */
+export interface LocalStore {
+  /**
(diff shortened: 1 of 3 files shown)


src/app/localStore.js
```diff
@@ -0,0 +1,26 @@
...
+ * An abstraction over the browser's `localStorage` API. This enables
+ * persistently storing and retrieving JSON-serializable values by a
+ * string key.
+ */
+export interface LocalStore {
+  /**
+   * Get the value stored under the given key, or `whenUnavailable` if
+   * this API is unavailable or the key does not exist or is corrupt.
+   */
+  get(key: string, whenUnavailable: any): any;
```
decentralion: Parameterize over T? doesn't need to be in this commit.

PR #526, opened 2018-07-25
Use `CheckedLocalStore` in the cred explorer

wchargin:
Summary:
Might as well have runtime type safety, in case we accidentally try to
store any more `Map`s or `undefined`s.

Test Plan:
Tests pass, but are likely not sufficient. Manual testing indicates that
the local storage still works, for both reads and writes, on a fresh
profile or with existing data, for both the repository owner/name and
the weight configuration.

wchargin-branch: use-checked-local-store


wchargin:
> TestLocalStore = Checked Memory Local Store

Done in #527.

PR #531, opened 2018-07-27
Default to selecting last loaded repository

decentralion:
In #529, I made the cred explorer populate a dropdown with the list of
repositories that are available to explore. That dropdown defaults to
selecting the alphabetically first repository.

This has an unfortunate consequence in that it makes it impossible for
us to explicitly set a default - for example, we would like
sourcecred.github.io/explorer to show sourcecred/sourcecred by default,
but instead it shows example-git.

So that we can choose the default, I've changed the logic so that it
instead shows the most-recently-loaded data first. This required
a breaking change to the repoRegistry serialized format, so I've also
refactored the module to use compat, which I should have done from the
beginning.

Test plan:
Unit tests for the repo selector are updated. The CLI load command
unfortunately has no tests, so I manually tested that it always provides
the lastest repository last, and appropriately handles the case where
the same repository is loaded multiple times.

### DIFF
--- a/src/app/credExplorer/App.js
+++ b/src/app/credExplorer/App.js
@@ -16,7 +16,8 @@ import type {DynamicPluginAdapter} from "../pluginAdapter";
 import {type EdgeEvaluator} from "../../core/attribution/pagerank";
 import {WeightConfig} from "./WeightConfig";
 import type {PagerankNodeDecomposition} from "../../core/attribution/pagerankNodeDecomposition";
-import RepositorySelect, {type Repo} from "./RepositorySelect";
+import RepositorySelect from "./RepositorySelect";
+import type {Repo} from "./repoRegistry";
 
 import * as NullUtil from "../../util/null";
 
--- a/src/app/credExplorer/RepositorySelect.js
+++ b/src/app/credExplorer/RepositorySelect.js
@@ -7,10 +7,8 @@ import deepEqual from "lodash.isequal";
 import * as NullUtil from "../../util/null";
 import type {LocalStore} from "../localStore";
 
-export const REPO_REGISTRY_API = "/api/v1/data/repositoryRegistry.json";
+import {type Repo, fromJSON, REPO_REGISTRY_API} from "./repoRegistry";
 export const REPO_KEY = "selectedRepository";
-
-export type Repo = {|+name: string, +owner: string|};
 
 export type Status =
   | {|+type: "LOADING"|}
@@ -95,17 +93,17 @@ export async function loadStatus(localStore: LocalStore): Promise<Status> {
       return {type: "FAILURE"};
     }
     const json = await response.json();
-    let availableRepos = Object.keys(json).map(repoStringToRepo);
-    availableRepos = sortBy(availableRepos, (r) => r.owner, (r) => r.name);
+    const availableRepos = fromJSON(json);
     if (availableRepos.length === 0) {
       return {type: "NO_REPOS"};
     }
     const localStoreRepo = localStore.get(REPO_KEY, null);
     const selectedRepo = NullUtil.orElse(
       availableRepos.find((x) => deepEqual(x, localStoreRepo)),
-      availableRepos[0]
+      availableRepos[availableRepos.length - 1]
     );
-    return {type: "VALID", availableRepos, selectedRepo};
+    const sortedRepos = sortBy(availableRepos, (r) => r.owner, (r) => r.name);
+    return {type: "VALID", availableRepos: sortedRepos, selectedRepo};
   } catch (e) {
     return {type: "FAILURE"};
   }
--- a/src/app/credExplorer/RepositorySelect.test.js
+++ b/src/app/credExplorer/RepositorySelect.test.js
@@ -10,8 +10,9 @@ import RepositorySelect, {
   loadStatus,
   type Status,
   REPO_KEY,
-  REPO_REGISTRY_API,
 } from "./RepositorySelect";
+
+import {toJSON, type RepoRegistry, REPO_REGISTRY_API} from "./repoRegistry";
 
 require("../testUtil").configureEnzyme();
 require("../testUtil").configureAphrodite();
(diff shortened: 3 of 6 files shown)


src/app/credExplorer/repoRegistry.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+// The repoRegistry is written by the CLI load command
+// (src/cli/commands/load.js) and is read by the RepositorySelect component
+// (src/app/credExplorer/RepositorySelect.js)
+export const REPO_REGISTRY_FILE = "repositoryRegistry.json";
+export const REPO_REGISTRY_COMPAT = {type: "REPO_REGISTRY", version: "0.1.0"};
```
wchargin: Two problems with this:
 1. By requiring clients to call `fromCompat` themselves, but only
    providing the most recent compat information, we prevent clients
    from using the backward compatibility feature of `compat.js`.
 2. With the current typings, a client can overwrite the attributes of
    this object, polluting other clients.

Why not treat this module as a provider of an actual data type? You can
`export type RepoRegistry`, and provide `toCompat` and `fromCompat`. You
could additionally export `emptyRegistry: RepoRegistry`, if you want.
Then, the `REPO_REGISTRY_COMPAT` is an implementation detail.

The two `_FILE` and `_API` strings can continue to exist.

src/cli/commands/load.js
```diff
@@ -153,6 +154,7 @@ function loadPlugin({basedir, plugin, repoOwner, repoName, githubToken}) {
 }
 
 const REPO_REGISTRY_FILE = "repositoryRegistry.json";
+const REPO_REGISTRY_COMPAT = {type: "REPO_REGISTRY", version: "0.1.0"};
```
wchargin: Dupe?

src/app/credExplorer/RepositorySelect.test.js
```diff
@@ -157,49 +153,37 @@ describe("app/credExplorer/RepositorySelect", () => {
...
+      mockRegistry(["foo/bar", "a/b", "a/z"]);
       const repos = [
         {owner: "a", name: "b"},
         {owner: "a", name: "z"},
         {owner: "foo", name: "bar"},
       ];
       const localStore = testLocalStore();
       localStore.set(REPO_KEY, 42);
-      return expectLoadValidStatus(fetchResult, localStore, repos, repos[0]);
+      return expectLoadValidStatus(localStore, repos, repos[1]);
```
wchargin: Why did you s/0/1/

decentralion:
Thanks for the feedback. I've refactored the code per your suggestion; it's cleaner and better tested (we now test the logic that adds new entries to the repository registry).

src/app/credExplorer/App.js
```diff
@@ -16,7 +16,8 @@ import type {DynamicPluginAdapter} from "../pluginAdapter";
 import {type EdgeEvaluator} from "../../core/attribution/pagerank";
 import {WeightConfig} from "./WeightConfig";
 import type {PagerankNodeDecomposition} from "../../core/attribution/pagerankNodeDecomposition";
-import RepositorySelect, {type Repo} from "./RepositorySelect";
+import RepositorySelect from "./RepositorySelect";
+import {type Repo} from "./repoRegistry";
```
wchargin: Slightly better to `import type {Repo} from "./repoRegistry"`, which
will have no runtime effect.

src/app/credExplorer/RepositorySelect.js
```diff
@@ -95,17 +93,17 @@ export async function loadStatus(localStore: LocalStore): Promise<Status> {
       return {type: "FAILURE"};
     }
     const json = await response.json();
-    let availableRepos = Object.keys(json).map(repoStringToRepo);
-    availableRepos = sortBy(availableRepos, (r) => r.owner, (r) => r.name);
+    const availableRepos = fromJSON(json);
```
wchargin: (Here is another place where it’s nice that we wrote `fromJSON`/`toJSON`
instead of forcing clients to work with strings.)

src/app/credExplorer/repoRegistry.js
```diff
@@ -0,0 +1,32 @@
...
+import deepEqual from "lodash.isequal";
+import {toCompat, fromCompat, type Compatible} from "../../util/compat";
+
+export const REPO_REGISTRY_FILE = "repositoryRegistry.json";
+export const REPO_REGISTRY_API = "/api/v1/data/repositoryRegistry.json";
+
+const REPO_REGISTRY_COMPAT = {type: "REPO_REGISTRY", version: "0.1.0"};
+
+export type Repo = {|+name: string, +owner: string|};
+export type RepoRegistry = Repo[];
```
wchargin: Should this be a `$ReadOnlyArray`?

src/app/credExplorer/repoRegistry.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+// The repoRegistry is written by the CLI load command
+// (src/cli/commands/load.js) and is read by the RepositorySelect component
+// (src/app/credExplorer/RepositorySelect.js)
+import deepEqual from "lodash.isequal";
+import {toCompat, fromCompat, type Compatible} from "../../util/compat";
+
+export const REPO_REGISTRY_FILE = "repositoryRegistry.json";
+export const REPO_REGISTRY_API = "/api/v1/data/repositoryRegistry.json";
```
wchargin: The value of this const isn’t really clear to me. Really, it seems like
`REPO_REGISTRY_API` should be `sourcecredDataUrl(REPO_REGISTRY_FILE)`,
where ``sourcecredDataUrl = (x) => `/api/v1/data/${x}` `` is used more
generally throughout the application, perhaps as part of some router.

I’m not totally sure what the right way to factor this is right now. The
current code seems okay; it would also be fine with me to delete this
export. What do you think?

src/app/credExplorer/repoRegistry.js
decentralion: Done.

src/app/credExplorer/App.js
decentralion: Done.

src/app/credExplorer/repoRegistry.js
decentralion: I agree that in the future we should have a `sourcecredDataUrl` method or similar and use the factorization you suggest. In the absence of that API, keeping the API const next to the FILE const is slightly nice, in that if you change the FILE const it's easy to find the corresponding API const that needs to change.

PR #532, opened 2018-07-27
Import the sourcecred.github.io deploy script

decentralion:
This copies the deploy script from sourcecred/sourcecred.github.io into
the sourcecred/sourcecred repository. It's been modified to reflect the
fact that it is no longer run from within the sourcecred.github.io
repository.

We’re moving this script into the main SourceCred repository so that it
can be versioned in the same tree as everything else, and in particular
so that the hash-of-built-commit depends on the build script itself.

The script was originally created by @wchargin in
https://github.com/sourcecred/sourcecred.github.io/pull/1 at hash
e8bc5b669b31df34b439b194cb62af4fd9a789d8, subsequently modified in
https://github.com/sourcecred/sourcecred.github.io/pull/2 and
https://github.com/sourcecred/sourcecred.github.io/pull/3 and
https://github.com/sourcecred/sourcecred.github.io/pull/4

Paired with @wchargin

Test plan: Ran locally (on dry-run) and inspected output.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,0 +1,146 @@
+#!/bin/sh
+set -eu
+
+: "${SOURCECRED_REMOTE:=git@github.com:sourcecred/sourcecred.git}"
+: "${SOURCECRED_REF:=origin/master}"
+
+: "${DEPLOY_REMOTE:=git@github.com:sourcecred/sourcecred.github.io.git}"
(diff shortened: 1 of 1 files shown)


wchargin:
Could you please add to the commit message:

```
We’re moving this script into the main SourceCred repository so that it
can be versioned in the same tree as everything else, and in particular
so that the hash-of-built-commit depends on the build script itself.
```

Issue #350, opened 2018-06-06
Investigate nondeterministic GitHub query rejections


wchargin:
In #349, we saw what appeared to be a bug in continuation resolution
when run on `ipfs/go-ipfs`:

```
TypeError: Cannot read property '_n0' of undefined
    at embeddings.reduce (/home/wchargin/git/sourcecred/bin/commands/plugin-graph.js:442:153)
    at Array.reduce (<anonymous>)
    at resolveContinuations (/home/wchargin/git/sourcecred/bin/commands/plugin-graph.js:442:82)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

I’ve investigated this. It is not a bug in our code, per se. What is
happening is the following. Sometimes, when we post a query to the
GitHub GraphQL API, the response is (prettified):

```
{
  "documentation_url": "https://developer.github.com/v3/#abuse-rate-limits",
  "message": "You have triggered an abuse detection mechanism. Please wait a few minutes before you try again."
}
```

Note that this links to the v3 API, even though we’re using the v4 API.
Also note that we are not using the API to rapidly create content,
polling (at all), or making multiple concurent requests, and it seems
surprising that the data that we are requesting should be
computationally expensive (e.g., we are not forcing computation of Git
diffs; we’re just fetching standard fields off the standard objects).

First of all, this is not a [valid GraphQL response][1], so it is not
surprising that we do not have logic to handle it.

[1]: https://facebook.github.io/graphql/October2016/#sec-Response-Format

Second, whether this error is produced appears to be nondeterministic.
Running the same sequence of queries multiple times can lead to exactly
the same results (failures and all), or it can be the case that the
first sequence fails and the second succeeds. One potential explanation
for this is that the abuse detection depends on actual compute resources
used by GitHub’s servers, and so initial queries cause the caches to be
hot for the particular resources in question. Or, it could just be a
fluke.

We should probably, nevertheless, attempt to fix this. For instance, we
could detect this error and, as the message suggests, wait some time
before re-issuing the query.

wchargin:
Useful debugging/investigation tool: https://gist.github.com/wchargin/a6336f5730d0d09a25562b83120d1a95

decentralion:
[...]

PR #536, opened 2018-07-27
Improve error messages when GitHub query fails

decentralion:
Currently, the GitHub graph fetcher will characteristically fail if:
1. it times out GitHub's server
2. it triggers the semidocumented abuse detection mechanism

In case 1, an intelligible error is posted to the console. In case 2, it
produces an unintelligible TypeError, because the response is not a
valid GraphQL response (the error field is not populated; it has a
custom message instead).

As of this commit, we gracefully catch both cases, and print a message
to console directing the user to #350, which has context on GitHub query
failures. This new catch works because in case 2, the data field is
empty, so we now properly recognize `x.data == null` as an error case.

Thanks to @wchargin for the investigatory work behind this commit.

Test plan:
We don't have unit tests that cover this case, but I did manually test
it by asking GitHub to fetch `ipfs/go-ipfs`, which consistently fails.


src/plugins/github/fetchGithubRepo.js
```diff
@@ -74,7 +74,15 @@ function postQuery({body, variables}, token) {
   })
     .then((x) => x.json())
     .then((x) => {
-      if (x.errors) {
+      if (x.errors || x.data == null) {
```
wchargin: Could we check `x.data === undefined` or `!("data" in x)` instead? It is
generally possible for the result of a valid query to be exactly `null`.

wchargin:
> we now properly recognize `x.data == null` as an error case.

I feel compelled to grumble that _our_ code was perfectly correct. [The
GraphQL spec clearly states][spec] that a GraphQL response must contain
a key `errors` if there were any errors, and also clearly states that
the top level response must not contain any entries other than those
with keys `data`, `errors`, and `extensions`. It is GitHub whose code is
“improper”, not us. `</grumble>`

[spec]: https://facebook.github.io/graphql/October2016/#sec-Response-Format

src/plugins/github/fetchGithubRepo.js
decentralion: Done.

PR #546, opened 2018-07-27
Upgrade Flow to v0.76.0

wchargin:
Summary:
In addition to a routine libdef update, we also need to work around a
particularly nasty new bug in Flow, which requires `any`-casts that are
even more unsafe than usual. That said, I think that it’s worth that
cost to remain up to date with Flow, so that we can amortize future such
issues.

Test Plan:
Running `yarn travis --full` passes.

wchargin-branch: upgrade-flow-v0.76.0


wchargin:
Relevant Flow bugs:
  - https://github.com/facebook/flow/issues/6648
  - https://github.com/facebook/flow/issues/6649

PR #552, opened 2018-07-28
UI Update: adding navbar

claireandcode:


decentralion:
Some requests based on how we do things 'round here:
- We always have pull descriptions that give context on this change. Since it's a UI change, adding before/after screenshots would be nice.
- We always have test plans for our pulls. Since it's a UI change, "Load the app and observe that the nav bar is pretty, and the links work" is a perfectly valid test plan.
- To fix the travis failure, add `// @flow` to the top of `src/app/style/styles.js`

src/app/Page.js
```diff
@@ -11,15 +13,75 @@ export default class Page extends React.Component<{|+children: Node|}> {
     return (
       <div>
         <header>
-          <nav>
-            <ul style={{listStyle: "none", paddingLeft: 0, margin: 0}}>
+          <nav className={css(style.nav)}>
```
wchargin: Styles like `nav` don’t need to be used by any other modules—and indeed
should not be. Could you define them locally instead of in a public
module? You can just inline them into the `StyleSheet.create` call.

If you have styles that you expect to be used by other modules, then I’m
tentatively okay with those being exposed publicly, but I think that it
might be better practice to just have a dedicated component. For
instance, the `header` style is arguably reusable, but it might be
better to just create a `<Header />` component that encapsulates the
appropriate styles.

src/app/Page.js
```diff
@@ -11,15 +13,75 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+            <ul className={css(style.navList)}>
+              <li className={css([style.navItem, style.navItemLeft])}>
```
wchargin: nit: I believe that this is equivalent to

```js
css(style.navItem, style.navItemLeft)
```

(i.e., without the array). If they’re equivalent, I have a weak
preference toward removing the array, because it makes the single-style
case more consistent with the multi-style case, which in turn minimizes
churn.

src/app/Page.js
```diff
@@ -11,15 +13,75 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+                  <svg
+                    className={css([style.ghLogo, style.ghLogoHover])}
+                    viewBox="0 0 1024 1024"
+                    xmlns="http://www.w3.org/2000/svg"
+                  >
+                    <path d="M512 0C229.25 0 0 229.25 0 512c0 226.25 146.688 418.125 350.156 485.812 25.594 4.688 34.938-11.125 34.938-24.625 0-12.188-0.469-52.562-0.719-95.312C242 908.812 211.906 817.5 211.906 817.5c-23.312-59.125-56.844-74.875-56.844-74.875-46.531-31.75 3.53-31.125 3.53-31.125 51.406 3.562 78.47 52.75 78.47 52.75 45.688 78.25 119.875 55.625 149 42.5 4.654-33 17.904-55.625 32.5-68.375C304.906 725.438 185.344 681.5 185.344 485.312c0-55.938 19.969-101.562 52.656-137.406-5.219-13-22.844-65.094 5.062-135.562 0 0 42.938-13.75 140.812 52.5 40.812-11.406 84.594-17.031 128.125-17.219 43.5 0.188 87.312 5.875 128.188 17.281 97.688-66.312 140.688-52.5 140.688-52.5 28 70.531 10.375 122.562 5.125 135.5 32.812 35.844 52.625 81.469 52.625 137.406 0 196.688-119.75 240-233.812 252.688 18.438 15.875 34.75 47 34.75 94.75 0 68.438-0.688 123.625-0.688 140.5 0 13.625 9.312 29.562 35.25 24.562C877.438 930 1024 738.125 1024 512 1024 229.25 794.75 0 512 0z" />
```
wchargin: Storing the SVG data inline in this module is fine with me. However,
could you extract React components `GithubLogo` and `TwitterLogo`, so
that the main body of the code can just read

```js
              <li className={css([style.navItem, style.navItemRight])}>
                <a
                  className={css([style.navLink, style.navLinkHover])}
                  href="https://github.com/sourcecred/sourcecred"
                >
                  <GithubLogo />
                </a>
              </li>
              <li className={css([style.navItem, style.navItemRight])}>
                <a
                  className={css([style.navLink, style.navLinkHover])}
                  href="https://twitter.com/sourcecred"
                >
                  <TwitterLogo />
                </a>
              </li>
```

I suspect that this will make the code much easier to read, in the same
way that extracting styles to an Aphrodite sheet does.

decentralion:
On hover, the navbar links get a darker font color, and an underline.
When I click the link, the style stays (even after I move the mouse away), only reverting once I've clicked somewhere else on the page.

This feels like a small bug to me (it feels like the hover is lagging or something). Can we fix that? (Repros in both Chrome and FF)

src/app/style/styles.js
```diff
@@ -0,0 +1,67 @@
+export default {
```
wchargin: This module exports a piece of global mutable state, which is something
that we try hard to avoid. If you decide to keep this module, could you
consider instead exporting a function to create this sheet definition?

src/app/style/styles.js
```diff
@@ -0,0 +1,67 @@
...
+  header: {
+    color: "#1297A1",
+    fontFamily: "Roboto Condensed",
```
wchargin: Do you want this to fall back to `sans-serif`?

src/app/style/styles.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+  body: {
+    padding: "0 100px",
+    flex: "3",
```
wchargin: You can use numeric literals instead of string literals in a bunch of
places: `{flex: 3}`, `{height: 60}`, `{paddingLeft: 0}`, etc. My
understanding is that this is the preferred style.

yarn.lock
```diff
@@ -16,6 +16,13 @@
     esutils "^2.0.2"
     js-tokens "^3.0.0"
 
+"@babel/runtime@^7.0.0-beta.49":
```
wchargin: It appears that `semantic-ui-react` requires Babel at runtime (!!!). I’m
still trying to dig into exactly how they use it, but this is deeply
disturbing. Depending on what they’re doing, it could cripple the
performance of our application. I’ll get back to you on this.

yarn.lock
```diff
@@ -4545,6 +4560,10 @@ jest@23.4.1:
     import-local "^1.0.0"
     jest-cli "^23.4.1"
 
+jquery@x.*:
```
wchargin: I’m also concerned by the dependency on jQuery. My understanding was
that the Semantic UI React family didn’t use jQuery at all. Do you know
what’s going on here?

src/app/Page.js
```diff
@@ -11,15 +13,75 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+                      to={path}
+                      className={css([style.navLink, style.navLinkHover])}
+                    >
+                      {navTitle}
+                    </Link>
                   </li>
                 ))
               )}
+              <li className={css([style.navItem, style.navItemRight])}>
+                <a
```
wchargin: These links are missing `alt`-text.

src/app/style/styles.js
```diff
@@ -0,0 +1,67 @@
...
+    display: "inline-block",
+  },
+  navList: {
+    listStyle: "none",
+    paddingLeft: "0",
+    margin: "0",
+    display: "flex",
+  },
+  navLink: {
+    color: "#3EBEC7",
```
wchargin: This color `#3ebec7` against white has a contrast ratio of 2.24, but at
the size at which it appears the minimum contrast ratio is 4.5. Could
you please increase the contrast to meet the accessibility guidelines?
(tota11y suggests using `#00848b`.)

src/app/style/index.css
```diff
@@ -0,0 +1,7 @@
+@import '~semantic-ui-css/semantic.min.css';
```
wchargin: This has the side effect of including 624KB\* of styles before the
content of the page. This means that no content can be rendered until
we’ve successfully downloaded, parsed, and interpreted the CSS necessary
to set up the sprite sheets for the flag of Christmas Island. This will
incur significant latency on lower-end connections and hardware, so it’s
definitely something that we want to avoid.

The long-term solution is to simply include the stylesheet by URL
instead of inlining it. (Inlining our previous version of `index.css`
was okay because that stylesheet was tiny by design, and was not
intended to grow.) I’m tentatively okay with merging this as is if you
can add a TODO and make an issue for it. What do you think?

\* measured with:

```
curl --silent http://localhost:4000/ | sed '0,/<style>/d; /^<body /q' | wc -c
```

src/app/routeData.js
```diff
@@ -26,7 +26,7 @@ const routeData /*: $ReadOnlyArray<RouteDatum> */ = [
     path: "/explorer",
     component: () => require("./credExplorer/App").default,
     title: "SourceCred explorer",
-    navTitle: "Explorer",
+    navTitle: "Explore",
```
wchargin: Interesting change!

wchargin:
This renders differently for me [in prod] vs. [in dev]. In prod,
headings are rendered in Lato and body is rendered in Roboto. In dev,
everything is rendered in my default sans-serif. Do you know why this
is? Do you have the same issue?

By “in prod” I mean `node bin/sourcecred.js start` after `yarn build`.
By “in dev” I mean `yarn start`.

[in prod](https://user-images.githubusercontent.com/4317806/43351927-e616ca98-91ce-11e8-9d5f-40aac738586c.png)
[in dev](https://user-images.githubusercontent.com/4317806/43351928-e87a4d28-91ce-11e8-906b-3094800c9a63.png)

src/app/style/styles.js
decentralion: Sad, I liked that teal.

wchargin:
Regarding @decentralion’s review:

> I'm basically asking for the :focus on the navlinks to be removed.

This would be fine with me. We have the default outline style (which is
good!), so the links are still keyboard-accessible even without the
added underline. (I think—correct me if I’m wrong.)

> But it might be nice to have the nav entry corresponding to the
> currently selected page be rendered bold (which would also make the
> link feel reactive to being clicked on)

:+1:

wchargin:
Let me take a step back to pose a high-level question.

I agree that standardizing on a UI toolkit is a good idea and something
that we should do. We don’t want to roll our own custom UI widgets,
because then we have to make sure that they’re accessible,
high-performance, consistently designed, and so on.

Whichever toolkit we select will quickly permeate our frontend, and the
decision may be hard to change later. Thus, we should choose carefully.

I’m concerned by the bundle size and build time penalties incurred by
Semantic UI:

| build ref        | `main.js` size |    build time |
|:-----------------|---------------:|--------------:|
| origin/master    |      439 646 B |      19.112 s |
| origin/ui-update |    1 408 500 B |      39.077 s |

Semantic UI is adding 968.8 KB to the build, and more than doubling the
build time. For comparison, [all of Material-UI is 95.4 KB][mui] (and
Material-UI supports directly importing individual components), and
[React itself (16.2.0 + DOM) is 97.5 KB][react].

(I built against origin/ui-update, not origin/update-navbar, because
this branch doesn’t actually use `semantic-ui-react`.)

[mui]: https://github.com/mui-org/material-ui/blob/0efc81dad607437fa448d6b19a2c985266790848/.size-limit.js#L27-L30
[react]: https://gist.github.com/Restuta/cda69e50a853aa64912d

So, I’d like to consider other options. Material-UI seems well supported
and maintained. React-Bootstrap is fine (I’ve used it in the past), if a
Bootstrap-like look is what you’re going for, and Elemental UI looks
pretty similar. What else should we be considering?

I also want to explicitly mention that I value the time and effort that
you’ve put into this set of changes. In the long run, though, I think
that it would be a mistake to base an architectural decision like this
one around the fact that we already have a bit of code written. Plus,
even if we end up scratching the technical implementation of this
changeset entirely, your work won’t have been wasted—in addition to the
actual code, you’ve spent time exploring different designs, and become
acquainted with the codebase, tools, and environment. All of that will
carry over.

@decentralion: anything I forgot to mention?

---

Another option that I want to float (though I suspect that it may be an
unpopular suggestion) is to use the unique UI toolkit that adds zero
build time and zero bundle size. You can go a long way by just styling
the standard DOM elements. I’ve used that approach in the past for [this
application][vt], which doesn’t look half bad, if you ask me. Choosing
to use the standard library is different from rolling your own
components: you get accessibility, performance, etc. for free.

I’m not saying that we should _definitely_ do this, but I do think that
we should consider alternatives in terms of their marginal benefits over
using the standard library.

[vt]: https://wchargin.github.io/varese/#/tree

src/app/GithubLogo.js
```diff
@@ -0,0 +1,22 @@
...
+
+type Props = {
+  className: string,
+  altText: string,
+};
+
+export default function GithubLogo(props: Props) {
+  return (
+    <svg
+      aria-labelledby="ghLogoTitle"
```
wchargin: Question from an SVG-noob: is it possible/okay to instead write

```jsx
<svg aria-label={props.altText} …>
  <title>{props.altText}</title>
  {/* … */}
</svg>
```

? I’m generally quite reluctant to use `id`s unless absolutely
necessary, because `id`s create global state on the page and thus break
encapsulation.

For instance, as the component is currently written, I can render

```jsx
<div>
  <GithubLogo className="" altText="Title the first" />
  <GithubLogo className="" altText="Title the second" />
</div>
```

and each SVG will have the correct _hover_ text, but both SVGs appear as
“Title the first” to screenreaders (according to tota11y, at least).

src/app/GithubLogo.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import React from "react";
+
+type Props = {
+  className: string,
+  altText: string,
+};
+
+export default function GithubLogo(props: Props) {
```
wchargin: Ah, interesting. Exporting these as separate files is totally fine with
me: they seem nice and self-contained. If you prefer to include them in
the `Page` module only, that’s also perfectly fine. (As you may have
noticed, we don’t maintain any “one component per file” rule.)

src/app/Page.js
```diff
@@ -28,3 +72,58 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+    margin: 0,
+    display: "flex",
+  },
+  navLink: {
+    color: "#00848B",
+    fontFamily: "Roboto Condensed",
+    fontSize: 18,
+    textDecoration: "none",
+  },
+  navLinkHover: {
```
wchargin: Is there a reason that we would want to style a link with `navLink` but
not `navLinkHover`? If not, you can include the hover state as part of
the main link style:

```js
  navLink: {
    color: "#00848B",
    fontFamily: "Roboto Condensed",
    fontSize: 18,
    textDecoration: "none",
    ":hover": {
      color: "#08A27F",
      textDecoration: "underline",
    },
  },
```

src/app/GithubLogo.js
claireandcode: I was basing it off of this guide: https://css-tricks.com/accessible-svgs/
But after further research, looks like `aria-label` should be fine.

src/app/GithubLogo.js
claireandcode: I generally tend towards one-component-per-file, because I like being able to scan files for exactly what I'm looking for :)

src/app/Page.js
claireandcode: 👍

src/app/Page.js
claireandcode: 👍 makes sense to me

decentralion:
This looks good to me! 
I'm going to let @wchargin give the official approval as I think he's a better reviewer for this code

src/app/Page.js
```diff
@@ -11,15 +13,47 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+                  >
+                    <Link to={path} className={css(style.navLink)}>
+                      {navTitle}
+                    </Link>
                   </li>
                 ))
               )}
+              <li className={css(style.navItem, style.navItemRight)}>
+                <a
+                  className={css(style.navLink, style.navLinkHover)}
```
wchargin: This `navLinkHover` now evaluates to `undefined` and can be removed.
Same below.

src/app/Page.js
```diff
@@ -28,3 +62,54 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+    margin: 0,
+    display: "flex",
+  },
+  navLink: {
+    color: "#00848B",
+    fontFamily: "Roboto Condensed",
+    fontSize: 18,
+    textDecoration: "none",
+    ":hover": {
+      color: "#08A27F",
```
wchargin: fyi: This color `#08a27f` against white has a contrast ratio of just
3.24, which is technically too low for the font size at which it’s
rendered (minimum is 4.5). However, I’m okay with it given that it
appears only as a hover style, and that the main link style does have
sufficient contrast. Up to you whether to take any action.

src/app/Page.js
```diff
@@ -28,3 +62,54 @@ export default class Page extends React.Component<{|+children: Node|}> {
...
+  navItemRight: {
+    marginLeft: 20,
+  },
+  logoImg: {
+    height: 20,
+  },
+  navLogo: {
+    height: 20,
+    width: 20,
+    fill: "#00848B",
```
wchargin: Please consider omitting these color declarations and instead
constructing your SVGs with `<path fill="currentColor" d={…} />`. One
advantage of this is that you only specify the colors in one location.
More importantly, this change would fix a bug in the current
implementation, wherein the `a` can be hovered even if the `svg` is not
(the `a` descends 4px below the baseline due to line-height), which
results in inconsistent styling: the cursor is `pointer` but the logo
does not appear hovered.

src/app/Page.js
```diff
@@ -28,3 +62,54 @@ export default class Page extends React.Component<{|+children: Node|}> {
     );
   }
 }
+
+const style = StyleSheet.create({
+  header: {
```
wchargin: Unused styles: `header`, `body`, `logoImg`. Can remove.

src/app/Page.js
claireandcode: I've changed where the fill is getting applied from to fix this- I wanted to keep fill out of the component itself to enable the hover color change.

src/app/Page.js
claireandcode: 👍 I've swapped the colors again to ones with better contrast

src/app/Page.js
claireandcode: 👍

PR #555, opened 2018-07-28
Create a Repo type and use throughout the project

decentralion:
Our data model orients on getting repos from GitHub, which are
alternatively represented as strings like "sourcecred/sourcecred", or
pairs of variables representing the owner and name, or objects with
owner and name properties. We also have a few different implementations
of repo validation, which are not applied consistently.

This commit changes all that. We now have a consistent Repo type which
is an object containing a string owner and string name. Thanks to a
clever suggestion by @wchargin, it is implemented as an opaque subtype
of an object containing those properties, so that the only valid way to
construct a Repo typed object is to use one of the functions that
consistently validates the repo.

As a fly-by fix, I noticed that there were some functions in the GitHub
query generation that didn't properly mark arguments as readOnly. I've
fixed these.

Test plan: No externally-observable behavior changes (except insofar as
there is a slight change in variable names in the GitHub graphql query,
which has also resulted in a snapshot diff). `yarn travis --full`
passes. `git grep repoOwner` presents no hits.

### DIFF
--- a/src/core/repo.test.js
+++ b/src/core/repo.test.js
@@ -1,0 +1,51 @@
+// @flow
+
+import {makeRepo, stringToRepo, repoToString, type Repo} from "./repo";
+
+describe("core/repo", () => {
+  describe("Repo type", () => {
+    it("manually constructing a Repo is illegal", () => {
+      // $ExpectFlowError
+      const _unused_repo: Repo = {owner: "foo", name: "bar"};
+    });
+    it("destructuring repo properties is legal", () => {
+      const repo: Repo = makeRepo("foo", "bar");
+      const _unused_owner: string = repo.owner;
+      const _unused_name: string = repo.name;
+    });
+  });
+  describe("makeRepoRepo", () => {
+    it("allows a simple repo", () => {
+      makeRepo("sourcecred", "sourcecred");
+    });
+    it("allows a repo with periods in name", () => {
+      makeRepo("sourcecred", "sourcecred.github.io");
+    });
+    it("allows a repo with hyphens", () => {
+      makeRepo("foo", "something-good");
+    });
+    it("disallows a repo with no owner", () => {
+      expect(() => makeRepo("", "foo")).toThrow("Invalid");
+    });
+    it("disallows a repo with no name", () => {
+      expect(() => makeRepo("foo", "")).toThrow("Invalid");
+    });
+    it("disallows a repo with underscores", () => {
+      expect(() => makeRepo("yep", "something_bad")).toThrow("Invalid");
+    });
+  });
+  describe("repo<->string", () => {
+    function testInvertible(owner, name) {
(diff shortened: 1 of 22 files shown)


src/core/repo.test.js
```diff
@@ -0,0 +1,51 @@
...
+      makeRepo("sourcecred", "sourcecred.github.io");
+    });
+    it("allows a repo with hypens", () => {
+      makeRepo("foo", "something-good");
+    });
+    it("disallows a repo with no owner", () => {
+      expect(() => makeRepo("", "foo")).toThrow("Invalid");
+    });
+    it("disallows a repo with no name", () => {
+      expect(() => makeRepo("", "")).toThrow("Invalid");
```
wchargin: Even if the repo name weren’t validated, wouldn’t this fail because the
owner is empty?

src/core/repo.test.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+    it("disallows a repo with underscores", () => {
+      expect(() => makeRepo("", "something_bad")).toThrow("Invalid");
```
wchargin: Even if the underscore case weren’t validated, wouldn’t this fail
because the owner is empty?

src/core/repo.test.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+  });
+  describe("makeRepoRepo", () => {
+    it("allows a simple repo", () => {
+      makeRepo("sourcecred", "sourcecred");
+    });
+    it("allows a repo with periods in name", () => {
+      makeRepo("sourcecred", "sourcecred.github.io");
+    });
+    it("allows a repo with hypens", () => {
```
wchargin: hyphens

src/core/repo.test.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+    it("disallows a repo with no name", () => {
+      expect(() => makeRepo("", "")).toThrow("Invalid");
+    });
+    it("disallows a repo with underscores", () => {
+      expect(() => makeRepo("", "something_bad")).toThrow("Invalid");
+    });
+  });
+  describe("repo<->string", () => {
+    function testCommutes(owner, name) {
```
wchargin: nit: you’re testing that the functions are inverses, not that they
commute (which doesn’t make sense in this case)

src/plugins/github/graphql.js
```diff
@@ -820,6 +821,6 @@ export function createFragments(): FragmentDefinition[] {
   ];
 }
 
-export function createVariables(repoOwner: string, repoName: string) {
-  return {repoOwner, repoName};
+export function createVariables(repo: Repo): Repo {
```
wchargin: I think that the type that you want here is actually `{+[string]: any}`
or something like that. The fact that the variables are exactly equal to
the fields of `Repo` is a coincidence/implementation detail that may
change over time.

PR #557, opened 2018-07-28
Make deploy script wait for valid response

decentralion:
Currently, any non-yes keystroke causes the deply script to abort.
This is frustrating, as it may take the user several minutes of waiting
to get to the prompt, only to have the program inadvertently terminate.

As of this change, the deploys script patiently re-prompts until it gets
a valid input.

Paired with @wchargin.

Test plan: Ran locally and tested the yes, no, and invalid response
cases. Also ran the script through shellcheck.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -113,8 +113,11 @@ build_and_deploy() {
     printf '\n'
     printf 'Please review the build output now---run:\n'
     printf '    cd "%s" && python -m SimpleHTTPServer\n' "${preview_dir}"
-    printf 'Do you want to deploy? yes/no> '
-    read -r line
+    line=
+    while [ "${line}" != yes ] && [ "${line}" != no ]; do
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -113,8 +113,13 @@ build_and_deploy() {
     printf '\n'
     printf 'Please review the build output now---run:\n'
     printf '    cd "%s" && python -m SimpleHTTPServer\n' "${preview_dir}"
-    printf 'Do you want to deploy? yes/no> '
-    read -r line
+    while true; do
```
wchargin: Alternately,

```shell
    line=
    while ! [ "${line}" = yes ] || [ "${line}" = no ]; do
        printf 'Do you want to deploy? yes/no> '
        read -r line
    done
```

which is more readable IMO.

decentralion: Done (in modified form), PTAL

PR #559, opened 2018-07-28
CLI takes repo strings as `owner/name`

decentralion:
Test plan:
```
$ yarn backend
$ node bin/sourcecred.js load sourcecred/sourcecred
```

### DIFF
--- a/src/cli/commands/load.js
+++ b/src/cli/commands/load.js
@@ -14,7 +14,7 @@ import {
   sourcecredDirectoryFlag,
 } from "../common";
 
-import {makeRepo} from "../../core/repo";
+import {repoToString, stringToRepo} from "../../core/repo";
 
 import {
   toJSON,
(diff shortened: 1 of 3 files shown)


src/cli/commands/load.js
```diff
@@ -31,14 +31,9 @@ export default class PluginGraphCommand extends Command {
...
-      name: "repo_owner",
+      name: "repo",
       required: true,
-      description: "owner of the GitHub repository for which to fetch data",
-    },
-    {
-      name: "repo_name",
-      required: true,
-      description: "name of the GitHub repository for which to fetch data",
+      description: "the GitHub repo to load, represented as $OWNER/$NAME. ",
```
wchargin: Let’s write `OWNER/NAME` instead of `$OWNER/$NAME`; the former is
standard man-page syntax.

Let’s also remove the trailing `". "` from this string, for consistency.

decentralion: Done.

PR #574, opened 2018-07-31
Write `CONTRIBUTING.md`

wchargin:
Summary:
Having a guide for contributors makes the experience more pleasant for
everyone. In the best case, contributors read the guide and submit
perfect PRs. In the event that a discrepancy occurs, we can gently point
the contributor to the guide—saving us from having to re-explain common
pitfalls, and signaling to the contributor that this is a matter of
practice and that we are not “out to get them”.

This guide, like all documentation, is intended to evolve over time.

Test Plan:
None.

wchargin-branch: contributing-md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,271 @@
+# Contributing
+
+Contributions are appreciated! If you’d like to contribute, please help
+us out by reading this short document.
+
+## Before writing code
+
+If you’re interested in adding a new feature, consider opening an issue
+suggesting it. Then, we can discuss the feature to make sure that
+everyone is on board.
+
+If you’re interested in helping but don’t know where to start, consider
+looking at open issues. [Issues marked **contributions-welcome**][cw]
+may be especially accessible to newcomers.
+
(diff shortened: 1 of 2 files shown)


wchargin:
I considered adding a section, “During pull request review”, but I
wasn’t sure that there was anything useful to add there. Standard advice
of “we’re trying to merge the best code that we can, not judging you as
a person” and “you and the maintainers are on the same team” applies,
but seems… almost self-evident? I’m not sure. Happy to add content there
if folks desire.

CONTRIBUTING.md
```diff
@@ -0,0 +1,271 @@
...
+
+Here is an example of a helpful commit message. [The commit in
+question][example-commit] doesn’t change very many lines of code, but
+the commit message explains the context behind the commit, links to
+relevant issues, thanks people who contributed to the commit, and
+describes a manual test plan. Someone reading this commit for the first
+time will have a much better understanding of the change by reading this
+commit message:
+
+```
```
wchargin: Also happy to format this as a block-quote instead of code.

wchargin: I looked into this a bit, and prefer it as is. Formatting as a block
quote changes the wrapping, the list formatting, and the inline code
formatting, making it harder for readers to understand what was actually
written without looking at the source.

PR #577, opened 2018-07-31
Refine the build directory cleaning logic

wchargin:
Summary:
We were asking the `clean-webpack-plugin` to remove the `build/`
directory in all cases. However, Webpack accepts a command-line
parameter `--output-path`. When such a parameter is passed, we would be
removing the wrong directory.

The proper behavior is to remove “whatever the actual output path is”.
Webpack exposes this information, but it appears that the
`clean-webpack-plugin` does not take advantage of it. Therefore, this
commit includes a small Webpack plugin to do the right thing.

Test Plan:
Test that the behavior is correct when no output directory is specified:
```
mkdir -p build && touch build/wat && yarn build && ! [ -e build/wat ]
```

Test that the behavior is correct with an explicit `--output-path`:
```
outdir="$(mktemp -d)" && touch "${outdir}/wat" && \
    yarn build --output-path "${outdir}" && \
    ! [ -e "${outdir}/wat" ]
```

wchargin-branch: clean-actual-build-directory

### DIFF
--- a/config/RemoveBuildDirectoryPlugin.js
+++ b/config/RemoveBuildDirectoryPlugin.js
@@ -1,0 +1,39 @@
+// @flow
+
+const fs = require("fs");
+const path = require("path");
+const rimraf = require("rimraf");
+
+// Note: the following type-import just resolves to `any`.
+/*:: import type {Compiler} from "webpack"; */
+
+module.exports = class RemoveBuildDirectoryPlugin {
+  apply(compiler /*: Compiler */) {
+    if (compiler.hooks) {
+      console.warn(
+        "" +
+          "You appear to be running Webpack >= 4. " +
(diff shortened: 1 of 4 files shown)


config/RemoveBuildDirectoryPlugin.js
```diff
@@ -0,0 +1,26 @@
...
+// Note: the following type-import just resolves to `any`.
+/*:: import type {Compiler} from "webpack"; */
+
+module.exports = class RemoveBuildDirectoryPlugin {
+  apply(compiler /*: Compiler */) {
+    if (compiler.hooks) {
+      console.warn(
+        "" +
+          "You appear to be running Webpack >= 4. " +
+          "The RemoveBuildDirectoryPlugin should be backward-compatible, " +
```
decentralion: nit: forward compatible

PR #579, opened 2018-08-01
Implement AppState and StateTransitionMachine

decentralion:
The cred explorer app has a variety of valid states. Currently, it is
thrown together without explicit documentation of what its states are,
how they transition, or error handling or testing. I worry that this
will be hard to maintain.

This commit creates the AppState type which explicitly reifies every
reachable state for the app, and a StateTransitionMachine which handles
transitions between states. The transitions are thoroughly tested,
including edge cases where the user makes a change while waiting for a
promise to resolve, or where one of the promises failes.

Test plan:
The unit tests are comprehensive. `yarn test` passes.

Summary:
[Please describe your changes.]

Test Plan:
[Please list the steps required to verify that your change is correct.]


src/app/credExplorer/state.js
```diff
@@ -0,0 +1,247 @@
...
+  repo: ?Repo,
+|};
+export type Initialized = {|
+  +type: "INITIALIZED",
+  +edgeEvaluator: EdgeEvaluator,
+  +repo: Repo,
+  +substate: AppSubstate,
+|};
+
+export type LoadingState = "NOT_LOADING" | "LOADING" | "FAILED";
```
wchargin: It seems like we will want to display some kind of failure message to
the user, so eventually we may have:

```js
export type LoadingState =
  | {|+type: "NOT_LOADING"|}
  | {|+type: "LOADING"|}
  | {|+type: "FAILED", +message: string|};
```

Presuming that you agree, would you like to make this change now, or
leave it for a subsequent commit as a test of how extensible this system
is? :-)

src/app/credExplorer/state.js
```diff
@@ -0,0 +1,247 @@
...
+          graphWithAdapters,
+          loading: "NOT_LOADING",
+        },
+      };
+    } catch (e) {
+      console.error(e);
+      newState = {...state, substate: {...substate, loading: "FAILED"}};
+    }
+    if (deepEqual(this.getState(), loadingState)) {
+      this.setState(NullUtil.get(newState));
```
wchargin: You can omit this `NullUtil.get`; Flow correctly infers that `newState`
cannot be `null`.

src/app/credExplorer/state.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+}
+
+export type GraphWithAdapters = {|
+  +graph: Graph,
+  +adapters: $ReadOnlyArray<DynamicPluginAdapter>,
+|};
+export function loadGraphWithAdapters(repo: Repo): Promise<GraphWithAdapters> {
```
wchargin: Hmm—I wonder. Instead of DIing a function returning
`Promise<GraphWithAdapters>`, do we want to DI a function returning
`Promise<Graph[]>`? I think we might want to. Then, the `Graph.merge`
logic is tested, and this also affords us the ability to show a loading
indicator that distinguishes between the “downloading graphs” and
“merging graphs” states. Along these lines, we might also (eventually)
want to have this function return `Promise<Graph>[]`, so that we can
`Promise.race` the results and display something like “graphs loading
(2/3): Git [done], GitHub [done], Spotlight [loading]”. Interested in
your thoughts, and how much if any of this should be implemented in this
commit.

src/app/credExplorer/state.js
```diff
@@ -0,0 +1,247 @@
...
+  type PagerankOptions,
+  pagerank,
+} from "../../core/attribution/pagerank";
+
+import type {DynamicPluginAdapter} from "../pluginAdapter";
+import {StaticPluginAdapter as GitAdapter} from "../../plugins/git/pluginAdapter";
+import {StaticPluginAdapter as GithubAdapter} from "../../plugins/github/pluginAdapter";
+
+/*
+  This models the UI states of the credExplorer/App as a state machine.
```
wchargin: You probably want to export the `initialState: () => AppState`.

src/app/credExplorer/state.js
decentralion: I think it's a reasonable idea, and we can implement it later once we actually want to populate the field. (For now, you can just check the console when something goes wrong.)

src/app/credExplorer/state.js
decentralion: I think it's a reasonable idea, and I'd like to put it off onto another commit.

src/app/credExplorer/state.js
decentralion: Done.

PR #581, opened 2018-08-01
Streamline the cred explorer UI

decentralion:
This commit removes some unecessary text and whitespace from the cred
explorer. The result is more minimal. It's not super pretty, but that
will come later.

Test plan: Visual inspection

Before:
image

After:
image

### DIFF
--- a/src/app/credExplorer/App.js
+++ b/src/app/credExplorer/App.js
@@ -1,7 +1,6 @@
 // @flow
 
 import React from "react";
-import {StyleSheet, css} from "aphrodite/no-important";
 
 import type {LocalStore} from "../localStore";
 import CheckedLocalStore from "../checkedLocalStore";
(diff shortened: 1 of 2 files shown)


src/app/credExplorer/App.js
```diff
@@ -65,15 +64,13 @@ export class App extends React.Component<Props, State> {
...
-        <header className={css(styles.header)}>
-          <h1>Cred Explorer</h1>
-        </header>
         <div>
           <RepositorySelect
             localStore={localStore}
             onChange={(selectedRepo) => this.setState({selectedRepo})}
           />
           <br />
+          <br />
```
wchargin: Eh… can we achieve the same result without the hacks by, making
`RepositorySelect` return a `<p>{whateverItReturnsNow()}</p>`?

src/app/credExplorer/App.js
```diff
@@ -102,9 +99,8 @@ export class App extends React.Component<Props, State> {
               });
             }}
           >
-            Run basic PageRank
+            Run PageRank
```
wchargin: +1, thanks

PR #583, opened 2018-08-01
Use AppStateTransitionMachine to drive App

decentralion:
Pull #579 reifies the cred explorer state as an explicit state
transition machine, with a well-tested implementation. This pull
re-writes `credExplorer/App.js` to use that implementation, and thoroughly
tests it.

The result is that `credExplorer/App.js` has much simpler code
(because it just binds the rendered components to the state machine),
and is much more thoroughly tested.

To ensure easy testability of the `App` class, it was refactored so that
the module exports a factory function which takes a method for creating
the `AppStateTransitionMachine` and returns an `App` class. This ensures
that in test code, we can easily mock the state transition machine. This
had no effect on external callers, since the higher-level `<AppPage>`
class, which already wraps over `LocalStore` choice, was already the
preferred call site.

I also added a loading indicator component, which presently displays a
status text corresponding to the state, such as "Loading graph...", or
"Error while running PageRank". This way, there is always some user
feedback during loading states (which could take a while).

Test plan:
Visual inspection, and the very thorough included unit tests.

Summary:
[Please describe your changes.]

Test Plan:
[Please list the steps required to verify that your change is correct.]

### DIFF
--- a/src/app/credExplorer/App.test.js
+++ b/src/app/credExplorer/App.test.js
@@ -1,114 +1,335 @@
 // @flow
+
 import React from "react";
 import {shallow} from "enzyme";
 
+import {Graph} from "../../core/graph";
+import {makeRepo} from "../../core/repo";
 import testLocalStore from "../testLocalStore";
-import {pagerank} from "../../core/attribution/pagerank";
-import {App} from "./App";
 
-import {Graph, NodeAddress, EdgeAddress} from "../../core/graph";
+import RepositorySelect from "./RepositorySelect";
+import {PagerankTable} from "./PagerankTable";
+import {WeightConfig} from "./WeightConfig";
+import {createApp, LoadingIndicator} from "./App";
+import {initialState} from "./state";
 
 require("../testUtil").configureEnzyme();
-require("../testUtil").configureAphrodite();
-
-function example() {
-  const graph = new Graph();
-  const nodes = {
-    fooAlpha: NodeAddress.fromParts(["foo", "a", "1"]),
-    fooBeta: NodeAddress.fromParts(["foo", "b", "2"]),
-    bar1: NodeAddress.fromParts(["bar", "a", "1"]),
-    bar2: NodeAddress.fromParts(["bar", "2"]),
-    xox: NodeAddress.fromParts(["xox"]),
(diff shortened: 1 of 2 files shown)


src/app/credExplorer/App.test.js
```diff
@@ -1,114 +1,338 @@
...
+    expect(el.state().appState).toBe(getState());
+  });
+  it("setState is wired properly", () => {
+    const {setState, el} = example();
+    expect(initialState()).not.toBe(initialState()); // sanity check
+    const newState = initialState();
+    setState(newState);
+    expect(el.state().appState).toBe(newState);
+  });
+  it.skip("localStore is wired properly", () => {
```
decentralion: @wchargin, would you mind taking a look into why this is failing? This test is not needed (as the actual requirement, which is that the RepositorySelect and WeightConfig are passed the right localStore, is tested elsewhere and passes), but I don't like that I don't understand why this test fails.

wchargin:
> would you mind taking a look into why this is failing?

Changing the offending assertion to

```js
    expect(el.instance().props.localStore).toBe(localStore);
```

causes the test to pass. This is because `el.props()` returns the props
for the element rendered by the component, not the component itself. So,
you’re seeing the props for the outermost `div`. (Confirm with
`expect(el.props()).toBe(Symbol())` and inspecting the error.

This is a strange decision, but it is documented:

<https://github.com/airbnb/enzyme/blob/290253d346200c2997a98fa053f76a1c0895e899/docs/api/ShallowWrapper/props.md>

decentralion:
Thanks; fixed.

src/app/credExplorer/App.test.js
```diff
@@ -1,114 +1,335 @@
...
+        const {el, loadGraph, setState} = example();
+        setState(stateFn());
+        el.update();
+        const button = el.findWhere(
+          (b) => b.text() === "Load graph" && b.is("button")
+        );
+        if (disabled) {
+          expect(button.props().disabled).toBe(true);
+        } else {
+          expect(button.props().disabled).not.toBe(true);
```
wchargin: Does `.toBe(false)` not work for some reason?

PR #592, opened 2018-08-02
Create script to build static site

wchargin:
Summary:
Currently, we create the static site and deploy it all at once in
`scripts/deploy.sh`. This commit creates a new script that only builds
the static site. This has the advantage that it is easier/less scary to
change that script (because it can be tested without worrying about
deploying to a local test target), and that we can write automated tests
for it.

Test Plan:
Run `yarn sharness`; note that it completes very quickly. Then, in a
shell with your GitHub token exported, run `yarn sharness-full`. Expect
all tests to pass.

For a sanity check, you can run:

```shell
outdir="$(mktemp -d --suffix .sourcecred-site)"
./scripts/build_static_site.sh --target "${outdir}" \
    --cname sourcecred.io \
    --repo sourcecred/example-git \
    --repo sourcecred/example-github \
    ;
(cd "${outdir}" && python -m SimpleHTTPServer)
```

and ensure that <http://localhost:8000/> is as expected.

One test case that is not covered is the following: _if_ the actual app
somehow tries to emit a `CNAME` file at root, _and_ our script’s logic
to catch this is broken, then we will not catch this failure. I’ve
tested the logic manually by adding `>"${cname_file}"` after definition
of that variable, but I don’t see a good way to test it automatically,
without adding flags like `--but-actually-emit-cname-too` to the build.
The compound probability of this happening is sufficiently low that this
doesn’t bother me.

wchargin-branch: build-static-site-script

### DIFF
--- a/scripts/build_static_site.sh
+++ b/scripts/build_static_site.sh
@@ -1,0 +1,136 @@
+#!/bin/bash
+set -eu
+
+usage() {
+    printf 'usage: build_static_site.sh --target TARGET\n'
+    printf '                            [--repo OWNER/NAME [...]]\n'
+    printf '                            [--cname DOMAIN]\n'
+    printf '                            [-h|--help]\n'
+    printf '\n'
+    printf 'Build the static SourceCred website, including example data.\n'
+    printf '\n'
+    printf '%s\n' '--target TARGET'
+    printf '\t%s\n' 'an empty directory into which to build the site'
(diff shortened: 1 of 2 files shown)


wchargin:
I’m actually going to make some changes here; will ping when ready for
review.

wchargin:
I have added tests.

This was really fun to write.

PR #597, opened 2018-08-02
Add `sharness` for shell-based testing

wchargin:
Summary:
We will shortly want to perform testing of shell scripts; it makes the
most sense to do so via the shell. We could roll our own testing
framework, but it makes more sense to use an existing one. By choosing
Sharness, we’re in good company: `go-ipfs` and `go-multihash` use it as
well, and it’s derived from Git’s testing library. I like it a lot.

For now, we need a dummy test file; our test runner will fail if there
are no tests to run. As soon as we have a real test, we can remove this.

This commit was generated by following the “per-project installation”
instructions at https://github.com/chriscool/sharness, and by
additionally including that repository’s `COPYING` file as
`SHARNESS_LICENSE`, with a header prepended. I considered instead adding
Sharness as a submodule, which is supported and has clear advantages
(e.g., you can update the thing), but opted to avoid the complexity of
submodules for now.

Test Plan:
Create the following tests in the `sharness` directory:

```shell
$ cat sharness/good.t
#!/bin/sh
test_description='demo of passing tests'
. ./sharness.sh
test_expect_success "look at me go" true
test_expect_success EXPENSIVE "this may take a while" 'sleep 2'
test_done
# vim: ft=sh
$ cat sharness/bad.t
#!/bin/sh
test_description='demo of failing tests'
. ./sharness.sh
test_expect_success "I don't feel so good" false
test_done
# vim: ft=sh
```

Note that `yarn sharness` and `yarn test` fail appropriately. Note that
`yarn sharness-full` fails appropriately after taking two extra seconds,
and `yarn test --full` runs the latter. Each failure message should
print the name of the failing test case, not just the suite name, and
should indicate that the passing tests passed.

Then, remove `sharness/bad.t`, and note that the above commands all
pass, with the `--full` variants still taking longer.

Finally, remove `sharness/good.t`, and note that the above commands all
pass (and all pass quickly).

wchargin-branch: add-sharness

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,9 @@ to run them all. This will run:
   - **Unit tests** (`yarn unit`). You can also run `yarn unit --watch`
     to automatically re-run tests when you change a relevant file.
 
+  - **Sharness** (`yarn sharness`). This runs shell-based tests, located
+    in the `sharness/` directory.
+
   - **Prettier** (`check-pretty`). You can simply run `yarn prettify` to
     reformat all files. It can be convenient to set up your editor to
     run `yarn prettier --write CURRENT_FILENAME` whenever you save a
--- a/config/test.js
+++ b/config/test.js
@@ -58,6 +58,16 @@ function makeTasks(mode /*: "BASIC" | "FULL" */) {
       deps: [],
     },
     {
+      id: {BASIC: "sharness", FULL: "sharness-full"}[mode],
+      cmd: [
+        "npm",
+        "run",
+        "--silent",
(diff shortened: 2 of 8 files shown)


wchargin:
> I don't think we should complicate `yarn unit`. The vast majority of
> development in this project is against js code, so let's make the
> testing surface area cater to the usual case, and add `yarn sharness`
> as a separate command. Of course, `yarn test` will run both.

Totally reasonable. Will do.

> Also, for posterity please document the "clear advantages" to adding
> yarness as a submodule in the commit message.

Will do.

wchargin:
Updated to integrate with `yarn test --full`. Re-review?

PR #598, opened 2018-08-03
Write prose for the home page

wchargin:
Summary:
@decentralion and I have spent a bunch of time on this prose, and we
think that it’s pretty good. Let’s put some content on our otherwise
bare site.

Test Plan:
Running `yarn start` suffices.

Paired with @decentralion.

wchargin-branch: home-page-prose

### DIFF
--- a/src/app/HomePage.js
+++ b/src/app/HomePage.js
@@ -1,9 +1,222 @@
 // @flow
 
-import React from "react";
+import {StyleSheet, css} from "aphrodite/no-important";
+import React, {type Node} from "react";
+import {Link} from "react-router";
 
 export default class HomePage extends React.Component<{||}> {
   render() {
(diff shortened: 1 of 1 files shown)


decentralion:
FYI @campoy, @vmarkovtsev, @mkosowsk, @owocki, @ajbouh - now that [sourcecred.io](sourcecred.io) is live and have some content, we consider the sourcecred alpha to have started :D

mkosowsk:
@decentralion Great! Will be watching with keen interest 👍

vmarkovtsev:
Cool! How about sorting the PageRank results in descending order?

wchargin:
@vmarkovtsev The PageRank results are sorted in descending order, but the scores are negative-log-score, so lower is better. We should clarify this in the UI.

PR #602, opened 2018-08-03
ensure-flow.sh: simplify, removing dep on GNU grep

### LINKED ISSUE #594: Test: ensure-flow-typing silently failing for Mac
The ensure-flow-typing test fails, but still is marked as a success within a local run of `yarn test`.
The issue seems twofold: 
* Incorrect syntax for grep within ensure-flow.sh (for a Mac)
* Error messages not being piped appropriately --> silent failure

wchargin:
Summary:
By using Git’s magic pathspecs instead of post-processing stream
operations, we reduce the pipeline to a single operation. Git implements
its own version of `grep`, so this should be platform-independent.
Previously, we had needed the `-z` argument to `grep(1)`, which is a GNU
extension.

Fixes #594.

Test Plan:
Ensure that the script passes with no output. Then,

```shell
mkdir -p src/flow-typed/
touch src/foo.js src/flow-typed.js src/flow-typed/foo.js
git add src/foo.js src/flow-typed.js src/flow-typed/foo.js
cd scripts
./ensure-flow.sh
```

and ensure that script exits with code 1, empty stdout, and stderr with:

```
../src/flow-typed.js
../src/flow-typed/foo.js
../src/foo.js
```

This verifies that the pathspec is properly excluding the root directory
`flow-typed`, but not files that just happen to have `flow-typed` in
their paths.

wchargin-branch: simplify-ensure-flow

### DIFF
--- a/scripts/ensure-flow.sh
+++ b/scripts/ensure-flow.sh
@@ -1,7 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 set -eu
-git grep -Fz --files-without-match -e '@flow' -e '@no-flow' -- '*.js' \
-    | grep -zv '^flow-typed/' \
-    | tr '\0' '\n' \
-    | tee >(cat >&2) \
-    | diff -q /dev/null - >/dev/null
+! git grep -F --files-without-match -e '@flow' -e '@no-flow' \
+        -- ':/*.js' ':!/flow-typed/' >&2


wchargin:
@claireandcode: Would appreciate if you could run the test plan when
reviewing; I don’t have a macOS computer to test on. I’m pretty sure
that it should be safe—it conforms to `man git-grep`—but it’d be nice to
make sure that this solves your underlying problem. Thanks!

wchargin:
I’m temporarily developing on macOS, and can confirm that this works as
detailed in the test plan (except for a typo: `ensure-flow.js` should of
course be `ensure-flow.sh`).

Even in light of #698, I’d like to merge this PR because the new code is
clearly better than the old code, irrespective of the fact that it
happens to work on vanilla Mac and POSIX `sh`.

Adding @decentralion for review. (This is not urgent.)

PR #605, opened 2018-08-04
Make PagerankTable props non-nullable

decentralion:
Historically, `credExplorer/App.js` instantiated a PagerankTable
regardless of its state, and would pass null props when the App didn't
have data needed to load the table. As of #583, we just don't create the
PagerankTable before its data is available, which is a simpler/better
contract. As such, the type signature of PagerankTable's props can be
simplified, and some logic for handling the null case may be removed.

Test plan: `yarn test` passes, which is sufficient.

### DIFF
--- a/src/app/credExplorer/PagerankTable.js
+++ b/src/app/credExplorer/PagerankTable.js
@@ -77,8 +77,8 @@ type SharedProps = {|
 |};
 
 type PagerankTableProps = {|
-  +pnd: ?PagerankNodeDecomposition,
-  +adapters: ?$ReadOnlyArray<DynamicPluginAdapter>,
+  +pnd: PagerankNodeDecomposition,
+  +adapters: $ReadOnlyArray<DynamicPluginAdapter>,
   +maxEntriesPerList: number,
 |};
 type PagerankTableState = {|topLevelFilter: NodeAddressT|};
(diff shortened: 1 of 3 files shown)


src/app/credExplorer/PagerankTable.js
```diff
@@ -92,14 +92,12 @@ export class PagerankTable extends React.PureComponent<
...
   render() {
-    if (this.props.adapters == null) {
-      return <p>You must load a graph before seeing PageRank analysis.</p>;
-    }
-    if (this.props.pnd == null) {
-      return <p>Please run PageRank to see analysis.</p>;
-    }
-    if (this.props.maxEntriesPerList == null) {
-      throw new Error("maxEntriesPerList not set");
+    if (
```
wchargin: Do we really need this check?

decentralion: I thought about this as well. I don't like having a habit of unnecessarily checking invariants that the type system should maintain. However, due to https://github.com/facebook/flow/issues/6400 (and similar) I'm not sure I trust flow to actually maintain its checks.

That class of bug does seem pretty rare, and I don't think I want to establish a consistent policy of always checking required props on React component construction, so I'm happy to take the check out.

Issue #502, opened 2018-07-07
Cred explorer: Aggregate over contributions


decentralion:
When looking at a highly connected nodes in the cred explorer (>20-30 connections), it can be hard to orient on what's going on. We can make this more legible by aggregating over this information. For every aggregated group of contributions, we will display a description, and the log of the summed scores in the aggregation. I propose we aggregate first by edge type, then by node type, following the rule that we aggregate whenever there are more than K nodes that would fall into the same group.

For example, if a user has many contributions, it might aggregate as follows:
- AUTHORED 100 nodes
  - AUTHORED 30 pull requests
    - ... pulls ...
  - AUTHORED 20 issues
    - ...issues...
  - AUTHORED 50 reviews
    - ...reviews...
- REFERENCED_BY 3 nodes
  - issue 123
  - comment 456
  - pull 678
 
The use of log scoring on the contributions will make this somewhat confusing; it will be counter-intuitive that the displayed scores in an aggregation don't sum to the value of the aggregation.

A concrete use case: ever since our first SourceCred prototype, @wchargin has had more cred than me, and I've never understood why. I think this will shed some light on the issue!

Issue #465, opened 2018-06-30
Formalize PluginAdapter semantics


decentralion:
- Can two plugins have the same prefix?
- Can two plugins have the same display name?
- Can one plugin have two types with the same prefix?
- Can one plugin have a prefix that is a prefix of another plugin's prefix?
- Can two plugins have types with the same prefix?

Tune in next episode for answers (and invariant assertions) to all these questions, and more!

wchargin:
Cool; I like this. I was going to take a stab at implementing this, but
it sounds like you have a lot of context/prior work on it, so I’ll hold
off on doing so.

> They both have names

Note that names of plugins could _arguably_ be required to be unique
(though I wouldn’t be thrilled about this), but names of types are
definitely not unique. Concepts of “user”, “issue”, “comment” will be
shared across lots of plugins. (No change required.)

> What if we unify them into a shared recursive abstraction

I don’t see how the abstraction that you propose is recursive. What do
you mean by that?

> Then, when finding the right method for a given node address, we choose the most specific Type available (and the presence of the BaseNodeType ensures that we always find a type).

I like the idea of having the base node type being a “fallback” on the
same level of everything else.

> It is illegal to have multiple types with the exact same prefix.

Yes, agreed. Perhaps we should also require that if `t1` is a child type
of `t0` (e.g., `t1` is a node type and `t0` is a plugin), then the
prefix of `t0` is a prefix of the prefix of `t1`. (That is, the function
from node types to node prefixes is a poset homomorphism.) Then, it is
possible to ensure that all types have distinct prefixes by reasoning
locally: you only have to consider your siblings, not all other types at
all levels, and in particular it will be valid to combine any set of
plugins as long as their _top-level_ prefixes are pairwise disjoint.

> `type NodeType`

Multiple functions of the `NodeType` (`render`, `weight`, at least) will
need to take more context than just a single node address—the graph, at
the very least, plus maybe some additional context (e.g., the GitHub
plugin’s relational view). But it sounds like you’re suggesting that the
node types be constants, as opposed to (e.g.) “creating” the GitHub node
types for a particular repository owner and name, receiving a
`Promise<NodeType[]>` in return. How is this resolved?

> As written, the types have all kinds of stuff attached (render,
> weight, etc), but clearly we should factor that out.

Agreed, though I’m willing to do this as needed.

decentralion:
Another idea, based on some discussions with @wchargin:

Often times, consumers of the static adapter actually want the data in this type:
[...]

PR #609, opened 2018-08-06
Add type signatures for aggregation logic

decentralion:
This is the first real step towards #502.

Factoring this out because deciding the type signatures was non-trivial,
and the work was paired with @wchargin

Test plan: `yarn test`

### DIFF
--- a/src/core/attribution/aggregation.js
+++ b/src/core/attribution/aggregation.js
@@ -1,0 +1,35 @@
+// @flow
+
+import type {NodeType, EdgeType} from "../graph";
+import type {ScoredConnection} from "./pagerankNodeDecomposition";
+
+export type ConnectionType =
+  | {|+type: "IN_EDGE", +edgeType: EdgeType|}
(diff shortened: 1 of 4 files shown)


src/core/attribution/aggregation.js
```diff
@@ -0,0 +1,35 @@
...
+export type ConnectionAggregation = {|
+  +connectionType: ConnectionType,
+  +aggregationSummary: AggregationSummary,
+  +nodeAggregations: NodeAggregation[],
+|};
+
+export type NodeAggregation = {|
+  +nodeType: NodeType,
+  +aggregationSummary: AggregationSummary,
+  +connections: ScoredConnection,
```
wchargin: Should this be an array?

wchargin:
This is on hold, yes?

decentralion:
Yep, this is on hold. I'm going to tackle #465 first, and then re-write this so it doesn't have any dependency on the NodeType/EdgeType concepts. Closing this for now.

PR #612, opened 2018-08-07
Add `NodeTrie` and `EdgeTrie` classes

decentralion:
For #465, I'm planning to create an abstraction over NodeTypes and
EdgeTypes which traverses a hierarchy of types and aggregates/reduces
information across all the matching types for a given Node/Edge address.

To do that efficiently, we will want tries[1].

Thanks to @wchargin for helping me figure out how to implement this.

Test plan: Unit tests. The code is a little tricky so please review it
closely.

[1]: https://en.wikipedia.org/wiki/Trie

### DIFF
--- a/src/core/trie.js
+++ b/src/core/trie.js
@@ -1,0 +1,99 @@
+// @flow
+
+import {type AddressModule} from "./address";
+import {
+  NodeAddress,
+  type NodeAddressT,
+  EdgeAddress,
+  type EdgeAddressT,
+} from "./graph";
+import * as NullUtil from "../util/null";
+
+const EMPTY_ENTRY_SYMBOL = Symbol("EMPTY");
+
+type Entry<V> = {|+map: RecursiveMap<V>, value: V | typeof EMPTY_ENTRY_SYMBOL|};
+type RecursiveMap<V> = Map<string, Entry<V>>;
+class BaseTrie<K, V> {
+  addressModule: AddressModule<K>;
+  entry: Entry<V>;
+
+  /**
+   * Create an empty trie backed by the given address module.
+   */
+  constructor(m: AddressModule<K>) {
+    this.addressModule = m;
+    this.entry = {value: EMPTY_ENTRY_SYMBOL, map: new Map()};
+  }
+
+  /**
+   * Add key `k` to this trie with value `v`. Return `this`.
+   */
+  add(k: K, val: V): this {
+    const parts = this.addressModule.toParts(k);
+    let entry = this.entry;
+    for (const part of parts) {
+      if (!entry.map.has(part)) {
+        entry.map.set(part, {map: new Map(), value: EMPTY_ENTRY_SYMBOL});
+      }
+      entry = NullUtil.get(entry.map.get(part));
+    }
+    if (entry.value !== EMPTY_ENTRY_SYMBOL) {
+      throw new Error(
+        `Tried to overwrite entry at ${this.addressModule.toString(k)}`
+      );
+    }
+    entry.value = val;
+    return this;
+  }
+
+  /**
+   * Get the values in this trie along the path to `k`.
+   *
+   * More specifically, this method has the following observable
+   * behavior. Let `inits` be the list of all prefixes of `k`, ordered
+   * by length (shortest to longest). Observe that the length of `inits`
+   * is `n + 1`, where `n` is the number of parts of `k`; `inits` begins
+   * with the empty address and ends with `k` itself. Initialize the
+   * result to an empty array. For each prefix `p` in `inits`, if `p`
+   * was added to this trie with value `v`, then append `v` to
+   * `result`. Return `result`.
+   */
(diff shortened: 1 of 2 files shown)


src/core/trie.js
```diff
@@ -0,0 +1,79 @@
...
+import {type AddressModule} from "./address";
+import {
+  NodeAddress,
+  type NodeAddressT,
+  EdgeAddress,
+  type EdgeAddressT,
+} from "./graph";
+import * as NullUtil from "../util/null";
+
+type Entry<T> = {map: RecursiveMap<T>, value: ?T};
```
wchargin: It seems that the implementation is unsound when `T` is nullable. You
could avoid this by either requiring that the values be non-null (via
`$NonMaybeType`), or by using a discriminated union, or by hacking
together a discriminated union with a fixed private `Symbol`.

src/core/trie.js
```diff
@@ -0,0 +1,79 @@
...
+  constructor(m: AddressModule<K>) {
+    this.addressModule = m;
+    this.entry = {value: null, map: new Map()};
+  }
+
+  add(k: K, val: T): this {
+    const parts = this.addressModule.toParts(k);
+    let i = 0;
+    let entry = this.entry;
+    while (i < parts.length) {
```
wchargin: You can drop the `let i` and the `while`; this can be:

```js
for (const part of this.addressModule.toParts(k))
```

src/core/trie.js
wchargin: Looks like this can be exact, and `map` can be read-only?

src/core/trie.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  get(k: K): T[] {
+    const parts = this.addressModule.toParts(k);
+    const result = [];
+
+    let entry = this.entry;
+    let i = 0;
+
+    while (i < parts.length) {
```
wchargin: Likewise with a `for`-loop.

src/core/trie.js
```diff
@@ -0,0 +1,79 @@
...
+  NodeAddress,
+  type NodeAddressT,
+  EdgeAddress,
+  type EdgeAddressT,
+} from "./graph";
+import * as NullUtil from "../util/null";
+
+type Entry<T> = {map: RecursiveMap<T>, value: ?T};
+type RecursiveMap<T> = Map<string, Entry<T>>;
+class BaseTrie<K, T> {
```
wchargin: I would like to see the semantics documented, please. Perhaps consider:

```js
class Trie<K, V> {
  /**
   * Create an empty trie backed by the given address module.
   */
  constructor(addressModule: AddressModule<K>): void;

  /**
   * Add key `k` to this trie with value `v`. Return `this`.
   */
  add(k: K, v: V): this;

  /**
   * Get the values in this trie along the path to `k`.
   *
   * More specifically, this method has the following observable
   * behavior. Let `inits` be the list of all prefixes of `k`, ordered
   * by length (shortest to longest). Observe that the length of `inits`
   * is `n + 1`, where `n` is the number of parts of `k`; `inits` begins
   * with the empty address and ends with `k` itself. Initialize the
   * result to an empty array. For each prefix `p` in `inits`, if `p`
   * was added to this trie with value `v`, then append `v` to
   * `result`. Return `result`.
   */
  get(k: K): V[];
}
```

src/core/trie.test.js
```diff
@@ -0,0 +1,92 @@
...
+    const x = new NodeTrie().add(empty, 1).add(foo, 2);
+    expect(x.get(foo)).toEqual([1, 2]);
+  });
+
+  it("swapping the order of addition doesn't change results", () => {
+    const x = new NodeTrie().add(foo, 2).add(empty, 1);
+    expect(x.get(foo)).toEqual([1, 2]);
+  });
+
+  it("get isn't phased by intermediary parts missing values", () => {
```
wchargin: nit: _fazed_, not _phased_

src/core/trie.js
wchargin: minor: Typically, the type parameters to a key-value data structure
are `<K, V>` instead of `<K, T>`.

src/core/trie.js
decentralion: used a symbol.

src/core/trie.js
decentralion: Done

src/core/trie.js
decentralion: Done

src/core/trie.js
decentralion: Done.

src/core/trie.js
decentralion: Done

src/core/trie.test.js
decentralion: Fixed

src/core/trie.js
decentralion: Fixed

PR #620, opened 2018-08-07
Add findEdgeType and findNodeType

decentralion:
These methods find the unique matching node or edge type for a node,
from a given pluginAdapter.

Test plan: unit tests

### DIFF
--- a/src/app/pluginAdapter.test.js
+++ b/src/app/pluginAdapter.test.js
@@ -13,6 +13,8 @@ import {
   staticDispatchByEdge,
   dynamicDispatchByNode,
   dynamicDispatchByEdge,
+  findNodeType,
+  findEdgeType,
 } from "./pluginAdapter";
 
 describe("app/pluginAdapter", () => {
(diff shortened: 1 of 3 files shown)


src/app/pluginAdapter.test.js
```diff
@@ -98,4 +122,40 @@ describe("app/pluginAdapter", () => {
       );
     });
   });
+
+  describe("findNodeType", () => {
+    // No need to check that it erros when there is not a unique type match,
```
wchargin: nit: _errors_ (sp.)

wchargin: I disagree in principle with the claim that “the implementation is
correct, so there’s no need to test it”. The implementation is of course
correct now, but I’m free to change the implementation so that it no
longer calls `findUniqueMatch`, and all that.

With the above in mind, if you still don’t want to test this error case,
I’m willing to let it slide.

decentralion:
Fair. I changed it to test all the relevant properties.

PR #624, opened 2018-08-08
Implement core aggregation logic

decentralion:
This implements two methods:
`aggregateByNodePrefix` groups `scoredConnection`s by the specified
`NodePrefixes`, along with summary statistics.

`aggregateByConnectionType` groups `scoredConnection`s by specified edge
prefixes (and direction, with a special category for the synthetic self
loop), and then sub-aggregates by `NodePrefixes` per
`aggregateByNodePrefix`.

This is progress towards #502.

Test plan: unit tests included.

### DIFF
--- a/src/app/credExplorer/pagerankTable/aggregate.test.js
+++ b/src/app/credExplorer/pagerankTable/aggregate.test.js
@@ -1,7 +1,343 @@
 // @flow
 
-describe("app/credExplorer/aggregation", () => {
-  it("is not yet implemented", () => {
-    return;
+import {EdgeAddress, NodeAddress} from "../../../core/graph";
+import {aggregateByNodeType, aggregateByConnectionType} from "./aggregate";
+
+describe("app/credExplorer/aggregate", () => {
+  function example() {
+    const nodes = {
+      root: NodeAddress.fromParts(["root"]),
+      zap: NodeAddress.fromParts(["zap"]),
+      kif: NodeAddress.fromParts(["kif"]),
+    };
+
+    const nodeTypes = {
+      root: {name: "root", prefix: nodes.root, defaultWeight: 0},
+      zap: {name: "zap", prefix: nodes.zap, defaultWeight: 0},
+      kif: {name: "kif", prefix: nodes.kif, defaultWeight: 0},
+      empty: {name: "empty", prefix: NodeAddress.empty, defaultWeight: 0},
+    };
+
+    const nodeTypesArray = [nodeTypes.root, nodeTypes.zap, nodeTypes.kif];
+
+    const edgeTypes = {
+      foo: {
+        forwardName: "foos",
+        backwardName: "foo'd by",
+        prefix: EdgeAddress.fromParts(["foo"]),
+      },
+      bar: {
+        forwardName: "bars",
+        backwardName: "bar'd by",
+        prefix: EdgeAddress.fromParts(["bar"]),
+      },
+      empty: {
+        forwardName: "empty",
+        backwardName: "emptied-by",
+        prefix: EdgeAddress.empty,
+      },
+    };
+    const edgeTypesArray = [edgeTypes.foo, edgeTypes.bar];
+
+    const edges = {
+      rfz4: {
+        src: nodes.root,
+        dst: nodes.zap,
+        address: EdgeAddress.fromParts(["foo", "1"]),
+      },
+      rfk1: {
+        src: nodes.root,
+        dst: nodes.kif,
+        address: EdgeAddress.fromParts(["foo", "2"]),
+      },
+      kfr3: {
+        src: nodes.kif,
+        dst: nodes.root,
+        address: EdgeAddress.fromParts(["foo", "3"]),
+      },
+      rbk2: {
+        src: nodes.root,
+        dst: nodes.kif,
+        address: EdgeAddress.fromParts(["bar", "1"]),
+      },
+    };
+
+    // All adjacencies are from perspective of "root"
+    const adjacencies = {
+      loop5: {type: "SYNTHETIC_LOOP"},
+      rfz4: {type: "OUT_EDGE", edge: edges.rfz4},
+      rfk1: {type: "OUT_EDGE", edge: edges.rfk1},
+      rbk2: {type: "OUT_EDGE", edge: edges.rbk2},
+      kfr3: {type: "IN_EDGE", edge: edges.kfr3},
+    };
+    const connections = {
+      loop5: {adjacency: adjacencies.loop5, weight: 0.2},
+      rfz4: {adjacency: adjacencies.rfz4, weight: 0.2},
+      rfk1: {adjacency: adjacencies.rfk1, weight: 0.2},
+      rbk2: {adjacency: adjacencies.rbk2, weight: 0.2},
+      kfr3: {adjacency: adjacencies.kfr3, weight: 0.2},
+    };
+    const scoredConnections = {
+      loop5: {
+        connection: connections.loop5,
+        source: nodes.root,
+        sourceScore: 0.2,
+        connectionScore: 5,
+      },
+      rfz4: {
+        connection: connections.rfz4,
+        source: nodes.zap,
+        sourceScore: 0.2,
+        connectionScore: 4,
+      },
+      kfr3: {
+        connection: connections.kfr3,
+        source: nodes.kif,
+        sourceScore: 0.2,
+        connectionScore: 3,
+      },
+      rbk2: {
+        connection: connections.rbk2,
+        source: nodes.kif,
+        sourceScore: 0.2,
+        connectionScore: 2,
+      },
+      rfk1: {
+        connection: connections.rfk1,
+        source: nodes.kif,
+        sourceScore: 0.2,
+        connectionScore: 1,
+      },
+    };
+    const scoredConnectionsArray = [
+      scoredConnections.loop5,
+      scoredConnections.rfz4,
+      scoredConnections.rfk1,
+      scoredConnections.rbk2,
+      scoredConnections.kfr3,
+    ];
+    return {
+      nodes,
+      edges,
+      nodeTypes,
+      nodeTypesArray,
+      edgeTypes,
+      edgeTypesArray,
+      adjacencies,
+      connections,
+      scoredConnections,
+      scoredConnectionsArray,
+    };
+  }
+  describe("aggregateByNodeType", () => {
+    function exampleNodeAggregation() {
+      const {nodeTypesArray, scoredConnectionsArray} = example();
+      const aggregations = aggregateByNodeType(
+        scoredConnectionsArray,
+        nodeTypesArray
+      );
+      return aggregations;
+    }
+    it("puts every connection in an aggregation", () => {
+      const {scoredConnectionsArray} = example();
+      let connectionsFound = 0;
+      for (const aggregation of exampleNodeAggregation()) {
+        connectionsFound += aggregation.connections.length;
+      }
+      expect(connectionsFound).toEqual(scoredConnectionsArray.length);
+    });
+    it("groups connections by node type", () => {
+      for (const aggregation of exampleNodeAggregation()) {
+        for (const connection of aggregation.connections) {
+          expect(
+            NodeAddress.hasPrefix(
+              connection.source,
+              aggregation.nodeType.prefix
+            )
+          ).toBe(true);
+        }
+      }
+    });
+    it("when multiple node types match, it uses the most specific one", () => {
+      const {nodeTypes, scoredConnections} = example();
+      const aggregations = aggregateByNodeType(
+        [scoredConnections.rfz4],
+        [nodeTypes.empty, nodeTypes.zap]
+      );
+      expect(aggregations).toHaveLength(1);
+      // It used the zap nodeType rather than the empty nodeType b/c
+      // it was more specific
+      expect(aggregations[0].nodeType).toEqual(nodeTypes.zap);
+    });
+    it("summarizes the group size appropriately", () => {
+      const {nodeTypesArray, scoredConnectionsArray} = example();
+      const aggregations = aggregateByNodeType(
+        scoredConnectionsArray,
+        nodeTypesArray
+      );
+      for (const aggregation of aggregations) {
+        expect(aggregation.summary.size).toEqual(
+          aggregation.connections.length
+        );
+      }
+    });
+    it("summarizes the group score appropriately", () => {
+      for (const aggregation of exampleNodeAggregation()) {
+        const scores = aggregation.connections.map((x) => x.connectionScore);
+        const totalScore = scores.reduce((a, b) => a + b);
+        expect(aggregation.summary.score).toEqual(totalScore);
+      }
+    });
+    it("does not create empty groups", () => {
+      for (const aggregation of exampleNodeAggregation()) {
+        expect(aggregation.summary.size).not.toEqual(0);
+      }
+    });
+    it("handles the case where there are no connections", () => {
+      const {nodeTypesArray} = example();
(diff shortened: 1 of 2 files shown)


src/core/attribution/aggregation.js
```diff
@@ -25,11 +30,129 @@ export type ConnectionAggregation = {|
...
+ * return incorrect results (example in test code).
+ */
+export function aggregateByNodePrefix(
+  xs: $ReadOnlyArray<ScoredConnection>,
+  nodePrefixes: $ReadOnlyArray<NodeAddressT>
+): $ReadOnlyArray<NodeAggregation> {
+  let connectionsFound = 0;
+  const result = [];
+  for (const nodePrefix of nodePrefixes) {
+    const connections = xs.filter(({source}) =>
```
wchargin: If you wanted to check the precondition listed in the contract, it seems
like it’d be quite simple to do: add a `const seen: Set<mixed>` at outer
level, and for each source of `connections` add the source to `visited`,
erroring if it already exists. You can rely on physical equality here
because you’re pulling objects from the same data structure, not
constructing them anew. I’d support this change if you’re willing to
make it. Then, you can remove the necessary-but-not-sufficient check
below.

src/core/attribution/aggregation.js
```diff
@@ -25,11 +30,129 @@ export type ConnectionAggregation = {|
...
+  const result = [syntheticAggregation];
+  for (const edgePrefix of edgePrefixes) {
+    const withCorrectEdgePrefix = xs.filter(({connection: {adjacency}}) => {
+      if (adjacency.type === "IN_EDGE" || adjacency.type === "OUT_EDGE") {
+        return EdgeAddress.hasPrefix(adjacency.edge.address, edgePrefix);
+      }
+      return false;
+    });
+
+    // It would be much cleaner if createAggregation actually returned the
```
wchargin: You can work around the inference failure without too much difficulty:

```diff
diff --git a/src/core/attribution/aggregation.js b/src/core/attribution/aggregation.js
index 3307a6a..ca97b89 100644
--- a/src/core/attribution/aggregation.js
+++ b/src/core/attribution/aggregation.js
@@ -102,10 +102,12 @@ export function aggregateConnections(
   connectionsFound += syntheticConnections.length;
   function createAggregation(
     connections: $ReadOnlyArray<ScoredConnection>,
-    type: "OUT_EDGE" | "IN_EDGE"
-  ) {
+    connectionType:
+      | {|+type: "IN_EDGE", edgePrefix: EdgeAddressT|}
+      | {|+type: "OUT_EDGE", edgePrefix: EdgeAddressT|}
+  ): ConnectionAggregation {
     const matchingConnections = connections.filter(
-      (x) => x.connection.adjacency.type === type
+      (x) => x.connection.adjacency.type === connectionType.type
     );
     const size = matchingConnections.length;
     connectionsFound += size;
@@ -115,6 +117,7 @@ export function aggregateConnections(
       nodePrefixes
     );
     return {
+      connectionType,
       aggregationSummary: {size, aggregateScore},
       nodeAggregations,
     };
@@ -127,24 +130,12 @@ export function aggregateConnections(
       }
       return false;
     });
-
-    // It would be much cleaner if createAggregation actually returned the
-    // aggregation, but unfortunately a flow bug necessitates this more verbose
-    // construction.
-    const outCollation = createAggregation(withCorrectEdgePrefix, "OUT_EDGE");
-    const outAggregation: ConnectionAggregation = {
-      connectionType: {type: "OUT_EDGE", edgePrefix},
-      aggregationSummary: outCollation.aggregationSummary,
-      nodeAggregations: outCollation.nodeAggregations,
-    };
-    const inCollation = createAggregation(withCorrectEdgePrefix, "IN_EDGE");
-    const inAggregation: ConnectionAggregation = {
-      connectionType: {type: "IN_EDGE", edgePrefix},
-      aggregationSummary: inCollation.aggregationSummary,
-      nodeAggregations: inCollation.nodeAggregations,
-    };
-    result.push(outAggregation);
-    result.push(inAggregation);
+    result.push(
+      createAggregation(withCorrectEdgePrefix, {type: "OUT_EDGE", edgePrefix})
+    );
+    result.push(
+      createAggregation(withCorrectEdgePrefix, {type: "IN_EDGE", edgePrefix})
+    );
   }
   if (connectionsFound !== xs.length) {
     // It would be extremely misleading to silently return an aggregation that
```

src/core/attribution/aggregation.js
```diff
@@ -25,11 +30,129 @@ export type ConnectionAggregation = {|
...
+    // did not successfully aggregate every connection, as this would lead to
+    // pieces of the graph silently disappearing.
+    throw new Error(
+      "Invariant violation: connections found is not same as connections in"
+    );
+  }
+  return result;
+}
+
+/* The correctness of this function depends on the following conditions:
```
wchargin: One unfortunate property of this implementation is that in the case that
we add new kinds of connection types, Flow will not direct us to modify
this function, and the connections will be silently omitted from the
aggregation. Do you have any ideas about how we might improve this?

src/core/attribution/aggregation.js
decentralion: Found an alternate solution by using the new Trie class

src/core/attribution/aggregation.js
decentralion: Fixed in the rewrite.

src/core/attribution/aggregation.js
decentralion: Fixed in the rewrite.

src/core/attribution/aggregation.test.js
```diff
@@ -1,7 +1,500 @@
...
+        connectionScore: 0.2,
+      },
+      kfr: {
+        connection: connections.kfr,
+        source: nodes.kif,
+        sourceScore: 0.2,
+        connectionScore: 0.2,
+      },
+    };
+    const scoredConnectionsInOrder = [
```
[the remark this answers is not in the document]
decentralion: Fixed in the rewrite.

src/core/attribution/aggregation.test.js
```diff
@@ -1,7 +1,500 @@
...
+          connections: [scoredConnections.loop],
+        },
+        {
+          nodePrefix: nodes.zap,
+          aggregationSummary: {size: 1, aggregateScore: 0.2},
+          connections: [scoredConnections.rfz],
+        },
+        {
+          nodePrefix: nodes.kif,
+          aggregationSummary: {size: 3, aggregateScore: 0.2 + 0.2 + 0.2},
```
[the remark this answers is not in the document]
decentralion: No longer an issue in the rewrite.

decentralion:
I've re-written the implementation to use Tries, which avoids the silent invariant failure, and as a bonus is forward-compatible with subtyping / hierarchical types. Also, the tests have been re-written to be dramatically clearer. PTAL.

src/app/credExplorer/pagerankTable/aggregation.js
```diff
@@ -29,11 +33,131 @@ export type ConnectionAggregation = {|
...
+    const type = types[types.length - 1];
+    if (!relevantMap.has(type)) {
+      relevantMap.set(type, []);
+    }
+    NullUtil.get(relevantMap.get(type)).push(x);
+  }
+
+  const result = [];
+
+  if (syntheticConnections.length > 0) {
```
wchargin: This can all just be:

```js
  if (syntheticConnections.length > 0) {
    result.push(
      createAggregation({type: "SYNTHETIC_LOOP"}, syntheticConnections)
    );
  }
```

(in which case it should be below the fundef).

src/app/credExplorer/pagerankTable/aggregation.test.js
```diff
@@ -1,7 +1,422 @@
...
+    });
+    it("when multiple node types match, it uses the most specific one", () => {
+      const {nodeTypes, scoredConnections} = example();
+      const aggregations = aggregateByNodeType(
+        [scoredConnections.rfz4],
+        [nodeTypes.empty, nodeTypes.zap]
+      );
+      expect(aggregations).toHaveLength(1);
+      // It used the zap nodeType rather than the empty nodeType b/c it was more specific
+      expect(aggregations[0].nodeType).toEqual(nodeTypes.zap);
```
wchargin: A more property-based (less example-based)way to do this would be to
check that if a node with source `a` is categorized under prefix `p`,
and `a` has prefix `p'`, then `p'` is a prefix of `p`. The way that
you’ve done it is not unreasonable, but you seemed to be enjoying the
property-based nature overall, so I humbly submit this alternative for
your consideration.

src/app/credExplorer/pagerankTable/aggregation.test.js
```diff
@@ -1,7 +1,422 @@
...
+      }
+    });
+    it("summarizes the group score appropriately", () => {
+      const {nodeTypesArray, scoredConnectionsArray} = example();
+      const aggregations = aggregateByNodeType(
+        scoredConnectionsArray,
+        nodeTypesArray
+      );
+      for (const aggregation of aggregations) {
+        const totalScore = aggregation.connections.reduce(
```
wchargin: Nit incoming, but one that I care about somewhat. You’ve a couple times
now written code of the form

```js
        const totalScore = aggregation.connections.reduce(
          (val, connection) => val + connection.connectionScore,
          0
        );
```

, which can be rewritten as

```js
        const scores = aggregation.connections.map((x) => x.connectionScore);
        const totalScore = scores.reduce((a, b) => a + b);
```

. I posit that the latter is better, both for “ivory-tower” “purity”
reasons\* and for practical reasons. The practical reason is simply the
following: while it is not _difficult_ to rederive at each occasion
whether the accumulator or the next item comes as the first argument to
the callback, it requires active thought each time that you read the
code (for me, at least, and I’m not alone). When your accumulator type
is different from your element type, the order matters. When you map
first so that everything is homogeneous, it doesn’t.\*\*

I think that this is consonant with the principle that code is read much
more often than it’s written; the obviously correct version is better
than the slightly-nonobviously correct version.

Another practical reason is that it’s impossible to get wrong if you map
first and then reduce. If you got the arguments wrong here, you’d get
`0` for an empty array or `"[object Object]undefined"` for a nonempty
array. Sure, tests will catch it, but why bother tracking down that bug?

\* The ivory-tower reason is roughly that when you perform a true
_reduction_, as opposed to a fold, you have much more freedom in terms
of execution strategy. For instance, it is trivial to parallelize a
reduction, but it is impossible to parallelize a fold. This really does
matter in principle (you sacrifice O(log *n*) for O(*n*)), but in
practice this code is not parallel, and if we want to make it parallel
then we can trivially make the transformation.

\*\* This is true even if the operator is non-commutative—ask me for
details if you care.

So—I care about this enough to write my own code in my preferred style,
and I want to let you know where I’m coming from. But if you have any
remotely good reasons for favoring the former style in any particular
case then I’m happy to let it slide.

\</nit> :-)

src/app/credExplorer/pagerankTable/aggregation.test.js
```diff
@@ -1,7 +1,422 @@
...
+    it("sorts the aggregations by total score", () => {
+      const {nodeTypesArray, scoredConnectionsArray} = example();
+      const aggregations = aggregateByNodeType(
+        scoredConnectionsArray,
+        nodeTypesArray
+      );
+      let lastSeenScore = Infinity;
+      for (const aggregation of aggregations) {
+        const score = aggregation.summary.score;
+        expect(lastSeenScore >= score).toBe(true);
```
wchargin: Prefer `expect(lastSeenScore).toBeGreaterThanOrEqual(score)`, yes?

src/app/credExplorer/pagerankTable/aggregation.test.js
```diff
@@ -1,7 +1,422 @@
...
+        edgeTypesArray
+      );
+      let lastSeenScore = Infinity;
+      for (const aggregation of aggregations) {
+        const score = aggregation.summary.score;
+        expect(lastSeenScore >= score).toBe(true);
+        lastSeenScore = score;
+      }
+    });
+    it("within each aggregation, sub-aggregations are sorted by score", () => {
```
wchargin: Just to check: this is done by `aggregateByNodeType` and doubly tested,
right? (No objection.)

src/app/credExplorer/pagerankTable/aggregation.js
```diff
@@ -29,11 +33,131 @@ export type ConnectionAggregation = {|
...
+  const nodeTypeToConnections = new Map();
+  for (const x of xs) {
+    const types = typeTrie.get(x.source);
+    if (types.length === 0) {
+      throw new Error(
+        `No matching NodeType for ${NodeAddress.toString(x.source)}`
+      );
+    }
+    const type = types[types.length - 1];
+    if (!nodeTypeToConnections.has(type)) {
```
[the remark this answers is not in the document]
decentralion: Nice suggestion; fixed.

src/app/credExplorer/pagerankTable/aggregation.js
decentralion: Nice catch.

src/app/credExplorer/pagerankTable/aggregation.test.js
decentralion: I like that this example very clearly exhibits the sub-prefixing behavior, and is very easy to reason about.

src/app/credExplorer/pagerankTable/aggregation.test.js
decentralion: Yep, this is a great suggestion. I got confused about the reduce argument order several times while writing the code. I'm surprised that reduces can be parallelized when non-commutative, if either of us remember tomorrow I want to hear why.

src/app/credExplorer/pagerankTable/aggregation.test.js
decentralion: Sure, done (here and elsewhere)

src/app/credExplorer/pagerankTable/aggregation.test.js
decentralion: Yep

decentralion:
Addressed review feedback; also refactored the test code so that there's less repetition of setting up the example aggregations

src/app/credExplorer/pagerankTable/aggregate.test.js
```diff
@@ -1,7 +1,344 @@
...
+      );
+      for (const aggregation of aggregations) {
+        expect(aggregation.summary.size).toEqual(
+          aggregation.connections.length
+        );
+      }
+    });
+    it("summarizes the group score appropriately", () => {
+      for (const aggregation of exampleNodeAggregation()) {
+        const totalScore = aggregation.connections.reduce(
```
wchargin: Can patch this `reduce`, too.

src/app/credExplorer/pagerankTable/aggregate.test.js
```diff
@@ -1,7 +1,344 @@
...
+      }
+    });
+    it("when multiple node types match, it uses the most specific one", () => {
+      const {nodeTypes, scoredConnections} = example();
+      const aggregations = aggregateByNodeType(
+        [scoredConnections.rfz4],
+        [nodeTypes.empty, nodeTypes.zap]
+      );
+      expect(aggregations).toHaveLength(1);
+      // It used the zap nodeType rather than the empty nodeType b/c it was more specific
```
wchargin: nit: please wrap (preferably at 72, but 80 is okay)

src/app/credExplorer/pagerankTable/aggregate.test.js
decentralion: Done

src/app/credExplorer/pagerankTable/aggregate.test.js
decentralion: Done

Issue #627, opened 2018-08-09
Rewrite the Git plugin


decentralion:
As currently implemented, the Git plugin is both extremely bloated, and of marginal value for cred attribution. The core issue is that we've closely mirrored Git's own data model: we store every commit, every tree and tree entry, blob, and so forth. This means we have an enormous number of edges that have no particular meaning vis-a-vis cred attribution. 

Consider an extreme example: a poorly-organized project which places every file in a single bloated directory. A user writes a commit which modifies one particular file. In our current model, this creates a new Tree, with new pointers to TreeEntries for every single file in the project. This flurry of new edges disguises the true signal: that there should be a connection between the commit and the particular file it touched.

While that is an extreme example, it is true in general that the Git graphs are regularly >10x the size of the GitHub graph, while containing very little discernable signal. We should rewrite the system.

@wchargin and I spent some time today discussing an alternate model, in which each commit has edges just to the files that it modifies. We need to spec out exactly how to handle merge commits, and how we can incorporate other information like directory structure so as to collect that files in the same directory are related.

decentralion:
In the short term, I want to put in a super-reduced Git plugin. It will have one node type (commit) and one edge type (commit-has-parent). This will be helpful for the near-term work I've got planned around re-writing the WeightConfig; having multiple plugins that have their own edge and node types changes the UI and I want to accommodate this from the start. It will also fix a bug where git commit nodes are currently rendered as unknown nodes with ugly descriptions, and serve as a hint to the user that we plan to incorporate commit nodes in richer ways in the future.

I see three principle ways to do this:

1. Modify the git plugin in-place so that it only returns the nodes and edges described, while leaving most of the logic and testing fundamentally the same
2. Hack the git plugin so that it only contains the logic relevant for commits, deleting the other logic
3. Fork the git plugin into git-minimal and git-deprecated (or similar), and then hack git minimal as described in **2**

I'm opposed to taking course 1. because it will leave the code in a confusing state where it's hard to tell which code is live and which is deprecated. I'm fine with either 2. or 3.; @wchargin do you have a preference between them?

wchargin:
> In the short term, I want to put in a super-reduced Git plugin. It
> will have one node type (commit) and one edge type
> (commit-has-parent).

Okay.

> This will be helpful for the near-term work I've got planned around
> re-writing the WeightConfig; having multiple plugins that have their
> own edge and node types changes the UI and I want to accommodate this
> from the start. 

Agreed.

[...]

PR #628, opened 2018-08-09
Disable the Git plugin

decentralion:
See #627 for context.

Removing the Git plugin results in an enormous performance improvement.
In my testing on `metamask/metamask-extension`: before this change, load
took 23s, and PageRank took >9 mins and then crashed. After this change,
load+PageRank took 5s combined.

Test plan: All unit tests pass; loading new data from the CLI works; and
I poked around the UI to make sure there were no spurious references to
the Git plugin.

Note: This does not break backcompat, there's no need to regenerate any
already-loaded data.


decentralion:
Good catch. I fixed the test while leaving the structure that iterates over a list of files to check, so it will be easy to add more files back in as we enable more plugins.

Issue #631, opened 2018-08-10
Expanding a pull request crashes the cred explorer


decentralion:
https://github.com/sourcecred/sourcecred/pull/628 introduced a serious bug where exploring a pull request crashes the cred explorer. There are 4 factors, all of which contributed to this bug making it into master:

1. The GitHub plugin depends on the Git plugin (it adds Git commit nodes to the graph)
2. The GitHub plugin's dependency on the Git is not expressed in any way that surfaces when, for example, someone yanks the Git plugin out of the frontend
3. The plugin dispatch code added in #619 throws an uncaught error if the right plugin adapter isn't found
4. We don't have tests that check that the app is robust to unknown nodes in the graph

We should think about how to resolve these issues at root cause. It's weird that the GitHub plugin introduces nodes into the graph that it cannot take responsibility for - because they are namespaced outside of the GitHub plugin's root namespace. Maybe we should consider allowing plugins to provide NodeTypes describing prefixes that are outside of its namespace? Let's think about how to resolve (1) and (2) more. We will also ensure that the cred explorer is robust to issues of this kind going forward.

edit: fixed a semantic typo

decentralion:
Arguably, a 5th cause is that I wasn't sufficiently thorough when testing #628. Given some thought, it does make sense that pull requests are a particular case where the plugins intersect, and I should have expanded them. However, I prefer to depend on automated testing / code invariants as as our approach to avoid regressions like this; in general, we can't expect humans to anticipate all the ways that parts of the system could interact to produce failures.

wchargin:
+1 to all. I briefly considered this case and concluded that it should
be fine. I was wrong, too.

It would have been nice for tests to catch this, but I’m not sure
exactly where that would have happened. It seems to me that the tests
were correctly testing the design, but we had a design flaw—we assumed
that every node would be owned by a plugin. This issue suggests that
perhaps that should not be the case. For example, suppose that the
GitHub plugin can create edges to the Git graph, the Phabricator graph,
and the mailing list graph. I posit that a project may choose not to use
the mailing list plugin, and the code should be robust to that.

Now that we know that this is a design goal, I think that it suffices to
adjust the tests accordingly. One resolution is to add a base plugin
with the empty prefix, and change logic that asserts uniqueness (e.g.,
`pluginAdapter.js`) to pick the most specific plugin. Another resolution
would be to make a bunch of types nullable (e.g., the dispatch methods
[...]

PR #632, opened 2018-08-10
Move core plugin adapter code to its own directory

decentralion:
This commit creates the directory `src/app/adapters` and moves the
following three files into it:
- `src/app/pluginAdapter.js`
- `src/app/pluginAdapter.test.js`
- `src/app/defaultPlugins.js`

This is in preparation for a principled fix for #631, which will add a
base plugin and some logic that ensures it's always included.


wchargin:
We should revert first to put out the fire, instead of adding commits on
top of a broken state.

PR #639, opened 2018-08-11
Tries allow multiple values for the same key

decentralion:
Given the type signature of tries, it is quite natural to allow storing
multiple values for the same key: get always returns an array, so if
multiple values are present for the key, they can all be added to the
array (in insertion order).

This also happens to be the correct semantics for implementing an
AdapterSet; having multiple (say) node types defined for the empty node
addresse is OK, so long as the order in which they are returned is
well-defined.

Test plan: unit tests

### DIFF
--- a/src/core/trie.js
+++ b/src/core/trie.js
@@ -9,9 +9,7 @@ import {
 } from "./graph";
 import * as NullUtil from "../util/null";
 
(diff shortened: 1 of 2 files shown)


decentralion:
Makes sense. Closing this PR.

PR #640, opened 2018-08-11
Add AdapterSet and FallbackAdapter

decentralion:
Issue #631 revealed that our current plugin-handling code is fragile -
we aren't robust to having nodes from a plugin without having that
plugin present in the frontend. This commit adds `StaticAdapterSet` and
`DynamicAdapterSet`, which are abstractions over finding the matching
plugin adapter or type for a node or edge. It's a robust abstraction,
because the adapter sets always include the `StaticFallbackAdapter` or
`DynamicFallbackAdapter`, which can match any node, so we'll never get
an error like #631 due to not having an adapter / type available.

Also relevant: #465

Test plan:
Unit tests included.

### DIFF
--- a/src/app/adapters/adapterSet.js
+++ b/src/app/adapters/adapterSet.js
@@ -1,0 +1,153 @@
+// @flow
+
+import {Graph, type NodeAddressT, type EdgeAddressT} from "../../core/graph";
+import {NodeTrie, EdgeTrie} from "../../core/trie";
+import type {Repo} from "../../core/repo";
+
+import type {
+  StaticPluginAdapter,
+  DynamicPluginAdapter,
+  NodeType,
+  EdgeType,
+} from "./pluginAdapter";
+
+import {FallbackStaticAdapter} from "./fallbackAdapter";
+
+export class StaticAdapterSet {
+  _adapters: $ReadOnlyArray<StaticPluginAdapter>;
+  _adapterNodeTrie: NodeTrie<StaticPluginAdapter>;
+  _adapterEdgeTrie: EdgeTrie<StaticPluginAdapter>;
+  _typeNodeTrie: NodeTrie<NodeType>;
+  _typeEdgeTrie: EdgeTrie<EdgeType>;
+
+  constructor(adapters: $ReadOnlyArray<StaticPluginAdapter>) {
+    this._adapters = [new FallbackStaticAdapter(), ...adapters];
+    this._adapterNodeTrie = new NodeTrie();
+    this._adapterEdgeTrie = new EdgeTrie();
+    this._typeNodeTrie = new NodeTrie();
+    this._typeEdgeTrie = new EdgeTrie();
+    const usedPluginNames = new Set();
+    this._adapters.forEach((a) => {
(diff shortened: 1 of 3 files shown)


src/app/adapters/adapterSet.js
```diff
@@ -0,0 +1,153 @@
...
+  DynamicPluginAdapter,
+  NodeType,
+  EdgeType,
+} from "./pluginAdapter";
+
+import {FallbackStaticAdapter} from "./fallbackAdapter";
+
+export class StaticAdapterSet {
+  _adapters: $ReadOnlyArray<StaticPluginAdapter>;
+  adapterNodeTrie: NodeTrie<StaticPluginAdapter>;
```
wchargin: Can you underscore-name all of these to be clear that they are all private?

decentralion: Done.

src/app/adapters/fallbackAdapter.js
```diff
@@ -0,0 +1,61 @@
...
+  type NodeAddressT,
+  EdgeAddress,
+} from "../../core/graph";
+import type {Repo} from "../../core/repo";
+
+import type {StaticPluginAdapter, DynamicPluginAdapter} from "./pluginAdapter";
+
+export const FALLBACK_NAME = "FALLBACK_ADAPTER";
+
+export class FallbackStaticAdapter implements StaticPluginAdapter {
```
wchargin: Implementation looks good. Do you think that any tests would be useful?

src/app/adapters/adapterSet.js
```diff
@@ -0,0 +1,153 @@
...
+    });
+    this.nodeTypes().forEach((t) => this._typeNodeTrie.add(t.prefix, t));
+    this.edgeTypes().forEach((t) => this._typeEdgeTrie.add(t.prefix, t));
+  }
+
+  adapters(): $ReadOnlyArray<StaticPluginAdapter> {
+    return this._adapters;
+  }
+
+  nodeTypes(): $ReadOnlyArray<NodeType> {
```
wchargin: These can return mutable array type.

src/app/adapters/adapterSet.test.js
```diff
@@ -0,0 +1,214 @@
...
+import {
+  NodeAddress,
+  EdgeAddress,
+  type NodeAddressT,
+  Graph,
+} from "../../core/graph";
+import type {DynamicPluginAdapter} from "./pluginAdapter";
+import {StaticAdapterSet} from "./adapterSet";
+import {FallbackStaticAdapter, FALLBACK_NAME} from "./fallbackAdapter";
+import type {JestMockFn} from "jest";
```
wchargin: This resolves to `any`, which is misleading.

src/app/adapters/adapterSet.js
decentralion: Done

src/app/adapters/fallbackAdapter.js
decentralion: Not really.

PR #641, opened 2018-08-11
In client-side code, only import GraphQL types

wchargin:
Summary:
An `import *` was used for convenience, but this effects a value import
in addition to a type import. By exploding the wildcard import to
directly import the required types, we can shave off 2.3% of our
post-gzip bundle size (131.82 KB to 128.74 KB). It’s unfortunate that we
lose the namespacing, but c’est la vie.

Test Plan:
`yarn flow` suffices.

wchargin-branch: explode-wildcard-type-import

### DIFF
--- a/src/plugins/github/relationalView.js
+++ b/src/plugins/github/relationalView.js
@@ -13,7 +13,16 @@ import type {
   CommentAddress,
   UserlikeAddress,
 } from "./nodes";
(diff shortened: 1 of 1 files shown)


wchargin:
Relevant: https://github.com/facebook/flow/issues/6726

PR #648, opened 2018-08-11
Add pluralized node names

decentralion:
This is preparation for #502 - we want to be able to describe groups of
nodes, e.g. "52 repositories", so we need a plural form for every node
name.

As a fly-by fix, I removed the parentheses around the node names in the
fallback adapter, as these proved to look ugly/inconsistent in the UI.

Test plan: `yarn test` is sufficient.

### DIFF
--- a/src/app/adapters/fallbackAdapter.js
+++ b/src/app/adapters/fallbackAdapter.js
@@ -27,15 +27,20 @@ export class FallbackStaticAdapter implements StaticPluginAdapter {
 
   nodeTypes() {
     return [
-      {name: "(unknown node)", prefix: NodeAddress.empty, defaultWeight: 1},
+      {
+        name: "unknown node",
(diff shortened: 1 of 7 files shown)


src/app/adapters/fallbackAdapter.js
```diff
@@ -27,15 +27,20 @@ export class FallbackStaticAdapter implements StaticPluginAdapter {
...
+      },
     ];
   }
 
   edgeTypes() {
     return [
       {
-        forwardName: "(unknown edge→)",
-        backwardName: "(unknown edge←)",
+        forwardName: "unknown edge→",
```
wchargin: FWIW, I think that these names are confusing. How is a user supposed to
know what `edge→` means? Something like “unknown out-edge” would be
clearer—but no change necessary for this PR.

PR #649, opened 2018-08-11
PagerankTable: separate indentDepth and colorDepth

decentralion:
This commit splits up the "depth" concept in PagerankTable into two
separate values: indentDepth, which determines how many steps the table
is indented, and colorDepth, which determines how many steps of color
shift are applied.

This is in preparation for #502 so that we can correctly indent within
the aggregated summary, without changing the color. Then, color can
represent "changing scope" to a different node, while indenting just
keeps things visually organized.

Test plan: There's no change in observed behavior right now (which I
verified), because the indent depth and color depth always change in
sync. The snapshot test for depth-based styling now shows how the styles
vary based on independent changes in the indent and color depth:
examining that snapshot should persuade the reader that the code works
as intended.

I made some minor improvements to the test code while I was in the area;
these should not be controversial.

### DIFF
--- a/src/app/credExplorer/pagerankTable/Connection.js
+++ b/src/app/credExplorer/pagerankTable/Connection.js
@@ -17,7 +17,8 @@ import {
 } from "./shared";
 
 type ConnectionRowListProps = {|
-  +depth: number,
+  +colorDepth: number,
+  +indentDepth: number,
   +node: NodeAddressT,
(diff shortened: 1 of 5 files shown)


decentralion:
Actually, I want to refactor out a consistent TableRow component before adding more complexity. Ignore this PR for now.

PR #650, opened 2018-08-11
Add Node’s `path.normalize` for the browser

wchargin:
Summary:
This function normalizes paths like `foo/bar//../baz` to `foo/baz`. The
implementation is directly copied from Node’s source code, which is
available under an MIT License.

I looked for a suitable NPM package, and rejected `path-normalize` and
`normalize-path`. (The former is closer and explicitly purports to be
the right thing, but actually isn’t.)

Test Plan:
Unit tests added, with full coverage except for one branch; I include a
proof that the branch is unreachable.

Tested on Node v8.9.4, Node v10.0.0, and Node v10.8.0. Tests pass in
each case. In the latter two cases, inconsistencies between the
implementation and the actual `path.posix.normalize` would cause a test
failure. In the former case, they do not. (All such cases verified.)

wchargin-branch: path-normalize

### DIFF
--- a/src/util/pathNormalize.js
+++ b/src/util/pathNormalize.js
@@ -1,0 +1,324 @@
+// @flow
+
+// This module reimplements Node's `require("path").posix.normalize`
+// function, so that it can be bundled in the browser.
+//
+// The code is copied directly from the Node implementation in
+// `lib/path.js` at Git commit e0395247c899af101f8a1f76a8554be1ff14040a.
+// The code being copied is published under the MIT License.
+//
+// The `normalize` function and its transitive dependencies are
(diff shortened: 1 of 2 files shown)


wchargin:
Okay. The code is substantively from Node, but not wholesale: I had to
cobble together a couple different pieces. It’s possible that I could
have made an error there, but it seems likely that such an error would
cause the whole thing to break, not introduce subtle failures.

If you want to check the proof in the notes, you’re welcome to.

wchargin:
The proof has two minor errors. (The result is correct, but there are
gaps in the argument.) I have corrections and will submit them shortly.

PR #653, opened 2018-08-12
PagerankTable filter defaults to GitHub users

### LINKED ISSUE #651: Make cred explorer default to "USERS" filter
Comparisons across every kind are less relevant - start with the users because that's what humans will be most interested in :)

Should be a simple change to the [default topLevelFilter state value](https://github.com/sourcecred/sourcecred/blob/f1c5d3756d77c3d4a6e550f33e4d7a39c1df6849/src/app/credExplorer/pagerankTable/Table.js#L26)

decentralion:
We humans tend to find information about humans more interesting than
information about commits or pulls. The UI should accomodate this by
defaulting to displaying GitHub user nodes in the cred explorer.

This is implemented as a new nullable argument to the PageRankTable. If
not present, then the filter defaults to showing all nodes. If the
default filter is present but doesn't match any available type, an error
is thrown.

Test plan: The new behavior is tested. Also, I checked it in the UI and
it works.

Closes #651

### DIFF
--- a/src/app/credExplorer/App.js
+++ b/src/app/credExplorer/App.js
@@ -9,6 +9,7 @@ import BrowserLocalStore from "../browserLocalStore";
 import {PagerankTable} from "./pagerankTable/Table";
 import {WeightConfig} from "./WeightConfig";
 import RepositorySelect from "./RepositorySelect";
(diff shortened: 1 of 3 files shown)


decentralion:
Good suggestion. I've implemented it that way, PTAL.

PR #655, opened 2018-08-13
Separate TableRow `depth` from `indent`

decentralion:
Currently, as we expand nodes or connections in the PagerankTable, the
rows both get more indented and attain a deeper color. Both of these
behaviors are controlled by the `depth` parameter.

We're going to switch the UI to a cyclic structure, where as you drill
down, once you get back to a particular node, the indentation resets to
base, but the color - which now indicates nested depth - continues to
change. This commit sets that change up, by splitting the behvaior into
two parameters: `depth`, which controls the color, and `indent`, which
controls the indentation level.

As a small additional tweak, the indentation formula is changed so that
buttons are always indented by 5 pixels. This results in a cleaner
display for nodes that have `depth>0` but `indent==0` (as the button
doesn't look squahsed against the color boundary).

Test plan:
The change is very simple; inspecting the updated snapshots should be
persuasive.

### DIFF
--- a/src/app/credExplorer/pagerankTable/TableRow.test.js
+++ b/src/app/credExplorer/pagerankTable/TableRow.test.js
@@ -8,10 +8,11 @@ import {COLUMNS} from "./sharedTestUtils";
 require("../../testUtil").configureEnzyme();
 
 describe("app/credExplorer/pagerankTable/TableRow", () => {
-  function example(depth: number = 3) {
+  function example() {
     return shallow(
       <TableRow
-        depth={depth}
+        depth={1}
+        indent={1}
         description={<span data-test-description={true} />}
         connectionProportion={0.5}
         score={133.7}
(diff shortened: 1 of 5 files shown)


src/app/credExplorer/pagerankTable/TableRow.test.js
```diff
@@ -8,33 +8,52 @@ import {COLUMNS} from "./sharedTestUtils";
...
+        indent={1}
         description={<span data-test-description={true} />}
         connectionProportion={0.5}
         score={133.7}
         children={<div data-test-children={true} />}
       />
     );
   }
-  it("has depth-based color", () => {
+  it("depth parameter changes the color", () => {
```
wchargin: Would be nice to test that this doesn’t also change the indentation, as
it does before the patch to the main code.

PR #658, opened 2018-08-14
ConnectionRow now shows the connection score

decentralion:
Previously, the ConnectionRow showed the score of the node that was the
source of the connection. I believe the UI will be more consistent and
useful if it instead shows the connection score, i.e. how important that
connection was to the node in scope. This combos well with PR #657.

Test plan: The change is very simple, and covered by unit tests. I also
verified the behavior by examining the cred explorer.


decentralion:
We could remove `sourceScore`, yes. Let's hold off on removing that, since (as discussed offline) there's not much cost to keeping it, and it might be useful later. I intend to take a pass at refactoring the connection data structures once things are more settled.

decentralion:
I added a TODO to consider removing the sourceScore so we don't forget.

Issue #643, opened 2018-08-11
Enable hosting the site at non-root


wchargin:
Various portions of the site currently assume that they are being served
from the root of the host. In other words, we assume that the public
path is `/`. This precludes hosting demos from a personal GitHub
repository like `decentralion/sourcecred-demos`.

A relatively straightforward fix is to thread the public path
configuration through whitherever it is required. Then, we can specify
the public path at build time: for instance, a demo for the Foo project
might be hosted at `--public-path /sourcecred-demos/foo`.

An alternative solution is to use relative paths throughout the
application, so that the public path does not need to be statically
known. This might require some care, but would probably be more robust.
For instance, this would let us serve instances of the site from an IPFS
gateway without needing control over the domain.

wchargin:
Here’s an update about what I'm working on.

As part of enabling our site to be served from arbitrary gateways, I've
created a new implementation of the `history@3` interface, which plugs
directly into our router. Getting the details right is critically
important, and it took some thought to figure out what the right
abstractions were. This, combined with the fact that `history` and
`react-router` together have a grand total of zero type definitions and
zero actual API references or specifications, suggests to me that the
module in question merits extensive automated testing.

Happily, it is the kind of thing for which it is possible to write
unusually good tests—both as core logic tests and a (nearly) end-to-end
test of a full React application, all within a normal Jest environment
(i.e., we don’t have to shell out to Webpack or anything). It’s taking
some time, but it’s going along well.

I considered adding Flow typedefs to `history@3`, which would mitigate
the problem somewhat. I decided not to do so for two reasons. First, the
module in question will _not_ permeate the whole application: its use is
scoped to a few files (approximately two clients). Second, `history@3`,
like many libraries designed without types in mind, is difficult to
write good types for: it has “shortcuts” for “convenience” that
significantly complicate the API to a point that it is difficult to
usefully express in Flow.

If we eventually ditch `react-router` entirely (which might be a good
idea), we’ll still be able to use important parts of this module, which
is nice.

PRs to come. :-)

wchargin:
Another technical update to explain the latest architecture.

Central to this issue is the `Assets` class introduced in #664, which
lets a client resolve a path like `/favicon.png` to either `favicon.png`
or `../favicon.png`, depending on the route from which it’s being
rendered. Implementing this and keeping it in sync with the URL is not
trivial, but it is self-contained.

At some point, however, we need to _use_ the `Assets` resolver. In
[...]

PR #663, opened 2018-08-14
Add a React context for resolving static assets

wchargin:
Summary:
This will enable components to obtain the path to a static asset, even
when the app is not hosted at the root of a server, as outlined in #643.
This module will be used for simple assets (images, etc.) and API data
(fetches from `/api/**`) alike.

Test Plan:
Unit tests included; `yarn test` suffices.

wchargin-branch: assets-context

### DIFF
--- a/src/app/assetsContext.js
+++ b/src/app/assetsContext.js
@@ -1,0 +1,46 @@
+// @flow
+
+import React from "react";
+
+import normalize from "../util/pathNormalize";
+
+/**
+ * Resolver for static assets (e.g., images, PDFs) and API data (e.g.,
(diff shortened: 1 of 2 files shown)


wchargin:
Due to continuing issues with Enzyme, I’m going to rethink the structure
of this PR. `Assets` will stay around, but `AssetsContext` may not. I
have the whole application working just fine, but getting the tests to
work is really awful, and is forcing the production code to be
structured awkwardly. We may want to hold off on using context until
Enzyme gets their act together.

Closing for now; will reopen a similar PR soon.

PR #667, opened 2018-08-15
PagerankTable displays aggregated connections

### LINKED ISSUE #502: Cred explorer: Aggregate over contributions

decentralion:
Previously, expanding a node would display the individual connections
that contributed cred to that node. For nodes with high degree, this was
a pretty noisy UI.

Now, expanding a node displays "aggregations": for every type of
adjacent connection (where type is the union of the edge type and the
adjacent node type), we show a summary of the total cred from
connections of that type. The result is a much more managable summary
view. Naturally, these aggregations can be further expanded to see the
individual connections.

Closes #502.

Test plan: The new behavior is unit tested. You can also launch the cred
explorer and experience the UI directly. I have used the new UI a lot,
as well as demo'd it to people, and I like it quite a bit.

### DIFF
--- a/src/app/credExplorer/pagerankTable/Aggregation.js
+++ b/src/app/credExplorer/pagerankTable/Aggregation.js
@@ -1,0 +1,111 @@
+// @flow
+
+import React from "react";
+import * as NullUtil from "../../../util/null";
+
+import type {NodeAddressT} from "../../../core/graph";
+import {ConnectionRowList} from "./Connection";
+
+import {aggregateFlat, type FlatAggregation, aggregationKey} from "./aggregate";
+
+import {Badge, type SharedProps} from "./shared";
+import {TableRow} from "./TableRow";
+
+type AggregationRowListProps = {|
+  +depth: number,
+  +node: NodeAddressT,
+  +sharedProps: SharedProps,
+|};
+
+export class AggregationRowList extends React.PureComponent<
+  AggregationRowListProps
+> {
+  render() {
+    const {depth, node, sharedProps} = this.props;
+    const {pnd, adapters} = sharedProps;
+    const {scoredConnections} = NullUtil.get(pnd.get(node));
+    const aggregations = aggregateFlat(
+      scoredConnections,
+      adapters.static().nodeTypes(),
+      adapters.static().edgeTypes()
+    );
+    return (
+      <React.Fragment>
+        {aggregations.map((agg) => (
+          <AggregationRow
+            key={aggregationKey(agg)}
+            depth={depth}
+            target={node}
+            sharedProps={sharedProps}
+            aggregation={agg}
+          />
+        ))}
+      </React.Fragment>
+    );
+  }
+}
+
+type AggregationRowProps = {|
+  +depth: number,
+  +target: NodeAddressT,
+  +aggregation: FlatAggregation,
+  +sharedProps: SharedProps,
+|};
+
+export class AggregationRow extends React.PureComponent<AggregationRowProps> {
+  render() {
+    const {sharedProps, target, depth, aggregation} = this.props;
+    const {pnd} = sharedProps;
+    const score = aggregation.summary.score;
+    const {score: targetScore} = NullUtil.get(pnd.get(target));
+    const connectionProportion = score / targetScore;
+
+    return (
+      <TableRow
+        depth={depth}
+        indent={1}
+        showPadding={false}
+        connectionProportion={connectionProportion}
+        score={score}
+        description={<AggregationView aggregation={aggregation} />}
+      >
+        <ConnectionRowList
+          key="children"
+          depth={depth}
+          node={target}
+          connections={aggregation.connections}
+          sharedProps={sharedProps}
+        />
+      </TableRow>
+    );
+  }
+}
+
+export class AggregationView extends React.PureComponent<{|
+  +aggregation: FlatAggregation,
+|}> {
+  render() {
+    const {aggregation} = this.props;
+    const {connectionType, summary, nodeType} = aggregation;
+    function connectionDescription() {
+      switch (connectionType.type) {
+        case "SYNTHETIC_LOOP":
+          return "synthetic loop";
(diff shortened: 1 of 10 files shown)


decentralion:
**Before:**
image

**After:**
image

src/app/credExplorer/pagerankTable/Aggregation.js
```diff
@@ -0,0 +1,135 @@
...
+  render() {
+    const {depth, node, sharedProps} = this.props;
+    const {pnd, adapters} = sharedProps;
+    const {scoredConnections} = NullUtil.get(pnd.get(node));
+    const aggregations = aggregateFlat(
+      scoredConnections,
+      adapters.static().nodeTypes(),
+      adapters.static().edgeTypes()
+    );
+    function aggregationKey(aggregation: FlatAggregation) {
```
wchargin: I don’t understand the implications of this function. Why can’t its
implementation just be

```js
    function aggregationKey(aggregation: FlatAggregation) {
      const result = {
        nodePrefix: aggregation.nodeType.prefix,
        connectionType: aggregation.connectionType.type,
      };
      return stringify(result);
    }
```

? This passes Flow and tests, and seems to work under `yarn start`. Are
additional considerations that should be taken into account?

src/app/credExplorer/pagerankTable/Aggregation.js
```diff
@@ -0,0 +1,135 @@
...
+        case "OUT_EDGE":
+          result.connectionType = {
+            type: "OUT_EDGE",
+            edgePrefix: aggregation.connectionType.edgeType.prefix,
+          };
+          break;
+        default:
+          throw new Error((aggregation.connectionType.type: empty));
+      }
+      return JSON.stringify(result);
```
wchargin: Should this be `json-stable-stringify`?

src/app/credExplorer/pagerankTable/Aggregation.js
```diff
@@ -0,0 +1,135 @@
...
+        case "OUT_EDGE":
+          return connectionType.edgeType.forwardName;
+        default:
+          throw new Error((connectionType.type: empty));
+      }
+    }
+    return (
+      <span>
+        <Badge>{connectionDescription()}</Badge>
+        <span> {summary.size} </span>
```
wchargin: Could we write `<span> {summary.size}× </span>` here, so that we see
“800× comment” instead of “800 comment”?

src/app/credExplorer/pagerankTable/Aggregation.js
```diff
@@ -0,0 +1,135 @@
...
+        <span style={{textTransform: "lowercase"}}>{nodeType.name}</span>
```
wchargin: This transformation is highly questionable in general, and we will
probably want to remove it later, but it is an improvement over not
having it right now, so I’m okay with including it.

src/app/credExplorer/pagerankTable/Aggregation.test.js
```diff
@@ -0,0 +1,217 @@
...
+    });
+  });
+  describe("AggregationView", () => {
+    const nodeType: NodeType = {
+      name: "whatDoes",
+      defaultWeight: 1,
+      prefix: NodeAddress.empty,
+    };
+    const edgeType: EdgeType = {
+      forwardName: "marcellus",
```
wchargin: important correction: `s/marcellus/marsellus/`

src/app/credExplorer/pagerankTable/Aggregation.js
decentralion: This would not be unique. Note that the connection type is just a string in `"IN_EDGE" | "OUT_EDGE" | "SYNTHETIC_LOOP"`. Because the edge prefix is not included, we can get a conflict when there are multiple edge types pointing to the same node type.

I've factored `aggregationKey` into an exported function from `aggregate.js`, and tested it. Using your proposed implementation would properly fail the test. (This happens in a separate commit; I'm planning to rebase and merge them)

src/app/credExplorer/pagerankTable/Aggregation.js
decentralion: Good point;fixed.

src/app/credExplorer/pagerankTable/Aggregation.js
decentralion: I changed it to use the plural names instead.

src/app/credExplorer/pagerankTable/Aggregation.test.js
decentralion: Fixed.

src/app/credExplorer/pagerankTable/Aggregation.js
decentralion: Ack

src/app/credExplorer/pagerankTable/Aggregation.js
wchargin: Great. Thanks for the explanation.

PR #679, opened 2018-08-15
Load plugin data even when hosted at non-root

wchargin:
Summary:
This commit approximately completes the implementation of #643.\* Plugin
adapters are now provided an `Assets` object at `load` time, which they
can use to resolve their plugin-specific API routes.

\* “Approximately” because there are some non-essential pieces of legacy
code that should be cleaned up.

Test Plan:
Unit tests modified, but it would be good to also manually test this.
Run `./scripts/build_static_site.sh` to build the site to, say,
`/tmp/gateway/`. Then spin up a static HTTP server serving `/tmp/` and
navigate to `/gateway/` in the browser. Note that the entire application
works.

wchargin-branch: use-assets-in-PluginAdapters


decentralion:
As discussed offline - this feature merits an entry in the changelog, so put it in now?

PR #680, opened 2018-08-16
Rename cred explorer table columns

decentralion:
The 'Score' column is renamed to 'Cred' (and its prop is renamed as
well). The column which shows how a connection or aggregation
contributes to a node's cred, as a percentage, has been rendered
nameless. It is pretty self explanatory, and the previous name
("Connection") was meaningless.

Test plan: Unit tests, also I inspected the frontend.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Rename cred explorer table columns (#680)
```
wchargin: Should this go at the bottom, for consistency? We can `tac` it when we
commit a release.

PR #682, opened 2018-08-16
Cleanup unnecessary lint suppresions

decentralion:
We often construct case statements over union-typed variables, and then
in the default case, we use a `(type: empty)` assertion to ensure that
failing to account for all the cases results in a flow error.

In the past, we created an extra line for this assertion, which required
some eslint suppressions. We've realized it's cleaner to inline the type
assertion in the runtime error that we throw in these defaults.

This code cleans everything to the new style, and removes every existing
`// no-unused-expressions` invocation in the codebase.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/git/edges.js
+++ b/src/plugins/git/edges.js
@@ -237,8 +237,6 @@ export function toRaw(x: StructuredAddress): RawAddress {
         ...lengthEncode(GitNode.toRaw(x.treeEntry))
       );
     default:
-      // eslint-disable-next-line no-unused-expressions
-      (x.type: empty);
-      throw new Error(x.type);
+      throw new Error((x.type: empty));
   }
 }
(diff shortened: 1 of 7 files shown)


src/plugins/git/edges.js
```diff
@@ -140,7 +140,7 @@ function multiLengthDecode(x: $ReadOnlyArray<string>, fail: () => Error) {
 }
 
 export function fromRaw(x: RawAddress): StructuredAddress {
-  function fail() {
+  function fail(_unused_assertion: any) {
```
wchargin: This particular change makes the code less clear. We shouldn’t add an
extra parameter to a function just to avoid a lint suppression.

The best version that I can think of is the original code.

Same comment applies to the other three `fail` functions.

PR #684, opened 2018-08-16
Radically simplify the git plugin

decentralion:
This commit radically simplifies the git plugin as described in [1].
Basically, we remove all the node types except for COMMIT and all of the
edge types except for HAS_PARENT (which links commits).

Implementing this patch was simple. First, I removed node types from
`git/nodes.js` and edge types from `git/edges.js`. Then, I iteratively
removed all the flow errors, which removed most of the code built on top
of those types. Finally, I updated the example repository data via
`src/plugins/git/loadRepositoryTest.sh` and fixed a few failing unit
tests.

Test plan: `yarn travis --full` passes. Also, patching the application
to use the new reduced Git plugin results in a cred explorer which
nicely displays Git commits (rather than rendering them as unknown
nodes), and links the Git commits together by their parent
relationships.

Here's the patch for using the reduced Git plugin. Note that if you
apply this, you should reset your `$SOURCECRED_DIRECTORY`. You should
also reset your application localStore state.

```diff
diff --git a/src/app/adapters/defaultPlugins.js b/src/app/adapters/defaultPlugins.js
index f9edaf3..4e1c11a 100644
--- a/src/app/adapters/defaultPlugins.js
+++ b/src/app/adapters/defaultPlugins.js
@@ -2,7 +2,8 @@

 import {StaticAdapterSet} from "./adapterSet";
 import {StaticPluginAdapter as GithubAdapter} from "../../plugins/github/pluginAdapter";
+import {StaticPluginAdapter as GitAdapter} from "../../plugins/git/pluginAdapter";

 export function defaultStaticAdapters(): StaticAdapterSet {
-  return new StaticAdapterSet([new GithubAdapter()]);
+  return new StaticAdapterSet([new GithubAdapter(), new GitAdapter()]);
 }
diff --git a/src/cli/common.js b/src/cli/common.js
index 4d4c87c..eb6f639 100644
--- a/src/cli/common.js
+++ b/src/cli/common.js
@@ -10,7 +10,7 @@ export function pluginNames(): PluginName[] {
 }

 export function defaultPlugins(): PluginName[] {
-  return ["github"];
+  return ["github", "git"];
 }

 function defaultStorageDirectory() {
```
[1]: https://github.com/sourcecred/sourcecred/issues/627#issuecomment-413435447

PR #683, opened 2018-08-16
Fork the git plugin into deprecated-git

decentralion:
Splitting out a current copy of the code, in advance of a radical
simplification as described in [1].

This patch was generated by copying the `plugins/git` directory, and
then modifying instances of the string `"/git/"` to
`"/deprecated-git/"`. Most notably, the file
`plugins/deprecated-git/loadRepositoryTest.sh` has been updated so that
it's example data file points to a data file in `deprecated-git`, and
the corresponding `loadAndPrintGitRepository` `bin/` target has been
forked so there is now a version specific to the deprecated plugin.

Test plan: `yarn travis --full` works, as does
`src/plugins/deprecated-git/loadRepositoryTest.sh`.

[1]: https://github.com/sourcecred/sourcecred/issues/627#issuecomment-413435447

PR #689, opened 2018-08-16
Add `createMinimalGraph` for a tiny git graph

decentralion:
This implements the approach suggested in [1]. Instead of forking the
git plugin entirely, we'll fork the createGraph method and the
pluginAdapter so that we have instances that produce a lightweight git
graph.

createMinimalGraph is a fork of createGraph that only adds commit nodes
and has_parent edges. New unit tests ensure that only the whitelisted
nodes and edges appear.

Supersedes #683 and #684.

Test plan: `yarn test`

[1]: https://github.com/sourcecred/sourcecred/issues/627#issuecomment-413623784

### DIFF
--- a/src/plugins/git/createMinimalGraph.js
+++ b/src/plugins/git/createMinimalGraph.js
@@ -1,0 +1,41 @@
+// @flow
+
+import {Graph} from "../../core/graph";
+
+import * as GT from "./types";
+import * as GN from "./nodes";
+import * as GE from "./edges";
+
+export function createMinimalGraph(repository: GT.Repository): Graph {
+  const creator = new MinimalGraphCreator();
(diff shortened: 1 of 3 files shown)


src/plugins/git/createMinimalGraph.js
```diff
@@ -0,0 +1,41 @@
...
+import * as GE from "./edges";
+
+export function createMinimalGraph(repository: GT.Repository): Graph {
+  const creator = new MinimalGraphCreator();
+  creator.addRepository(repository);
+  return creator.graph;
+}
+
+class MinimalGraphCreator {
+  graph: Graph;
```
wchargin: Can be read-only.

PR #692, opened 2018-08-16
Add `GitState`, `Environment` to the `VersionInfo`

wchargin:
Summary:
The version number displayed in the application now displays much more
specific information. It now lists the Git commit from which the build
was constructed, and will identify whether we have accidentally deployed
a development instance (which would be slow) or an instance with
uncommitted changes (which would be bad).

The version information is computed during the initialization of the
Webpack config. For development, this means that it is computed when you
run `yarn start`, and not updated thenafter. If the stale information
presents actual confusion, we would need to backport Webpack 4’s support
for runtime values in `DefinePlugin` to Webpack 3 (or upgrade Webpack
by a major version).

Test Plan:
The logic for `GitState` and `Environment` has existing tests. With both
a clean tree and a dirty tree, run `yarn start` and build the static
site, and check that the resulting versions are correct.

wchargin-branch: use-rich-version-types

### DIFF
--- a/config/env.js
+++ b/config/env.js
@@ -1,7 +1,12 @@
 // @flow
+const {spawnSync, execFileSync} = require("child_process");
 const fs = require("fs");
+const stringify = require("json-stable-stringify");
 const path = require("path");
+
 const paths = require("./paths");
+
+/*:: import type {GitState} from "../src/app/version"; */
 
 // Make sure that including paths.js after env.js will read .env variables.
 delete require.cache[require.resolve("./paths")];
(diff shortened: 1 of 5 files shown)


config/env.js
```diff
@@ -55,11 +60,69 @@ process.env.NODE_PATH = (process.env.NODE_PATH || "")
...
+
+  const commitTimestamp = execFileSync(
+    "git",
+    [
+      "-C",
+      __dirname,
+      "show",
+      "--no-patch",
+      "--format=%cd",
+      "--date=format:%Y%m%d-%M%S",
```
decentralion: might be nice to know the hour too

PR #695, opened 2018-08-17
build_static_site.sh: create target if nonexistent

wchargin:
Summary:
The current version of the build script has the safe but annoying
property that the target directory must be an existing, empty directory.
It seems reasonable and convenient to allow the build script to create
the directory with `mkdir -p`. It still fails if the directory is not
empty or is a file.

Test Plan:
Unit tests updated; run `yarn sharness-full`.

wchargin-branch: build-mkdir-p

### DIFF
--- a/scripts/build_static_site.sh
+++ b/scripts/build_static_site.sh
@@ -45,9 +45,7 @@ parse_args() {
                 fi
                 shift
                 if [ $# -eq 0 ]; then die 'missing value for --target'; fi
-                if ! target="$(readlink -e "$1")"; then
-                    die "target does not exist: $1"
-                fi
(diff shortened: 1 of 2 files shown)


wchargin:
> How would you feel about adding a --force option to overwrite if it
> currently exists?

That’d be fine with me. This raises an existing issue regarding testing:
it’s easy to test the failure cases, but it’s not easy to test the
success case without also spending ~30s actually building the site.
Something like a “validate args only” would solve the problem, but
doesn’t actually make sense to include; perhaps a “dry run” mode is what
we want?

All this for a separate change, though; I’ll merge this as is.

PR #699, opened 2018-08-22
Retry GitHub queries with exponential backoff

wchargin:
Summary:
This patch adds independent exponential backoff to each individual
GitHub GraphQL query. We remove the fixed `GITHUB_DELAY_MS` delay before
each query in favor of this solution, which requires no additional
configuration (thus resolving a TODO in the process).

We use the NPM module `retry` with its default settings: namely, a
maximum of 10 retries with factor-2 backoff starting at 1000ms.
Empirically, it seems very unlikely that we should require much more
than 2 retries for a query. (See Test Plan for more details.)

This is both a short-term unblocker and a good kind of thing to have in
the long term.

Test Plan:
Note that `yarn test --full` passes, including `fetchGithubRepoTest.sh`.
Consider manual testing as follows.

Add `console.info` statements in `retryGithubFetch`, then load a large
repository like TensorFlow, and observe the output:

```shell
$ node bin/sourcecred.js load --plugin github tensorflow/tensorflow 2>&1 | ts -s '%.s'
0.252566 Fetching repo...
0.258422 Trying...
5.203014 Trying...
[snip]
1244.521197 Trying...
1254.848044 Will retry (n=1)...
1260.893334 Trying...
1271.547368 Trying...
1282.094735 Will retry (n=1)...
1283.349192 Will retry (n=2)...
1289.188728 Trying...
[snip]
1741.026869 Ensuring no more pages...
1742.139978 Creating view...
1752.023697 Stringifying...
1754.697116 Writing...
1754.697772 Done.
```

This took just under half an hour, with 264 queries total, of which:
  - 225 queries required 0 retries;
  - 38 queries required exactly 1 retry;
  - 1 query required exactly 2 retries; and
  - 0 queries required 3 or more retries.

wchargin-branch: github-backoff

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -5,6 +5,7 @@
  */
 
 import fetch from "isomorphic-fetch";
+import retry from "retry";
 
 import {stringify, inlineLayout} from "../../graphql/queries";
 import {createQuery, createVariables, postQueryExhaustive} from "./graphql";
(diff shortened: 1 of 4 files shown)


wchargin:
Interested parties may fetch the resulting TensorFlow graph from IPFS
via either of the following:

  - `ipfs cat QmQYXQBBYCu6QkkRt31EUHZ128kP9PTyvBMqssC8vtELZ4 | gzip -d`
  - `ipfs cat QmbgBtjprTZTGM9LDYUVbS2dLRA99zuTsJ8D4i9roZwHEy`.

More details: <https://gist.github.com/wchargin/2e4352b5e875cf764e1ea64f56ed13b7>

src/plugins/github/fetchGithubRepo.js
```diff
@@ -49,6 +50,75 @@ export default function fetchGithubRepo(
...
+              ({
+                type: "GITHUB_INTERNAL_EXECUTION_ERROR",
+                retry: true,
+                error: x,
+              }: GithubResponseError)
+            );
+          } else {
+            return Promise.reject(
+              ({
+                type: "GRAPHQL_ERROR",
```
decentralion: Hmm, so if GitHub changed the message they put on their timeout failure, we'll start registering them as GRAPHQL_ERRORs. in that case, it would reject without retry and the error would be printed to console, and we would quickly see what went wrong. so I think this is fine.

wchargin: Precisely correct. I agree that this is fine. If we start getting
spurious errors for whatever reason, we can of course revisit this
logic.

Issue #705, opened 2018-08-24
Better cred normalization algorithm


decentralion:
Currently, "cred" is computed by normalizing the scores, so that the highest score in the graph has a score of 1000. Often times, this node is the repository node. Since we are mostly interested in users' cred, we should probably normalize based on the users' scores explicitly.

I'm considering two alternatives:
1. Normalize based on the highest user's score
2. Normalize based on the sum of the users' scores

1 has some nice properties, namely that the displayed cred values are not "zero sum".
However, I think 2 is a more principled approach, and makes for easier & more meaningful comparison of numbers across repositories.

So if no one objects I will likely go with option 2. 

cc: @wchargin

wchargin:
> Often times, this node is the repository node. Since we are mostly
> interested in users' cred, we should probably normalize based on the
> users' scores explicitly.

I don’t follow.

What’s wrong with normalizing against the highest score in the graph?

So, sometimes the top user has score 1000, and sometimes the top user
has score 850. Why is that a problem?

> I'm considering two alternatives:
> 
>  1. Normalize based on the highest user's score
>  2. Normalize based on the sum of the users' scores

(1) has the minor downside that the user with the most cred is not
strictly numerically incentivized to earn more cred—by earning more
cred, they just push others downward. This doesn’t seem very nice.

(2) is reasonable, but interesting in the context of a long-tailed
distribution like user cred. Recall, for instance, that long-tailed
distributions [can have infinite moments!][pareto] I don’t object to
trying it out.

[pareto]: https://en.wikipedia.org/wiki/Pareto_distribution#Properties

decentralion:
> So, sometimes the top user has score 1000, and sometimes the top user
> has score 850. Why is that a problem?

If it only varied in that range, it would be OK. What about when the top user has a score of 100? Or 12? Or 2? Especially as most users won't look at the repository node, it will add confusion and make results intuitively less comparable across repositories.

wchargin:
> What about when the top user has a score of 100? Or 12? Or 2?

This sounds like interesting signal to me. What is 10 or 80 or 500 times
more important than the most important user? Why do we want to hide that
information?

> less comparable across repositories.

In what ways is cred meaningfully comparable across repositories now?

Ratios are meaningfully comparable: Alice has twice as much cred as Bob
in Project A and half as much cred as Carol in Project 2. This
comparison makes sense regardless of the normalization.

Is it meaningful to compare absolutes—that Alice has twice as much cred
[...]

PR #708, opened 2018-08-29
Implement scoreByConstantTotal

decentralion:
This commit adds the logic for computing scores so that the total score,
summed across all nodes matching a NodePrefix, is a fixed constant.

See #705 for context.

Test plan: The logic is quite simple, and adequate unit tests are
included.

Note to reviewer: There is a spurious whitespace diff in the test file
because the tests for the previous test block were not correctly scoped.

### DIFF
--- a/src/core/attribution/nodeScore.js
+++ b/src/core/attribution/nodeScore.js
@@ -1,10 +1,11 @@
 // @flow
 
-import type {NodeAddressT} from "../graph";
+import {NodeAddress, type NodeAddressT} from "../graph";
 import type {NodeDistribution} from "./graphToMarkovChain";
 
 export type NodeScore = Map<NodeAddressT, number>;
 
+/* Normalize scores so that the maximum score has a fixed value */
 export function scoreByMaximumProbability(
   pi: NodeDistribution,
   maxScore: number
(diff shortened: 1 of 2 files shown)


src/core/attribution/nodeScore.js
```diff
@@ -26,3 +27,34 @@ export function scoreByMaximumProbability(
...
+
+  let unnormalizedTotal = 0;
+  for (const [addr, prob] of pi) {
+    if (NodeAddress.hasPrefix(addr, nodeFilter)) {
+      unnormalizedTotal += prob;
+    }
+  }
+  if (unnormalizedTotal === 0) {
+    throw new Error(
+      "Invariant violation: Tried to normalize based on nodes with no score. " +
```
wchargin: If it's reachable, it's not an invariant violation, it's an invalid argument.

src/core/attribution/nodeScore.js
```diff
@@ -26,3 +27,34 @@ export function scoreByMaximumProbability(
...
+  let unnormalizedTotal = 0;
+  for (const [addr, prob] of pi) {
+    if (NodeAddress.hasPrefix(addr, nodeFilter)) {
+      unnormalizedTotal += prob;
+    }
+  }
+  if (unnormalizedTotal === 0) {
+    throw new Error(
+      "Tried to normalize based on nodes with no score. " +
+        "This probably means that there were no nodes matching prefix: " +
```
wchargin: FYI, I just hit this on a real repo (wchargin/varese), because it has no
GitHub activity. No change requested, but we should note that this can
happen even if you don’t hack up your graphs manually.

PR #709, opened 2018-08-29
Users have 1000 cred in aggregate

### LINKED ISSUE #705: Better cred normalization algorithm

decentralion:
This commit changes the cred normalization algorithm so that the total
cred of all GitHub user nodes always sums to 1000. For rationale on the
change, see #705.

Fixes #705.

Test plan: There is an included unit test which verifies that the
new argument gets threaded through the state properly. But this is
mostly a config change, so it's best tested by actually inspecting
the cred explorer. I have done so, and can verify that the behavior is
as expected: the sum of users' cred now sums to 1000, and e.g. modifying
the weight on the repository node doesn't produce drastic changes to
cred scores.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Normalize scores so that 1000 cred is split amongst users (#STOPSHIP)
```
wchargin: Update STOPSHIP?

PR #711, opened 2018-08-29
Allow combining repositories at load time

wchargin:
Summary:
As a first pass toward support for analyzing whole organizations, we
allow loading multiple repositories with `sourcecred load`, combining
them into a single relational view and a single Git graph at load time.

Test Plan:
Run

```
node bin/sourcecred.js \
    load \
    sourcecred/example-git \
    sourcecred/example-github \
    sourcecred/sourcecred \
    --output sourcecred/examples \
    ;
```

and select `sourcecred/examples` from the web view. Filter “Repository”
nodes, and note that there are three.

Note that loading a single repository without `--output` still works,
that loading a single repository with `--output` still works (respecting
the alias name), and loading not exactly one repository without
`--output` yields an appropriate error message.

Note that `yarn sharness-full` still works.

wchargin-branch: load-combined

### DIFF
--- a/src/cli/commands/load.js
+++ b/src/cli/commands/load.js
@@ -30,11 +30,14 @@ const execDependencyGraph = require("../../tools/execDependencyGraph").default;
 export default class PluginGraphCommand extends Command {
   static description = "load data required for SourceCred";
 
+  static strict = false;
+
   static args = [
     {
-      name: "repo",
+      name: "repos",
       required: true,
-      description: "the GitHub repo to load, represented as OWNER/NAME",
+      description:
+        "GitHub repos to load (one per argument), represented as OWNER/NAME",
     },
   ];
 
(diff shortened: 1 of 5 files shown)


src/cli/commands/load.js
```diff
@@ -58,30 +67,46 @@ export default class PluginGraphCommand extends Command {
...
+        output: defaultOutput,
         "github-token": githubToken,
         "sourcecred-directory": basedir,
         "max-old-space-size": maxOldSpaceSize,
         plugin,
       },
     } = this.parse(PluginGraphCommand);
-    const repo = stringToRepo(args.repo);
+    const repos = argv.map((s) => stringToRepo(s));
+    const outputRepo =
```
decentralion: I dislike nested ternary expressions, as I find them hard to read. (And nested ternary expressions with IIFEs are right out.)

wchargin: Changed, but for the record I think that nested conditional expressions
get way more hate than they deserve. :-)

PR #717, opened 2018-08-29
Add logo variants for CredBot and Discourse

wchargin:
Summary:
The logos are the same as the original, but with different colors for
the field.

Test Plan:
None.

wchargin-branch: credbot-discourse-logos


wchargin:
I’ll set the Discourse logo once this is actually merged, so that I can
link to the file in `master`.

Issue #588, opened 2018-08-02
Make explorer weight sliders prettier/nicer


decentralion:
The sliders are a bit ugly right now, IMO.

Also, I think the range on weights is too high - I think limiting to +/- 5 would be sufficient. (+5 = 32x more important than default)

decentralion:
I'm closing this as obsolete. We've made a number of improvements to weight sliders since this issue was opened.

PR #736, opened 2018-08-30
Better handling of log weights

decentralion:
This commit isolates all of the log-weight behavior in the weight
slider. That slider moves in log space, but the numbers printed and
passed around the WeightConfig code are now always in linear-space.

This should reduce confusion in the UI and for developers.

This commit contains two other improvements: (#588)
- Changes the (log space) range on the sliders from ±10 to ±5
- Change the order from slider, weight, name to name, slider, weight, so
that there is more visual separation between the name and the weight.

Test plan: Changes to the weight slider are tested. Changes to the
WeightConfig aren't (#604) so I manually tested the UI.


src/app/credExplorer/weights/WeightSlider.js
```diff
@@ -9,28 +9,36 @@ export class WeightSlider extends React.Component<{|
...
+            this.props.onChange(2 ** logValue);
           }}
         />{" "}
-        <span>{formatWeight(this.props.weight)}</span>
-        <span>{this.props.name}</span>
+        <span
+          style={{minWidth: 45, display: "inline-block", textAlign: "right"}}
+        >
+          {formatWeight(this.props.weight)}
+        </span>{" "}
```
wchargin: Why the trailing space?

src/app/credExplorer/weights/WeightSlider.js
```diff
@@ -9,28 +9,36 @@ export class WeightSlider extends React.Component<{|
...
       </label>
     );
   }
 }
 
 export function formatWeight(n: number) {
-  let x = n.toFixed(1);
-  if (!x.startsWith("-")) {
-    x = "+" + x;
+  if (n < 0 || !isFinite(n)) {
```
wchargin: Let's disallow zero. Otherwise, (e.g.) you get a `value={-Infinity}` on the slider.

Or, allow zero, but consistently and with tests.

decentralion: Done.

PR #741, opened 2018-08-31
cli: add `main`, `sourcecred`, and `help`

wchargin:
Summary:
This commit includes a minimal usage of an actual CLI application. It
provides the `help` command and no actual functionality.

Test Plan:
Unit tests added, with full coverage. To see it in action, first run
`yarn backend`, then run `node bin/cli.js help`.

wchargin-branch: cli-beginnings

### DIFF
--- a/src/cli/sourcecred.js
+++ b/src/cli/sourcecred.js
@@ -1,0 +1,29 @@
+// @flow
+// Implementation of the root `sourcecred` command.
+
+import type {Command} from "./command";
+
+import {VERSION_SHORT} from "../app/version";
+
+import help from "./help";
+
+const sourcecred: Command = async (args, std) => {
+  if (args.length === 0) {
+    help([], {out: std.err, err: std.err});
+    return 1;
+  }
+  switch (args[0]) {
+    case "--version":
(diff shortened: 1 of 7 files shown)


src/cli/sourcecred.js
```diff
@@ -0,0 +1,27 @@
...
+import help from "./help";
+
+const sourcecred: Command = async (args, std) => {
+  if (args.length === 0) {
+    help([], {out: std.err, err: std.err});
+    return 1;
+  }
+  switch (args[0]) {
+    case "--version":
+      std.out("sourcecred version 0.1.0");
```
decentralion: Please import the version string

wchargin:
> Please import the version string [from `src/app/version.js`]

Discussed offline. I posited that the app version and CLI version should
not necessarily be tied to one another. But @decentralion pointed out
that as long as we’re in a single repo and have a single system of tags,
it makes sense to couple them. It also prevents potential confusion: if
we bump the app version but not the CLI version, then users might run
`--version` and think that they’re on an older version.

I’m persuaded. I’ll implement this, but it depends on #746.

wchargin:
Changes made; please take a look.

PR #749, opened 2018-08-31
Configure forward/backward edge weights separately

### LINKED ISSUE #596: Remove "Edge directionality" from Weight Sliders
Separating the edge directionality from the edge weights creates confusion. Removing directionality as a separate category and instead incorporating it into the edge weights provides a clearer UI.

decentralion:
This commit introduces a new component, `EdgeTypeConfig`, which is
responsible for configuring the weights for a given edge type. The
config creates two `WeightSlider`s: one for the forward direction, and
one for the backward direction. The `DirectionalitySlider` is no longer
used, and is removed. This fixes #596.

The EdgeTypeConfig is tested, so this is continuing progress towards
solving #604.

Test plan: I manually verified that modifying edge weights has the
expected effect on cred scores. Also, some new unit tests are included.


wchargin:
> This component renders the "forward name" with the "backward weight"
> and vice versa. While counter-intuitive, this is the correct UI
> decision.

I am highly skeptical of this claim.

First, this change makes it much more confusing to reason about what is
happening in the actual graph and Markov chain, because it does
literally the opposite of what it says.

Second, recall that we want to and plan to add the dual view of the cred
explorer wherein you see where nodes flow cred _to_ instead of where
they get it _from_. If we make this change, then _that_ view will appear
to be backward.

That is, this decision (1) is confusing and (2) only punts the problem
down the road.

We need to be explicit about what is going on. Suppose that we have a
display like this:

> Flow cred from **B** to **A** when:
>
> |                              |             |      |
> | :--                          | :-:         | --:  |
> | **A** has child **B**        | `----o----` | 1×   |
> | **A** has parent **B**       | `----o----` | 1×   |
> | **A** references **B**       | `---o-----` | 1/2× |
> | **A** is referenced by **B** | `------o--` | 4×   |

This is now entirely unambiguous. If we switch to the dual view, we can
swap the **A** and the **B** (perfectly clear), or swap the sliders
(might be less jarring). Neither of these options would be viable in the
current model because there would be no indication that anything had
actually changed—the UI would just be showing different weigts!

This is one solution, but I’m sure it’s not the only one, and I’m not
dead-set on it: as long as a solution makes explicit and unambiguous
what the user’s transformations actually _mean_, I’m comfortable with
it.

wchargin:
Needs rebase?

src/app/credExplorer/weights/EdgeTypeConfig.js
```diff
@@ -0,0 +1,104 @@
...
+    backwardWeight: 1,
+  };
+}
+
+export class EdgeTypeConfig extends React.Component<{
+  +weightedType: WeightedEdgeType,
+  +onChange: (WeightedEdgeType) => void,
+}> {
+  render() {
+    // This component renders the "forward name" with the "backward weight" and
```
wchargin: This comment is obsolete, right?

src/app/credExplorer/weights/EdgeTypeConfig.js
```diff
@@ -0,0 +1,104 @@
...
+          }}
+        />
+      </div>
+    );
+  }
+}
+
+export function styledVariable(letter: string) {
+  return (
+    <span style={{fontWeight: 700, fontStyle: "italic", marginRight: 3}}>
```
wchargin: Please indicate that the `marginRight` is an _italic correction_.

Really, the unit of measurement should be `em`, not `px`, because the
dimension scales (linearly) with the font size. Something like `0.1em`
would probably be a good starting point.

src/app/credExplorer/weights/EdgeTypeConfig.js
```diff
@@ -0,0 +1,104 @@
...
+      {letter}
+    </span>
+  );
+}
+
+export class EdgeWeightSlider extends React.Component<WeightSliderProps> {
+  render() {
+    const modifiedName = (
+      <span style={{display: "flex"}}>
```
wchargin: This `flex` seems superfluous; remove?

Can the whole outer `span` be a `React.Fragment`?

src/app/credExplorer/weights/EdgeTypeConfig.test.js
```diff
@@ -0,0 +1,87 @@
...
+  describe("EdgeWeightSlider", () => {
+    function example() {
+      const onChange = jest.fn();
+      const element = shallow(
+        <EdgeWeightSlider weight={3} name="foo" onChange={onChange} />
+      );
+      const weightSlider = element.find(WeightSlider);
+      return {element, onChange, weightSlider};
+    }
+    it("renders the name along with some greek characters", () => {
```
wchargin: s/greek/Greek/

Issue #586, opened 2018-08-02
Combine "Load Graph" and "Run Pagerank" into one button


decentralion:

decentralion:
A bit of nuance to consider: if I push the explore button, and then switch repo before the loadGraph finishes, the loadGraph promise will fail to apply (due to state change) and runPagerank should not get triggered. This  might require changing the return signature of loadGraph so that it returns a boolean, which is whether the loadGraph was successfully applied.

Also, if implemented naively the `loading` state will briefly toggle between `LOADING` to `NOT_LOADING` when `loadGraph` is successful, and then swap back to `LOADING` as soon as `runPagerank` triggers. We may want to test that it doesn't do this, as it could cause some spurious UI flickering.

PR #757, opened 2018-09-03
StateTransitionMachine.loadGraph reports success

decentralion:
Step one towards #586. This will enable us to chain runPagerank after
loadGraph only if the load went through successfully.

### DIFF
--- a/src/app/credExplorer/state.js
+++ b/src/app/credExplorer/state.js
@@ -79,7 +79,7 @@ export function initialState(): AppState {
 export interface StateTransitionMachineInterface {
   +setRepo: (Repo) => void;
   +setEdgeEvaluator: (EdgeEvaluator) => void;
-  +loadGraph: (Assets) => Promise<void>;
+  +loadGraph: (Assets) => Promise<boolean>;
   +runPagerank: (NodeAddressT) => Promise<void>;
 }
 /* In production, instantiate via createStateTransitionMachine; the constructor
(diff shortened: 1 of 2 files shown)


wchargin:
> Why use `Promise<boolean>` instead of rejecting on failure?

In particular, what is the difference in interpretation between a
promise that resolves to `false` and a promise that rejects? The client
will need to handle both anyway.

decentralion:
> In particular, what is the difference in interpretation between a
> promise that resolves to false and a promise that rejects? The client
> will need to handle both anyway.

Consider the case where the state transition is computed correctly, but should not be applied because the transition was pre-empted (e.g. while waiting for PageRank to converge, the user selected a different repo). If the promise rejects, then this will send an unhandled promise rejection to the console, which is noise from the user's perspective - the UI still functions fine. So that is a case where we want to return false, but not reject.

Alternatively, we could have the async methods always reject when they fail to apply their transition, but then we would need to add another layer of error-handling code to differentiate the "transition pre-empted" rejections (which should be suppressed) from other rejections. However, this means that the "transition pre-empted" concept would need to leak from state.js and into all callers of StateTransitionMachine's methods, i.e. it increases the conceptual surface area without giving any more power to callers.

There is a case to be made for switching the promises so that when there is a real failure, they reject rather than resolving. This would remove the need for the custom code that calls `console.error` on the failure before suppressing it. If you like, I can make that change in a separate commit. However, I do not see any concrete benefit from making this change.

If there were many async state-transitioning methods in STM, then we might benefit from switching to a DRY pattern around handling rejects/loading|failure states/transition pre-empting, and refactoring around promise rejections might make sense. But as there are only two such methods, there is little benefit to be gained from additional abstraction.

PR #760, opened 2018-09-03
Move WeightConfig button to the right

decentralion:
By moving the WeightConfig button to the side and away from the main
controls, I think we visually signal that this button is less important
/ isn't part of the main flow.

Test plan: Change is purely visual.

image

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -63,6 +63,7 @@ export class WeightConfig extends React.Component<Props, State> {
     return (
       <React.Fragment>
         <button
+          style={{float: "right"}}
           onClick={() => {
             this.setState(({expanded}) => ({expanded: !expanded}));
           }}


decentralion:
Yeah, the way that WeightConfig places itself right now is wrong, and is gonna change, as I continue refactoring + testing the weight configuration code. I think I'm going to move the entire concept down into PagerankTable and change how the button works.

I view this as a temporary / hacky improvement on the code, which will be refactored as I continue fixing up this part of the codebase. I prefer to put it in because in light of #759 I think it's a big UI improvement. In either case, within the next few days I'll have better code that accomplishes the same goal.

decentralion:
After considering your concerns, I think it does indeed make more sense to fix this properly later. There's little risk that I'll forget to make the fix if we don't add the hack, but there's some risk that I'll forget to remove the hack if I do add it. And we have no immediate deadlines pushing for a fast fix here.

PR #762, opened 2018-09-03
Add default weights for edge types

decentralion:
This commit modifies the edge type so that it has a
`defaultForwardWeight` and `defaultBackwardWeight`, and these defaults
are respected by the `WeightConfig`.

I came up with reasonable-seeming defaults for the Git and GitHub
plugins; these will undoubtably be more methodically tuned in the
future.

Test plan: `yarn test` passes, also opening the cred explorer now has
the specified default weights in the WeightConfig. (Note that the
forward/backward directions are reversed as described in #749.)

### DIFF
--- a/src/app/adapters/demoAdapters.js
+++ b/src/app/adapters/demoAdapters.js
@@ -33,13 +33,17 @@ export const machineNodeType: NodeType = Object.freeze({
 
 export const assemblesEdgeType: EdgeType = Object.freeze({
   forwardName: "assembles",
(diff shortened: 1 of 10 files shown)


wchargin:
Needs rebase.

nit: Of `2 ** -2`, `0.25`, and `1/4`, the latter two are easier for me to read than the former. The new UI represents everything in linear space, anyway; is there any reason that some parts of this patch use the exponent form?

PR #763, opened 2018-09-03
 Factor out NodeTypeConfig

decentralion:
This factors `NodeTypeConfig` out of the `WeightConfig` component. The
scope for a `NodeTypeConfig` is that it configures a single node type.
Right now it just renders a single `WeightSlider`, but I like factoring
out both for consistency with the `EdgeTypeConfig` (see #749) and
because I expect we may want to add more complexity later.

Test plan: The new component has some tests, also I manually tested the
frontend.

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -6,9 +6,14 @@ import sortBy from "lodash.sortby";
 import {type EdgeEvaluator} from "../../core/attribution/pagerank";
 import {byEdgeType, byNodeType} from "./edgeWeights";
 import {defaultStaticAdapters} from "../adapters/defaultPlugins";
-import type {NodeType, EdgeType} from "../adapters/pluginAdapter";
+import type {EdgeType} from "../adapters/pluginAdapter";
 import {WeightSlider} from "./weights/WeightSlider";
 import {DirectionalitySlider} from "./weights/DirectionalitySlider";
+import {
+  NodeTypeConfig,
+  defaultWeightedNodeType,
+  type WeightedNodeType,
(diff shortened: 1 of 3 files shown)


src/app/credExplorer/WeightConfig.js
```diff
@@ -175,36 +172,30 @@ class EdgeConfig extends React.Component<{
...
-    const sortedWeights = sortBy(
-      this.props.nodeWeights,
-      ({type}) => type.prefix
-    );
-
-    const controls = sortedWeights.map(({type, weight}) => {
-      const onChange = (value) => {
-        const nodeWeights = this.props.nodeWeights.filter(
-          (x) => x.type.prefix !== type.prefix
+  controls() {
```
wchargin: nit: It is customary to call this `_renderControls` (underscore optional).

src/app/credExplorer/WeightConfig.js
```diff
@@ -175,36 +172,30 @@ class EdgeConfig extends React.Component<{
 }
 
 class NodeConfig extends React.Component<{
-  nodeWeights: NodeWeights,
-  onChange: (NodeWeights) => void,
+  nodeWeights: WeightedNodeType[],
```
wchargin: That this array should be mutable is quite surprising. Is it intentional? (Likewise in state, though less so.)

src/app/credExplorer/WeightConfig.js

src/app/credExplorer/WeightConfig.js
decentralion: Fixed.

wchargin:
Reapproved.

PR #764, opened 2018-09-04
Fix a bug that can lock the cred explorer UI

decentralion:
Fix a bug that can lock the cred explorer UI

Currently, it's possible to lock the cred explorer UI via the following
sequence of actions:

1. Press the analyze cred button
2. While it is running, modify the weights

After following this repro, the UI is stuck: it will say "loading"
forever, and the analyze cred button is disabled.

The issue is that loadGraph and runPagerank do not apply their success
(or failure) state transitions if they are pre-empted by another state
change. If a repo change occurs, that's the right behavior: the repo
change puts the state back to `"READY_TO_LOAD_GRAPH"`, which means the
UI is ready to re-load, and showing the PageRank results for the wrong
repo would be very confusing.

However, if an edge evaluator change occurs while loadGraph or
runPagerank is happening, the state is left in the "LOADING" status
(which means the analyze cred button  is disabled).

This commit fixes the issue via a refactor: per @wchargin's suggestion,
responsibility for the edge evaluator moves from the state module out to
`credExplorer/App.js`. This dramatically simplifies the state module, as
we no longer need a `Substate` concept: we can simplify the state into a
single sequence of states.

As of the refactor, the bug is impossible.

Test plan: Unit tests have been updated to maintain coverage on the
refactored code. I manually tested that the bug no longer repro's, and
that the cred explorer UI continues to function. I did not add a new
test to protect against regression, because in the new codepath, the bug
is basically impossible.

### DIFF
--- a/src/app/credExplorer/App.js
+++ b/src/app/credExplorer/App.js
@@ -2,11 +2,13 @@
 
 import React from "react";
 
+import * as NullUtil from "../../util/null";
 import type {Assets} from "../assets";
 import type {LocalStore} from "../localStore";
 import CheckedLocalStore from "../checkedLocalStore";
 import BrowserLocalStore from "../browserLocalStore";
 
+import {type EdgeEvaluator} from "../../core/attribution/pagerank";
 import {PagerankTable} from "./pagerankTable/Table";
 import {WeightConfig} from "./WeightConfig";
 import RepositorySelect from "./RepositorySelect";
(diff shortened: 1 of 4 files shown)


src/app/credExplorer/App.js
```diff
@@ -61,14 +65,11 @@ export function createApp(
       const {localStore} = this.props;
       const {appState} = this.state;
       const loadingState =
-        appState.type === "INITIALIZED" ? appState.substate.loading : null;
+        appState.type === "UNINITIALIZED" ? null : appState.loading;
```
wchargin: The nullity of this variable makes it extra confusing to clients: a
client who writes `loadingState === "LOADING"` will be unintentionally
forgetting the `null` case, which is not locally apparent.

In other words, clients should expect that `loadingState: LoadingState`,
but this constraint does not in fact hold.

As this variable is only used once, and the caller has a redundant check
against the app state anyway, can we inline it?

PR #769, opened 2018-09-05
Fail on console errors when testing enzyme

### LINKED ISSUE #668: Factor out console error/warn assertions in tests
An increasing number of tests have code like:

```js
  beforeEach(() => {
    // $ExpectFlowError
    console.error = jest.fn();
    // $ExpectFlowError
    console.warn = jest.fn();
  });
  afterEach(() => {
    expect(console.warn).not.toHaveBeenCalled();
    expect(console.error).not.toHaveBeenCalled();
  });
```

These assertions are good. In fact, they should probably be in every
test! We should factor them out.

Doing so will require maintaining support for the cases where we expect
warnings to be raised. We currently do so with

```js
    expect(console.warn).toHaveBeenCalledTimes(0);
    someQuestionableAction();
    expect(console.warn).toHaveBeenCalledTimes(1);
    expect(console.warn).toHaveBeenCalledWith("hold up");
    // $ExpectFlowError
    console.warn = jest.fn();
```

which should be provided as a helper function.

decentralion:
`testUtil.configureEnzyme` now additionally asserts, after every test,
that `console.error` and `console.warn` were not called. Tests that
explicitly expect such calls can still be written by manually re-mocking
the relevant console method (and several examples already exist).

The code that explicitly specifies this for various enzyme test files
has been removed.

Test plan: `git grep "not.toHaveBeenCalled"` shows only unrelated usage.
`yarn test` passes. Adding a spurious console.warn to a passing test
causes it to fail.

Fixes #668

### DIFF
--- a/src/app/testUtil.js
+++ b/src/app/testUtil.js
@@ -4,6 +4,16 @@ export function configureEnzyme() {
   const Enzyme = require("enzyme");
   const Adapter = require("enzyme-adapter-react-16");
   Enzyme.configure({adapter: new Adapter()});
+  beforeEach(() => {
+    // $ExpectFlowError
+    console.error = jest.fn();
+    // $ExpectFlowError
+    console.warn = jest.fn();
+  });
+  afterEach(() => {
+    expect(console.warn).not.toHaveBeenCalled();
+    expect(console.error).not.toHaveBeenCalled();
+  });
 }
 
 export function configureAphrodite() {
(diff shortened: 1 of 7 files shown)


wchargin:
Lovely.

Why Enzyme only?

src/app/testUtil.js
```diff
@@ -4,6 +4,16 @@ export function configureEnzyme() {
   const Enzyme = require("enzyme");
   const Adapter = require("enzyme-adapter-react-16");
   Enzyme.configure({adapter: new Adapter()});
+  beforeEach(() => {
```
wchargin: Superfluous?

decentralion:
It's enzyme only because @wchargin and I spent some time trying to get it to work everywhere, but did not succeed.

Issue #595, opened 2018-08-02
Separate Weight Sliders by Plugin, better namespacing


claireandcode:
As the project evolves, it is likely that the number of plugins will increase.
Similarly, a user may wish to weight different plugins relative to each other, e.g. all "git" nodes are weighted less heavily than all "github" nodes. 
To start building the infrastructure to facilitate those changes, we want to separate the weight sliders by plugin, and also take advantage of the namespacing in the plugin adapter to provide better labelling, and thus a friendlier UI.

Credit to @decentralion and @wchargin for architectural planning discussions.

decentralion:
FYI, I'm making some changes related to #465 which should simplify how the plugin-level weights will work. Please hold off on implementing this until my changes land

decentralion:
Nevermind, I didn't make any relevant changes.

decentralion:
Closed by #773.

PR #773, opened 2018-09-05
Organize weights by plugin

decentralion:
This commit adds PluginWeightConfig, which is responsible for
adding all the weights for an individual plugin. The top-level
WeightConfig now creates multiple PluginWeightConfigs. It also takes
responsibility for hiding the FallbackPlugin.

Test plan: The PluginWeightConfig is tested (and fairly simple). The
top-level WeightConfig is not yet tested (#604), so I manually tested
that the weights in the app still function.

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -1,30 +1,31 @@
 // @flow
 
 import React from "react";
-import sortBy from "lodash.sortby";
 
 import {type EdgeEvaluator} from "../../core/attribution/pagerank";
 import {byEdgeType, byNodeType} from "./edgeWeights";
-import {defaultStaticAdapters} from "../adapters/defaultPlugins";
+import type {StaticAdapterSet} from "../adapters/adapterSet";
 import {
-  NodeTypeConfig,
-  defaultWeightedNodeType,
+  type WeightedTypes,
+  PluginWeightConfig,
+} from "./weights/PluginWeightConfig";
+import {
   type WeightedNodeType,
+  defaultWeightedNodeType,
 } from "./weights/NodeTypeConfig";
 import {
-  EdgeTypeConfig,
-  defaultWeightedEdgeType,
   type WeightedEdgeType,
+  defaultWeightedEdgeType,
 } from "./weights/EdgeTypeConfig";
-import {styledVariable} from "./weights/EdgeTypeConfig";
+import {FALLBACK_NAME} from "../adapters/fallbackAdapter";
 
 type Props = {|
+  +adapters: StaticAdapterSet,
   +onChange: (EdgeEvaluator) => void,
 |};
 
 type State = {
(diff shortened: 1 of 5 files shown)


src/app/credExplorer/WeightConfig.js
```diff
@@ -65,26 +62,49 @@ export class WeightConfig extends React.Component<Props, State> {
...
           </div>
         )}
       </React.Fragment>
     );
   }
 
+  pluginWeightConfigs() {
+    return this.props.adapters
+      .adapters()
+      .filter((x) => x.name() !== FALLBACK_NAME)
```
wchargin: Can we filter by the prefix instead of the name? The former is
_required_ to be unique, while the latter merely _should_ be.

src/app/credExplorer/WeightConfig.js
```diff
@@ -65,26 +61,49 @@ export class WeightConfig extends React.Component<Props, State> {
...
+      const weights = this.state.pluginNameToWeights.get(adapter.name());
+      const newNodeWeights =
+        weights == null
+          ? adapter.nodeTypes().map(defaultWeightedNodeType)
+          : weights.nodes;
+      const newEdgeWeights =
+        weights == null
+          ? adapter.edgeTypes().map(defaultWeightedEdgeType)
+          : weights.edges;
+      nodeWeights = nodeWeights.concat(newNodeWeights);
```
wchargin: Note that this may be quadratic in the number of plugins (or, depending
on the array implementation, may be linear). This is probably fine.

src/app/credExplorer/weights/PluginWeightConfig.js
```diff
@@ -0,0 +1,99 @@
...
+          onChange={onChange}
+        />
+      );
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <h3>{`${this.props.adapter.name()}`}</h3>
```
wchargin: `` `${"Superfluous interpolation."}` ``

src/app/credExplorer/WeightConfig.js
decentralion: Name is required unique.
https://github.com/sourcecred/sourcecred/blob/master/src/app/adapters/adapterSet.js#L34

src/app/credExplorer/weights/PluginWeightConfig.js
decentralion: Fixed

decentralion:
> If I understand correctly: the behavior that the fallback adapter is
elided is not tested because the implementation is in WeightConfig,
which still has ~no testing; as you continue to clean up this part of
the code base, this feature will eventually be tested. Is that right?

Correct.

src/app/credExplorer/WeightConfig.js
wchargin: Great; that works, too. Thanks!

wchargin:
Does this close #595?

decentralion:
Yep, updated.

PR #787, opened 2018-09-06
Create a `weights` module with types and utils

decentralion:
This commit creates a central `weights` module that defines all of the
weight-related types, and provides some utilities for dealing with them.

This way users of weight-concepts do not need to depend on a lot of
random modules just to get the relevant types. The utility methods are
implicitly defined a few places in the codebase: now we can avoid
re-writing them, and test them more thoroughly.

Test plan: Unit tests pass.

### DIFF
--- a/src/app/credExplorer/weights/weights.js
+++ b/src/app/credExplorer/weights/weights.js
@@ -1,0 +1,74 @@
+// @flow
+
+import {NodeAddress, EdgeAddress} from "../../../core/graph";
+import type {NodeType, EdgeType} from "../../adapters/pluginAdapter";
+import type {StaticPluginAdapter} from "../../adapters/pluginAdapter";
+import type {StaticAdapterSet} from "../../adapters/adapterSet";
+
+export type WeightedNodeType = {|+type: NodeType, +weight: number|};
+
+export type WeightedEdgeType = {|
+  +type: EdgeType,
+  +forwardWeight: number,
+  +backwardWeight: number,
+|};
+
+export type WeightedTypes = {|
+  +nodes: $ReadOnlyArray<WeightedNodeType>,
+  +edges: $ReadOnlyArray<WeightedEdgeType>,
+|};
+
+export function defaultWeightedNodeType(type: NodeType): WeightedNodeType {
+  return {
+    type,
+    weight: type.defaultWeight,
+  };
+}
+
+export function defaultWeightedEdgeType(type: EdgeType): WeightedEdgeType {
+  return {
+    type,
+    forwardWeight: type.defaultForwardWeight,
+    backwardWeight: type.defaultBackwardWeight,
+  };
+}
+
+export function defaultWeightsForAdapter(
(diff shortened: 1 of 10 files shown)


src/app/credExplorer/weights/weights.js
```diff
@@ -0,0 +1,74 @@
...
+  return {
+    nodes: adapter.nodeTypes().map(defaultWeightedNodeType),
+    edges: adapter.edgeTypes().map(defaultWeightedEdgeType),
+  };
+}
+
+export function combineWeights(
+  ws: $ReadOnlyArray<WeightedTypes>
+): WeightedTypes {
+  const seenPrefixes = new Set();
```
wchargin: Note that if you didn’t care about saying specifically which prefix was duplicated, then this could just be

```js
if (new Set(nodes.map((x) => x.type.prefix)).size !== nodes.length) {
  throw new Error("Duplicate prefix among: " + JSON.stringify(nodes));
}
```

But this is fine with me, too.

src/app/credExplorer/weights/weights.js
```diff
@@ -0,0 +1,74 @@
...
+  const nodes = [].concat(...ws.map((x) => x.nodes));
+  for (const {
+    type: {prefix},
+  } of nodes) {
+    if (seenPrefixes.has(prefix)) {
```
wchargin: We check for duplicate prefix, but not for duplicate name. Isn’t this
inconsistent with the adapter sets?

src/app/credExplorer/weights/weights.test.js
```diff
@@ -0,0 +1,93 @@
...
+        nodes: [defaultWeightedNodeType(inserterNodeType)],
+        edges: [defaultWeightedEdgeType(assemblesEdgeType)],
+      };
+      const weights2 = {
+        nodes: [defaultWeightedNodeType(machineNodeType)],
+        edges: [defaultWeightedEdgeType(transportsEdgeType)],
+      };
+      expect(combineWeights([weights1, weights2])).toEqual(defaultWeights());
+    });
+    it("treats empty weights as an identity", () => {
```
wchargin: Good. Thanks.

src/app/credExplorer/weights/weights.js
decentralion: No. Adapter set requires that plugin *names* be distinct, but not that their types' names be unique.

I.e. the Git Plugin and Phabricator plugin can both have a node type called 'issue'. Attempting to require that every node type name be unique across all plugins would not be feasible. Likewise for edges.

src/app/credExplorer/weights/weights.js
decentralion: Ack.

src/app/credExplorer/weights/weights.js
wchargin: Right. Makes sense. Thanks.

PR #790, opened 2018-09-06
flow: add typing to Jest transform modules

wchargin:
Test Plan:
`yarn flow` passes, and the [Jest docs][1] suggest that the added type
annotations are correct.

[1]: https://jestjs.io/docs/en/configuration#transform-object-string-string

wchargin-branch: flow-jest-transforms

### DIFF
--- a/config/jest/cssTransform.js
+++ b/config/jest/cssTransform.js
@@ -1,4 +1,4 @@
-// @no-flow
+// @flow
 // This is a custom Jest transformer turning style imports into empty objects.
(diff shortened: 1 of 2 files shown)


wchargin:
As of this commit, we have only one module without Flow types:

```js
$ git grep --name-only '@no-flow' -- '*.js'
config/polyfills.js
```

This remaining module has one Flow error, which I would like to resolve
via flow-typed/flow-typed#2705 instead of an `any`-cast or similar.

PR #792, opened 2018-09-06
`credExplorer/App` stores weights, not evaluator

decentralion:
This commit refactors `credExplorer/App` so that instead of storing an
`EdgeEvaulator` in its state, it stores `WeightedTypes` instead. This
has a few benefits:

- It's trivial to generate the right default value for `WeightedTypes`,
so we no longer allow the variable to be nullable in the state. This
simplifies logic, removes an error case, and means that we don't require
the `WeightConfig` to mount before the app is usable.
- `WeightedTypes` are serializable and can be tested for equality, so
they are a better-behaved piece of state
- We put off the information-destroying transformation as long as
possible
- In the future, I think we may want to move the weights/types concept
into core, at which point the `WeightedTypes` will directly be consumed
by the `core/attribution` module.

Test plan: Unit tests are pretty thorough; to be safe, I tested the UI
myself.

### DIFF
--- a/src/app/credExplorer/App.js
+++ b/src/app/credExplorer/App.js
@@ -2,15 +2,18 @@
 
 import React from "react";
 
-import * as NullUtil from "../../util/null";
 import type {Assets} from "../assets";
 import type {LocalStore} from "../localStore";
 import CheckedLocalStore from "../checkedLocalStore";
 import BrowserLocalStore from "../browserLocalStore";
 
-import {type EdgeEvaluator} from "../../core/attribution/pagerank";
+import {defaultStaticAdapters} from "../adapters/defaultPlugins";
 import {PagerankTable} from "./pagerankTable/Table";
 import {WeightConfig} from "./WeightConfig";
+import {
+  type WeightedTypes,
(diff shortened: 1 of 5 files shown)


wchargin:
I definitely agree with the goal and rationale.

It looks like the types are now stored in the state at two levels: the
top-level app and the weight config.

How about making `WeightConfig` a controlled component? It could accept
the current weight types as a prop, and an `onChange` handler, but not
store any state itself. It would decompose its prop into the
plugin-specific weights, and render some subcomponents—each of which is
already a controlled component. When the subcomponent changes, it can
merge in the change and invoke its own `onChange`.

This reduces the amount of state, instead relying on React to propagate
changes to props. In other words, it removes a whole class of
desynchronization bugs.

What do you think?

decentralion:
This is a great suggestion. It's a suggestion that will be easier to implement after this PR is merged - because then we have the WeightedTypes at the top level state, and can canonicalize it and pass it down to the children. If you approve this I'll send a PR implementing your suggestion shortly.

wchargin:
Okay.

PR #794, opened 2018-09-06
Add `MapUtil.merge` for combining maps

decentralion:
See the docstring for details.

Test plan: Unit tests.

### DIFF
--- a/src/util/map.js
+++ b/src/util/map.js
@@ -122,3 +122,27 @@ export function mapEntries<K, V, InK, InV>(
   }
   return result;
 }
+
+/**
+ * Merge maps without mutating the arguments.
+ *
+ * Merges multiple maps, returning a new map which has every key from
+ * the source maps, with their corresponding values. None of the inputs
+ * are mutated. In the event that multiple maps have the same key, an
+ * error will be thrown.
+ */
+export function merge<K, V>(
+  maps: $ReadOnlyArray<Map<$Subtype<K>, $Subtype<V>>>
+): Map<K, V> {
+  const result = new Map();
+  let updates = 0;
+  for (const map of maps) {
+    for (const [key, value] of map.entries()) {
+      result.set(key, value);
+      if (result.size !== ++updates) {
+        throw new Error(`Maps have duplicate key: ${String(key)}`);
+      }
+    }
+  }
+  return result;
+}
--- a/src/util/map.test.js
+++ b/src/util/map.test.js
@@ -245,4 +245,53 @@ describe("util/map", () => {
       expect(output).toEqual(new Map().set(11, "wat").set(12, "wat"));
     });
   });
+  describe("merge", () => {
+    it("combines two simple maps", () => {
+      const a = new Map().set("a", 1);
+      const b = new Map().set("b", 2);
+      const c = new Map().set("c", 3);
+      expect(MapUtil.merge([a, b, c])).toEqual(
+        new Map()
+          .set("a", 1)
+          .set("b", 2)
+          .set("c", 3)
+      );
+    });
+    it("treats empty map as an identity", () => {
+      const m = new Map().set("a", 11).set("b", 22);
+      expect(MapUtil.merge([new Map(), m, new Map()])).toEqual(m);
+    });
+    it("errors if there are any duplicate keys", () => {
+      const a = new Map().set("a", null);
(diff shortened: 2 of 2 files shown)


src/util/map.test.js
```diff
@@ -245,4 +245,32 @@ describe("util/map", () => {
...
+      const b = new Map().set("b", 2);
+      const c = new Map().set("c", 3);
+      expect(MapUtil.concat([a, b, c])).toEqual(
+        new Map()
+          .set("a", 1)
+          .set("b", 2)
+          .set("c", 3)
+      );
+    });
+    it("treats empty map as an identity", () => {
```
wchargin: Can you also explicitly test that `expect(concat([])).toEqual(new Map())`?

src/util/map.js
```diff
@@ -122,3 +122,26 @@ export function mapEntries<K, V, InK, InV>(
...
+export function concat<K, V>(maps: $ReadOnlyArray<Map<K, V>>): Map<K, V> {
+  const result = new Map();
+  let updates = 0;
+  for (const map of maps) {
+    for (const [key, value] of map.entries()) {
+      updates++;
+      result.set(key, value);
+    }
+  }
+  if (Array.from(result.keys()).length !== updates) {
```
wchargin: Why not `if (result.size !== updates)`?

src/util/map.js
```diff
@@ -122,3 +122,26 @@ export function mapEntries<K, V, InK, InV>(
...
+ * the source maps, with their corresponding values. None of the inputs
+ * are mutated. In the event that multiple maps have the same key, an
+ * error will be thrown.
+ */
+export function concat<K, V>(maps: $ReadOnlyArray<Map<K, V>>): Map<K, V> {
+  const result = new Map();
+  let updates = 0;
+  for (const map of maps) {
+    for (const [key, value] of map.entries()) {
+      updates++;
```
wchargin: Note that you could find exactly which key was the duplicate without a
`has`-check at each iteration, by making sure that `result.size`
increments by exactly one at each iteration instead of checking it at
the end. This requires *n* accesses to `size`, which is certainly
cheaper than *n* calls to `has`, and probably free in practice.

I have no strong preference, so I’ll leave it up to you.

src/util/map.js
```diff
@@ -122,3 +122,26 @@ export function mapEntries<K, V, InK, InV>(
   }
   return result;
 }
+
+/**
```
wchargin: Note that the analogous operation on `Graph` is called `merge`, and also
this function has a different API from `Array.prototype.concat` (and a
better one). Weak preference toward `merge` for those reasons, and also
because `concat` suggests a notion of order that maps don’t\* have.

src/util/map.js
```diff
@@ -122,3 +122,26 @@ export function mapEntries<K, V, InK, InV>(
...
+ * Concatenate maps without mutating the arguments.
+ *
+ * Combines multiple maps, returning a new map which has every key from
+ * the source maps, with their corresponding values. None of the inputs
+ * are mutated. In the event that multiple maps have the same key, an
+ * error will be thrown.
+ */
+export function concat<K, V>(maps: $ReadOnlyArray<Map<K, V>>): Map<K, V> {
```
wchargin: It would be nice for test cases like this to work:

```js
const numberMap: Map<number, number> = new Map().set(1, 2);
const stringMap: Map<string, string> = new Map().set("one", "two");
type NS = number | string;
const polyMap: Map<NS, NS> = concat([numberMap, stringMap]);
// (result could also be `Map<mixed, mixed>`)
```

Or, perhaps more usefully:

```js
const gitNodeDescriptions: Map<GitAddress, string> = new Map();
const githubNodeDescriptions: Map<GithubAddress, string> = new Map();
const nodeDescriptions: Map<NodeAddressT, string> = concat([
  gitNodeDescriptions,
  githubNodeDescriptions,
]);
```

I believe (but have not tested) that (a) this test case would not pass
currently, and (b) it would pass if you changed the type declaration to

```js
concat<K, V>(maps: $ReadOnlyArray<Map<$Subtype<K>, $Subtype<V>>>): Map<K, V>;
```

What do you think of this suggestion?

src/util/map.js
decentralion: Done, thanks.

src/util/map.js
decentralion: Done, thanks.

PR #795, opened 2018-09-06
Change WeightedTypes to contain maps

decentralion:
This will make it easier to re-organize the weight components so that
the WeightedTypes have a single source of truth, as described in
https://github.com/sourcecred/sourcecred/pull/792#issuecomment-419234721

Test plan: Unit tests suffice.

### DIFF
--- a/src/app/credExplorer/weights/PluginWeightConfig.js
+++ b/src/app/credExplorer/weights/PluginWeightConfig.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from "react";
-import deepEqual from "lodash.isequal";
+import * as MapUtil from "../../../util/map";
 import {NodeTypeConfig} from "./NodeTypeConfig";
 import {EdgeTypeConfig} from "./EdgeTypeConfig";
 import {StaticPluginAdapter} from "../../adapters/pluginAdapter";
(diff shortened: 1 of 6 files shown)


src/app/credExplorer/weights/weights.js
```diff
@@ -14,8 +15,8 @@ export type WeightedEdgeType = {|
 |};
 
 export type WeightedTypes = {|
-  +nodes: $ReadOnlyArray<WeightedNodeType>,
-  +edges: $ReadOnlyArray<WeightedEdgeType>,
+  +nodes: Map<NodeAddressT, WeightedNodeType>,
```
wchargin: Should document what these addresses mean: it is not obvious that the
key is required to map to the prefix of the type of its corresponding
value.

src/app/credExplorer/weights/PluginWeightConfig.js
```diff
@@ -37,45 +37,43 @@ export class PluginWeightConfig extends React.Component<Props, State> {
...
-    return this.state.weights.nodes.map((wnt: WeightedNodeType) => {
-      const onChange = (newType: WeightedNodeType) => {
-        const index = this.state.weights.nodes.findIndex((x) =>
-          deepEqual(x.type, wnt.type)
+    return Array.from(this.state.weights.nodes.values()).map(
+      (wnt: WeightedNodeType) => {
+        const onChange = (newType: WeightedNodeType) => {
+          const newNodes = MapUtil.copy(this.state.weights.nodes);
+          newNodes.set(newType.type.prefix, newType);
+          const weights = {nodes: newNodes, edges: this.state.weights.edges};
```
wchargin: yay, one fewer `slice`.

src/app/credExplorer/weights/PluginWeightConfig.js
```diff
@@ -37,45 +37,43 @@ export class PluginWeightConfig extends React.Component<Props, State> {
...
 
   _renderNodeWeightControls() {
-    return this.state.weights.nodes.map((wnt: WeightedNodeType) => {
-      const onChange = (newType: WeightedNodeType) => {
-        const index = this.state.weights.nodes.findIndex((x) =>
-          deepEqual(x.type, wnt.type)
+    return Array.from(this.state.weights.nodes.values()).map(
+      (wnt: WeightedNodeType) => {
+        const onChange = (newType: WeightedNodeType) => {
+          const newNodes = MapUtil.copy(this.state.weights.nodes);
```
wchargin: This computation needs to be in a `setState` callback, because you read
from `this.state` and use the result to construct the new state.

Something like:

```js
this.setState((state) => {
  const newNodes = MapUtil.copy(state.weights.nodes);
  newNodes.set(newType.type.prefix, newType);
  const weights = {...state.weights, newNodes};
  return {weights};
}, () => this.fire());
```

with various levels of inlining possible depending on your preference.

(Recall that React batches state updates, so the “pending state” may be
more recent than `this.state`. The code as is risks stomping updates.)

Likewise below for edges.

src/app/credExplorer/weights/PluginWeightConfig.js
wchargin: nit: prefer `{...this.state.weights, nodes: newNodes}`, unless a reason
otherwise? It’s a clearer signal to the reader that you intend to update
just one field.

src/app/credExplorer/weights/PluginWeightConfig.js
decentralion: Good catch. Done.

src/app/credExplorer/weights/PluginWeightConfig.js
decentralion: Done.

PR #796, opened 2018-09-06
Make `credExplorer/App` control the WeightedTypes

### LINKED ISSUE #604: WeightConfig component has no testing

decentralion:
This commit implements a [suggestion] to make `credExplorer/App` a
single source of truth on the `WeightedTypes`. As such, both
`WeightConfig` and `PluginWeightConfig` have been refactored to be
(essentially) stateless components that are controlled from above. I say
essentially because `WeightConfig` still has its expanded state, but
that will go away soon.

Along the way, I've improved testing and added some new invariant
checking (e.g. that `PluginWeightConfig` was passed the correct set of
weights for its adapter). For the first time, there are now tests for
the `WeightConfig` itself! I'm not totally done with the weight
re-write, but this seems like a good time to close #604, as the whole
logical sequence for setting weights is now tested.

Test plan: There are new unit tests. Also, to be sure, I manually tested
the UI as well.

[suggestion]: https://github.com/sourcecred/sourcecred/pull/792#issuecomment-419234721

### DIFF
--- a/src/app/credExplorer/WeightConfig.js
+++ b/src/app/credExplorer/WeightConfig.js
@@ -2,38 +2,29 @@
 
 import React from "react";
 import * as NullUtil from "../../util/null";
+import * as MapUtil from "../../util/map";
 
-import type {StaticPluginAdapter} from "../adapters/pluginAdapter";
 import type {StaticAdapterSet} from "../adapters/adapterSet";
-import {
-  type WeightedTypes,
-  defaultWeightsForAdapter,
-  combineWeights,
-} from "./weights/weights";
+import type {WeightedTypes} from "./weights/weights";
 import {PluginWeightConfig} from "./weights/PluginWeightConfig";
 import {FALLBACK_NAME} from "../adapters/fallbackAdapter";
 
 type Props = {|
   +adapters: StaticAdapterSet,
+  +weightedTypes: WeightedTypes,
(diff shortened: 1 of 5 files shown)


src/app/credExplorer/WeightConfig.js
```diff
@@ -55,43 +46,55 @@ export class WeightConfig extends React.Component<Props, State> {
...
-  pluginWeightConfigs() {
+  _renderPluginWeightConfigs() {
     return this.props.adapters
       .adapters()
       .filter((x) => x.name() !== FALLBACK_NAME)
       .map((adapter) => {
         const onChange = (weightedTypes) => {
-          this.state.pluginNameToWeights.set(adapter.name(), weightedTypes);
-          this.fire();
+          const newWeightedTypes = {
```
wchargin: To clarify: this is like

```js
          const newWeightedTypes = {
            nodes: MapUtil.merge([
              MapUtil.copy(this.props.weightedTypes.nodes),
              weightedTypes.nodes,
            ]),
            edges: MapUtil.merge([
              MapUtil.copy(this.props.weightedTypes.edges),
              weightedTypes.edges,
            ]),
          };
```

except that you need to _replace_ duplicates instead of failing if they
exist, yes?

decentralion: That's right.

PR #798, opened 2018-09-07
Move WeightConfig into PagerankTable

decentralion:
Previously, the WeightConfig (and the button that expanded it) were in
the credExplorer App. This was a little weird, as there's no reason to
play with the weights before you have some Pagerank results to
investigate; additionally, it risked confusing new users with a concept
that was not yet applicable.

Also, the implementation was wonky: the WeightConfig had responsibility
for expanding/hiding itself, which gave poor ability to position the
button and the WeightConfig separately.

Finally, the codepath was untested (vestiges of #604).

This commit fixes all three issues:
- The WeightConfig and button have moved into PagerankTable
- The WeightConfig is now a stateless component, and the parent takes
responsibility for deciding when to mount it
- Logic for showing/hiding the WeightConfig is now tested.

### DIFF
--- a/src/app/credExplorer/pagerankTable/Table.test.js
+++ b/src/app/credExplorer/pagerankTable/Table.test.js
@@ -7,42 +7,87 @@ import {NodeAddress, type NodeAddressT} from "../../../core/graph";
 
 import {PagerankTable} from "./Table";
 import {example, COLUMNS} from "./sharedTestUtils";
+import {NodeRowList} from "./Node";
+import {WeightConfig} from "../weights/WeightConfig";
+import {defaultWeightsForAdapter} from "../weights/weights";
+import {FactorioStaticAdapter} from "../../adapters/demoAdapters";
 
 require("../../testUtil").configureEnzyme();
 describe("app/credExplorer/pagerankTable/Table", () => {
   describe("PagerankTable", () => {
-    it("renders thead column order properly", async () => {
-      const {pnd, adapters} = await example();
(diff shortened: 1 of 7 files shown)


src/app/credExplorer/pagerankTable/Table.test.js
```diff
@@ -7,50 +7,95 @@ import {NodeAddress, type NodeAddressT} from "../../../core/graph";
...
-      async function setup(defaultNodeFilter?: NodeAddressT) {
-        const {pnd, adapters} = await example();
-        const element = shallow(
-          <PagerankTable
-            defaultNodeFilter={defaultNodeFilter}
-            pnd={pnd}
-            adapters={adapters}
-            maxEntriesPerList={1}
-          />
+    describe("has a WeightConfig", () => {
```
wchargin: Instead of testing against the particular state implementation, why not
test:

  - by default:
      - there is no weight config
      - there is a button that says “Show weight config”
  - when you click the button:
      - there is a weight config
      - the button says “Hide weight config”

Then, the tests are robust to the particular shape of the state field.

decentralion: Ok. Then we also mix in the onWeightedTypeChange test into the "when you click the button" case, which is likely fine.

wchargin: Yep, fine with me, as that is how the user flow works.

PR #805, opened 2018-09-07
Add `MapUtil.pushValue` for maps of arrays

decentralion:
With some frequency we find ourselves needing to maintain maps whose
values are arrays that we append to. `MapUtil.pushValue` is a utility
method for these cases.

Existing usage in `aggregate.js` has been modified to use the new
function.

### DIFF
--- a/src/util/map.js
+++ b/src/util/map.js
@@ -146,3 +146,19 @@ export function merge<K, V>(
   }
   return result;
 }
+
+/**
+ * Given a map whose values are arrays, push an element onto the array
+ * corresponding to the given key. If the key is not in the map, first
+ * insert it with value a new empty array.
+ *
+ * If the key is already in the map, its value will be mutated, not
+ * replaced.
+ */
+export function pushValue<K, V>(map: Map<K, V[]>, key: K, value: V): void {
+  let arr = map.get(key);
+  if (arr == null) {
+    map.set(key, (arr = []));
+  }
+  arr.push(value);
+}
--- a/src/util/map.test.js
+++ b/src/util/map.test.js
@@ -294,4 +294,27 @@ describe("util/map", () => {
       ]);
     });
   });
+  describe("pushValue", () => {
+    it("works when the map has no matching key", () => {
+      const map = new Map();
+      MapUtil.pushValue(map, "foo", 3);
+      expect(map).toEqual(new Map().set("foo", [3]));
+    });
+    it("works when the map has a matching key", () => {
+      const map = new Map().set("foo", [3]);
+      MapUtil.pushValue(map, "foo", 4);
+      expect(map).toEqual(new Map().set("foo", [3, 4]));
+    });
+    it("works when the map already has an empty array", () => {
+      const map = new Map().set("foo", []);
+      MapUtil.pushValue(map, "foo", 1);
+      expect(map).toEqual(new Map().set("foo", [1]));
+    });
+    it("preserves array identity", () => {
+      const arr = [];
+      const map = new Map().set("foo", arr);
+      MapUtil.pushValue(map, "foo", 1);
+      expect(map.get("foo")).toBe(arr);
+    });
+  });
 });
--- a/src/app/credExplorer/pagerankTable/aggregate.js
+++ b/src/app/credExplorer/pagerankTable/aggregate.js
@@ -2,6 +2,7 @@
 
 import sortBy from "lodash.sortby";
 import stringify from "json-stable-stringify";
+import * as MapUtil from "../../../util/map";
 import {NodeTrie, EdgeTrie} from "../../../core/trie";
 import type {NodeType, EdgeType} from "../../adapters/pluginAdapter";
 import type {ScoredConnection} from "../../../core/attribution/pagerankNodeDecomposition";
@@ -53,11 +54,7 @@ export function aggregateByNodeType(
   const nodeTypeToConnections = new Map();
   for (const x of xs) {
     const type = typeTrie.getLast(x.source);
-    const connections = nodeTypeToConnections.get(type) || [];
-    if (connections.length === 0) {
-      nodeTypeToConnections.set(type, connections);
-    }
-    connections.push(x);
+    MapUtil.pushValue(nodeTypeToConnections, type, x);
   }
   const aggregations: NodeAggregation[] = [];
   for (const [
@@ -109,11 +106,7 @@ export function aggregateByConnectionType(
     }
     const edge = x.connection.adjacency.edge;
     const type = typeTrie.getLast(edge.address);
-    const connections = relevantMap.get(type) || [];
-    if (connections.length === 0) {
-      relevantMap.set(type, connections);
-    }
-    connections.push(x);
+    MapUtil.pushValue(relevantMap, type, x);
   }
 
   function createAggregation(


src/util/map.js
```diff
@@ -146,3 +148,58 @@ export function merge<K, V>(
...
+
+  values(): Iterator<$ReadOnlyArray<V>> {
+    return this._internalMap.values();
+  }
+
+  *entries(): Iterator<[K, $ReadOnlyArray<V>]> {
+    for (const [key, array] of this._internalMap.entries()) {
+      yield [key, array];
+    }
+    // TODO: get this more efficient implementation to typecheck
```
decentralion: @wchargin your suggestions are invited

wchargin:
Is a new class the right thing to do here?

For `NullUtil` and `MapUtil`, I deliberately avoided creating new
classes because they have nontrivial runtime elision, which makes them
more difficult to use in existing code. For instance, `NullUtil` could
ahve been designed around a new `class Optional<T>` with methods like
`map<U>(f: (T) => U): Optional<U>`. But then you’d have to marshal
between `Optional<T>` and `?T`, which makes the module much less
attractive than if it were to simply operate on maybe-types.

You could instead write functions on the built-in `Map` type:

```js
/**
 * Push `v` onto the array value for key `k`, initializing that value to
 * an empty array if `k` is not in the map.
 */
pushValue<K, V>(
  m: Map<K, V[]>,
  k: K,
  v: V
): void;

/**
 * Create a new map with the key-value pairs in `m` except for `k`,
 * which maps to `[...(m.get(k) || []), v]`.
 */
withPushedValue<K, V>(
  m: Map<K, $ReadOnlyArray<V>>,
  k: K,
  v: V
): Map<K, $ReadOnlyArray<V>>;
```

---

The behavior with mutation is very surprising. From reading the
documentation that “this method is not dangerous”, I was sure that this
meant that you were doing a shallow copy, so that observers would not be
able to tell that there was any mutation going on internally. After
reading the implementation, I was confused enough that I had to read the
tests to convince myself that this is not the case.

Exposing a mutable surface means that you can’t use this in React props
or state. It also brings all the usual problems of mutability. Strong
suggestion to not offer a class tied to a mutable API unless there’s
quite a good reason to do so. (A mutable API over an existing map type,
as with `pushValue` above, is not a problem.)

---

As far as the iterators go: I’d prefer to `yield* _internalMap.keys()`,
because this ensures that we have control over the implementation of the
returned iterator. In particular, we would not be returning a built-in
`MapIterator` object. In addition to this being an observable difference
(if only trivially so), it is more robust to, e.g., a change to the ES
spec to add a `delete()` method to the `MapIterator` protocol, [as is
the case in Java][it-remove]. If we directly returned the iterator, this
would open up a hole in our contract; by `yield*`ing from it, we’re
safe.

[it-remove]: https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#remove--

---

If my suggestion for point (1) is implemented, points (2) and (3) are
fixed for free.

decentralion:
Thanks for the suggestions, @wchargin. (1) is a much cleaner way to do things, and made the code much easier to integrate (including getting that usage I wanted from graph.js). PTAL.

src/util/map.js
```diff
@@ -146,3 +146,9 @@ export function merge<K, V>(
   }
   return result;
 }
+
+export function pushValue<K, V>(map: Map<K, V[]>, key: K, value: V): void {
```
wchargin: This definitely needs docs.

What does it do?

Does it guarantee that the identity of an existing array value is
preserved?

src/util/map.js
```diff
@@ -146,3 +146,9 @@ export function merge<K, V>(
...
+  const arr = map.get(key) || [];
+  arr.push(value);
+  map.set(key, arr);
```
wchargin: Since this is factored out into a utility function, we might as well use
the twice-as-fast-in-the-usual-case implementation:

```js
let arr = map.get(key);
if (arr == null) {
  map.set(key, arr = []);
}
arr.push(value);
```

Or am I missing something?

src/util/map.test.js
```diff
@@ -294,4 +294,24 @@ describe("util/map", () => {
...
+      MapUtil.pushValue(map, "foo", 3);
+      expect(map.get("foo")).toEqual([3]);
+    });
+    it("works when the map has a matching key", () => {
+      const map = new Map();
+      MapUtil.pushValue(map, "foo", 3);
+      MapUtil.pushValue(map, "foo", 4);
+      expect(map.get("foo")).toEqual([3, 4]);
+    });
+    it("doesn't add anything else to the mapp", () => {
```
wchargin: s/mapp/map/

src/util/map.test.js
```diff
@@ -294,4 +294,24 @@ describe("util/map", () => {
       ]);
     });
   });
+  describe("MapUtil.pushValue", () => {
```
wchargin: Test when the value is an empty array?

Test that array identity is preserved?

src/util/map.js
decentralion: Done

decentralion:
Updated, PTAL

src/util/map.js
```diff
@@ -146,3 +146,16 @@ export function merge<K, V>(
   }
   return result;
 }
+
+/**
+ * Push a value into the sub-array at key k, creating the array if needed.
```
wchargin: What do you mean by “sub-array”?

This doesn’t technically indicate what the initial value of the array
“if needed” is, but probably the reader can infer this.

src/util/map.test.js
```diff
@@ -294,4 +294,35 @@ describe("util/map", () => {
...
+      const arr = [];
+      const map = new Map().set("foo", arr);
+      MapUtil.pushValue(map, "foo", 1);
+      expect(map.get("foo")).toBe(arr);
+    });
+    it("doesn't add anything else to the map", () => {
+      const map = new Map();
+      MapUtil.pushValue(map, "foo", 3);
+      MapUtil.pushValue(map, "foo", 4);
+      expect(Array.from(map.entries())).toEqual([["foo", [3, 4]]]);
```
wchargin: Are these two `expect`s a vestige of a `class`ier implementation? The
first one seems okay, though it’d be better\* to write

```js
expect(map).toEqual(new Map().set("foo", [3, 4]));
```

But the second `expect` is quite confusing and should probably be
removed, unless I’m missing something.

\* “better” because this does not depend on the order of the key–value
pairs returned by `entries`—which is trivial in this case of a size-`1`
map, thus merely “better” and not “mandatory”. :-)

decentralion: Changed.

src/util/map.js
decentralion: Rewrote it

src/util/map.js
```diff
@@ -146,3 +146,20 @@ export function merge<K, V>(
   }
   return result;
 }
+
+/**
+ * Push a value into a map which has an array for every key.
```
wchargin: > Push a value into a map which has an array for every key.

This doesn’t make sense to me. “Has an array for every key” suggests
that the keys are arrays, and the use of “every” suggests that the map
is total over some (natural) domain, which certainly isn’t the case.

It sounds like you want to say that the _values_ of the map should all
be arrays, though this is already implied by the type. (I see how the
original text could be interpreted to mean this, but it wasn’t clear to
me, which suggests that it should be rewritten.)

What do you think of

```js
/**
 * Given a map whose values are arrays, push an element onto the array
 * corresponding to the given key. If the key is not in the map, first
 * insert it with value a new empty array.
 *
 * If the key is already in the map, its value will be mutated, not
 * replaced.
 */
```

?

src/util/map.test.js
```diff
@@ -294,4 +294,34 @@ describe("util/map", () => {
       ]);
     });
   });
+  describe("pushValue", () => {
+    it("works when the map has no matching key", () => {
+      const map = new Map();
+      MapUtil.pushValue(map, "foo", 3);
+      expect(map.get("foo")).toEqual([3]);
+    });
+    it("works when the map has a matching key", () => {
```
wchargin: This test is strictly weaker than “doesn’t add anything else to the
map”. Can we replace this test’s implementation with that test’s, and
delete the latter?

While we’re at it, we might as well replace the assertion on line 301
with `expect(map).toEqual(new Map().set("foo", [3]))`—this, too, looks
like a vestige of a former patch that used a new class instead of the
existing `Map` type.

decentralion: Done. Also modified the first test to follow the same pattern, and modified the second test to test its stated expectation more clearly.

Issue #804, opened 2018-09-07
Heuristic for comment relationships


decentralion:
An interesting [suggestion] from @whyrusleeping on valuing comments:

> I think that comments need a better way to be valued. Emoji responses will
> likely help here, but also I think that **increasing a comments weight if its
> author has someone respond to them later in the thread**.

[suggestion]: https://discuss.sourcecred.io/t/ipfs-cred-feedback/17/2

What if instead of increasing a comment's weight, we created new edges
according to this heuristic? If a comment mentions a user, and there were
comments by that user earlier in the thread, then the comment creates edges
pointing to those comments.

Interpreting it via edges rather than weight modification makes it harder to
use it for gaming. If we modify the weight, then a user could create a
sockpuppet account, and have the sockpuppet reference their comments, which
would increase the value of the comments out of proportion to the cred in the
sockpuppet account. Using edges ensures that the new cred flows still respect
the rest of the distribution.

PR #806, opened 2018-09-12
Add logic for finding `MentionsAuthorReference`s

decentralion:
A `MentionsAuthorReference` is created when a post mentions a user, and
that user has authored at least one post in the same thread. Then there
is a `MentionsAuthorReference` from the post to the other posts by that
author.

For context, see the docstrings in `mentionsAuthorReference.js`, and
see #804.

Test plan:
Thorough unit tests have been added, which test the entire pipeline,
from ingesting the data via GitHub's graphql responses, through to
detecting the references. Edge cases such as self-reference and
multi-reference are tested.

Thanks to @wchargin for help writing this commit.

### DIFF
--- a/src/plugins/github/heuristics/mentionsAuthorReference.js
+++ b/src/plugins/github/heuristics/mentionsAuthorReference.js
@@ -1,0 +1,110 @@
+// @flow
+
+import * as MapUtil from "../../../util/map";
+import * as RV from "../relationalView";
+import * as N from "../nodes";
+
+// MentionsAuthorReferences are inferred edges between posts.  If a post
+// contains an @-mention of a user, and that user has authored any posts
+// in the same thread, then we create MentionsAuthorReferences from the
+// mentioning post to any posts in the same thread that were authored by
+// the mentioned author.
+//
+// A concrete example: Suppose I write a post, and I @-mention you. Then
+// my post will have MentionsAuthorReferences pointing to any posts that
+// you write. The inference is that because I mentioned you by name, my
+// post probably values your posts (either because I am referring to
+// content you wrote in a previous post, or because I appreciate that
+// you showed up to engage with me after being mentioned)
+//
+// See #804 for context:
+// https://github.com/sourcecred/sourcecred/issues/804
+
+export type MentionsAuthorReference = {|
+  // The post that generated an mentions author reference (via @-mention)
+  +src: N.TextContentAddress,
+  // The post authored by the person mentioned in the src
+  +dst: N.TextContentAddress,
+  // The person that was mentioned in src.  Storing `who` is necessary
+  // because in the case of paired authorship, there might be multiple
+  // pointers from the same src to the same dst
+  +who: N.UserlikeAddress,
+|};
+
+export function* findMentionsAuthorReferences(
+  view: RV.RelationalView
+): Iterator<MentionsAuthorReference> {
+  function* issuesAndPulls() {
+    yield* view.issues();
+    yield* view.pulls();
+  }
+  for (const post of issuesAndPulls()) {
+    const thread = createThread(post);
+    yield* referencesFromThread(thread);
+  }
+}
+
+// A post in a thread. This is either a top-level post (like an issue or
+// pull request) or a child (like a comment).
+type ThreadPost = {|
+  // Address of the post.
+  +address: N.TextContentAddress,
+  // List of all authors of the post. Order is undefined (this is a set
+  // encoded as a list).
+  +authors: $ReadOnlyArray<N.UserlikeAddress>,
+|};
+
+type Thread = {|
+  // Includes the root post.
+  // Ordered chronologically.
+  +posts: $ReadOnlyArray<ThreadPost>,
+  // Map from raw addresses of userlike kind to an array of all indices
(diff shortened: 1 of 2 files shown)


src/plugins/github/heuristics/mentionedAuthorReference.js
```diff
@@ -0,0 +1,107 @@
...
+import * as RV from "../relationalView";
+import * as N from "../nodes";
+
+// MentionedAuthorReferences are inferred edges between posts.
+// If a post contains an @-mention of a user, and that user has authored any posts
+// in the same thread, then we create MentionedAuthorReferences from the mentioning
+// post to any posts authored by the mentioned author.
+// A concrete example: Suppose I write a post, and I @-mention you. Then my post will
+// have MentionedAuthorReferences pointing to any posts that you write. The inference
+// is that because I mentioned you by name, my post probably values your posts (either because
```
wchargin: nit: this line has length 94; please wrap (preferably at 72)

src/plugins/github/heuristics/mentionedAuthorReference.test.js
```diff
@@ -0,0 +1,302 @@
...
+    /** The test cases below are organized so that for every
+     * post in the thread rooted at Issue #2, we expect all
+     * of the new MentionedAuthorReferences that were created
+     * as a result of that post (whether because that post is
+     * the src or the dst). At the end, we verify that every
+     * MentionedAuthorReference is accounted for. */
+
+    // If an author references themself, there will be a self-directed
+    // MentionedAuthorReference.
+    expectReferences("issue:2", [
```
wchargin: As these strings translate to test case names, can we make them describe
the expected behavior at the domain level (e.g., “a self-reference on a
comment”) rather than the data-specific level (“comment:1_1”)?

src/plugins/github/heuristics/mentionedAuthorReference.test.js
```diff
@@ -0,0 +1,302 @@
...
+    ]);
+    expectReferences("comment:2_2", [
+      {
+        src: commentAddress(2, 2),
+        dst: commentAddress(2, 1),
+        who: users.holoPearl,
+      },
+      // Because Steven self-referenced, that post references all of Steven's
+      // other posts.  (If people use this for gaming, we can reconsider, but
+      // it shoudl be fairly obvious)
```
wchargin: shoudl

src/plugins/github/heuristics/mentionedAuthorReference.js
```diff
@@ -0,0 +1,107 @@
...
+  // there might be multiple pointers from the same src to the same dst
+  +who: N.UserlikeAddress,
+|};
+
+export function* findMentionedAuthorReferences(
+  view: RV.RelationalView
+): Iterator<MentionedAuthorReference> {
+  function* issuesAndPulls() {
+    yield* view.issues();
+    yield* view.pulls();
```
wchargin: If you delete `yield* view.pulls()`, the tests all pass.

Can we make the tests just a wee bit stronger?

src/plugins/github/heuristics/mentionedAuthorReference.test.js
decentralion: We could, but at the expense of readability and maintainability. When writing these tests, it's important to have a clear place where every individual reference should live. As written, there's a simple system: we place each reference with the {src} or {dst} of that reference, choosing whichever one was chronologically last. If I re-write them to have descriptive names, it won't be unambiguous which block should own any particular reference, making the code harder to maintain.

src/plugins/github/heuristics/mentionedAuthorReference.js
decentralion: Ok.

Just a wee bit, though.

src/plugins/github/heuristics/mentionsAuthorReference.js
```diff
@@ -0,0 +1,108 @@
+// @flow
+
+import * as MapUtil from "../../../util/map";
+import * as RV from "../relationalView";
+import * as N from "../nodes";
+
+// MentionsAuthorReferences are inferred edges between posts.  If a post
+// contains an @-mention of a user, and that user has authored any posts
+// in the same thread, then we create MentionsAuthorReferences from the
+// mentioning post to any posts authored by the mentioned author.  A
```
wchargin: Mind being a bit more explicit and writing “any posts authored by the
mentioned author within the same thread”? The fact that this is
thread-scoped is only hinted at in the introductory sentence.

src/plugins/github/heuristics/mentionsAuthorReference.js
```diff
@@ -0,0 +1,108 @@
...
+  +src: N.TextContentAddress,
+  // The post authored by the person mentioned in the src
+  +dst: N.TextContentAddress,
+  // The person that was mentioned in src.  Storing `who` is necessary
+  // because in the case of paired authorship, there might be multiple
+  // pointers from the same src to the same dst
+  +who: N.UserlikeAddress,
+|};
+
+export function* findMentionsAuthorReferences(
```
wchargin: I like that this and its return type are the only two exports. Good encapsulation.

PR #808, opened 2018-09-12
Add `MentionsAuthor` edges to the graph

### LINKED ISSUE #804: Heuristic for comment relationships

decentralion:
This commit builds on the work in #806, adding the
`MentionsAuthorReference`s to the graph. It thus resolves #804.

Empirically, the addition of these edges does not change the users' cred
distribution much.  Consider the results with the following 3 forward
weights for the edge (results for ipfs/go-ipfs):

| User          | w=1/32 | w=1/2  | w=2    |
|---------------|-------:|-------:|-------:|
| whyrusleeping | 228.04 | 225.69 | 223.86 |
| jbenet        | 102.04 | 100.26 | 99.53  |
| kubuxu        | 66.60  | 67.80  | 69.36  |
| ...           | —      | —      | —      |
| btc           | 22.69  | 22.29  | 21.38  |

The small effect on users' cred is not that surprising: the
MentionsAuthor references always "shadow" a direct comment->user
reference. In principle, the overall cred going to the user should be
similar; the difference is that now some more cred flows in between the
various comments authored by that user, on the way to the user. (And if
those other comments had references, then it flows out from them, etc.)

Empirically, the variance on comments' scores seems to increase as a
result of having this heuristic, which is great—the fact that all
comments had about the same score was a bug, not a feature.

Sadly, we don't have good tooling for proper statistical analysis of the
effect this is having. We'll want to study the effect of this heuristic
more later, as we build tooling and canonical datasets that makes that
analysis feasible. 

We choose to add this heuristic, despite the ambiguous effect on users'
cred, because we think it is principled, and adds meaningful structure
to the graph.

Test plan:
The commit is a pretty straightforward generalization of our existing
GitHub edge logic. All of the interesting logic was thoroughly tested in
the preceding pull, so this commit just tests the integration. Observe
that standard (de)serialization of the edge works, that the snapshot is
updated with a MentionsAuthor reference edge, and that the graph
invariant checker, after update, does not throw errors. Also, I manually
tested this change on the ipfs/go-ipfs repo. (It does not require
regenerating data.)


src/plugins/github/pluginAdapter.js
```diff
@@ -100,6 +100,14 @@ export class StaticPluginAdapter implements IStaticPluginAdapter {
         defaultBackwardWeight: 1 / 16,
         prefix: E._Prefix.references,
       },
+      {
+        forwardName: "mentions author of",
+        backwardName: "has author mentioned by",
+        defaultForwardWeight: 1,
+        // TODO: Probably change this to 0
```
wchargin: Please mark `TODO(@username)` or `TODO(#issueno)`.

src/plugins/github/edges.js
```diff
@@ -47,12 +50,17 @@ export type ReferencesAddress = {|
   +referrer: GithubNode.TextContentAddress,
   +referent: GithubNode.ReferentAddress,
 |};
+export type MentionsAuthorAddress = {|
+  +type: typeof MENTIONS_AUTHOR_TYPE,
+  +reference: MentionsAuthorReference,
```
wchargin: Interesting decision to tie this to the reference type, as opposed to
transcluding the fields directly. If the property names on the reference
type are changed/added/removed, then Flow will catch an error on line
216 of `edges.js`, right? If so, this is fine with me.

decentralion: Correct.

Issue #815, opened 2018-09-13
Create authorship edges between commits and users


decentralion:
This will result in much higher quality signal for repositories that don't have a 1:1 mapping between pulls and commits. TensorFlow is one such example.

PR #816, opened 2018-09-13
Give GitHub plugin support for Commit addresses

decentralion:
The Git plugin owns Commits, but the GitHub plugin also creates commits.
This commit reifies that relationship by making a Git commit address a
valid GitHub structured address. This is precursor work for #815, which
will require adding a commit entity to the GitHub relational view.

Also, this commit surfaces and fixes a minor type bug, wherein a map
from strings to referent addresses was typed to hold any structured
address, rather than just referent addresses.

Test plan: The unit tests confirm that serializing/deserializing a Git
commit address using the GitHub plugin's methods works as intended.
Also, unit tests were added that verify that (de)serializing Git
addresses for non-commit objects is still an error.

### DIFF
--- a/src/plugins/github/nodes.js
+++ b/src/plugins/github/nodes.js
@@ -1,6 +1,7 @@
 // @flow
 
 import {NodeAddress, type NodeAddressT} from "../../core/graph";
+import * as GitNode from "../git/nodes";
 
 export opaque type RawAddress: NodeAddressT = NodeAddressT;
 
@@ -70,7 +71,8 @@ export type StructuredAddress =
   | PullAddress
   | ReviewAddress
   | CommentAddress
-  | UserlikeAddress;
+  | UserlikeAddress
+  | GitNode.CommitAddress;
 
 // Each of these types has 0 or more "AUTHORS" edges, each of which
 // leads to a UserlikeAddress.  Note: It is not true that every
@@ -124,6 +126,14 @@ const _unused_static = (_: CommentableAddress): ParentAddress => _;
 export function fromRaw(x: RawAddress): StructuredAddress {
   function fail() {
     return new Error(`Bad address: ${NodeAddress.toString(x)}`);
+  }
+  if (NodeAddress.hasPrefix(x, GitNode.Prefix.base)) {
+    const structured: GitNode.StructuredAddress = GitNode.fromRaw((x: any));
+    if (structured.type === GitNode.COMMIT_TYPE) {
+      return (structured: GitNode.CommitAddress);
+    } else {
+      throw fail();
+    }
   }
   if (!NodeAddress.hasPrefix(x, GITHUB_PREFIX)) {
     throw fail();
@@ -280,6 +290,8 @@ export function toRaw(x: StructuredAddress): RawAddress {
         default:
           throw new Error((x.subtype: empty));
       }
+    case GitNode.COMMIT_TYPE:
+      return GitNode.toRaw(x);
     default:
       throw new Error(`Unexpected type ${(x.type: empty)}`);
   }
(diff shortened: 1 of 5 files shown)


src/plugins/github/relationalView.js
```diff
@@ -176,6 +176,8 @@ export class RelationalView {
         return this.comment(address);
       case "USERLIKE":
         return this.userlike(address);
+      case "COMMIT":
+        throw new Error("Commit entity not yet supported");
```
wchargin: Would prefer to return `null` here, as clients could conceivably request
the endpoint of a PR’s merged-as edge.

src/plugins/github/nodes.js
```diff
@@ -125,6 +127,12 @@ export function fromRaw(x: RawAddress): StructuredAddress {
   function fail() {
     return new Error(`Bad address: ${NodeAddress.toString(x)}`);
   }
+  if (NodeAddress.hasPrefix(x, GitNode.Prefix.commit)) {
```
wchargin: This dereference really concerns me. See #818. I won’t block this PR on
these grounds.

src/plugins/github/nodes.js
```diff
@@ -125,6 +127,12 @@ export function fromRaw(x: RawAddress): StructuredAddress {
...
+    const commitAddress: GitNode.CommitAddress = (GitNode.fromRaw(
```
wchargin: You’re relying on _another_ implementation detail here, which is that
the `Prefix.commit` prefix maps via `fromRaw` into the specifically the
`Commit` subtype of structured addresses. (You may object that this
obviously holds. But the Git plugin is within its rights to add
addresses like [sourcecred, git, commit, submodule_commit, repo, hash],
introducing a new structured type `SubmoduleCommit` disjoint from other
commit types.)

This comment and the one above it can both be handled by instead
writing (untested):

```js
if (NodeAddress.hasPrefix(x, GitNode.Prefix.base)) {
  const structured: GitNode.StructuredAddress = GitNode.fromRaw((x: any));
  if (structured.type === GitNode.COMMIT_TYPE) {
    return (structured: GitNode.CommitAddress);
  } else {
    throw new Error("unexpected Git node address: " + stringify(structured));
  }
}
```

Would you mind making this change?

(Relatedly: please do test the error case of providing, say, a Git tree
address.)

src/plugins/github/nodes.test.js
```diff
@@ -45,6 +46,10 @@ describe("plugins/github/nodes", () => {
     subtype: "USER",
     login: "decentralion",
   });
+  const commit = (): GitNode.CommitAddress => ({
+    type: GitNode.COMMIT_TYPE,
+    hash: "7ff6fd7b565057b2ff6d9105fd89ae0b372f6c09",
```
wchargin: What is `7ff6fd7b565057b2ff6d9105fd89ae0b372f6c09` a hash of? I don’t
see it in SourceCred or either of our example repositories.

src/plugins/github/relationalView.js
```diff
@@ -393,7 +395,7 @@ export class RelationalView {
     //   https://github.com/sourcecred/sourcecred/pull/416
     // - A # followed by a number, such as #416
     // - An @ followed by a login name, such as @decentralion
-    const refToAddress: Map<string, N.StructuredAddress> = new Map();
+    const refToAddress: Map<string, N.ReferentAddress> = new Map();
```
wchargin: No objection to this change. Consider mentioning in the commit message.

src/plugins/github/nodes.js
decentralion: Done. Added test case as requested.

src/plugins/github/nodes.test.js
decentralion: Just a random hash I had in my working state. I changed it to 000...000

src/plugins/github/nodes.js
decentralion: Fixed per your suggestion.

src/plugins/github/relationalView.js
decentralion: Done.

PR #824, opened 2018-09-13
Add commits in the history to the RelationalView

decentralion:
This builds on #821 so that every commit in the default ref's history is
added as a Commit entity to the GitHub relational view. This means that
these commits are also added to the graph by the GitHub plugin. In
general, this will have no effect on real graphs, because these commits
were already available via the Git plugin.

Test plan:
Observe that the snapshot changes just correspond to new commits being
available to the RelationalView, and correspondingly added to the GitHub
graph. `yarn test --full` passes.

### DIFF
--- a/src/plugins/github/relationalView.js
+++ b/src/plugins/github/relationalView.js
@@ -4,6 +4,8 @@ import {toCompat, fromCompat, type Compatible} from "../../util/compat";
 import stringify from "json-stable-stringify";
 import {parseReferences} from "./parseReferences";
 import * as N from "./nodes";
+import dedent from "../../util/dedent";
+
 // Workaround for https://github.com/facebook/flow/issues/6538
 import type {
   RepoAddress,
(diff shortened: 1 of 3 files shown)


src/plugins/github/relationalView.js
```diff
@@ -288,6 +288,16 @@ export class RelationalView {
...
     this._repos.set(raw, entry);
+    this._addCommitHistory(json);
+  }
+
+  _addCommitHistory(json: RepositoryJSON) {
+    if (json.defaultBranchRef != null) {
+      const target = json.defaultBranchRef.target;
+      if (target.__typename === "Commit") {
+        target.history.nodes.forEach((commit) => this._addCommit(commit));
+      }
```
wchargin: If I understand correctly, we believe (at least weakly) that flow can
never reach here. Consequently, we don’t know what the right thing to do
is. Can we throw an error in this case, so that (a) we don’t do anything
unexpected, and (b) we can find out the cause?

PR #832, opened 2018-09-14
Modify parseReferences to detect refs to commits

teamdandelion:
We add a new function, `findCommitReferences`, which can find both
explicit url references to commits, and commit hashes.

Since the commit url includes the commit hash, some extra logic is added
to deduplicate them in this instance. Tests verify that this is done
properly.

Test plan: Unit tests cover the cases of having commit hashes, having
commit urls, and having both at once.

### DIFF
--- a/src/plugins/github/parseReferences.js
+++ b/src/plugins/github/parseReferences.js
@@ -29,6 +29,7 @@ function parseReferencesFromRawString(textBlock: string): ParsedReference[] {
     ...findRepoNumericReferences(textBlock),
     ...findGithubUrlReferences(textBlock),
     ...findUsernameReferences(textBlock),
+    ...findCommitReferences(textBlock),
   ];
 }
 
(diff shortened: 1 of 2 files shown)


src/plugins/github/parseReferences.js
```diff
@@ -98,6 +99,47 @@ function findGithubUrlReferences(textBlock: string): ParsedReference[] {
   }));
 }
 
+function findCommitReferences(textBlock: string): ParsedReference[] {
+  const hashReferences = findAllMatches(
+    /(?:\W|^)([a-fA-F0-9]{40})(?=\W|$)/gm,
```
wchargin: Git is in the process of moving from SHA-1 to SHA-256, which changes the
hex digest from 40 characters to 64. Do you want to modify your regular
expression, changing `{40}` to `{40,}`, to be forward-compatible?

See [`git/git:Documentation/technical/hash-function-transition.txt`[1]
for more details.

[1]: https://github.com/git/git/blob/1d4361b0f344188ab5eec6dcea01f61a3a3a1670/Documentation/technical/hash-function-transition.txt

src/plugins/github/parseReferences.js
```diff
@@ -98,6 +99,47 @@ function findGithubUrlReferences(textBlock: string): ParsedReference[] {
...
+  }));
+  const commitUrlRegex = new RegExp(
+    "" +
+      /(?:\W|^)/.source +
+      "(" +
+      /https?:\/\/github.com\//.source +
+      `(?:${githubOwnerPattern})` +
+      /\//.source +
+      `(?:${githubRepoPattern})` +
+      /\/commit\/([a-z0-9]{40})/.source +
```
wchargin: `[a-z0-9]` should be `[a-fA-F0-9]`.

Do you want to add a test case with uppercase in the URL?

src/plugins/github/parseReferences.js
```diff
@@ -98,6 +99,47 @@ function findGithubUrlReferences(textBlock: string): ParsedReference[] {
...
+      ")" +
+      /(?=[^\w/]|$)/.source,
+    "gm"
+  );
+  const urlsAndHashes = findAllMatches(commitUrlRegex, textBlock).map((x) => ({
+    url: x[1],
+    hash: x[2],
+  }));
+  // Every urlReference will also generate a hash reference (because the url contains the hash).
```
wchargin: Please wrap (preferably at 72, but certainly less than 96).

src/plugins/github/parseReferences.test.js
```diff
@@ -182,6 +190,43 @@ https://github.com/sourcecred/exa_mple-git.hub/pull/3#issuecomment-369162222
...
   });
 
+  describe("commit references", () => {
+    const commitHash = "6bd1b4c0b719c22c688a74863be07a699b7b9b34";
+    it("are valid if they are 40-chars long and all hex", () => {
+      expect(parseReferences(commitHash)).toEqual([
+        {refType: "BASIC", ref: commitHash},
+      ]);
+    });
+    it("are canonicalized  to lower-case capitalization", () => {
```
wchargin: nit: extra space

PR #839, opened 2018-09-14
Request reactions data from GitHub

decentralion:
This commit updates the GitHub graphql query to also fetch reactions.
We update the JSON typedefs to include this new information, add
continuations from comments, and update existing continuation and query
code. Also, I added a safety check when updating comments for issues
that was previously unnecessary but is now needed.

Test plan:
- `yarn test --full` passes.
- Setting the page limits to 1 and running on the example-github does
not error with unexhausted pages, and loads all the expected reactions.
- Running on a larger repository (go-ipfs) works as expected.
- I have written dependent code that consumes these reactions in the
RelationalView, and works as intended, which suggests that the type
signatures are correct.


src/plugins/github/graphql.js
```diff
@@ -885,6 +983,32 @@ function commitHistoryFragment(): FragmentDefinition {
...
+  | "THUMBS_DOWN"
+  | "LAUGH"
+  | "CONFUSED"
+  | "HEART"
+  | "HOORAY";
+
+export type ReactionJSON = {|
+  +id: string,
+  +content: ReactionContent,
+  +user: AuthorJSON,
```
wchargin: Please mark that this is nullable, even though we do not know why.

(If a user deletes their account, their reactions are removed.)

src/plugins/github/graphql.js
```diff
@@ -324,7 +326,7 @@ function* continuationsFromIssue(
   path: $ReadOnlyArray<string | number>
 ): Iterator<Continuation> {
   const b = build;
-  if (result.comments.pageInfo.hasNextPage) {
+  if (result.comments && result.comments.pageInfo.hasNextPage) {
```
wchargin: Curious whether your tests would have caught the absence of this change?

src/plugins/github/graphql.js
```diff
@@ -885,6 +983,39 @@ function commitHistoryFragment(): FragmentDefinition {
   ]);
 }
 
+export const THUMBS_UP = "THUMBS_UP";
```
wchargin: Can you namespace these (preferably into an object)?

src/plugins/github/graphql.js
decentralion: I think so.

PR #841, opened 2018-09-17
Add tensorflow-gardener to the list of bots

decentralion:
Test plan: I verified that the spelling is correct


src/plugins/github/bots.js
```diff
@@ -16,5 +16,6 @@ export function botSet(): Set<string> {
     "stickler-ci",
     "tensorflow-jenkins",
     "tensorflowbutler",
+    "tensorflow-gardener",
```
wchargin: Can this be placed in sorted order, please?

PR #848, opened 2018-09-17
Prepare to enable flow-type eslint rules

decentralion:
This commit upgrades the flow-type eslint plugin to latest, and writes
new rules into the eslintrc. To keep the diff clean, the rules are
disabled: I will turn them on individually (fixing errors) in followon
commits.

Test plan: `yarn test`.
Uncommenting the lines produces many lint errors (but the linter still operates as expected).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,7 @@
 // @flow
 module.exports = {
+  parser: "babel-eslint",
+  plugins: ["flowtype"],
   extends: "react-app",
   rules: {
     "no-unused-vars": [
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -12,5 +14,9 @@ module.exports = {
     ],
     "no-use-before-define": ["off"],
     "no-useless-constructor": ["off"],
+    // TODO(@decentralion): Enable this rule.
+    //    "flowtype/no-mutable-array": [2],
+    // TODO(@decentralion): Enable this rule.
+    //    "flowtype/require-exact-type": [2, "always"],
```
wchargin: Cool. There are definitely other rules that I would like to add. I’ll PR
them after this is landed.

Issue #622, opened 2018-08-08
Design and implementation of the Mirror module (formerly: Incremental/resumable GraphQL loading notes)


wchargin:
Documenting design efforts here.

*Editor’s note:* Comments before 2018-08-10 describe older ideas of how the system might be designed, and were never implemented. Discussion of the design that we use today starts here: <https://github.com/sourcecred/sourcecred/issues/622#issuecomment-412228254>

wchargin:
Daily dump of some stuff that is reasonably solid.

Object model:

```js
// A reference to a GraphQL node. The typename is its GraphQL type, as
// defined in a relevant schema. The ID is the unique ID assigned by the
// remote server (i.e., GitHub).
type Ref = {|+typename: Typename, +id: string|};

// The interface used by the GitHub plugin. Described below.
declare class GraphqlMirror {
  constructor(options: {|
    +storage: Storage,
    +schema: Schema,
    +ttl: Duration
  |}): void;
  update(root: Ref): void;
  denormalize(root: Ref): any;
}

// An abstraction over persistent storage.
interface Storage {
  write(filename: string, contents: string): void;
  read(filename: string): string | void;
  sync(): Promise<void>;
}

// Storage backed by a physical filesystem; `sync` writes to disk.
declare class DiskStorage implements Storage {
  constructor(basedir: string): void;
  ...Storage;
}

// Memory-only storage, for tests and such; `sync` is NOP.
declare class MemoryStorage implements Storage {
  constructor(initialState: Map<string, string>): void;
  ...Storage;
}
```

Top-level options (constructor to `GraphqlMirror`):

  - `storage`: Persistent storage, wholly scoped to the plugin and
    repository. In principle, this could be structurally shared across
    repositories, but we’ll keep it simple for now.
  - `schema`: A structured sub-GraphQL schema, as described above. This
    is required to remain fixed across invocations with the same cache
    directory; it may be persisted on disk to check for
    data-incompatible code upgrades.
  - `ttl`: Time to live; a duration after which data is to be considered
    stale.\* This may vary across invocations.

\* For liveness, we may wish to relax the definition of staleness to
prevent a case wherein our fetch rate can’t keep up with the TTL.
Semantics to be determined. This isn’t very important for an initial
version of the system, so we omit discussion.

Desired operations (methods on `GraphqlMirror`):

  - `update(root: Ref)`: Make any needed updates for `root` and any
    transitive dependencies. Writes to persistent storage, `sync`ing
[...]

PR #849, opened 2018-09-17
mirror: embed GraphQL schema into SQL

wchargin:
Summary:
This commit augments the `Mirror` constructor to turn the provided
GraphQL schema into a SQL schema, with which it initializes the backing
database. The schema is roughly as originally described in #622, with
some changes (primarily: we omit `WITHOUT ROWID`; we add indexes; we
store `total_count` on connections; and we use milliseconds instead of
seconds for epoch time).

Test Plan:
Unit tests included, with full coverage; run `yarn unit`.

wchargin-branch: mirror-sql-schema

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -85,11 +85,12 @@ export class Mirror {
    * of each other and of own-data.
    *
    * Note that any object in the database should have entries in the
-   * `connections` and `links` table for all relevant fields, even if
-   * the node has never been updated. This is for convenience of
-   * implementation: it means that the first fetch for a node is the
-   * same as subsequent fetches (a SQL `UPDATE` instead of first
-   * requiring an existence check).
+   * `connections` and `links` table for all relevant fields, as well as
+   * an entry in the relevant primitives table, even if the node has
+   * never been updated. This is for convenience of implementation: it
+   * means that the first fetch for a node is the same as subsequent
+   * fetches (a SQL `UPDATE` instead of first requiring an existence
+   * check).
    *
    * Finally, a table `meta` is used to store metadata about the mirror
    * itself. This is used to make sure that the mirror is not loaded
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -3,6 +3,7 @@
 import type Database from "better-sqlite3";
 import stringify from "json-stable-stringify";
 
+import dedent from "../util/dedent";
 import * as Schema from "./schema";
 
 /**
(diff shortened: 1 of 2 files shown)


src/graphql/mirror.js
```diff
@@ -35,6 +36,39 @@ export class Mirror {
     this._initialize();
   }
 
+  /**
+   * Embed the GraphQL schema into the database, initializing it for use
+   * as a mirror.
+   *
+   * We store the data for an object across three tables, depending on
```
decentralion: This blcok is confusing. Are there three tables for each kind of object? or are there three tables overall? please be more specific (an example would be nice)

decentralion: It would be nice if we start at a high level: the database has one connections table for all objects' connections, one links table, and then, for each object type, one type-specific table that has columns for every primitive field. (assuming this is true). getting in to the columnar structure for the connections and links table would be great. also, how does the meta table fit into this precisely? does every object type have its own meta entry?

src/graphql/mirror.js
```diff
@@ -35,6 +36,39 @@ export class Mirror {
...
+   * the field type:
+   *
+   *   - Connections are stored in the `connections `table.
+   *   - Node references are stored in the `links` table.
+   *   - Primitive data is stored in a type-specific data table.
+   *
+   * We refer to node and primitive data together as "own data", because
+   * this is the data that can be queried uniformly for all elements of
+   * a type; querying connection data, in contrast, requires the
+   * object-specific end cursor.
```
decentralion: it would be great to have an example here, i'm a  little confused

src/graphql/mirror.js
```diff
@@ -35,6 +36,39 @@ export class Mirror {
...
+   *
+   * The aforementioned tables are each keyed by object ID. An object's
+   * metadata appears in the `objects` table, including the object's
+   * typename and the last update for the object's own-data. Each
+   * connection has its own last update value, because connections can
+   * be updated independently of each other and of own-data.
+   *
+   * Note that any object in the database should have entries in the
+   * `connections` and `links` table even if the node has never been
+   * updated.
```
decentralion: I'm a little confused how this is the case. It would be great to have a little example of the lifecycle and how/why we maintain this property

src/graphql/mirror.js
```diff
@@ -43,26 +77,170 @@ export class Mirror {
...
+                    throw new Error("invalid field name: " + fieldname);
+                  }
+                  primitiveFieldNames.push(fieldname);
+                  break;
+                // istanbul ignore next
+                default:
+                  throw new Error((field.type: empty));
+              }
+            }
+            const tableName = `"data_${typename}"`;
```
decentralion: Note: you said you want to change this to primitives_

wchargin:
@jart: If you would like to review this, you’re more than welcome to. :-)

wchargin:
@decentralion: Regarding the own-data questions: would it help you to
see the implementation of `_queryOwnData`? (This will come in a later
PR.) I’ve posted the implementation at:
<https://gist.github.com/wchargin/08809d782e08da11f239a4ee1f9a552b>

wchargin:
Thanks. I’ll go ahead and merge this. Post hoc reviews are still
welcome: this design is not final.

PR #853, opened 2018-09-17
Use read only arrays by default

decentralion:
This enables an eslint rule disallowing the use of mutable arrays in
favor of read only arrays. There are a few cases where we genuinely need
mutable arrays, so we added a lint exception. Note that in the future,
you can still write `[]` for convenience and then call `yarn lint --fix`
to fix them all up before committing.

There are a few cases where we have now marked function outputs as
readonly when returning a mutable instance would be totally fine. We
could mark them all as exceptions, but I like adopting a policy of "read
only unless actually required mutable" because it is simpler to
enforce/maintain. A caller who needs to mutate their copy can always
just call  `.slice()`.

Test plan: `yarn test`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,8 +14,7 @@ module.exports = {
     ],
     "no-use-before-define": ["off"],
     "no-useless-constructor": ["off"],
-    // TODO(@decentralion): Enable this rule.
-    //    "flowtype/no-mutable-array": [2],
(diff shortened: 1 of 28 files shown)


decentralion:
Based on the [review feedback](https://github.com/sourcecred/sourcecred/pull/853#pullrequestreview-156158114) I've decided to abandon this change (in favor of #872).

PR #854, opened 2018-09-17
Require exact types by default

decentralion:
This enables the flowtype lint rule for requiring exact object types. I
added an exception in queries.js because it was a huge mess to properly
label it as exact. I also added a flyby fix to mark some objects as
readonly at @wchargin's suggestion.

Test plan: `yarn test`

### DIFF
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -51,7 +51,10 @@ export type InlineFragment = {|
   +selections: Selection[],
 |};
 
-export type Arguments = {[string]: Value};
+// The next line should be exact, but flow's support isn't great
+// and doing so is a miserable change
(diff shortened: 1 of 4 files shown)


src/graphql/queries.js
```diff
@@ -238,7 +241,7 @@ function formatList<T>(
  * join the keys and their corresponding results with a formatter.
  */
 function formatObject<T>(
-  object: {[string]: T},
+  object: {+[string]: T},
```
wchargin: This is not exact. Why does it not raise a lint error?

decentralion:
You're right, the rule doesn't seem to be working well. I'm abandoning this change (in favor of #872).

PR #855, opened 2018-09-18
Fix a crash when a review comment has >5 reactions

decentralion:
Our continuation-fetching code failed to properly get continuations for
pull request review comments, because it was only asking for more
reactions on `"IssueComment"` fragments. This caused the
`ensureNoMorePages` function to properly throw an error rather than
proceding with incomplete data.

This commit fixes the root cause by splitting
`continuationsFromComment`into `continuationsFromReviewComment` and
`continuationsFromIssueComment`. (Pull and issue comments are both
considered 'IssueComment's.) The example-github repository has been
updated to include 10 reactions to a single review comment; the
example-data was updated in this commit, and all reactions have been
loaded.

I've also added a `console.error` statement in `ensureNoMorePages`. This
only triggers when the program is about to fail, and it's useful for
debugging.

Paired with @wchargin


src/plugins/github/fetchGithubRepo.js
```diff
@@ -188,6 +188,7 @@ function ensureNoMorePages(result: any, path = []) {
   }
   if (result.pageInfo) {
     if (result.pageInfo.hasNextPage) {
+      console.error(result);
```
wchargin: I’m okay with this, but only because the code is untested (though
there’s no reason that it should be).

PR #860, opened 2018-09-19
Display urls in the cred explorer

### LINKED ISSUE #590: Cred Explorer should hyperlink to GitHub entities
It's frustrating that the cred explorer shows GitHub entities without offering a link to them. We should add hyperlinks. We have urls available for every entity via the [relationalView](https://github.com/sourcecred/sourcecred/blob/master/src/plugins/github/relationalView.js#L540) so it shouldn't be hard to add this.

A simple approach would be to just make the whole nodeDescription row into a hyperlink (when available). We can also hyperlink directly to any Git commit, but that isn't as important.

decentralion:
This commit modifies the plugin adapter's `nodeDescription` method so
that it may return a React node.

This enables the GitHub plugin's `nodeDescription` method to include
hyperlinks directly to the referenced content on GitHub. This makes
examining e.g. comment cred much easier.

I've also made two other changes to the descriptions:
- Pull requests diffs now color-encode the additions and deletions
- Descriptions for comments and reviews no longer include the authors

The Git plugin's behavior is unchanged.

Test plan:
I loaded a large repository in the cred explorer and verified that
exploring comments and pulls and issues is much easier. The descriptions
are as expected for every category of node. Snapshot tests updated.

Fixes #590.

### DIFF
--- a/src/app/adapters/pluginAdapter.js
+++ b/src/app/adapters/pluginAdapter.js
@@ -1,5 +1,6 @@
 // @flow
 
+import {type Node as ReactNode} from "react";
 import {Graph, type NodeAddressT, type EdgeAddressT} from "../../core/graph";
 import type {Assets} from "../assets";
 import type {Repo} from "../../core/repo";
@@ -30,6 +31,6 @@ export interface StaticPluginAdapter {
 
 export interface DynamicPluginAdapter {
   graph(): Graph;
-  nodeDescription(NodeAddressT): string;
+  nodeDescription(NodeAddressT): ReactNode;
   static (): StaticPluginAdapter;
 }
--- a/src/plugins/github/render.js
+++ b/src/plugins/github/render.js
@@ -1,31 +1,96 @@
 // @flow
 
+import React, {type Node as ReactNode} from "react";
 import * as R from "./relationalView";
 
-export function description(e: R.Entity) {
-  const withAuthors = (x: R.AuthoredEntity) => {
-    const authors = Array.from(x.authors());
-    if (authors.length === 0) {
-      // ghost author - probably a deleted account
-      return "";
-    }
-    return "by " + authors.map((x) => description(x)).join(" & ") + " ";
-  };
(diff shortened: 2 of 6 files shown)


src/app/adapters/pluginAdapter.js
```diff
@@ -30,6 +31,6 @@ export interface StaticPluginAdapter {
 
 export interface DynamicPluginAdapter {
   graph(): Graph;
-  nodeDescription(NodeAddressT): string;
+  headline(NodeAddressT): ReactNode;
```
wchargin: > ```js
> headline(NodeAddressT): ReactNode;
> ```

I don’t think that this is the right API. It’s more typical to return a
`React.ComponentType<{|+address: NodeAddressT|}>`. This also has the
benefit that clients can provide a stateful implementation if they want.

src/plugins/github/render.test.js
```diff
@@ -1,16 +1,16 @@
...
 import {exampleEntities} from "./example/example";
-import {description} from "./render";
+import {headline} from "./render";
 
 describe("plugins/github/render", () => {
-  it("descriptions are as expected", () => {
+  it("headlines are as expected", () => {
     const examples = exampleEntities();
-    const withDescriptions = {};
+    const withheadlines = {};
```
wchargin: capitalization (rogue `%s`?)

src/plugins/github/render.js
```diff
@@ -1,31 +1,96 @@
...
-    const authors = Array.from(x.authors());
-    if (authors.length === 0) {
-      // ghost author - probably a deleted account
-      return "";
-    }
-    return "by " + authors.map((x) => description(x)).join(" & ") + " ";
-  };
+function EntityUrl(props) {
+  return (
+    <a href={props.entity.url()} target="_blank">
```
wchargin: Can you use `rel="nofollow noopener"`, and maybe `noreferrer` if you want?

PR #871, opened 2018-09-20
Add `mergeRepository` for merging Git repositories

decentralion:
This commit adds a utility method, `mergeRepository`, which can merge
multiple Git repository data structures. `loadGitData` has been updated
to create a merged repository and then subsequently generate a graph
from it.

Test plan:
New unit tests were added. `yarn test --full` passes. Loading a project
and viewing its git data in the cred explorer works.

### DIFF
--- a/src/plugins/git/mergeRepository.js
+++ b/src/plugins/git/mergeRepository.js
@@ -1,0 +1,30 @@
+// @flow
+
+import deepEqual from "lodash.isequal";
+import type {Repository} from "./types";
+
+export function mergeRepository(
+  repositories: $ReadOnlyArray<Repository>
+): Repository {
+  const newRepository = {commits: {}, trees: {}};
+  for (const {trees, commits} of repositories) {
+    for (const treeHash of Object.keys(trees)) {
+      const existingTree = newRepository.trees[treeHash];
+      if (existingTree != null && !deepEqual(existingTree, trees[treeHash])) {
+        throw new Error(`Conflict between trees at ${treeHash}`);
+      }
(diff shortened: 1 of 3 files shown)


src/plugins/git/mergeRepository.js
```diff
@@ -0,0 +1,18 @@
...
+
+import type {Repository} from "./types";
+
+export function mergeRepository(
+  repositories: $ReadOnlyArray<Repository>
+): Repository {
+  const newRepository = {commits: {}, trees: {}};
+  for (const {trees, commits} of repositories) {
+    for (const treeHash in trees) {
+      newRepository.trees[treeHash] = trees[treeHash];
```
wchargin: Let’s fail here if the contents are not deep-equal. (likewise below.)

src/plugins/git/mergeRepository.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import type {Repository} from "./types";
+
+export function mergeRepository(
+  repositories: $ReadOnlyArray<Repository>
+): Repository {
+  const newRepository = {commits: {}, trees: {}};
+  for (const {trees, commits} of repositories) {
+    for (const treeHash in trees) {
```
wchargin: Looks like `for`-`in` is safe when the object in question is just a struct; if you have inherited/non-enumerable properties, it’s less so. This is not technically guaranteed for us because the `commits` and `trees` are not exact. But it should be safe with our existing code thus far. Up to you whether to convert to `Object.keys`.

decentralion: Done.

PR #872, opened 2018-09-20
Remove obsolete eslint TODOs

decentralion:
I wanted to add two new flowtype lint rules. However, as seen in #853
and #854, these rules don't work very well. As such, I'm giving up on
those two and removing the TODOs.

Test plan: n/a


wchargin:
OK, no problem.

PR #874, opened 2018-09-20
Git plugin serializes the repository and graph

decentralion:
This modifies the behavior when loading the Git plugin so that it
serializes the Repository as well as the graph. This will allow us to
get extra information, like the commit headline, to the Git plugin in
the frontend.

As an added bonus, we can now refactor `loadRepositoryTest` to depend on
`sourcecred.js load` rather than `loadAndPrintRepository`. As this was
the only use for `loadAndPrintRepository`, we can safely delete it. This
improves our test quality because it means we are also testing the
actual CLI behavior.

Note that the switch from using `stringify` to `json.tool` for
pretty-printing has resulted in a trivial diff in the snapshot.


src/plugins/git/loadRepositoryTest.sh
```diff
@@ -23,8 +23,12 @@ usage() {
 
 fetch() {
   tmpdir="$(mktemp -d)"
-  node "${SOURCECRED_BIN:-./bin}/createExampleRepo.js" "${tmpdir}"
-  node "${SOURCECRED_BIN:-./bin}/loadAndPrintGitRepository.js" "${tmpdir}"
+  SOURCECRED_DIRECTORY="${tmpdir}" \
+    node "${SOURCECRED_BIN:-./bin}/sourcecred.js" \
+    load --plugin git \
+    sourcecred/example-git
+  cat "${tmpdir}"/data/sourcecred/example-git/git/repository.json \
```
wchargin: This is a “useless use of `cat(1)`”. You can simply:

```shell
<"${tmpdir}"/data/sourcecred/example-git/git/repository.json \
  python -m json.tool
```

(The redirection can also go after the command, if you prefer.)

It looks like `python -m json.tool` also accepts a filename as a
command-line argument? If this is the case, then it’s definitely better
to use that form. Tools can be orders of magnitude faster with this
formulation, because they get a file as opposed to an arbitrary input
stream, which grants them more fine-grained control over I/O. (I’ve
noticed this especially with `jq(1)`.)

PR #875, opened 2018-09-20
mirror: add internal method `_findOutdated`

wchargin:
Summary:
This function finds all objects whose own data has not been updated
since a given time, and all connections whose entries have not been
updated since that time.

Note that this is scoped to the entirety of the database. In #622,
I discussed using a recursive common table expression to identify only
those transitive dependencies of the root. I think that this is overkill
for the `_findOutdated` method: you’ll usually want to update everything
in the database. Don’t worry—the cool recursive query will still be used
in the `extract` function. :-)

This commit makes progress toward #622.

Test Plan:
Unit tests added, with full coverage; run `yarn unit`.

wchargin-branch: mirror-findoutdated

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -353,6 +353,58 @@ export class Mirror {
       addConnection.run({id, fieldname});
     }
   }
+
+  /**
+   * Find objects and connections that are not known to be up-to-date.
+   *
+   * An object is up-to-date if its own data has been loaded at least as
+   * recently as the provided date.
+   *
+   * A connection is up-to-date if it has been fetched at least as
+   * recently as the provided date, and at the time of fetching there
+   * were no more pages.
+   */
+  _findOutdated(since: Date): QueryPlan {
+    const db = this._db;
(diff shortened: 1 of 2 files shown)


src/graphql/mirror.js
```diff
@@ -353,6 +353,58 @@ export class Mirror {
...
+   * recently as the provided date, and at the time of fetching there
+   * were no more pages.
+   */
+  _findOutdated(since: Date): QueryPlan {
+    const db = this._db;
+    return _inTransaction(db, () => {
+      const objects: $PropertyType<QueryPlan, "objects"> = db
+        .prepare(
+          dedent`\
+            SELECT typename AS typename, id AS id
```
wchargin: Re: `typename AS typename`, see:
  - <https://sqlite.org/c3ref/column_name.html> ("The name of a result column is the value of the "AS" clause for that column, if there is an AS clause. If there is no AS clause then the name of the column is unspecified and may change from one release of SQLite to the next.")
  - <https://github.com/JoshuaWise/better-sqlite3/issues/167>

PR #876, opened 2018-09-20
Load commits' short hash and summary

decentralion:
This modifies the Git Commit type to includea short hash and a oneline
summary, and modifies `loadRepository` so that we actually get that
data.

Test plan: Observe that the snapshot is updated, and the updates are
correct. `yarn test --full` passes.

### DIFF
--- a/src/plugins/git/types.js
+++ b/src/plugins/git/types.js
@@ -7,4 +7,8 @@ export type Hash = string;
 export type Commit = {|
   +hash: Hash,
   +parentHashes: $ReadOnlyArray<Hash>,
+  // a shorter version of the hash;
+  // shortHash is not guaranteed unique.
+  +shortHash: string,
+  +summary: string, // Oneline commit summary
 |};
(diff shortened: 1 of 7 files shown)


wchargin:
Formatting with `--pretty=%H %h %P | %s`, then splitting on `|` and
trimming, has the wrong behavior if a commit message includes a literal
pipe, or includes leading or trailing whitespace.

How about using `--format=%H %h %P|%s`, then taking `indexOf("|")` and a
substring to split on just the first pipe?

src/plugins/git/types.js
```diff
@@ -7,4 +7,6 @@ export type Hash = string;
 export type Commit = {|
   +hash: Hash,
   +parentHashes: $ReadOnlyArray<Hash>,
+  +shortHash: string, // short unambiguous identifier
```
wchargin: I take issue with the use of _unambiguous_ here, because it means that `mergeRepository` is no longer correct. What should we do about this?

PR #879, opened 2018-09-20
Cred explorer: display commit short hash + summary

decentralion:
This modifies how commits are displayed in the cred explorer. Rather
than printing the full hash, we now print a short hash followed by the
summary.

Test plan:
Snapshot is updated, also I tested it by running SourceCred on a real
repository.


src/plugins/git/render.test.js
```diff
@@ -2,15 +2,32 @@
...
+        shortHash: exampleHash.slice(7),
+        summary: "This is an example commit",
+        parentHashes: [],
+      },
+    },
+  });
+
   it("commit snapshots as expected", () => {
-    expect(description(examples.commit())).toMatchSnapshot();
+    expect(description(exampleCommit, exampleRepository)).toMatchSnapshot();
```
wchargin: Not a product of this commit, but it seems a bit sketchy to be
snapshotting an arbitrary React node. No change requested—I’ll fix these
up in a separate PR.

src/plugins/git/pluginAdapter.js
```diff
@@ -42,32 +43,50 @@ export class StaticPluginAdapter implements IStaticPluginAdapter {
...
-  constructor(graph: Graph) {
+  +_repository: Repository;
+  constructor(graph: Graph, repository: Repository) {
     this._graph = graph;
+    this._repository = repository;
   }
   graph() {
     return this._graph;
   }
+  repository() {
```
wchargin: Remove? Not specified by interface, never used.

src/plugins/git/render.js
```diff
@@ -1,19 +1,21 @@
...
-    case "BLOB":
-      return `blob ${address.hash}`;
-    case "TREE_ENTRY":
-      return `entry ${JSON.stringify(address.name)} in tree ${
-        address.treeHash
-      }`;
+      const hash = address.hash;
+      const commit = repository.commits[hash];
+      if (commit == null) {
+        throw new Error(`Unable to find commit with hash ${hash}`);
```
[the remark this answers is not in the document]
decentralion: Good point. I've changed it to error to console instead (and tested this behavior).

src/plugins/git/render.test.js
wchargin: Well, never mind. Enzyme is still broken. This will have to do.

src/plugins/git/render.test.js
```diff
@@ -2,15 +2,35 @@
...
+  const exampleHash = "3715ddfb8d4c4fd2a6f6af75488c82f84c92ec2f";
+  const exampleCommit: GN.CommitAddress = Object.freeze({
+    type: GN.COMMIT_TYPE,
+    hash: exampleHash,
+  });
+  const exampleRepository: Repository = Object.freeze({
+    commits: {
+      [exampleHash]: {
+        hash: exampleHash,
+        shortHash: exampleHash.slice(7),
```
wchargin: `slice(0, 7)` :-)

PR #883, opened 2018-09-21
mirror: query and process own-data updates

wchargin:
Summary:
This commit adds internal functions to (a) emit a GraphQL query to fetch
data for own-data of an object, and (b) ingest the results of said query
back into the database.

The API and implementation differ from the connection-updating analogues
introduced in #878 in that the query for own data is independent of an
object’s ID: it depends only on the object’s type. This affords us more
flexibility in composing queries.

As described in a internal documentation comment, values are stored in
the database in JSON-stringified form: we cannot use the obvious raw SQL
values, because there is no native encoding of booleans (`0`/`1` is
conventional), and we need to distinguish them from other data types.
There are other ways to solve this problem. Notably:

 1. We could take inspiration from OCaml: encode stronger static types
    and a simpler runtime representation. That is, we could change the
    schema field types from simply “primitive” to the various specific
    primitive types. Then, when reading data out from the database, we
    could reinterpret the values appropriately.

 2. We could take advantage of the fact that we are not using all of
    SQLite’s data types. In particular, we do not store anything as a
    binary blob, so we could encode `false` as a length-0 zeroblob and
    `true` as a length-1 zeroblob, for instance. Again, when reading
    data out from the database, we would reinterpret the values—but in
    this approach we would not need an explicit schema from the user.

For now, we take the easiest and simplest approach just to get ourselves
off the ground. We can easily move to the second option described above
later.

This commit makes progress toward #622.

Test Plan:
Unit tests included, with full coverage. While these tests check that
the GraphQL queries are as expected, they cannot check that they are
actually valid in production. To check this, follow the instructions in
the added snapshot test.

wchargin-branch: mirror-own-data-updates

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -69,6 +69,7 @@ export class Mirror {
    * the corresponding table.
    *
    * In more detail:
+   *
    *   - The `connections` table has a row for each `(id, fieldname)`
    *     pair, where `fieldname` is the name of a connection field on the
    *     object with the given ID. This stores metadata about the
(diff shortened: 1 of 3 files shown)


src/graphql/mirror.js
```diff
@@ -684,6 +697,236 @@ export class Mirror {
...
+      throw new Error(
+        "Cannot update data for non-object type: " +
+          `${JSON.stringify(typename)} (${this._schema[typename].type})`
+      );
+    }
+
+    const db = this._db;
+    const objectType = this._schemaInfo.objectTypes[typename];
+
+    // Convert a prepared statement into a JS function that executes
```
decentralion: Pull this out?

src/graphql/mirror.js
```diff
@@ -684,6 +697,236 @@ export class Mirror {
...
+        const value: PrimitiveResult | NodeFieldResult = entry[fieldname];
+        const linkData: NodeFieldResult = (value: any);
+        if (linkData === undefined) {
+          const s = JSON.stringify;
+          throw new Error(
+            `Missing node reference ${s(fieldname)} on ${s(entry.id)} ` +
+              `of type ${s(typename)} (got ${(linkData: empty)})`
+          );
+        }
+        let childId = null;
```
decentralion: possibly we can abstract over this pattern? it seems kind of generic and we might like to fold it out of this particular update

wchargin:
Thanks for your comments. I’ve factored out some pieces as you suggest
(see #888, #889) and will wait until those merge to rebase and land this
pull request.

wchargin:
Diff updated; please take a look when you get a chance. I’ve used the
factors introduced in #888 and #889, and transposed the implementation
of `updateOwnData` in (I think) the way that you suggested. Further
thoughts appreciated.

PR #884, opened 2018-09-21
Git: track the repositories containing each commit

decentralion:
This modifies the Git `Repository` data structure so that for every
commit, we track the `RepoId`s of repos containing that commit. This way
we will be able to do things like hyperlink to the right url for that
commit.

`loadRepository` has been modified to set the initial `repoId`.
`mergeRepository` has been updated to ensure that it concatenates the
`repoId`s properly.
Tests were added for both cases.

The example-git snapshot has been updated accordingly.

Test plan: `yarn test --full`

### DIFF
--- a/src/plugins/git/loadRepository.js
+++ b/src/plugins/git/loadRepository.js
@@ -9,9 +9,15 @@
  */
 // @flow
 
+import * as MapUtil from "../../util/map";
 import type {GitDriver} from "./gitUtils";
-import type {Repository, Hash, Commit} from "./types";
+import type {Repository, Commit} from "./types";
 import {localGit} from "./gitUtils";
+import {
+  repoIdToString,
+  type RepoId,
+  type RepoIdString,
+} from "../../core/repoId";
 
 /**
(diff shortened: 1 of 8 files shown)


decentralion:
I updated the pull per your suggestions.

src/plugins/git/loadRepository.js
```diff
@@ -32,18 +39,18 @@ export function loadRepository(
...
-
-function objectMap<T: {+hash: Hash}>(ts: $ReadOnlyArray<T>): {[Hash]: T} {
-  const result = {};
-  ts.forEach((t) => {
-    result[t.hash] = t;
+  const rawCommits = findCommits(git, rootRef);
+  const commits = MapUtil.toObject(new Map(rawCommits.map((x) => [x.hash, x])));
+  const repoIdString = repoIdToString(repoId);
+  const repoIdStringSet: () => {[RepoIdString]: true} = () => ({
+    [((repoIdString: RepoIdString): any)]: true,
```
wchargin: Note for posterity: this is <https://github.com/facebook/flow/issues/6569>.

PR #887, opened 2018-09-22
Hyperlink Git commits to GitHub

decentralion:
This modifies the `nodeDescription` code for the Git plugin so that when
given a Git commit, it will hyperlink to that commit on GitHub. It does
this by looking up the corresponding `RepoId`s from the newly-added
`commitToRepoId` field in the `Repository` (#884).

Per a [suggestion in review], rather than hardcoding the GitHub url
logic in the Git plugin, we provide them via a `GitGateway`.

[suggestion in review]: https://github.com/sourcecred/sourcecred/pull/887#issuecomment-424059649

When no `RepoId` is found, it errors to console and does not include a
hyperlink. When multiple `RepoId`s are available, it chooses to link to
one arbitrarily. (In the future, we could amend this behavior to add
links to every valid repo). This behavior is tested.

Test plan:
I ran the application on newly-generated data and verified that it sets
up commit hyperlinks appropriately. Also, see unit tests.

### DIFF
--- a/src/plugins/git/pluginAdapter.js
+++ b/src/plugins/git/pluginAdapter.js
@@ -10,8 +10,14 @@ import {description} from "./render";
 import type {Assets} from "../../app/assets";
 import type {RepoId} from "../../core/repoId";
 import type {Repository} from "./types";
+import type {GitGateway} from "./gitGateway";
 
 export class StaticPluginAdapter implements IStaticPluginAdapter {
+  _gitGateway: GitGateway;
+
+  constructor(gg: GitGateway): void {
+    this._gitGateway = gg;
+  }
   name() {
     return "Git";
   }
@@ -65,16 +71,22 @@ export class StaticPluginAdapter implements IStaticPluginAdapter {
       loadGraph(),
       loadRepository(),
     ]);
-    return new DynamicPluginAdapter(graph, repository);
+    return new DynamicPluginAdapter(this._gitGateway, graph, repository);
   }
 }
 
 class DynamicPluginAdapter implements IDynamicPluginAdapter {
   +_graph: Graph;
   +_repository: Repository;
-  constructor(graph: Graph, repository: Repository) {
+  +_gitGateway: GitGateway;
+  constructor(
+    gitGateway: GitGateway,
+    graph: Graph,
+    repository: Repository
+  ): void {
     this._graph = graph;
     this._repository = repository;
+    this._gitGateway = gitGateway;
   }
   graph() {
     return this._graph;
@@ -83,9 +95,9 @@ class DynamicPluginAdapter implements IDynamicPluginAdapter {
     // This cast is unsound, and might throw at runtime, but won't have
     // silent failures or cause problems down the road.
     const address = N.fromRaw((node: any));
-    return description(address, this._repository);
+    return description(address, this._repository, this._gitGateway);
   }
   static() {
-    return new StaticPluginAdapter();
+    return new StaticPluginAdapter(this._gitGateway);
   }
 }
--- a/src/plugins/git/render.js
+++ b/src/plugins/git/render.js
@@ -1,11 +1,16 @@
 // @flow
 
+import React from "react";
+import Link from "../../app/Link";
 import * as N from "./nodes";
 import type {Repository} from "./types";
+import {type RepoIdString, stringToRepoId} from "../../core/repoId";
+import type {GitGateway} from "./gitGateway";
 
 export function description(
   address: N.StructuredAddress,
-  repository: Repository
+  repository: Repository,
+  gateway: GitGateway
 ) {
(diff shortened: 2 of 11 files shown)


wchargin:
This couples the Git plugin directly to GitHub in the render code.

What do you think about adding:

```js
type URL = string;
export interface GitGateway {
  commit(repo: RepoId, hash: Hash): URL;
}

// src/plugins/git/pluginAdapter.js
export class StaticPluginAdapter implements IStaticPluginAdapter {
  constructor(gateway: GitGateway): void;
  // …
}

// src/app/adapters/defaultPlugins.js
class GithubGateway implements GitGateway {
  commit(repo: RepoId, hash: Hash) {
    return `https://github.com/${repo.owner}/${repo.name}/commit/${hash}`;
  }
}
export function defaultStaticAdapters(): StaticAdapterSet {
  return new StaticAdapterSet([new GithubAdapter(), new GitAdapter()]);
}
```

This way, we avoid locking the Git plugin—which should definitely be
agnostic with respect to the hosting service—into a GitHub backend.

src/plugins/git/pluginAdapter.js
```diff
@@ -10,8 +10,14 @@ import {description} from "./render";
 import type {Assets} from "../../app/assets";
 import type {RepoId} from "../../core/repoId";
 import type {Repository} from "./types";
+import type {GitGateway} from "./gitGateway";
 
 export class StaticPluginAdapter implements IStaticPluginAdapter {
+  _gitGateway: GitGateway;
+
+  constructor(gg: GitGateway) {
```
wchargin: Return `void` here, please, to avoid facebook/flow#6400. Likewise below.

src/plugins/github/githubGitGateway.test.js
```diff
@@ -0,0 +1,17 @@
...
+import {GithubGitGateway} from "./githubGitGateway";
+
+describe("src/plugins/github/githubGitGateway", () => {
+  describe("commitUrl", () => {
+    it("works for a simple example", () => {
+      const repoId = makeRepoId("sourcecred", "example-github");
+      const hash = "0123456adef";
+      const url = new GithubGitGateway().commitUrl(repoId, hash);
+      expect(url).toMatchInlineSnapshot(
+        `"https://github.com/sourcecred/example-github/commit/0123456adef"`
```
wchargin: It would be nice to use an actual commit hash here, so that we can
simply the link to check that it is valid. Say: `ec91adb718a6`.

src/plugins/git/render.js
```diff
@@ -1,24 +1,56 @@
...
-        return hash;
+        return <code>{hash}</code>;
       }
-      const {shortHash, summary} = commit;
-      return `${shortHash}: ${summary}`;
+      const repoIdsForCommit = repository.commitToRepoId[hash];
+      const repoIdStrings: $ReadOnlyArray<RepoIdString> =
+        repoIdsForCommit == null
+          ? []
+          : (Object.keys(repository.commitToRepoId[hash]): any);
```
wchargin: Replace `repository.commitToRepoId[hash]` with `repoIdsForCommit`?

Or replace the whole block with

```js
// This `any`-cast courtesy of facebook/flow#6927.
const repoIdStrings: $ReadOnlyArray<RepoIdString> =
  (Object.keys(repository.commitToRepoId[hash] || {}): any);
```

?

src/plugins/github/githubGitGateway.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import {type RepoId, repoIdToString} from "../../core/repoId";
+import type {Hash} from "../git/types";
+import type {GitGateway, URL} from "../git/gitGateway";
+
+export class GithubGitGateway implements GitGateway {
+  commitUrl(repoId: RepoId, hash: Hash): URL {
+    const repoString = repoIdToString(repoId);
+    return `https://github.com/${repoString}/commit/${hash}`;
```
wchargin: Hmm… `repoIdToString` doesn’t guarantee any particular representation. If I were changing that module, I would feel free to change its representation to `(rid) => stringify(rid)`.

What do you think about using `` `https://github.com/${repoId.owner}/${repoId.name}/commit/${hash}` `` here instead?

src/plugins/git/render.js
```diff
@@ -1,24 +1,58 @@
...
       const hash = address.hash;
       const commit = repository.commits[hash];
       if (commit == null) {
         console.error(`Unable to find data for commit ${hash}`);
-        return hash;
+        return <code>{hash}</code>;
       }
-      const {shortHash, summary} = commit;
-      return `${shortHash}: ${summary}`;
+      const repoIdsForCommit = repository.commitToRepoId[hash];
```
wchargin: travis says remove

PR #888, opened 2018-09-22
mirror: factor out `_makeSingleUpdateFunction`

wchargin:
Summary:
This helpful utility is already used in some test code, and will shortly
be used in main code. @decentralion suggested factoring it out in
<https://github.com/sourcecred/sourcecred/pull/883#discussion_r219647781>.

Test Plan:
Unit tests included, with full coverage; run `yarn unit`.

wchargin-branch: mirror-make-update-helper

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -1,6 +1,6 @@
 // @flow
 
-import type Database from "better-sqlite3";
+import type Database, {BindingDictionary, Statement} from "better-sqlite3";
 import stringify from "json-stable-stringify";
 
 import dedent from "../util/dedent";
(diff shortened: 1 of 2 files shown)


src/graphql/mirror.js
```diff
@@ -886,3 +886,54 @@ function _primitivesTableName(typename: Schema.Typename) {
...
+ * exists---if either existence or uniqueness fails, this method will
+ * raise an error quickly instead of leading to a corrupt state.
+ *
+ * For example, this code...
+ *
+ *     const setName: ({|+userId: string, +newName: string|}) => void =
+ *       _makeSingleUpdateFunction(
+ *         "UPDATE users SET name = :newName WHERE id = :userId"
+ *       );
+ *     setName({userId: "user:foo", newName: "The Magnificent Foo"});
```
decentralion: Good to see a return of the Magnificent Foo Plugin

src/graphql/mirror.js
```diff
@@ -886,3 +886,54 @@ function _primitivesTableName(typename: Schema.Typename) {
...
+ * The prepared statement must not return data (e.g., INSERT and UPDATE
+ * are okay; SELECT is not).
+ *
+ * The statement is not executed inside an additional transaction, so in
+ * the case that the assertion fails, the effects of the statement are
+ * not rolled back by this function.
+ *
+ * This is useful when the statement is like `UPDATE ... WHERE id = ?`
+ * and it is assumed that `id` is a primary key for a record already
+ * exists---if either existence or uniqueness fails, this method will
```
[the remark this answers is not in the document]
wchargin: a: When writing in a plain-text, non–user facing format (primarily code
comments), I generally try to stick to ASCII when reasonable. This
primarily means omitting smart quotes, smart ellipses, and correct
dashes and minus signs. I do this primarily for the benefit of people
who would have a harder time editing the nicer forms, perhaps because
they get confused by a diff of `-` to `−`, or they don’t know how to
input `df…` in Vim, or they’ve SSHed into an old system that doesn’t
have good Unicode support.

(Within ASCII, the choices are usually easy: `---` is the standard
ligature for “—” since TeX days, `...` is obvious for “…”, etc.)

I am not dogmatic about this. For example, our `config/env.js` file
contains the line

```
  // Useful for determining whether we’re running in production mode.
```

which I have not excised, and I certainly don’t correct others on this
regard.

src/graphql/mirror.js
wchargin: I thought that you might enjoy that.

PR #890, opened 2018-09-24
app: add and use a shared Link component

wchargin:
Summary:
This will enable us to style links consistently across our application.

Our previous link colors for base and `:visited` were so similar that I
didn’t actually realize that they were different. In this change, I’ve
kept the same base color, and selected a more contrasting `:visited`
color. I also added an `:active` color, which is good for usability
(color chosen via <http://paletton.com/>’s “triad” suggestion).

Example screenshot, with active, visited, and base links:
![Screenshot as in this commit][img-underline]

I also considered implementing the link underlines with `border-bottom`
instead of `text-decoration` (an oft-touted suggestion that has always
smelled fishy to me, but [the W3 does it][w3], so I guess it’s okay).
That would look like this:

![Screenshot with `border-bottom` underlines][img-border]

…but I did not do so ([rationale in a comment on #890][rationale]).

[w3]: https://www.w3.org/TR/WCAG20-TECHS/F73.html
[img-underline]: https://user-images.githubusercontent.com/4317806/45987381-f154f580-c025-11e8-8b0f-63c1e1ddce02.png
[img-border]: https://user-images.githubusercontent.com/4317806/45926176-a081b780-bed4-11e8-96f2-d0d24d11c8f7.png
[rationale]: https://github.com/sourcecred/sourcecred/pull/890#issuecomment-424146773

We can certainly change these decisions later—that’s one of the purposes
of having this abstraction—so I’m not inclined to bikeshed on them too
much in this commit.

Implementation adapted from:
<https://github.com/wchargin/wchargin.github.io/blob/80263b190e7fb172bd7f022905ad3338ee0af839/src/components/Link.js>

Test Plan:
Check that `<a>` elements and React Router link elements are used only
in `Link` and snapshots:

```
$ git grep --name-only -Fw '<a'
src/app/Link.js
src/assets/logo/discourse_512.png
src/plugins/github/__snapshots__/render.test.js.snap
$ git grep '"react-router"' | grep 'Link'
src/app/Link.js:import {Link as RouterLink} from "react-router";
src/app/Link.test.js:import {Link as RouterLink} from "react-router";
src/app/createRelativeHistory.test.js:import {Router, Route, Link} from "react-router";
src/app/withAssets.test.js:import {IndexRoute, Link, Router, Route} from "react-router";
```

Check that the primary color now appears in just one spot:

```
$ git grep -i 0872a2
src/app/Link.js:    ...colorAttributes("#0872A2"),
```

Then, run `yarn start` and click all the links. Note in particular that
the SVG icons in the header have the correct colors in the active state
as well as the base state.

wchargin-branch: app-link

### DIFF
--- a/src/app/ExternalRedirect.js
+++ b/src/app/ExternalRedirect.js
@@ -2,6 +2,8 @@
 
 import React from "react";
 
+import Link from "./Link";
+
 export default class ExternalRedirect extends React.Component<{|
   +redirectTo: string,
 |}> {
@@ -11,7 +13,7 @@ export default class ExternalRedirect extends React.Component<{|
         <h1>Redirecting…</h1>
         <p>
           Redirecting to:{" "}
-          <a href={this.props.redirectTo}>{this.props.redirectTo}</a>
+          <Link href={this.props.redirectTo}>{this.props.redirectTo}</Link>
         </p>
       </div>
     );
--- a/src/app/HomePage.js
+++ b/src/app/HomePage.js
@@ -1,10 +1,9 @@
 // @flow
 
-import {StyleSheet, css} from "aphrodite/no-important";
-import React, {type Node} from "react";
-import {Link} from "react-router";
+import React from "react";
 
 import type {Assets} from "./assets";
+import Link from "./Link";
 
 export default class HomePage extends React.Component<{|+assets: Assets|}> {
   render() {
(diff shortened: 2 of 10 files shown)


wchargin:
> We have also observed that our link underlines are annoyingly thick.
> A common solution is to use `border-bottom` instead of `text-decoration`
> to effect the underline. This is done [even by W3 itself][w3], which
> makes me comfortable enough to use it.
> 
> [w3]: https://www.w3.org/TR/WCAG20-TECHS/F73.html

You know, I take this back. All of our links look fine in Firefox.
In Chrome, links on the home page at default zoom do indeed look a bit
thick, but this is solved by decreasing the font size from 20px to
anything smaller (i.e., 19px does the job).

Proper underlines, when thin, definitely look better than faux
underlines effected with `border-bottom`. This is also less of a
surprise for other modules that need to know about this style, like
`Page.js`. If no objection, I’ll fix up this patch (later) to use normal
underlines, without changing the font size on the home page.

PR #907, opened 2018-10-02
markdown: ignore references in HTML code elements

### LINKED ISSUE #903: Spurious reference detection inside code blocks
Consider the following pull request: https://github.com/ipfs/js-ipfs-block-service/pull/36

It has numeric references inside code blocks (e.g. `#62`). Per documentation [here] the numeric references should not be considered, as they are in code blocks. However, they do in fact get detected. See image below for an example.

Relevant: https://github.com/sourcecred/sourcecred/issues/432

[here]: https://github.com/sourcecred/sourcecred/blob/163b2c137757a16b2dce7324660e24410cf5540e/src/plugins/github/parseMarkdown.js#L9
image

wchargin:
Summary:
Fixes #903. We already ignore Markdown code syntax (backticks), but
prior to this commit we treated the contents of all HTML elements,
including `<code>`, as normal text. As of this commit, `<code>` elements
are stripped entirely. Other HTML elements, like `<em>`, are unaffected.

Test Plan:
Unit tests added. Also, load data for `ipfs/js-ipfs-block-service`, and
observe in the UI that PR `#36` (Update aegir to version 9.0.0) no
longer has any outward references.

wchargin-branch: markdown-html-code

### DIFF
--- a/src/plugins/github/parseMarkdown.test.js
+++ b/src/plugins/github/parseMarkdown.test.js
@@ -50,6 +50,54 @@ describe("plugins/github/parseMarkdown", () => {
       ];
       expect(textBlocks(input)).toEqual(expected);
     });
+
+    it("includes text inside of non-code HTML elements", () => {
+      const input = "My <strong>#1</strong> pal";
+      expect(textBlocks(input)).toEqual(["My #1 pal"]);
+    });
+
+    it('strips HTML "code" elements', () => {
+      const input = "My <code>#1</code> pal";
+      expect(textBlocks(input)).toEqual(["My  pal"]);
+    });
+
+    it('strips subtrees rooted at HTML "code" elements', () => {
+      const input = "My <code>#1 <strong>*and* #2</strong></code> pals";
+      expect(textBlocks(input)).toEqual(["My  pals"]);
+    });
+
+    it('strips "code" elements within "code" elements', () => {
+      const input = "see <code>#1 and <code>#2</code> okay</code>";
+      expect(textBlocks(input)).toEqual(["see "]);
+    });
+
+    it('handles comments and CDATA within "code" elements', () => {
+      // These are "html_inline" nodes, but are not HTML elements. They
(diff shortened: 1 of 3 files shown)


src/plugins/github/parseMarkdown.test.js
```diff
@@ -50,6 +50,54 @@ describe("plugins/github/parseMarkdown", () => {
...
+    });
+
+    it('strips HTML "pre" blocks and subtrees', () => {
+      // "pre" is not handled specially; all blocks are skipped.
+      const input =
+        "Hello\n\n<pre>some pre-formatted <code>code</code></pre>\n\nworld";
+      expect(textBlocks(input)).toEqual(["Hello", "world"]);
+    });
+
+    it('strips non-"pre" blocks and subtrees', () => {
```
decentralion: This is a little weird, but I'm OK merging this for now.

wchargin: Agreed that it’s a bit weird that we ignore `div`s and `details`s and
such. It bears noting that this is the existing behavior. In particular,
CommonMark treats anything inside a block-level HTML element as raw. For
instance, [`<div>a **b** c</div>`][dingus] renders as a div containing
some literal asterisks and no `strong`s.

So, to make this _not_ be the case, we’d need to recursively seek out
these HTML blocks and re-run Markdown parsing on their contents. We
could consider that as a feature request, but it’s certainly outside the
purview of this commit.

[dingus]: https://spec.commonmark.org/dingus/?text=%3Cdiv%3Ea%20**b**%20c%3C%2Fdiv%3E

decentralion: Note this deviates from GitHub's reference detection.

<div>
**#907**
</div>

Observe that this created a reference to this pull.

wchargin: Yep, this is a good point.

PR #926, opened 2018-10-19
schema: allow annotating primitive field types

wchargin:
Summary:
Prior to this change, primitive fields were un\[i\]typed. This commit
allows adding type annotations. Such annotations do not change the
semantics at all, but we will be able to use them to generate Flow types
corresponding to a schema.

This commit also strengthens the validation on schemata to catch some
errors that would previously have gone unnoticed at schema construction
time: e.g., a node reference to a type that does not exist.

Test Plan:
Unit tests updated, retaining full coverage. The demo script continues
to work when loading `example-github` or `sourcecred`.

wchargin-branch: schema-annotated-primitives

### DIFF
--- a/src/graphql/schema.js
+++ b/src/graphql/schema.js
@@ -32,6 +32,9 @@ export type ObjectId = string;
 //   - Connections are supported as object fields, but arbitrary lists
 //     are not.
 //
+// Primitive and enum fields on an object type may optionally be
+// annotated with their representations.
+//
 // To accommodate schemata where some object types do not have IDs,
 // objects may have "nested" fields of primitive or node-reference type.
 // These may be nested to depth exactly 1. Suppose that `Foo` is an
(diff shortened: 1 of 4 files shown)


src/graphql/schema.js
```diff
@@ -32,6 +32,9 @@ export type ObjectId = string;
 //   - Connections are supported as object fields, but arbitrary lists
 //     are not.
 //
+// Primitive and enum fields on an object type may optionally be
```
decentralion: Why is it optional?

wchargin: Good question.

This was a slightly easier change to make: it means that I didn’t have
to update all the schemata in test code and such.

There’s something to be said for not requiring people to annotate their
schemata if they don’t care about Flow types. But maybe that’s not a
good argument. For instance, if the annotations are required, then the
Mirror module can store some things more efficiently (it wouldn’t have
to JSON-encode values, because we could represent booleans as SQL `0` or
`1` instead of `'true'` and `'false'`, because we could tell at
extraction time what the type of the result should be.)

I’d be happy to make it required in a future commit.

PR #927, opened 2018-10-19
graphql: generate Flow types from a schema

wchargin:
Summary:
Generating Flow types from a structured schema is both straightforward
and terribly useful. This commit implements it.

Test Plan:
Inspect the snapshot for correctness manually. Then, copy it into a new
file, remove backslash-escapes, and verify that it passes Flow.
A subsequent commit will generate types for the actual GitHub schema.

wchargin-branch: graphql-generate-flow-types


src/graphql/generateFlowTypes.js
```diff
@@ -0,0 +1,118 @@
...
+      return field.annotation.elementType;
+    } else {
+      return "null | " + field.annotation.elementType;
+    }
+  }
+  function formatNodeField(field: NodeFieldType): string {
+    return "null | " + field.elementType;
+  }
+  function formatConnectionField(field: ConnectionFieldType): string {
+    return `$ReadOnlyArray<null | ${field.elementType}>`;
```
decentralion: Just checking my understanding: it's <null | elementType> because if the object was deleted or someting, we may still have a connection pointing to it, but won't actually be able to fetch it? (so it will be in the array as null, rather than implicitly filtered out)

wchargin: The real reason that it’s `null | elementType` is that all the
connections provided by GitHub return lists of nullable items:

  - <https://developer.github.com/v4/object/issuecommentconnection/>
  - <https://developer.github.com/v4/object/userconnection/>
  - <https://developer.github.com/v4/object/reactionconnection/>
  - etc.

Therefore, to faithfully match the schema, we have to mark these results
as nullable.

As usual, the GitHub docs do not specify _when_ such a connection might
contain a `null` item. Your explanation may be the case, but I can’t
say. I don’t believe that I’ve ever actually seen a `null` element in an
actual connection result.

Issue #920, opened 2018-10-05
Decide what to do about Commit histories


wchargin:
## Background

When we first started designing what is now the mirror module, we made
the foundational assumption that we treat entities of the same type in
the same way, regardless of how we reached them while traversing the
graph. The user-defined schema specifies what fields we want from each
type of object, and we grab exactly those fields.

But #821 changed our existing continuations system to _not_ have this
property. In particular: the `Commit` object has a `history` connection,
which has nothing special about it from a schema perspective: it is a
connection of commits. But when we query GitHub, we _only_ request the
commit history of the default branch ref’s target commit, not of other
commits that we find.

## Problem

In the current Mirror model, we have two choices: include the connection
for all commits, or include it for no commits.

Including the connection for all commits is a non-starter, because it
would require sending queries quadratic in the number of commits. We
wouldn’t need to fetch each _commit_ a large number of times, but we
would need to exhaustively enumerate all commits’ history connections.
As humans, we can reason about how the commits’ history connections are
not independent—they overlap significantly—but we of course have no way
to express these relationships to the query engine so that it can
“deduplicate” the histories.

Omitting the history connection means that we will not fetch commits
that are not otherwise reachable from the graph. Effectively, this means
that we will only find commits that are _directly reachable_ from the
non-commit subgraph: e.g., one that is the “merged-as” commit of a
pull request or that is the endpoint of a reachable reference event.

## Alternatives

Neither of the above options is attractive. Here are some alternative
proposals.

### Explicit parent traversal

We could omit the history connection from the main schema, but add the
`parents` connection, which lists all of a commit’s direct parents. Say
that a commit is _reachable_ if it is directly reachable or is the
parent of a reachable commit. Then the number of queries required by
this approach is proportional to the longest shortest path to a
reachable commit from the set of directly reachable commits.

This is fine if a project uses pull requests regularly, even if these
only account for a small proportion of the commits. If, for instance, a
project mostly uses direct commits but has a pull request about once
every 10 commits, then after we’ve found all the pull requests, we can
find all the commits in just about 10 extra queries.

In cases where a repository has a fully linear history and does not use
pull requests at all, this requires queries linear in the number of
commits. We fall back to traversing a linked list.

A minor disadvantage of this approach is that the structured form of the
result will be highly nested as opposed to mostly flat. This could be(?)
bad for JSON serialization, but is probably not a big deal. Honestly,
it’s only by chance that the structure isn’t actually circular already.
[...]

PR #928, opened 2018-10-19
github: add GraphQL schema and Flow types

wchargin:
Summary:
The included schema is forked from the one in `graphql/demo.js`.
Primitive types have been added, and the `parents` connection has been
added to commit objects per #920. (We do not include this in the demo
script because without prefetching it would take a long time to load.)

Test Plan:
Unit tests added; run `yarn unit`. Then run `yarn backend` and verify
that `node ./bin/generateGithubGraphqlFlowTypes.js` generates exactly
the same output as in the types file:

```
$ node ./bin/generateGithubGraphqlFlowTypes.js |
> diff -u - ./src/plugins/github/graphqlTypes.js
$ echo $?
0
```

Change the `graphqlTypes.js` file and verify that `yarn unit` fails.

As the build config has been changed, a `yarn test --full` is warranted.
It passes.

Finally, I have manually verified that the schema is consistent with the
documentation at <https://developer.github.com/v4/object/repository/>
and related pages.

wchargin-branch: github-schema-flow-types


wchargin:
IMO, this approach fits better with the Unix approach of filters that
operate on stdin and stdout. For instance, I can verify that the output
file doesn’t contain any JavaScript errors by running

```
node -e "$(node ./bin/generateGithubGraphqlFlowTypes.js | flow-remove-types)"
```

or I can get the list of all exported types with

```
$ node ./bin/generateGithubGraphqlFlowTypes.js |
> grep '^export type' | awk '{print $3}'
Actor
Blob
Bot
[snip]
```

or I can diff the result against the current file without updating it
by running

```
$ node ./bin/generateGithubGraphqlFlowTypes.js |
> diff -u - src/plugins/github/graphqlTypes.js 
```

In other words, saving the output to a fixed filepath is just one
potential application of the script.

Issue #923, opened 2018-10-18
Integrate the Mirror module into the GitHub plugin


wchargin:
Per #622, we now have a way to fetch a GitHub repository incrementally
and cache the results in a local SQLite database. The next step is to
actually use this functionality in the GitHub plugin.

A few things need to happen.

The data format ingested by the relational view’s `addData` is not the
same as the data format returned by the Mirror module’s `extract`. The
former has GraphQL artifacts like `pageInfo` and `nodes` fields, whereas
the latter is just [a structured representation of the object graph][1].
We should write a module to transform the former to the latter, and then
change the relational view to ingest the new format (but still from the
old source for now).

[1]: https://github.com/sourcecred/sourcecred/blob/7c5923959e6d7a63de903bbdb030a7e34017f517/src/graphql/mirror.js#L1424-L1470

Additionally, as described in #920, we need to change the way that we
represent commit histories. Currently, we fetch a linearized history of
commits at top level, and also each pull request’s merged-as commit. In
the new world, we’ll fetch all _reachable_ commits, where a commit is
reachable if it is the HEAD commit, if it is the merged-as commit of a
pull request, or if it is a parent of a reachable commit.

This is primarily a difference of representation, but it is semantically
different if a pull request’s merged-as commit is not Git-reachable from
the head of the repository (say, if there was a force-push, or if a PR
was merged into a branch that was never merged into the default branch).
In those cases, the new semantics are more correct—the old semantics
would silently be missing pieces of the history.

To account for this in the transition period, we’ll update the existing
GitHub queries to fetch parent OIDs for each commit that we currently
fetch. Then, the translation module described above will dereference
commits appropriately. If during the translation process we find a
commit OID reference for which we do not have the commit, we can print
an error and a dummy commit. This should only happen in exceptional
cases (force-push, merge to non-default branch) and only until we switch
over to the Mirror API, at which point the problem simply goes away.

wchargin:
I’ve decided not to do prefetching:
<https://github.com/sourcecred/sourcecred/issues/920#issuecomment-433732215>

Thus, this issue is closed by #937.

PR #934, opened 2018-10-22
relationalView: use new data format

wchargin:
Summary:
This makes significant progress toward #923. As of this commit, it is
possible to use the Mirror module for the whole loading pipeline. This
process may be slow for repositories that do not use pull requests at
all (more precisely, that have large connected commit subgraphs none of
whose nodes is the merge commit of a pull request; see #920 for details)
so it is not yet the default codepath.

Test Plan:
Existing unit tests should suffice. For extra testing, I’ve added a
script that fetches a repository both via the old continuations logic
and the new Mirror logic, then constructs relational views and checks
whether the data is the same. For `example-github`, the views are
identical. For `sourcecred`, they are not: the old continuations logic
erroneously omits two commits, which the Mirror logic includes.

You can run the test like this:

```
$ node ./bin/testContinuations.js \
> sourcecred sourcecred MDEwOlJlcG9zaXRvcnkxMjAxNDU1NzA= \
> /tmp/continuations.json /tmp/mirror.json \
> 2> >(jq . >&2)
{
  "child": "0d38dde23a6de831315f3643a7d2bc15e8df7678",
  "parent": "cb8ba0eaa1abc1f921e7165bb19e29b40723ce65",
  "type": "UNKNOWN_PARENT_OID"
}
{
  "child": "d152f48ce4c2ed1d046bf6ed4f139e7e393ea660",
  "parent": "de7a8723963d9cd0437ef34f5942a071b850c0e7",
  "type": "UNKNOWN_PARENT_OID"
}
Different. Saving to disk...
```

Use `diff -u <(jq . /tmp/continuations.json) <(jq . /tmp/mirror.json)`
to inspect the differences, and note that exactly the two missing
commits have been added and that there are no other changes. (The diff
is small: just 51 lines of nicely formatted JSON.) The full log is here:
<https://gist.github.com/wchargin/e159cac9dcf3cc3b1efbd54f59e24e0b>

I also generated the `sourcecred/sourcecred` cred attribution and viewed
it with `yarn start`, which seems to work fine.

wchargin-branch: relationalview-new-data-format

### DIFF
--- a/config/paths.js
+++ b/config/paths.js
@@ -29,6 +29,9 @@ module.exports = {
   backendEntryPoints: {
     sourcecred: resolveApp("src/cli/main.js"),
     //
+    testContinuations: resolveApp(
+      "src/plugins/github/bin/testContinuations.js"
+    ),
     generateGithubGraphqlFlowTypes: resolveApp(
       "src/plugins/github/bin/generateGraphqlFlowTypes.js"
     ),
--- a/src/plugins/github/bin/testContinuations.js
+++ b/src/plugins/github/bin/testContinuations.js
@@ -1,0 +1,106 @@
+// @flow
+// Ad hoc testing script for RelationalView input format consistency.
+
+import Database from "better-sqlite3";
+import fs from "fs-extra";
+import stringify from "json-stable-stringify";
+import deepEqual from "lodash.isequal";
+
+import {makeRepoId} from "../../../core/repoId";
+import {Mirror} from "../../../graphql/mirror";
+import fetchGithubRepo, {postQuery} from "../fetchGithubRepo";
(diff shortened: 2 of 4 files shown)


config/paths.js
```diff
@@ -29,6 +29,9 @@ module.exports = {
   backendEntryPoints: {
     sourcecred: resolveApp("src/cli/main.js"),
     //
```
decentralion: I know it's not added in this commit, but: is there any reason for the empty comment here? If not, maybe we can  fly-by remove it? (approved either way)

wchargin: When we had an OClif CLI, it separated the OClif commands from the other
various entry points. I’m fine with removing it.

Issue #704, opened 2018-08-24
Make an API for accessing cred


decentralion:
We should make a user's cred programatically accessible. For the time being, we may just want to make an API for getting the full score/cred distribution for a project available as a JSON blob. This combos well with #703 as we need to run PageRank at load time for both.

cc: @karthik @mikeal

decentralion:
This is conclusively fixed as of #1223! :tada:

Issue #703, opened 2018-08-23
Server-side render the cred attribution


decentralion:
Right now, we don't run PageRank on the backend, or provide pre-rendered cred results on the cred explorer load. This means that users need to run PageRank on their own machines before seeing any results. 

As of 0.0.2, when projects are publicly posting their cred attributions, this will be a bad user experience - visitors should be able to see the cred, according to the project's canonical weights, immediately without needing to run PageRank themselves. This will reduce user confusion, and also be better for people with slow CPUs or visiting on mobile. 

To improve this, we should SSR the cred explorer results. Power users who want to tweak the weights and re-run will still be able to do so. This change will have some implications around e.g. storing weights in localStore, and might make the power user experience slightly less convenient, but I think it's actually a good thing. Storing weight tweaks in localStore makes the experience less reproducible and means that different people may see different cred without realizing why.

decentralion:
Here's a proposal for what gets included with the server side render:

When the page loads, it comes bundled with every user's cred, and every connection that contributed to every user's cred. Then, when the page loads, the UI works normally as you explore where each user got their cred from. However, once you get down to the individual connection level (e.g. REFERENCES pull 703), you can't expand down to the node until you load the data. 

This way, on page load, you immediately have a workable UI that covers ~90% of regular behavior, i.e. checking out the leaderboard and doing some basic exploration. Once you want to get super far into it, then you need to load the whole graph.

We could consider doing the same thing for issues and pulls (or for top 100 issues and pulls).

Beanow:
The current solution in TimelineCred is to run Timeline PageRank ahead of time and read summarized data in the frontend. Needing only the recalculate when changing parameters.

I think this covers the use-case for this issue. Suggest picking up further improvements when we rework the UI again.

PR #942, opened 2018-10-29
Add skeleton of `sourcecred analyze`

decentralion:
The `analyze` command is the first step towards #704 and #703. When
fully implemented, it will run PageRank for a loaded repository,
generating a complete graph and cred attribution.

For now, this just adds a scaffold. It does basic argument parsing, and
has help text, but the actual command is not yet implemented.

Test plan:
Unit tests verify that the analyze command is hooked into `sourcecred`
and `sourcecred help`, and that it responds to the `--help` command and
parses its arguments appropriately.


src/cli/help.js
```diff
@@ -32,6 +34,7 @@ function usage(print: (string) => void): void {
 
     Commands:
       load      load repository data into SourceCred
+      analyze   analyze cred for a repository
```
wchargin: for an already-loaded repository?

PR #947, opened 2018-10-29
Factor out plugin declarations

decentralion:
The plugin adapters are specific to `app/` and have logic for fetching
data from the backend, producing React nodes for descriptions, et
cetera.

However, they also have information that is generic to the plugin
itself: its name, its node/edge prefixes, and its types.

This method factors out the generic info into a `PluginDeclaration`,
which is a type (rather than an interface). Then, the plugin adapter has
a `declaration` method which returns the declaration.

Current users of the plugin adapters get additional mechanical
complexity because they need to call `.declaration().property` rather
than `.property()`, but this is not too significant.

The main benefit is that #704 is unblocked, as the cli `analyze` command
will be able to get necessary information from the declaration. As an
added benefit, the organization of plugin code is somewhat improved.

Test plan: `yarn test` sufficies, but I also ran `yarn start` and
checked the UI behavior to be safe.

### DIFF
--- a/src/analysis/pluginDeclaration.js
+++ b/src/analysis/pluginDeclaration.js
@@ -1,0 +1,12 @@
+// @flow
+
+import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
+import type {EdgeType, NodeType} from "./types";
+
+export type PluginDeclaration = {|
+  +name: string,
+  +nodePrefix: NodeAddressT,
+  +edgePrefix: EdgeAddressT,
+  +nodeTypes: $ReadOnlyArray<NodeType>,
+  +edgeTypes: $ReadOnlyArray<EdgeType>,
+|};
--- a/src/plugins/git/declaration.js
+++ b/src/plugins/git/declaration.js
@@ -1,0 +1,31 @@
+// @flow
+
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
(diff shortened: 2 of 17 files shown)


src/analysis/pluginDeclaration.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
+import type {EdgeType, NodeType} from "./types";
+
+export type PluginDeclaration = {|
+  +name: string,
+  +nodePrefix: NodeAddressT,
+  +edgePrefix: EdgeAddressT,
+  +nodeTypes: NodeType[],
```
wchargin: Now that this is a data type, it should provide `$ReadOnlyArray`s.

src/plugins/git/declaration.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import * as N from "./nodes";
+import * as E from "./edges";
+
+const CommitNodeType = Object.freeze({
```
wchargin: `commitNodeType` or `COMMIT_NODE_TYPE`, please.

wchargin:
> but this is not too significant.

strong +1

PR #958, opened 2018-10-31
test: run Sharness verbosely on nightly jobs

wchargin:
Summary:
Tests that run only on nightly builds (`yarn test --full`) and fail only
on CI (not locally) are a bit more inconvenient to debug when they fail.
This patch makes the `yarn test --full` script print all the
intermediate output in Sharness tests, which can be helpful. We don’t do
this for `yarn test` simply because it generates a ton of spam even on
successful tests.

Test Plan:

    $ yarn test --full 2>&1 | wc -l
    1173

wchargin-branch: test-full-verbose

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "test": "node ./config/test.js",
     "unit": "BABEL_ENV=test NODE_ENV=test jest --env=jsdom",
     "sharness": "make -sC ./sharness prove PROVE_OPTS=-f TEST_OPTS='--chain-lint'",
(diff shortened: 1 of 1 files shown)


wchargin:
@chriscool, do you know if there is an incantation that we can use to
make Sharness print verbose output only on failures? Currently, we use

    make -sC ./sharness prove PROVE_OPTS=-f TEST_OPTS='--chain-lint'

but this prints a lot of log output even on successful tests (about 1000
lines for us), which is annoying.

PR #965, opened 2018-10-31
Add the demo plugin

decentralion:
This pull request contains two commits. The first commit factors the demo adapters into their own demo plugin, and the second commit deletes the demo adapters and re-routes all the references.

Should be merged via rebase-and-merge.


src/plugins/demo/analysisAdapter.js
```diff
@@ -0,0 +1,23 @@
...
+import {Graph} from "../../core/graph";
+import type {RepoId} from "../../core/repoId";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {IAnalysisAdapter} from "../../analysis/analysisAdapter";
+
+import {declaration} from "./declaration";
+import {graph} from "./graph";
+
+export class AnalysisAdapter implements IAnalysisAdapter {
+  loadingMock: Function;
```
wchargin: Can we type this as

```js
  loadingMock: (assets: Assets, repoId: RepoId) => Promise<mixed>;
``

instead? The `Function` type is deliberately unsound (basically `any`)
and should be avoided.

You can apply the following patch to fix the resulting type error in the test code:

```diff
diff --git a/src/app/adapters/adapterSet.test.js b/src/app/adapters/adapterSet.test.js
index af7c2a78..f0e3caf9 100644
--- a/src/app/adapters/adapterSet.test.js
+++ b/src/app/adapters/adapterSet.test.js
@@ -103,8 +103,7 @@ describe("app/adapters/adapterSet", () => {
     });
     it("loads a dynamicAdapterSet", async () => {
       const {x, sas} = example();
-      x.loadingMock = jest.fn();
-      x.loadingMock.mockResolvedValue();
+      x.loadingMock = jest.fn().mockResolvedValue();
       expect(x.loadingMock).toHaveBeenCalledTimes(0);
       const assets = new Assets("/my/gateway/");
       const repoId = makeRepoId("foo", "bar");
```

src/plugins/demo/appAdapter.js
```diff
@@ -0,0 +1,50 @@
...
+    return Promise.resolve(result);
+  }
+}
+
+export class FactorioDynamicAdapter implements DynamicAppAdapter {
+  graph() {
+    return graph();
+  }
+  nodeDescription(x: NodeAddressT) {
+    return NodeAddress.toString(x);
```
wchargin: Can this prepend `"[Factorio] "` or something, so that we can tell in
test code where this is being called versus a `toString` fallback?

src/plugins/demo/README.md
```diff
@@ -0,0 +1,10 @@
+# The Demo Plugin
+
+This plugin exists for testing convenience. It's nice sometimes to have a super
```
wchargin: Is this plugin also supposed to be a structural model for real plugins?

If not, can we call that out explicitly?

src/plugins/demo/declaration.js
```diff
@@ -0,0 +1,45 @@
...
+  defaultForwardWeight: 1,
+  backwardName: "is transported by",
+  defaultBackwardWeight: 2 ** -1,
+  prefix: EdgeAddress.fromParts(["factorio", "transports"]),
+});
+
+export const declaration: PluginDeclaration = Object.freeze({
+  name: "Factorio demo adapter",
+  nodePrefix: NodeAddress.fromParts(["factorio"]),
+  nodeTypes: [inserterNodeType, machineNodeType],
```
wchargin: Please freeze these objects.

src/plugins/demo/analysisAdapter.js
decentralion: Yes, for consistency with the app adapter.

src/plugins/demo/README.md
decentralion: Great point. Updated README.

src/plugins/demo/analysisAdapter.js
wchargin: Okay, makes sense.

Issue #967, opened 2018-10-31
Refactor so that PageRank can run on the backend


decentralion:
Right now a lot of the essential logic for running PageRank--namely generating the weights and turning the weights into an EdgeEvaluator--lives in `app/credExplorer/weights`. This means that the logic for running PageRank is commingled with the frontend code, which blocks #704 and (as a consequence) #935 and #703.

This issue is tracking work to move all of the essential logic for computing PageRank out of `app` and into the new `analysis` module. 

Work so far:
#944, #946, #947, #948, #949, #950, #965, #966

decentralion:
This was completed a long time ago.

PR #968, opened 2018-11-01
Declare the fallback plugin in analysis

decentralion:
Now that the `analysis` module owns the Node and Edge types, it should
own the "fallback plugin" too. (Note that it's not actually a plugin,
though it somewhat acts like one.)

We now declare the fallback type in `analysis`, along with a fallback
analysis adapter. `app/adapters` then declares a fallback app adapter.

Test plan: `yarn test`

Progress towards #967.

### DIFF
--- a/src/analysis/fallbackDeclaration.js
+++ b/src/analysis/fallbackDeclaration.js
@@ -1,0 +1,32 @@
+// @flow
+
+import {NodeAddress, EdgeAddress} from "../core/graph";
+
+import type {PluginDeclaration} from "./pluginDeclaration";
+
+import type {EdgeType, NodeType} from "./types";
+
+export const FALLBACK_NAME = "FALLBACK_ADAPTER";
+
+export const fallbackNodeType: NodeType = Object.freeze({
+  name: "node",
+  pluralName: "nodes",
+  prefix: NodeAddress.empty,
+  defaultWeight: 1,
+});
+
+export const fallbackEdgeType: EdgeType = Object.freeze({
+  forwardName: "points to",
+  backwardName: "is pointed to by",
+  defaultForwardWeight: 1,
+  defaultBackwardWeight: 1,
+  prefix: EdgeAddress.empty,
+});
+
+export const fallbackDeclaration: PluginDeclaration = Object.freeze({
+  name: FALLBACK_NAME,
+  nodePrefix: NodeAddress.empty,
+  edgePrefix: EdgeAddress.empty,
+  nodeTypes: Object.freeze([fallbackNodeType]),
+  edgeTypes: Object.freeze([fallbackEdgeType]),
+});
--- a/src/analysis/fallbackAdapter.js
+++ b/src/analysis/fallbackAdapter.js
@@ -1,0 +1,21 @@
+// @flow
+
+import {Graph} from "../core/graph";
+import type {RepoId} from "../core/repoId";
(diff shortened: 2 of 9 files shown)


wchargin:
The fallback plugin is very different from the demo plugin. Core code
depends on it by name. The core code may not depend on `plugins/`.

It sounds like this should go in `analysis/`?

decentralion:
PTAL. I rewrote it as suggested.

src/analysis/fallbackAnalysisAdapter.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+import {Graph} from "../core/graph";
+import type {RepoId} from "../core/repoId";
+import type {PluginDeclaration} from "./pluginDeclaration";
+import type {IAnalysisAdapter} from "./analysisAdapter";
+
+import {fallbackDeclaration} from "./fallbackDeclaration";
+
+export class AnalysisAdapter implements IAnalysisAdapter {
```
wchargin: The name `FallbackAnalysisAdapter` (maybe dropping the `Analysis`) makes more sense to me. Thoughts?

src/analysis/fallbackAnalysisAdapter.js
```diff
@@ -0,0 +1,24 @@
...
+  loadingMock: (sourcecredDirectory: string, repoId: RepoId) => Promise<mixed>;
```
wchargin: This adapter should not have a loading mock; it is used in production.

src/analysis/fallbackDeclaration.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const fallbackNodeType: NodeType = Object.freeze({
+  name: "node",
+  pluralName: "nodes",
+  prefix: NodeAddress.empty,
+  defaultWeight: 1,
+});
+
+export const fallbackEdgeType: EdgeType = Object.freeze({
+  forwardName: "forward edge",
```
wchargin: Aren't these supposed to be verb phrases?

What do you think of "points to"/"is pointed to by"?

src/app/adapters/fallbackAppAdapter.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+export class FallbackDynamicAdapter implements DynamicAppAdapter {
+  graph() {
+    return new Graph();
+  }
+
+  nodeDescription(x: NodeAddressT) {
+    return `[fallback]: ${NodeAddress.toString(x)}`;
```
wchargin: We probably don't want the `[fallback]: ` here, right? as this can appear in production?

decentralion: Fallback *shouldn't* appear in production. If it does, having the string [fallback] making it clear that the fallback plugin is present is probably a feature. It makes it easy to grep for the string and find why it appears.

src/analysis/fallbackAnalysisAdapter.js
decentralion: Done.

src/analysis/fallbackDeclaration.js
decentralion: Done.

src/app/adapters/fallbackAppAdapter.js
wchargin: Okay, I'm persuaded.

PR #974, opened 2018-11-01
Rename `app/` to `homepage/`

decentralion:
Now that we've moved the explorer out of app, it is more concisely
described as the homepage.

Test plan: Rename only. Run `yarn test`.


decentralion:
Note: as of this commit, `version.js` lives under `homepage/`. I don't think that's the right location, and plan to move it in a subsequent commit.

wchargin:
@decentralion This breaks `yarn test --full`; can you please fix?

PR #975, opened 2018-11-01
Fix a build failure induced by #974

decentralion:
The referenced pull request mistakenly didn't update `config/paths.js`,
which caused build_static_site to fail in `yarn test full`.

Test plan: `yarn test --full` now passes.


wchargin:
`yarn test --full`?

decentralion:
That's right. Fixed.

PR #981, opened 2018-11-01
webpack: expose repo registry at build time

decentralion:
Summary:
We want to remove the repository selector dropdown on the cred explorer
homepage and instead render a separate web page for each project. To do
this, we need to know which pages to render statically. We choose to
ingest this information from the state of the repository registry at
build time.

This commit adds an environment variable `REPO_REGISTRY` whose contents
are the stringified version of the repository registry, or `null` if
SourceCred has been built for the backend. This variable is defined with
Webpack’s `DefinePlugin`, so any code bundled by Webpack can refer to it
via `process.env.REPO_REGISTRY` both on the server and in the browser.

Paired with @wchargin.

Test Plan:
Sharness tests modified; running `yarn test --full` suffices.

### DIFF
--- a/config/webpack.config.web.js
+++ b/config/webpack.config.web.js
@@ -5,7 +5,9 @@ import type {
   $Application as ExpressApp,
   $Response as ExpressResponse,
 } from "express";
+import type {RepoIdRegistry} from "../src/explorer/repoIdRegistry";
 */
+const fs = require("fs");
 const os = require("os");
 const path = require("path");
 const webpack = require("webpack");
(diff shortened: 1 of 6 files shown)


config/webpack.config.web.js
```diff
@@ -19,8 +21,39 @@ const getClientEnvironment = require("./env");
...
 // Source maps are resource heavy and can cause out of memory issue for large source files.
 const shouldUseSourceMap = process.env.GENERATE_SOURCEMAP !== "false";
+
+function loadRepoRegistry() /*: RepoIdRegistry */ {
+  const env = process.env.SOURCECRED_DIRECTORY;
+  // TODO(#945): de-duplicate finding the directory with src/cli/common.js
+  const defaultDirectory = path.join(os.tmpdir(), "sourcecred");
+  const scDirectory = env != null ? env : defaultDirectory;
+  // TODO(@dandelion): Remove hacks around compat usage here
+  // TODO(@dandelion): Import rather than hardcode the registry file name
```
wchargin: Yep. For posterity: this is all a bit hacky because most of our modules
are written with ES6 syntax, but our Webpack config needs to run in
vanilla Node, so we’re inlining things left and right.

Issue #943, opened 2018-10-29
Optimize PagerankNodeDecomposition


decentralion:
@wchargin and I were doing some perf explorations and found that PagerankNodeDecomposition/decompose is quite slow (~5.7s when run on mrdoob/three.js, out of ~30s total time).

It looks like the issue is the JSON.stringify here: https://github.com/sourcecred/sourcecred/blob/d3e79e3c4e65f562c64866bbc33de9eedaf62544/src/core/attribution/pagerankNodeDecomposition.js#L50

We should optimize away the need for that JSON.stringify and post before/after performance analysis.

PR #995, opened 2018-11-02
Implement to/fromJSON for PagerankNodeDecomp

decentralion:
For #967, we need to actually serialize the results from running
PageRank so that others can consume it. This commit adds toJSON and
fromJSON methods for PagerankNodeDecomposition so that we can do just
that.

Test plan: Unit tests to verify round trip stability of the JSON have
been added. Run `yarn test`.

### DIFF
--- a/src/analysis/pagerankNodeDecomposition.js
+++ b/src/analysis/pagerankNodeDecomposition.js
@@ -11,6 +11,7 @@ import {
 import type {NodeScore} from "./nodeScore";
 import * as MapUtil from "../util/map";
 import * as NullUtil from "../util/null";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
 
 export type ScoredConnection = {|
   +connection: Connection,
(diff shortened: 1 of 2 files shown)


decentralion:
While testing this pull request, I realized that serializing the pagerankNodeDecomposition is a bad idea. The reason is that the pND is a huge expansion over the underlying data that it describes (specifically, its size is potentially quadratic in the size of the scored graph). Therefore, serializing this data structure makes very little sense; actually, even precomputing the whole thing is a bit silly, since we could be computing the local sections of it that we need on-the-fly (thus reducing the 5sec load penalty induced by computing the whole pND as described in #943).

decentralion:
Oops, I think you're right: it's linear in the number of edges.

In practice though, it's really quite big. (data for sourcecred/sourcecred)
```
464K	git
260K	github
2.7M	graph.json
260K	graph.json.gz
18M	pagerankNodeDecomposition.json
904K	pagerankNodeDecomposition.json.gz
```

wchargin:
Would love to see data on a larger repo, like twbs/bootstrap. Should be easy to generate that if you still have the Mirror cache laying around (if not, I can send you one).

PR #1005, opened 2018-11-15
Update WeightSlider.js to allow 0 weights

ianjdarrow:
We changed the weight slider so the minimum value maps to zero instead
of a power of two. Also changed declarations to default at zero when
appropriate to keep UI state consistent.

test plan: Tested UI, confirmed ability to move weights to 0, added unit
tests

### DIFF
--- a/src/explorer/weights/WeightSlider.js
+++ b/src/explorer/weights/WeightSlider.js
@@ -2,6 +2,9 @@
 
 import React from "react";
 
+export const WEIGHT_SLIDER_MIN = -5;
+export const WEIGHT_SLIDER_MAX = 5;
+
 export type Props = {|
   +weight: number,
   +name: React$Node,
@@ -14,13 +17,15 @@ export class WeightSlider extends React.Component<Props> {
         <span style={{flexGrow: 1}}>{this.props.name}</span>
         <input
           type="range"
-          min={-5}
-          max={5}
+          min={WEIGHT_SLIDER_MIN}
+          max={WEIGHT_SLIDER_MAX}
           step={1}
-          value={Math.log2(this.props.weight)}
+          value={convertWeightToSliderValue(this.props.weight)}
           onChange={(e) => {
-            const logValue = e.target.valueAsNumber;
-            this.props.onChange(2 ** logValue);
+            const sliderValue = convertSliderValueToWeight(
+              e.target.valueAsNumber
+            );
+            this.props.onChange(sliderValue);
           }}
         />{" "}
         <span
@@ -33,11 +38,19 @@ export class WeightSlider extends React.Component<Props> {
   }
 }
 
+function convertWeightToSliderValue(weight: number): number {
+  return weight === 0 ? WEIGHT_SLIDER_MIN : Math.log2(weight);
+}
+
+function convertSliderValueToWeight(sliderValue: number): number {
+  return sliderValue === WEIGHT_SLIDER_MIN ? 0 : 2 ** sliderValue;
+}
+
 export function formatWeight(n: number) {
-  if (n <= 0 || !isFinite(n)) {
+  if (n < 0 || !isFinite(n)) {
     throw new Error(`Invalid weight: ${n}`);
   }
-  if (n >= 1) {
+  if (n >= 1 || n === 0) {
     return n.toFixed(0) + "×";
   } else {
     return `1/${(1 / n).toFixed(0)}×`;
--- a/src/explorer/weights/WeightSlider.js
+++ b/src/explorer/weights/WeightSlider.js
@@ -2,14 +2,47 @@
 
 import React from "react";
 
-export const WEIGHT_SLIDER_MIN = -5;
-export const WEIGHT_SLIDER_MAX = 5;
+/**
+ * The Weight is the number that the WeightSlider actually provides to the
+ * instantiating component. The Weight varies in the range
+ * [0, 2**MAX_SLIDER_VALUE].
+ */
+export type Weight = number;
+
+/**
+ * The SliderPosition is the current position of the [range input] that the
+ * WeightSlider instantiates. It varies in the range [MIN_SLIDER, MAX_SLIDER].
+ *
+ * [range input]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range
+ *
+ * The returned Weight is exponential in the SliderPosition, so that e.g. if
+ * the SliderPosition is 2 then the Weight is 4, if the SliderPosition is 3
+ * then the Weight is 8, and so forth. The exception is when the
+ * SliderPosition==MIN_SLIDER_VALUE. In this case, the Weight will be 0.
+ */
+export type SliderPosition = number;
+
+export const MIN_SLIDER: SliderPosition = -5;
+export const MAX_SLIDER: SliderPosition = 5;
 
 export type Props = {|
-  +weight: number,
+  // The current Weight the WeightSlider is set to
+  +weight: Weight,
+  // The name associated with the slider.
+  // Will be displayed adjacent to the slider.
   +name: React$Node,
-  +onChange: (number) => void,
+  // onChange handler so the parent can recieve (and then set) the new weight
+  +onChange: (Weight) => void,
 |};
+
+/**
+ * The WeightSlider is a React component for letting the user input
+ * a numeric weight that varies over an exponential range.
+ * The WeightSlider instantiates a slider that lets the user change
+ * the weight, displays the current weight value, and displays
+ * a "name" for the slider (which may be an arbitrary inline React
+ * component).
+ */
 export class WeightSlider extends React.Component<Props> {
   render() {
     return (
@@ -17,15 +50,13 @@ export class WeightSlider extends React.Component<Props> {
         <span style={{flexGrow: 1}}>{this.props.name}</span>
         <input
           type="range"
-          min={WEIGHT_SLIDER_MIN}
-          max={WEIGHT_SLIDER_MAX}
+          min={MIN_SLIDER}
+          max={MAX_SLIDER}
           step={1}
-          value={convertWeightToSliderValue(this.props.weight)}
+          value={weightToSlider(this.props.weight)}
           onChange={(e) => {
-            const sliderValue = convertSliderValueToWeight(
-              e.target.valueAsNumber
-            );
-            this.props.onChange(sliderValue);
+            const weight: Weight = sliderToWeight(e.target.valueAsNumber);
+            this.props.onChange(weight);
           }}
         />{" "}
         <span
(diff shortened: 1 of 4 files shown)


src/explorer/weights/WeightSlider.js
```diff
@@ -33,11 +38,19 @@ export class WeightSlider extends React.Component<Props> {
   }
 }
 
+function convertWeightToSliderValue(weight: number): number {
+  return weight === 0 ? WEIGHT_SLIDER_MIN : Math.log2(weight);
```
wchargin: It’s confusing that `convertWeightToSliderValue` can return values that
are smaller than `WEIGHT_SLIDER_MIN`, and also (relatedly) that it’s
discontinuous. How about the following instead:

```js
  return Math.max(WEIGHT_SLIDER_MIN, Math.log2(weight));
```

When `weight` is `0`, its `log2` will be `-Infinity`, so everything
works out.

src/explorer/weights/WeightSlider.js
```diff
@@ -33,11 +38,19 @@ export class WeightSlider extends React.Component<Props> {
...
+}
+
+function convertSliderValueToWeight(sliderValue: number): number {
+  return sliderValue === WEIGHT_SLIDER_MIN ? 0 : 2 ** sliderValue;
```
wchargin: Could you please add docs to this function to clarify its contract?

  - Are clients allowed to call it with values outside the range set by
    `WEIGHT_SLIDER_MIN` and `WEIGHT_SLIDER_MAX`?
  - If this is disallowed, what should happen if a client does this?
    Should you throw an error? Return `NaN`? Is the behavior undefined?

Currently, the function is discontinuous at `WEIGHT_SLIDER_MIN`, which
is a bit strange. Depending on what you want the semantics to be, you
might s/===/<=/ or you might add some error checks.

src/plugins/github/declaration.js
```diff
@@ -99,35 +99,31 @@ const mentionsAuthorEdgeType = Object.freeze({
   forwardName: "mentions author of",
   backwardName: "has author mentioned by",
   defaultForwardWeight: 1,
-  // TODO(#811): Probably change this to 0
-  defaultBackwardWeight: 1 / 32,
+  defaultBackwardWeight: 0,
```
wchargin: Please put these in a separate change.

src/explorer/weights/WeightSlider.js
wchargin: Is it legal to call `convertWeightToSliderValue` with a negative value?
Currently, you return `NaN`.

Is it legal to call `convertWeightToSliderValue` with a value higher
than `2 ** WEIGHT_SLIDER_MAX`? Currently, you return a value outside of
the implied range.

src/explorer/weights/WeightSlider.test.js
```diff
@@ -3,7 +3,7 @@
 import React from "react";
 import {shallow} from "enzyme";
 
-import {WeightSlider, formatWeight} from "./WeightSlider";
+import {WeightSlider, formatWeight, WEIGHT_SLIDER_MIN} from "./WeightSlider";
```
wchargin: Usually, when we have tests converting among two representations, we
like to see “roundtrip tests”: that converting in one direction and then
the other gets you the original value. In this case, this isn’t exactly
what we want, because one of the directions is _supposed_ to be lossy.
But converting from slider-value to weight-value and back to
slider-value should always be the identity operation. It would be nice
to test that on some interesting values.

wchargin:
Welcome! Glad to see this change.

Tests look good—thanks! Left some inline comments.

Your commits don’t seem to be properly associating with your GitHub
account, because your local Git client has your email address set to
`iandarrow@Ians-MacBook-Pro.local`. Please [configure your local Git
client][git]. You don’t have to fix these commits, because they’ll all
be squashed under your GitHub identity, but if you want to rebase them
to update the authorship you can.

I note that setting a node weight to zero can now mean that a node
actually has 0 cred, because even its synthetic loop becomes
zero-weighted. (This yields some “NaN%” in the UI.) The synthetic loop
should probably not be subject to node weights, so I’d be happy to see a
change to that effect in a separate PR either before or after this one.

[git]: https://help.github.com/articles/setting-your-commit-email-address-in-git/

src/plugins/github/declaration.js
decentralion: Note: if they are in a separate change, then the UI is briefly in an inconsistent state, because the sliders start at 1/32, (with a reasonable-looking visual position), but once the slider is moved, it can never return to 1/32. I suppose it's not too objectionable, but that was why I suggested just putting the weight value change in this commit.

wchargin: Ah, good point. Either way is fine with me, then.

decentralion:
Just to follow on to @wchargin's comment about getting the git email address right: the incentive for doing this is that SourceCred gives cred for commits based on the email field, so it does matter. If you want to keep your email private, a good choice is ianjdarrow@users.noreply.github.com. (GitHub will map this onto your user account, and the way we collect commit authorship at present is by asking GitHub :wink:)

Also: since we pair programmed on this commit, you may want to add `Paired with @decentralion` to the bottom of the pull request message and commit message, in which case, in the benevolent eye of SourceCred, we will split the cred for this pull request. But it's up to you. :smile:

src/explorer/weights/WeightSlider.js
decentralion: Took this suggestion in my commit, although I think it's now unnecessary because I explicitly check the domain.

src/explorer/weights/WeightSlider.js
decentralion: Added docs, and also added error handling.

src/explorer/weights/WeightSlider.js
decentralion: Added error handling and made all out-of-bounds inputs throw errors.

src/explorer/weights/WeightSlider.test.js
decentralion: Added round-trip tests.

decentralion:
Since it's been stale for quite some time, I'm assuming ownership of this PR so it can make it into the codebase. @wchargin, please re-review when you get the chance.

src/explorer/weights/WeightSlider.js
```diff
@@ -2,30 +2,61 @@
...
 import React from "react";
 
-export const WEIGHT_SLIDER_MIN = -5;
-export const WEIGHT_SLIDER_MAX = 5;
+/**
+ * The Weight is the number that the WeightSlider actually provides to the
+ * instantiating component. The Weight varies in the range
+ * [0, 2**MAX_SLIDER_VALUE].
+ */
+export type Weight = number;
```
wchargin: Did you consider making `Weight` and `SliderPosition` opaque subtypes of
`number`? As it stands, in all modules they are interchangeable.

src/explorer/weights/WeightSlider.js
```diff
@@ -2,30 +2,61 @@
...
+
+/**
+ * The SliderPosition is the current position of the [range input] that the
+ * WeightSlider instantiates. It varies in the range [MIN_SLIDER, MAX_SLIDER].
```
wchargin: nit: maybe backtick `` `[MIN_SLIDER, MAX_SLIDER]` `` to contrast with
the Markdown link syntax `[range input]` used immediately both above and
below.

src/explorer/weights/WeightSlider.js
```diff
@@ -2,30 +2,61 @@
...
 
 export type Props = {|
-  +weight: number,
+  // The current Weight the WeightSlider is set to
+  +weight: Weight,
+  // The name associated with the slider.
+  // Will be displayed adjacent to the slider.
   +name: React$Node,
-  +onChange: (number) => void,
+  // onChange handler so the parent can recieve (and then set) the new weight
```
wchargin: nit: *receive* (sp.)

wchargin: Commenting `onChange: (Weight) => void` with “onChange handler” isn’t
very useful. Suggest something like “Callback: invoked with the new
weight after the user has adjusted the slider.”

src/explorer/weights/WeightSlider.js
```diff
@@ -2,30 +2,61 @@
...
+  +onChange: (Weight) => void,
 |};
+
+/**
+ * The WeightSlider is a React component for letting the user input
```
wchargin: It would probably be helpful to explicitly note that this is a
[“controlled component”][cc].

[cc]: https://reactjs.org/docs/forms.html#controlled-components

src/explorer/weights/WeightSlider.js
```diff
@@ -2,30 +2,61 @@
...
+ * a numeric weight that varies over an exponential range.
+ * The WeightSlider instantiates a slider that lets the user change
+ * the weight, displays the current weight value, and displays
+ * a "name" for the slider (which may be an arbitrary inline React
```
wchargin: What is an “inline React component”?

src/explorer/weights/WeightSlider.js
```diff
@@ -2,30 +2,61 @@
...
-export const WEIGHT_SLIDER_MAX = 5;
+/**
+ * The Weight is the number that the WeightSlider actually provides to the
+ * instantiating component. The Weight varies in the range
+ * [0, 2**MAX_SLIDER_VALUE].
+ */
+export type Weight = number;
+
+/**
+ * The SliderPosition is the current position of the [range input] that the
```
wchargin: Is `SliderPosition` intended to be an integer-valued type only?

What about `Weight`?

src/explorer/weights/WeightSlider.js
```diff
@@ -38,12 +69,43 @@ export class WeightSlider extends React.Component<Props> {
...
-function convertWeightToSliderValue(weight: number): number {
-  return weight === 0 ? WEIGHT_SLIDER_MIN : Math.log2(weight);
+/**
+ * Converts a Weight into a SliderPosition.
+ *
+ * If the Weight corresponds to an illegal slider value (i.e. weight < 0 or
+ * weight > 2**MAX_SLIDER), then an error will be thrown.
+ * be thrown.
+ *
+ * Over the domain of legal weights, `weightToSlider` and `sliderToWeight`
```
wchargin:
```js
 * Over the domain of legal weights, `weightToSlider` and `sliderToWeight`
 * compose to form the identity function.
```

This doesn’t seem accurate. The weight (say) `0.01` is accepted by this
function and doesn’t throw, but also doesn’t roundtrip:

    > WS.sliderToWeight(WS.weightToSlider(0.01))
    0

src/explorer/weights/WeightSlider.js
decentralion: Done.

src/explorer/weights/WeightSlider.js
decentralion: Good catch, thanks.

decentralion: Replaced with your proposed text.

src/explorer/weights/WeightSlider.js
decentralion: Moved it elsewhere in the docs, also backticked it.

src/explorer/weights/WeightSlider.js
decentralion: I could do that, but IMO it adds a little bit of complexity without any real benefit, because the module where "all of the action happens" (WeightSlider.js) will still allow improper conversions. And exporting the types from a separate module doesn't seem worth the effort. Underlying these tradeoffs: the SliderPosition isn't used elsewhere in the codebase (it is exported for testing purposes only) so if we did invest the effort to make this change robustly, we would not benefit from it.

src/explorer/weights/WeightSlider.js
decentralion: Nothing intelligble. Rewritten.

src/explorer/weights/WeightSlider.js
decentralion: SliderPosition is intended to be integer; Weight may be any finite nonnegative number. Updated the docstrings accordingly.

src/explorer/weights/WeightSlider.js
decentralion: You're right. Updated the documentation.

src/explorer/weights/WeightSlider.js
decentralion: It was a handwavey gesture at the idea that the `name` element should be inline (in a span). I've reworded the section.

src/explorer/weights/WeightSlider.js
decentralion: SliderPosition should be integer. Weight can be any nonnegative finite number. I've updated docs accordingly.

src/explorer/weights/WeightSlider.js
decentralion: You're right. Updated the docs.

decentralion:
@wchargin, should be ready for re-review.

PR #1006, opened 2018-11-16
Optimize `Graph.neighbors`

decentralion:
Callers of `Graph.neighbors` can choose not to do any prefix filtering
(i.e. get all neighbors) by passing `NodeAddress.empty` and
`EdgeAddress.empty` as the prefix filters. This means that we call
`Address.hasPrefix`, which calls `Address.isValid` on twice. This is
redundant work for a function that we know will return true.

This commit introduces a slight optimization so that in the case where
the prefix was the empty address, we shortcut this work. The commit was
motivated by seeing `neighbors`, `hasPrefix`, and `isValid` showing up
on some perf profiling of node decomposition related functionality. In
my profiling, this commit did not actually significantly improve the
perf. However, it's a simple and principled commit that makes sense, so
I'm submitting it anyway.

Test plan: Unit tests already have good coverage of calling `neighbors`
with an empty node or edge address.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -552,8 +552,14 @@ export class Graph {
     options: NeighborsOptions,
     initialModificationCount: ModificationCount
   ): Iterator<Neighbor> {
-    const nodeFilter = (n) => NodeAddress.hasPrefix(n, options.nodePrefix);
-    const edgeFilter = (e) => EdgeAddress.hasPrefix(e, options.edgePrefix);
+    const nodeFilter =
+      options.nodePrefix === NodeAddress.empty
+        ? (_) => true
+        : (n) => NodeAddress.hasPrefix(n, options.nodePrefix);
+    const edgeFilter =
+      options.edgePrefix === EdgeAddress.empty
+        ? (_) => true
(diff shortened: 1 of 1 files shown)


wchargin:
> In my profiling, this commit did not actually significantly improve
> the perf. However, it's a simple and principled commit that makes
> sense, so I'm submitting it anyway.

Hmm. Really? The code that you’re optimizing is `x.startsWith("N\0")`.
I can run about 30 million of those in a second, independent of the
length of `x`: <https://jsperf.com/startswith-short-string/1>

This code makes the code a bit harder to read, because the reader has to
wonder why the check is there even though the two branches appear to be
equivalent. It makes that part of the code marginally harder to change
later: e.g., we’ve talked about allowing unions of filters, like
`options.nodePrefixes`.

The commit is simple, but makes the code less simple than it currently
is. It’s principled in the sense that it’s a sound transformation, but
other than that I’m not sure what you mean by that claim.

I’m pretty reluctant to merge any performance change that doesn’t have a
performance impact.

decentralion:
I believe #1011 obviated the need for this PR.

PR #1007, opened 2018-11-17
Improve performance of pagerank `decompose`

### LINKED ISSUE #943: Optimize PagerankNodeDecomposition

decentralion:
When I implemented this function, I incorrectly assumed that
`lodash.sortBy` only calls subsequent accessor functions if there is a
tie from the first accessor. Actually, it calls it every time. We can
avoid lots of wasteful JSON.serialization by just grabbing the exact
properties of interest.

Test plan:

For correctness: `yarn test` suffices, as this functionality is already
tested.

For performance improvement: I ran the full load+analyze workflow, in
Chrome, on twbs/bootstrap. Before this change, decompose took 6.9s;
after this change, it takes 1.3s, for a 5.3x speedup.

Close #943.


src/analysis/pagerankNodeDecomposition.js
```diff
@@ -44,8 +44,11 @@ export function decompose(
         }
       ),
       (x) => -x.connectionScore,
-      // The following should be called rarely and on small objects.
-      (x) => JSON.stringify(x.connection.adjacency)
+      (x) => x.connection.adjacency.type,
+      (x) =>
+        x.connection.adjacency.type === "SYNTHETIC_LOOP"
```
wchargin: Please convert this to a `switch` with an empty-assertion. As it stands,
it is vulnerable to silently doing the wrong thing in the faces of
particular new kinds of adjacencies.

wchargin:
Oh. I see. `sortBy` does a [Schwartzian transform], which totally makes
sense, but makes it easier to see why they eagerly evaluate all
accessors. I’ll ask whether this is something that they’d like to fix by
computing these values lazily.

[Schwartzian transform]: https://en.wikipedia.org/wiki/Schwartzian_transform

decentralion:
Updated per your request.

Issue #1004, opened 2018-11-11
Find an efficient way to represent PageRank results


decentralion:
Currently, we represent the results of PageRank via the `PagerankNodeDecomposition`. The NodeDecomposition is a rich data type in that it allows us to answer the following questions:

- what nodes exist?
- for any node, what is its score?
- for any node, how can we attribute its score to its neighbors?

Thus, the PagerankNodeDecomposition has all the data needed to run the cred analysis workflow of the explorer. (Although the explorer wants some additional plugin-specific metadata so that it can display node names prettily.)

Unfortunately, the PagerankNodeDecomposition is [quite large](https://github.com/sourcecred/sourcecred/pull/995#issuecomment-437439528).
Now that we want [to serialize these results to disk](https://github.com/sourcecred/sourcecred/issues/967), we should find a better representation, lest we get blocked by the 100MB file size limit on GitHub pages.

## Size of a PagerankNodeDecomposition
Why is the PagerankNodeDecomposition so large? Let's analyze it.

Let `n` be the number of nodes in the graph, `e` be the number of edges, `A` be the size of an address (for convenience we assume that node and edge addresses are the same length), `F` be the size
of a float, and `K` be the size of a string constant (`"SYNTHETIC_LOOP", "IN_EDGE", "OUT_EDGE"`). 

Then the size of a PagerankNodeDecomposition is:

`n * (K + 3F + A) + e * (2K + 8A + 4F)`

Addresses are often quite long, so this expression is likely dominated by `n*A + e*8A`

<details>
<summary>(Derivation)</summary>

At a top level, we have:

```
export type PagerankNodeDecomposition = Map<
  NodeAddressT,
  {|
    +score: number,
    // Connections are sorted by `adjacencyScore` descending,
    // breaking ties in a deterministic (but unspecified) order.
    +scoredConnections: $ReadOnlyArray<ScoredConnection>,
  |}
>;
```

The size of a scoredConnection depends on whether it is a connection that came from an edge, or a connection that came from a synthetic self loop. However, every scoredConnection appears only once in the pagerankNodeDecomposition. And there is exactly one synthetic loop connection per node, and two edge connections
per edge. 
So at a top level, the size is:

`n*F + n * (loopConnection) + 2e * (edgeConnection)`

Now we analyze the connections:
```
export type ScoredConnection = {|
  +connection: Connection,
  +source: NodeAddressT,
  +connectionScore: number,
|};

export type Connection = {|
[...]

PR #1009, opened 2018-11-17
Use `WeightedGraph` in `createConnections`

decentralion:
This commit takes the `WeightedGraph` type added in #1008, and uses it
in `createConnections`.

`createConnections` now has a simplified signature (it just takes a
`WeightedGraph`), and a simplified implementation (it no longer needs to
calculate totalOutWeights).

Test plan: `createConnections` already has thorough unit testing (both
direct and indirect tests). Run `yarn test`.

Part of the #967 work group.

### DIFF
--- a/src/analysis/pagerank.js
+++ b/src/analysis/pagerank.js
@@ -6,6 +6,7 @@ import {
   createConnections,
   createOrderedSparseMarkovChain,
   type EdgeWeight,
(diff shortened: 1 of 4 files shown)


decentralion:
I'm closing this issue, as the design is likely obsoleted by something that comes out of #1004.

PR #1010, opened 2018-11-17
Add `singleNodeConnections`

decentralion:
This adds the function `singleNodeConnections` to the
`graphToMarkovChain` module. The contract is that for any node `n` in
the WeightedGraph `wg`, `singleNodeConnections(wg, n)` is equal to
`createConnections(wg).get(n)`, modulo connection ordering. The benefit
is that if you only need connections for a single node,
`singleNodeConnections` is much faster.

In the future, we could consider removing `createConnections` and having
all clients use `singleNodeConnections`. However, in the case where you
will need every connection (e.g. for creating the full markov chain),
`createConnections` is more efficient, so we would need to investigate
performance implications before making any such switch.

Test plan: I added two unit tests. One verifies that the results are
consistent with calling `createConnections` on the advanced graph.
The other explicitly models a small graph and ensures that score
normalization works properly. (Previous versions of this commit had bugs
in that logic, and this test would have caught them.)

### DIFF
--- a/src/core/attribution/graphToMarkovChain.js
+++ b/src/core/attribution/graphToMarkovChain.js
@@ -5,6 +5,9 @@ import {
   type Graph,
   type NodeAddressT,
   type EdgeAddressT,
+  Direction,
+  NodeAddress,
+  EdgeAddress,
 } from "../graph";
 import type {Distribution, SparseMarkovChain} from "./markovChain";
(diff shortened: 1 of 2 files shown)


decentralion:
I'm closing this PR, as the design is likely obsoleted by something that comes out of #1004.

PR #1012, opened 2018-11-17
Add `scoredConnections`

decentralion:
This adds `scoredConnections`, which allows you to compute the
`ScoredConnection[]` for a particular node. (The `WeightedGraph` is
required too).

This function is basically the single-node analogue of the full
`decompose` function from `pagerankNodeDecomposition`. I intend for this
function to wholly replace the `decompose` function. Once clients are no
longer calling `decompose`, I will remove that function, rename the
file, and slim down the `ScoredConnection` interface.

Test plan: I duplicated the testing for `decompose` so that we use the
same logic to test `scoredConnections`, and leave the path clear to
remove `decompose` down the line. I verified that the extra snapshot is
identical to the existing snapshot.

### DIFF
--- a/src/analysis/__snapshots__/pagerankNodeDecomposition.test.js.snap
+++ b/src/analysis/__snapshots__/pagerankNodeDecomposition.test.js.snap
@@ -169,3 +169,173 @@ Map {
   },
 }
 `;
+
+exports[`analysis/pagerankNodeDecomposition scoredConnections has the expected output on a simple asymmetric chain 1`] = `
+Map {
+  "NodeAddress[\\"n1\\"]" => Object {
+    "score": 0.19117656878499834,
+    "scoredConnections": Array [
+      Object {
+        "connection": Object {
+          "adjacency": Object {
+            "edge": Object {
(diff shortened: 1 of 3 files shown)


decentralion:
One future change I'm considering, which would change the interface to
`scoredConnections`, is creating a `ScoredWeightedGraph` type as
follows:

```
export type ScoredWeightedGraph = {|
  +weightedGraph: WeightedGraph,
  +scores: NodeScore,
|}
```

In general, the scores are almost meaningless without the associated
`WeightedGraph`, so in practice wherever we pass scores, we tend to pass
the Graph or WeightedGraph as a sibling. Explicitly packaging them
together in that type saves the trouble of passing two variables around.
It also would allow us to enforce at the type level that you cannot mix
and match the scores from one call to PageRank with a different graph.
(By exporting the ScoredWeightedGraph as an opaque subtype of a
structurally identical type.)

The disadvantage is that typing
`scoredWeightedGraph.weightedGraph.graph` is fairly tedious. (Or
`swg.weightedGraph.graph` in any case.) Maybe `swg.wg.graph`?

I welcome thoughts on this proposal.

PR #1013, opened 2018-11-18
Add `ReadOnlyGraph` interface

decentralion:
The `ReadOnlyGraph` is an interface for a graph that cannot be mutated.
As such, it only presents its read methods, e.g. `nodes`.

This change is motivated by my desire to create "enriched" graph
classes, e.g. `WeightedGraph`, which wrap the graph along with
additional metadata. For those enriched graphs to maintain integrity,
independently modifying the underlying graph must be disallowed, so this
is the interface they should implement.

In principle, I'd like to add `toJSON` and `equals` to this interface,
but it's not clear how to do that properly.

Test plan: `yarn flow` suffices.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -68,7 +68,24 @@ export opaque type GraphJSON = Compatible<{|
 
 type ModificationCount = number;
 
-export class Graph {
+// For clients that need to be able to read the state of a graph,
+// without ever mutating it.
+export interface ReadOnlyGraph {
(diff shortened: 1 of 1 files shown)


decentralion:
Leaving this up to get reviewer commentary, but as discussed in https://github.com/sourcecred/sourcecred/issues/1004#issuecomment-439726352, I think that I will not merge this PR.

PR #1014, opened 2018-11-18
Add the `WeightedGraph` class

decentralion:
As discussed in depth in #1004, it makes sense to package a Graph along
with the metadata needed to run Pagerank on it: hence, the
WeightedGraph, which is basically a Graph bundled with weights for every
edge, and a synthetic self loop weight.

In #1008, I created WeightedGraph as a simple struct. This commit
explores a different approach: WeightedGraph is reified as a full class
(a sibling to Graph) with a constructor that validates its invariants,
JSON (de)serialization, equality testing, and functions that implement
basic behaviors on top of it. The potential benefits of this approach
are discussed in [this comment](https://github.com/sourcecred/sourcecred/issues/1004#issuecomment-439641568)

The intention is that consumers that need read-only access to a Graph
should be able to use a WeightedGraph as well. To that end, the
WeightedGraph implements the ReadOnlyGraph interface.

One possible improvement: we could add `addNode` and `addEdge` to the
WeightedGraph, but with semantics that weights must be present. So,
`addEdge` requires an edge and a weight to both be provided. It wouldn't
be hard to implement. This is not needed by any existing client, but
having it could make it much easier to directly create PageRank-able
graphs. (If you don't need the decoupling offered by providing an edge
evaulator separately from specifiying the graph, it is really annoying.
I learned this while trying to create WeightedGraph test cases.)

Test plan: New, thorough unit tests have been added. Run `yarn test`.

### DIFF
--- a/src/core/__snapshots__/weightedGraph.test.js.snap
+++ b/src/core/__snapshots__/weightedGraph.test.js.snap
@@ -1,0 +1,44 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`core/weightedGraph to/from JSON snapshots on a simple graph 1`] = `
+Array [
+  Object {
+    "type": "sourcecred/weightedGraph",
+    "version": "0.1.0",
+  },
+  Object {
+    "graph": Array [
+      Object {
+        "type": "sourcecred/graph",
+        "version": "0.4.0",
+      },
+      Object {
+        "edges": Array [
+          Object {
+            "address": Array [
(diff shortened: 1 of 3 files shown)


decentralion:
As discussed in https://github.com/sourcecred/sourcecred/issues/1004#issuecomment-439726352, I think I'm going to rework this and create a new PR with a slightly different interface. However, I'm still interested in review comments.

decentralion:
Superceded by #1016

PR #1015, opened 2018-11-19
Retrieve sorted nodes/edges from GraphJSON

decentralion:
As discussed in #1004, we want to be able to package metadata with a
graph's nodes and edges. We can do this much more compactly if we store
the metadata as an array, ordered by the corresponding node/edge
address, rather than storing a map. The disadvantage is that clients
then need to manually sort the graph addresses to deserialize.

This commit adds public methods that allow a client to efficiently
retrieve the sorted addresses from the GraphJSON (where they are already
serialized). This behavior is tested. Note that we appropriately don't
allow clients to peek and directly depend on the exact representation of
GraphJSON, we just promise that sorted address retrieval is possible.

Test plan: Unit tests added (and I verified that breaking the sorting
will fail the test).

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -623,3 +623,35 @@ export function edgeToParts(
   const dstParts = NodeAddress.toParts(edge.dst);
   return {addressParts, srcParts, dstParts};
 }
+
+/*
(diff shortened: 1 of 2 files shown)


decentralion:
Merging on self-review; I need it for #1020.

PR #1017, opened 2018-11-21
Elide assertValid & assertValidParts in production

### LINKED ISSUE #1011: Elide `{Node,Edge}Address.assertValid` in production
I've found empirically that `assertValid` can be 5-10% of CPU usage in some production workflows (e.g. on-the-fly score decompositions). We should elide this in production; having it run in test code is sufficient to give us confidence that we aren't producing illegal addresses.

decentralion:
This commit substantially improves SourceCred's performance in
production.

Measurement methodology: I create a new tab in Chrome, navigate to my local
prototypes, and select go-ipfs. I then turn on profiling, and click the
analyze button, and then turn off profiling when analysis is done. I
then go to the "bottom-up" tab in the JS analysis box on the bottom and
sort by "Total Time".

__Before this commit:__

|        fn        | total time | time as % |
|:---------------- | ----------:| ---------:|
| assertValid      |      815ms |      8.6% |
| assertValidParts |      261ms |      2.7% |

__After this commit:__

|        fn        | total time | time as % |
|:---------------- | ----------:| ---------:|
| assertValid      |       21ms |      0.2% |
| assertValidParts |       23ms |      0.3% |

Test plan: `yarn test`, also performance measurement as described above.

Fixes #1011.

### DIFF
--- a/src/core/address.js
+++ b/src/core/address.js
@@ -126,9 +126,11 @@ export function makeAddressModule(options: Options): AddressModule<string> {
   );
 
   function assertValid(address: Address, what?: string): void {
-    // TODO(perf): If this function becomes a bottleneck, consider
-    // omitting it entirely in production. If this is undesirable, a
-    // number of micro-optimizations can be made.
+    // istanbul ignore if
+    if (process.env.NODE_ENV !== "test") {
+      // Catching invalid addresses in test code should be sufficient
+      return;
+    }
     const prefix = what == null ? "" : `${what}: `;
     if (address == null) {
       throw new Error(prefix + `expected ${name}, got: ${String(address)}`);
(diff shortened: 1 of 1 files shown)


src/core/address.js
```diff
@@ -126,9 +126,10 @@ export function makeAddressModule(options: Options): AddressModule<string> {
   );
 
   function assertValid(address: Address, what?: string): void {
-    // TODO(perf): If this function becomes a bottleneck, consider
-    // omitting it entirely in production. If this is undesirable, a
-    // number of micro-optimizations can be made.
+    if (process.env.NODE_ENV !== "test") {
```
wchargin: Mind adding `// istanbul ignore if` before this block? (and below)

decentralion:
I'm not sure if it disappears entirely in production. It's hard to tell with my instrumentation because all the function names are lost.

wchargin:
From the minified-then-prettified code (comments are my own):

```js
        function t(e, t) {  // assertValid
          return;
        }
        function r(e, t) {  // assertValidParts
          return;
        }
        function o(e) {  // fromParts
          return r(e), h + y + i(e);
        }
        function u(e) {  // toParts
          t(e);
          var r = e.split(y);
          return r.slice(1, r.length - 1);
        }
        function f(e, r) {  // hasPrefix
          return t(e, "address"), t(r, "prefix"), e.startsWith(r);
        }
```

So it looks like the functions continue to exist (which they must,
because they’re part of the `AddressModule` interface and thus part of
the return value) with now-empty bodies, and the call sites also
continue to exist.

An optimizer _should_ be able to remove the call sites at least in
`fromParts` and friends (maybe not in more distant call sites in other
modules, where the indirection is through a module object: it should
still be valid because we `Object.freeze` the module, but this is far
less obvious). But it also probably doesn’t matter very much, because
any decent JIT will likely optimize these call sites away, too.

Issue #998, opened 2018-11-04
Make Mirror robust to input data with incorrect `__typename`s


wchargin:
It has come to our attention that the GitHub API occasionally returns
results that are internally inconsistent, in that a node’s returned
`__typename` can be incorrect, and nodes can appear in places where they
are disallowed by the schema. The instances in which we know that this
can occur are documented in #996.

This issue describes a proposal for a way to cope with this incorrect
input data, without changing the direct public API of the Mirror module,
without changing the spirit in which it operates, and with only a local
performance impact. This proposal does not require making any Mirror
functions asynchronous or performing any cross-cutting external queries.

## Assumptions

Since it has become clear that the GitHub API is sometimes simply wrong,
we should be clear and explicit about our assumptions. Until now, we
have made the following assumption:

> Whenever the `id` and `__typename` of a `Node` are selected, the node
> with that ID will always have the same `__typename`, in all contexts.

This is implied by [§3 of the Relay Global Object Identification
Specification][relay-spec] and [GitHub’s own docs][github-docs], but it
is unfortunately clear that this assumption is false.

We relax our assumption to the following:

> Whenever a `Node` is queried by `id` via [the `node` field of the
> top-level `Query` object][node-field] or the `nodes` field on the same
> object, the returned `__typename` is the same.

In other words, a node’s type is well-defined, is constant, and has an
effective source of truth.

[relay-spec]: https://facebook.github.io/relay/graphql/objectidentification.htm#sec-Node-root-field
[github-docs]: https://developer.github.com/v4/guides/using-global-node-ids/#2-find-the-object-type-in-graphql
[node-field]: https://developer.github.com/v4/query/#node

This _is_ consistent with our observations. Note, for instance, that
even [when a user transforms into an organization, the node’s ID
changes][id-change]. Note also that GitHub’s IDs are base64-encoded
strings of the form `` `${int}:${graphqlTypename}${databaseId}` ``\*,
like:

```shell
$ ghquery '{ user(login: "wchargin") { id } }' | jq -r .data.user.id | base64 -d
04:User4317806
```

Of course, IDs must be treated as opaque—we cannot try to construct them
ourselves—but we can use them to infer implementation details about
GitHub’s systems. In particular, the fact that the node typename is
encoded directly into the ID, together with the fact that the GitHub
schema does not have any typenames that end with a decimal digit…

```shell
$ ghquery '{ __schema { types { name } } }' |
[...]

PR #1018, opened 2018-11-28
Update the blacklisted object ids

decentralion:
This adds a blacklisted id for @greenkeeper, a bot which used to be a
user. This is a temporary fix until we solve #998.

Test plan: `yarn test` passes. Before this commit, attempting to load
`probot/probot` fails. After this commit, it succeeds.


mikeal:
+1, bots users are actually excluded already from `gharchive` and a few other APIs, I'm surprised they show up here still.

decentralion:
Note: I requested @mikeal as a proxy reviewer for this rather trivial change. I want to merge it in because I want to be able to publish a probot cred attribution (and my local script always uses the master copy). Given his +1, I'm now going to push the red "squash and merge" button. :cowboy_hat_face:

decentralion:
> and my local script always uses the master copy

Heh heh, yes, I wrote that sheepishly

PR #1025, opened 2018-12-29
Add documentation to the Graph module

decentralion:
This commit adds a module-level docstring that gives an overview of the
Graph class and its importance to SourceCred, as well as adding
docstrings to specific methods.

Test plan:
In addition to review by the SourceCred maintainers, this should be
reviewed by at least one person who is not familiar with the codebase,
so that we can verify that it's actually working as documentation. :)

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -7,6 +7,80 @@ import {makeAddressModule, type AddressModule} from "./address";
 import {toCompat, fromCompat, type Compatible} from "../util/compat";
 import * as NullUtil from "../util/null";
 
+/**
+ * This module contains the Graph, which is one of the most fundamental pieces
+ * of SourceCred. SourceCred uses this graph to model all of the contributions
+ * that make up a project, and the relationships between those contributions.
+ *
+ * The Graph serves a simple function: it keeps track of which Nodes exist, and
+ * what Edges join those nodes to each other. Nodes and Edges are both identified
+ * by Addresses; specifically, `NodeAddressT`s and `EdgeAddressT`s.
+ *
+ * In both cases, addresses are modeled as arrays of strings. For example,
+ * you might want to give an address to your favorite node. You can do so as
+ * follows:
+ *
+ * const myAddress: NodeAddressT = NodeAddress.fromParts(["my", "favorite"])
+ *
+ * Edge Addresses are quite similar, except you use the EdgeAddress module.
+ *
+ * We model addresses as arrays of strings so that plugins can apply
+ * hierarchical namespacing for the address. In general, for any address, the
+ * first piece should be the name of the organization that owns the plugin, and
+ * the second piece should be the name of the plugin. Pieces thereafter are
+ * namespaced by the plugin's internal logic. For example, SourceCred has a
+ * Git plugin, and that plugin produces addresses like
+ * ["sourcecred", "git", "commit", "9cba0e9e212a287ce26e8d7c2d273e1025c9f9bf"].
+ *
+ * This enables "prefix matching" for finding only certain types of nodes. For
+ * example, if we wanted to find every Git commit in the graph, we
+ * could use the following code:
+ *
+ * const commitPrefix = NodeAddress.fromParts(["sourcecred", "git", "commit"])
+ * const commitNodes = `graph.nodes({prefix: commitPrefix})`
+ *
+ * Under the hood, addresses are represented as strings. Basically, we
+ * concatenate the pieces of the address together with the null seperator, and
+ * prepend a nonce ("E" or "N") to distinguish node and edge addresses. So the
+ * address for the node with pieces ["my", "favorite"] would be
+ * "N\0my\0favorite". Representing addreses as strings means that we can use
+ * them as keys for maps. To see the implementation, and the full address model
+ * API, check out src/core/address.js.
+ *
+ * In the case of a node, the address is all the graph knows about. If you have
+ * any other metadata you want to store about the node—like its commit message
+ * or its favorite color—you can store that in another database. The Graph is
+ * just a graph, not a key value store.
+ *
+ * Edges need a little more data, since the graph needs to know what nodes the
+ * edge connects. Each edge is an object with `src` and `dst` fields which are
+ * both NodeAddressT, and an address which is EdgeAddressT.
+ * Here's a toy example:
+ *
+ * const pr = NodeAddress.fromParts(["pull_request", "1"])
+ * const me = NodeAddress.fromParts(["user", "decentralion"])
+ * const authored = EdgeAddress.fromParts(["authored", "pull_request", "1"])
+ * const edge = {src: me, dst: pr, address: authored}
+ *
+ * Creating a graph is as simple as invoking the constructor and adding nodes and edges:
+ *
+ * const g = new Graph()
+ * g.addNode(pr);
+ * g.addNode(me);
+ * g.addEdge(edge);
+ *
+ * Graph has a number of accessor methods:
+ * - `hasNode` to check if a node is in the Graph
+ * - `nodes` to iterate over every node
+ * - `hasEdge` to check if an edge address is in the Graph
+ * - `edge` to retrieve an edge by its address
+ * - `edges` to query over every node
+ * - `neighbors` to find all the edges and nodes adjacent to a node
+ *    (also supports filtering by direction, by node prefix, or edge prefix)
+ *
+ * The Graph also has a few other convenience methods, like toJSON/fromJSON
+ * for serialization, and `Graph.merge` for combining multiple graphs.
+ */
 export opaque type NodeAddressT: string = string;
 export opaque type EdgeAddressT: string = string;
 export const NodeAddress: AddressModule<NodeAddressT> = (makeAddressModule({
@@ -20,6 +94,9 @@ export const EdgeAddress: AddressModule<EdgeAddressT> = (makeAddressModule({
   otherNonces: new Map().set("N", "NodeAddress"),
 }): AddressModule<string>);
 
+/**
+ * Represents an individual Edge in the Graph.
+ */
 export type Edge = {|
   +address: EdgeAddressT,
   +src: NodeAddressT,
@@ -122,6 +199,11 @@ export class Graph {
     this._maybeCheckInvariants();
   }
 
+  /**
+   * Adds a new NodeAddressT to the Graph.
+   *
+   * Adding the same node multiple times is safe.
+   */
   addNode(a: NodeAddressT): this {
     NodeAddress.assertValid(a);
     if (!this._nodes.has(a)) {
@@ -134,6 +216,15 @@ export class Graph {
     return this;
   }
 
+  /**
+   * Remove a NodeAddressT from the Graph.
+   *
+   * If the node is incident to any edges, those edges must be removed
+   * before removing the node. Attempting to remove a node that is incident
+   * to some edges will throw an error.
+   *
+   * Attempting to remove a node that is not in the graph is safe.
+   */
   removeNode(a: NodeAddressT): this {
     NodeAddress.assertValid(a);
     const existingInEdges = this._inEdges.get(a) || [];
@@ -156,6 +247,9 @@ export class Graph {
     return this;
   }
 
+  /**
+   * Test whether a given NodeAddressT is present in the Graph.
+   */
   hasNode(a: NodeAddressT): boolean {
     NodeAddress.assertValid(a);
     const result = this._nodes.has(a);
@@ -163,6 +257,13 @@ export class Graph {
     return result;
   }
 
+  /**
+   * Returns an interator over all of the NodeAddressTs in the Graph.
+   *
+   * Optionally, the caller can provide a node prefix as a NodeAddressT. If
+   * provided, the iterator will only contain NodeAdressTs matching that
+   * prefix.
+   */
   nodes(options?: {|+prefix: NodeAddressT|}): Iterator<NodeAddressT> {
     const prefix = options != null ? options.prefix : NodeAddress.empty;
     if (prefix == null) {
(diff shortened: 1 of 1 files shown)


decentralion:
@anthrocypher @lpatmo @julianjohannesen it would be great to get a review from you!

src/core/graph.js
```diff
@@ -524,6 +684,14 @@ export class Graph {
     this._maybeCheckInvariants();
   }
 
+  /**
+   * Checks whether this Graph is equal to another Graph.
+   *
+   * Two Graphs are considered equal if they have identical sets of nodes
+   * and edges.
+   *
+   * Runs in O(n + e) where n is the number of nodes and e is the number of edges.
```
wchargin:
```
   * Checks whether this Graph is equal to another Graph.
   *
   * Two Graphs are considered equal if they have identical sets of nodes
   * and edges.
```

Good. Could choose to explicitly call out that insertion order and
duplication are irrelevant.

```
   * Runs in O(n + e) where n is the number of nodes and e is the number of edges.
```

Specify “Runs in time `O(n + e)`” (or “worst-case time” if you prefer).
Add comma before “where”.

src/core/graph.js
```diff
@@ -535,12 +703,29 @@ export class Graph {
     return result;
   }
 
+  /**
+   * Produce a copy of a Graph.
```
wchargin:
```
   * Produce a copy of a Graph.
```

…of _this_ graph, not just of _a_ graph.

```
   *
   * The copy is equal to the original, but distinct, so that they may be
   * modified independently.
```

Good.

src/core/graph.js
```diff
@@ -535,12 +703,29 @@ export class Graph {
...
+   * modified independently.
+   */
   copy(): Graph {
     const result = Graph.merge([this]);
     this._maybeCheckInvariants();
     return result;
   }
 
+  /**
+   * Serialize a Graph into JSON.
```
wchargin:
```
   * Serialize a Graph into JSON.
```

Ambiguous: “JSON” could well be interpreted to mean a string rather than
what you mean, which is a “plain old JavaScript object” or “JSON
object”.

```
   *
   * We store the nodes in sorted order. This allows us to serialize the edges
   * in an indexed representation, where each edge has an address and then
   * represents its src and dst by sorted node indices.
   *
   * This greatly reduces the size of the representation, and means that this
   * method runs in O(n log n + e) where n is the number of nodes and e is the
   * number of edges.
```

These are implementation details and should _not_ be documented. If we
want an API that exposes the performance benefits of this representation
to other clients, then that is a totally separate matter.

Suggest replacing with:

> Runs in time `O(n log n + e)`, where `n` is the number of nodes and
> `e` is the number of edges.

src/core/graph.js
```diff
@@ -575,6 +763,18 @@ export class Graph {
     return result;
   }
 
+  /**
+   * Allows merging multiple graphs.
```
wchargin:
```
   * Allows merging multiple graphs.
```

It doesn’t “allow” anything; it actually merges the graphs! And I can
infer that from the method name—if I’m reading these docs, I’m probably
trying to figure out what “merging” graphs actually means. This is not a
helpful summary.

Suggest:

> Compute the union of the given graphs. The result is a new graph that
> has all the nodes and edges from all the provided graphs.
>
> If two of the given graphs have edges with the same address, the edges
> must be equal (i.e., must have the same source and destination in each
> graph). If this is not the case, an error will be thrown.

```
   *
   * Example usage:
   * const g1 = new Graph().addNode(a).addNode(b).addEdge(x);
   * const g2 = new Graph().addNode(c);
   * const g3 = Graph.merge([g1, g2]);
```

Good to have an example, but this example isn’t particularly meaningful. In particular, the two graphs that are merged are disjoint to begin with. Suggest something like:

> ```js
> const g1 = new Graph().addNode(a).addNode(b).addEdge(e);
> const g2 = new Graph().addNode(b).addNode(c).addEdge(f);
> const g3 = Graph.merge([g1, g2]);
> Array.from(g3.nodes()).length;  // 3
> Array.from(g3.edges()).length;  // 2
> ```

```
   *
   * The new Graph will contain the union of all the nodes and edges in each of
   * the merged graphs. It is a separate instance and may be mutated
   * independently.
```

Good (though may merit some rewording in response to comments on first
paragraph).

src/core/graph.js
```diff
@@ -589,6 +789,9 @@ export class Graph {
   }
 }
 
+/**
+ * Conert an edge into a human readable string.
```
wchargin:
```
 * Conert an edge into a human readable string.
```

sp.: *Convert*

Need to specify that the precise behavior is an implementation detail
and is subject to change.

Should we specify that the function is injective?

src/core/graph.js
```diff
@@ -7,6 +7,80 @@ import {makeAddressModule, type AddressModule} from "./address";
...
+ * ["sourcecred", "git", "commit", "9cba0e9e212a287ce26e8d7c2d273e1025c9f9bf"].
+ *
+ * This enables "prefix matching" for finding only certain types of nodes. For
+ * example, if we wanted to find every Git commit in the graph, we
+ * could use the following code:
+ *
+ * const commitPrefix = NodeAddress.fromParts(["sourcecred", "git", "commit"])
+ * const commitNodes = `graph.nodes({prefix: commitPrefix})`
+ *
+ * Under the hood, addresses are represented as strings. Basically, we
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/graph.js
```diff
@@ -7,6 +7,80 @@ import {makeAddressModule, type AddressModule} from "./address";
...
+ * Here's a toy example:
+ *
+ * const pr = NodeAddress.fromParts(["pull_request", "1"])
+ * const me = NodeAddress.fromParts(["user", "decentralion"])
+ * const authored = EdgeAddress.fromParts(["authored", "pull_request", "1"])
+ * const edge = {src: me, dst: pr, address: authored}
+ *
+ * Creating a graph is as simple as invoking the constructor and adding nodes and edges:
+ *
+ * const g = new Graph()
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/graph.js
```diff
@@ -20,6 +94,9 @@ export const EdgeAddress: AddressModule<EdgeAddressT> = (makeAddressModule({
   otherNonces: new Map().set("N", "NodeAddress"),
 }): AddressModule<string>);
 
+/**
+ * Represents an individual Edge in the Graph.
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/graph.js
```diff
@@ -304,6 +386,15 @@ export class Graph {
     return this;
   }
 
+  /**
+   * Remove a NodeAddressT from the Graph.
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/graph.js
```diff
@@ -326,13 +417,23 @@ export class Graph {
...
+   */
   hasNode(a: NodeAddressT): boolean {
     NodeAddress.assertValid(a);
     const result = this._nodes.has(a);
     this._maybeCheckInvariants();
     return result;
   }
 
+  /**
+   * Returns an interator over all of the NodeAddressTs in the Graph.
```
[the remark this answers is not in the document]
decentralion: Done. (Is it OK that I didn't capitalize MUST NOT and MAY?)

src/core/graph.js
```diff
@@ -421,20 +537,43 @@ export class Graph {
...
+   */
   hasEdge(address: EdgeAddressT): boolean {
     EdgeAddress.assertValid(address);
     const result = this._edges.has(address);
     this._maybeCheckInvariants();
     return result;
   }
 
+  /**
+   * Returns the Edge matching a given EdgeAddressT, if such an edge exists, or
```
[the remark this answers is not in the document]
decentralion: Ack.

src/core/graph.js
```diff
@@ -421,20 +537,43 @@ export class Graph {
...
+   */
   edge(address: EdgeAddressT): ?Edge {
     EdgeAddress.assertValid(address);
     const result = this._edges.get(address);
     this._maybeCheckInvariants();
     return result;
   }
 
+  /**
+   * Returns an iterator over every Edge in the Graph.
```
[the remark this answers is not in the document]
decentralion: Great suggestion, thanks.

src/core/graph.js
```diff
@@ -395,6 +505,12 @@ export class Graph {
     return this;
   }
 
+  /**
+   * Remove the edge matching a given EdgeAddressT from the graph.
```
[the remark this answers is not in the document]
decentralion: Makes sense. Done.

src/core/graph.js
decentralion: Fixed. I don't think we need to specify that it's injective.

decentralion:
@wchargin I've incorporated your suggestions; PTAL.

src/core/graph.js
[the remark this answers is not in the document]
wchargin: Using the lowercase versions is fine with me. To be clear, all-caps “MUST”, “SHOULD”, “MAY”, and similar have specific meanings defined in [RFC 2119]. If you use the all-caps versions in documentation or elsewhere in “permanent artifacts” (like commit messages), you SHOULD refer to that RFC somewhere in the docs (as the RFC itself suggests). When I use them in discussion threads or chat or other “non-permanent” media, I usually intend them in the same way but don’t bother linking the RFC; it’s just a convenient vocabulary.

[RFC 2119]: https://www.ietf.org/rfc/rfc2119.txt

src/core/graph.js
```diff
@@ -7,6 +7,72 @@ import {makeAddressModule, type AddressModule} from "./address";
...
+ * namespaced by the plugin's internal logic. For example, SourceCred has a
+ * Git plugin, and that plugin produces addresses like
+ * ["sourcecred", "git", "commit", "9cba0e9e212a287ce26e8d7c2d273e1025c9f9bf"].
+ *
+ * This enables "prefix matching" for finding only certain types of nodes. For
+ * example, if we wanted to find every Git commit in the graph, we
+ * could use the following code:
+ *
+ * const commitPrefix = NodeAddress.fromParts(["sourcecred", "git", "commit"]);
+ * const commitNodes = `graph.nodes({prefix: commitPrefix})`;
```
lpatmo: Is `graph.nodes...` meant to be inside a template literal?

src/core/graph.js
```diff
@@ -134,6 +211,18 @@ export class Graph {
...
+   * If the node does not exist in the graph, no action is taken and no error
+   * is thrown. (This operation is idempotent.)
+   *
+   * If the node is incident to any edges, those edges must be removed
+   * before removing the node. Attempting to remove a node that is incident
+   * to some edges will throw an error.
+   *
+   * Returns `this` for chaining.
+   */
   removeNode(a: NodeAddressT): this {
```
lpatmo: Never seen `removeNode(a: NodeAddressT): this {` syntax before. Guessing it's syntax from unit tests?

src/core/graph.js
wchargin: Good catch; it shouldn’t be.

(Aside: In Rust, if you put code into a doc comment, it’s automatically
turned into a test case. Would be nice to have something similar here.)

src/core/graph.js
```diff
@@ -7,6 +7,72 @@ import {makeAddressModule, type AddressModule} from "./address";
...
+ * g.addNode(pr);
+ * g.addNode(me);
+ * g.addEdge(edge);
+ *
+ * Graph has a number of accessor methods:
+ * - `hasNode` to check if a node is in the Graph
+ * - `nodes` to iterate over every node
+ * - `hasEdge` to check if an edge address is in the Graph
+ * - `edge` to retrieve an edge by its address
+ * - `edges` to query over every node
```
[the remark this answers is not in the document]
wchargin: The text should say “to iterate over every edge” (but this is a bit of a
simplification, as `g.edges` accepts options). Yes, `g.edges()` returns
an iterator that includes every `Edge` object.

src/core/graph.js
wchargin: These are [Flow] types: the argument has to be a `NodeAddressT`, and the
return type is guaranteed to be the `this` object. (The latter is a bit
stronger than just declaring a return type of `Graph`: if we declare a
class `Subgraph` that extends `Graph`, then we’ll know that `Subgraph`’s
`removeNode` method also returns a `Subgraph` and not just a `Graph`.)

It’s not related to unit tests, no.

[Flow]: https://flow.org/

src/core/graph.js
```diff
@@ -7,6 +7,80 @@ import {makeAddressModule, type AddressModule} from "./address";
...
 import * as NullUtil from "../util/null";
 
+/**
+ * This module contains the Graph, which is one of the most fundamental pieces
+ * of SourceCred. SourceCred uses this graph to model all of the contributions
+ * that make up a project, and the relationships between those contributions.
+ *
+ * The Graph serves a simple function: it keeps track of which Nodes exist, and
+ * what Edges join those nodes to each other. Nodes and Edges are both identified
+ * by Addresses; specifically, `NodeAddressT`s and `EdgeAddressT`s.
```
anthrocypher: Here we may want to add a link to a resource introducing the concept of graphs more generally. This seems like a good candidate: https://softwareengineering.stackexchange.com/questions/168058/what-are-graphs-in-laymens-terms#168067

src/core/graph.js
```diff
@@ -7,6 +7,80 @@ import {makeAddressModule, type AddressModule} from "./address";
...
+ * them as keys for maps. To see the implementation, and the full address model
+ * API, check out src/core/address.js.
+ *
+ * In the case of a node, the address is all the graph knows about. If you have
+ * any other metadata you want to store about the node—like its commit message
+ * or its favorite color—you can store that in another database. The Graph is
+ * just a graph, not a key value store.
+ *
+ * Edges need a little more data, since the graph needs to know what nodes the
+ * edge connects. Each edge is an object with `src` and `dst` fields which are
```
anthrocypher: What is the significance of the `src` and `dst` fields?

src/core/graph.js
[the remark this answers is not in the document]
decentralion: Fixed the text, thank you @lpatmo

src/core/graph.js
decentralion: Done, thanks

src/core/graph.js
decentralion: The "source" and "destination" of the edge. Re-wrote the paragraph to clarify :+1:

decentralion:
Thank you for the reviews @wchargin, @lpatmo, and @anthrocypher. I've updated the PR and will now merge.

PR #1026, opened 2018-12-29
Graph: move invariant checker to bottom of class

decentralion:
This moves the invariant checking code from the top of the Graph class
to the bottom. Most readers of this file will probably be more
interested in seeing the API, and reading the invariant checker first
is likely to be confusing and off-putting.

Test plan: `yarn test` suffices. No semantic change.


decentralion:
Force merging as trivial reorg not requiring review.

PR #1029, opened 2019-01-02
add timestamp to registry entries

expravit:
This attaches a timestamp to every entry as it is loaded into the registry.

I have not begun any new testing on this in any way yet. I have modified the relevant tests that deal with loading the entries into the registry. Which seems to accept the new form of the entry data-structure, however it fails to recieve the the expected timestamp. I'm very new to testing but I think I have a way to fix up my crude timestamp 'monkey-patching' in the tests using a feature in jest called Snapshot Property Matchers. I believe this would be a much better solution than the weird timestamp export that i have placed inside of load.test.js

Any comments, advice, or further discussion would be greatly appreciated.

### DIFF
--- a/src/cli/load.js
+++ b/src/cli/load.js
@@ -213,7 +213,8 @@ function addToRepoIdRegistry(repoId) {
   } else {
     registry = RepoIdRegistry.emptyRegistry();
   }
-  registry = RepoIdRegistry.addRepoId(registry, {repoId});
+  let timestamp = new Date();
+  registry = RepoIdRegistry.addRepoId(registry, {repoId, timestamp});
 
   fs.writeFileSync(outputFile, stringify(RepoIdRegistry.toJSON(registry)));
 }
--- a/src/cli/load.test.js
+++ b/src/cli/load.test.js
@@ -18,6 +18,8 @@ jest.mock("../plugins/git/loadGitData", () => ({
   loadGitData: jest.fn(),
 }));
 
+export const fakeTimestamp = new Date();
+
 type JestMockFn = $Call<typeof jest.fn>;
 const execDependencyGraph: JestMockFn = (require("../tools/execDependencyGraph"): any);
 const loadGithubData: JestMockFn = (require("../plugins/github/loadGithubData")
@@ -430,7 +432,7 @@ describe("cli/load", () => {
           .toString();
         const registry = RepoIdRegistry.fromJSON(JSON.parse(blob));
         const expected: RepoIdRegistry.RepoIdRegistry = [
-          {repoId: stringToRepoId("foo/combined")},
+          {repoId: stringToRepoId("foo/combined"), timestamp: fakeTimestamp},
         ];
         expect(registry).toEqual(expected);
       });
@@ -441,8 +443,14 @@ describe("cli/load", () => {
           path.join(sourcecredDirectory, RepoIdRegistry.REPO_ID_REGISTRY_FILE),
           JSON.stringify(
             RepoIdRegistry.toJSON([
-              {repoId: stringToRepoId("previous/one")},
-              {repoId: stringToRepoId("previous/two")},
+              {
+                repoId: stringToRepoId("previous/one"),
+                timestamp: fakeTimestamp,
+              },
+              {
+                repoId: stringToRepoId("previous/two"),
+                timestamp: fakeTimestamp,
+              },
             ])
           )
         );
@@ -455,9 +463,9 @@ describe("cli/load", () => {
           .toString();
         const registry = RepoIdRegistry.fromJSON(JSON.parse(blob));
         const expected: RepoIdRegistry.RepoIdRegistry = [
-          {repoId: stringToRepoId("previous/one")},
-          {repoId: stringToRepoId("previous/two")},
-          {repoId: stringToRepoId("foo/combined")},
+          {repoId: stringToRepoId("previous/one"), timestamp: fakeTimestamp},
+          {repoId: stringToRepoId("previous/two"), timestamp: fakeTimestamp},
+          {repoId: stringToRepoId("foo/combined"), timestamp: fakeTimestamp},
         ];
         expect(registry).toEqual(expected);
       });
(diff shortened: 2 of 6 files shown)


expravit:
I think I was supposed to merge to my master branch, and then make the PR - will this request still be sufficient?

src/core/repoIdRegistry.js
```diff
@@ -14,6 +14,7 @@ const REPO_ID_REGISTRY_COMPAT = {type: "REPO_ID_REGISTRY", version: "0.2.0"};
 
 export type RegistryEntry = {|
   +repoId: RepoId,
+  +timestamp: Date,
```
decentralion: JS date objects scare me, so I'd likely prefer timestamp in ms since epoch, as retrievable from [Date.now](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now). So as to be unambiguous, the field can be called `timestampMsSinceEpoch`. 
Pinging @wchargin  in case he has another opinion

src/core/repoIdRegistry.test.js
```diff
@@ -9,6 +9,7 @@ import {
 } from "./repoIdRegistry";
 
 import {makeRepoId} from "../core/repoId";
+import {fakeTimestamp} from "../cli/load.test";
```
decentralion: src/core should never depend on src/cli. if we needed a fake timestamp export, it should be exported from core and then imported in src/cli. but exporting a variable that is just a fake timestamp is unnecessary, I think--each file can define its own fake timestamp

src/cli/load.test.js
```diff
@@ -18,6 +18,8 @@ jest.mock("../plugins/git/loadGitData", () => ({
   loadGitData: jest.fn(),
 }));
 
+export const fakeTimestamp = new Date();
```
decentralion: Having test code that references the actual current date makes me uncomfortable. It's an invitation for flaky tests. It would be better to make the fakeTimestamp a specific fixed date, e.g. new Date(1000)

src/cli/load.js
```diff
@@ -213,7 +213,8 @@ function addToRepoIdRegistry(repoId) {
   } else {
     registry = RepoIdRegistry.emptyRegistry();
   }
-  registry = RepoIdRegistry.addRepoId(registry, {repoId});
+  let timestamp = new Date();
```
decentralion: We use `const` in the codebase by default, and only use `let` when we need to re-assign the variable. it makes the code a little easier to grok, if you know that once something is assigned, it can never be re-assigned

src/cli/load.test.js
wchargin: Yep, definitely.

Not sure why we need to export a global `fakeTimestamp` at all. Can’t
each test instantiate its own?

Because date objects are mutable, please don’t use a global constant,
but rather a global function:

```js
const fakeTimestamp = () => new Date(123);
```

src/core/repoIdRegistry.js
expravit: Okay, sounds good, I'll change it in a bit.

wchargin: Eh… JS `Date` objects certainly aren’t great\*, but I don’t see any
reason to be scared of them. Just treat them as if they were immutable
and you should be okay. I have a moderate preference for `Date` over
`timestampMsSinceEpoch`: `Date` encapsulates the units perfectly well.

\* Primarily: they’re mutable, and `getDay()` does not get the day.

src/core/repoIdRegistry.test.js
expravit: This sounds like a much better idea to me as well. Trying to go about this in a practical sense was very difficult for me, testing is still a foreign concept to me at this point. I'll fix this up though and perhaps we can find a better solution later on.

I watched a presentation from the Jest team the other night [https://youtu.be/cAKYQpTC7MA](url), they were discussing the new release(v23) of Jest and some of it's new features. At 42:47 in the video one of the devs begin to talk about using mock dates with a feature called snapshot property matchers, this sounds like it might be relevant and useful. I definitely need to look more into this today.

src/cli/load.js
expravit: I don't know why I did this, it must've been amatuer-hour or something - I know better, thank you for pointing that out though.

expravit:
I have no problem working with either of them; ultimately, as far as i'm concernced, it's whatever everyone else is comfortable using. You guys can make the final call on this one.

src/cli/load.test.js
expravit: Definitely! This is the missing piece I was looking for that was giving me trouble with the current unit testing.

src/cli/load.js
decentralion: You're welcome, and no worries! Our code reviews are certified judgement free and produced without any animal testing. (Just unit testing.)

decentralion:
My actual concern in using Date is around JSON serialization/deserialization. Since dates are not natively supported in JSON, they will serialize as strings; would we need to update the fromJSON function to parse those strings as dates? I notice that we already have round-trip serialization tests; do those pass with the code as currently written? (And are we sure they aren't spuriously passing due to some weird coercive equality semantics?)

expravit:
@decentralion yeah so I think the tests are failing due to Jest's Assymetric Equality thing. I'm about to look into it and fix this code up. I'll make the timestamp change to ms for the JSON serialization and will leave it alone until further notice.

wchargin:
The tests are passing right now because `fromJSON` and `toJSON` really
are inverses. But `fromJSON` and `toJSON` don’t make it all the way into
the string layer: they return JSON-objects, not JSON-strings. Jest is
fine; it isn’t performing coercive tests. A test using functionality as
in `checkedLocalStore.js` would catch the roundtrip-through-string
behavior.

Yes, `fromJSON` needs to be updated. But that’s the whole point of
providing `toJSON` and `fromJSON`: they enable us to use appropriate
object types in JS-land and then marshal to plain values at the
serialization boundary.

The forcing change that should be made is to change

```js
export type RepoIdRegistryJSON = Compatible<RepoIdRegistry>;
```

to

```js
export type RepoIdRegistryJSON = Compatible<$ReadOnlyArray<{|
  +repoId: RepoId,
  +timestamp: string,  // ISO 8601, as `Date.prototype.toJSON()`
|}>>;
```

(which should probably also be opaque…?) and then fix the resulting type
errors.

Issue #1019, opened 2018-11-28
Redirect `/prototypes` to `/prototype`


decentralion:
Consider the following url:

https://sourcecred.io/prototypes/sourcecred/sourcecred/

If you wanted to find other prototypes, it would be natural to truncate off the `sourcecred/sourcecred`, resulting in

https://sourcecred.io/prototypes

Unfortunately, this is a dead link. The correct link is: https://sourcecred.io/prototype

We can easily solve this by adding an internal redirect from `/prototypes` to `/prototype`. Likely we should copy from or extend [ExternalRedirect.js](https://github.com/sourcecred/sourcecred/blob/master/src/homepage/ExternalRedirect.js)

wchargin:
Let’s rather redirect `/prototype/` to `/prototypes/`, yes?

As the name suggests, `ExternalRedirect` is not the right tool for the
job. You’ll want a new `InternalRedirect`. The server-side rendering
code will probably have to change a bit; see [the docs for `match`][1].

[1]: https://github.com/ReactTraining/react-router/blob/v3/docs/API.md#match-routes-location-history-options--cb

expravit:
Why is `/prototypes/` even a route? 

I mean semantically - yes - `/prototypes/` infers that this route holds multiple items(plural) that one may be apt to check out. However we really only have one prototype we're dealing with(in this case, the application itself, not the instances/routes we've laoded into the app), don't we?

Why not disregard `/prototypes/` altogether and just use `/prototype/` for all of our various routing needs? Am I completely off-base with this, why not disregard the redirect and strictly use one route(both for landing on the prototype page, and for traversing the appended repo-entries) instead of using two routes that are more-or-less inconsistent with one another?

wchargin:
@expravit: Yes, we should definitely pick either `prototype` or
`prototypes` and use that name consistently within all routes. I don’t
have a strong preference as to whether it should be singular or plural.
(Both are common, in my experience: this issue URL, for instance, is
`/issues/1019`, whereas pull requests are like `/pull/1025`.)

The discussion of a redirect is just for backward compatibility, to not
break existing permalinks that we’ve shared. But if we end up using the
singular form throughout then this might not be necessary, as we’ve
mostly told people to visit `/prototype/`.

decentralion:
I don't think we've shared many links to /prototypes/, so I feel good about just deprecating the prototypes/ route as @expravit suggested, without setting up the redirect. Save the effort for things that will really matter to the project :+1:

PR #1030, opened 2019-01-02
redirect routing for prototype

expravit:
This PR is a fix for Issue #1019 

Using just the prototype route seems to be working perfectly fine and I think this might be ready for master.

However when I omit the trailing-slash from the route the page will not display. I don't know if making the trailing-slash at the end of the route would break anything but I think it would be kinda useful/helpful.

### DIFF
--- a/src/homepage/PrototypesPage.js
+++ b/src/homepage/PrototypesPage.js
@@ -26,7 +26,7 @@ export default function makePrototypesPage(
           <ul>
             {registry.map((x) => (
               <li key={stringify(x)}>
-                <Link to={`/prototypes/${x.repoId.owner}/${x.repoId.name}/`}>
+                <Link to={`/prototype/${x.repoId.owner}/${x.repoId.name}/`}>
                   {`${x.repoId.owner}/${x.repoId.name}`}
                 </Link>
               </li>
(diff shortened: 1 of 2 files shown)


wchargin:
Every route must have a trailing slash. We have a test case for this
that explains why:
<https://github.com/sourcecred/sourcecred/blob/7c7fa2d83d4fd5ba38efb2b2f4e0244235ac1312/src/homepage/routeData.test.js#L27-L51>

I would welcome a patch to make `Link` verify that the value of its `to`
prop ends in a slash, throwing a helpful error message (maybe linking to
that test case) otherwise.

wchargin:
> This PR is a fix for Issue #1019

Prefer to write “Fixes #1019”. If you use _fixes_, _resolves_, or
_closes_ followed by an issue or pull request number, the issue or PR
will automatically be closed when this pull request is merged.

> However when I omit the trailing-slash from the route the page will not display

Correct (see my previous comment)—but if you omit the trailing slash
from the route and then run `yarn unit`, then you should see a helpful
error message. :-)

> please add a test plan to the pull request

Yes, please do so. Let me also suggest that that test plan include
something like, “Running `git grep -F '/prototypes/'` returns no
results; before this commit, it yielded 2 results.” More importantly, I
think that you’ll find that adding `git grep` steps to your mental test
plan for changes like this will catch actual bugs more often than you
might think!

This PR will have my approval once the description is updated.

wchargin:
You may find it helpful to read: <https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#when-writing-commit-messages>

decentralion:
I've run William's proposed test, and am going to merge this PR with an updated description.

PR #1032, opened 2019-01-03
Add project title to explorer page

### LINKED ISSUE #1027: Add a project title on Project Explorer pages
A title on individual project explorer pages that signifies which project you're looking at, eg `repoId.owner}/repoId.name` ? Above the Analyze Cred Button ?

BrianLitwin:
Resolves #1027

Using `repoId.owner/repoId.name` for the project title
because that is how projects are identified on `PrototypePage`.

Created a `<ProjectDetail />` component inside `<App />`  that consumes a `RepoId`
and renders a title.

**Test Plan:**

Added two unit tests: 

The first verifies that the parent `<App />` component
instantiates a `<ProjectDetai />` component with the correct props. 
The current correct prop is a `RepoId` object. 

The second test verifies that the `<ProjectDetail />` component renders
the title correctly given the `RepoId`, ie as a `<p>` element
with `repoId.owner/repoId.name` for text.

Visual tests verify that the title is above the Analyze Cred
button, and that clicking from one project to another renders
the appropriate title for separate projects.

Attaching a screenshot as a comment at #1032
for reference: 

screenshot 2019-01-04 13 40 03

### DIFF
--- a/src/explorer/App.js
+++ b/src/explorer/App.js
@@ -112,6 +112,7 @@ export function createApp(
             {spacer()}
             <Link href={feedbackUrl}>feedback</Link>
           </p>
+          <ProjectDetail repoId={this.props.repoId} />
           <button
             disabled={
               appState.type === "UNINITIALIZED" ||
@@ -134,6 +135,14 @@ export function createApp(
       );
     }
   };
+}
+
+export class ProjectDetail extends React.PureComponent<{|
+  +repoId: RepoId,
+|}> {
+  render() {
+    return <p>{`${this.props.repoId.owner}/${this.props.repoId.name}`}</p>;
+  }
 }
 
 export class LoadingIndicator extends React.PureComponent<{|
--- a/src/explorer/App.test.js
+++ b/src/explorer/App.test.js
@@ -12,7 +12,7 @@ import {FactorioStaticAdapter} from "../plugins/demo/appAdapter";
 import {defaultWeightsForAdapter} from "./weights/weights";
 
 import {PagerankTable} from "./pagerankTable/Table";
-import {createApp, LoadingIndicator} from "./App";
+import {createApp, LoadingIndicator, ProjectDetail} from "./App";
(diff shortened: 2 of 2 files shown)


src/explorer/App.js
```diff
@@ -112,6 +112,9 @@ export function createApp(
             {spacer()}
             <Link href={feedbackUrl}>feedback</Link>
           </p>
+          <ProjectDetail
```
decentralion: If we're going to have a ProjectDetail component, then that component should have the logic for how we construct a title for a given repoId. Otherwise, we could just directly create the <p> here.

So, I suggest one of the following two changes:

- Create a ProjectDetail component which takes a `RepoId` as a prop, and then uses that to render a title
OR
- Just directly create a <p> here with the right title, without the indirection of an additional component.

I have a preference for option 1 as it will scale better in complexity as we add more "project detail" type information (which I imagine we will)

src/explorer/App.test.js
```diff
@@ -111,6 +111,14 @@ describe("explorer/App", () => {
     expect(link.prop("href")).toMatch(/https?:\/\//);
   });
 
+  it("should have a project title", () => {
```
decentralion: Assuming that we stick with a ProjectDetail component, there should be two tests.

1. The containing component instantiates a ProjectDetail with the right props
2. The ProjectDetail renders an example RepoId correctly

src/explorer/App.js
decentralion: Yup!

src/explorer/App.test.js
decentralion: Yep, exactly. It's nice to have very narrowly scoped tests, makes it easy to understand and resolve failures.

BrianLitwin:
@decentralion I've updated the code + tests as per your comments... thanks for your review! I'm learning a lot from your and @wchargin's comments. Let me know if there's anything else.

decentralion:
Looks good! Merging!

PR #1038, opened 2019-01-04
Replace invocations of `npm` with `yarn`

decentralion:
Currently, our underlying test script uses npm rather than yarn to
execute the tests. This is awkward, because we use yarn everywhere else
in lieu of npm. It turns out that some setups have node available
without npm, and in such environments our tests fail with a cryptic
ENOENT error.

This changes the tests to use yarn instead.

Thanks to @wpank for help uncovering this issue.

### DIFF
--- a/config/test.js
+++ b/config/test.js
@@ -54,40 +54,31 @@ function makeTasks(mode /*: "BASIC" | "FULL" */) {
     },
     {
       id: "check-pretty",
-      cmd: ["npm", "run", "--silent", "check-pretty"],
+      cmd: ["yarn", "run", "--silent", "check-pretty"],
       deps: [],
     },
     {
       id: "lint",
-      cmd: ["npm", "run", "--silent", "lint"],
+      cmd: ["yarn", "run", "--silent", "lint"],
       deps: [],
     },
     {
       id: "flow",
-      cmd: [
-        "npm",
-        "run",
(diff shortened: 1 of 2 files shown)


decentralion:
It was cryptic in that many tests would pass, but some of them would fail ENOENT. E.g. `backend` would fail, but `flow` would pass. Looking at this from understanding the code, it's clear that this is because some tasks directly invoked npm and others didn't. But looking at it from the outside, it seemed plausible that there was some other obscure configuration issue causing certain tasks to fail in a way that was directly related to the domain complexity of that individual test.

If you want to keep using `npm` rather than `yarn`, we could just add another mini script in the style of `check-gnu-coreutils` for verifying that npm is available. (`check-npm-available`?)

wchargin:
> It was cryptic in that many tests would pass, but some of them would
> fail ENOENT.

Got it; that makes sense.

> If you want to keep using `npm` rather than `yarn`

It would seem strange to require that you have a package manager that we
don’t actually use for its package management capabilities, so no, I’m
not going to argue for this. I’m fine with the patch; removing the `--`s
is the only thing that I see that should be changed.

wchargin:
Or, for perhaps a more familiar usage: `bazel run //tensorboard -- --logdir /tmp/logs`. :-)

PR #1043, opened 2019-01-09
Add documentation for `NodeType` and `EdgeType`

decentralion:
This commit adds some docstrings for the concepts of NodeType and
EdgeType. I also swapped the order so that `NodeType` comes first,
which is more consistent with usage across the codebase.

This commit makes no changes to the actual code; the only effects
are re-organization and documentation.

Test plan: `yarn test` && human inspection

### DIFF
--- a/src/analysis/types.js
+++ b/src/analysis/types.js
@@ -2,17 +2,95 @@
 
 import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
 
-export type EdgeType = {|
-  +forwardName: string,
-  +backwardName: string,
-  +defaultForwardWeight: number,
-  +defaultBackwardWeight: number,
-  +prefix: EdgeAddressT,
+/**
+ * This module defines `NodeType`s and `EdgeType`s, both of which are
+ * data structures containing shared metadata that describes many nodes or edges
+ * in the graph. Nodes can be "members" of zero or more `NodeType`s, and edges can
+ * be "members" of zero or more `EdgeType`s. Membership is determined based on the
+ * type's `prefix`, which is an address. A node or edge is considered a member of
+ * a type if that type's prefix is a prefix of the node's address.
+ *
+ * To make this more concrete, let's consider a specific example. Suppose we define
+ * the following nodes:
+ *
+ * const pullNode = NodeAddress.fromParts(["github", "pull", "repo", "34"]);
+ * const commitNode = NodeAddress.fromParts(["git", "commit", "e1337"]);
+ * const pullType: NodeType = {
+ *  name: "Pull request",
+ *  prefix: NodeAddress.fromParts(["github", "pull"]),
+ *  // ... more properties as required
+ * };
+ * const githubType: NodeType = {
+ *   name: "GitHub node",
+ *   prefix: NodeAddress.fromParts(["github"])
+ * };
+ *
+ * Then the `pullNode` is considered a member of the `pullType` and `githubType`,
+ * while the `commitNode` is not a member of either type.
+ *
+ * The concept of a "type" is useful to SourceCred because it let's us express
+ * that groups of nodes are conceptually related, and that we should treat them similarly.
+ * Most saliently, we use types to assign default weights to groups of nodes, and to
+ * aggregate them for better UI organization. The fact that the SourceCred UI can group
+ * all pull requests together, and assign a default weight to all of them, is possible
+ * because the GitHub plugin defines a pull request node type.
+ *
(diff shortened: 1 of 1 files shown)


src/analysis/types.js
```diff
@@ -2,17 +2,70 @@
 
 import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
 
+/**
+ * Represents a "Type" of node in the graph.
+ *
+ * An node is considered a member of this type if
```
anthrocypher: "A node"

decentralion: Fixed, thanks.

src/analysis/types.js
```diff
@@ -2,17 +2,70 @@
...
+ * its address has this type's prefix as an address prefix,
```
anthrocypher: While I think I can deduce what you're saying, the repetition of words makes the pieces of functionality you're referring to a bit obscure. Perhaps edit to say "A Node is considered a member of NodeAddressT if..." and include a link to `Address.hasPrefix` semantics (which, I believe, is here: https://github.com/sourcecred/sourcecred/blob/83fa29688ed97c996212fe8d02abe7d2a32ac4c1/src/core/graph.js#L299)

src/analysis/types.js
```diff
@@ -2,17 +2,70 @@
...
+export type NodeType = {|
+  // The name for an individual node of this type.
+  // For example, for the GitHub PULL_REQUEST node type, the name is "Pull
+  // request". The first letter of the name should be capitalized.
+  +name: string,
+  // The name for multiple nodes of this type.
+  // For example, for the GitHub PULL_REQUEST node type, the pluralName is
+  // "Pull requests". The first letter of this name should be capitalized.
+  +pluralName: string,
+  // The address prefix that will be used to test whether a node is a member
```
anthrocypher: Might be nice to specify here and bring it back around for your reader: "`NodeAddressT` is the address prefix that will be used to test whether a node is a member of this NodeType".

src/analysis/types.js
```diff
@@ -2,17 +2,70 @@
...
+ *
+ * Edge types are used for organization in the UI, and for
+ * applying type-level weights.
+ */
 export type EdgeType = {|
+  // A brief descriptive name of what the "forward" direction of the edge
+  // means. For example, for the GitHub REFERENCES edge type, the forwardName
+  // is "references"
   +forwardName: string,
+  // A brief descriptive name of what the "backward" direction of the edge
```
anthrocypher: I'm not sure what you mean by "backward" direction of the edgeType. Can we say a little more about how this works and why it's important? Or, if it's a really foundational concept, maybe we can spend more time on it in a better medium, like office hours.

src/analysis/types.js
```diff
@@ -2,17 +2,70 @@
...
   +backwardName: string,
+  // The default weight for the forward direction of this edge.
+  // We use `1` as a default value ("of normal importance").
+  // The weights have linear importance, i.e. 2 is twice as important as 1.
   +defaultForwardWeight: number,
+  // The default weight for the backward direction of this edge.
+  // We use `1` as a default value ("of normal importance").
+  // The weights have linear importance, i.e. 2 is twice as important as 1.
   +defaultBackwardWeight: number,
+  // The address prefix that will be used to test whether an edge is a member
```
anthrocypher: Again, seems good to specify: "`EdgeAddressT` is the address prefix that will be used to test whether an edge is a member of this EdgeType"

src/analysis/types.js
decentralion: Thanks for this comment. I realized that I can do a better job explaining how types work, and re-wrote the documentation. Please take another look and tell me what you think.

src/analysis/types.js
decentralion: Rewrote it to be more explicit.

src/analysis/types.js
decentralion: Rewrote it to be more explicit.

src/analysis/types.js
decentralion: It's a pretty foundational concept, it deserves its own documentation elsewhere and then we can link to it here. Created an issue so I don't lose track of this: https://github.com/sourcecred/sourcecred/issues/1049

src/analysis/types.js
anthrocypher: New version is a really solid intro. 👍  Does a great job of presenting first principles in a highly readable form, then showing how they combine/build on each other.

Issue #1020, opened 2018-12-05
[designdoc] Create `PagerankGraph`


decentralion:
The core SourceCred infrastructure exists to:
1. Create Contribution Graphs which represent contributions to a project as
   nodes in the graph, with edges representing relationships between
   contributions
2. Run a (PageRank-derived) credit attribution algorithm on those graphs,
   assigning a score to every node in the graph. (Normalization is also
   optionally applied, so that e.g. the users' scores can sum to 1000.)

For 1., we have the [Graph class] which provides an object-oriented interface
for creating graphs. The graph has a thoughtful interface that supports
common operations like filtering by prefix, or exploring the 'neighborhood' of
a node. The graph class also provides utility functionality, notably the
ability to (de)serialize graphs, check them for equality, copy them, etc.

[Graph class]: https://github.com/sourcecred/sourcecred/blob/bbe773bb67ef6132554335c643d676fb9d4fef9d/src/core/graph.js#L71

For 2., we have a variety of low-level functions which transform a Graph into a
markov chain, and support taking running PageRank steps on that markov chain.
We also have a [pagerank invocation] that takes the graph, along with
non-serializable arguments like an evaluator function, and outputs a
non-serializable `PagerankNodeDecomposition.` The `PagerankNodeDecomposition`
provides score information, in terms of both the nodes in the graph, and the
`Connection`s in the markov chain.

[pagerank invocation]: https://github.com/sourcecred/sourcecred/blob/bbe773bb67ef6132554335c643d676fb9d4fef9d/src/analysis/pagerank.js#L46

In contrast to 1., 2. has a messy interface. The `Connection`s and `ScoredConnection`s
are new concepts that mostly overlap with edges; there's no support for serializing
the inputs and outputs, the API is split across multiple files in a subfolder.

The non-serializability of the inputs and outputs to `pagerank` are a blocking
issue for #967. We could just write a serializer for
`PagerankNodeDecomposition`, which was prototyped in #995, or we could slightly
modify the interface so that we serialize different data, which is discussed at
length in #1004. However, having spent time exploring implementation of these
alternatives (cf #1008, #1009, #1010, #1012), I think we
should instead re-imagine how we handle step 2 entirely.

I have the following goals in mind:

1. The API for running PageRank should be clean, and easy to depend on. It
   should live in `core/`, rather than in `analysis/` (and it should merit
   inclusion in `core/`) 
2. The inputs and outputs to PageRank should both be serializable.
[...]

PR #1053, opened 2019-01-17
Report Markov Chain convergence statistics

decentralion:
This commit modifies `markovChain.findStationaryDistribution` so that
in addition to returning the final distribution, it also provides
information on how many iterations were required to get that
distribution, and on what the final convergence delta was.

This is motivated by the proposed API for the new PagerankGraph (see
[#1020]). Also, I think it makes a nice addition to the test code.

Test plan:
Unit tests are updated, and `yarn test` passes.

[#1020]: https://github.com/sourcecred/sourcecred/issues/1020

Thanks to @BrianLitwin for semi-pair-programming it

### DIFF
--- a/src/core/attribution/markovChain.js
+++ b/src/core/attribution/markovChain.js
@@ -7,6 +7,17 @@
  */
 export type Distribution = Float64Array;
 
+export type StationaryDistributionResult = {|
+  // The final distribution after attempting to find the stationary distribution
+  // of the Markov chain.
+  +distribution: Distribution,
+  // How many steps were taken in this convergence attempt.
+  +nIterations: number,
+  // The maximum delta between entries in the distribution as of the final
+  // markov chain step. If no steps were taken, convergenceDelta will be NaN.
+  +convergenceDelta: number,
+|};
+
 /**
  * A representation of a sparse transition matrix that is convenient for
  * computations on Markov chains.
@@ -102,12 +113,16 @@ function* findStationaryDistributionGenerator(
   chain: SparseMarkovChain,
   options: {|
     +verbose: boolean,
+    // A distribution is considered stationary if the action of the Markov
+    // chain on the distribution does not change any component by more than
+    // `convergenceThreshold` in absolute value.
     +convergenceThreshold: number,
+    // We will run maxIterations markov chain steps at most.
     +maxIterations: number,
   |}
-): Generator<void, Distribution, void> {
-  let pi = uniformDistribution(chain.length);
-  let scratch = new Float64Array(pi.length);
+): Generator<void, StationaryDistributionResult, void> {
+  let distribution = uniformDistribution(chain.length);
+  let scratch = new Float64Array(distribution.length);
   function computeDelta(pi0, pi1) {
     let maxDelta = -Infinity;
     // Here, we assume that `pi0.nodeOrder` and `pi1.nodeOrder` are the
@@ -118,26 +133,27 @@ function* findStationaryDistributionGenerator(
     });
     return maxDelta;
   }
-  let iteration = 0;
+  let nIterations = 0;
+  let convergenceDelta = NaN;
   while (true) {
-    if (iteration >= options.maxIterations) {
+    if (nIterations >= options.maxIterations) {
       if (options.verbose) {
-        console.log(`[${iteration}] FAILED to converge`);
+        console.log(`[${nIterations}] FAILED to converge`);
       }
-      return pi;
+      return {distribution, nIterations, convergenceDelta};
     }
-    iteration++;
-    sparseMarkovChainActionInto(chain, pi, scratch);
-    const delta = computeDelta(pi, scratch);
-    [scratch, pi] = [pi, scratch];
+    nIterations++;
+    sparseMarkovChainActionInto(chain, distribution, scratch);
+    convergenceDelta = computeDelta(distribution, scratch);
+    [scratch, distribution] = [distribution, scratch];
     if (options.verbose) {
-      console.log(`[${iteration}] delta = ${delta}`);
+      console.log(`[${nIterations}] delta = ${convergenceDelta}`);
     }
-    if (delta < options.convergenceThreshold) {
+    if (convergenceDelta < options.convergenceThreshold) {
       if (options.verbose) {
-        console.log(`[${iteration}] CONVERGED`);
+        console.log(`[${nIterations}] CONVERGED`);
       }
-      return pi;
+      return {distribution, nIterations, convergenceDelta};
     }
     yield;
   }
@@ -154,7 +170,7 @@ export function findStationaryDistribution(
     +maxIterations: number,
     +yieldAfterMs: number,
   |}
-): Promise<Distribution> {
+): Promise<StationaryDistributionResult> {
   let gen = findStationaryDistributionGenerator(chain, {
     verbose: options.verbose,
     convergenceThreshold: options.convergenceThreshold,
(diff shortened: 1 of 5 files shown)


src/core/attribution/markovChain.js
```diff
@@ -118,26 +132,27 @@ function* findStationaryDistributionGenerator(
...
+      return {distribution, nIterations, convergenceDelta};
     }
-    iteration++;
-    sparseMarkovChainActionInto(chain, pi, scratch);
-    const delta = computeDelta(pi, scratch);
-    [scratch, pi] = [pi, scratch];
+    nIterations++;
+    sparseMarkovChainActionInto(chain, distribution, scratch);
+    convergenceDelta = computeDelta(distribution, scratch);
+    [scratch, distribution] = [distribution, scratch];
```
wchargin: I’m not going to request that this be reverted, but IMHO `pi` is much
clearer than `distribution`. _Both_ `scratch` and `pi` are
distributions, but one is scratch space and the other is our
current-best approximation of the stationary distribution.

src/core/attribution/markovChain.js
```diff
@@ -102,12 +113,15 @@ function* findStationaryDistributionGenerator(
   chain: SparseMarkovChain,
   options: {|
     +verbose: boolean,
+    // The chain is considered converged once the maximum delta in the
```
wchargin: This is at best unclear. What does “maximum delta in the distribution”
mean? It suggests “the largest difference between any two components in
the distribution”, but of course that is not what is meant.

Suggest: “A distribution is considered stationary if the action of the
Markov chain on the distribution does not change any component by more
than `convergenceThreshold` in absolute value.”

src/core/attribution/markovChain.js
```diff
@@ -7,6 +7,17 @@
  */
 export type Distribution = Float64Array;
 
+export type StationaryDistributionResult = {|
+  // The final distribution after attempting to find the stationary distribution
```
wchargin: Please capitalize “Markov chain” and end these sentences with periods.

src/core/attribution/markovChain.test.js
```diff
@@ -174,40 +185,61 @@ describe("core/attribution/markovChain", () => {
...
         yieldAfterMs: 1,
       });
-      expectStationary(chain, pi);
+
+      expect(result.convergenceDelta).toBeLessThanOrEqual(1e-7);
+      expect(result.nIterations).toBeLessThanOrEqual(255);
+
+      // This should be wholly deterministic, so having an inline snapshot
+      // should not produce test flakes.
+      // This value should not be updated without a good reason (i.e. we
```
wchargin: `s/i.e./e.g.,/`

src/core/attribution/markovChain.test.js
```diff
@@ -174,40 +185,61 @@ describe("core/attribution/markovChain", () => {
...
+      // realized that our convergence semantics were wrong)
+      expect(result.convergenceDelta).toMatchInlineSnapshot(
+        `9.536743161842054e-8`
```
wchargin: Curious on your thought process when writing these snapshot tests. Do
they seem “totally reasonable”, or are they “uncomfortable but better
than nothing”, or something else?

I agree that they’re not _flaky_ in the sense that the code is
deterministic. But they’re certainly brittle: valid mathematical
transformations to the implementation may change the results. For
example, changing `sparseMarkovChainInto` to iterate over the vertices
in reverse order will yield different values:

    > 0.1 + 0.2 + 0.3 == 0.3 + 0.2 + 0.1
    false

(For this particular set of tests, I was careful to find graphs that
were both nontrivial _and_ had transition matrices all of whose nonzero
entries are small integer multiples of reciprocal powers of two, so this
particular problem shouldn’t arise, but I hope that the point stands.)

Suppose that I make a change that causes this test to fail: the expected
convergence delta changes from `9.54e-8` to `9.58e-8` and the number of
iterations increases by 1. A priori, this could be a genuine off-by-one
error, or it could just be due to the fact that floating-point numbers
do not model the theory of reals. How will this test help me determine
whether I’ve actually made a mistake?

src/core/attribution/markovChain.js
```diff
@@ -7,6 +7,17 @@
...
+  // of the markov chain
+  +distribution: Distribution,
+  // How many steps were taken in this convergence attempt
+  +nIterations: number,
+  // The maximum delta between entries in the distribution as of the final
+  // markov chain step. If no steps were taken, convergenceDelta will be NaN
```
wchargin: This doesn’t seem right. What I care is the loss for the _returned
distribution_ under the action, not the _previous_ distribution. If you
return $x$, the delta should be $\norm{A x - x}_\infty$, which is well
defined even if $x$ was initialized to the uniform distribution and
never modified. No?

wchargin:
I can see why clients might want to know the final ∞-norm value: that’s
a property of the r_result_ that describes its accuracy. But why would
they want to know the number of iterations? That’s an implementation
detail, and I don’t see how it’s useful.

src/core/attribution/markovChain.js
decentralion: Done.

src/core/attribution/markovChain.js
decentralion: Done.

src/core/attribution/markovChain.test.js
decentralion: My rationale for the snapshot tests was that they would act as a sentinel that the behavior had changed at all. But I agree they aren't actually that valuable, so I've removed them.

src/core/attribution/markovChain.test.js
decentralion: Removed.

src/core/attribution/markovChain.js
decentralion: Ack.

src/core/attribution/markovChain.js
decentralion: I don't understand, can you please explain more simply? What is the `A` in `A x - x`, where does the sub infinity factor in?

wchargin: Sorry, was writing in a bit of shorthand, forgot to expand.

$A$ is the Markov chain transition matrix. If $x$ is a distribution
vector, then $A x$ is the distribution after one step in the Markov
chain. (`sparseMarkovChainAction(A, x)` computes `A x`.)

$\norm{\delta}_\infty$ is the [∞-norm][un] of $\delta$; equivalently,
the largest value in $\delta$.

Basically, if $x$ is actually a stationary distribution for the Markov
chain given by $A$, then $A x = x$ by definition. So $A x - x$ is a
vector that will be “smaller” when $x$ is a better approximation of a
stationary distribution. That is, $\norm{A x - x}$ is the loss function
that we’re trying to minimize. In particular, the norm that we choose is
the max-norm, though it would also be valid to minimize another norm
like the Euclidean (sum-of-squares) norm.

The purpose of returning the `convergenceDelta` (i.e., loss) to the
client is so that the client can get an idea of how good the returned
distribution is. But as documented and implemented, you’re returning the
final distribution but the loss corresponding to the penultimate
distribution. My claim is that the returned loss should actually be the
loss of the distribution that you return, not of a different
distribution.

[un]: https://en.wikipedia.org/wiki/Uniform_norm

decentralion:
Got it, thanks. I've updated the PR per your feedback. I also removed the nIterations return value, I was thinking it's useful to have for calibrating the max iterations setting, but you can actually set verbose=True already for doing that.

decentralion:
Ok, I'm going to merge this on self-review, (discussed with @wchargin that I was going to start doing this when I feel confident in the state of the code).

I've incorporated @wchargin's suggestion that we change the semantics of the returned convergence delta so that it describes the returned distribution, rather than the previous distribution. To be consistent with this behavior, we now return a slightly less-converged distribution than previously, i.e. we return the first distribution that would have a sufficiently low convergence delta, rather than the second distribution that would be converged. This is why the snapshotted values have slightly changed.

PR #1056, opened 2019-01-20
Fix markov chain semantics for empty graphs

decentralion:
While working on #1020, I discovered that attempting to find the
stationary distribution on an empty graph currently throws an error.
This is awkward, since the empty graph is a totally reasonable graph for
our code to work on.

This commit modifies the markov chain semantics so that we can consider
empty chains (which have no probability mass). Attempting to find the
stationary distribution of one immediately returns an empty
distribution, after 0 iteration steps and with a convergence delta of
`NaN`.

Test plan: Unit tests included; `yarn test` should suffice.

### DIFF
--- a/src/analysis/pagerank.js
+++ b/src/analysis/pagerank.js
@@ -58,13 +58,16 @@ export async function pagerank(
     fullOptions.selfLoopWeight
   );
   const osmc = createOrderedSparseMarkovChain(connections);
-  const distribution = await findStationaryDistribution(osmc.chain, {
+  const distributionResult = await findStationaryDistribution(osmc.chain, {
     verbose: fullOptions.verbose,
     convergenceThreshold: fullOptions.convergenceThreshold,
     maxIterations: fullOptions.maxIterations,
     yieldAfterMs: 30,
   });
-  const pi = distributionToNodeDistribution(osmc.nodeOrder, distribution);
+  const pi = distributionToNodeDistribution(
+    osmc.nodeOrder,
+    distributionResult.distribution
+  );
   const scores = scoreByConstantTotal(
     pi,
     fullOptions.totalScore,
--- a/src/analysis/pagerankNodeDecomposition.test.js
+++ b/src/analysis/pagerankNodeDecomposition.test.js
@@ -7,7 +7,10 @@ import {
   createOrderedSparseMarkovChain,
 } from "../core/attribution/graphToMarkovChain";
 import {findStationaryDistribution} from "../core/attribution/markovChain";
-import {decompose} from "./pagerankNodeDecomposition";
+import {
+  decompose,
+  type PagerankNodeDecomposition,
+} from "./pagerankNodeDecomposition";
 import * as MapUtil from "../util/map";
 
 import {advancedGraph} from "../core/graphTestUtil";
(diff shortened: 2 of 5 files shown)


wchargin:
It is not at all clear to me that these semantics _do_ make sense (which
is why I originally wrote them to throw an error). Can you elaborate on
the motivation for this change, and why you think that these semantics
are correct?

decentralion:
@wchargin: Please see [this test case][1].

If you don't think the semantics should be modified in this file, it would be easy for me to simply special case the empty graph at the PagerankGraph layer.

[1]: https://github.com/sourcecred/sourcecred/pull/1057/files#diff-7b94c16cf1633f3f8594196b1dce1e49R166

wchargin:
I see that test case, but it doesn’t really convince me at all. You’re
saying that it’s somehow natural/useful/meaningful to say that the
convergence delta of an empty graph is `NaN` and that it ran for zero
iterations? Any change in which a previously guaranteed-non-`NaN` value
is permitted to become `NaN` hardly seems natural to me!

Even “ran for zero iterations” is only tautologically true. Usually, we
say that the computation ran for *k* iterations if after *k* iterations
it converged or hit the iteration cap. But here neither of those is the
case: we can’t even assess whether it converged because its convergence
is not well defined. It “runs for zero iterations” because that’s what
you’ve chosen to be the behavior for an empty chain—why?

(And all this beside the fact that the resulting “distribution” doesn’t
sum to 1…)

If I take my bicycle to the gas station and ask for the full-service
treatment, and that they should please fill the tank to at least 90% in
no more than an hour, they shouldn’t say “we got it NaN% full in zero
hours flat”—they should say, “pardon me, sir, this is a bike”.

decentralion:
Your bicycle at the gas station made me chuckle out loud, thanks for that.

I see your point that there isn't really any coherent justification for returning 0 and NaN. What semantics would you think are best? Do you think we should just throw an error when calling `runPagerank` on an empty graph?

decentralion:
I've discussed this offline with @wchargin, and decided not to merge it.

wchargin:
To roughly serialize from our discussion offline:

> I see your point that there isn't really any coherent justification
> for returning 0 and NaN. What semantics would you think are best? Do
> you think we should just throw an error when calling `runPagerank` on
> an empty graph?

Yes. I don’t think that empty graphs are part of the valid domain for
`runPagerank` (because there is no Markov chain on an empty graph), so I
think that throwing an error is the right behavior.

> I was actually basing the original design on your "empty values should
> not need exceptional handling" heuristic, and returning 0 and NaN
> seemed like a less exceptional option than throwing an error. But
> maybe I am misreading the intent of your heuristic.

I think that you’re reading the _intent_ correctly, but I don’t think
that that heuristic applies here. The heuristic is useful when the empty
values fit in with the rest of the domain, usually by means of some nice
compositional properties. A hint that this isn’t the case here is that
the domain of the `convergenceDelta` needs to be broadened from
“non-negative reals” to “non-negative reals plus `NaN`”, which is not a
good sign!

It also bears noting that this is just one API design heuristic. It is
often a useful one, but it is not a rule to be followed at all costs.
For most nontrivial APIs, the heuristics won’t all pull you in the same
direction—weighing their relative merits is part of the fun!

PR #1078, opened 2019-01-28
Readme: correct node link

BrianLitwin:
There's a bug in #1076 where  the Node link at
the bottom of the Readme catches the Node
link reference created earlier in the Readme and
links to 'https://nodejs.org/en/' instead of
'https://github.com/nodejs/node`.

Decided to put the link inline so we
could keep the word "Node" linked, instead
of using "Node.js", which would suggest
the organization instead of the GitHub project.

Test Plan:
Testing that these links work in the live ReadMe on my
fork, instead of copy/pasting the changed bits into
a Markdown parser.


BrianLitwin:
Niiice @wpank will update!

BrianLitwin:
Updated! 🚀

PR #1075, opened 2019-01-25
Changelog: add rocket and eyes reaction types

BrianLitwin:
Test Plan:
Make sure the pull request number is correct

PR #1082, opened 2019-02-07
Fix a build break due to merge conflicts

decentralion:
PR #1075 added a new EdgeType, and #1080 added a new field to EdgeTypes.
Both PRs merged and this broke the build.

This very trivial commit fixes the build breakage in a noncontroversial
way (copies the placeholder edge description used for every other edge
over).


decentralion:
Since this is trivial, and a build fix, I am going to merge this on self review.

PR #1083, opened 2019-02-07
Fill in edge type descriptions

decentralion:
Pull #1080 added in a description field for edge types, but put in a
placeholder message for each actual description. This pull adds in
descriptions for each edge type.

Test plan: `yarn test` passes, and additionally
`git grep 'TODO: Add a description for this edge type'` returns no hits.

### DIFF
--- a/src/plugins/github/declaration.js
+++ b/src/plugins/github/declaration.js
@@ -78,8 +78,9 @@ const authorsEdgeType = Object.freeze({
   defaultBackwardWeight: 1,
   prefix: E.Prefix.authors,
   description: dedent`\
-    An authors edge connects a GitHub userlike account to a post\
-    that they authored. Examples of posts include issues, pull requests, and comments.
+    Connects a GitHub account to a post that they authored.
+
+    Examples of posts include issues, pull requests, and comments.
   `,
 });
 
@@ -90,7 +91,8 @@ const hasParentEdgeType = Object.freeze({
   defaultBackwardWeight: 1 / 4,
   prefix: E.Prefix.hasParent,
   description: dedent`\
-    A hasParent edge connects a GitHub entity to its child entities.
+    Connects a GitHub entity to its child posts.
+
     For example, a Repository has Issues and Pull Requests as children, and a
     Pull Request has comments and reviews as children.
   `,
@@ -103,8 +105,7 @@ const mergedAsEdgeType = Object.freeze({
   defaultBackwardWeight: 1,
   prefix: E.Prefix.mergedAs,
   description: dedent`\
-    A mergedAs edge connects a GitHub pull request to the commit that it merges,
-    assuming that the pull request has merged.
+    Connects a GitHub pull request to the Git commit that it merges.
   `,
 });
 
@@ -115,10 +116,11 @@ const referencesEdgeType = Object.freeze({
   defaultBackwardWeight: 0,
   prefix: E.Prefix.references,
   description: dedent`\
-    A references edge connects a GitHub post to a GitHub entity that it
-    references. For example, if you write a GitHub issue comment that says
-    "thanks @username for pull #1337", it will create references edges to both
-    the user @username, and to pull #1337 in the same repository.
+    Connects a GitHub post to an entity that it references.
+
+    For example, if you write a GitHub issue comment that says "thanks
+    @username for pull #1337", it will create references edges to both the user
+    @username, and to pull #1337 in the same repository.
   `,
 });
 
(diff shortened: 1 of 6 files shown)


src/plugins/github/declaration.js
```diff
@@ -113,7 +129,14 @@ const mentionsAuthorEdgeType = Object.freeze({
...
   prefix: E.Prefix.mentionsAuthor,
-  description: "TODO: Add a description for this EdgeType",
+  description: dedent`\
+    When a post in a particular thread mentions a particular user, that post
+    will have a \`MentionsAuthorEdge\` connecting it to any posts written by the
+    mentioned user. The intuition is that if a post is mentioning an author by name,
+    their contributions in that thread are probably particularly valuable.
+
+    This is an experimental feature and may be removed in a future version of SourceCred.
+  `,
```
BrianLitwin: Maybe just revisit this description.. the word 'particular' is repeated 3x. 

"When a post in a thread mentions a user, that post will have a \`MentionsAuthorEdge\` connecting it to every post by that user in that particular thread". Idk if that's any clearer.

BrianLitwin:
A general note is that there is a lot of camel case in the descriptions.. do you think that will be consistent with the way the edges are represented in the UI?

src/plugins/github/declaration.js
```diff
@@ -123,7 +146,9 @@ const reactsHeartEdgeType = Object.freeze({
   // TODO(#811): Probably change this to 0
   defaultBackwardWeight: 1 / 32,
   prefix: E.Prefix.reactsHeart,
-  description: "TODO: Add a description for this EdgeType",
+  description: dedent`\
+    A reactsHeart edge connects users to posts that they gave a ❤️ reaction to.
```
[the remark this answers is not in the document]
decentralion: I've updated with your suggestion.

src/plugins/github/declaration.js
```diff
@@ -85,7 +89,11 @@ const hasParentEdgeType = Object.freeze({
   defaultForwardWeight: 1,
   defaultBackwardWeight: 1 / 4,
   prefix: E.Prefix.hasParent,
-  description: "TODO: Add a description for this EdgeType",
+  description: dedent`\
+    A hasParent edge connects a GitHub entity to its child entities.
```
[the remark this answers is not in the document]
decentralion: Good point, I changed it per your suggestion.

src/plugins/github/declaration.js
```diff
@@ -76,7 +77,10 @@ const authorsEdgeType = Object.freeze({
   defaultForwardWeight: 1 / 2,
   defaultBackwardWeight: 1,
   prefix: E.Prefix.authors,
-  description: "TODO: Add a description for this EdgeType",
+  description: dedent`\
+    An authors edge connects a GitHub userlike account to a post\
```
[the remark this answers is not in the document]
decentralion: Good point, I think "GitHub account" is sufficiently clear.

decentralion:
> A general note is that there is a lot of camel case in the descriptions.. do you think that will be consistent with the way the edges are represented in the UI?

Good catch--the camecasing is a code convention, but the UI never uses camelcasing. So, it will be more consistent to eschew camelcasing here. I've re-written the text based on this suggestion, PTAL @BrianLitwin.

src/plugins/github/declaration.js
```diff
@@ -85,7 +90,12 @@ const hasParentEdgeType = Object.freeze({
   defaultForwardWeight: 1,
   defaultBackwardWeight: 1 / 4,
   prefix: E.Prefix.hasParent,
-  description: "TODO: Add a description for this EdgeType",
+  description: dedent`\
+    Connects a GitHub entity to its child posts.
```
wchargin: Child posts only? Would a label not be a child of a repository?

src/plugins/github/declaration.js
```diff
@@ -112,7 +130,15 @@ const mentionsAuthorEdgeType = Object.freeze({
   defaultForwardWeight: 1,
   defaultBackwardWeight: 0,
   prefix: E.Prefix.mentionsAuthor,
-  description: "TODO: Add a description for this EdgeType",
+  description: dedent`\
+    Connects a post which mentions a user to posts in the same thread that
```
wchargin: nit:s/which/that/

src/plugins/github/declaration.js
decentralion: Updated for future proofing (although right now it's just posts).

src/plugins/github/declaration.js
decentralion: Fixed.

Issue #1023, opened 2018-12-12
Error while processing tensorflow/tensorflow repository


Dim25:
I was able to run `node bin/sourcecred.js load sourcecred/sourcecred` successfully, but when I try to process the [tensorflow](https://github.com/tensorflow/tensorflow) repository with `node bin/sourcecred.js load tensorflow/tensorflow` I get the following error:

> Starting tasks
>   GO   load-git
>   GO   load-github
>  DONE  load-git
>  FAIL  load-github
> Exit code: 1
> Contents of stderr:
>     Error: Cannot update data for nonexistent node: undefined
>         at Mirror._nontransactionallyUpdateOwnData (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:1081:415)
>         at Mirror._nontransactionallyUpdateData (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:885:489)
>         at _inTransaction (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:890:605)
>         at _inTransaction (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:1324:142)
>         at Mirror._updateStep (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:890:522)
>         at process.internalTickCallback (internal/process/next_tick.js:77:7)
> 
> 
> Full results
>  DONE  load-git
>  FAIL  load-github
> Exit code: 1
> Contents of stderr:
>     Error: Cannot update data for nonexistent node: undefined
>         at Mirror._nontransactionallyUpdateOwnData (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:1081:415)
>         at Mirror._nontransactionallyUpdateData (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:885:489)
>         at _inTransaction (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:890:605)
>         at _inTransaction (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:1324:142)
>         at Mirror._updateStep (/home/diver/in_work/sourcecred/sourcecred/bin/sourcecred.js:890:522)
>         at process.internalTickCallback (internal/process/next_tick.js:77:7)
> 
> 
> Overview
> Failed tasks:
>   - load-github
> Final result:  FAILURE 
> 

Node v11.1.0
Yarn 1.10.1
[...]

Issue #996, opened 2018-11-02
Known contract violation errors in the GitHub GraphQL API


wchargin:
This issue documents instances in which the GitHub GraphQL API specifies
that it returns objects of certain types, but in fact returns objects of
the wrong type (and lies about the objects’ actual types).

Depending on what fields exist on the relevant types, queries can also
sometimes crash or return invalid data.

SourceCred developers should know about these because we depend on the
GitHub API. We should try to avoid querying them when possible. This may
not always be possible—we definitely need `Reaction.user`, for instance.
But we can probably avoid requesting `IssueComment.issue` even if it may
be marginally convenient in some case.

Will post as individual comments for direct linking.

Table of contents:

  - [`Reaction.user`]
  - [`GitActor.user`]
  - [`IssueComment.issue`]
  - [`Release.author`]

[`Reaction.user`]: https://github.com/sourcecred/sourcecred/issues/996#issuecomment-435495422
[`GitActor.user`]: https://github.com/sourcecred/sourcecred/issues/996#issuecomment-435495461
[`IssueComment.issue`]: https://github.com/sourcecred/sourcecred/issues/996#issuecomment-435495537
[`Release.author`]: https://github.com/sourcecred/sourcecred/issues/996#issuecomment-456718463

wchargin:
## `Reaction.user`

The [`Reaction` object] has a field `user` of type `User`. In fact,
sometimes the returned objects are of type `Organization`. This can
occur if a user reacts to a post and then converts their account to an
organization: existing reactions are not deleted, and the GraphQL API
returns the wrong result.

Steps to reproduce:

  - Create new accounts **A** and **B**.
  - On account **A**, react to some issue, issue comment, etc.
  - Transform account **A** to an organization owned by **B**.
  - Use the GraphQL API to request the reactions for the post in
    question, and select `nodes { user { __typename id } }`. The result
    will have `__typename: "User"`, but the `id` will correspond to an
    organization ID: re-fetching with `node(id: …) { __typename }` will
    yield `Organization`.

Sample query:

```graphql
{
  reactionNode: node(id: "MDg6UmVhY3Rpb24yMTY3ODkyNQ==") {
    __typename
    ... on Reaction {
      user {
        __typename
        id
        login
      }
    }
  }
  userNode: node(id: "MDEyOk9yZ2FuaXphdGlvbjE3OTUyOTI1") {
    __typename
    id
  }
}
```

```json
{
  "data": {
    "reactionNode": {
[...]

Issue #997, opened 2018-11-03
Don’t fetch GitHub commits recursively


wchargin:
Summary:

  - Don’t fetch `parents` on GitHub `Commit` objects.
  - Instead, identify Git commit authors with GitHub users manually.
  - This makes us more resilient to schema changes, and makes our
    performance more predictable.
  - This makes us way faster on repositories that do not use pull
    requests as discussed in #920 and #923.
  - This is more portable to other Git hosting providers, like GitLab or
    Phabricator or an internal/private host.
  - This ties in nicely to solving identity resolution, which we also
    need to do in the short-to-medium term.
  - This gives us better data anyway (GitHub’s user detection is quite
    limited; Git has tools to do better: see #921).

Currently, whenever we find a GitHub `Commit` object, we fetch its
parents. Transitively, this means that we fetch the full commit graph
reachable from any root commits that we know about.

As we fetch more data from GitHub, this has the potential to
accidentally pull in a lot of commits that we do not want. For instance,
suppose that we start fetching issue and pull request timeline items,
because we want to see when labels are added and removed. `LabeledEvent`
and `UnlabeledEvent` are two kinds of timeline items, but so are
`Commit`s. We would have to be very careful to not transitively pull in
any references from commits in arbitrary repositories that reference our
issues and pull requests, lest we start downloading their entire commit
graphs!

The primary reason that we fetch `Commit` objects from GitHub at all is
that we want the `author.user` field, so that we can attribute commits
to GitHub users even when the commits are not merged via pull request.
This is a valuable user-level feature, and we should not get rid of it.

But we can implement this feature without having to fetch GitHub commit
objects: we can get the data from Git directly. A Git commit is
associated with a GitHub user via the email address in the commit. We
can simply perform this resolution ourselves, and in fact we will be
able to do a better job than GitHub by using Git’s `.mailmap` (see #921
for details).

Depending on how we implement this, we might want to have “Git
identities” as first-class entities that are explicitly linked to
“GitHub users”. In such a case, we would want to make sure to get the
semantics for identity resolution right. That is something that we
certainly will need to do eventually (e.g., once we add a Discourse
plugin, we’ll certainly need to identify GitHub users with Discourse
users). But if we need to do this in a pinch, we could probably
implement it as a plugin that directly creates GitHub authorship edges
from Git commits to GitHub authors, sidestepping the more fundamental
problem for now.

Another obvious benefit of not fetching the commit parents is that that
[...]

PR #1084, opened 2019-02-09
Mirror inconsistent typenames

wpank:
### Context
This pull requests aims to fix inconsistent typenames reported by Github's GraphQL API, as described in #996,  #998, and #1023. The crux of the issue revolves around users that transform into organizations, whereby certain nodes reported by Github will provide historical interactions, prior to the transformation, and provide the typename for these interactions as belonging to a user, while still maintaining the same id. 

[This issue](https://github.com/wchargin/ghost-playground/issues/2) and [this issue](https://github.com/casper0001/special-octo-waffle/issues/1) provide further practical examples of queries where Github will report inconsistent typenames. 

---
### Implementation

The following changes follow @wchargin's suggestions in #998. 

This entails adding a `Fidelity` type that is either `FAITHFUL` or `UNFAITHFUL` to `NodeFieldType` and `NestedFieldType`. Given the current github schema, these are the only places that exist an `Actor` type that can have inconsistent typenames.  

The schema for `objects` has been changed, which requires upgrading the mirror module to v4. An object's `typename` is allowed to be `null`. The reasoning behind this is that the object of questionable fidelity will have an inconsistent `typename` until a global node lookup is made. This should solely apply to anything that implements the `Actor` interface.

In order to initiate a global node lookup, `_findOutdated` has been updated to include a `typenames` `QueryPlan`. The result of `typenames` is a SQL query for all object ids that have a `null` `typename`. The SQL query for `objects` has been updated to filter out any object with a `null` `typename`.

`_queryFromPlan` has been updated to account for the `typenames` `QueryPlan`, listed above. This includes a new `typenamesLimit` variable, which is an option of how many ids per page to include in pagination. The assumption here is that `_findOutdated` will instead first put all those into a query plan to fetch their typenames first. 

For those `typenames` in the `queryPlan` above, a new function was made to handle generation of the GraphQL query: `queryTypename()`, which will generate a fragment based on the `Actor` union type defined in the schema. The assumption here is that the bug in Github's Graphql api is isolated only to `User`s, `Organization`s, and `Bot`s. 

A new kind of `UpdateResult` as been created to address the kind of changes for querying for typenames: `TypenameUpdateResult`. This consists of a `$ReadOnlyArray` of objects that each have an `id` and `__typename` field. 

`_nontransactionallyUpdateData` has been updated to account for this new kind of `TypenameUpdateResult`. It has been updated to look for a new `TYPENAME_UPDATE` `FIELD_PREFIX`, and if the update in the result is named appropriate, it will pass this to a new function, `_nontransactionallyUpdateTypename`. 

---

### DIFF
--- a/src/graphql/generateFlowTypes.js
+++ b/src/graphql/generateFlowTypes.js
@@ -48,9 +48,6 @@ export default function generateFlowTypes(
     }
   }
   function formatNodeField(field: NodeFieldType): string {
-    if (field.fidelity.type === "UNFAITHFUL") {
-      throw new Error("Unfaithful Fidelity not yet supported");
-    }
     return "null | " + field.elementType;
   }
   function formatConnectionField(field: ConnectionFieldType): string {
--- a/src/graphql/generateFlowTypes.js
+++ b/src/graphql/generateFlowTypes.js
@@ -57,9 +57,6 @@ export default function generateFlowTypes(
     return `$ReadOnlyArray<null | ${field.elementType}>`;
   }
   function formatNestedField(field: NestedFieldType): string {
-    if (field.fidelity.type === "UNFAITHFUL") {
-      throw new Error("Unfaithful Fidelity not yet supported");
-    }
     const eggs = [];
     for (const eggName of Object.keys(field.eggs).sort()) {
       eggs.push({eggName, rhs: formatField(field.eggs[eggName])});
--- a/src/graphql/generateFlowTypes.js
+++ b/src/graphql/generateFlowTypes.js
@@ -48,12 +48,18 @@ export default function generateFlowTypes(
     }
   }
   function formatNodeField(field: NodeFieldType): string {
+    if (field.fidelity.type === "UNFAITHFUL") {
+      throw new Error("Unfaithful Fidelity not yet supported");
+    }
     return "null | " + field.elementType;
   }
   function formatConnectionField(field: ConnectionFieldType): string {
     return `$ReadOnlyArray<null | ${field.elementType}>`;
   }
   function formatNestedField(field: NestedFieldType): string {
+    if (field.fidelity.type === "UNFAITHFUL") {
+      throw new Error("Unfaithful Fidelity not yet supported");
+    }
     const eggs = [];
     for (const eggName of Object.keys(field.eggs).sort()) {
       eggs.push({eggName, rhs: formatField(field.eggs[eggName])});
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -175,7 +175,7 @@ export class Mirror {
     // it requires bumping the version, bump it: requiring some extra
     // one-time cache resets is okay; doing the wrong thing is not.
     const blob = stringify({
-      version: "MIRROR_v3",
+      version: "MIRROR_v4",
       schema: this._schema,
       options: {
         blacklistedIds: this._blacklistedIds,
@@ -231,8 +231,9 @@ export class Mirror {
         dedent`\
           CREATE TABLE objects (
               id TEXT NOT NULL PRIMARY KEY,
-              typename TEXT NOT NULL,
+              typename TEXT,
               last_update INTEGER,
+              CHECK((last_update IS NOT NULL) <= (typename IS NOT NULL)),
               FOREIGN KEY(last_update) REFERENCES updates(rowid)
           )
         `,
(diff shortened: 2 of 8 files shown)


wchargin:
Thanks for the PR! I’ll review this this evening.

wpank:
@wchargin awesome, currently editing the description with some more context. This is more of a work in progress looking for some feedback. It doesn't _quite_ work yet, but is about 95% of the way there I think. `js-ipfs-unixfs-engine` will load without a blacklist, but it also will load without specifying unfaithful types in the schema. `tensorflow/tensorflow` still doesn't load. Hoping you could help figure out what might be missing here.

wchargin:
> This is more of a work in progress looking for some feedback

Okay. If you’re looking for design feedback that doesn’t pertain
directly to the code itself (e.g., “because of ABC, I took approach XYZ
instead of what you originally suggested; is this reasonable?”, or “this
unforeseen problem came up; I handled it with MNO, and I’m curious what
you think about that?”), then feel free to ask. Before I can provide any
substantive feedback about the code itself, please separate this one
giant PR into smaller PRs, each of which comprises a single logical
change. You can use the line items of my [suggested implementation
plan][plan] to delineate the commit boundaries. This will make it much
easier to comment and discuss the code in the GitHub UI.

In the end, each PR will be merged as a single commit in the repository
history. You don’t have to create all the PRs at once: we can review
them one at a time.

[plan]: https://github.com/sourcecred/sourcecred/issues/998#issuecomment-437548721

As you update these commits, please keep the following in mind:

  - The codebase needs to pass `yarn test` at every commit (tests,
    types, and style). Each commit needs to be correct without relying
    on all the commits that come after it. It should be possible to
    merge commits one at a time as they’re reviewed and approved, rather
    than waiting until the whole stack is done and nervously flipping a
    big flag.

    You can run `yarn prettify` to format all files, though for the
    future I highly recommend setting up your editor to automatically
    reformat on save.

  - Each commit headline should describe what _that commit_ does, not
    what the whole family of commits does. If these commits were merged
    as is, we’d have ~7 commits of the same name, which makes the blame
    and log harder to read.

  - Avoid commenting out code when possible. If it’s no longer needed,
    remove it.

  - Test plans describe a reproducible sequence that are sufficient to
    verify that the commit is correct, and should make it clear _why_
    the sequence is sufficient. For instance, “Inspected via SQLite
    shell” doesn’t actually tell me what you did: what database were you
    reading? what queries did you issue? what results did you get? why
    does this mean that the code is correct? Likewise, “This commit adds
    a check in `_updateData`…” isn’t a test plan.

    Sometimes, the test plan really is just “added new unit tests,
    retaining full coverage; `yarn test` passes” or “this is a refactor;
    existing unit tests suffice”. In those cases, that’s perfectly fine.

    Keep in mind that part of the purpose of the test plan is to help
    you improve the code. It’s extremely common for me to write a patch,
    start writing the commit message, get to the “Test Plan” section,
    and then realize that I really need to add more tests, or redesign
    the interfaces, or split the commit into multiple smaller ones. If
    the test plan is trying to tell you something, listen! :-)

  - Commits shouldn’t include `(#issue_number)` in the headline. When
    the pull request is merged, `(#pr_number)` will be appended to the
    headline, so the headline will look like `Do a thing (#123) (#456)`,
    which is confusing. Instead, just reference the relevant issue
    within the body.

  - The commit message body should be flush-left, not tab-indented.

And, on a technical note, please note that `graphql/schema.js` should be
entirely agnostic to the actual contents of the schema: it shouldn’t
know about users, organizations, and bots, for instance.

> It doesn't _quite_ work yet, but is about 95% of the way there I think

Okay. Do you suspect that the issues are implementation problems or a
problem with the design? If the latter, we might want to hold off on
reviewing the code in detail until we’re confident that the high-level
approach makes sense. If the former, it’s up to you how much we want to
troubleshoot these issues before reviewing the main code.

> `tensorflow/tensorflow` still doesn't load

What goes wrong?

(Note that this is a fairly large repository; something like #997 might
be lurking.)

---

Please take as much time as you need; I want to make sure that this
isn’t a stressful experience. Feel free to ping me or @decentralion with
any questions (process, code, or anything else); I’ll try to get back to
you within a day.

decentralion:
Closing this sequence of pulls for now, as we don't have bandwidth to get them merged, and we're prioritizing going deep (better cred on SourceCred) before going wide (applying SourceCred to many other projects), and atm these pulls are needed for going wide.

cf https://discourse.sourcecred.io/t/deep-then-wide/102/4

PR #1085, opened 2019-02-10
Enable loading private git repositories

decentralion:
This commit enables loading private repositories, assuming that the user
has ssh-agent configured with keys to allow cloning the private
repository, and has provided a GitHub API token with permissions for the
repository in question.

I have not added automated testing. I don't think a cost-benefit
analysis favors adding such tests at this time:
- This code changes very infrequently, and so is unlikely to break
- If it does break, it will be pretty easy to catch and to fix
- the @sourcecred org is on a free plan, which doesn't allow private
repos, so setting up the test case is a bit of a pain

Test plan: `yarn test --full` passes, so I haven't broken existing Git
clone behavior. Locally, I am able to load private repositories.

### DIFF
--- a/src/plugins/git/gitUtils.js
+++ b/src/plugins/git/gitUtils.js
@@ -4,6 +4,7 @@ import {execFileSync} from "child_process";
 import fs from "fs";
 import mkdirp from "mkdirp";
 import path from "path";
+import process from "process";
 
 export interface Utils {
   exec: GitDriver;
@@ -35,6 +36,10 @@ export function localGit(repositoryPath: string): GitDriver {
       // Ignore global Git settings, for test isolation.
       GIT_CONFIG_NOSYSTEM: "1",
       GIT_ATTR_NOSYSTEM: "1",
+      // Bring over the SSH configuration so that loading private repos is possible
+      // This post has some useful information on SSH_AUTH_SOCK:
+      // http://blog.joncairns.com/2013/12/understanding-ssh-agent-and-ssh-add/
+      SSH_AUTH_SOCK: process.env.SSH_AUTH_SOCK,
       ...(env || {}),
     };
     const options = {env: fullEnv};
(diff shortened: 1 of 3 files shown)


src/plugins/git/gitUtils.js
```diff
@@ -4,6 +4,7 @@ import {execFileSync} from "child_process";
 import fs from "fs";
 import mkdirp from "mkdirp";
 import path from "path";
+import process from "process";
```
wchargin: `process` is on the global object; no need to import.

src/plugins/git/gitUtils.js
```diff
@@ -35,6 +36,10 @@ export function localGit(repositoryPath: string): GitDriver {
       // Ignore global Git settings, for test isolation.
       GIT_CONFIG_NOSYSTEM: "1",
       GIT_ATTR_NOSYSTEM: "1",
+      // Bring over the SSH configuration so that loading private repos is possible
+      // This post has some useful information on SSH_AUTH_SOCK:
+      // http://blog.joncairns.com/2013/12/understanding-ssh-agent-and-ssh-add/
+      SSH_AUTH_SOCK: process.env.SSH_AUTH_SOCK,
```
wchargin: This adds `SSH_AUTH_SOCK=undefined` to the environment when the user
does not have this environment variable set, which is a bit unclean, but
from my testing this doesn’t look like it actually interferes with the
operation. No change requested.

src/plugins/git/gitUtils.js
decentralion: Done, thanks.

src/plugins/git/gitUtils.js
decentralion: Good catch. I experimented with setting it conditionally on SSH_AUTH_SOCK being present, but it meant I would need to put in a flow any-cast, or else write out an interface for the object, so I decided to just leave it as is.

PR #1087, opened 2019-02-14
Add `PagerankGraph.equals`

decentralion:
Part of ongoing work for #1020.

Adds an equals method for the PagerankGraph. This is really quite
straightforward, the logic is based on the matching logic for
`Graph.equals`.

Tests added.

Test plan: The added tests are comprehensive, and they pass.

### DIFF
--- a/src/core/pagerankGraph.test.js
+++ b/src/core/pagerankGraph.test.js
@@ -11,6 +11,11 @@ describe("core/pagerankGraph", () => {
   const nonEmptyGraph = () =>
     new Graph().addNode(NodeAddress.fromParts(["hi"]));
 
+  function examplePagerankGraph() {
+    const g = advancedGraph().graph1();
+    return new PagerankGraph(g, defaultEvaluator);
+  }
+
   it("cannot construct PagerankGraph with empty Graph", () => {
     const eg1 = new Graph();
     const eg2 = new Graph()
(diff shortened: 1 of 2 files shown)


src/core/pagerankGraph.test.js
```diff
@@ -175,4 +175,58 @@ describe("core/pagerankGraph", () => {
...
+    });
+    it("unequal syntheticLoopWeight => unequal", () => {
+      const pg1 = new PagerankGraph(nonEmptyGraph(), defaultEvaluator, 0.1);
+      const pg2 = new PagerankGraph(nonEmptyGraph(), defaultEvaluator, 0.2);
+      expect(pg1.equals(pg2)).toBe(false);
+    });
+    it("unequal graph => unequal", () => {
+      const pg1 = new PagerankGraph(nonEmptyGraph(), defaultEvaluator, 0.1);
+      const g2 = nonEmptyGraph().addNode(NodeAddress.empty);
+      const pg2 = new PagerankGraph(g2, defaultEvaluator, 0.2);
```
BrianLitwin: Are you wanting to isolate the difference in graph nodes here or is this meant to be a catch-all for different kinds of inequalities? Perhaps keep the synthetic loop weights equal if its the former.

decentralion: Good catch! I intended to just look at the difference in the underlying graph, putting different SLW was a mistake. Will fix.

PR #1088, opened 2019-02-14
PagerankGraph: Add toJSON/fromJSON

decentralion:
This commit adds serialization logic to `PagerankGraph`. As with many
things in PagerankGraph, it's based on the corresponding logic in `Graph`.
Much like graph, it stores data associated with nodes and edges (in this
case, the scores and edge weights) in an ordered array rather than a
map, so as to avoid repetitiously serializing the node and edge
addresses.

Test plan: Unit tests added, and they should be sufficient. Also take a
look at the included snapshot.

### DIFF
--- a/src/core/pagerankGraph.js
+++ b/src/core/pagerankGraph.js
@@ -36,7 +36,13 @@ export type WeightedEdge = {|
 
 export opaque type PagerankGraphJSON = Compatible<{|
   +graphJSON: GraphJSON,
+  // Score for every node, ordered by the sorted node address.
   +scores: $ReadOnlyArray<number>,
+  // Weights for every edge, ordered by sorted edge address.
+  // We could save the EdgeWeights directly rather than having separate
+  // arrays for toWeights and froWeights, but this would lead to an inflated
+  // JSON representation because we would be needlessly duplicating the keys
+  // "toWeight" and "froWeight" themselves.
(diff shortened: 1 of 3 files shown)


src/core/pagerankGraph.js
```diff
@@ -296,6 +315,85 @@ export class PagerankGraph {
...
+   * For space efficency, we store the node scores as an array of numbers in
+   * node-address-sorted order, and we store the edge weights as two arrays of
+   * numbers in edge-address-sorted-order.
+   */
+  toJSON(): PagerankGraphJSON {
+    this._verifyGraphNotModified();
+
+    const graphJSON = this.graph().toJSON();
+    const nodes = sortedNodeAddressesFromJSON(graphJSON);
+    const scores = new Array(nodes.length);
```
BrianLitwin: should you add a type to `scores` eg like `toWeights` and `froWeights` below ?

BrianLitwin: Also, out of curiosity, why the for-loops and not: 
```javascript
const scores = nodes.map((n) => NullUtil.get(this._scores.get(n))
```

decentralion: Thanks for the review! I'm not sure why I used the for loops, these maps are cleaner; I've modified the code. I also added the explicit type annotations per your suggestion.

PR #1089, opened 2019-02-16
[draft] Add an org-mode to `sourcecred load`

decentralion:
It's common for a project to be split up under many repositories; e.g.
ipfs has 238. We have a feature that allows the user to aggregate cred
for many repositories into one output, but doing so for 238 repos is
incredibly tedious.

This commit adds a `--organization` flag to `sourcecred load` which
loads all non-fork repositories for a given organization, and saves it
to the repoId `organization/combined`.

The list of non-fork repositories is retrieved by using the GitHub
GraphQL search API. Currently, this is implemented without the benefit
of re-using any of our other GraphQL logic; we should probably
de-duplicate it before merging.

Test plan: I've used this for several orgs for demos and it's worked;
hoewver, we should test it before merging.

### DIFF
--- a/src/cli/load.js
+++ b/src/cli/load.js
@@ -7,7 +7,12 @@ import mkdirp from "mkdirp";
 import path from "path";
 
 import * as RepoIdRegistry from "../core/repoIdRegistry";
-import {repoIdToString, stringToRepoId, type RepoId} from "../core/repoId";
+import {
+  repoIdToString,
+  stringToRepoId,
+  makeRepoId,
+  type RepoId,
+} from "../core/repoId";
 import dedent from "../util/dedent";
 import type {Command} from "./command";
 import * as Common from "./common";
@@ -15,6 +20,7 @@ import * as Common from "./common";
 import execDependencyGraph from "../tools/execDependencyGraph";
 import {loadGithubData} from "../plugins/github/loadGithubData";
 import {loadGitData} from "../plugins/git/loadGitData";
+import {fetchReposForOrg} from "../plugins/github/fetchReposForOrg";
 
 function usage(print: (string) => void): void {
   print(
@@ -78,9 +84,10 @@ function die(std, message) {
 }
 
 const load: Command = async (args, std) => {
-  const repoIds = [];
+  let repoIds = [];
   let explicitOutput: RepoId | null = null;
   let plugin: Common.PluginName | null = null;
+  let organization: string | null = null;
   for (let i = 0; i < args.length; i++) {
     switch (args[i]) {
       case "--help": {
@@ -105,12 +112,37 @@ const load: Command = async (args, std) => {
         plugin = arg;
         break;
       }
+      case "--organization": {
+        if (organization != null)
+          return die(std, "'--organization' given multiple times");
+        if (++i >= args.length)
+          return die(std, "--organization' given without value");
+        organization = args[i];
+        break;
+      }
       default: {
         // Should be a repository.
         repoIds.push(stringToRepoId(args[i]));
         break;
       }
     }
+  }
+
+  if (organization != null) {
+    if (repoIds.length > 0) {
+      die(std, "'--organization' cannot be used in conjunction with repoIds");
+    }
+    if (explicitOutput == null) {
+      explicitOutput = makeRepoId(organization, "combined");
+    }
+    const token = Common.githubToken();
+    if (token == null) {
+      // TODO(#638): This check should be abstracted so that plugins can
+      // specify their argument dependencies and get nicely formatted
+      // errors.
+      return die(std, "no GitHub token specified");
+    }
+    repoIds = await fetchReposForOrg(organization, token);
   }
 
   let output: RepoId;
--- a/src/plugins/github/fetchReposForOrg.js
+++ b/src/plugins/github/fetchReposForOrg.js
@@ -1,0 +1,70 @@
+// @flow
+
+import {type RepoId, makeRepoId} from "../../core/repoId";
+import dedent from "../../util/dedent";
+
+const GITHUB_GRAPHQL_SERVER = "https://api.github.com/graphql";
+async function ghquery(query, token): any {
+  const body = JSON.stringify({query});
+  const fetchOptions = {
+    method: "POST",
+    body,
+    headers: {
+      Authorization: `bearer ${token}`,
+    },
+  };
+  const response = await fetch(GITHUB_GRAPHQL_SERVER, fetchOptions);
+  const json = await response.json();
+  if (json.errors) {
+    throw json.errors;
+  }
+  return json;
+}
+
+export async function fetchReposForOrg(
+  org: string,
+  token: string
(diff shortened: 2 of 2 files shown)


src/plugins/github/fetchReposForOrg.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+
+import {type RepoId, makeRepoId} from "../../core/repoId";
+import dedent from "../../util/dedent";
+
+const GITHUB_GRAPHQL_SERVER = "https://api.github.com/graphql";
+async function ghquery(query, token): any {
```
wchargin: Why not use `postQuery` from `fetchGithubRepo.js`? That function already
exists and also has better error checking (rate limits, internal
execution errors, …) and the correct return type (`Promise<any>`).

(Also, `ghquery` is a convenient name for a shell alias, but a JS
function deserves a clearer name, like `postQuery`. :-) )

src/plugins/github/fetchReposForOrg.js
```diff
@@ -0,0 +1,70 @@
...
+  return json;
+}
+
+export async function fetchReposForOrg(
+  org: string,
+  token: string
+): Promise<$ReadOnlyArray<RepoId>> {
+  const query = dedent`\
+  {
+    search(query: "org:${org}", type: REPOSITORY, first: 50) {
```
wchargin: > ```js
> "org:${org}"
> ```

Please don’t do this. It’s an injection vector. There’s no need to open
this up.

Suppose, for instance, that you exposed a simple web service that
displayed SourceCred for a user-provided organization on request. It so
happens that my favorite GitHub organization is called

    ") { totalCount } } mutation Z { deleteProject(…) } query Q2 { search(query: "

which is unfortunate for whoever owns your API key.

That “this is draft/prototype code” is not a sufficient excuse. Draft
and prototype code turns into production code very quickly and without
anyone noticing. In the early days of SourceCred I made sure that even
our prototype code did not have this exact vulnerability.

We have a perfectly safe mechanism for doing this: use GraphQL variables
rather than string interpolation.

src/plugins/github/fetchReposForOrg.js
```diff
@@ -0,0 +1,70 @@
...
+          name
+        }
+      }
+      pageInfo {
+        endCursor
+        hasNextPage
+      }
+    }
+  }
+  `;
```
wchargin: Please prefer the structured API over manual string interpolation:

```js
  const b = Queries.build;
  const payload = {
    body: [
      b.query(
        "PerformSearch",
        [b.param("searchQuery", "String!")],
        [
          b.field(
            "search",
            {
              query: b.variable("searchQuery"),
              type: b.literal("REPOSITORY"),
              first: b.literal("30"),
            },
            [
              b.field("nodes", {}, [
                b.inlineFragment("Repository", [b.field("name")]),
              ]),
              b.field("pageInfo", {}, [
                b.field("endCursor"),
                b.field("hasNextPage"),
              ]),
            ]
          ),
        ]
      ),
    ],
    variables: {searchQuery: `org:${org}`},
  };
  const result = await postQuery(payload);
```

This emits

```graphql
query PerformSearch($searchQuery: String!) {
  search(query: $searchQuery type: "REPOSITORY" first: "30") {
    nodes {
      ... on Repository {
        name
      }
    }
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```

just like your version, but it is guaranteed at compile time that the
query is structurally correct (e.g., you have the correct number of
closing braces, everything is quoted properly, GraphQL syntax is
correct). Also, this sends fewer bytes over the wire because the
`inlineLayout` version is

```
query PerformSearch($searchQuery: String!) { search(query: $searchQuery type: "REPOSITORY" first: "30") { nodes { ... on Repository { name } } pageInfo { endCursor hasNextPage } } }
```

which is only 77% of the size of the original.

src/plugins/github/fetchReposForOrg.js
```diff
@@ -0,0 +1,70 @@
...
+  let json = await ghquery(query, token);
+  console.log(json);
+  let repos = json.data.search.nodes.map((x) => makeRepoId(org, x.name));
+  while (json.data.search.pageInfo.hasNextPage) {
+    const query = dedent`\
```
wchargin: After you switch from string interpolation to the structured API, it
will also be easier to make this second query reuse the components of
the original. With the structured API, GraphQL queries and their
components are simple immutable values, rather than arbitrary sequences
of Unicode code points. Please take advantage of this to ensure that
you’re querying for the same selection set on the initial and subsequent
queries.

src/plugins/github/fetchReposForOrg.js
```diff
@@ -0,0 +1,70 @@
...
+      pageInfo {
+        endCursor
+        hasNextPage
+      }
+    }
+  }
+  `;
+    json = await ghquery(query, token);
+    const newRepos = json.data.search.nodes.map((x) => makeRepoId(org, x.name));
+    repos = repos.concat(newRepos);
```
wchargin: Note that this is quadratic. Even though the quadratic factors are
dwarfed by network overhead, it would be better to store a list of lists
of repositories and then flatten them all at once at the end (when you
know the total count and can allocate the correct length of array).

src/plugins/github/fetchReposForOrg.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+  console.log(repos);
```
wchargin: Remove console logs?

src/cli/load.js
```diff
@@ -105,6 +112,14 @@ const load: Command = async (args, std) => {
         plugin = arg;
         break;
       }
+      case "--organization": {
```
wchargin: Before this commit, the `cli/` package had 100% test coverage. Please
maintain that.

src/cli/load.js
```diff
@@ -113,6 +128,23 @@ const load: Command = async (args, std) => {
     }
   }
 
+  if (organization != null) {
+    if (repoIds.length > 0) {
+      die(std, "'--organization' cannot be used in conjunction with repoIds");
+    }
+    if (explicitOutput == null) {
+      explicitOutput = makeRepoId(organization, "combined");
```
wchargin: \*sigh\*, yeah. I guess we don’t really have a better mechanism than
faking a repository ID for now.

src/plugins/github/fetchReposForOrg.js
decentralion: Unfortunately, what your version emits is _not_ the same as what my version emits. In particular, my version emits:

```

decentralion: Update: I've realized this is covered by the ENUM use case.

wchargin: Er, yes, sorry, that should have been `b.enumLiteral`. My bad.

decentralion:
I'm planning to split this into smaller PRs (starting with #1117) so I'll close this one.

PR #1090, opened 2019-02-17
Add optional node prefix filter to pagerankGraph

BrianLitwin:
Continuing work on #1020.
Adding an optional parameter to `nodes()` which enables optional
node prefix filtering.

Test plan:

@decentralion suggested on Discord that the tests should verify:
1) the parameter was passed to `_graph` correctly
2) the augmentation logic was applied correctly

The tests I added are identical to the tests in `graph.test`, except
that they verify that the result of `pagerankGraph` matches that of
`graph`. On one hand, this creates a dependence on `graph`,
as these tests don't verify that the filter works correctly, only that
graph has applied the filter and returned the iterator.
However, my prevailing thought is that it isn't `pagerankGraph's` responsibility
to test the behavior of `graph`, and so testing the exact filter results
of `pagerankGraph` like we do in `graph.test` isn't the best strategy, and
testing that `pagerankGraph`'s results equal `graph`'s results is a better strategy.

The tests also check that a `score` is provided alongside each `node` in the iterator,
to minimally satisfy @decentralion's second spec.

yarn test passes.

### DIFF
--- a/src/core/pagerankGraph.test.js
+++ b/src/core/pagerankGraph.test.js
@@ -1,7 +1,13 @@
 // @flow
 
 import sortBy from "lodash.sortby";
-import {Graph, NodeAddress, EdgeAddress, type Edge} from "./graph";
+import {
+  Graph,
+  NodeAddress,
+  EdgeAddress,
+  type NodeAddressT,
+  type Edge,
+} from "./graph";
 import {PagerankGraph} from "./pagerankGraph";
 import {advancedGraph} from "./graphTestUtil";
 import * as NullUtil from "../util/null";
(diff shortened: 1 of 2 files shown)


src/core/pagerankGraph.test.js
```diff
@@ -63,6 +69,50 @@ describe("core/pagerankGraph", () => {
     });
   });
 
+  describe("node prefix filter matches graph filter", () => {
```
decentralion: I would slightly restructure this describe block. I think the test that verifies that a prefix is required when options are passed should be outside of the describe, then the describe block should contain just the graph<>pagerankGraph consistency checks.

src/core/pagerankGraph.test.js
```diff
@@ -63,6 +69,50 @@ describe("core/pagerankGraph", () => {
...
+    const n4 = NodeAddress.fromParts(["zod", "bar"]);
+    const g = () =>
+      new Graph()
+        .addNode(n1)
+        .addNode(n2)
+        .addNode(n3)
+        .addNode(n4);
+    const pg = () => new PagerankGraph(g(), defaultEvaluator);
+
+    function expectPagerankGraphToEqualGraph(
```
decentralion: Nice use of a helper function for testing!

src/core/pagerankGraph.test.js
```diff
@@ -63,6 +69,50 @@ describe("core/pagerankGraph", () => {
...
+    it("requires a prefix when options are specified", () => {
+      // $ExpectFlowError
+      expect(() => Array.from(pg().nodes({}))).toThrow("prefix");
+    });
+
+    it("with prefix filter", () => {
+      expectPagerankGraphToEqualGraph({prefix: n2});
+    });
+
+    it("with prefix that matches nothing", () => {
```
decentralion: For completeness, let's also check with the prefix `NodeAddress.empty` (which matches everything).

BrianLitwin: In a separate PR, ought I add that test to `Graph.test` too?

src/core/pagerankGraph.test.js
BrianLitwin: Good call!

BrianLitwin:
Thanks for your review @decentralion, RAFL 👀!

src/core/pagerankGraph.test.js
decentralion: Sure. I would be fine with doing it in the same PR too, tbh.

PR #1092, opened 2019-02-18
PagerankGraph: Add `totalOutWeight`

decentralion:
This commit adds a `totalOutWeight` method to `PagerankGraph`.
For any given node, `totalOutWeight` reports the total weight traveling
away from the node on edges (including the synthetic loop edge). Using
totalOutWeight makes it possible to normalize the weights to get the
actual markov transition probabilities.

Test plan: Unit tests verify the following properties:
- An error is thrown if the requested node does not exist.
- An error is thrown if the graph has been modified.
- The out weights are computed correctly in the standard case.
- The out weights are computed correctly in the case where there are no
weights (except the synthetic loop weight)
- The out weights are still computed correctly after
JSON-deserialization.

### DIFF
--- a/src/core/pagerankGraph.test.js
+++ b/src/core/pagerankGraph.test.js
@@ -17,9 +17,11 @@ describe("core/pagerankGraph", () => {
   const nonEmptyGraph = () =>
     new Graph().addNode(NodeAddress.fromParts(["hi"]));
 
-  function examplePagerankGraph(): PagerankGraph {
+  function examplePagerankGraph(
+    edgeEvaluator = defaultEvaluator
+  ): PagerankGraph {
     const g = advancedGraph().graph1();
-    return new PagerankGraph(g, defaultEvaluator);
+    return new PagerankGraph(g, edgeEvaluator);
   }
   async function convergedPagerankGraph(): Promise<PagerankGraph> {
     const pg = examplePagerankGraph();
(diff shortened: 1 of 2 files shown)


src/core/pagerankGraph.test.js
```diff
@@ -168,6 +168,64 @@ describe("core/pagerankGraph", () => {
...
+      for (const {edge, weight} of pg.edges()) {
+        addWeight(edge.src, weight.toWeight);
+        addWeight(edge.dst, weight.froWeight);
+      }
+      for (const node of pg.graph().nodes()) {
+        expect(pg.totalOutWeight(node)).toEqual(outWeight.get(node));
+      }
+    }
+    it("computes outWeight correctly on the example graph", () => {
+      const eg = examplePagerankGraph();
```
BrianLitwin: Thought: 
The edge evaluator (`defaultEvaluator`) for `examplePagerankGraph()` sets `froWeight` to 0, so if you removed [this line], would this test not still pass? Because the example graph's `dst` node has two out edges that are both set to 0, so it's `totalOutWeight` will be it's synthetic loop weight. So maybe use a different edge evaluater for this test if those assumptions are correct.  

[this line]: https://github.com/sourcecred/sourcecred/blob/eea041a0fe2ef3938273387a5a3394b98bc33adb/src/core/pagerankGraph.js#L175

BrianLitwin: verified those assumptions above are correct => removing that line that accumulates `froWeights` to `edge.dst` has no effect on the outcome of the test. 

And the following seems to fix that => 

```javascript 
function examplePagerankGraph(
    edgeEvaluator = defaultEvaluator
  ): PagerankGraph {
    const g = advancedGraph().graph1();
    return new PagerankGraph(g, edgeEvaluator);
  }

it("computes outWeight correctly on the example graph", () => {
      const edgeEvaluator = (edge) => ({toWeight: 1, froWeight: 2});
      const eg = examplePagerankGraph(edgeEvaluator);
      verifyOutWeights(eg);
  });
```

src/core/pagerankGraph.test.js
```diff
@@ -168,6 +168,64 @@ describe("core/pagerankGraph", () => {
...
+      expect(() =>
+        eg.totalOutWeight(NodeAddress.fromParts(["nonexistent"]))
+      ).toThrowError("non-existent node");
+    });
+    function verifyOutWeights(pg: PagerankGraph) {
+      const outWeight: Map<NodeAddressT, number> = new Map();
+      for (const node of pg.graph().nodes()) {
+        outWeight.set(node, pg.syntheticLoopWeight());
+      }
+      function addWeight(node: NodeAddressT, weight: number) {
```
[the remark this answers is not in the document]
decentralion: Ok, done.

src/core/pagerankGraph.test.js
```diff
@@ -168,6 +168,64 @@ describe("core/pagerankGraph", () => {
...
+        const newWeight = NullUtil.get(outWeight.get(node)) + weight;
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/pagerankGraph.test.js
decentralion: Great catch, thanks! Fixed.

PR #1093, opened 2019-02-18
Add a test helper function for converged graphs

decentralion:
Really minor refactor, adds a `convergedPagerankGraph` helper method
which provides a converged pagerank graph. :)

Test plan: `yarn test` suffices.


decentralion:
Merging on self review as the pull is trivial.

PR #1094, opened 2019-02-18
PagerankGraph: add neighbors + score decomposition

decentralion:
This commit adds a `neighbors` method to `PagerankGraph`. This is an
augmented version of `Graph.neighbors`. It returns the base data from
`Graph.neighbors` as well as the score, the edge weights, and the score
contribution. The score contribution basically means how much score was
contributed from the target node by this particular neighbor connection.

When the graph is well-converged, a node's score will be the sum of all
its neighbors' score contributions, as well as the contribution it
received from its synthetic loop edge. So, for completeness sake, I
added another method, `syntheticLoopScoreContribution`, which computes
how much score a node received from its synthetic loop edge. (This value
should usually be close to 0).

You can think of these two methods as providing a replacement for the
`PagerankNodeDecomposition` logic.

Test plan: I've added tests that verify:
- That neighbors returns results consistent with Graph.neighbors
- That neighbors' score contributions are computed correctly
- That neighbors errors if the graph has been modified
- That synthetic score contributions are computed correctly
- That a node's score is the sum of all its contributions

Test plan: Unit tests included. Run `yarn test`.

### DIFF
--- a/src/core/pagerankGraph.js
+++ b/src/core/pagerankGraph.js
@@ -13,6 +13,7 @@ import {
   sortedEdgeAddressesFromJSON,
   sortedNodeAddressesFromJSON,
   NodeAddress,
+  type NeighborsOptions,
 } from "./graph";
 import {
   distributionToNodeDistribution,
@@ -23,6 +24,8 @@ import {
 import {findStationaryDistribution} from "../core/attribution/markovChain";
 import * as NullUtil from "../util/null";
 
+export {Direction} from "./graph";
+export type {DirectionT, NeighborsOptions} from "./graph";
 export type {EdgeWeight} from "./attribution/graphToMarkovChain";
 export type EdgeEvaluator = (Edge) => EdgeWeight;
 
@@ -34,6 +37,16 @@ export type ScoredNode = {|
 export type WeightedEdge = {|
   +edge: Edge,
   +weight: EdgeWeight,
+|};
+
+export type ScoredNeighbor = {|
+  // The neighbor node, with its score
+  +scoredNode: ScoredNode,
+  // The edge connecting the target to its neighbor node, with its weight
+  +weightedEdge: WeightedEdge,
+  // How much score (in absolute terms) was provided to the target by
+  // the neighbor node through this weightedEdge
+  +scoreContribution: number,
 |};
 
 export opaque type PagerankGraphJSON = Compatible<{|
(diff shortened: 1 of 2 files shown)


src/core/pagerankGraph.js
```diff
@@ -286,6 +299,88 @@ export class PagerankGraph {
...
+      let relevantEdgeWeight = 0;
+      if (edge.src === target) {
+        relevantEdgeWeight += weightedEdge.weight.froWeight;
+      }
+      if (edge.dst === target) {
+        relevantEdgeWeight += weightedEdge.weight.toWeight;
+      }
+      // We normalize this edge weight by the total outWeight for `node`.
+      const normalizedEdgeWeight =
+        relevantEdgeWeight / this.totalOutWeight(node);
```
BrianLitwin: If you ask for outgoing edges in `options`, do you still want to divide those weights by `totalOutWeight` of `node`?Wouldn't you be dividing `node`'s incoming edge weight by its `totalOutWeight`? And the total incoming edge weight does not equal the total outgoing edge weight? (but total incoming score does equal total outgoing score, doesn't it?? Which sort of makes this difficult to model on a napkin)

So maybe if you are asking for outgoing edges, divide by the `totalOutWeight` of `target`? So if you ask for incoming edges, the `scoredContribution` is `node` 's contribution to `target`, and if you ask for outgoing edges, you get `target`'s contribution to `node` ?

decentralion: Remember that for the purposes of pagerank/the weighted graph, every edge is actually two edges: one pointing from src to dst, and one pointing from dst to src. So even if we are filtering for outgoing edges (i.e. where target is the src), we still have an edge coming back from `node` to `target`. Then to get the normalized weight of the edge from `node` to `target`, we need to normalize by `node`'s total weight (because the markov chain rule is that the weight of all the connections out of a node need to sum to one).

Basically, the concept of "outgoing edge" at the neighbor API level is orthogonal to the concept of "outgoing connection", since every edge (whether inbound or outbound) corresponds to two connections.

Does that make sense?

BrianLitwin: I got `froWeight` and `toWeight` mixed up. I was reading the following snippet as accumulating weights from `target` to `node` instead of `node` to `target`: 
```javascript 
 if (edge.src === target) {
        relevantEdgeWeight += weightedEdge.weight.froWeight;
  }
```

PR #1096, opened 2019-02-18
Mirror Refinements 2: adds typenames field to `QueryPlan`

wpank:
This commit addresses (#998) and (#996) and corresponds to
2.) in @wchargin's suggested implementation plan.

This adds a `typenames` field to the `QueryPlan` type
in Mirror. `typenames` is a list of GraphQL id's to be queried for
their canonical typename.

`_findOutdated` in Mirror has been updated to an empty list for
this new field.

`_queryFromPlan` has been updated to throw a `Not Yet Implemented`
Error if `queryPlan.typenames` is not empty.

Test Plan: Tests in `mirror.test.js` have been updated to include
an list for a `typenames` field anywhere a `QueryPlan` is.

A test has been added to `_queryFromPlan` that checks that
`typenames` will throw a NYE error given a non-null list.

This PR depends on #1095

### DIFF
--- a/src/graphql/generateFlowTypes.js
+++ b/src/graphql/generateFlowTypes.js
@@ -48,12 +48,18 @@ export default function generateFlowTypes(
     }
   }
   function formatNodeField(field: NodeFieldType): string {
+    if (field.fidelity.type === "UNFAITHFUL") {
+      throw new Error("Unfaithful Fidelity not yet supported");
+    }
     return "null | " + field.elementType;
   }
   function formatConnectionField(field: ConnectionFieldType): string {
     return `$ReadOnlyArray<null | ${field.elementType}>`;
   }
   function formatNestedField(field: NestedFieldType): string {
+    if (field.fidelity.type === "UNFAITHFUL") {
+      throw new Error("Unfaithful Fidelity not yet supported");
+    }
     const eggs = [];
     for (const eggName of Object.keys(field.eggs).sort()) {
       eggs.push({eggName, rhs: formatField(field.eggs[eggName])});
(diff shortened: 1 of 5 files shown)


decentralion:
@wpank I recommend modifying this PR so that the base branch points to #1095 rather than to master. It'll make the PR a lot cleaner.

(Note, we then need to be careful when merging, so that it actually does merge into master (after #1095 has merged) rather than into an obsolete branch.)

wpank:
@decentralion :+1: updated all these to ahve #1095 as the base.

decentralion:
@wpank are you sure? for me it still says the base for this one is sourcecred:master rather than wpank:mirror-inconsistent-typename-1. (and https://github.com/sourcecred/sourcecred/pull/1097 points to sourcecred:master rather than wpank:mirror-inconsistent-typename-2 and so forth)

wchargin:
@decentralion: I believe that the base branch is required to correspond
to the repository in which the pull request was opened, so this PR
cannot be against `wpank:*`.

@wpank: I’d say don’t worry about it. For incremental reviews, we can
always look at the last commit in each PR; after we merge each one, you
can rebase the rest. The GitHub workflow for dependent pull requests is
really a bit annoying, but there aren’t too many ways to slice it. (As
it happens, I’m currently working on a more polished version of some
scripts that I have to automate this…but that probably won’t be safe for
mass consumption before these PRs are merged.)

wpank:
@decentralion Ah I rebased them all against wpank:*. [Here's the network graph](https://github.com/wpank/sourcecred/network). Like @wchargin mentioned, it seems you can't change the base branch in a PR to correspond to another repo (which makes sense, otherwise you'd just make the PR in that repo instead).

PR #1104, opened 2019-02-21
Move EdgeOptions error checking into helper function

BrianLitwin:
This is motivated by the desire to check for errors at the call site
of `edges()` in both `graph` and `pagerankGraph`, ie at the iterator
wrapper method, as discussed w/ @decentralion on Discord. PagerankGraph
will need to do the same error checking at it's `edges()` method, and
this helper function will enable us to re-use the error checking logic
in both classes.

Test plan:
yarn test passes.
I'm not sure if it makes sense to test this function separately, since
it's creation was motivated by the desire to test the consistency of
`graph` and `pagerankGraph`.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -457,25 +457,10 @@ export class Graph {
    * undefined.
    */
   edges(options?: EdgesOptions): Iterator<Edge> {
-    if (options == null) {
-      options = {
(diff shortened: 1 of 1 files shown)


decentralion:
I think we can render this pull request unnecessary: see https://github.com/sourcecred/sourcecred/pull/1105#pullrequestreview-206496537

PR #1105, opened 2019-02-21
pagerankGraph: add edge filter

BrianLitwin:
Part of ongoing work for #1020.

Test plan:
Added tests that mirror the edge filtering tests in `graph.test`
to check that `graph` and `pagerankGraph` return the same edges
with the given `EdgesOptions` parameter. Also added a sanity check
that a `weight` prop is returned from the iterator along with the edge.

Given the dependence on a helper function to test the edge
iterator's equality between graphs, I would suggest reviewers give
particular attention to that function:
`expectConsistentEdges()`

### DIFF
--- a/src/core/pagerankGraph.test.js
+++ b/src/core/pagerankGraph.test.js
@@ -7,6 +7,7 @@ import {
   EdgeAddress,
   type NodeAddressT,
   type Edge,
+  type EdgesOptions,
 } from "./graph";
 import {PagerankGraph} from "./pagerankGraph";
 import {advancedGraph} from "./graphTestUtil";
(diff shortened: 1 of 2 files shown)


BrianLitwin:
Updated @decentralion ! RAFL.

src/core/pagerankGraph.test.js
```diff
@@ -165,6 +166,131 @@ describe("core/pagerankGraph", () => {
...
+        [src1, src2, dst1, dst2].reduce((g, n) => g.addNode(n), new Graph())
+      );
+
+    const pagerankGraph = () => new PagerankGraph(graph(), defaultEvaluator);
+
+    function expectPagerankGraphToEqualGraph(options: EdgesOptions | void) {
+      const pagerankGraphEdges = Array.from(
+        pagerankGraph().edges(options)
+      ).sort();
+      const graphEdges = Array.from(graph().edges(options)).sort();
```
decentralion: FWIW, the sort should be unnecessary--the order should be the same. Hoewver, we don't make guarantees about iteration order, so either is fine.

src/core/pagerankGraph.test.js
```diff
@@ -165,6 +166,131 @@ describe("core/pagerankGraph", () => {
...
+
+      const edgesAreEqual = (e1, e2) =>
```
decentralion: We don't need to define functions like this, since we can just import "lodash.isequal". We usually import it as `deepEqual`, so `git grep deepEqual` will find you examples.

Also, be aware that expect(object).toEqual(object) is a thing, and also uses deep-equal semantics

src/core/pagerankGraph.test.js
```diff
@@ -165,6 +166,131 @@ describe("core/pagerankGraph", () => {
...
+    };
+    const graph = () =>
+      [e11, e12, e21, e22].reduce(
+        (g, e) => g.addEdge(e),
+        [src1, src2, dst1, dst2].reduce((g, n) => g.addNode(n), new Graph())
+      );
+
+    const pagerankGraph = () => new PagerankGraph(graph(), defaultEvaluator);
+
+    function expectPagerankGraphToEqualGraph(options: EdgesOptions | void) {
```
decentralion: this function name is a little confusing, it makes me think we are checking graph equality rather than edges equality. Maybe `expectConsistentEdgesFromGraphAndPagerankGraph` or just `expectConsistentEdges`.

src/core/pagerankGraph.test.js
```diff
@@ -165,6 +166,131 @@ describe("core/pagerankGraph", () => {
...
+      pagerankGraphEdges.forEach((pgEdge, i) =>
+        expect(edgesAreEqual(pgEdge, graphEdges[i])).toBe(true)
+      );
+    }
+
+    describe("edge filter matches graph edge filter", () => {
+      it("finds all edges when no options are specified", () => {
+        expectPagerankGraphToEqualGraph(undefined);
+      });
+      it("finds all edges when universal filters are specified", () => {
```
decentralion: this isn't a universal filter, it's a filter that gets every edge in the graph.
a universal filter would be:
```
{
addressPrefix: EdgeAddress.empty,
srcPrefix: NodeAddress.emtpy,
dstPrefix: NodeAddress.empty,
}
```

src/core/pagerankGraph.test.js
```diff
@@ -165,6 +166,131 @@ describe("core/pagerankGraph", () => {
...
+        pagerankGraph().edges(options)
+      ).sort();
+      const graphEdges = Array.from(graph().edges(options)).sort();
+
+      const edgesAreEqual = (e1, e2) =>
+        e1.edge.address === e2.address &&
+        e1.edge.src === e2.src &&
+        e1.edge.dst === e2.dst;
+
+      pagerankGraphEdges.forEach((pgEdge, i) =>
```
decentralion: rather than checking each edge in an individual expect, I recommend checking the whole set at once:

```
const pagerankGraphEdges = Array.from(pagerankGraph.edges(options).map((e) => e.edge);
const graphEdges = Array.from(pagerankGraph.edges(options);
expect(pagerankGraphEdges).toEqual(graphEdges);
```

src/core/pagerankGraph.test.js
BrianLitwin: okay, going to leave `sort()` out then if it's just noise.

src/core/pagerankGraph.test.js
BrianLitwin: Definitely agree, went with `expectConsistentEdges`, thanks for the naming suggestion.

src/core/pagerankGraph.test.js
BrianLitwin: The thesaurus turned up `all-inclusive`.. that feels like a precise description.

src/core/pagerankGraph.test.js
BrianLitwin: Gotcha! I remember how you did this live coding pgGraph now.

src/core/pagerankGraph.test.js
BrianLitwin: Thanks! Remembering now that William [spoke on this](https://www.youtube.com/watch?v=xKfIOcN-xnA#t=1h47m21s) at live coding pgGraph.

BrianLitwin:
I think this is RFAL @decentralion. Besides the changes based on your comments, I added a sanity check in `expectConsistentEdges()` to ensure the iterator is returning a `weight` param w/ the edges.

src/core/pagerankGraph.test.js
```diff
@@ -165,6 +166,122 @@ describe("core/pagerankGraph", () => {
...
+      pagerankGraphEdges.forEach((e) => {
+        expect(e.weight.froWeight).toBe(0);
+        expect(e.weight.toWeight).toBe(1);
+      });
+      const graphEdges = Array.from(graph().edges(options));
+      expect(pagerankGraphEdges.map((e) => e.edge)).toEqual(graphEdges);
+    }
+
+    describe("edge filter matches graph edge filter", () => {
+      it("finds all edges when no options are specified", () => {
```
decentralion: If you wanted to code golf this section, you could do so with something like:

```js
const mkOptions = (edgeParts, srcParts, dstParts) => ({
  addressPrefix: EdgeAddress.fromParts(edgeParts),
  srcPrefix: NodeAddress.fromParts(srcParts),
  dstPrefix: NodeAddress.fromParts(dstParts),
})

const testCases = [
  { description: "when no options are specified", options: null },
  {
    description: "when all-inclusive filters are specified", 
    options: mkOptions(["e"], ["src"], ["dst"])
  },
  ...
]

for (const case of testCases) {
  it(case.description,() => {
    expectConsistentEdges(case.options);
  })
}
```

Change not requested, since your current approach is fine + totally readable.
And we aren't running out of paper, so there's no _need_ for code golf. It can
be fun, though. :)

wchargin: My $0.02: I prefer the way that @BrianLitwin wrote it, primarily because
it’s simpler. Programmatically generating tests can be useful especially
when the parameters themselves are programmatically generated: e.g.,

<https://github.com/sourcecred/sourcecred/blob/656a2d15433f37f358027f03f7f0e54dd8b0fa01/src/plugins/github/render.test.js#L12-L19>

But here it’s not clear to me that you really buy anything from
generating the tests, and the cost is nonzero; for instance:

  - Reading `mkOptions(["e"], ["src"], ["dst"])` doesn’t tell you as
    much as reading the object with named fields. With the latter, not
    only are the parameters named so that you know which is which, but
    if you’re just jumping to this test case (which is likely, assuming
    that it’s just failed) then the names help you gain context on what
    kind of code this is testing.

  - The “golfed” form impedes searchability. If the test fails with an
    error message like “bad addressPrefix: `foo`”, I might grep the
    codebase or file for `addressPrefix.*foo`. With `mkOptions`, that
    won’t find the right thing. Similarly, `it.*all-inclusive filters`
    doesn’t find the test case definition.

(Definitely agreed that code golf can be fun, though, and you’ll often
learn a thing or two!)

decentralion: +1 to what William said. :)

BrianLitwin: My $0.02 preference is try to keep the code in `graph.test` and `pagerankGraph.test` consistent because they're testing the same conditions.. what do y'all think about moving these graph/edge generators into a new test utility file (or `graphTestUtil`) and sharing them between `graph.test` and `pagerankGraph.test` ? 

Things that it would make easier: 
1) being certain the test cases are consistent and subsequent efforts to reason about test failures in one or both files. 
2) keeping the workflow/verbiage more consistent i.e. more readable.  E.g. "universal" vs "all-inclusive" filters are being used to describe the same edge filters in `graph.test` and `pagerankGraph.test`, respectively.  The same `graph` being created w/ `forEach` in one file and `reduce` in the other. 
2) adding new test cases 


And I think we could still golf 🏌️but keep the slightly more verbose version for clarity 
e.g.  => 
```javascript 
// create the edges, nodes, and graph in helper file 

const testCases = [
  {
    desc: "finds all edges when all-inclusive filters are specified",
    options: {
      addressPrefix: EdgeAddress.fromParts(["e"]),
      srcPrefix: NodeAddress.fromParts(["src"]),
      dstPrefix: NodeAddress.fromParts(["dst"]),
    }, 
    expected: [e11, e12, e21, e22]
  },
  {
    desc: "finds edges by address prefix",
    options: {
      addressPrefix: EdgeAddress.fromParts(["e", "1"]),
      srcPrefix: NodeAddress.empty,
      dstPrefix: NodeAddress.empty,
    },
   expected: [e11, e12]
  },
];

```

decentralion: I'm fine with a refactor to make the tests more consistent, but my preference would be that this commit merges more-or-less as is,  and then there's a separate pull that refactors the two to be consistent.

My reasoning being that there's some work I want to do that depends on having these methods in PagerankGraph, so preference towards merging it sooner. (Although admittedly there are a few other things I want to work on between now and then, so it's not a real rush.)

BrianLitwin:
@decentralion summary: 
* no golf on account of a few discussed factors :( 
* updated the `reduce()` to a `forEach()` in the graph generation step, as you suggested.

PR #1111, opened 2019-03-05
Add CLI command to clear sourcecred data directory

### LINKED ISSUE #1067: Add CLI command for clearing cache or data directory
It's relatively common at this point that we make backwards-incompatible changes to our serialized data formats, or to the GitHub mirror cache. When we do so, users need to reset their SourceCred data directory, which they may do by invoking either:

`> rm -r $SOURCECRED_DIRECTORY` (if they set the flag)
`> rm -r /tmp/sourcecred` (if they have not)

Since there are two different ways to do this, and only one is correct depending on how the user has configured their environment, it makes it complicated to help people who are in a corrupted state. In the medium term, we should fix the root issue (i.e. automatically upgrade the data formats so the user doesn't need to take any special action when the data format changes). However, for now, we could make this a little easier by adding a command like:

`sourcecred clear --all` -- removes the whole $SOURCECRED_DIRECTORY
`sourcecred clear --cache` -- removes just the cache directory
`sourcecred clear` -- prompts the user to be more specific

Then those errors that are thrown can report exactly how to fix them, i.e. when we have the wrong version of repoIdRegistry, we can ensure that "try running `sourcecred clear --all`" is printed to console.

BrianLitwin:
Resolves #1067

Adds the CLI commands:
`sourcecred clear --all` -- removes the $SOURCECRED_DIRECTORY
`sourcecred clear --cache` -- removes the cache directory
`sourcecred clear --help` -- provides usage info
`sourcecred clear` -- prompts the user to be more specific

Test plan:
The unit tests ensure that the command is properly wired into the
 sourcecred CLI, including help text integration. However, just to be
safe, we can start by verifying that calling `sourcecred` without
arguments lists the `clear` command as a valid option, and that
calling `sourcecred help clear` prints help information. (Note: it's
necessary to run `yarn backend` before testing these changes)

The unit tests also ensure that the command removes the proper
directories, so there isn't really a need to manually test it,
although the reviewer may choose to do so to be safe.

Although out of scope for unit tests on this function, we can also do
integration tests, to make sure that running the clear command doesn't
leave the sourcecred directory in an invalid state from the perspective of the `load` command.

```js
$ yarn backend;
$ node bin/sourcecred.js load sourcecred/example-github;
$ node bin/sourcecred.js clear --cache;
$ node bin/sourcecred.js load sourcecred/example-github;
$ node bin/sourcecred.js clear --all;
$ node bin/sourcecred.js load sourcecred/example-github;
```
The expected behavior of the above command block is that the load command never fails or throws an error.

@decentralion and I discussed the scenario where `rimraf` errors.
We decided that testing this scenario wasn't necessary, because
`rimraf` doesn't error if a directory doesn't exist, and
rimraf's maintainer suggests [monkey-patching the fs module]
to get rimraf to error in testing scenarios.

Thanks @decentralion for reviewing and pair-programming this with me.

[monkey-patching the fs module]: https://github.com/isaacs/rimraf/issues/31#issuecomment-29534796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
 ## [Unreleased]
+
 - Add the `pagerank` command (#1114)
+- Add the `clear` command (#1111)
 - Add description tooltips for node and edge types in the weight configuration UI (#1081)
 - Add the `export-graph` command (#1110)
 - Enable loading private repositories (#1085)
@@ -11,6 +13,7 @@
 <!-- Please add new entries to the _top_ of this section. -->
 
 ## [0.2.0]
+
 - Cache GitHub data, allowing for incremental and resumable loading (#622)
 - Hyperlink Git commits to GitHub (#887)
 - Relicense from MIT to MIT + Apache-2 (#812)
@@ -23,6 +26,7 @@
 - Add `MentionsAuthor` edges to the graph (#808)
 
 ## [0.1.0]
+
 - Organize weight config by plugin (#773)
 - Configure edge forward/backward weights separately (#749)
 - Combine "load graph" and "run pagerank" into one button (#759)
--- a/src/cli/clear.js
+++ b/src/cli/clear.js
@@ -1,0 +1,111 @@
+// @flow
+// implementation of `sourcecred clear`
+
+import path from "path";
+import rimraf from "rimraf";
+
+import dedent from "../util/dedent";
+import {type Command} from "./command";
+import * as Common from "./common";
+
+function usage(print: (string) => void): void {
+  print(
+    dedent`\
+    usage: sourcecred clear --all
+           sourcecred clear --cache
+           sourcecred clear --help
+
+    Remove the SOURCECRED_DIRECTORY, i.e. the directory where data, caches,
+    registries, etc. owned by SourceCred are stored.
+
+    Arguments:
+        --all
+            remove entire SOURCECRED_DIRECTORY
+
+        --cache
+            remove only the SourcCred cache directory
+
+        --help
+            Show this help message and exit, as 'sourcecred help clear'.
+
+    Environment Variables:
+        SOURCECRED_DIRECTORY
+            Directory owned by SourceCred, in which data, caches,
+            registries, etc. are stored. Optional: defaults to a
+            directory 'sourcecred' under your OS's temporary directory;
+            namely:
+                ${Common.defaultSourcecredDirectory()}
+
+    `.trimRight()
+  );
+}
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  std.err("fatal: run 'sourcecred help clear' for help");
+  return 1;
+}
+
+export function makeClear(removeDir: (string) => Promise<void>): Command {
+  return async function clear(args, std) {
+    async function remove(dir) {
+      try {
+        await removeDir(dir);
+        return 0;
+      } catch (error) {
+        return die(std, `${error}`);
+      }
+    }
+
+    switch (args.length) {
+      case 0:
+        return die(std, "no arguments provided");
+      case 1:
+        switch (args[0]) {
+          case "--help":
+            usage(std.out);
+            return 0;
+
+          case "--all":
+            return remove(Common.sourcecredDirectory());
+
+          case "--cache":
+            return remove(path.join(Common.sourcecredDirectory(), "cache"));
+
+          default:
+            return die(std, `unrecognized argument: '${args[0]}'`);
+        }
+      default:
+        return die(
+          std,
+          `expected 1 argument but recieved: ${args.length} arguments`
+        );
+    }
+  };
+}
+
+export function removeDir(p: string): Promise<void> {
+  return new Promise((resolve, reject) =>
+    rimraf(p, (error) => {
+      if (error) {
+        reject(error);
+      } else {
+        resolve();
+      }
+    })
+  );
+}
+
+export const help: Command = async (args, std) => {
+  if (args.length === 0) {
+    usage(std.out);
+    return 0;
+  } else {
+    usage(std.err);
+    return 1;
+  }
+};
+
+export const clear = makeClear(removeDir);
+
+export default clear;
--- a/src/cli/clear.test.js
+++ b/src/cli/clear.test.js
@@ -1,0 +1,134 @@
+// @flow
+
+import path from "path";
+import tmp from "tmp";
+import fs from "fs";
+
+import {makeClear, removeDir, help} from "./clear";
+import {run} from "./testUtil";
+import * as Common from "./common";
+
+describe("cli/clear", () => {
+  describe("'help' command", () => {
+    it("prints usage when given no arguments", async () => {
+      expect(await run(help, [])).toEqual({
+        exitCode: 0,
+        stdout: expect.arrayContaining([
+          expect.stringMatching(/^usage: sourcecred clear/),
+        ]),
+        stderr: [],
+      });
+    });
+
+    it("fails when given arguments", async () => {
+      expect(await run(help, ["foo/bar"])).toEqual({
+        exitCode: 1,
+        stdout: [],
+        stderr: expect.arrayContaining([
+          expect.stringMatching(/^usage: sourcecred clear/),
+        ]),
+      });
+    });
+  });
+
+  describe("'makeClear' command", () => {
+    it("prints usage with '--help'", async () => {
+      const clear = makeClear(jest.fn());
+      expect(await run(clear, ["--help"])).toEqual({
+        exitCode: 0,
+        stdout: expect.arrayContaining([
+          expect.stringMatching(/^usage: sourcecred clear/),
+        ]),
+        stderr: [],
+      });
+    });
+
+    it("fails when no arguments specified", async () => {
+      const clear = makeClear(jest.fn());
+      expect(await run(clear, [])).toEqual({
+        exitCode: 1,
+        stdout: [],
+        stderr: [
+          "fatal: no arguments provided",
+          "fatal: run 'sourcecred help clear' for help",
+        ],
+      });
+    });
+
+    it("fails when an invalid argument is specified", async () => {
+      const clear = makeClear(jest.fn());
+      expect(await run(clear, ["invalid"])).toEqual({
+        exitCode: 1,
+        stdout: [],
+        stderr: [
+          "fatal: unrecognized argument: 'invalid'",
+          "fatal: run 'sourcecred help clear' for help",
+        ],
+      });
+    });
+
+    it("fails when more than one argument specified", async () => {
+      const clear = makeClear(jest.fn());
+      expect(await run(clear, ["1", "2"])).toEqual({
+        exitCode: 1,
+        stdout: [],
+        stderr: [
+          "fatal: expected 1 argument but recieved: 2 arguments",
+          "fatal: run 'sourcecred help clear' for help",
+        ],
+      });
+    });
+
+    it("passes correct param to removeDir with `--all`", async () => {
+      const rmDir = jest.fn();
+      const clear = makeClear(rmDir);
+      await run(clear, ["--all"]);
+      expect(rmDir).toHaveBeenCalledWith(Common.sourcecredDirectory());
+    });
+
+    it("passes correct param to removeDir with `--cache`", async () => {
+      const rmDir = jest.fn();
+      const clear = makeClear(rmDir);
+      await run(clear, ["--cache"]);
+      const cacheDir = path.join(Common.sourcecredDirectory(), "cache");
+      expect(rmDir).toHaveBeenCalledWith(cacheDir);
+    });
+
+    function throwError() {
+      return Promise.reject(new Error("test error"));
+    }
+
+    it("--all returns error if removeDir errors", async () => {
+      const clear = makeClear(throwError);
+      expect(await run(clear, ["--all"])).toEqual({
+        exitCode: 1,
+        stdout: [],
+        stderr: [
+          "fatal: Error: test error",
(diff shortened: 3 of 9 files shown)


src/cli/clear.js
```diff
@@ -0,0 +1,92 @@
...
+      removePath(cacheDir);
+      return 0;
+    case undefined:
+      return die(std, "no arguments specified");
+    default:
+      return die(std, "invalid arguments");
+  }
+};
+
+const removePath = (p) => {
```
wchargin: I think it’s worth using `rimraf` here; it has error handling for
various OS-specific quirks (e.g., “fix EPERM on Windows”).

src/cli/clear.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  std.err("fatal: run 'sourcecred help clear' for help");
+  return 1;
+}
+
+export const clear: Command = async (args, std) => {
+  switch (args[0]) {
```
wchargin: Should fail if there is not exactly one argument, rather than silently
swallowing all but the first.

src/cli/clear.js
```diff
@@ -0,0 +1,92 @@
...
+      removePath(Common.sourcecredDirectory());
+      return 0;
+    case "--cache":
+      const cacheDir = path.join(Common.sourcecredDirectory(), "cache");
+      removePath(cacheDir);
+      return 0;
+    case undefined:
+      return die(std, "no arguments specified");
+    default:
+      return die(std, "invalid arguments");
```
wchargin: Can we include the value of the invalid argument?

src/cli/clear.js
```diff
@@ -0,0 +1,92 @@
...
+
+    Arguments:
+        --all
+            remove entire SOURCECRED_DIRECTORY
+
+        --cache
+            remove only the SourcCred cache directory
+
+        --help
+            Show this help message and exit, as 'sourcecred help export-graph'.
```
wchargin: s/export-graph/clear/

src/cli/clear.test.js
```diff
@@ -0,0 +1,81 @@
+// @flow
+
+import fs from "fs";
+import path from "path";
+import tmp from "tmp";
+
+import {clear} from "./clear";
+import {run} from "./testUtil";
+import {help} from "./clear";
+
```
wchargin: Consider mocking out `rimraf`/`removePath` here—not for testing
convenience, but for safety in case the code under test or the test
harness is wrong. I could easily see a misconfigured test accidentally
removing the test runner’s actual SourceCred directory, which would be
quite bad.

src/cli/clear.js
BrianLitwin: I've updated that function

src/cli/clear.js
BrianLitwin: updated to `'${args[0]}' is not a valid argument`

src/cli/clear.test.js
BrianLitwin: Updated.. trying to copy what @decentralion did w/ `exportGraph` in #1110.

src/cli/clear.js
BrianLitwin: Thanks, fixed.

src/cli/clear.js
wchargin: To clarify: <https://www.npmjs.com/package/rimraf>

src/cli/clear.test.js
wchargin: Yep, understood. The same general approach is good. It’s just that
`export-graph`’s primary effect is fairly benign (printing to stdout),
whereas a mistake in `clear` could actually be destructive to the host.

src/cli/clear.js
```diff
@@ -0,0 +1,99 @@
...
+      case "--cache":
+        const cacheDir = path.join(Common.sourcecredDirectory(), "cache");
+        return removeDir(cacheDir);
+      default:
+        return die(std, `'${args[0]}' is not a valid argument`);
+    }
+  };
+}
+
+//how do we handle callbacks
```
decentralion: We generally prefer to use promises and async rather than callbacks. The other commands import "fs-extra" rather than "fs". 

https://www.npmjs.com/package/fs-extra

src/cli/clear.js
```diff
@@ -0,0 +1,99 @@
...
+export const removeDir = (p: string) => {
+  const onError = (e) => {
+    return die(`Error removing directory: ${e}`);
+  };
+  rimraf(p, onError);
```
decentralion: if you want to make it async, I think the following should work:

```
export function removeDir(p: string): Promise<void> {
  return new Promise((resolve, reject) => {
    rimraf(p, (error) => ({
      if (error) {
        reject(error);
      } else {
        resolve();
      }
    })
  })
}
```

src/cli/clear.js
```diff
@@ -0,0 +1,99 @@
...
+  return 1;
```
decentralion: why does it return 1?

src/cli/clear.js
```diff
@@ -0,0 +1,99 @@
...
+  );
+}
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  std.err("fatal: run 'sourcecred help clear' for help");
+  return 1;
+}
+
+export function makeClear(removeDir: (string) => void): Command {
```
decentralion: the signature here suggests that removeDir operates synchronously, which it shouldn't be. So the signature should be:
`removeDir: (string) => Promise<void>`

src/cli/clear.test.js
```diff
@@ -0,0 +1,95 @@
...
+      expect(await run(help, ["foo/bar"])).toEqual({
+        exitCode: 1,
+        stdout: [],
+        stderr: expectToPrint(/^usage: sourcecred clear/),
+      });
+    });
+  });
+
+  describe("'clear' command", () => {
+    const mockFunc = () => jest.fn();
```
decentralion: mockFunc is equivalent to `jest.fn`. It doesn't serve any purpose that I can see.

src/cli/clear.js
```diff
@@ -0,0 +1,99 @@
...
+        const cacheDir = path.join(Common.sourcecredDirectory(), "cache");
+        return removeDir(cacheDir);
+      default:
+        return die(std, `'${args[0]}' is not a valid argument`);
+    }
+  };
+}
+
+//how do we handle callbacks
+export const removeDir = (p: string) => {
```
decentralion: Currently, this method is not tested by unit tests. It should be. I would add another describe block to the tests for checking that this function works as intended.

src/cli/help.js
```diff
@@ -41,6 +43,7 @@ function usage(print: (string) => void): void {
       analyze       analyze cred for a loaded repository
       export-graph  print a raw SourceCred graph
       help          show this help message
+      clear         clear soucrcecred data
```
decentralion: typo, should say SourceCred

BrianLitwin: Thanks!

BrianLitwin:
@decentralion - 

My thoughts about this: 
1) I think the current code is good, addresses all previous comments, but still lacks: 

2) a test for a `rimraf` error. I test that the `makeClear` function handles errors thrown by the func passed in as a parameter, but not that `removeDir` behaves correctly if `rimraf` were to err. 

3) Out of scope for this PR, but it would be useful to be able to clear specific projects/directories and not be forced to re-download them all. E.g. if you've loaded Tensorflow and Sourcecred/PM, and you only want to reload Sourcecred/PM.

decentralion:
> 3. How can I do a better job of including command line tests in the test plan test to verify that everything is working?

here's a proposed test plan (this is basically what I did):

Test plan:
The unit tests ensure that the command is properly wired into the sourcecred CLI, including help text integration. However, just to be safe, we can start by verifying that calling `sourcecred` without arguments lists the `clear` command as a valid option, and that calling `sourcecred help clear` prints help information.

The unit tests also ensure that the command removes the proper directories, so there isn't really a need to manually test it, although the reviewer may choose to do so to be safe.

Although out of scope for unit tests on this function, we can also do integration tests, to make sure that running the clear command doesn't leave the sourcecred directory in an invalid state from the perspective of the `load` command.

```js
$ yarn backend;
$ node bin/sourcecred.js load sourcecred/example-github;
$ node bin/sourcecred.js clear --cache;
$ node bin/sourcecred.js load sourcecred/example-github;
$ node bin/sourcecred.js clear --cache;
$ node bin/sourcecred.js load sourcecred/example-github;
```
The expected behavior of the above command block is that the load command never fails or throws an error.

src/cli/clear.test.js
```diff
@@ -0,0 +1,134 @@
...
+        stdout: [],
+        stderr: expect.arrayContaining([
+          expect.stringMatching(/^usage: sourcecred clear/),
+        ]),
+      });
+    });
+  });
+
+  describe("'makeClear' command", () => {
+    function throwError() {
```
decentralion: This should actually be returning a promise that rejects, rather than throwing an error, to satisfy the type signature, right?

decentralion: Also, I would move this declaration down to line 100 (just above the tests that use it) to make the code a bit easier to read.

decentralion:
Here's the workflow for rebasing:

Ensure we have the latest copy of the code:
```
$ git fetch origin/master;
```
 
Next, we interactively rebase your commits on top of the master branch. This will open a
text editor that shows all your commits in order, allowing you to re-arrange
them, remove some, squash them together, "fix them up" (modify them in place),
or reword their commit messages. In this case you have only one commit so you
could do without the -i flag, but using -i is the better habit because it makes
it easy to see what you're about to apply, and decide whether you want to clean
it up. In this case, we can save+exit the editor to go with the default plan (which is to
use our one commit).

 ```
$ git rebase -i origin/master;
 ```

Once you apply the commit, you will have conflicts.
Lets use git status to see the conflicts:

```
$ git status
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   package.json
	new file:   src/cli/clear.js
	new file:   src/cli/clear.test.js
	modified:   src/cli/help.test.js
	modified:   src/explorer/weights/WeightConfig.js
	modified:   yarn.lock

Unmerged paths:
  (use "git reset HEAD <file>..." to unstage)
  (use "git add <file>..." to mark resolution)

	both modified:   CHANGELOG.md
	both modified:   src/cli/help.js
	both modified:   src/cli/sourcecred.js
	both modified:   src/cli/sourcecred.test.js
```

Now you need to fix all of the unmerged paths. Git will show you the conflicts, demarcated by ">>>".
Here's an example from the CHANGELOG:

```
<<<<<<< HEAD
- Add the `pagerank` command (#1114)
- Add description tooltips for node and edge types in the weight configuration UI (#1081)
=======
- Add the `clear` command (#1111)
>>>>>>> 7f56ab7... Add CLI command to clear sourcecred data directory
```

On the top (HEAD) is the state of the master branch, at the bottom is the state in your commit.
We merge by changing the text to the desired state. In this case, we modify so that we have:

```
- Add the `clear` command (#1111)
- Add the `pagerank` command (#1114)
- Add description tooltips for node and edge types in the weight configuration UI (#1081)
```

Once we've fixed the file, we add it to mark the change as resolved:
$ git add CHANGELOG.md

For another example, we can look at help.js:

```
    analyze: analyzeHelp,
<<<<<<< HEAD
    pagerank: pagerankHelp,
=======
    clear: clearHelp,
>>>>>>> 7f56ab7... Add CLI command to clear sourcecred data directory
    "export-graph": exportGraphHelp,
  };
  if (subHelps[command] !== undefined) {
```

We update this to:

```
    analyze: analyzeHelp,
    pagerank: pagerankHelp,
    clear: clearHelp,
    "export-graph": exportGraphHelp,
  };
  if (subHelps[command] !== undefined) {
```

Once we've fixed all the conflicts in the file, we add it, as above.

Once all the changes are resolved: *do not make a new commit*. That will put you in a confusing state, since Git expects you are just merging changes that are already associated with a comimt. Instead, do `git rebase --continue`
to continue in the rebase.

Once it's done, you can push the new, rebased commit via `git push --force-with-lease`. Woo!

decentralion:
4. Out of scope for this PR, but it would be useful to be able to clear specific projects/directories and not be forced to re-download them all. E.g. if you've loaded Tensorflow and Sourcecred/PM, and you only want to reload Sourcecred/PM.

I think this is a great idea. Could you file an issue suggesting this and tag it "good first issue"? (And feel free to work on it too! But to make sure we don't lose track of it, let's file the issue.)

src/explorer/weights/WeightConfig.js
```diff
@@ -27,7 +27,6 @@ export class WeightConfig extends React.Component<Props> {
           style={{
             display: "flex",
             flexWrap: "wrap",
-            justifyContent: "space-between",
```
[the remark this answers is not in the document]
BrianLitwin: Updated

src/cli/clear.test.js
BrianLitwin: Updated (and moved declaration): 

```js
function throwError() {
 return Promise.reject(new Error("test error"));
}
```

BrianLitwin:
I thinkkk we are gtg @decentralion 

Thank you for the git workflow explanation @decentralion and @wchargin - both were very helpful !!

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Changelog
 
 ## [Unreleased]
+
 - Add the `pagerank` command (#1114)
+- Add the `clear` command (#1111)
 - Add description tooltips for node and edge types in the weight configuration UI (#1081)
+>>>>>>> fd5b21d... Add CLI command to clear sourcecred data directory
```
decentralion: Oops, remove this diff marker.

BrianLitwin: updated

PR #1112, opened 2019-03-08
Move default pagerank settings to pagerankGraph

decentralion:
This commit moves the default Pagerank options out of
`analysis/pagerank` and to `core/pagerankGraph`. This reflects the
gradual migration of core pagerank logic into `pagerankGraph`.

Test plan: `yarn test` should suffice. It's a trivial change.


decentralion:
Merging on self review, as the change is trivial.

PR #1114, opened 2019-03-08
Add `sourcecred pagerank` for backend pagerank

decentralion:
This commit adds a new CLI command, `pagerank`, which runs PageRank on a
given repository. At present, the command only ever uses the default
weights, although I plan to make this configurable in the future. The
command then saves the resultant pagerank graph in the SourceCred
directory.

On its own, this command is not yet very compelling, as it doesn't
present any easily-consumed information (e.g. users' scores). However,
it is the first step for building other commands which do just that. My
intention is to make running this command the last step of `sourcecred
load`, so that future commands may assume the existence of pagerank
scores for any loaded repository.

Test plan: The new command is thoroughly tested; see
`cli/pagerank.test.js`. It also has nearly perfect code coverage (one
line missing, the dependency-injected real function for loading graphs).

Additionally, the following sequence of commands works:
```
$ yarn backend
$ node bin/sourcecred.js load sourcecred/pm
$ node bin/sourcecred.js pagerank sourcecred/pm
$ cat $SOURCECRED_DIRECTORY/data/sourcecred/pm/pagerankGraph.json
```

Material progress on #967.

### DIFF
--- a/src/cli/pagerank.js
+++ b/src/cli/pagerank.js
@@ -1,0 +1,210 @@
+// @flow
+// Implementation of `sourcecred pagerank`.
+
+import fs from "fs-extra";
+import path from "path";
+
+import {Graph} from "../core/graph";
+import {
+  PagerankGraph,
+  DEFAULT_SYNTHETIC_LOOP_WEIGHT,
+  DEFAULT_CONVERGENCE_THRESHOLD,
+  DEFAULT_MAX_ITERATIONS,
+} from "../core/pagerankGraph";
+import {repoIdToString, stringToRepoId, type RepoId} from "../core/repoId";
+import dedent from "../util/dedent";
+import type {Command} from "./command";
+import * as Common from "./common";
+import stringify from "json-stable-stringify";
+import {loadGraph, type LoadGraphResult} from "../analysis/loadGraph";
+
+import {
+  type WeightedTypes,
+  combineWeights,
+  defaultWeightsForDeclaration,
+} from "../analysis/weights";
+import {weightsToEdgeEvaluator} from "../analysis/weightsToEdgeEvaluator";
+import type {IAnalysisAdapter} from "../analysis/analysisAdapter";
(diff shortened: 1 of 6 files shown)


src/cli/pagerank.js
```diff
@@ -0,0 +1,202 @@
...
+  adapters: $ReadOnlyArray<IAnalysisAdapter>
+): WeightedTypes {
+  const declarations = adapters.map((a) => a.declaration());
+  return combineWeights(declarations.map(defaultWeightsForDeclaration));
+}
+
+export const defaultAdapters = () => [
+  new GithubAnalysisAdapter(),
+  new GitAnalysisAdapter(),
+  new FallbackAdapter(),
```
decentralion: The astute reviewer will note that this is a little different from the other commands, in that we include the FallbackAdapter. This is needed for tests to pass, as otherwise the pagerank command would fail if a node was included in the graph that no plugin accounted for (as would be the case, for example, if the GitHub plugin were run on its own, as it creates git nodes without providing corresponding NodeTypes). Handling this case is why we have a fallbackAdapter to begin with.

In a future-but-soon PR, I plan to refactor the list of default plugins out of the individual commands and into a shared file, along with an interface that makes it clear what a plugin needs to package.

decentralion:
As currently written, this command blows heap if run on a large repository.
It can be remedied by running `node --max_old_space_size=8192 bin/sourcecred.js pagerank foo/bar`,
which is what the `load` command already does, under the hood, for all the sub-commands it calls out to.

May want to add a layer of indirection so that this just happens by default for this command.

decentralion:
Going to merge this on self-review, since it's been outstanding for a while and is important to the project.

On reflection, I decided not to add a layer of indirection to avoid the heap size issues. Since we are already planning to have this command auto-run on `sourcecred load` (see #1115), and that already does heap-size indirection, manually running this command is likely to be limited to power users (aka me) for the time being. Instead, I added info on the heap size flag to the usage string. If we later make this a command that end users use a lot, we can wrap it in indirection at that time. (Or we could just optimize Pagerank's memory usage.)

PR #1122, opened 2019-03-26
Fixes broken link README

s-ben:
This commit fixes three broken links (two in the README, one in the prototype app) that were still pointing to `https://discuss.sourcecred.io/`. 

Test plan:
Verify that there are no other bad links to the old Discourse location, by running git grep "discuss.sourcecred.io".

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Please also take a look at our [contributing guide].
 
 If you’re looking for a place to start, we’ve tagged some [good first issues].
 
-[forum]: https://discuss.sourcecred.io
+[forum]: https://discourse.sourcecred.io/ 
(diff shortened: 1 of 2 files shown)


decentralion:
Btw, this is a good example of why we value having each commit come with a test plan, as described in [our Contributing guide](https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#test-plan). In this case, I think a good test plan would be: 

_Verify that there are no other bad links to the old Discourse location, by running `git grep "discuss.sourcecred.io"`._

Running that test plan, I discover that there is one other bad link in the codebase that should get fixed.

s-ben:
Shield fixed. Running grep also found a broken link in the prototype app (the 'what is this?' link when you click on a project).

PR #1123, opened 2019-03-29
refactor: `load` uses dependency injection

decentralion:
This commit refactors the `sourcecred load` CLI command so that it uses
dependency injection, much like the testing setup #1110. This makes it
feasible to test "surface logic" of how the CLI parses flags and
transforms them into data separately from the "piping logic" of invoking
the right API calls using that data.

This is motivated by the fact that I have other pulls on the way that
modify the `load` command (e.g. #1115) and testing them within the
current framework is onerous.

Test plan:
This is a pure refactoring commit, which substantially re-writes the
unit tests. The new unit tests pass (`yarn test --full` is happy).

Note that `yarn test -full` also includes a sharness test that does an
E2E usage of `sourcecred load`
(see sharness/test_load_example_github.t), so we may be confident that
the command still works as intended.

### DIFF
--- a/src/cli/load.js
+++ b/src/cli/load.js
@@ -4,6 +4,8 @@
 import mkdirp from "mkdirp";
 import path from "path";
 
+import * as NullUtil from "../util/null";
+
 import * as RepoIdRegistry from "../core/repoIdRegistry";
 import {repoIdToString, stringToRepoId, type RepoId} from "../core/repoId";
 import dedent from "../util/dedent";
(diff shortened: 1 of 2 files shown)


decentralion:
Note to reviewer: This commit is kind of noisy. If you don't get a chance to look at it in the next couple days, no worries, I will just merge it on self-review.

decentralion:
Merging on self review.

PR #1129, opened 2019-04-19
refactor: group Pagerank args as PagerankParams

decentralion:
This straightforward refactor switches the core markov chain methods
that implement PageRank from accepting positional arguments to accepting
a named object of arguments. We plan to add additional features to the
PagerankGraph.runPagerank API, and this refactor will make testing those
features easier.

This is a port of [sourcecred/odyssey-hackathon#5] and
[sourcecred/odyssey-hackathon#6] to mainline SourceCred.

[sourcecred/odyssey-hackathon#5]: https://github.com/sourcecred/odyssey-hackathon/pull/5
[sourcecred/odyssey-hackathon#6]: https://github.com/sourcecred/odyssey-hackathon/pull/6

Test plan: Straightforward refactor; `yarn test` passes.

Paired with @mzargham

PR #1130, opened 2019-04-21
refactor args to findStationaryDistribution

decentralion:
In [#1128: Add support for seed vectors][#1128], we significantly
increase the number of arguments to
markovChain.findStationaryDistribution. To clean up the invocations, I
added a followon PR (#1129) which converts findStationaryDistribution to
use a `PagerankParams` object instead.

However, I think it will be cleaner to land the PagerankParams refactor
before adding new features in #1128, so I'm making this PR as
pre-cleanup.

Test plan: This is a trivial refactor. `yarn test` passes.

[#1128]: https://github.com/sourcecred/sourcecred/pull/1128

### DIFF
--- a/src/analysis/pagerank.js
+++ b/src/analysis/pagerank.js
@@ -19,7 +19,11 @@ import {
 
 import {scoreByConstantTotal} from "./nodeScore";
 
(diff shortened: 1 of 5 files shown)


decentralion:
Merging on self review.

PR #1131, opened 2019-04-21
Make PagerankGraph convergence options optional

decentralion:
Right now PagerankGraph requires that the user choose specific values
for maxIterations and convergenceThreshold when running PageRank.

I also rename `PagerankConvergenceOptions` to `PagerankOptions`.

The motivation is that I want to add future arguments to the same
options dict (e.g. alpha and the seed vector), so the rename is
appropriate, and allowing the options to be unset (and thus inherit
default values) will make the whole API much cleaner as I add more
options.

Test plan: Unit test added. `yarn test` passes.

### DIFF
--- a/src/core/pagerankGraph.js
+++ b/src/core/pagerankGraph.js
@@ -24,7 +24,7 @@ import {
 import {
   findStationaryDistribution,
   type PagerankParams,
-  type PagerankOptions,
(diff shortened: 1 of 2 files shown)


decentralion:
Merging on self review.

PR #1132, opened 2019-04-21
Port skeleton of Odyssey frontend

decentralion:
Port skeleton of Odyssey frontend

This commit integrates an bare skeleton of the odyssey frontend that we
implemented in the [odyssey-hackathon] repository. You can see the
working frontend that we are trying to port over at
[sourcecred.io/odyssey-hackathon/][scio].

The prototype in the other repository has some tooling choices which are
incompatible/redundant with decisions in our codebase (sass vs
aphrodite), and requires some tools not yet present here
(svg-react-loader). This commit includes the build and integration work
needed to port the prototype frontend into mainline SourceCred. The
frontend scaffold isn't yet integrated with any "real" Odyssey data.

One potential issue: right now, every page that is rendered from the
SourceCred homepage is contained within a [homepage/Page], meaning that
it has full SourceCred website styling, along with the SourceCred
website header. The [application][scio] also has a header. Currently, I
work around this by having the Odyssey UI cover up the base header (via
absolute positioning), which works but is hacky. We can consider more
principled solutions:

- Finding a way to specify routes which aren't contained by
[homepage/Page]; maybe by adding a new top-level route
[here][route-alternative].
- Unify the headers for the Odyssey viewer and the page as a whole
(sounds like inappropriate entanglement?)
- Have a website header and also an application header (sounds ugly?)

[homepage/Page]: https://github.com/sourcecred/sourcecred/blob/ee1d2fb996718fe41325711271542a54c197a1ed/src/homepage/Page.js
[route-alternative]: https://github.com/sourcecred/sourcecred/blob/ee1d2fb996718fe41325711271542a54c197a1ed/src/homepage/createRoutes.js#L17

Test plan: Run `yarn start`, and then navigate to
`localhost:8080/odyssey/`. observe that a working website is displayed,
and that the cred logo next to the word "SourceCred" is loaded properly
(i.e. svg-react-loader is integrated properly). Observe that there are
no build/compile errors from either `yarn start` or `yarn build`. Also,
observe that the UI looks passably nice, and that if the number of
elements in the entity lists is larger than can be displayed, the
sidebar pane scrolls independently.

The UI was tested in both Chrome and Firefox.

[odyssey-hackathon]: https://github.com/sourcecred/odyssey-hackathon
[scio]: https://sourcecred.io/odyssey-hackathon/

Thanks to @jmnemo, as the implementation is based on [his work].

[his work]: https://github.com/jmnemo/hackathon-event/


decentralion:
@wchargin: Please review this from a build/tooling perspective rather than worrying too much about the actual html/css. My intention is to let this be a "high velocity" part of the codebase and increase the quality discipline (e.g. unit testing) once it has settled a bit. (In other words, I'm willing to take on some technical debt.) Happy to talk about this tradeoff more here, or offline, if you like.

By the way, here's the current state of the UI:
image

decentralion:
I've polished this PR some more so that it actually merges the UI in a nice state.

Here's a screenshot in Chrome:
image

And here's a screenshot in Firefox: 
image

decentralion:
Merging on (light) self review.

PR #1134, opened 2019-04-24
Initial data model for the Odyssey plugin

decentralion:
This commit puts in a basic data model for the Odyssey plugin. It's
built around the `OdysseyInstance` class, which is basically a Graph
that keeps around descriptions for every node, and ensures that
nodes/edges are typed in accordance with the Odyssey plugin declaration.

In the future, I want to enable instances to declare their own node/edge
types, in which case the instance will assume responsibility for
tracking and serializing the types.

To make the Odyssey plugin a 'proper plugin', I've also added a plugin
declaration, as well as analysis and explorer adapters. I haven't
decided exactly where data for Odyssey instances should be stored, so
for now the plugin adapters always return an example instance which is
based on our experience at the Odyssey hackathon.

Test plan: The instance has unit tests for its logic.

If you want to see what the plugin looks like right now when it's
integrated, you can apply the following diff, and then load the
prototype. It will contain Odyssey nodes (find them using the node type
dropdown). Note that without a seed vector to move cred back to the
values/artifacts, the cred distribution in the Odyssey subgraph is
degenerate; the users are all sinks and have postiive cred scores, but
all the other nodes converge to 0 cred.

```diff
diff --git a/src/homepage/homepageExplorer.js b/src/homepage/homepageExplorer.js
index cae4548..48987f1 100644
--- a/src/homepage/homepageExplorer.js
+++ b/src/homepage/homepageExplorer.js
@@ -6,6 +6,7 @@ import type {Assets} from "../webutil/assets";
 import {StaticExplorerAdapterSet} from "../explorer/adapters/explorerAdapterSet";
 import {StaticExplorerAdapter as GithubAdapter} from "../plugins/github/explorerAdapter";
 import {StaticExplorerAdapter as GitAdapter} from "../plugins/git/explorerAdapter";
+import {StaticExplorerAdapter as OdysseyAdapter} from "../plugins/odyssey/explorerAdapter";
 import {GithubGitGateway} from "../plugins/github/githubGitGateway";
 import {AppPage} from "../explorer/App";
 import type {RepoId} from "../core/repoId";
@@ -14,6 +15,7 @@ function homepageStaticAdapters(): StaticExplorerAdapterSet {
   return new StaticExplorerAdapterSet([
     new GithubAdapter(),
     new GitAdapter(new GithubGitGateway()),
+    new OdysseyAdapter(),
   ]);
 }
```

### DIFF
--- a/src/plugins/odyssey/declaration.js
+++ b/src/plugins/odyssey/declaration.js
@@ -1,0 +1,83 @@
+// @flow
+
+import {EdgeAddress, NodeAddress} from "../../core/graph";
+
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {NodeType} from "../../analysis/types";
+
+export type OdysseyNodeTypeIdentifier =
+  | "ARTIFACT"
+  | "CONTRIBUTION"
+  | "VALUE"
+  | "PERSON";
+
+export type OdysseyEdgeTypeIdentifier = "DEPENDS_ON";
+
+export function isOdysseyNodeTypeIdentifier(
+  n: OdysseyNodeTypeIdentifier
+): boolean {
+  return ["ARTIFACT", "CONTRIBUTION", "VALUE", "PERSON"].indexOf(n) !== -1;
+}
+
+export const NODE_PREFIX = NodeAddress.fromParts(["sourcecred", "odyssey"]);
+
+export function isOdysseyEdgeTypeIdentifier(x: string): boolean {
+  return x === "DEPENDS_ON";
+}
+export const EDGE_PREFIX = EdgeAddress.fromParts(["sourcecred", "odyssey"]);
+
+const artifactNodeType: NodeType = Object.freeze({
+  name: "Artifact",
+  pluralName: "Artifacts",
+  prefix: NodeAddress.append(NODE_PREFIX, "ARTIFACT"),
+  defaultWeight: 2,
+  description:
+    "Represents a durably valuable piece of a project, e.g. a major subcomponent.",
+});
+
+const contributionNodeType: NodeType = Object.freeze({
+  name: "Contribution",
+  pluralName: "Contributions",
+  prefix: NodeAddress.append(NODE_PREFIX, "CONTRIBUTION"),
+  defaultWeight: 1,
+  description:
+    "Represents any specific work or labor that went into a project.",
+});
+
+const valueNodeType: NodeType = Object.freeze({
+  name: "Value",
+  pluralName: "Values",
+  prefix: NodeAddress.append(NODE_PREFIX, "VALUE"),
+  defaultWeight: 4,
+  description: "Represents a high-level value of the project.",
+});
+
+const personNodeType: NodeType = Object.freeze({
+  name: "Person",
+  pluralName: "People",
+  prefix: NodeAddress.append(NODE_PREFIX, "PERSON"),
+  defaultWeight: 1,
+  description: "Represents an individual contributor.",
+});
+
+const dependsOnEdgeType = Object.freeze({
+  forwardName: "depends on",
+  backwardName: "is depended on by",
+  prefix: EdgeAddress.append(EDGE_PREFIX, "DEPENDS_ON"),
+  defaultForwardWeight: 1,
+  defaultBackwardWeight: 0,
+  description: "Generic edge for flowing credit in the Odyssey plugin",
+});
+
+export const declaration: PluginDeclaration = Object.freeze({
+  name: "Odyssey",
+  nodePrefix: NODE_PREFIX,
+  edgePrefix: EDGE_PREFIX,
+  nodeTypes: [
+    contributionNodeType,
+    valueNodeType,
+    personNodeType,
+    artifactNodeType,
+  ],
+  edgeTypes: [dependsOnEdgeType],
+});
--- a/src/plugins/odyssey/instance.js
+++ b/src/plugins/odyssey/instance.js
@@ -1,0 +1,225 @@
+// @flow
+
+/**
+ * Core "model" logic for the Odyssey plugin.
+ * Basically allows creating a data store of priorities, contributions, and people,
+ * and compiling that data store into a cred Graph.
+ */
+import {
+  Graph,
+  EdgeAddress,
+  NodeAddress,
+  type NodeAddressT,
+  type GraphJSON,
+  sortedNodeAddressesFromJSON,
+} from "../../core/graph";
+
+import deepEqual from "lodash.isequal";
+
+import {
+  NODE_PREFIX,
+  EDGE_PREFIX,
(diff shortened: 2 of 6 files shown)


decentralion:
Here's what the example odyssey data looks like, when integrated into the existing explorer ui.
image

src/plugins/odyssey/declaration.js
```diff
@@ -0,0 +1,83 @@
+// @flow
+
+import {EdgeAddress, NodeAddress} from "../../core/graph";
+
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {NodeType} from "../../analysis/types";
+
+export type OdysseyNodeTypeIdentifier =
```
decentralion: Maybe add a TODO to the code that documents that we want to switch the node types into the instance itself (eventually, nonurgent while SC is primary consumer.)

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+  isOdysseyNodeTypeIdentifier,
+  type OdysseyEdgeTypeIdentifier,
+  isOdysseyEdgeTypeIdentifier,
+} from "./declaration";
+
+import {toCompat, fromCompat, type Compatible} from "../../util/compat";
+
+import * as NullUtil from "../../util/null";
+
+export type Node = {|
```
decentralion: TODO (not this PR): add weights (except the interpretation of a 'weight' here will probably be as a pagerank seed?)

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+ * [declaration file](./declaration.js) but we intend to allow instance-specified
+ * types in the future.
+ *
+ * The OdysseyInstance maintains an internal graph which actually stores the
+ * node identities, as well as added nodes. You can get a copy of this graph
+ * by calling the `.graph()` method.
+ *
+ * Entities are identified by an incrementing id (`._count`). This is convenient
+ * for implementation, although it will make reconciling simultaneous edits challenging.
+ * Once that becomes a real issue, we should consider switching to a different node/edge
```
decentralion: `s/consider switching/switch/`

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+  _descriptions: Map<NodeAddressT, string>;
+  _count: number;
+
+  /**
+   * Construct an Odyssey Instance.
+   */
+  constructor() {
+    this._graph = new Graph();
+    this._count = 0;
+    this._descriptions = new Map();
```
decentralion: nit: order properties the same in the declaration as in the class declaration

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  /**
+   * Add a new node to the instance (and a corresponding node to the graph).
+   *
+   * Requires a valid node type identifier (aka a string that uniquely identifies
```
decentralion: a string that uniquely identifies... see [declaration.js](./declaration.js)

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+   * Retrieve the Node corresponding to a given node address, if it exists.
+   */
+  node(address: NodeAddressT): ?Node {
+    if (!this._graph.hasNode(address)) {
+      return null;
+    }
+    const parts = NodeAddress.toParts(address);
+    // We know it is an OdysseyNodeTypeIdentifier because the instance's internal
+    // graph only has Odyssey nodes in it.
+    const nodeTypeIdentifier: OdysseyNodeTypeIdentifier = (parts[2]: any);
```
decentralion: for safety- maybe call the helper function and throw error if it's not?

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+    const description = NullUtil.get(this._descriptions.get(address));
+    return {address, nodeTypeIdentifier, description};
+  }
+
+  nodes(typeIdentifier?: OdysseyNodeTypeIdentifier): Iterator<Node> {
+    if (
+      typeIdentifier != null &&
+      !isOdysseyNodeTypeIdentifier(typeIdentifier)
+    ) {
+      throw new Error(`Invalid type identifier: ${typeIdentifier}`);
```
decentralion: I think it should allow bad identifiers and just not return any nodes. Safer if e.g. in the future we actually do have instances that have defined non-default types (and potentially stopped using default types)

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+      typeIdentifier == null
+        ? NODE_PREFIX
+        : NodeAddress.append(NODE_PREFIX, typeIdentifier);
+    return this._nodesIterator(prefix);
+  }
+
+  *_nodesIterator(prefix: NodeAddressT): Iterator<Node> {
+    for (const a of this._graph.nodes({prefix})) {
+      const node = NullUtil.get(this.node(a));
+      yield node;
```
decentralion: nit: collapse the variable declaration and the yield together

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+    }
+  }
+
+  // Cred for these edge flows from src to dst.
```
decentralion: Nit: convert comment to docstring

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+    }
+    const parts = NodeAddress.toParts(address);
+    // We know it is an OdysseyNodeTypeIdentifier because the instance's internal
+    // graph only has Odyssey nodes in it.
+    const nodeTypeIdentifier: OdysseyNodeTypeIdentifier = (parts[2]: any);
+    const description = NullUtil.get(this._descriptions.get(address));
+    return {address, nodeTypeIdentifier, description};
+  }
+
+  nodes(typeIdentifier?: OdysseyNodeTypeIdentifier): Iterator<Node> {
```
decentralion: nit: docstring

src/plugins/odyssey/instance.js
```diff
@@ -0,0 +1,226 @@
...
+    const sortedNodeAddresses = sortedNodeAddressesFromJSON(graphJSON);
+    for (let i = 0; i < sortedNodeAddresses.length; i++) {
+      descriptions.set(sortedNodeAddresses[i], sortedDescriptions[i]);
+    }
+    instance._descriptions = descriptions;
+    return instance;
+  }
+
+  /**
+   * Returns whether two Odyssey instances are equal.
```
decentralion: consider renaming this from `equals` to `historicallyIdentical` because it's really ... not... equality

decentralion:
Addressed all my self-review comments. Merging :)

PR #1135, opened 2019-04-24
Add support for seed vectors to PagerankGraph

decentralion:
This commit modifies `PagerankGraph.runPagerank` so that the user can
provide an alpha and seed vector. The seed vector is specified via a map
of weights, which will be normalized into a probability distribution
over all the nodes in the graph. In the event that the map is empty (or
the total weight is otherwise 0), a uniform distribution is created.

To effect this change, a helper function called `weightedDistribution`
has been added (and thoroughly tested) in the `graphToMarkovChain`
module. Then, that function is used in `pagerankGraph.runPagerank`
(along with light testing).

Currently, the default alpha is set to 0, to ensure consistency with the
legacy pagerank implementation in `analysis/pagerank`. Once that has
been replaced with `PagerankGraph`, we can consider changing the defualt
alpha to non-zero (thus removing the need for synthetic self-loops).

I took a different approach in the [odyssey-hackathon repo][commit].
The previous approach was a much more complicated (and fairly redundant)
API, that allowed specifying "NO_SEED", "UNIFORM_SEED", "SELECTED_SEED",
and "SPECIFIED_SEED". I'm much happier with this API and implementation.

[commit]: https://github.com/sourcecred/odyssey-hackathon/commit/ed07861073f1a5c803a378b39515b338b80693eb

Test plan: Unit tests included; run `yarn test`.

### DIFF
--- a/src/core/attribution/graphToMarkovChain.js
+++ b/src/core/attribution/graphToMarkovChain.js
@@ -1,7 +1,11 @@
 // @flow
 
-import {type Edge, type Graph, type NodeAddressT} from "../graph";
-import type {Distribution, SparseMarkovChain} from "./markovChain";
+import {type Edge, type Graph, type NodeAddressT, NodeAddress} from "../graph";
+import {
+  type Distribution,
+  type SparseMarkovChain,
+  uniformDistribution,
+} from "./markovChain";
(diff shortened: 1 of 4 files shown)


decentralion:
Merging on self-review.

Issue #39, opened 2018-02-28
design a graph visualizer


decentralion:
Since SourceCred's main data structure is a graph, we're going to want to build a tool for visualizing it.

### fractal graph, meaningful at different levels of scale

the graph will span levels of abstraction from the repository and project level (visualize GitHub projects as nodes, edges are dependencies between projects) down to the individual commit and comment level (a pull request is a node with edges to all its comments, and edges to its commits, and to the files it modifies).

the graph visualizer should be able to seamlessly transition between these levels of scale, and explore the boundaries in between (e.g. the specific artifacts that create connections across repository boundaries)

### some nodes will have very high degree

for example, user nodes will have an edge to every contribution and comment they have ever authored. we'll need to abstract over cases like this, rather than overwhelm the display

### some nodes are much more important than others

so we will want to surface highest-importance nodes within the viewing context, and maybe aggregate less important nodes together. fortunately, we will have a metric (cred) for determining the relative importance of nodes (that's the whole point! 😄)

### heterogenous node and edge types

the nodes and edges are provided by plugins (see: #26) and each plugin may have its own logic for how to display that node or edge. the UI will need a plugin area that provides this context.

## my current vision

What I'm imagining right now is a zoom-and-pannable "map" depicting nodes and edges. To get some inspiration, please take a look at the [TensorFlow graph explorer].

At any time, there is an active "scope" that determines which part of the graph is being visualized; the scope can be global (show most important K repos, and edges between them), repo specific (show most important K projects and contributions, with edges between them), user specific (show most important contributions that user has made, and edges...), etc.

For any given scope, most of the graph will be hidden, either by being abstracted into aggregate nodes, or just elided from view. Plugins should be able to define their own scoping logic; for example, the GitHub plugin may define comments get scoped with their containing issue or pull request. 

The user should be able to interactively explore the graph, both by drilling down into scopes (e.g. click on an aggregated node to expand it an explore its contents), and by traversing the graph to different locations (e.g. follow a path of package dependencies to core deps). 

In contrast to the TensorFlow graph explorer, I don't expect that there will be a canonical fixed layout that the whole graph fits into; instead, the layout is determined dynamically, possibly via [force directed layout](https://bl.ocks.org/mbostock/4062045) or similar.

[...]

PR #1139, opened 2019-05-12
Graph Visualizer: Add NodeRenderer

decentralion:
This commit is the first substantive step towards implementing a graph visualizer (#39).
We're basically porting the hackathon graph visualizer seen [here](https://sourcecred.io/odyssey-hackathon/) into the proper SourceCred codebase.

This commit adds a `NodeRenderer` class for visualizing nodes in the
SourceCred graph. The NodeRenderer is a react component which maintains
a d3 selection under the hood. This approach allows us to build the
visualizer around a React architecture, while still benefiting from d3
transitions. See [this blog post] for details.

[this blog post]: https://medium.com/@sxywu/on-d3-react-and-a-little-bit-of-flux-88a226f328f3

Rather than writing unit tests, I wrote an 'Inspection Test',
which is basically a visual component which drives all the features of
the NodeRenderer, and allows manual inspection.

As compared to a unit test, it has several advantages:
- Allows testing transitions
- Easier/faster to write
- Since the output is visual, visual inspection can be better at finding
bugs than unit testing

Its main disadvantage is that since it's a manual test, people might
forget to run it, and inadvertently break the application. However,
since the visualizer is quite self-contained (few dependencies into the
rest of the application), I don't think this is a big concern.

Test plan: Run the inspection test (by running `yarn start` and then
navigating to `/test/NodeRenderer/`), and manually verify the behavior
is correct.

Thanks to @sxywu for the blog post which inspired this design.


decentralion:
Closing as inactive.

PR #1143, opened 2019-05-18
explorer: tweak weights on a per-node basis

decentralion:
This pull request adds a weight slider to every NodeRow in the explorer,
enabling the user to manually set a weight for that node. The weights are
multiplicative with the type level weights, so that they can be changed
independently (e.g. you can have a comment that is weighted 2x higher than
regular comments, but still have comments get a low weight in general).

This pull coordinates a number of different changes across the codebase, all of
which are tested:

- Adding support for manual weights in the `weights` and
`weightsToEdgeEvaluator` modules.
- Modifying `pagerankTable.TableRow` so that it can show a slider in the second
column.
- Adding piping for manual weights into the `PagerankTable` shared props, and
into the explorer app
- Adding the slider to the `NodeRow` class that displays the current weight,
and can trigger the upstream weight change
- Ensuring that the `runPagerank` call in the explorer actually uses the manual
weights

At present, there is no way to save these weights (they are ephemeral in the
frontend) and so this is clearly a prototype/tech demo level feature rather
than being ready for real usage. Correspondingly, CLI pagerank command always
uses an empty set of manual weights. I plan to remedy this in a follow-on pull
request.

Test plan: Run the included unit tests (`yarn test`) and also spin up the UI,
verify that it visually looks good in both Firefox and Chrome, and verify that
changing the weights and then re-running PageRank actually causes the cred of
the modified node to change.

Review plan: In addition to carefully reading the code, ensure that all of the
changes described a few paragraphs up are actually tested.

Merge plan: Squash and merge.

Thanks to @s-ben for proposing this feature in Discord, and to everyone
discussing its implications in [this Discourse thread].

[this Discourse thread]: https://discourse.sourcecred.io/t/maintainers-user-stories-discussion/110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Allow tweaking weights on a per-node basis (#1143)
 - Add the `pagerank` command (#1114)
 - Add the `clear` command (#1111)
 - Add description tooltips for node and edge types in the weight configuration UI (#1081)
--- a/src/analysis/weights.js
+++ b/src/analysis/weights.js
@@ -19,6 +19,8 @@ export type WeightedTypes = {|
   +edges: Map<EdgeAddressT, WeightedEdgeType>,
 |};
 
+export type ManualWeights = Map<NodeAddressT, number>;
+
 export function defaultWeightedNodeType(type: NodeType): WeightedNodeType {
   return {
     type,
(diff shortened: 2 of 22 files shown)


decentralion:
merging on self review.

decentralion:
Note: when i squashed and merged this PR, to my surprise, it made a commit whose description was the concatenation of all squashed commit messages, rather than the PR description I provided.

I didn't like this, so I force-pushed a re-written commit message onto master. As such, this PR really merged as https://github.com/sourcecred/sourcecred/commit/d2559960bb9a9e5cc8b7c01e70ccab2a269545d4

PR #1147, opened 2019-05-20
Add JSON serialization support for weights

decentralion:
Test plan: New unit tests added.

### DIFF
--- a/src/analysis/weights.js
+++ b/src/analysis/weights.js
@@ -1,6 +1,8 @@
 // @flow
 
+import * as MapUtil from "../util/map";
 import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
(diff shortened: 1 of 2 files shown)


decentralion:
Fun fact: This pull request was an example of "make the hard thing easy (warning: may be hard), then do the easy thing". I initially started work on this PR prior to #1146, and it was really ugly, because I needed to implement special logic in the serializer to filter out default weights, so that when users didn't change any weights, they would continue to use the default weights, even if the defaults changed. (And this meant there was no way for a user to force the weight to stay at a value equal to its defaults). The issue was that the data structure was storing every weight, not every user choice.

After the refactor, adding this with the right logic is totally trivial--we just save an object that represents all of the user's explicit choices.

PR #1161, opened 2019-05-30
Quicker failure and description when invalid token supplied

### LINKED ISSUE #1156: `sourcecred load` fails slowly on bad credentials
When an invalid GitHub token is provided, SourceCred takes a very long time to fail to load the GitHub plugin. Eventually, it produces this output:

```
$ export  SOURCECRED_GITHUB_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
$ node bin/sourcecred.js load sourcecred/research 
sourcecred/git/remotes/origin/HEAD 

Starting tasks
  GO   load-git
  GO   load-github
 DONE  load-git
 FAIL  load-github
Exit code: 1
Contents of stderr:
    GitHub query failed! We're tracking these issues at https://github.com/sourcecred/sourcecred/issues/350.
    If the error is a timeout or abuse rate limit, you can try loading a smaller repo, or trying again in a few minutes.
    The actual failed response can be found below:
    =================================================
    { message: 'Bad credentials',
      documentation_url: 'https://developer.github.com/v4' }
    undefined


Overview
Failed tasks:
  - load-github
Final result:  FAILURE 
```

In fetchGithubRepo, we define a number of query failure types and whether or not we should retry:

https://github.com/sourcecred/sourcecred/blob/61627531bbfcd4a3d07796150f145cc318608c65/src/plugins/github/fetchGithubRepo.js#L81-L86

However, we do not consider the case where a bad GitHub API token has been provided. I believe this results in failing through to this section: 

https://github.com/sourcecred/sourcecred/blob/61627531bbfcd4a3d07796150f145cc318608c65/src/plugins/github/fetchGithubRepo.js#L128-L134

We should fix this part of the codebase so that we identify when the error message is `'Bad credentials'` and print a helpful error message (and do it quickly, by setting `retry: false`.

Thanks to @nothingismagick, who first encountered this bug.

tylerdmace:
Fixes #1156

When users export a GitHub API token that has insufficient privileges
or has been revoked, we have been using a catch all error with retry
to handle it. This change adds a new error type for bad credentials
and does not retry.

Test plan:
There are no unit tests that cover this, however, you can test the
change by supplying a revoked token and attempting to load a GitHub
repo.

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -83,6 +83,7 @@ type GithubResponseError =
   | {|+type: "GRAPHQL_ERROR", retry: false, error: mixed|}
   | {|+type: "RATE_LIMIT_EXCEEDED", retry: false, error: mixed|}
   | {|+type: "GITHUB_INTERNAL_EXECUTION_ERROR", retry: true, error: mixed|}
+  | {|+type: "BAD_CREDENTIALS", retry: false, error: mixed|}
   | {|+type: "NO_DATA", retry: true, error: mixed|};
 
 // Fetch against the GitHub API with the provided options, returning a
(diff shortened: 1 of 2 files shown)


src/plugins/github/fetchGithubRepo.js
```diff
@@ -126,10 +127,20 @@ function tryGithubFetch(fetch, fetchOptions): Promise<any> {
           }
         }
         if (x.data === undefined) {
-          // See https://github.com/sourcecred/sourcecred/issues/350.
-          return Promise.reject(
-            ({type: "NO_DATA", retry: true, error: x}: GithubResponseError)
-          );
+          // See https://github.com/sourcecred/sourcecred/issues/350
```
decentralion: I think this comment is meant to be bound to the `({type: "NO_DATA", retry: true, ...` line, as #350 essentially justifies why we would try to retry in that particular case. Could you move the comment down a few lines so it's still connected to that line?

tylerdmace: Good call! Fix incoming.

PR #1165, opened 2019-05-30
Add creationTimeline

decentralion:
This has the methods for organizing the creation history of a graph into
fixed-time-width intervals where each interval is associated with the
nodes created in that time slice.

Test plan: Unit tests included, `yarn test` passes.


decentralion:
Obsoleted by #1212

Issue #1158, opened 2019-05-28
Stop storing the Git graph on disk


decentralion:
Right now, when the Git plugin is loaded, it saves two files to disk. One contains a serialized Git repository type:
https://github.com/sourcecred/sourcecred/blob/61627531bbfcd4a3d07796150f145cc318608c65/src/plugins/git/types.js#L5-L10

The other is a serialized graph.

We can see the two files being saved here:

https://github.com/sourcecred/sourcecred/blob/61627531bbfcd4a3d07796150f145cc318608c65/src/plugins/git/loadGitData.js#L18-L31

This is a little redundant, because we can already create a Git graph from the repository:
https://github.com/sourcecred/sourcecred/blob/61627531bbfcd4a3d07796150f145cc318608c65/src/plugins/git/createGraph.js#L9-L13

Therefore, instead of persisting both the graph and repository to disk, it would suffice to persist just the Repository, and then create the Graph from it when required. 

The first step will be to modify loadGitData so that it only saves the repository.json. Of course, that will also break the application, because currently the AnalysisAdapter (plugin adapter for cred calculations) and ExplorerAdapter (plugin adapter for the frontend UI) depend on loading graph.json.

In the AnalysisAdapter load method, we should change the following section of code:
https://github.com/sourcecred/sourcecred/blob/e01247a64264b1e55ce8112f6990e31aa16ded1e/src/plugins/git/analysisAdapter.js#L36-L41

Instead of loading graph.json there, we can just load the repository, and then invoke the createGraph method on that repository.

Similarly, in the explorer adapter, we can delete the code for loading the graph, and instead call createGraph on the repository after it is loaded.

https://github.com/sourcecred/sourcecred/blob/e01247a64264b1e55ce8112f6990e31aa16ded1e/src/plugins/git/explorerAdapter.js#L25-L48

After the change has been made, we can test it in the following two ways:

1. Manually
```shell
# rebuild SourceCred
$ yarn backend
# Ensure that no out-dated files are around
$ node bin/sourcecred.js clear --all
# check that load still works (this includes using the analysis adapter for pagerank)
$ node bin/soureccred.js load sourcecred/example-github 
$ yarn start 
# navigate to http://localhost:8080/prototype/sourcecred/example-github/
# and make sure that it still works
```
2. Automatically
```shell
$ export SOURCECRED_GITHUB_TOKEN=acd....0123
# Update the loaded snapshots to reflect the change to the stored data
[...]

PR #1166, opened 2019-05-30
WIP: Enhancement: Stop storing the Git graph on disk #1158

pkafei:
Git plugin no longer stores Git graph on disk. In relation to issue #1158.


src/plugins/git/explorerAdapter.js
```diff
@@ -44,6 +36,7 @@ export class StaticExplorerAdapter implements IStaticExplorerAdapter {
     const [graph, repository] = await Promise.all([
```
decentralion: we can remove the `Promise.all` since now there's only one promise.

`const repository = await loadRepository()`

src/plugins/git/explorerAdapter.js
```diff
@@ -44,6 +36,7 @@ export class StaticExplorerAdapter implements IStaticExplorerAdapter {
...
       loadGraph(),
       loadRepository(),
+      createGraph()
```
decentralion: createGraph takes a repository as an argument.

pkafei:
Thanks for reviewing the changes! I ran `scripts/update_snapshots.sh` but the `rc/plugins/git/analysisAdapter.test.js` is failing and is outputting this error message:

`Error: ENOENT: no such file or directory, open 'sharness/__snapshots__/example-github-load/data/sourcecred/example-github/git/graph.json'`

I am not sure if I'm getting this error because graph.json no longer exists.

decentralion:
@pkafei, it looks like there's a test in git/analysisAdapter.test.js that explicitly references the graph.json. It will need to be updated:

https://github.com/sourcecred/sourcecred/blob/1d32141f7640091223cffaf67d0a57287f3fbd8e/src/plugins/git/analysisAdapter.test.js#L27-L42

We can update the test to verify that it has the correct graph, with logic like:

```js
// modify the loading code so we load the repository instead of the graph
// from disk. Then:
const repository = JSON.parse(repositoryBufer.toString()
const actualGraph = aa.graph()
const expectedGraph = createGraph(repository)
expect(actualGraph.equals(expectedGraph)).toBe(true)
```

decentralion:
I'm going to close this pull request as inactive. @pkafei, you're more than welcome to reach out if you want to keep hacking on SourceCred. :)

PR #1167, opened 2019-05-30
README & CONTRIBUTING updates

tylerdmace:
Nothing major here -- added a note about SSH keys, about Changelog update format, and some misc formatting to make things a tad easier to read.


CONTRIBUTING.md
```diff
@@ -124,7 +124,7 @@ This is the same set of tests that is run on our CI system, CircleCI.
 If your patch makes a change that would be visible or interesting to a
 user of SourceCred—for example, fixing a bug—please add a description of
 the change under the `[Unreleased]` heading of `CHANGELOG.md`. Your new
-change should be the first entry in the section.
+change should be the first entry in the section. The format of your entry should be: ```<description of change> (#<PR number>)```.
```
decentralion: Nit (here and elsewhere): could we keep this document wrapped consistently? (I believe it's 80 chars)

README.md
```diff
@@ -43,36 +44,40 @@ First, make sure that you have the following dependencies:
...
-yarn install
-yarn backend
-export SOURCECRED_GITHUB_TOKEN=YOUR_GITHUB_TOKEN
-node bin/sourcecred.js load REPO_OWNER/REPO_NAME
-# this loads sourcecred data for a particular repository
-# it can take a few mins to run and will exit when finished
-yarn start
-# then navigate to localhost:8080 in your browser
+```Bash
+$ git clone https://github.com/sourcecred/sourcecred.git
```
wchargin: The downside to including the prompt tokens is that you can no longer
copy and paste the sequence of commands into their shell. Omitting the
prompt is therefore pretty common (at a glance: [Flow], [TensorBoard],
[TensorFlow], and [fzf] all do, and those are the first that I tried).
Suggest reverting this change.

[fzf]: https://github.com/junegunn/fzf#using-homebrew-or-linuxbrew
[Flow]: https://github.com/facebook/flow#installing-flow-per-project
[TensorBoard]: https://github.com/tensorflow/tensorboard#usage
[TensorFlow]: https://github.com/tensorflow/tensorflow#installation

README.md
```diff
@@ -43,36 +44,40 @@ First, make sure that you have the following dependencies:
...
-cd sourcecred
-yarn install
-yarn backend
-export SOURCECRED_GITHUB_TOKEN=YOUR_GITHUB_TOKEN
-node bin/sourcecred.js load REPO_OWNER/REPO_NAME
-# this loads sourcecred data for a particular repository
-# it can take a few mins to run and will exit when finished
-yarn start
-# then navigate to localhost:8080 in your browser
+```Bash
```
wchargin: nit: It’s ```` ```Bash ````, not ```` ```bash ````.

CONTRIBUTING.md
wchargin: Yes, please, but it’s wrapped at 72 characters, not 80.

README.md
```diff
@@ -31,6 +31,7 @@ First, make sure that you have the following dependencies:
   - Install [Node] (tested on v8.x.x).
   - Install [Yarn] (tested on v1.7.0).
   - Create a [GitHub API token]. No special permissions are required.
+  - Ensure that you have an SSH key (```id_rsa```) for ```git``` to use. On unix-like systems, this is usually stored in ```~/.ssh/```. If you do not, you can generate one by following these [instructions](https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent).
```
wchargin: nit: Use `` `single backticks` `` for code spans.

wchargin: Hmm, @decentralion, didn’t we try to make it unnecessary to have SSH
keys unless you’re accessing a private repository? Did that regress?

(No action required for this change if you do actually lways need keys.)

wchargin: Link text should be descriptive for accessibility; links like “this
link” or “here” or “instructions” lack context. Rephrase—perhaps:

```markdown
> If you don’t have an SSH key, you can follow [GitHub’s guide to
> SSH keys][ssh-keys] to set one up.

[ssh-keys]: https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent
```

wchargin: nit: “Unix-like” is capitalized.

decentralion: I probably induced this regression in https://github.com/sourcecred/sourcecred/pull/1085/files, specifically by changing our clone from https to git:
https://github.com/sourcecred/sourcecred/blob/4b2dcfa0ff87cde524b35398c8df12c822f4ec96/src/plugins/git/cloneAndLoadRepository.js#L19

Not sure what the best solution is. We should avoid needing extra steps for users to try out SourceCred, so fixing the bug will be important. but, it would be annoying to have an extra flag for cloning private repositories.

maybe we could have it try cloning with https, and then if it fails re-try with git, and print appropriate error messages depending on what happens? or too hacky?

CONTRIBUTING.md
tylerdmace: How do you all generally enforce this? Is there something in place that will do this for all files or should I be manually doing this using something like `fold`?

README.md
tylerdmace: If I had to opt for one option over the other, I'd go with flags over keys every time. As it stands right now, too, we are forcing key gen (or at least a check for some people) for everyone no matter the type of repo they are running SourceCred against. A flag might be an extra step but it's an extra step for a smaller number of people (only those who need to run against a private repo).

README.md
tylerdmace: Good call. Just getting them into a `bash` block will look better than what we had which was really all I was trying to do there.

README.md
decentralion: Yep, I agree. I wrote a detailed description along with an implementation plan here: https://github.com/sourcecred/sourcecred/issues/1168. As a very first step, we can just swap back from git to https; AFAIK no one except me has ever used the private git loading feature, so I'm fine privileging the common case over the exceptional case. @tylerdmace would you be willing to take this on? (either Just Fixing it hackily, or adding the flag.)

For now, let's keep this change to the README (since it is accurate information for SourceCred as it presently exists) and let's plan to take it back out once we fix the underlying issue.

CONTRIBUTING.md
decentralion: It would be great to have some tooling that does this automatically. (Like prettier but for Markdown?) Personally, I code/write in vim, and re-wrapping a paragraph is easy as `gqap`

README.md
tylerdmace: Absolutely, I can take it!

README.md
```diff
@@ -5,91 +5,114 @@
 [![Discord](https://img.shields.io/discord/453243919774253079.svg)](https://discord.gg/tsBTgc9)
 
 SourceCred creates reputation networks for open-source projects.
-Any open-source project can create its own _cred_, which is a reputational metric showing how much credit contributors deserve for helping the project.
-To compute cred, we organize a project’s contributions into a graph, whose edges connect contributions to each other and to contributors.
-We then run PageRank on that graph.
+Any open-source project can create its own _cred_, which is a
```
wchargin: Please don’t reformat the rest of the file with unrelated changes; this
kills the blame. It looks like `README.md` uses one-sentence-per-line
wrapping, so within this file your additions should do the same and your
changes shouldn’t reflow other lines.

(The purpose of one-sentence-per-line is made editing and diffing
easier; reflowing breaks that.)

tylerdmace: So should `README.md` not  adhere to the 72-character wrap then? I'm a little confused. The reason I made these changes was because I thought both files needed to be limited in width and I noticed that a number of lines in `README.md` fell outside that constraint.

wchargin: The important thing is to be consistent with the existing formatting of
the files that you edit. Occasionally, these standards are themselves
inconsistent across files. Some of our shell scripts have 2-space
indents; some have 4-space indents. The `CONTRIBUTING.md` file
hard-wraps at column 72, while `README.md` wraps at the end of every
sentence instead of a fixed column. This may not be ideal, but it’s
better to remain locally consistent and accept the global inconsistency
than it is to reblame the entire file on an unrelated change.

We would only make these files globally consistent if we also configured
a lint rule to verify this prior to merging any PRs, and preferably an
autoformatter so we don’t have to think about this.

So: `README.md` should indeed not adhere to the 72-character wrapping
found in `CONTRIBUTING.md`.

tylerdmace: Awesome, thank you for the clarification. I'll be sure to pay attention to these details going forward.

wchargin: Thanks, and sorry for the confusion—having to explain this at all is a
big part of the reason that we love autoformatters. :-)

tylerdmace:
I've reverted the wrap changes made to `README.md` and believe we're ready for approval/merge now.

README.md
decentralion: I think the intent was to remove the `$` tokens, but that revert doesn't seem to have made it into the change.

tylerdmace:
@decentralion Good catch; that revert brought back the tokens and also the use of triple back ticks over single. I've fixed both.

decentralion:
Thanks @tylerdmace! Unfortunately, it looks like we can't merge it quite yet, due to conflicts. :( Please rebase it on master.

(To make things simpler: I would consider first squashing the commits together via an [interactive rebase], and then rebasing the single squashed commit on top of updated `origin/master`, and then force-pushing.)

decentralion:
Note: as of #1210 I'm making the Git plugin not load by default, so we can remove the bit about setting up ssh keys. (Getting a fix so that the Git plugin uses https:// by default would still be nice, since I plan to keep the Git plugin around--I will re-enable it once we are using the data, e.g. to track file level cred. However, that fix is now less urgent.)

Issue #1136, opened 2019-04-24
Refactor graph to support limited node metadata


decentralion:
updated plan
---

I intend to refactor the Graph so that rather than storing nodes only by identity (`NodeAddressT`), we have a `Node` type with a few fields, such as:

```
type Node = {|
  +address: NodeAddressT,
  +timestampMs: number | null,
  +description: string,
|}
```

This is motivated by the fact that a few extra fields (e.g. timestamp) are needed for generally computing cred, and packaging them together will greatly simplify data plumbing for SourceCred. See [this comment for justification and discussion of tradeoffs](https://github.com/sourcecred/sourcecred/issues/1136#issuecomment-498078988).

This involves a major refactor to the Graph class. In particular:
- We need to modify the Graph so nodes are stored as objects.
- We need to ensure that adding a node produces an error if the contents conflict (but is idempotent if the node is identical)
- We need to ensure that it is legal to add dangling edges (so that e.g. the GitHub plugin can add an edge to a Git node without having sufficient information to add it to the graph)
- We need to ensure that dangling edges are handled appropriately (e.g. they should be wholly ignored for PageRank purposes)
- We need to ensure that Graph.merge throws an error if node contents conflict
- We need to ensure that JSON serialization of node contents and of dangling edges work as expected

original issue
---
I'm interested in modifying the Graph class so that every node is stored along with a markdown description.

This would be convenient from the perspective of the current implementation. We could, for example, more-or-less remove the Dynamic Explorer adapter (its other function, providing the graph, is getting obsoleted by generating PagerankGraphs on the backend (#967)). It would also save us from writing new Graph wrappers that mostly serve to add descriptions to nodes, e.g. the OdysseyInstance in #1134.

Naturally, "it would be convenient right now" is not sufficient justification to add a field to the core data structure (more than just "add a field": it  increase the number of fields from 0 to 1). However, I do think that having a description built-in to the graph would be of general value. Any time that we want to be able to introspect on the graph, or debug or interpret scores, having info on what the nodes represent will be valuable. Also, using Markdown gives a lot of flexibility, since it supports hyperlinks, formatting, etc. 

Basically, having seen that the first two frontends to SourceCred (explorer and odyssey-ui) both need descriptions, I suspect that they all will, and if so, putting descriptions directly in the graph will make things a lot simpler and more consistent from an implementation perspective.

The downsides to this approach include:
- potential for enormous graph bloating
- waste in the cases where we don't need introspection (pure score calculation)
[...]

PR #1172, opened 2019-06-03
(metadata-2) Refactor Graph to support object Nodes

decentralion:
Presently, the base Graph tracks nodes as `NodeAddressT`s, without the
possibility of storing any metadata with the node.

This commit changes that: now the Graph has a `Node` type, which is an
object. At present, that type only contains an address; it will be
extended in future commits.

See #1136 for some discussion of this approach, and in particular [this
comment] for justification.

[this comment]: https://github.com/sourcecred/sourcecred/issues/1136#issuecomment-498078988

Because this changes the basic APIs in Graph, it churns the entire
codebase. To minimize future churn, I've defined helper functions,
`partsNode` and `node`, in `graphTestUtil`. These helper functions
create simple dummy nodes in situations where we don't much care about
the nodes' fields. This way, as we add new fields we won't need to
update all of the test code again.

I removed the short example from the Graph module docstring. My
reasoning was, we aren't actually typechecking that code, so it could go
out of date as we add fields to Graph. I'd rather keep the docstring
more high-level for now and ensure its sync'd.

Test plan: `yarn test --full` passes. Check the documentation on the
Graph module, and take a look at snapshot changes.

### DIFF
--- a/src/analysis/loadGraph.test.js
+++ b/src/analysis/loadGraph.test.js
@@ -16,6 +16,7 @@ import type {
 import * as RepoIdRegistry from "../core/repoIdRegistry";
 import {makeRepoId, type RepoId} from "../core/repoId";
 import {loadGraph} from "./loadGraph";
+import {node} from "../core/graphTestUtil";
 
 const declaration = (name) => ({
   name,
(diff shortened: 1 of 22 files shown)


decentralion:
Obsoleted by #1188.

PR #1177, opened 2019-06-05
tools: make `update_snapshots.sh` reentrant

wchargin:
Summary:
General cleanup to `update_snapshots.sh`, primarily such that it is free
of race conditions and does not clobber your build output directory. The
mechanism is the same as used in `yarn test`, via `SOURCECRED_BIN`.

Test Plan:
Remove your build directories (`rm -r ./bin ./build`), then run this
script (`./scripts/update_snapshots.sh`) and check that each subprocess
is properly invoked and all tests pass. Check that the temporary build
directory is cleaned up.

wchargin-branch: update-snapshots-cleanup

### DIFF
--- a/scripts/update_snapshots.sh
+++ b/scripts/update_snapshots.sh
@@ -6,16 +6,33 @@ set -eu
 toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
 cd "${toplevel}"
 
-yarn backend
-echo "Updating GitHub graphql flow types"
-node ./bin/generateGithubGraphqlFlowTypes.js > src/plugins/github/graphqlTypes.js
-echo "Updating for sharness/test_load_example_github.t"
+tmpdir="$(mktemp -d)"
+cleanup() {
+    rm -r "${tmpdir}"
+}
+trap cleanup EXIT
+
+SOURCECRED_BIN="${tmpdir}/bin"
+yarn run --silent backend --output-path "${SOURCECRED_BIN}"
+export SOURCECRED_BIN  # for Sharness and shell tests
+export NODE_PATH="${toplevel}/node_modules${NODE_PATH:+:${NODE_PATH}}"
+
+echo "Updating GitHub GraphQL Flow types"
+cp .prettierrc.json "${SOURCECRED_BIN}/"
+node "${SOURCECRED_BIN}/generateGithubGraphqlFlowTypes.js" \
+    >src/plugins/github/graphqlTypes.js
+
+echo "Updating sharness/test_load_example_github.t"
 (cd sharness; UPDATE_SNAPSHOT=1 ./test_load_example_github.t -l)
+
 echo "Updating git/loadRepositoryTest.sh"
-./src/plugins/git/loadRepositoryTest.sh -u
+./src/plugins/git/loadRepositoryTest.sh -u --no-build
+
 echo "Updating github/fetchGithubOrgTest.sh"
 ./src/plugins/github/fetchGithubOrgTest.sh -u --no-build
+
 echo "Updating github/fetchGithubRepoTest.sh"
 ./src/plugins/github/fetchGithubRepoTest.sh -u --no-build
+
 echo "Updating Jest snapshots"
 yarn unit -u


scripts/update_snapshots.sh
```diff
@@ -6,16 +6,33 @@ set -eu
...
+}
+trap cleanup EXIT
+
+SOURCECRED_BIN="${tmpdir}/bin"
+yarn run --silent backend --output-path "${SOURCECRED_BIN}"
+export SOURCECRED_BIN  # for Sharness and shell tests
+export NODE_PATH="${toplevel}/node_modules${NODE_PATH:+:${NODE_PATH}}"
+
+echo "Updating GitHub GraphQL Flow types"
+cp .prettierrc.json "${SOURCECRED_BIN}/"
```
decentralion: Why is this needed?

scripts/update_snapshots.sh
```diff
@@ -6,16 +6,33 @@ set -eu
...
-echo "Updating for sharness/test_load_example_github.t"
+tmpdir="$(mktemp -d)"
+cleanup() {
+    rm -r "${tmpdir}"
+}
+trap cleanup EXIT
+
+SOURCECRED_BIN="${tmpdir}/bin"
+yarn run --silent backend --output-path "${SOURCECRED_BIN}"
+export SOURCECRED_BIN  # for Sharness and shell tests
```
decentralion: Why not just export SOURCECRED_BIN when declaring it?

scripts/update_snapshots.sh
```diff
@@ -6,16 +6,33 @@ set -eu
...
+cp .prettierrc.json "${SOURCECRED_BIN}/"
+node "${SOURCECRED_BIN}/generateGithubGraphqlFlowTypes.js" \
+    >src/plugins/github/graphqlTypes.js
+
+echo "Updating sharness/test_load_example_github.t"
 (cd sharness; UPDATE_SNAPSHOT=1 ./test_load_example_github.t -l)
+
 echo "Updating git/loadRepositoryTest.sh"
-./src/plugins/git/loadRepositoryTest.sh -u
+./src/plugins/git/loadRepositoryTest.sh -u --no-build
```
decentralion: Good catch.

wchargin: I’d actually assumed that this was intentional—you omit `--no-build` on
the first one so that it builds it into the main repository, then the
rest depend on that. The unpleasant asymmetry is what first prompted me
to write the PR. :-)

scripts/update_snapshots.sh
wchargin: The Flow types are formatted according to the active Prettier config. If
this is omitted, then the output file will omit trailing commas (etc.),
and so it will not actually be the right snapshot.

scripts/update_snapshots.sh
wchargin: The presence of `SOURCECRED_BIN` generally means “this directory
contains built SourceCred binaries”, but that’s not actually true until
the previous line has run. It won’t matter unless `yarn backend` itself
depends on `SOURCECRED_BIN` (which would be very bad).

Separately, you generally want to separate assignments and exports when
the rhs of the assignment may fail, because `export` will mask that
status: `FOO="$(false)"` has exit status 1 because `false` fails, but
`export FOO="$(false)"` has exit status 0 because the `export` itself
succeeds. This doesn’t matter here, because if `${tmpdir}` fails to
expand then the `export` statement will fail. (Shellcheck will tell you
if you’re doing this wrong.)

Would you like me to move this up?

scripts/update_snapshots.sh
decentralion: Got it. It would be good to have a comment explaining in the source code, since it's quite non-obvious.

scripts/update_snapshots.sh
decentralion: Slight preference for moving it up, since IMO it's a little bit confusing, and it seems the considerations that motivate it don't apply in this particular case. but overall I'm fine either way.

scripts/update_snapshots.sh
decentralion: I always had a `yarn backend` call at the top. I think I just forgot to pass --no-build to the first one.

PR #1203, opened 2019-07-06
Switch to flow v0.100.0 to try to fix CI issues

decentralion:
Since upgrading to flow 0.102.0, we've been having CI issues where flow
fails as "out of retries". In my testing, downgrading flow seems to
resolve this, although it's hard to be certain as the issue strikes
sporadically.

Test plan: This commit definitely works locally; hopefully it will
consistently pass in CI as well.


decentralion:
Thanks to @wchargin for help with this PR. Also, cf https://github.com/facebook/flow/issues/7893

PR #1211, opened 2019-07-09
Removed incorrect information from README.md

tylerdmace:
#1167 added some info to the README about how the user needs to have ssh keys setup. This was true at the time, but changed as a result of #1210. This commit fixes that up by removing the now-outdated information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ First, make sure that you have the following dependencies:
 
 [Node]: https://nodejs.org/en/
 [Yarn]: https://yarnpkg.com/lang/en/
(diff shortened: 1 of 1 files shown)


decentralion:
Could you change the pull description to have a bit more context? Like:

```
Remove some outdated info from README

#1167 added some info to the README about how the user needs to have ssh keys setup. This was true at the time, but changed as a result of #1210. This commit fixes that up by removing the now-outdated information.
```

I recognize it doesn't matter hugely in this case, but I like that SourceCred generally has a lot of context in commit messages, and it's been quite useful so far.

PR #1217, opened 2019-07-11
Add ipfs/go-ipfs to website

decentralion:
I've been using it a lot as a test for a larger repo. May as well
canonicalize it.


decentralion:
(I removed sourcecred/pm because we aren't using it very much at all.)

Issue #1047, opened 2019-01-10
Show sourcecred data on a project's readme file


asood123:
It would be cool to generate an SVG that could be easily embedded in a readme file. It can show the top contributors and their percentages. Maybe it shows top 5 or anyone above 5%?

This would be a great way to spread the word about sourcecred.

decentralion:
I like this idea! We'll need to solve #967 first.

wpank:
For reference I think [This Repo](https://github.com/dawnlabs/carbon) is a good inspiration for this sort of feature.

wpank:
I found something that does actually exactly this, with a similar motivation behind sourcecred: showing appreciation for contributors. 

[hall-of-fame](https://github.com/sourcerer-io/hall-of-fame) creates a widget for readme's that shows new, trending and top contributors. 

Taken from their README:

```
Every hour, Hall-of-fame looks into recent commits of your
 repo using GitHub API. It then selects three
 categories of contributors: new (those who made their first 
commit to your repo in the last 7 days), 
trending (those with most commits in the last 7 days), and top 
(those with most commits of all time). 
It then selects up to three new contributors, up to 4 trending 
contributors, and if total is less than 7, 
it fill up the remaining spots with top contributors.

Each contributor in the Hall-of-fame is represented with their 
avatar, and a badge new/trending/top 
with the corresponding number of commits. Each avatar links
 to a contributor's profile. Which means
 that anyone has a chance to be prominently featured on your 
README for some time!

Hall-of-fame works with Sourcerer (https://sourcerer.io), but it's
 not required for your contributors. 
If a contributor has a Sourcerer profile, their Sourcerer avatar 
with a halo is used, and the avatar 
links to their Sourcerer profile. If they don't, an avatar based on 
their GitHub is used, and it is linked 
to their GitHub profile.
```

Example:

### [iterative/dvc](https://github.com/iterative/dvc)

[![](https://sourcerer.io/fame/efiop/iterative/dvc/images/0)](https://sourcerer.io/fame/efiop/iterative/dvc/links/0)
[![](https://sourcerer.io/fame/efiop/iterative/dvc/images/1)](https://sourcerer.io/fame/efiop/iterative/dvc/links/1)
[![](https://sourcerer.io/fame/efiop/iterative/dvc/images/2)](https://sourcerer.io/fame/efiop/iterative/dvc/links/2)
[...]

PR #1223, opened 2019-07-12
add `sourcecred/scores`

decentralion:
The scores are lightly processed from their internal representation.
Example usage:

```
$ yarn backend;
$ node bin/sourcecred.js load sourcecred/sourcecred
$ node bin/sourcecred.js scores sourcecred/sourcecred > scores.json
```

The data structure is as follows:

```js
export type NodeOutput = {|
  +id: string,
  +totalCred: number,
  +intervalCred: $ReadOnlyArray<number>,
|};

export type ScoreOutput = Compatible<{|
  +users: $ReadOnlyArray<NodeOutput>,
  +intervals: $ReadOnlyArray<Interval>,
|}>;
```

Test plan: I added sharness tests at `sharness/test_cli_scores.t`.
In the past, we've used javascript tests for CLI commands. However,
those are pretty time-consuming to write, and are less robust than
simply running the command from bash. Check the snapshot for a sense of
what the new data format looks like. Also, the snapshot updater now
updates this snapshot too.

Relevant for #1047.
Thanks to @Beanow for feedback on the output format and design.


src/cli/help.js
```diff
@@ -7,6 +7,7 @@ import dedent from "../util/dedent";
 import {help as loadHelp} from "./load";
 import {help as analyzeHelp} from "./analyze";
 import {help as pagerankHelp} from "./pagerank";
```
Beanow: `pagerankHelp` is somehow missing below and shows in the linter as unused import.

Beanow:
This looks like a very usable output format. Should we include a format version for upgrade sanity sake?

src/cli/help.js
decentralion: Fixed.

decentralion:
@Beanow I added versioning info per your suggestion. I also added sharness tests with a snapshot, so you can see what the output format looks like.

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
...
+        export SOURCECRED_BIN="${toplevel}/bin"
+    fi &&
+    export NODE_PATH="${toplevel}/node_modules${NODE_PATH:+:${NODE_PATH}}" &&
+    test_set_prereq SETUP
+'
+
+run() (
+  set -eu
+  rm -f out err
+  node "${SOURCECRED_BIN}"/sourcecred.js "$@" >out 2>err
```
decentralion: @wchargin would you be willing to take a look at this file and tell me what you think?

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
...
+export GIT_CONFIG_NOSYSTEM=1
+export GIT_ATTR_NOSYSTEM=1
+
+# shellcheck disable=SC1091
+. ./sharness.sh
+
+test_expect_success "environment and Node linking setup" '
+    toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)" &&
+    snapshot_directory="${toplevel}/sharness/__snapshots__/" &&
+    SOURCECRED_DIRECTORY="${snapshot_directory}/example-github-load" &&
```
wchargin: You haven’t exported `SOURCECRED_DIRECTORY`, so it’s not visible to the
child processes. The children work against your actually-populated cache
locally, but fail with an empty cache on CI (or on my machine,
presumably, though I didn’t run this).

Immediate fix:

```sh
    SOURCECRED_DIRECTORY="${snapshot_directory}/example-github-load" &&
    export SOURCECRED_DIRECTORY &&
```

Long-term suggestion: In the top-level test config, make a temp
directory and export `` `${tmpdir}/ERROR_NO_SOURCECRED_DIRECTORY_SET` ``
to make tests more hermetic.

Also:

  - In your test, use `diff -u` rather than `diff -q` so that the output
    is printed on failure.

  - In your `run`, use something like

    ```sh
    run() (
      set -eu
      rm -f out err
      code=0
      node "${SOURCECRED_BIN}"/sourcecred.js "$@" >out 2>err || code=$?
      if [ "${code}" -ne 0 ]; then
        printf '%s failed with %d\n' "sourcecred $*"
        printf 'stdout:\n'
        cat out
        printf 'stderr:\n'
        cat err
      fi
      return "${code}"
    )
    ```

    so that you can see the output when it fails.

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
...
+    test_set_prereq SETUP
+'
+
+run() (
+  set -eu
+  rm -f out err
+  node "${SOURCECRED_BIN}"/sourcecred.js "$@" >out 2>err
+)
+
+test_expect_success SETUP, \
```
wchargin: nit: `s/SETUP,/SETUP/`

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
...
+  "should print help message when called without scores" '
+  test_must_fail run scores &&
```
wchargin: nit: Please fix indentation throughout this file. It’s 4-space indent,
but more importantly the elements of an `&&`-chain should all be at the
same level of indentation.

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
+#!/bin/sh
+
+# Disable these lint rules globally:
+#   2034 = unused variable (used by sharness)
+#   2016 = parameter expansion in single quotes
+#   1004 = backslash-newline in single quotes
+# shellcheck disable=SC2034,SC2016,SC1004
+:
+
+test_description='tests for scripts/build_static_site.sh'
```
wchargin: doubt it

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
...
+'
+
+if [ -n "${UPDATE_SNAPSHOT}" ]; then
+    test_set_prereq UPDATE_SNAPSHOT
+fi
+
+test_expect_success SETUP,UPDATE_SNAPSHOT \
+    "should update the snapshot" '
+    run scores sourcecred/example-github &&
+    mv out ${snapshot_file}
```
wchargin: Quotes, please! `"${snapshot_file}"`

sharness/test_cli_scores.t
```diff
@@ -0,0 +1,86 @@
...
+'
+
+test_expect_success SETUP \
+    "should be identical to the snapshot" '
+    run scores sourcecred/example-github &&
+    diff -q out ${snapshot_file}
```
wchargin: Quotes!

sharness/test_cli_scores.t
wchargin: Done.

sharness/test_cli_scores.t
Beanow: Related to #1226

sharness/test_cli_scores.t
decentralion: Done.

sharness/test_cli_scores.t
decentralion: Fixed

sharness/test_cli_scores.t
decentralion: Fixed

sharness/test_cli_scores.t
decentralion: Done.

sharness/test_cli_scores.t
decentralion: Thanks for the suggestions! I adopted the all of them except the longer-term suggestion, which I didn't understand. Can you please elaborate on it? Are there any examples of a similar pattern already in the codebase?

Also, why should we separate defining SOURCECRED_DIRECTORY and exporting it into two separate statements? Couldn't we just as easily combine them into one?

wchargin: > I adopted the all of them except the longer-term suggestion, which I
> didn't understand. Can you please elaborate on it?

Sure.

When you run `config/test.js`, we invoke (e.g.) `yarn run -s sharness`
in a subprocess. That subprocess inherits your working environment,
including your `SOURCECRED_DIRECTORY` variable. If the Sharness task
doesn’t explicitly set the `SOURCECRED_DIRECTORY` variable itself, then
it will use your global cache, which is not hermetic: it may pass for
you but fail for everyone else, as happened here.

The simplest form of what I’m suggesting is to just add

```js
process.env["SOURCECRED_DIRECTORY"] = "/ERROR_SOURCECRED_DIRECTORY_NOT_SET";
```

to the top of `config/test.js`. Then, if the Sharness task fails to
explicitly set the `SOURCECRED_DIRECTORY` variable, it will inherit the
“dummy” value; any attempt to read from or write to this path, which is
a non-existent directory in a write-protected directory, will be much
more likely to fail.

In the case of the bug in the original version of this PR, I think that
the test would have just failed with

```
fatal: repository ID sourcecred/example-github not loaded
Try running `sourcecred load sourcecred/example-github` first.
```

when run locally, which is probably enough to point you in the right
direction. If the test had tried to run `sourcecred load …`, then the error message would have been

```
Contents of stderr:
    EACCES: permission denied, mkdir '/ERROR_SOURCECRED_DIRECTORY_NOT_SET'
```

which explicitly points to the problem in the test configuration.

(I say “simplest” because ideally we’d (a) change `yarn sharness`
similarly, to prevent skew between `yarn sharness` and `yarn test`, and
(b) use a directory name that is guaranteed to not exist by running
something like `tmpdir="$(mktemp -d)" && chmod 000 "${tmpdir}"` before
exporting `"${tmpdir}/ERROR_SOURCECRED_DIRECTORY_NOT_SET"`.)


> Are there any examples of a similar pattern already in the codebase?

Not exactly, no. But the general pattern is occasionally quite useful.

> Also, why should we separate defining SOURCECRED_DIRECTORY and
> exporting it into two separate statements? Couldn't we just as easily
> combine them into one?

Using `export NAME=word` is in general dangerous when `word` is not a
constant, because `export` returns success if the export itself succeeds
even if the expansion of `word` failed. For instance:

```shell
$ (set -e; FOO="$(false)"; printf 'got here\n'; export FOO); echo $?
1
$ (set -e; export FOO="$(false)"; printf 'got here\n'); echo $?
got here
0
```

It’s actually safe in this case, because the only non-constant portion
of the variable assignment is a parameter expansion, and failures of
parameter expansion (either unbound variables with `set -u`, or syntax
errors in the expansion form) _will_ cause the whole `export` to fail.
But this isn’t something that I really want to have to remember, hence
the blanket advice to separate `export`s from assignments whenever the
RHS of the assignment is non-constant.

That is: if you made this particular `export` also assign, I wouldn’t
object; the above habit is simply the reason that I wrote my suggestion
the way that I did.

There is a shellcheck lint for this:
<https://github.com/koalaman/shellcheck/wiki/SC2155>

(and the shellcheck lint _will_ only warn in the actually-dangerous
case, which is nice). I highly recommend integrating shellcheck-on-save
into your editor. It’ll catch all of those missing quotes!

decentralion: Thanks!

PR #1224, opened 2019-07-12
Add --weights as load option

Beanow:
This will let you use the same format `weights.json` file as the frontend allows you to upload/download and use it doing the load command.

### DIFF
--- a/src/cli/load.js
+++ b/src/cli/load.js
@@ -16,7 +16,7 @@ import {Graph} from "../core/graph";
 import {loadGraph, type LoadGraphResult} from "../analysis/loadGraph";
 import {type IBackendAdapterLoader} from "../analysis/analysisAdapter";
 import {TimelineCred} from "../analysis/timeline/timelineCred";
-import {defaultWeights} from "../analysis/weights";
+import {defaultWeights, fromJSON as weightsFromJSON} from "../analysis/weights";
 
 import execDependencyGraph from "../tools/execDependencyGraph";
 import {loadGithubData} from "../plugins/github/loadGithubData";
@@ -55,6 +55,10 @@ function usage(print: (string) => void): void {
             'github'. If not specified, data for all default plugins will be
             loaded.
 
+        --weights WEIGHTS_FILE
+            Path to a json file which contains a weights configuration.
+            This will be used instead of the default weights and persisted.
+
         --help
             Show this help message and exit, as 'sourcecred help load'.
 
@@ -88,6 +92,7 @@ function die(std, message) {
 export type LoadOptions = {|
   +output: RepoId,
   +repoIds: $ReadOnlyArray<RepoId>,
+  +weightsPath?: string | null,
 |};
 
 export function makeLoadCommand(
@@ -107,6 +112,7 @@ export function makeLoadCommand(
     const repoIds: RepoId[] = [];
     let explicitOutput: RepoId | null = null;
     let plugin: Common.PluginName | null = null;
+    let weightsPath: string | null = null;
 
     for (let i = 0; i < args.length; i++) {
       switch (args[i]) {
@@ -120,6 +126,14 @@ export function makeLoadCommand(
           if (++i >= args.length)
             return die(std, "'--output' given without value");
           explicitOutput = stringToRepoId(args[i]);
+          break;
+        }
+        case "--weights": {
+          if (weightsPath != null)
+            return die(std, "'--weights' given multiple times");
+          if (++i >= args.length)
+            return die(std, "'--weights' given without value");
+          weightsPath = args[i];
           break;
         }
         case "--plugin": {
@@ -150,7 +164,7 @@ export function makeLoadCommand(
       return die(std, "output repository not specified");
     }
 
-    const options: LoadOptions = {output, repoIds: repoIds};
+    const options: LoadOptions = {output, repoIds: repoIds, weightsPath};
 
     if (plugin == null) {
       try {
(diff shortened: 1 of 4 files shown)


Beanow:
Edit: realized we use `scripts/build_static_site.sh` so I've added the option there as well.

src/cli/load.js
```diff
@@ -177,7 +191,7 @@ export type SaveGraph = (
   RepoId
 ) => Promise<Graph>;
 
-export type SaveCred = (Graph, RepoId) => Promise<void>;
+export type SaveCred = (Graph, RepoId, string | null) => Promise<void>;
```
decentralion: To make this a little clearer to the reader, how about we make it:

```js
export type SaveCred = (graph: Graph, repoId: RepoId, weightsPath?: string) => Promise<void>
```

this way someone reading the type signature can see what the string argument means. (Another option would be to make a type alias `export type WeightsPath = string)

Using `weightsPath?` means that it's an optional arg, so we won't need to do the `| null` stuff.

src/cli/load.js
```diff
@@ -88,6 +92,7 @@ function die(std, message) {
 export type LoadOptions = {|
   +output: RepoId,
   +repoIds: $ReadOnlyArray<RepoId>,
+  +weightsPath?: string | null,
```
decentralion: if using `weightsPath?` then you don't need to do `string | null`, since the `?` already makes it optional.

src/cli/load.js
```diff
@@ -220,22 +234,42 @@ export const makeLoadDefaultPlugins = (
     }
     addToRepoIdRegistry(options.output);
     const graph = await saveGraph(defaultAdapterLoaders(), options.output);
-    await saveCred(graph, options.output);
+    await saveCred(graph, options.output, options.weightsPath || null);
```
decentralion: if we make it an optional arg (see above) then we won't need the `|| null`

src/cli/load.js
```diff
@@ -220,22 +234,42 @@ export const makeLoadDefaultPlugins = (
...
+  }
+
+  const raw = await fs.readFile(path, "utf-8");
+  return weightsFromJSON(JSON.parse(raw));
+};
+
+export const saveCred = async (
+  graph: Graph,
+  repoId: RepoId,
+  weightsPath: string | null
```
decentralion: as above, can be `weightsPath?: string`

src/cli/load.js
Beanow: Changed it.

src/cli/load.js
Beanow: Changed it.

src/cli/load.js
Beanow: Changed it.

src/cli/load.js
Beanow: Changed it.

src/cli/load.js
```diff
@@ -122,6 +128,14 @@ export function makeLoadCommand(
           explicitOutput = stringToRepoId(args[i]);
           break;
         }
+        case "--weights": {
+          if (weightsPath != undefined)
```
Beanow: In general an optional field `weightsPath?: string` doesn't match nulls so I used undefined here.

src/cli/load.test.js
```diff
@@ -544,7 +553,27 @@ describe("cli/load", () => {
         repoIds: [fooBar, fooBaz],
       });
       expect(saveCred).toHaveBeenCalledTimes(1);
-      expect(saveCred).toHaveBeenCalledWith(graph, fooCombined);
+      expect(saveCred).toHaveBeenCalledWith(graph, fooCombined, undefined);
+    });
+
+    it("calls saveCred with the weight path if set in the options", async () => {
```
Beanow: Along the lines of the previous test case, I've included the case where a path _is_ set.

Beanow:
The tests passed for me locally, I'm not sure what's up with the CI one.

Beanow:
Rebased from master, which fixed the CI tests.

decentralion:
CI is having issues. You can ignore ENOMEM issues or `out of retries, exiting` issues from flow. It started happening after I upgraded all the deps, I still need to look into it.

src/cli/load.test.js
```diff
@@ -302,7 +308,10 @@ describe("cli/load", () => {
           successCase({
             name: "with a multiple repositories",
             args: ["foo/bar", "bar/zod", "--output", "bar/zod"],
-            loadOptions: {output: barZod, repoIds: [fooBar, barZod]},
+            loadOptions: {
```
decentralion: I'm confused where all these unrelated changes are coming from (prettier doesn't usually churn the codebase).

Beanow: It failed on the prettier check on `yarn test`, this is the result of `yarn prettify`.

decentralion: interesting; on my machine, both before and after were `pretty` enough. we must have slightly different config.

decentralion:
@Beanow I'd like this to get merged as squash+merge (rather than rebase+merge). Would you write up a commit message for the combined commit? Something that just has some context on the change and how it was tested.

Beanow:
Add --weights as load option

Includes a change to cli load and build_static_site.sh to accept a --weights WEIGHTS_FILE argument.
This allows overriding the default weights at build-time using a weights.json that has the same format as previously generated in the frontend.

Adds an additional test-case as well for propagating the optional parameter.
The file I/O of loading and parsing a weights.json file was tested manually. As analysis/weights' fromJSON() is tested elsewhere as is passing weight parameters.

PR #1225, opened 2019-07-12
Blacklist additions quasar test

Beanow:
Setting up a demo for quasar, another user that became an org appeared and caused issues loading the graph. These blacklisted IDs got it to work for me.


decentralion:
I think this is a bit too trivial for a CHANGELOG entry; can you remove that commit? Otherwise looks good.

Beanow:
Force pushed to only include the first commit. Changelog should now be unchanged.

PR #1228, opened 2019-07-14
Fix #1226

### LINKED ISSUE #1226: build_static_site.sh now requires SOURCECRED_DIRECTORY env var
The change in https://github.com/sourcecred/sourcecred/commit/ff026817d014124914af51028c72254052c196dd

Uses the `SOURCECRED_DIRECTORY` env var before it's exported with a default value.
Meaning it's now required to be set or will crash the script. It should probably be wrapped in an if statement to skip linking the cache if the env is not provided.

decentralion:
This resolves an issue where build_static_site.sh fails if the
$SOURCECRED_DIRECTORY is not set.

I tried to add a test for this, but was unsuccessful; see discussion in
[a575638].

Test plan: I manually verified that running build_static_site.sh with an
unset SOURCECRED_DIRECTORY fails before this commit, and passes
afterwards.

[a575638]: https://github.com/sourcecred/sourcecred/commit/a575638fa47265cb624663894c4e28ba2bddc2e4

### DIFF
--- a/scripts/build_static_site.sh
+++ b/scripts/build_static_site.sh
@@ -108,9 +108,16 @@ parse_args() {
 
 build() {
     sourcecred_data="$(mktemp -d --suffix ".sourcecred-data")"
-    # Give sourcecred access to the full cache. This will greatly speed
-    # up site builds on repos that have already been loaded.
-    ln -s "${SOURCECRED_DIRECTORY}/cache" "${sourcecred_data}/cache"
+
+    if [ -n "${SOURCECRED_DIRECTORY:-}" ]; then
+        # If $SOURCECRED_DIRECTORY is available, then give sourcecred access to
+        # the cache. This will greatly speed up site builds on repos that have
+        # already been loaded.
+        # Note this speedup will only apply if the SOURCECRED_DIRECTORY has been
+        # explicitly set.
+        ln -s "${SOURCECRED_DIRECTORY}/cache" "${sourcecred_data}/cache"
+    fi
+
     export SOURCECRED_DIRECTORY="${sourcecred_data}"
 
     if [ "${BACKEND}" -ne 0 ]; then


wchargin:
Are you on an ancient version of bash bundled with macOS? I can’t
reproduce this with GNU bash 4.3.48.

```shell
$ cat test.sh
#!/bin/bash
set -eu

main() {
  trap cleanup EXIT
  case "${1-}" in
      with)
          failure_with_exit_code
          ;;
      without)
          failure_without_exit_code
          ;;
      *)
          printf >&2 'bad argument "%s"; use "with" or "without"\n' "${1-}"
          exit 1
          ;;
  esac
}

cleanup() {
  rv=$?
  printf 'CLEANUP: rv=%d\n' "${rv}"
  exit "${rv}"
}

failure_without_exit_code() {
    echo "${UNSET_VARIABLE}"
}

failure_with_exit_code() {
  false
}

main "$@"
```

```
$ ./test.sh with; echo $?
CLEANUP: rv=1
1
```

```
$ ./test.sh without; echo $?
./test.sh: line 27: UNSET_VARIABLE: unbound variable
CLEANUP: rv=1
1
```

```
$ bash --version | head -n 1
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
$ sh ./test.sh with; echo $?
CLEANUP: rv=1
1
```

```
$ sh ./test.sh without; echo $?
./test.sh: 27: ./test.sh: UNSET_VARIABLE: parameter not set
CLEANUP: rv=2
2
```

(`sh` gives a _different_ failure code for unset variables, but that’s
fine per spec.)

Aside: To generate an error, please use `false` or `return 37` or
something rather than `rm`, even if the file is “unlikely” to exist.

decentralion:
Hmm, you're right: due to licensing issues, MacOS ships with a version of bash that was current in 2007!

However, finding a fix is not as simple as `brew install bash` -- that will install a later version of bash in `/usr/local/bin/bash`. However, our shell scripts invoke `bin/bash` directly. It seems that overriding `bin/bash` is not as simple as `sudo`ing, I would need to restart in recovery mode and disable system integrity protection.

Do you see any issue with moving from `#!/bin/bash` to `#!/usr/bin/env bash` to work around this issue?

wchargin:
(Just saw this comment.)

@csilvers always advised me to avoid `#!/usr/bin/env`, primarily (iirc)
as a security/reproducibility issue. I’m inclined to trust his judgment,
so I’ve avoided it. He also advised me to prefer POSIX `sh` when
possible, rather than `bash` or GNU `bash`, which largely ameliorates
the issue. I’ve been quite happy with the results: if I need more
functionality than POSIX `sh` offers (e.g., arrays other than `"$@"`),
then I’d usually be better off with a more general-purpose programming
language, anyway.

Note that [Git uses the same pattern][1] to capture the failed exit
code, and they’re quite diligent about [working on all reasonable
platforms in practice][2]. They also run tests under `sh`, not `bash`.
And Sharness was written for Git (or, rather, extracted from Git); their
lead is more relevant here than just a generic opinion of best practice.

[1]: https://github.com/git/git/blob/ff66981f4593aec0f3b3eeace0eacb7dbe44fd8c/t/test-lib.sh#L635-L651
[2]: https://github.com/git/git/blob/ff66981f4593aec0f3b3eeace0eacb7dbe44fd8c/Documentation/CodingGuidelines#L4-L15

Issue #701, opened 2018-08-23
Project-level cred attribution


decentralion:
We should enable SourceCred instances for entire projects, rather than just individual repositories. This will be a good fit for projects like @ipfs or @ropensci which contain many different repositories. 

We basically need to fetch a listing of all the repositories owned by an org, fetch all of those repos individually, and then merge all the graphs together into an org-level graph. 

Factored out of #697.

Some TODOs:
- [ ] Change data model to explicitly support "projects" (lists of repos?)
- [ ] Automatically create a project with all the repos in an organization
- [ ] Allow filtering down to individual repo (or group of repos?)
- [ ] Aggregate across the repository first (if multiple repos available?)
- [ ] Solve data sharding problems (#731)

PR #1238, opened 2019-07-18
Add `core/project` and `core/project_io`

decentralion:
This creates a new `Project` type which will replace `RepoId` as the
index type for saving and loading data.

The basic data type is added to `project.js`. Rather than having a
`RepoIdRegistry`, I intend to infer the registry at build time by
scanning for available projects saved in the sourcecred directory. I've
added the `project_io` module for this task. It has methods for setting
up a project subdirectory, and loading the `Project` info from that
subdirectory.

To ensure that projects ids can be encoded even if they have symbols
like `/` and `@`, we base64 encode them.

To ensure that project ids can be retrieved at build time, the
`getProjectIds` method is factored out into its own plain ECMAScript
module. For all non-build time needs, it is re-exported from
`project_io`.

Test plan: Unit tests added; run `yarn test`.

Note: This is factored out of #1233 and is progress towards #701.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "aphrodite": "^2.1.0",
+    "base-64": "^0.1.0",
     "better-sqlite3": "^5.4.0",
     "chalk": "2.4.2",
     "commonmark": "^0.29.0",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "aphrodite": "^2.1.0",
-    "base-64": "^0.1.0",
+    "base64url": "^3.0.1",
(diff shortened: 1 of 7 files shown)


decentralion:
Based on feedback from @beanow, I modified the code to use base64-url encoding, and to make _getProjectIds async.

Beanow:
Particularly windows would need testing here as it's very quirky with casing. And base64 is case sensitive. NTFS does preserve caps internally but in use its optional within windows afaik.

decentralion:
I don't know of any prospective windows users, and don't have a testing environment for windows, so I'm not going to put effort into supporting windows yet. For now, people who want to deploy SourceCred instances and have windows machines can use aws, gcp, or CI.

Issue #1243, opened 2019-07-19
Persistent CI issues


decentralion:
Over the past few weeks, we've had persistent CI issues. They started when I did a round of dependency updates; see #1195, #1196, and #1197. Since then, we've had sporadic CI flakes. Interestingly, the flake can take many different forms, and break various tests. I've seen:

## lint:

```
 FAIL  lint
Exit code: 137
Contents of stderr:
    Killed
    error Command failed with exit code 137.
```

## flow:
```
 FAIL  flow
Exit code: 7
Contents of stderr:
    Out of retries, exiting!
    error Command failed with exit code 7.
```

## unit:
```
 FAIL  unit
Exit code: 1
Contents of stderr:
...
    FAIL src/plugins/git/example/exampleRepo.test.js (6.443s)
      ● plugins/git/example/exampleRepo › createExampleRepo › is deterministic
    
        spawnSync git ENOMEM
```

On rare occasions, I've seen similar failures from check-pretty or sharness.

The common theme here seems to be memory exhaustion (processes get killed by CircleCI if it runs out of memory, which results in exit code 137). So my best theory is that one of the updates resulted in lint, flow, or jest requiring a lot more memory, such that now (depending on execution order) we exhaust memory.

In local testing, I found that jest was the biggest memory hog, so I merged #1230, which made the `unit` task run before anything else in CI. In CI testing, that seemed to help, but I'm still seeing flakey failures since that merge.

Next, I think I will try isolating not just unit, but also flow and lint, before running other tests.

decentralion:
Apparently it's possible for the flow task to fail (out of retries) when it is the first task, and nothing else at all is happening on the server: https://circleci.com/gh/sourcecred/sourcecred/1403

decentralion:
This seems to have been fixed by #1282.

PR #1244, opened 2019-07-19
Another stab at fixing CI flakiness

decentralion:
See #1243 for context. This is basically a more aggressive version of
pull #1230 -- instead of just running unit tests in isolation, we also
run flow in isolation, and kill the servers afterwards.

Test plan: See how this fares in CI :)

### DIFF
--- a/config/test.js
+++ b/config/test.js
@@ -1,5 +1,7 @@
 // @flow
 
+/*:: import type {Task} from "../src/tools/execDependencyGraph"; */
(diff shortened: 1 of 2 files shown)


decentralion:
It passed CI, which doesn't persuade me that it fixed the issue (I tested #1230 extensively in CI and still saw flakey failures). But it's a start. I'm going to merge this and see what happens.

PR #1249, opened 2019-07-21
Replace `Object.freeze` with `deepFreeze`

decentralion:
Throughout the codebase, we freeze objects when we want to ensure that
their properties are never altered -- e.g. because they are a plugin
declaration, or are being re-used for various test cases.

We generally use `Object.freeze`. This has the disadvantage that it does
not work recursively, so a frozen object's mutable fields and properties
can still be mutated. (E.g. if `const obj = Object.freeze({foo: []})`,
then `obj.foo.push(1)` will succeed in mutating the 'frozen' object).

Sometimes we anticipate this and explicitly freeze the sub-fields (which
is tedious); sometimes we forget (which invites errors). This change
simply replaces all instances of Object.freeze with [deep-freeze], so we
don't need to worry about the issue at all anymore.

Test plan: `yarn test` passes (after updating snapshots);
`git grep Object.freeze` returns no hits.

[deep-freeze]: https://www.npmjs.com/package/deep-freeze


wchargin:
FWIW I’m not sure that `generateFlowTypes` should emit modules that have
external dependencies, given that the `src/graphql/` package was always
intended to be well isolated and general-purpose, and especially given
that it has zero benefit: these objects will always be shallow, and they
are machine-generated, and both types and tests ensure this.

decentralion:
I'm fine with reverting the change locally.

PR #1260, opened 2019-08-01
add timelineCredView

zixuanzh:
Add TimelineCredView that shows cred for a specified interval which will enable zoom in/zoom out feature in timelineCred later on.

TODO
- [x] More testing on TimelineCredView

### DIFF
--- a/src/analysis/timeline/timelineCred.js
+++ b/src/analysis/timeline/timelineCred.js
@@ -1,8 +1,5 @@
 // @flow
 
-import {sum} from "d3-array";
-import sortBy from "lodash.sortby";
-import * as NullUtil from "../../util/null";
 import {toCompat, fromCompat, type Compatible} from "../../util/compat";
 import {type Interval} from "./interval";
 import {timelinePagerank} from "./timelinePagerank";
@@ -11,7 +8,6 @@ import {
   Graph,
   type GraphJSON,
   type NodeAddressT,
-  NodeAddress,
   type Node,
 } from "../../core/graph";
 import {
@@ -30,18 +26,6 @@ import {
 } from "./filterTimelineCred";
 
 export type {Interval} from "./interval";
-
-/**
- * A Graph Node wrapped with cred information.
- */
-export type CredNode = {|
-  // The Graph Node in question
-  +node: Node,
-  // The total aggregated cred. (Summed across every interval).
-  +total: number,
-  // The timeline sequence of cred (one score per interval).
-  +cred: $ReadOnlyArray<number>,
-|};
 
 /**
  * Parameters for computing TimelineCred
@@ -128,6 +112,10 @@ export class TimelineCred {
     return this._config;
   }
 
+  availableNodes(): NodeAddressT[] {
+    return Array.from(this._cred.addressToCred.keys());
+  }
+
   /**
    * Creates a new TimelineCred based on the new Parameters.
    * Holds the graph and config constant.
@@ -143,35 +131,6 @@ export class TimelineCred {
    */
   intervals(): $ReadOnlyArray<Interval> {
     return this._cred.intervals;
-  }
-
-  /**
-   * Get the CredNode for a given NodeAddress.
-   *
-   * Returns undefined if the node is not in the filtered results.
-   *
-   * Note that it's possible that the node is present in the Graph, but not the
-   * filtered results; if so, it will return undefined.
-   */
-  credNode(a: NodeAddressT): ?CredNode {
-    const cred = this._cred.addressToCred.get(a);
-    if (cred === undefined) {
-      return undefined;
-    }
-    const total = sum(cred);
-    const node = NullUtil.get(this._graph.node(a));
-    return {cred, total, node};
-  }
-
-  /**
-   * Return all the nodes matching the prefix, along with their cred,
-   * sorted by total cred (descending).
-   */
-  credSortedNodes(prefix: NodeAddressT): $ReadOnlyArray<CredNode> {
-    const match = (a) => NodeAddress.hasPrefix(a, prefix);
-    const addresses = Array.from(this._cred.addressToCred.keys()).filter(match);
-    const credNodes = addresses.map((a) => this.credNode(a));
-    return sortBy(credNodes, (x: CredNode) => -x.total);
   }
 
   toJSON(): TimelineCredJSON {
@@ -202,6 +161,22 @@ export class TimelineCred {
   ): Promise<TimelineCred> {
     const ftc = await _computeTimelineCred(graph, params, config);
     return new TimelineCred(graph, ftc, params, config);
+  }
+
+  /**
+   * Returns the Node matching a given NodeAddressT, if such a node exists,
+   * or undefined otherwise.
+   */
+  node(a: NodeAddressT): ?Node {
+    return this.graph().node(a);
+  }
+
+  /**
+   * Returns an array of cred from a given NodeAddressT from FilteredTimelineCred,
+   * if such a node exists, or undefined otherwise.
+   */
+  cred(a: NodeAddressT): ?$ReadOnlyArray<number> {
+    return this._cred.addressToCred.get(a);
   }
 }
 
--- a/src/analysis/timeline/timelineCred.test.js
+++ b/src/analysis/timeline/timelineCred.test.js
@@ -1,7 +1,5 @@
 // @flow
 
-import {sum} from "d3-array";
-import sortBy from "lodash.sortby";
 import {utcWeek} from "d3-time";
 import {NodeAddress, Graph} from "../../core/graph";
 import {TimelineCred, type TimelineCredConfig} from "./timelineCred";
@@ -15,6 +13,13 @@ describe("src/analysis/timeline/timelineCred", () => {
     types: {nodeTypes: [], edgeTypes: []},
   });
 
+  const users = [
+    ["starter", (x) => Math.max(0, 20 - x)],
+    ["steady", (_) => 4],
+    ["finisher", (x) => (x * x) / 20],
+    ["latecomer", (x) => Math.max(0, x - 20)],
+  ];
+
   function exampleTimelineCred(): TimelineCred {
     const startTimeMs = +new Date(2017, 0);
     const endTimeMs = +new Date(2017, 6);
@@ -26,12 +31,6 @@ describe("src/analysis/timeline/timelineCred", () => {
         endTimeMs: +boundaries[i + 1],
       });
     }
-    const users = [
-      ["starter", (x) => Math.max(0, 20 - x)],
-      ["steady", (_) => 4],
-      ["finisher", (x) => (x * x) / 20],
-      ["latecomer", (x) => Math.max(0, x - 20)],
-    ];
 
     const graph = new Graph();
     const addressToCred = new Map();
@@ -42,7 +41,7 @@ describe("src/analysis/timeline/timelineCred", () => {
         description: `[@${name}](https://github.com/${name})`,
         timestampMs: null,
       });
-      const scores = intervals.map((_unuesd, i) => generator(i));
+      const scores = intervals.map((_unused_interval, i) => generator(i));
       addressToCred.set(address, scores);
     }
     const filteredTimelineCred: FilteredTimelineCred = {
@@ -60,28 +59,37 @@ describe("src/analysis/timeline/timelineCred", () => {
     expect(tc.graph()).toEqual(tc_.graph());
     expect(tc.params()).toEqual(tc_.params());
     expect(tc.config()).toEqual(tc_.config());
-    expect(tc.credSortedNodes(NodeAddress.empty)).toEqual(
-      tc.credSortedNodes(NodeAddress.empty)
-    );
   });
 
-  it("cred sorting works", () => {
+  it("`node` returns undefined for absent nodes", () => {
     const tc = exampleTimelineCred();
-    const sorted = tc.credSortedNodes(NodeAddress.empty);
-    const expected = sortBy(sorted, (x) => -x.total);
-    expect(sorted).toEqual(expected);
+    expect(tc.node(NodeAddress.fromParts(["baz"]))).toBe(undefined);
   });
 
-  it("cred aggregation works", () => {
+  it("`node` returns node for node address", () => {
     const tc = exampleTimelineCred();
-    const nodes = tc.credSortedNodes(NodeAddress.empty);
-    for (const node of nodes) {
-      expect(node.total).toEqual(sum(node.cred));
+    for (const [name, _unused_generator] of users) {
+      const address = NodeAddress.fromParts(["foo", name]);
+      const expectedNode = {
+        address,
+        description: `[@${name}](https://github.com/${name})`,
+        timestampMs: null,
+      };
+      expect(tc.node(address)).toEqual(expectedNode);
     }
   });
 
-  it("credNode returns undefined for absent nodes", () => {
+  it("`cred` returns undefined for absent nodes", () => {
     const tc = exampleTimelineCred();
-    expect(tc.credNode(NodeAddress.fromParts(["baz"]))).toBe(undefined);
+    expect(tc.cred(NodeAddress.fromParts(["baz"]))).toBe(undefined);
+  });
+
+  it("`cred` returns cred for node address", () => {
+    const tc = exampleTimelineCred();
+    for (const [name, _unused_generator] of users) {
+      const address = NodeAddress.fromParts(["foo", name]);
+      const expectedCred = tc._cred.addressToCred.get(address);
+      expect(tc.cred(address)).toEqual(expectedCred);
+    }
   });
 });
--- a/src/analysis/timeline/timelineCredView.js
+++ b/src/analysis/timeline/timelineCredView.js
@@ -1,0 +1,134 @@
+// @flow
+
+import {sum} from "d3-array";
+import sortBy from "lodash.sortby";
+import stringify from "json-stable-stringify";
+import * as NullUtil from "../../util/null";
+import {TimelineCred} from "./timelineCred";
+import {type Interval, weekIntervals} from "./interval";
+import {type Node, type NodeAddressT, NodeAddress} from "../../core/graph";
+
+/**
+ * A Graph Node wrapped with cred information.
+ */
+export type CredNode = {|
+  // The Graph Node in question
+  +node: Node,
+  // The total aggregated cred. (Summed across every interval).
+  +total: number,
+  // The timeline sequence of cred (one score per interval).
+  +cred: $ReadOnlyArray<number>,
+|};
+
+/**
+ * TimelineCredView allows a retrieving information about nodes' cred
+ * for a specific interval of time. The TimelineCredView wraps a TimelineCred
+ * instance, and provides methods like `credNode` and `credSortedNodes` for
+ * retrieving information on the available nodes and how much cred they earned.
+ * Every TimelineCredView is scoped to a certain view interval, which defines
+ * the min and max time of cred being viewed. When `CredNode` results are returned,
+ * the `total` field will correspond to the view interval of the TimelineCredView.
+ * By default, the view interval is chosen to include every underlying interval
+ * in the TimelineCred`.
+ */
+export class TimelineCredView {
+  _inputInterval: Interval;
+  _viewIntervals: Interval[];
+  _timelineCred: TimelineCred;
+
+  // We are displaying a range of the intervals in the base TimelineCred
+  // We track the start index and end index of the range, so we can efficiently
+  // slice the cred
+  _startIntervalIndex: number;
+  _endIntervalIndex: number;
+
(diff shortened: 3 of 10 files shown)


src/analysis/timeline/timelineCred.test.js
```diff
@@ -57,31 +58,36 @@ describe("src/analysis/timeline/timelineCred", () => {
...
     expect(tc.params()).toEqual(tc_.params());
     expect(tc.config()).toEqual(tc_.config());
-    expect(tc.credSortedNodes(NodeAddress.empty)).toEqual(
-      tc.credSortedNodes(NodeAddress.empty)
+    expect(tcv.credSortedNodesView(NodeAddress.empty)).toEqual(
+      tcv_.credSortedNodesView(NodeAddress.empty)
     );
   });
 
   it("cred sorting works", () => {
```
decentralion: These tests no longer apply to the `timelineCred` class, and should be moved to `timelineCredView.test.js`

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+import sortBy from "lodash.sortby";
+import stringify from "json-stable-stringify";
+import * as NullUtil from "../../util/null";
+import {TimelineCred, type CredNode} from "./timelineCred";
+import {type Interval, weekIntervals} from "./interval";
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
+
+// test intervals in timelineCredView subsset of the
+// intervals from the base TimelineCred
+export class TimelineCredView {
```
decentralion: Needs a class docstring that explains what this class is for. As an example, consider the docstrings in the `timelineCred.js` module.

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+      if (!baseIntervalStrings.has(v)) {
+        throw new Error(`invariant violation: can't find interval ${v}`);
+      }
+    }
+    this._startIntervalIndex = Array.from(baseIntervalStrings).indexOf(
+      viewIntervalStrings[0]
+    );
+    this._endIntervalIndex = Array.from(baseIntervalStrings).indexOf(
+      viewIntervalStrings[viewIntervalStrings.length - 1]
+    );
```
decentralion: Having an invariant checker that verifies that neither index is `-1` would make me feel better.

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+    this._timelineCred = timelineCred;
+    this._inputInterval = inputInterval;
+    this._viewIntervals = weekIntervals(
+      inputInterval.startTimeMs,
+      inputInterval.endTimeMs
+    );
+    this._findIntervalIndices();
+  }
+
+  _findIntervalIndices() {
```
decentralion: Given that this method will only be called once, and in the constructor, I think we may want to "unwrap it" directly into the constructor... but also add some comments explaining what the logic is for.

Consider this a 'nit', not needed to merge.

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
+// @flow
+
+import {sum} from "d3-array";
+import sortBy from "lodash.sortby";
+import stringify from "json-stable-stringify";
+import * as NullUtil from "../../util/null";
+import {TimelineCred, type CredNode} from "./timelineCred";
```
decentralion: We should move the `CredNode` type into this module, since now the `timelineCred.js` module does not create or return any `CredNode`s.

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+    const total = sum(cred);
+    const node = NullUtil.get(this._timelineCred._graph.node(a));
+    return {cred, total, node};
+  }
+
+  /**
+   * Return all the nodes matching the prefix, along with their cred,
+   * sorted by total cred (descending).
+   */
+  credSortedNodesView(prefix: NodeAddressT): $ReadOnlyArray<CredNode> {
```
decentralion: Why the `View` in `credSortedNodesView`? I think `credSortedNodes` would suffice?

I think we could also go for simpler names: `node` and `nodes`. That it is a credNode is communicated by the types, and implied by the class we're calling the methods on. That the nodes are sorted by descending cred can also be communicated by the docstring. But in general I'm OK with more verbose/descriptive names too. Use your judgement. :)

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+    const match = (a) => NodeAddress.hasPrefix(a, prefix);
+    const addresses: NodeAddressT[] = Array.from(
+      this._timelineCred._cred.addressToCred.keys()
+    ).filter(match);
+    const credNodes = addresses.map((a) => this.credNode(a));
+    return sortBy(credNodes, (x: CredNode) => -x.total);
+  }
+}
+
+export function getDefaultTimelineCredView(tc: TimelineCred) {
```
decentralion: Rather than exporting an additional function for this, let's make the input interval optional in the constructor. If the input interval is not provided, then we use this logic to make a view for the whole timeline.

src/explorer/TimelineCredChart.js
```diff
@@ -48,14 +50,22 @@ export class TimelineCredChart extends React.Component<Props> {
   render() {
     const {timelineCred, displayedNodes} = this.props;
     const intervals = timelineCred.intervals();
+    let timelineCredViewInterval: Interval = {
+      startTimeMs: intervals[0].startTimeMs,
+      endTimeMs: intervals[intervals.length - 1].endTimeMs,
+    };
+    const timelineCredView = new TimelineCredView(
```
decentralion: Assuming we refactor the constructor to make the interval optional, then we shouldn't need to do the manual interval creation here.

src/explorer/TimelineCredViewContainer.js
```diff
@@ -30,10 +31,11 @@ const DEFAULT_ENTRIES_PER_CHART = 6;
  * and then navigating to:
  * http://localhost:8080/test/TimelineCredView/
  */
-export class TimelineCredView extends React.Component<Props> {
+export class TimelineCredViewContainer extends React.Component<Props> {
```
decentralion: I think `TimelineCredViewContainer` is a confusing name. As it's a React component, the concept of `Container` would (IMO) more naturally apply to the React components that it contains, so in that sense it contains a `TimelineCredChart`, not a `TimelineCredView`.

How about we call it a `TimelineCredRenderer`?

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Get the CredNode for a given NodeAddress.
+   *
+   * Returns undefined if the node is not in the filtered results.
+   *
+   * Note that it's possible that the node is present in the Graph, but not the
+   * filtered results; if so, it will return undefined.
+   */
+  credNode(a: NodeAddressT): ?CredNode {
+    const fullView = this._timelineCred._cred.addressToCred.get(a);
```
decentralion: We shouldn't be poking into the private properties of the `TimelineCred` class. Instead, we should have an API on `timelineCred` that allows getting back a `Node` (not a `CredNode`, but a graph `Node`) by address. (Accessing it via `timelineCred.graph().node(...` looks acceptable, but actually `TimelineCred` has a filtered set of nodes compared to the graph, so it should be a separate API that checks what is really available on `TimelineCred`.)

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Return all the nodes matching the prefix, along with their cred,
+   * sorted by total cred (descending).
+   */
+  credSortedNodesView(prefix: NodeAddressT): $ReadOnlyArray<CredNode> {
+    const match = (a) => NodeAddress.hasPrefix(a, prefix);
+    const addresses: NodeAddressT[] = Array.from(
+      this._timelineCred._cred.addressToCred.keys()
```
decentralion: As with the `credNode` method: we shouldn't poke into the internals of `timelineCred`, so we need to add a new API on `timelineCred` for accessing all the nodes available. (That API will return Graph nodes, not CredNodes)

src/analysis/timeline/timelineCred.js
```diff
@@ -203,6 +158,21 @@ export class TimelineCred {
...
+   * or undefined otherwise.
+   */
+  getNodeByAddress(a: NodeAddressT): ?Node {
+    return this.graph().node(a);
+  }
+
+  /**
+   * Returns an array of cred from a given NodeAddressT from FilteredTimelineCred
+   */
+  getCredByAddress(a: NodeAddressT) {
```
decentralion: Should have an explicit type signature. (Not needed technically, but will make the code easier to read.)

decentralion: Assuming it gives interval aligned data, there should also be an API for getting the intervals.

decentralion: Also, the docstring should document what happens when the address is not present (does it throw an error? does it return undefined?)

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,130 @@
...
+    return {cred, total, node};
+  }
+
+  /**
+   * Return all the nodes matching the prefix, along with their cred,
+   * sorted by total cred (descending).
+   */
+  credSortedNodes(prefix: NodeAddressT): $ReadOnlyArray<CredNode> {
+    const match = (a) => NodeAddress.hasPrefix(a, prefix);
+    const addresses: NodeAddressT[] = this._timelineCred
```
decentralion: It's not guaranteed that every node address in the graph will be present in the timelineCred. (timelineCred filters out most of the nodes for space efficiency.) That means that this code will violate its type signature, and return an array containing [undefined].

Instead, we need a method on the base `timelineCred` that returns all the node addresses that it has available.

Consider adding a test that will catch this bug, and then fixing the bug.

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,130 @@
...
+
+  // We are displaying a range of the intervals in the base TimelineCred
+  // We track the start index and end index of the range, so we can efficiently
+  // slice the cred
+  _startIntervalIndex: number;
+  _endIntervalIndex: number;
+
+  constructor(timelineCred: TimelineCred, inputInterval: ?Interval) {
+    this._timelineCred = timelineCred;
+    if (inputInterval === undefined || inputInterval === null) {
```
decentralion: This is the one case where using `==` is preferable to `===`. We can just do:

```js
if (inputInterval == null) {
  this._inputInterval = this._getDefaultInterval(timelineCred);
} else {
  this._inputInterval = inputInterval;
}
```

The following are also fine choices:
```js
this._inputInterval = inputInterval == null 
    ? this._getDefaultInterval(timelineCred) 
    : inputInterval
```
or
```js
this._inputInterval = NullUtil.orElse(
    inputInterval, 
    this._getDefaultInterval(timelineCred)
)
```

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,130 @@
...
+   * Get the CredNode for a given NodeAddress.
+   *
+   * Returns undefined if the node is not in the filtered results.
+   *
+   * Note that it's possible that the node is present in the Graph, but not the
+   * filtered results; if so, it will return undefined.
+   */
+  credNode(a: NodeAddressT): ?CredNode {
+    const fullView = this._timelineCred.getCredByAddress(a);
+    if (fullView === undefined || fullView === null) {
```
decentralion: as above: `fullView == null` suffices.

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+    const addressToCred = new Map();
+    for (const [name, generator] of users) {
+      const address = NodeAddress.fromParts(["foo", name]);
+      graph.addNode({
+        address,
+        description: `[@${name}](https://github.com/${name})`,
+        timestampMs: null,
+      });
+      const scores = intervals.map((_unuesd, i) => generator(i));
+      addressToCred.set(address, scores);
```
decentralion: to add a test that will catch the case above where we fail if there is a node that is in the graph but not in addressToCred: you could just filter one of the users here, so they get added to the graph but not the addressToCred. I would create a new user called "filteredUser" for that purpose.

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+    const params = {alpha: 0.05, intervalDecay: 0.5, weights: defaultWeights()};
+    return new TimelineCred(graph, filteredTimelineCred, params, credConfig());
+  }
+
+  it("JSON serialization works", () => {
+    const json = exampleTimelineCred().toJSON();
+    const tc_ = TimelineCred.fromJSON(json, credConfig());
+    const tcv = new TimelineCredView(exampleTimelineCred());
+    const tcv_ = new TimelineCredView(tc_);
+    expect(tcv.credSortedNodes(NodeAddress.empty)).toEqual(
```
decentralion: the JSON serialization of TimelineCred is already tested in timelineCred.js. And timelineCredView doesn't implement any new JSON serialization. so, we don't need this test.

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+    const tcv = new TimelineCredView(tc);
+    expect(tc.intervals()).toEqual(tcv.intervals());
+  });
+
+  it("single input interval works", () => {
+    const tc = exampleTimelineCred();
+    const startTimeMs = +new Date(2017, 1);
+    const inputInterval = {startTimeMs: startTimeMs, endTimeMs: startTimeMs};
+    const tcv = new TimelineCredView(tc, inputInterval);
+    expect(tcv.intervals().length).toEqual(1);
```
decentralion: Let's add a bit more to this test:

```js
expect(tcv.intervals()).toHaveLength(1);
const interval = tcv.intervals()[0];
expect(interval.startTimeMs).toBeLessThanOrEqualTo(startTimeMs);
expect(interval.endTimeMs).toBeGreaterThanOrEqualTo(startTimeMs);
```

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  it("straddle input interval works", () => {
+    const tc = exampleTimelineCred();
+    const startTimeMs = +new Date(2017, 0, 14);
+    const endTimeMs = +new Date(2017, 0, 14 + 7 + 2);
+    const inputInterval = {startTimeMs: startTimeMs, endTimeMs: endTimeMs};
+    const tcv = new TimelineCredView(tc, inputInterval);
+    expect(tcv.intervals().length).toEqual(3);
```
decentralion: similar to the test above, it would be nice if the test was more specific about how the intervals should relate to the startTimeMs (e.g: startTimeMs is included within the first interval, endTimeMs is included within the last interval). However, I'm OK merging without this change.

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  it("cred sorting works", () => {
```
decentralion: it would be nice to have a test that address filtering works

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+    expect(sorted).toEqual(expected);
+  });
+
+  it("cred aggregation works", () => {
+    const tcv = new TimelineCredView(exampleTimelineCred());
+    const nodes = tcv.credSortedNodes(NodeAddress.empty);
+    for (const node of nodes) {
+      expect(node.total).toEqual(sum(node.cred));
+    }
+  });
```
decentralion: so far, we haven't checked consistency between the intervals and the cred. so it could be that when I request cred for a node, it gives me cred for all of the intervals rather than the filtered ones. or, it could give me cred for the wrong set of intervals, even if it has the right length. we should add a test to ensure the cred and intervals match up properly.

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,109 @@
...
+import {sum} from "d3-array";
+import sortBy from "lodash.sortby";
+import {utcWeek} from "d3-time";
+import {NodeAddress, Graph} from "../../core/graph";
+import {TimelineCred, type TimelineCredConfig} from "./timelineCred";
+import {TimelineCredView} from "./timelineCredView";
+import {type FilteredTimelineCred} from "./filterTimelineCred";
+import {defaultWeights} from "../weights";
+
+describe("src/analysis/timeline/timelineCred", () => {
```
decentralion: sharness is failing because this has the wrong filename.
suggest changing to:

`describe("analysis/timeline/timelineCredView"`

src/core/graph.js
```diff
@@ -432,6 +432,13 @@ export class Graph {
     return result;
   }
 
+  /**
```
decentralion: Please remove this method, as it's not actually needed for this PR (since the timelineCred and the graph may have different sets of nodes).

right now, across the codebase, when we do need node addresses from the graph, we do `Array.from(graph.nodes()).map((x) => x.address)`. if we were to add a node address method to the graph, we would want it in its own PR where we refactor all the calls of that form to use the new method. Also, the node address method would need to return the node addresses in sorted order (like the other nodes methods on the graph) rather than the insertion order of the map.

src/explorer/TimelineCredChart.js
decentralion: (You marked this resolved, but I think it still applies.)

zixuanzh: good catch

src/analysis/timeline/timelineCredView.js
```diff
@@ -0,0 +1,128 @@
...
+  // The Graph Node in question
+  +node: Node,
+  // The total aggregated cred. (Summed across every interval).
+  +total: number,
+  // The timeline sequence of cred (one score per interval).
+  +cred: $ReadOnlyArray<number>,
+|};
+
+/**
+ * TimelineCredView is a subset view of a TimelineCred for a user
```
decentralion: I think this docstring is inaccurate. Describing it as a "Subset view" of TimelineCred made more sense when TimelineCred still had the `.credNodes` API, etc. But now that Timeline Cred doesn't give that info, this isn't a subset view. Also, "TimelineCredView will extend this interval to be contained in the base intervals" is quite confusing.

Rather, I would say something like:

```
TimelineCredView allows a retrieving information about nodes' cred for a specific interval of time. The TimelineCredView wraps a TimelineCred instance, and provides methods like `credNode` and `credSortedNodes` for retrieving information on the available nodes and how much cred they earned. 

Every TimelineCredView is scoped to a certain view interval, which defines the min and max time of cred being viewed. When `CredNode` results are returned, the `total` field will correspond to the view interval of the TimelineCredView. By default, the view interval is chosen to include every underlying interval in the TimelineCred`.
```

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,143 @@
+// @flow
+/*eslint no-unused-vars: 0 */
```
decentralion: Don't disable lint rules wholesale. :)

src/analysis/timeline/timelineCredView.test.js
```diff
@@ -0,0 +1,142 @@
...
+  it("cred sorting works", () => {
+    const tcv = new TimelineCredView(exampleTimelineCred());
+    const sorted = tcv.credSortedNodes(NodeAddress.empty);
+    const expected = sortBy(sorted, (x) => -x.total);
+    expect(sorted).toEqual(expected);
+  });
+
+  it("cred interval consistency checks", () => {
+    const boundaries = utcWeek.range(+new Date(2017, 0), +new Date(2017, 6));
+    const startTimeIndex = Math.floor(Math.random() * boundaries.length - 1);
```
decentralion: Using `Math.random()` in test code is a bad practice; it invites flakey test failures.

For example, suppose that there are 60 possible configurations of startTimeIndex and endTimeIndex using the Math.random testing style, and only 1 of them is broken. Then in expectation, this commit will merge without test failure, and a week or two later we'll first notice the flakey test failure.

yarn.lock
```diff
@@ -6697,7 +6697,7 @@ prelude-ls@~1.1.2:
   resolved "https://registry.yarnpkg.com/prelude-ls/-/prelude-ls-1.1.2.tgz#21932a549f5e52ffd9a827f570e04be62a97da54"
   integrity sha1-IZMqVJ9eUv/ZqCf1cOBL5iqX2lQ=
 
-prettier@1.18.2:
+prettier@^1.18.2:
```
decentralion: This shouldn't be changed in this PR.

decentralion:
Closing as stale.

PR #1261, opened 2019-08-04
Updates logo

ericronne:
Beta version of new logo (color and b/w, with stacked and horizontal versions).


decentralion:
Thanks for the pull @ericronne! I updated it with a few commits that remove some sketch branding from the svgs, and updated the rasterizer to generate pngs from the logos.

A question for you. In our direct messages, you sent over a colored version of the logo that looks like this:

image


However, it seems the svg in this pull looks a bit different:
image

Is this intentional? (I like the top version better.)

Also, I quite liked how the older version looked like an eye:
image

Any chance we could see a variant like that one that's more clearly eye-like?

ericronne:
The color version was a simplified alt approach; I’ll upload the dm’ed and 👁 versions, too 👍

decentralion:
Sounds good, thanks. I like the detailing on the original colored logo; I'm curious to see it set on a dark circle to get the "eye" effect. I think that one might wind up being a winner for me.

ericronne:
@decentralion 👍 ?

decentralion:
Looks great! Thank you @ericronne!

Issue #1232, opened 2019-07-16
Publish sourcecred on npm


decentralion:
We should make depending on SourceCred a lot easier, by publishing it on npm. Users should be able to install the sourcecred CLI; once that's widely available, we can start making it easier to use (e.g. use the sourcecred CLI to create and update an instance, rather than needing `build_static_site.sh`). 

Ideally, we'll also make it possible to take code dependencies on pieces of the SourceCred API--e.g. importing the SourceCred graph and cred calculator.

I'm not sure how difficult this will be, or what the best practices are for publishing webpack-ified tools to npm.

Also, we can consider whether we want to publish everything to `sourcecred`, or use namespacing like `@sourcecred/cli`, `@sourcecred/core`. (Would that also imply splitting up sourcecred/sourcecred into sub repos?)

cc @Beanow and @wchargin

Beanow:
I believe namespacing would make sense. Especially considering various plugin systems and library dependencies are something you may want. But on top of that a meta-package would be cool.

### From the typical user / getting started perspective

From a users perspective who _just_ wants sourcecred added to their project a `sourcecred` meta package may be useful. This could be a 'typical' use-case including perhaps many dependencies like `@sourcecred/core`, `@sourcecred/cli`, `@sourcecred/widgets`, `@sourcecred/themes`, `@sourcecred/github`, `@sourcecred/discourse`, etc.

On top of that, I would ideally use a configuration file for *everything*. Weights, repos to load, project definitions, list of widgets to generate, output directories, caching directories, plugins to use... something like `.sourcecred.yaml`.

Along with that a workflow like `npm i -g sourcecred` to install, `sourcecred init` to generate a sample config file and `sourcecred update` to load, calculate and regenerate everything. Maybe `sourcecred preview` to have a look at it from localhost.

Many tools now use a similar approach. Flow, hugo, prettier, eslint. All with minimal cli arguments and a lot of declarative config.

### A bit more advanced setup

When messing around with manual deployments and multiple versions of `npm i -g sourcecred` starts causing trouble. You'll probably want something a bit more contained and automated.

You could use a `package.json` just to declare exactly what selection of sourcecred packages you're using. Along with your `.sourcecred.yaml` configuration file. Maybe some cache or static files committed that are referenced by your config. And a CI config file if you're automating it.

An idea of such a package.json.
```json
{
  "name": "my-sourcecred-updater",
  "private": true,
  "dependencies": {
[...]

PR #1262, opened 2019-08-05
yarn.lock: update out-of-date `flow-bin` entry

wchargin:
Summary:
In #1259, `flow-bin` was upgraded to 0.104.0 in `package.json`, but no
corresponding change was made in the lock file.

Test Plan:
Running `yarn` is now a no-op.

wchargin-branch: lock-flow-bin-0.104.0


decentralion:
Thanks. I indeed need to configure Greenkeeper to update the lockfile automatically. For public packages it now does that by default, for private ones it needs greenkeeper-lockfile (below). I haven't done it yet because I'm planning to make a public SC package Soon(tm) (#1232)

https://github.com/greenkeeperio/greenkeeper-lockfile

Issue #865, opened 2018-09-20
Create a Discourse Plugin


decentralion:
Today, SourceCred gets data from two [plugins]: the git and github plugins.

[plugins]: https://github.com/sourcecred/sourcecred/tree/master/src/plugins

The plugins load data from Git and GitHub, and map that data into a [graph] showing how contributions in the domain are interrelated. 
Contributions, like Git commits or GitHub pull requests, become nodes in that graph, and relationships between them (authorship, references, etc) are edges.

[graph]: https://github.com/sourcecred/sourcecred/blob/master/src/core/graph.js

SourceCred uses [Discourse] for its community forum, available at https://discourse.sourcecred.io.
We should add a third plugin, the Discourse plugin, which loads data from arbitrary Discourse instances.

[Discourse]: https://www.discourse.org/

To start, it should have the following nodes and edges:

**Nodes:**
- USER, a Discourse login identity
- THREAD, a top-level thread
- POST, an individual pots in a thread (including the first message in the thread)
- CATEGORY, a Discourse category of threads

**Edges:**
- AUTHORS, user authors a post or thread
- REFERENCES, a post references a user, thread, or post (via url or @-refernece)
- QUOTES, a post quotes another post (there's an explicit tag)
- CONTAINS, categories contain threads and threads contain posts
- LIKES, a user likes a post

This is a relatively involved undertaking; whoever decides to do this, please
feel free to drop by our [Discord](https://sourcecred.io/discord-invite) to
chat about how the architecture works.

decentralion:
Closing, because we aren't making active use of Discourse atm.

decentralion:
I spent a bit of time today looking at the [discourse API](https://docs.discourse.org/).

Some thoughts:
- Getting data from the API requires a key to be created by the forum maintainers. This means, in contrast to GitHub, it will be impossible to generate cred for a Discourse forum without explicit buy-in from the admins
- There doesn't seem to be any way to scope API keys with specific permissions, so they're quite sensitive. We should be careful with these.
- The [Topics api](https://docs.discourse.org/#tag/Topics) has endpoints for getting recent topics and top topics, but not for getting all topics. This could be a problem, except:
[...]

PR #1265, opened 2019-08-06
Add class for fetching data from Discourse

decentralion:
The `DiscourseFetcher` class abstracts over fetching from the Discourse
API, and post-processing and filtering the result into a form that's
convenient for us.

Testing is a bit tricky because the Discourse API keys are sensitive
(they are admin keys) and so I'm reluctant to commit them, even for our
test instance. As a workaround, I've added a shell script which
downloads some data from the SourceCred test instance, and saves it with
a filename which is an encoding of the actual endpoint. Then, in
testing, we can use a mocked fetch which actually hits the snapshots
directory, and thus validate the processing logic on "real" data from
the server. We also test that the fetch headers are set correctly, and
that we handle non-200 error codes appropriately.

Test plan: In addition to the included tests, I have an end-to-end test
which actually uses this fetcher to fully populate the mirror and then
generate a valid SourceCred graph.

This builds on API investigations
[here](https://github.com/sourcecred/sourcecred/issues/865#issuecomment-478026449),
and is general progress towards #865. Thanks to @erlend-sh, without whom
we wouldn't have a test instance.


decentralion:
FYI  @wchargin; feel free to review the whole PR, but I've added you just to take a look at the snapshot updating shell script. (In my head you have "shell readability", and I don't. :wink:)

scripts/update_snapshots.sh
```diff
@@ -34,5 +34,12 @@ echo "Updating github/fetchGithubOrgTest.sh"
 echo "Updating github/fetchGithubRepoTest.sh"
 ./src/plugins/github/fetchGithubRepoTest.sh -u --no-build
 
+if [ -z "${DISCOURSE_TEST_API_KEY:-}" ]; then
+  echo "Updating Discourse API snapshots"
+  ./src/plugins/discourse/update_discourse_api_snapshots.sh
+else
+  echo "Not updating Discourse API snapshots (need DISCOURSE_TEST_API_KEY)"
+fi
```
wchargin: lgtm for file

src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -0,0 +1,34 @@
...
+set -eu
+
+snapshots_dir=src/plugins/discourse/snapshots
+test_instance_url="https://sourcecred-test.discourse.group"
+test_instance_username="credbot"
+
+if [ -z "${DISCOURSE_TEST_API_KEY:-}" ]; then
+  printf >&2 'Please set the DISCOURSE_TEST_API_KEY environment variable.\n'
+  printf >&2 'Contact the SourceCred maintainers to get the key.\n'
+  return 1
```
wchargin: This will fail—you can only `return` from a function:

```
Please set the DISCOURSE_TEST_API_KEY environment variable.
Contact the SourceCred maintainers to get the key.
./test.sh: line 11: return: can only `return' from a function or sourced script
```

Try `exit 1`, or wrap the whole thing in a `main` function.

src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+cd "${toplevel}"
+
+fetch() {
+  url="${test_instance_url}${1}"
```
wchargin: style nit: We generally don’t brace positional parameters (though we
still do quote them, of course): suggest `$1` rather than `${1}`.

src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -0,0 +1,34 @@
...
+  filename=$(echo -n "${url}" | base64 -w 0 | tr -d '=' | tr '/+' '_-')
```
wchargin: Nice pipeline; good use of filters.

wchargin: Prefer `printf '%s' "${url}"` to `echo -n "${url}"`. The behavior of
`echo -n` is not portable, and even bare `echo` is not portable on
arbitrary inputs (e.g., if any argument contains a backslash).

[POSIX recommends][1] without qualification that new applications use
`printf` rather than `echo`.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html#tag_20_37_16

src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -0,0 +1,34 @@
...
+  path="${snapshots_dir}/${filename}"
+  curl $url \
```
wchargin: Quotes! It matters this time: some of your `url`s have question mark
characters, which are interpreted specially as file expansions by the
shell.

src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -0,0 +1,34 @@
...
+    -H "Api-Key:${DISCOURSE_TEST_API_KEY}" \
```
wchargin: nit: HTTP headers are written with a space after the colon.

src/plugins/discourse/update_discourse_api_snapshots.sh
wchargin: Recommend quoting `filename="$(…)"`. It’s not technically necessary in
variable assignments, but if you quote then it’s _clearly_ correct
rather than _merely_ correct.

src/plugins/discourse/update_discourse_api_snapshots.sh
wchargin: I typically incant `curl -sfL`: silent (omit progress bar), fail on HTTP
errors like 404, and follow HTTP redirects. You may or may not want `-s`
here, but you probably do want `-f` and `-L`.

src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -0,0 +1,34 @@
...
+    > "${path}"
+}
+
+rm -r "${snapshots_dir}"
+mkdir "${snapshots_dir}"
+fetch "/latest.json?order=created"
+fetch "/posts.json"
+fetch "/t/11.json"
+fetch "/posts/14.json"
+
```
wchargin: nit: Extraneous blank line.

wchargin:
Will review for SQL; stay tuned.

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+    if (response.status === 404) {
+      throw new Error(`404 from ${response.url}; maybe bad serverUrl?`);
+    }
+    if (response.status === 403) {
+      const {apiUsername, apiKey} = this.options;
+      throw new Error(
+        `got status 403; bad api username (${apiUsername}) or key (${apiKey})`
+      );
+    }
+    if (response.status !== 200) {
```
wchargin: Use `response.ok`, which checks for 2xx status codes. There are valid
status codes other than 200, and more may be added in the future.

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+
+  async latestPosts(): Promise<$ReadOnlyArray<Post>> {
+    const response = await this._fetch("/posts.json");
+    if (response.status === 404) {
+      throw new Error(`404 from ${response.url}; maybe bad serverUrl?`);
+    }
+    if (response.status === 403) {
+      const {apiUsername, apiKey} = this.options;
+      throw new Error(
+        `got status 403; bad api username (${apiUsername}) or key (${apiKey})`
```
wchargin: Again on the security note.

Perhaps encapsulate this logic into a function? It’s duplicated verbatim
across these two functions.

wchargin: Again on the security note.

Perhaps encapsulate this logic into a function? It’s duplicated verbatim
across these two functions.

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+    const json = await response.json();
+    return parsePost(json);
+  }
+}
+
+function parsePost(json: any): Post {
+  return {
+    id: json.id,
+    timestampMs: +new Date(json.created_at),
```
wchargin: This is [the same as][1] `Date.parse(json.created_at)`.

[1]: https://tc39.es/ecma262/#sec-date-value

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+  // individually.
+  +posts: $ReadOnlyArray<Post>,
+|};
+
+/**
+ * Interface over the external Discourse API, structured to suit our particular needs.
+ * We have an interface (as opposed to just an implementation) to enable easy mocking and
+ * testing.
+ */
+export interface DiscourseInterface {
```
wchargin: Some of these methods can resolve to `null` in some cases _and_ reject
in others. The documentation needs to discuss this for clients to be
able to implement this interface and consume implementations.

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+  // Retrieve the Topic with Posts for a given id. May be null, either because the
+  // topic is hidden from the api user, or it 404s. (Not sure why this happens.)
+  topicWithPosts(id: TopicId): Promise<TopicWithPosts | null>;
+  // Retrieve an individual Post by its id. May be null, e.g. because the post is
+  // hidden from the api user.
+  post(id: PostId): Promise<Post | null>;
+  // Retrieve the latest posts from the server.
+  // Vital so that we can then enumerate and fetch every Post that we haven't
+  // encountered.
+  latestPosts(): Promise<$ReadOnlyArray<Post>>;
```
wchargin: Should this return a promise of mutable-array of `Post`? That’s strictly
more useful for clients, and shouldn’t be a burden on implementations
because the array is intended to be constructed immediately before being
returned, rather than being a reference into a shared cache.

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+  +title: string,
+  +timestampMs: number,
+  +authorUsername: string,
+|};
+
+export type Post = {|
+  +id: PostId,
+  +topicId: TopicId,
+  // Which number post this was within the topic (starts at 1)
+  +postNumber: number,
```
wchargin: Consider calling this `indexInTopic` or similar? It’s not clear from the
names why “`id`” and “`postNumber`” should be two separate fields. (If
so, also rename `replyToPostIndex` for consistency.)

wchargin: Consider calling this `indexInTopic` or similar? It’s not clear from the
names why “`id`” and “`postNumber`” should be two separate fields. (If
so, also rename `replyToPostIndex` for consistency.)

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+  //
+  // We do use these Posts though, as it allows us to save requesting them all
+  // individually.
+  +posts: $ReadOnlyArray<Post>,
+|};
+
+/**
+ * Interface over the external Discourse API, structured to suit our particular needs.
+ * We have an interface (as opposed to just an implementation) to enable easy mocking and
+ * testing.
```
wchargin: Well, and also the Discourse `Mirror` should implement this interface,
right?

src/plugins/discourse/fetch.js
wchargin: No need for `interface FooInterface`. Can it just be `Discourse`?

In general, interfaces should get “the good names” if there’s a
conflict, because clients work primarily with interfaces, not concrete
implementations.

src/plugins/discourse/fetch.js
decentralion: 👍

src/plugins/discourse/fetch.js
```diff
@@ -0,0 +1,212 @@
...
+
+  async latestTopicId(): Promise<TopicId> {
+    const response = await this._fetch("/latest.json?order=created");
+    if (response.status === 404) {
+      throw new Error(`404 from ${response.url}; maybe bad serverUrl?`);
+    }
+    if (response.status === 403) {
+      const {apiUsername, apiKey} = this.options;
+      throw new Error(
+        `got status 403; bad api username (${apiUsername}) or key (${apiKey})`
```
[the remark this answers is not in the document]
decentralion: Fixed, thanks.

src/plugins/discourse/fetch.js
decentralion: Done.

src/plugins/discourse/fetch.js
decentralion: Done.

src/plugins/discourse/fetch.js
decentralion: Sure, why not.

src/plugins/discourse/fetch.js
decentralion: I chose postNumber to mirror the Discourse API. But I'm happy to rename their fields; indexWithinTopic it is.

src/plugins/discourse/fetch.js
decentralion: No, this interface is the right interface if you want to do an update (find out what the latest indices are, compare to stored, and fetch new stuff). If you want to create a graph from scratch, then you just want methods that give you every topic and every post so you can iterate over them conveniently. So the mirror has a different interface.

src/plugins/discourse/fetch.js
decentralion: Sure

decentralion:
Thanks for the review @wchargin!

Issue #1256, opened 2019-07-24
Make SourceCred robust on large repositories


decentralion:
Right now, there are a lot of issues with downloading large GitHub repositories that conspire to make SourceCred brittle on large repos. #1023 is an example of this.

Here are the sub-issues I'm aware of
- [ ] Not robust to GitHub entities that have changed their typename (#998)
- [ ] Not robust to GitHub node deletion (#938)
- [ ] We need to handle the known contract violation errors in the api ([#996])

This issue will track progress towards the overall goal of making SC robust and reliable on large repos, like TensorFlow.

Beanow:
Perhaps useful to document problematic repos for testing, I ran into more but let me update this as I bump into them.

Error: Cannot update data for nonexistent node: undefined
- passbolt/passbolt_api

decentralion:
@Beanow one of the ways this issue comes up is if a node was deleted. Possibly during the update itself, but most likely between when you created your cache and your latest update. (See #938). So if you haven't already, I recommend clearing your cache and trying again.

The very least we can do is try to get better error messages on all of these issues, so we at least know why the update failed, give a hint as to whether a cache clear may fix it, and link to the right issue on GitHub.

@wchargin would you be willing to take on just updating the error messages so they point to the right failure cause?

Beanow:
I believe it wasn't the cache. Try running from a container with a fresh data folder:

```sh
time SOURCECRED_GITHUB_TOKEN="xxxxxx" \
  docker run -ti --rm --env SOURCECRED_GITHUB_TOKEN \
  -p 8080:8080 sourcecred/sourcecred dev-preview quasarframework/quasar-cli
```

<pre><span style="background-color:#3465A4"><font color="#EEEEEC"><b>  GO  </b></font></span> load-quasarframework/quasar-cli
<span style="background-color:#3465A4"><font color="#EEEEEC"><b>  GO  </b></font></span> github/quasarframework/quasar-cli
Error: Cannot update data for nonexistent node: undefined
    at Mirror._nontransactionallyUpdateOwnData (webpack:///./src/graphql/mirror.js?:412:161)
    at Mirror._nontransactionallyUpdateData (webpack:///./src/graphql/mirror.js?:239:235)
    at eval (webpack:///./src/graphql/mirror.js?:244:638)
    at _inTransaction (webpack:///./src/graphql/mirror.js?:560:142)
    at Mirror._updateStep (webpack:///./src/graphql/mirror.js?:244:555)
[...]

PR #1270, opened 2019-08-07
Include full projects on sourcecred.io

decentralion:
This updates the deploys script so that we now load full projects for
@libp2p, @ipld, @sourcecred, and @filecoin-project.

I'd like to support @ipfs, but we need to tackle #1256 first.

Test plan: Ran deploy, verified that it produced good output, and then actually deployed.


Beanow:
Are repo's still supported? If so, wouldn't including the previous two ipfs repos offer a nice smoke test?

decentralion:
Repos are still supported (via `--project ipfs/js-ipfs` for instance). Yeah, I'd be fine adding them back. I had an issue loading go-ipfs earlier (maybe a new bad node appeared?) but maybe it was ephemeral (deletion while I was loading, or such). Assuming there's no issue loading, i"d be happy to have the projects back.

PR #1280, opened 2019-08-13
Fix type error in FileUploader

wchargin:
Summary:
Introduced in #1277.

Test Plan:
Run `yarn start` and visit <http://localhost:8080/test/FileUploader/>.
Conduct the test plan as specified on that page.

wchargin-branch: fileuploader-target

### DIFF
--- a/src/util/FileUploader.js
+++ b/src/util/FileUploader.js
@@ -24,8 +24,11 @@ export class FileUploader extends React.Component<Props> {
     const onUpload = (e) => {
       const file = e.target.files[0];
       const reader = new FileReader();
-      reader.onload = (e) => {
+      reader.onload = () => {
         const jsonString = reader.result;
+        if (typeof jsonString !== "string") {
+          throw new Error("Unexpected: jsonString is not string");
+        }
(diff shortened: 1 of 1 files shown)


decentralion:
Hmm, it seems your commit had a type and lint error. I pushed up a fix, take a look?

wchargin:
Lint error is my fault; should have checked that. I can also reproduce
the Flow error now, though it passed before. (Flow server in a bad
state, maybe?) Thanks for the patch! PTAL.

decentralion:
It's a little funny to me that all of this will boost the cred of #1277, by virtue of the fact that I added an uncaught build error.

Then again, this is where manual mode could come in. Force the weight of that pull to 1/8 since it broke the build!

wchargin:
Yeah, a bit funny, indeed.

I considered just immediately reverting that commit, as one usually does
for build/test failures, but opted to fix forward because I expected the
bottleneck to be review/merge time rather than fix time, anyway. But
another argument for rolling back is that it creates the correct cred
signal.

PR #1286, opened 2019-08-14
mirror: allow updating blacklisted ids

decentralion:
Now the GitHub mirror will warn when the blacklisted ids are
inconsistent, rather than throwing an error.

This bit of tolerance is quite convenient, because you might know that
the change in blacklisted ids was for a totally different repo, and have
a reason to want to keep your cache around. Of course, you'll eventually
want to reset the cache so that you stop seeing the warning message. :)

Test plan: Tests updated; `yarn test` passes.

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -1,6 +1,7 @@
 // @flow
 
 import deepFreeze from "deep-freeze";
+import deepEqual from "lodash.isequal";
 import type Database, {BindingDictionary, Statement} from "better-sqlite3";
 import stringify from "json-stable-stringify";
 
@@ -9,6 +10,19 @@ import * as MapUtil from "../util/map";
 import * as NullUtil from "../util/null";
 import * as Schema from "./schema";
 import * as Queries from "./queries";
+
+// The following version number must be updated if there is any
+// change to the way in which a GraphQL schema is mapped to a SQL
+// schema or the way in which the resulting SQL schema is
+// interpreted. If you've made a change and you're not sure whether
+// it requires bumping the version, bump it: requiring some extra
+// one-time cache resets is okay; doing the wrong thing is not.
+const MIRROR_VERSION = "MIRROR_v3";
+type MirrorConfig = {|
+  +version: string,
+  +schema: Schema.Schema,
+  +blacklistedIds: {|+[Schema.ObjectId]: true|},
+|};
 
 /**
  * A local mirror of a subset of a GraphQL database.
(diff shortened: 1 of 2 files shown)


wchargin:
What unexpected behavior may result? Data loss? Incorrect responses?
Failure to terminate?

src/graphql/mirror.js
```diff
@@ -200,16 +207,36 @@ export class Mirror {
...
+            "Database already populated with incompatible schema."
+          );
+        } else if (
+          !deepEqual(existingConfig.blacklistedIds, mirrorConfig.blacklistedIds)
+        ) {
+          console.error(dedent`\
+            Warning: Database already loaded with different blacklistedIds.\
+            Unexpected behavior may result. Consider resetting your cache.`);
+          // We don't need to do the rest of the initialization, because it was
+          // setup, albiet with different options.
```
wchargin: sp.: *albeit* (“although be it”)

Also, *set up* as a phrasal verb is two words.

src/graphql/mirror.js
```diff
@@ -200,16 +207,36 @@ export class Mirror {
...
+            "Database already populated with incompatible version."
+          );
+        } else if (!deepEqual(existingConfig.schema, mirrorConfig.schema)) {
+          throw new Error(
+            "Database already populated with incompatible schema."
+          );
+        } else if (
+          !deepEqual(existingConfig.blacklistedIds, mirrorConfig.blacklistedIds)
+        ) {
+          console.error(dedent`\
```
wchargin: A warning should use `console.warn`, no?

src/graphql/mirror.js
```diff
@@ -200,16 +207,36 @@ export class Mirror {
...
+            Warning: Database already loaded with different blacklistedIds.\
```
wchargin: Also, this won’t format as you expect; it will have 12 interior spaces.

decentralion:
I actually have no idea what kinds of bad things may happen. They may happen, and they are truly unexpected. :)

wchargin:
FWIW, the fact that this PR was not merged just saved me a bit of
trouble. I’d run

```
SOURCECRED_DIRECTORY=/tmp/testsc/ node ./bin/sourcecred.js load aragon/aragon
```

to test some behavior on a clean cache. I went off and did some
investigation, then came back and wanted to retry the load from where it
failed, so I reran my last `node ./bin/sourcecred.js` invocation—but
this omitted the environment variable, keying against my primary cache
instead. Happily, the load pipeline failed immediately with

> Error: Database already populated with incompatible schema, options,
> or version

rather than corrupting the database that had the proper set of
blacklisted IDs.

decentralion:
Closing as I don't feel we have consensus this is a good idea.

Issue #1272, opened 2019-08-08
Adding a frontend CLI command


Beanow:
As a preparation for NPM packages #1232
Having a lightweight `sourcecred frontend` command of sorts would be swell.

Ideally we would not need to use webpack, babel, flow, etc for this command.
It would take a pre-compiled frontend template and make the necessary modifications based on the loaded sourcecred data.

Looking into the current situation, I think that means preparing `/api/v1/data` and injecting some titles and hrefs in `/(timeline|prototype)/<org>/<repo>/index.html`.

Doing this would allow an NPM package to ship and have both frontend and backend capabilities without depending on all the webpack tooling. Slimming down the package significantly.

decentralion:
This is a great suggestion @Beanow. Getting something like this working well is vital to making the future sourcecred command fully usable for setting up an instance.

I think the trickiest bit will be getting the routing for the projects to work correctly. The way the project ids get piped around is a little weird, they get put into the process environment in [config/env.js](https://github.com/sourcecred/sourcecred/blob/master/config/env.js#L145), you can grep for `PROJECT_IDS` to find the usage.

Beanow:
Note from #1232 musings, the frontend generation may be part of the update command, so like the refactored load command is probably best as a combination of a javascript api + cli code to call that.

Beanow:
New suggestion here https://discourse.sourcecred.io/t/sourcecred-instance-system/244/2
To avoid needing to compile the frontend entirely.

PR #1288, opened 2019-08-14
Adding Docker container with instructions for running sourcecred

vsoch:
This pull request will add a Dockerfile to build a container that can be used to easily run sourcecred, without needing to deal with dependencies. I've also pushed a version to dockerhub [vanessa/sourcecred](https://hub.docker.com/r/vanessa/sourcecred) if you want to test without building!

Complete instructions (and images) are added to the README, but I'll reiterate here.

The user can optionally build the container:

```bash
$ docker build -t sourcecred .
```

But they don't have to - they can use [vanessa/sourcecred](https://hub.docker.com/r/vanessa/sourcecred)

```bash
$ docker pull vanessa/sourcecred
$ docker tag vanessa/sourcecred sourcecred
```

They still need to export a Github access token and provide it to the container, and then run the container providing the token, exposing a port, and providing the GitHub repo of interest:

```bash
REPOSITORY=sfosc/sfosc
$ docker run -d --name sourcecred --rm --env SOURCECRED_GITHUB_TOKEN=$(cat .token) -p 8080:8080 sourcecred $REPOSITORY
```
After running the command, you can inspect it's progress like this:

```bash
$ docker logs sourcecred
  GO   load-sfosc/sfosc
  GO   github/sfosc/sfosc
 DONE  github/sfosc/sfosc: 25s
  GO   compute-cred
 DONE  compute-cred: 1s
 DONE  load-sfosc/sfosc: 26s
...
```

It will take about 30 seconds to do the initial build, and when the web server is running you'll see
this at the end:

```bash
...
[./node_modules/react/index.js] 190 bytes {main} {ssr} [built]
[./src/homepage/index.js] 1.37 KiB {main} [built]
[./src/homepage/server.js] 5.61 KiB {ssr} [built]
    + 1006 hidden modules
ℹ ｢wdm｣: Compiled successfully.
```

Then you can open up to [http://127.0.0.1:8080](http://127.0.0.1:8080) to see the interface! I added images to the README for the user to see what is expected.

I'd like to add @Beanow widgets so that the container can be used to generate the svg files for output, that way any user can (with one command) use the tool to generate the images (which is what I sort of want for my repos :P)


scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [ -z "${SOURCECRED_GITHUB_TOKEN}" ]; then
+    echo "Export SOURCECRED_GITHUB_TOKEN with the container."
+    echo "docker run --env SOURCECRED_GITHUB_TOKEN=$(cat .token) -p 8080:8080 sourcecred <repository>"
+    exit 1;
+fi
+
+# serves on 8080 in docker container, should be bound to host
+node /code/bin/sourcecred.js load "${@}"
+yarn start --host 0.0.0.0
```
Beanow: While this is great for development, I think we should definitely move away from this when publishing an image. The underlying webpack dev-server is as the name suggest intended for development with hot-reloading of files and whatnot. I wouldn't dare run it (as root :scream:) on a public facing image by default.

Until we have a better CLI (discussed in #1232), perhaps (for production use) we should encourage users to use this image to build static files for example for gh-pages or as part of a multi-stage build into nginx?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    apt-get update && \
+    apt-get install -y yarn build-essential && \
+    mkdir -p /code
+WORKDIR /code
+ADD . /code
+RUN curl --compressed -o- -L https://yarnpkg.com/install.sh | bash && \
+    yarn && \
+    yarn backend
+ENTRYPOINT ["/bin/bash", "/code/scripts/docker-entrypoint.sh"]
```
Beanow: Making the `sh` file executable you can use it's crunchbang instead of explicit bash calls.

scripts/docker-entrypoint.sh
Beanow: From the top of my head, this call may also need to be run with `exec` in order to pass on signals.

scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ -z "${SOURCECRED_GITHUB_TOKEN}" ]; then
+    echo "Export SOURCECRED_GITHUB_TOKEN with the container."
+    echo "docker run --env SOURCECRED_GITHUB_TOKEN=$(cat .token) -p 8080:8080 sourcecred <repository>"
```
wchargin: This will actually print out the token in the message; the command
substitution will be expanded. That needs to be escaped, so that the
token doesn’t end up in logs or the user’s terminal window.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:10
+# docker build -t sourcecred .
+# available at vanessa/sourcecred if you don't want to build
+RUN apt-get update && \
+    apt-get install -y apt-transport-https python
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
```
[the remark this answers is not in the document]
vsoch: You got it! I'm so new with front end packages I was just Tim Gunning it... make it work! :)

scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ -z "${SOURCECRED_GITHUB_TOKEN}" ]; then
```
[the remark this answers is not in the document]
vsoch: Does the command spit out a nice human error message if it's not provided? If so, great, and I'll remove this here.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    apt-get update && \
+    apt-get install -y yarn build-essential && \
+    mkdir -p /code
+WORKDIR /code
+ADD . /code
+RUN curl --compressed -o- -L https://yarnpkg.com/install.sh | bash && \
+    yarn && \
```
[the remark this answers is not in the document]
vsoch: Good suggestion, I'll try this now.

Dockerfile
wchargin: I’m no Docker expert, but is `curl | apt-key add -` really the safest
way to do this? Adding data from the internet as _package signing keys_
seems pretty alarming.

Dockerfile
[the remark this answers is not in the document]
vsoch: And if unset, where does it go (where is it going now?). It would be good to have a consistent export location, and then we can give instruction to the user to bind as a volume. If it doesn't matter, I'll set it as SOURCECRED_DIRECTORY=/data.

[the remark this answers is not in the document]
vsoch: And I'll do a build arg so some future user can change it.

scripts/docker-entrypoint.sh
vsoch: That's the only way to expose it to the host, it's still only intended for development. I will add a note to the README to exemplify this... but if we don't set that hostname it's not viewable from the host.

vsoch: Yes the best use case would be to generate those beautiful svg's you showed me - for now I was just starting with the basics (as described in this readme) because the widgets are separate, no?

Dockerfile
vsoch: The sh file is executable, but I always make explicit calls to the binary I want used - it's good practice.

scripts/docker-entrypoint.sh
vsoch: This would make sense - when I tried without detached it hung and I couldn't control C. Can you show me an example of what it should be (`yarn exec --host 0.0.0.0`) ?

README.md
```diff
@@ -61,6 +67,77 @@ yarn start
...
+```
+
+You will still need to export a GitHub token, and then provide it to the container
+when you run it. Notice that we are also binding port 8080 so we can
+view the web interface that will be opened up.  The only argument needed is the 
+GitHub repository to generate the sourcecred for:
+
+```bash
+REPOSITORY=sfosc/sfosc
+$ docker run -d --name sourcecred --rm --env SOURCECRED_GITHUB_TOKEN=$(cat .token) -p 8080:8080 sourcecred $REPOSITORY
```
wchargin: Using `--env SOURCECRED_GITHUB_TOKEN="$(cat .token)"` s unsafe: it puts
the contents of the token in the command line, and the command lines of
processes are visible to _all users_ on the system.

The standard mechanism to pass secrets securely across process
boundaries is the environment, because the environment of a process is
not visible to other users. Docker supports this: passing `--env FOO`
instead of `--env FOO=bar` will pass the environment’s value of `FOO`
down to the container:

```shell
SOURCECRED_GITHUB_TOKEN="$(cat .token)" \
    docker run --env SOURCECRED_GITHUB_TOKEN …
```

wchargin: This is also unsafe as written because `"$(cat .token)"` needs to be
quoted, lest it be subject to glob expansion and word splitting. So does
`"${REPOSITORY}"`, and you should probably include a `--` before that
argument to prevent the repository name from being interpreted as a
Docker option.

Dockerfile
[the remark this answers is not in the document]
vsoch: Just a heads up - installing from npm spits out an error about missing webpack, and from experience I'd expect webpack-cli next. And then when I change the order of adding files, it fails massively again (it took me several hours to get around this error)

```bash
$ NODE_ENV=development webpack --config config/webpack.config.backend.js
/code/node_modules/webpack-cli/bin/cli.js:93
				throw err;
				^

Error: Cannot find module '/code/config/webpack.config.backend.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (/code/node_modules/v8-compile-cache/v8-compile-cache.js:161:20)
    at WEBPACK_OPTIONS (/code/node_modules/webpack-cli/bin/utils/convert-argv.js:116:13)
    at requireConfig (/code/node_modules/webpack-cli/bin/utils/convert-argv.js:118:6)
    at /code/node_modules/webpack-cli/bin/utils/convert-argv.js:125:17
    at Array.forEach (<anonymous>)
    at module.exports (/code/node_modules/webpack-cli/bin/utils/convert-argv.js:123:15)
    at yargs.parse (/code/node_modules/webpack-cli/bin/cli.js:71:45)
    at Object.parse (/code/node_modules/webpack-cli/node_modules/yargs/yargs.js:567:18)
    at /code/node_modules/webpack-cli/bin/cli.js:49:8
    at Object.<anonymous> (/code/node_modules/webpack-cli/bin/cli.js:365:3)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/code/node_modules/webpack/bin/webpack.js:156:2)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
error Command failed with exit code 1.
```
so I'm not going to be able to change either of those things, maybe you can take a look and test?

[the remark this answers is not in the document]
vsoch: For now I'm leaving as I have it, which is a working build.

Dockerfile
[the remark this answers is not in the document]
vsoch: Note: this didn't work.

scripts/docker-entrypoint.sh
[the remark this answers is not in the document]
vsoch: I'm going to leave this in for now - it can be a silent error if the user doesn't provide it (just tested)

```bash
$ docker logs sourcecred
fatal: SOURCECRED_GITHUB_TOKEN not set
fatal: run 'sourcecred help load' for help
yarn run v1.17.3
$ NODE_ENV=development webpack-dev-server --config config/webpack.config.web.js --host 0.0.0.0
(node:35) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
ℹ ｢wds｣: Project is running at http://0.0.0.0:8080/webpack-dev-server/
ℹ ｢wds｣: webpack output is served from /
ℹ ｢wds｣: Content not from webpack is served from /code
```
It should not continue given the variable is missing.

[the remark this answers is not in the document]
vsoch: When the command is customized (e.g., scores added) we can do a check for that and act appropriately. For now it just serves to load and then run, so I'm leaving it.

README.md
```diff
@@ -39,17 +39,23 @@ First, make sure that you have the following dependencies:
 [Node]: https://nodejs.org/en/
 [Yarn]: https://yarnpkg.com/lang/en/
 [GitHub API token]: https://github.com/settings/tokens
-
 [macos-gnu]: https://github.com/sourcecred/sourcecred/issues/698#issuecomment-417202213
 
+When you create a token, a good strategy is to write it into a local file that
+is ignored by the repository. For example, we can write it into `.token`.
```
wchargin: We’ve generally avoided explicit recommendations as to where to store
tokens. For instance, if storing in a token file, you’d need to be sure
to `chmod go-rwx` the token file. Adding it to the `.gitignore` is
obviously a requirement, but an accidental `git add -f .` will still
immediately copy it into the—unsecured!—Git object store as a new blob.
Tarring up the whole repository directory would include the secret, too.
It’ll be removed by a `git clean -qfdx`. And so on.

Can we remove the `.token` suggestions from the gitignore (etc.) and
copy, and instead only describe that the token needs to be provided as
an environment variable?

vsoch:
@Beanow I've made all the changes that I'm able to (that don't result in errors!) If you want to give a go at trying some of the suggestions you made, please do and let me know if you can get it to work. Much apology that I'm not super great at debugging npm / yarn :P I'm quite afraid of them actually, haha.

scripts/docker-entrypoint.sh
vsoch: It can't - the command is run from inside the container, and the .token is in the .dockerignore. But it _will_ spit out an error to not have the file, so I'll fix the quotes for that!

scripts/docker-entrypoint.sh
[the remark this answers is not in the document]
Beanow: > I'm going to leave this in for now - it can be a silent error if the user doesn't provide it (just tested)
> 
> It should not continue given the variable is missing.

I think this is because you're not using the exit value of the `load` command. Try adding appending `|| exit 1` or using `&&` between load and start.

scripts/docker-entrypoint.sh
vsoch: Fixed.

Dockerfile
vsoch: Installing yarn with the package manager (as I was originally doing) we have to add the key to get the package. But since I'm using the install script later, there is a good chance I can just remove this now. Testing now!

README.md
vsoch: Good thing I'm the only user on my system :) But happy to make this change.

vsoch: yep we discussed this in your first comment! It won't be interpreted as a docker option because it's coming after the container name.

Dockerfile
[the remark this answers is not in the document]
Beanow: > And if unset, where does it go (where is it going now?). It would be good to have a consistent export location, and then we can give instruction to the user to bind as a volume. If it doesn't matter, I'll set it as SOURCECRED_DIRECTORY=/data.

The default I believe is `${PWD}/sourcecred_data` meaning it would start writing it to a new layer of the container.

What's the advantage of a build arg over env + volume here?

README.md
vsoch: Sure, works for me.

README.md
vsoch: Done.

Dockerfile
vsoch: Good catch here! We're good without the extra step.

scripts/docker-entrypoint.sh
Beanow: > That's the only way to expose it to the host, it's still only intended for development. I will add a note to the README to exemplify this... but if we don't set that hostname it's not viewable from the host.

Oh I don't mean the `0.0.0.0` host. This is pretty common for docker containers and have no issues with that. I mean `yarn start` is handled by webpack's dev-server and has a lot of features unsafe for production (and a history of having vulnerabilities bubbling up from large dependency list). I would rather this is part of local development only and not the image's default.

For example, I've used `docker-compose` a lot as a way to describe development commands. While the Dockerfile image defaults to production commands.

Dockerfile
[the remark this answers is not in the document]
vsoch: The build arg just means that if someone wants to re-build the container, they can provide the `--build-arg`  flag for some custom value. It will default to what we specified.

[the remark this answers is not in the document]
vsoch: Docker always has one writable overlay layer (hence why you can write) and it would use it for the default or /data, regardless of the path I think. Singularity would be another story - we wouldn't have writable at all and would need to specify an overlay or bind to the host.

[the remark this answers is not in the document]
vsoch: Oh oups, I need to set the build arg to be the environment variable, totally goofed that. :)

scripts/docker-entrypoint.sh
Beanow: It's a built-in sh/bash function and it goes in front of your command.

```sh
exec yarn --host 0.0.0.0
```

```
$ help exec
exec: exec [-cl] [-a name] [command [arguments ...]] [redirection ...]
    Replace the shell with the given command.
    
    Execute COMMAND, replacing this shell with the specified program.
    ARGUMENTS become the arguments to COMMAND.  If COMMAND is not specified,
    any redirections take effect in the current shell.
    
    Options:
      -a name	pass NAME as the zeroth argument to COMMAND
      -c	execute COMMAND with an empty environment
      -l	place a dash in the zeroth argument to COMMAND
    
    If the command cannot be executed, a non-interactive shell exits, unless
    the shell option `execfail' is set.
    
    Exit Status:
    Returns success unless COMMAND is not found or a redirection error occurs.

```

vsoch: I was just following instructions in the README, I have no way to know what "production" sourcecred looks like, but I'm happy to adjust if you tell me :)

Dockerfile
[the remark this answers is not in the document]
vsoch: okay I fixed my goof! Just pushed a change so now the variable is export:
```bash
$ docker run -it --entrypoint bash sourcecred
root@a28e47d07cad:/code# echo $SOURCECRED_DIRECTORY 
/data
```

scripts/docker-entrypoint.sh
vsoch: Oh of course - you want to pass execution off to yarn. Sure let me try that now, it would be better usability to be able to run and kill with control C than in detached.

.dockerignore
```diff
@@ -0,0 +1 @@
+.token
```
Beanow: I just realized important additions for this would also be:
- `bin` to skip built backends
- `node_modules` for consistency

For CI it shouldn't matter, but on my dev machine it reduced the build context from 291.4MB to 14.08MB.

README.md
wchargin: Thank you!

Beanow:
@vsoch I've pushed a version of the dockerfile that includes the `npm install -g yarn` approach and splits node module dependencies from the rest of the source code. What do you think of it?

A thing I haven't modified yet is `VOLUME` isn't used here. Can it use the `ARG` in the same way as the `ENV`?

README.md
wchargin: I’m also the only user on my system, except that I’m not:

```
$ ps -e -o user= | sort -u
avahi
colord
geoclue
messagebus
nobody
ntp
root
rtkit
syslog
uuidd
wchargin
```

For instance, `geoclue` is a service that determines my location and
provides it to applications like `redshift`. To do so, it may make
network requests against a GeoIP service. It shouldn’t be able to read
my token file.

(And of course lots of people use servers with multiple actual human
users, and we can’t just leak their credentials.)

Thanks for fixing. :-)

wchargin: > It won't be interpreted as a docker option because it's coming after
> the container name.

Good to know; thanks.

scripts/docker-entrypoint.sh
vsoch: exec doesn't work, so I'll be happy to add it, but ti doesn't allow control C :(

I'm testing the changes you just pushed now!

vsoch:
I don't think we should enforce how the user wants to define volumes - we are providing the base image they can extend upon, either with a bootstrapped Dockerfile, a runtime argument, or docker-compose.

The new recipe works great! And I just tried a rebuild, it goes _much_ faster. Like holy crap, this is so much better (thank you!)

scripts/docker-entrypoint.sh
vsoch: SO MUCH FASTER! :rocket:  And no wacky error messages. Dinosaur approved!

Dockerfile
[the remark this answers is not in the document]
vsoch: okay you've fixed this, resolving conversation.

scripts/docker-entrypoint.sh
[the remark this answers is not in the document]
vsoch: okay trying this now.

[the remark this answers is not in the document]
vsoch: That does it! Since there are other commands that might be run, I'm going to remove load as well, and have the user provide it.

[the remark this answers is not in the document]
vsoch: I'll add instructions to the README about mounting the volume too.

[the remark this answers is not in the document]
vsoch: okey doke! Just added a complete example to mount the data (so it remains on the host) along with using a build argument, and I generalized the command so that the user doesn't have to load (but can do something else... something that I don't know how to do yet :P)

Beanow:
Interesting. I've seen a fair number of images that use a data volume to store for example sqlite databases just declare it as a volume, for example [portainer](https://github.com/portainer/portainer/blob/develop/build/linux/Dockerfile). But indeed others with pretty much the same usecase do not and leave it up to you, for example [grafana](https://github.com/grafana/grafana/blob/master/packaging/docker/Dockerfile). Although both recommend you use volumes in the documentation.

For people looking to use it as a base image, I guess it makes more sense to not use a volume declaration. Although a lot of the time you'll probably want to reuse the same image with different data directory volumes to work on multiple projects.

vsoch:
@Beanow this is why I added an example of binding a volume to the host, and I think (if a more production version is desired) a docker-compose could be used to achieve the same.

vsoch:
@Beanow and @wchargin we did a lot of good work this morning! Let me know what additional changes you'd like to discuss - I left open the discussions for which I wasn't absolutely sure we had resolution, and overall the changes we made look good :)

Beanow:
I did a little testing with volumes vs using the writable layer.
Using `docker run --rm ...` there's no real difference. The data is discarded after use.
Using it without `--rm` just leaves a container in `docker ps -a`, while having a volume it will also leave a volume in `docker volumes ls`.

Both keeps the data directory isolated between multiple calls unless explicitly restarted from an existing container. So I suppose there's no reason to add a `VOLUME`. To manage persistent data directories you'll want to explicitly declare them yourself anyway.

That being the case, I wonder if we need the `ARG SOURCECRED_DEFAULT_DIRECTORY`. Since `RUN yarn backend` doesn't use this. Passing `SOURCECRED_DIRECTORY` as an env would allow you to change it without rebuilding. And images using `FROM sourcecred` can override it as well.

vsoch:
If there could be a usecase of bootstrapping the container (and having the location used for something during build) I think it's worth keeping. Note that the SOURCECRED_DIRECTORY can still be passed as an environment variable, either way, without rebuilding.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM node:10
```
decentralion: is there a particular reason for node 10? i'd have a slight preference for 12, although either is fine. i believe node 12 does a better job sizing the heap by default, which is convenient since sourcecred can sometimes need a lot of heap.

README.md
```diff
@@ -61,6 +60,115 @@ yarn start
...
+set that as a `--build-arg`:
+
+```bash
+$ docker build --build-arg SOURCECRED_DEFAULT_DIRECTORY=/tmp/data -t sourcecred .
+```
+
+If you don't want to build, the container is also provided at [vanessa/sourcecred](https://hub.docker.com/r/vanessa/sourcecred)
+
+```bash
+$ docker pull vanessa/sourcecred
```
decentralion: I'm uncomfortable referencing a "third party" docker file in the README, especially since SourceCred changes pretty frequently, so unless we do a good job coordinating, the docker file would go out of date.

what's the workstream involved in keeping an official SourceCred docker file up to date? can we have a script like `scripts/update_dockerfile.sh` that we can invoke in [deploy.sh](https://github.com/sourcecred/sourcecred/blob/master/scripts/deploy.sh) and use to ensure that we update our dockerfile on every release?

Dockerfile
vsoch: I think that I tried a newer version of node, and it had trouble building, but I can't be 100% certain it was because of the version. I chose node 10 based on looking at @Beanow Drone.io deployment in sfosc/sourcecred, so he might have comment about the choice of version.

README.md
vsoch: I'd be happy to remove this (and just pushed a commit to do that)! The only thing that we'd have to worry about keeping up to date is an actual provided container, the typical use case without that would be for the user to clone and build, at which point they build the latest from the code. If you are interested in providing an actual image (you could grab sourcecred on Docker Hub) I'd be glad to do another PR with a build -> deploy script (Travis, Circle, now GitHub CI, etc, choose your preference).

Dockerfile
decentralion: I'd be surprised if it had trouble with a new version of node--we run CI on both node 10 and 12, and I develop on node 12.

README.md
decentralion: Yeah, I think maintaining an official docker container makes sense; especially if you'll provide the build/deply script. I just grabbed the `sourcecred` on Docker Hub. Let's go with CircleCI integration since we already use that for CI.

vsoch: It’s a plan! If it’s ok with you, I’ll do that in a separate PR to keep each well scoped and smaller in size.

Dockerfile
Beanow: I opted for 10 in the drone build for consistency. Node 12 is the current release and will ship new features. While node 10 is the active lts and only receives backports with additional scrutiny from the release group.

Both are within the test matrix of sourcecred so I opted for the more stable one. It should be no problem to make 12 work reliably though. As we only need to worry about testing the particular version we get with `FROM node:12` just before pushing a new release. Which is not the case for the drone build, because the node version and code we ship update independently.

vsoch: okay, if I'm hearing correctly, @decentralion has preference for 12 and @Beanow is okay and suggests testing - so I just tested building with 12 locally (works great!) and updated the PR. I'll put in a new PR later tomorrow (or worst case) this weekend with setup on CircleCI to deploy to sourcecred/sourcecred, and during that CI build @Beanow we will be sure to test before pushing a new release. Sound like a plan?

@Beanow I'm off to bed so if there are further tweaks needed, please go forth and do! Catch up tomorrow!

Beanow:
With the same reasoning as going to node:12, I think we can try out node:12-alpine so long as we don't run into issues that will save us some space.

Comparing the build so far:
```
sourcecred          latest              506f6de7b2cf        18 seconds ago      1.52GB
sourcecred          alpine              ea297c38688d        2 minutes ago       901MB
```

Going to run the test suites see if anything comes up.

Beanow:
The final issue I have is with the entrypoint. Still, shipping with webpack dev-server as the default seems like a security nightmare to me. This will be a lot better once we have the new CLI done, but for now I think we should have an alternative. Let me work on that a bit to propose something.

README.md
decentralion: sgtm

Beanow:
I've changed the entrypoint to be something similar to
https://github.com/klakegg/docker-hugo/blob/master/dist/debian/files/run.sh

This exposes all of the normal `bin/sourcecred.js` CLI, plus:
- `dev-preview <repo>` a shortcut for sourcecred load + yarn start
- `dev-server` alias for yarn start
- `build [...args]` alias for yarn build

This lets you do things like one-offs for development purposes.

```
$ docker run --rm -ti sourcecred --version
sourcecred v0.4.0

$ docker run --rm -tip 8080:8080 -e SOURCECRED_GITHUB_TOKEN=`cat .token` \
sourcecred dev-preview sfosc/wizard

  GO   load-sfosc/wizard
  GO   github/sfosc/wizard
 DONE  github/sfosc/wizard: 6s
  GO   compute-cred
 DONE  compute-cred: 17ms
 DONE  load-sfosc/wizard: 6s
...
ℹ ｢wds｣: Project is running at http://0.0.0.0:8080/webpack-dev-server/
...
PROJECT_IDS: ["sfosc/wizard"]
...
ℹ ｢wdm｣: Compiled successfully.
```

Store your sourcecred data in a named volume to do multiple operations.

```
$ docker run --rm -ti -v sourcecred_data:/data -e SOURCECRED_GITHUB_TOKEN=`cat .token` \
sourcecred load sfosc/sourcecred

  GO   load-sfosc/sourcecred
  GO   github/sfosc/sourcecred
 DONE  github/sfosc/sourcecred: 491ms
  GO   compute-cred
 DONE  compute-cred: 41ms
 DONE  load-sfosc/sourcecred: 562ms

$ docker run --rm -ti -v sourcecred_data:/data -e SOURCECRED_GITHUB_TOKEN=`cat .token` \
sourcecred load sfosc/wizard

  GO   load-sfosc/wizard
  GO   github/sfosc/wizard
 DONE  github/sfosc/wizard: 392ms
  GO   compute-cred
 DONE  compute-cred: 21ms
 DONE  load-sfosc/wizard: 439ms

$ docker run --rm -ti -v sourcecred_data:/data -p 8080:8080 sourcecred dev-server
...
ℹ ｢wds｣: Project is running at http://0.0.0.0:8080/webpack-dev-server/
...
PROJECT_IDS: ["sfosc/sourcecred","sfosc/wizard"]
...
ℹ ｢wdm｣: Compiled successfully.
```

And build for production with such a volume.

```
$ docker run --rm -ti -v sourcecred_data:/data sourcecred build --output-path /tmp/example
Removing contents of build directory: /tmp/example
...
PROJECT_IDS: ["sfosc/sourcecred","sfosc/wizard"]
...
```

It's not perfect, (help missing the extra commands for example may be confusing), but it's something I'd be a lot more comfortable release into the wild until we have a better CLI.

vsoch:
@Beanow did you update the instructions in the README to reflect your changes?

Beanow:
> @Beanow did you update the instructions in the README to reflect your changes?

I haven't. Was interested in your opinions on this approach first.

Beanow:
@decentralion question for you. Currently I didn't include the `cp -r "${sourcecred_data}" "${target}/api/v1/data"` stuff in the build command.
However as it stands, the build already requires your data folder to find the PROJECT_IDS. Should we just expand on that assumption and make it a webpack responsibility to copy the right data? We can add the same cp logic to the entrypoint script, but that's just one more place to maintain it.

Beanow:
Mmh, having said that. Both the `soucrecred update` idea and #1272 are trying to move away from webpack. So It might not be worth the trouble adding this now vs just copying the shell script equivalent.

Beanow:
Speaking of readme, since we're unlikely to have access to `library/sourcecred` which can be shortend as `sourcecred` anytime soon. Should we refer to `sourcecred/sourcecred` as the image name?

Beanow:
@vsoch I've taken the current state of this PR and created https://github.com/sourcecred/widgets/pull/15 to see how they would be used together for generating SVGs.

.gitignore
```diff
@@ -13,6 +13,8 @@
 /bin
 
 # misc
+data/
```
decentralion: does data/ need to be in the gitignore if it's only used within the docker image?
having `data/` as a default data directory would be reasonable (likely much more convenient than status quo) but i'd keep that change to a separate PR. also, once we launch the [sourcecred instance](https://github.com/sourcecred/sourcecred/issues/1232#issuecomment-519538494) concept, we should have a more principled way to do this.

also, can we take `.token` out of the gitignore? i think we decided to keep doing it in the env by default?

(i personally have `SOURCECRED_GITHUB_TOKEN` set by my `.zshrc`, would rather encourage people to take that approach. although i recognize that futzing with zshrc is more of a pain than just writing a file, so I see the appeal to the `.token` approach too.)

README.md
```diff
@@ -39,17 +39,16 @@ First, make sure that you have the following dependencies:
 [Node]: https://nodejs.org/en/
 [Yarn]: https://yarnpkg.com/lang/en/
 [GitHub API token]: https://github.com/settings/tokens
-
 [macos-gnu]: https://github.com/sourcecred/sourcecred/issues/698#issuecomment-417202213
 
-Then, run the following commands to clone and build SourceCred:
+You'll stil need to create a GitHub token to use as an environment variable
```
decentralion: Creating a GitHub token is already discussed in the bullet list above; is there a reason we are mentioning it again?

README.md
```diff
@@ -61,6 +60,108 @@ yarn start
...
+If you want to build and customize the `SOURCECRED_DIRECTORY`, you can
+set that as a `--build-arg`:
+
+```bash
+$ docker build --build-arg SOURCECRED_DEFAULT_DIRECTORY=/tmp/data -t sourcecred .
+```
+
+You will still need to export a GitHub token, and then provide it to the container
+when you run it. Notice that we are also binding port 8080 so we can
+view the web interface that will be opened up.  The only argument needed is a command to load the GitHub repository to generate the sourcecred for:
```
decentralion: Please wrap this text for consistency with the rest of the file

README.md
```diff
@@ -61,6 +60,108 @@ yarn start
...
+
+```bash
+REPOSITORY=sfosc/sfosc
```
decentralion: technically, the argument isn't a repository, but a project spec. right now, the project can specify a repository (as in `sfosc/sfosc`) or a whole organization (as in `@sourcecred`). would be nice to show that both are possible.

README.md
```diff
@@ -61,6 +60,108 @@ yarn start
...
+below, we have a folder "data" in the present working directory that we bind to
+"/data" in the container, the default `SOURCECRED_DIRECTORY`.
+
+
+```bash
+$ SOURCECRED_GITHUB_TOKEN="xxxxxxxxxxxxxxxxx" \
+    docker run -d --name sourcecred --rm --env SOURCECRED_GITHUB_TOKEN -p 8080:8080 -v $PWD/data:/data sourcecred load "${REPOSITORY}"
+```
+
+We are running in detached mode (-d) so it's easier to remove the container after. After running the command, you can inspect it's progress like this:
```
decentralion: (also wrap)

.gitignore
vsoch: Data is the suggested folder for the user to bind to the host, so even in the case that only 0.25 of users try it out, it's good to have there. The (internal container) folder can be very specific, but the user is most likely to want to bind somewhere obvious in the present working directory. The instructions say $PWD/data so it's logical to add this, it doesn't hurt but has potential to help.

README.md
vsoch: Redundancy is a good strategy for documentation. It's likely that a user sees one or the other, skips over things.

README.md
vsoch: You mean one sentence per line, with wrapping as necessary? I'll fix this now. @Beanow I'll also rename containers to sourcecred/sourcecred.

README.md
vsoch: I'll add this now, although likely @Beanow will change it with the updated use.

README.md
vsoch: Fixed! Pushing soon (testing new changes to entrypoint script)

Dockerfile
vsoch: We are good with the version so I'm going to resolve this particular conversation (it's getting a bit long on the threads!) Re-open if you see necessary.

README.md
vsoch: Great! If you want to create an issue an assign me, that works, otherwise I'll just remember and add to my TODO. Resolving this discussion.

README.md
decentralion: Heh, I should stop bothering people about this and just find an automated solution to impose some consistent style. Do whatever seems best :)

Dockerfile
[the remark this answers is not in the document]
vsoch: This is fixed, thank you @Beanow!

Beanow:
For build, it's no different than the existing `yarn build` most of it's options are webpack related and requires a lot of code digging to get any use out of. So I would only explain two important points.

1. It requires a prepopulated sourcecred data directory. So volume/bind mounts for that would be good to show.
2. The `--output-path <path>` option. Along with perhaps a second bind mount to store the output. Perhaps using `-v $PWD/docs:/output` with `--output-path /output` as a github pages /docs setup?

Given we need a prepopulated data directory, I think `load` is a logical one to show as a wildcard example.

Finally `--version` or `help <command>` are somewhat of a hello world. So these may be useful.

vsoch:
Sure! Can you add that?

Beanow:
On the train and away for the weekend so I think that would take me a few days possibly. For examples I did include some in https://github.com/sourcecred/sourcecred/pull/1288#issuecomment-521965997

And https://github.com/sourcecred/widgets/pull/15

Others are familiar with the rest of the CLI so I think the main suggestion I could give is to explain use of volumes / bind mounts well. I have used named volumes in my examples, maybe for consistency you want bind mounts.

vsoch:
oh noo! Okay I'll try doing it.

vsoch:
okay, done! https://github.com/sourcecred/sourcecred/pull/1288/commits/fd9b10450b5783bd531798f397d2d6b517fb102b your descriptions were very useful @Beanow I was able to get both use cases working, and view output.

Beanow:
I'll sugest a few more changes because dev-server also needs a populated data folder. dev-preview is there to skip that by combining load + dev-server.

README.md
```diff
@@ -61,6 +59,190 @@ yarn start
...
+ORGANIZATION=@sfosc
+$ SOURCECRED_GITHUB_TOKEN="xxxxxxxxxxxxxxxxx" \
+    docker run -d --name sourcecred --rm --env SOURCECRED_GITHUB_TOKEN -p 8080:8080 sourcecred/sourcecred dev-preview "${ORGANIZATION}"
+```
+
+If you want to bind the data folder to the host, you can do that too. 
+In the example below, we have a folder "data" in the present working directory that we bind to "/data" in the container, the default `SOURCECRED_DIRECTORY`.
+
+```bash
+$ SOURCECRED_GITHUB_TOKEN="xxxxxxxxxxxxxxxxx" \
```
Beanow: The load command will terminate when done. So I would suggest not using `-d` but `-ti` instead.

Beanow: Also does not require the published port.

README.md
```diff
@@ -61,6 +59,190 @@ yarn start
...
+  GO   load-sfosc/sfosc
+  GO   github/sfosc/sfosc
+ DONE  github/sfosc/sfosc: 25s
+  GO   compute-cred
+ DONE  compute-cred: 1s
+ DONE  load-sfosc/sfosc: 26s
+...
+```
+
+It will take about 30 seconds to do the initial build, and when the web server is running you'll see this at the end:
```
Beanow: The load command does not build or run a server. So here we can show the dev-server command.

```
docker run -d --name sourcecred --rm -p 8080:8080 -v $PWD/data:/data sourcecred/sourcecred dev-server
```

README.md
```diff
@@ -61,6 +59,190 @@ yarn start
...
+Cool!
+
+
+#### Development Server
+
+If you only want to start the development server, you can run the container like this:
+
+```bash
+$ SOURCECRED_GITHUB_TOKEN="xxxxxxxxxxxxxxxxx" \
+    docker run -d --name sourcecred --rm --env SOURCECRED_GITHUB_TOKEN -p 8080:8080 sourcecred/sourcecred dev-server
```
Beanow: As per previous comment, we _do need_ a data directory here, but _don't need_ a github token. (All data should already be loaded)

```
docker run -d --name sourcecred --rm -p 8080:8080 -v $PWD/data:/data sourcecred/sourcecred dev-server
```

README.md
vsoch: No that's a typo that should be dev-preview too (not load). Sorry about that!

README.md
vsoch: oup and you just beat me to my last comment :)

README.md
vsoch: This isn't correct - it spits out an error asking for the token.

Beanow: `dev-server` does? That shouldn't be the case.

README.md
Beanow: Ah, I thought load + dev-server is a more natural example of when you'd use volumes, dev-preview being a bit more self-contained. But I'm not to picky about it either way. Planned behaviour would be more like dev-preview anyway.

README.md
vsoch: Yeah you're right! I had forgotten to rebuild the container, derp. We're good :)

README.md
vsoch: okay I just edited it to be (I think) in vision of what you want. If I did it wrong, please go ahead and fix it (because I might not be sure what you have in mind). I did my best!

vsoch:
I guess we need 6 reviews? Holy cow that's a lot.

wchargin:
Don’t worry about the “six approvals” requirement. It’s just a hack that
I came up with to establish finer-grained permissions. We want to say,
“permit this collaborator to create branches and push to any branch that
isn’t `master`”, because committing to `master` requires a bit more
caution and thought. But GitHub doesn’t have such an option, so we
instead create a branch protection rule for `master` with an practically
unattainable threshold while allowing it to be overridden by repository
administrators only. :-)

I’ll let @decentralion approve and merge this one, as they have more
context here than I do.

decentralion:
Yup, as @wchargin said, the 6-approvals is just a hack so that we can give "make new branch" permissions easily and "merge to master" sparingly.

This looks good to me to merge -- thank you @vsoch for writing it and @beanow for the thorough review! Since I don't have much knowledge of docker myself I very much appreciate having a confident reviewer. :)

@vsoch I think it will be best to squash this into a single commit to keep the master commit log clean, although we can leave all the commits here in the pull for cred history. Can you update the pull request description to be a good combined message for the squashed commit, and then either @wchargin and I will merge it in.

vsoch:
I won't dare try to do this on the command line, but since you are squashing and merging (and there is a place you can copy paste when you do that) here is the text to use:


```
Adding docker container recipe and instructions in README for running sourcecred
```

It does seem a bit ironic to squash (and give it only one commit, shared between two people) for a software that is all about giving credit, no? But I understand the need, so please go ahead. Maybe there are other metrics taken into account like lines contributed? I guess I need to look more at how sourcecred is actually working :)

decentralion:
> It does seem a bit ironic to squash (and give it only one commit, shared between two people) for a software that is all about giving credit

Right now we don't depend on commits very much in cred attribution; it's more based on the pull request and review process--and the cred for this pull request will reflect the fact that you were the primary author, that @beanow reviewed extensively and added several commits, and that @wchargin and I both had a moderate amount of activity. So by squashing the commits together we aren't losing much from a cred attribution standpoint--but we do make the git history of the project a lot cleaner, which helps when doing `git bisect`, or grepping over the log, etc.

In the future, I think we'll build more coupling between the cred graph, the commits, and the file structure. But I think we'll continue to tie the commit more closely to the pull request than to the nominal author.

vsoch:
okay, sounds good to me then! I'm so looking forward to having this merged - because next we'll do automated builds, and after automated builds sourcecred (and the wizard) can be run in just a few lines pretty much anywhere (GitHub, CircleCI, Travis CI). If we make a GitHub CI / Action I'm fairly certain this will quickly take over the internet because it's just plain cool :)

So let's do that!

Beanow:
Should squashes end up undervalued somehow I think that would be a thing to address in Sourcecred, but all the buzz in this PR will definitely amount to some cred. I agree though 26 commits  for 8 changed files isn't what you want.

wchargin:
Yep, exactly. SourceCred will find the signal in a discussion with 126
comments (very much so). The individual commits can certainly be nice
for review, but “changing quotes”, “fixing usage example”, and the like
shouldn’t be in the permanent repository history for everyone who pulls.

vsoch:
Totally agree. It's a great feature that you can have best practices (clean, single commits to correspond to a feature) while still holding true to the goal of the software.

decentralion:
Tested it locally as a final sanity check and then merged it! Thanks @vsoch!

PR #1293, opened 2019-08-16
mirror: allow fetching all usernames

decentralion:
This is a minor change to the Discourse mirror so that it supports a
query to get all users from the server. It will be convenient for a
followon change which makes `update` search for every user's likes.

I also modified createGraph so that it uses the new method, which
results in code that is cleaner and slightly more efficient.

Test plan: Unit tests updated.

### DIFF
--- a/src/plugins/discourse/createGraph.js
+++ b/src/plugins/discourse/createGraph.js
@@ -130,17 +130,19 @@ export function createGraph(serverUrl: string, data: DiscourseData): Graph {
   const g = new Graph();
   const topicIdToTitle: Map<TopicId, string> = new Map();
 
+  for (const username of data.users()) {
+    g.addNode(userNode(serverUrl, username));
+  }
+
   for (const topic of data.topics()) {
     topicIdToTitle.set(topic.id, topic.title);
     g.addNode(topicNode(serverUrl, topic));
-    g.addNode(userNode(serverUrl, topic.authorUsername));
     g.addEdge(authorsTopicEdge(serverUrl, topic));
   }
 
   for (const post of data.posts()) {
     const topicTitle = topicIdToTitle.get(post.topicId) || "[unknown topic]";
     g.addNode(postNode(serverUrl, post, topicTitle));
-    g.addNode(userNode(serverUrl, post.authorUsername));
     g.addEdge(authorsPostEdge(serverUrl, post));
     g.addEdge(topicContainsPostEdge(serverUrl, post));
     let replyToPostIndex = post.replyToPostIndex;
--- a/src/plugins/discourse/createGraph.test.js
+++ b/src/plugins/discourse/createGraph.test.js
@@ -40,6 +40,16 @@ describe("plugins/discourse/createGraph", () => {
     posts(): $ReadOnlyArray<Post> {
       return this._posts;
     }
+    users(): $ReadOnlyArray<string> {
+      const users = new Set();
+      for (const {authorUsername} of this.posts()) {
+        users.add(authorUsername);
+      }
+      for (const {authorUsername} of this.topics()) {
+        users.add(authorUsername);
+      }
+      return Array.from(users);
+    }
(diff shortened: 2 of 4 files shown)


src/plugins/discourse/createGraph.js
```diff
@@ -0,0 +1,165 @@
...
+}
+export function postAddress(serverUrl: string, id: PostId): NodeAddressT {
+  return NodeAddress.append(postNodeType.prefix, serverUrl, String(id));
+}
+export function userAddress(serverUrl: string, username: string): NodeAddressT {
+  return NodeAddress.append(userNodeType.prefix, serverUrl, username);
+}
+
+export function userNode(serverUrl: string, username: string): Node {
+  const url = `${serverUrl}/u/${username}/`;
```
wchargin: Presumably the server URL is trusted input; are there any restrictions
on the usernames, or should this be validated/sanitized?

src/plugins/discourse/createGraph.js
```diff
@@ -0,0 +1,165 @@
...
+  return {address, description, timestampMs: topic.timestampMs};
+}
+
+export function postNode(
+  serverUrl: string,
+  post: Post,
+  topicTitle: string
+): Node {
+  const url = `${serverUrl}/t/${String(post.topicId)}/${String(post.id)}`;
+  const descr = `[post #${post.indexWithinTopic} on ${topicTitle}](${url})`;
```
wchargin: This topic title almost certainly needs to be sanitized, right? Should
also probably be `“quoted”` just for readability.

src/plugins/discourse/createGraph.test.js
```diff
@@ -0,0 +1,317 @@
...
+    findPostInTopic(topicId: TopicId, indexWithinTopic: number): ?PostId {
+      const post = this._posts.filter(
+        (p) => p.topicId === topicId && p.indexWithinTopic === indexWithinTopic
+      )[0];
+      return post ? post.id : null;
+    }
+  }
+
+  function example() {
+    const url = "https://url.com";
```
wchargin: nit: It’s conventional to use <https://example.com> as an example URL,
because it’s owned by ICANN and explicitly allowed to be used in such
cases. By contrast, <https://url.com> is owned by some company.

src/plugins/discourse/mirror.js
```diff
@@ -131,12 +138,14 @@ export class Mirror implements DiscourseData {
...
     const tables = [
+      "CREATE TABLE users (username TEXT PRIMARY KEY)",
       dedent`\
         CREATE TABLE topics (
             id INTEGER PRIMARY KEY,
             title TEXT NOT NULL,
             timestamp_ms INTEGER NOT NULL,
-            author_username TEXT NOT NULL
+            author_username TEXT NOT NULL,
+            FOREIGN KEY(author_username) REFERENCES users(username)
```
wchargin: Yes, good.

src/plugins/discourse/mirror.js
```diff
@@ -200,6 +210,13 @@ export class Mirror implements DiscourseData {
       }));
   }
 
+  users(): $ReadOnlyArray<string> {
+    return this._db
+      .prepare("SELECT (username) FROM users")
```
wchargin: nit: Parens are unnecessary.

src/plugins/discourse/mirror.js
```diff
@@ -200,6 +210,13 @@ export class Mirror implements DiscourseData {
...
+      .all()
+      .map(({username}) => username);
```
wchargin: This can just be `.pluck().all()`.

src/plugins/discourse/createGraph.js
decentralion: discourse usernames are only letters, underscores, and numbers. should be fine.

decentralion:
Re sanitzation: @wchargin you have a good point, but it's out of scope for this PR (only showed up due to diffbase issues). I made issue https://github.com/sourcecred/sourcecred/issues/1297 to track it.

PR #1298, opened 2019-08-18
discourse mirror updates likes

decentralion:
The Discourse mirror class now keeps an up-to-date record of all of the
likes within an instance. It does this by iterating over every user in
the history, and requesting their likes. If at any point we hit a like
we've already seen, we move on to the next user. In the future, we can
improve this so we only query users we haven't checked in a while, or
users who were recently active.

Test plan: Tests verify that we correctly store all the likes, including
after partial updates, and that we don't issue unnecessary queries.

### DIFF
--- a/src/plugins/discourse/mirror.js
+++ b/src/plugins/discourse/mirror.js
@@ -9,11 +9,12 @@ import {
   type PostId,
   type Topic,
   type Post,
+  type LikeAction,
 } from "./fetch";
 
 // The version should be bumped any time the database schema is changed,
 // so that the cache will be properly invalidated.
-const VERSION = "discourse_mirror_v2";
+const VERSION = "discourse_mirror_v3";
 
 /**
  * An interface for retrieving all of the Discourse data at once.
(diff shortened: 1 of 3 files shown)


src/plugins/discourse/mirror.js
```diff
@@ -343,5 +369,62 @@ export class Mirror implements DiscourseData {
...
+    // since our last scan. This would likely improve the performance of this
+    // section of the update significantly.
+    for (const user of this.users()) {
+      let offset = 0;
+      let upToDate = false;
+      while (!upToDate) {
+        const likeActions = await this._fetcher.likesByUser(user, offset);
+        possiblePageSize = Math.max(likeActions.length, possiblePageSize);
+        for (const {timestampMs, postId, username} of likeActions) {
+          const alreadySeenThisLike = db
```
decentralion: @wchargin: let me know if there's a way to condense these queries (e.g. to get a "return value" from the INSERT that tells us if the field was already present).

wchargin: There is, and it’s all in the types:
https://github.com/sourcecred/sourcecred/blob/bf5c3a953bfba08501a5e88864d780b4f07bf6c6/flow-typed/npm/better-sqlite3_vx.x.x.js#L87-L88

I’ll do you one better and turn 2*n* queries into 1: #1301, #1302.

PR #1305, opened 2019-08-18
Fix 404s in timeline and legacy-mode interlinks

### LINKED ISSUE #1304: 404s on timeline and legacy-mode interlinks
Repros in prod (`v0.4.0-b304a7ae51f4-20190818-1625-clean-production`):

  -   1. Visit <https://sourcecred.io/cred/timeline/@sourcecred/>.
      2. Click the “(legacy)” link.

  -   1. Visit <https://sourcecred.io/cred/prototype/@sourcecred/>.
      2. Click the “timeline mode” link.

Each of these should properly resolve a URL, but instead 404s.

This is due to use of raw `<a>` elements instead of `Link` components.
I’ll send a fix.

wchargin:
Summary:
All links in SourceCred must use the `Link` component, providing either
an external URL `href={…}` or an internal route `to={…}`. Any uses of a
raw `<a>` element for internal routes will incur 404s when the
application is hosted on a non-root path, as is currently the case on
the production website.

The change to `FileUploader` is not strictly necessary, as the link has
no styled text and uses a `data:` URL, but there’s no reason not to.

Fixes #1304.

Test Plan:
Build the static site:

```
scripts/build_static_site.sh --target cred --project sourcecred/example-github
```

Then run `python3 -m http.server` from the repository root directory—not
the `cred/` subdirectory—and navigate to the timeline cred view:
<http://localhost:8000/cred/timeline/sourcecred/example-github/>

Observe that the “(legacy)” link now has the correct styling and
correctly navigates to the legacy mode page when clicked: prior to this
change, it would navigate to a URL without the proper `/cred/` path
prefix, yielding a 404. On the legacy page, verify that the “timeline
mode” link has the same properties.

Then, visit <http://localhost:8000/cred/test/FileUploader/> and verify
that the inspection test still passes.

Added a regression test to catch further such errors. Note that
reverting the code changes in this commit causes the test to fail, and
that running it with `--verbose` prints the problematic files.

wchargin-branch: fix-bad-routing-404s

### DIFF
--- a/sharness/test_no_raw_anchor_elements.t
+++ b/sharness/test_no_raw_anchor_elements.t
@@ -1,0 +1,32 @@
+#!/bin/sh
+# shellcheck disable=SC2016
+#
+# Components must use the `Link` component from `webutil/Link.js` rather
+# than raw `<a>` elements. The `Link` component properly handles
+# client-side routing, as well as providing consistent styles.
+#
+# See <https://github.com/sourcecred/sourcecred/pull/1305> for an
+# example of how to fix this error.
+
+export GIT_CONFIG_NOSYSTEM=1
+export GIT_ATTR_NOSYSTEM=1
+
+# shellcheck disable=SC2034
+test_description='check that bare <a> elements are never directly used'
+
+# shellcheck disable=SC1091
+. ./sharness.sh
+
+# shellcheck disable=SC1004
+test_expect_success "application components must use <Link> instead of <a>" '
+    test_must_fail git grep -nF "</a>" \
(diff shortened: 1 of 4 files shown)


wchargin:
Added a regression test.

Using `</a>` isn’t exactly sufficient because of self-closing tags, but
`<a>` elements usually shouldn’t be self-closing anyway, so I expect
that this should be fine.

The exception on `webutil/Link.js` is actually superfluous—it doesn’t
literally use `</a>` anywhere—but I’ve kept it to show how to add more
exceptions should the need arise.

Issue #1308, opened 2019-08-21
Fix skew between dependency implementations and typings


wchargin:
Recent updates to dependencies (including additions of new dependencies)
have not always added the relevant typing definitions. This leads to
incorrect type errors, incorrect missing type errors, and unnecessary
`any`-proliferation.

The following type definitions are out of date:

  - [x] (#1311) flow-typed/npm/chalk_v1.x.x.js
  - [x] (#1314) flow-typed/npm/enzyme_v3.x.x.js
  - [x] (#1312) flow-typed/npm/express_v4.16.x.js
  - [x] (#1311) flow-typed/npm/flow-bin_v0.x.x.js
  - [x] (#1311) flow-typed/npm/isomorphic-fetch_v2.x.x.js
  - [x] (#1311) flow-typed/npm/jest_v24.x.x.js
  - [x] (#1311) flow-typed/npm/object-assign_v4.x.x.js
  - [x] (#1307) flow-typed/npm/prettier_v1.x.x.js

The following type definitions exist in the `flow-typed` repository, but
not in our project:

  - [x] (#1311) flow-typed/npm/chalk_v2.x.x.js
  - [x] (#1310) flow-typed/npm/deep-freeze_v0.0.1.js
  - [x] (#1311) flow-typed/npm/rimraf_v2.x.x.js
  - [x] **wontfix** ~~flow-typed/npm/webpack_v4.x.x.js~~

Known issues:

  - The `webpack` typings are simply incomplete, omitting valid and
    documented attributes like `compiler.hooks` and `compiler.options`,
    which we use. We’re going to decline to add these typings for now:
    <https://github.com/sourcecred/sourcecred/issues/1308#issuecomment-523975401>

  - The `enzyme` typings have been changed to improve preservation of
    type parameters, which came (perhaps unwittingly) at the cost of a
    seemingly valid pattern in our code. Merged a PR upstream to fix
    this: <https://github.com/flow-typed/flow-typed/pull/3522>

wchargin:
I’ve looked into the `webpack` typings a bit more. They don’t seem too
useful to us. They have five false-positive errors, which we can just
suppress. After introducing a bunch of intentional type errors to the
config, I was unable to get Flow to actually flag any of them as an
error. So I’ll suggest that we just treat those libdefs as if they don’t
exist.

wchargin:
I’ve resolved all the line items in this issue, so I’m going to close
it. When adding and updating dependencies in the future, let’s make sure
[...]

PR #1310, opened 2019-08-21
flow: add libdefs for `deep-freeze`

wchargin:
Summary:
This dependency was added in #1249 without typedefs, and so is
implicitly `any`-typed.

Depends on #1309 to fix a bug that would otherwise be a true positive
type error.

Addresses part of #1308.

Generated with `flow-typed install deep-freeze@0.0.1`.

Test Plan:
Running `yarn flow` passes, but fails if you remove the `nodePrefix` or
`edgePrefix` attributes of the Discourse plugin declaration.

wchargin-branch: libdefs-deep-freeze


decentralion:
Thank you for catching this @wchargin! I independently encountered the same issue in my local discourse-plugin development.

wchargin:
Sure thing.

PR #1320, opened 2019-08-23
Adding docker build (all but master) and deploy (master) to CircleCI

vsoch:
This pull request will use the [CircleCI Docker Orb](https://circleci.com/orbs/registry/orb/circleci/docker) to add two workflows:

 - one to build the container and preview the tag (all branches but master)
 - one to build and deploy the container on merge to master

Specifically, the testing should work okay as is, but for deploy an admin will need to add `DOCKER_LOGIN` and `DOCKER_PASSWORD` to the environment in CircleCI. I'm not sure if you can see this, but [here is](https://circleci.com/gh/good-labs/sourcecred/14) the test build and tag preview on my branch.

Both build the images "sourcecred/sourcecred:latest" - and derive the version from the container (and push both latest and the tagged version, only on merge to master).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,10 @@
 #
 version: 2.1
 
+orbs:
+  # https://circleci.com/orbs/registry/orb/circleci/docker
+  docker: circleci/docker@0.5.13
+
 executors:
   node10:
     docker:
@@ -51,22 +55,72 @@ jobs:
       - run: yarn test --full --ci
 
 workflows:
-  version: 2
+  version: 2.0
   commit:
     jobs:
       - test
+      - docker/publish:
+          deploy: false
+          image: sourcecred/sourcecred
+          tag: latest
+          requires:
+            - test
+          filters:
+            branches:
+              ignore: 
+                - master
+          after_build:
+            - run:
+                name: Preview Docker Tag for Build
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
+                   echo "Version that would be used for Docker tag is ${DOCKER_TAG}"
+
+      - docker/publish:
+          image: sourcecred/sourcecred
+          tag: latest
+          requires:
+            - test
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/
+          after_build:
+            - run:
+                name: Publish Docker Tag with Sourcecred Version
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
+                   echo "Version for Docker tag is ${DOCKER_TAG}"
+                   docker tag sourcecred/sourcecred:latest sourcecred/sourcecred:${DOCKER_TAG}
+                   docker push sourcecred/sourcecred
+
       - test_full:
           filters:
             branches:
               only:
                 - master
                 - /ci-.*/
+
       - test_full_10:
           filters:
             branches:
               only:
                 - master
                 - /ci-.*/
+
+      - docker/publish:
+          image: sourcecred/sourcecred
+          requires:
+            - test
+            - test_full
+            - test_full_10
+          tag: dev
+          filters:
+            branches:
+              only: master
+
+
   nightly:
     triggers:
       - schedule:


Beanow:
From what I understand this looks for the env vars to have permissions to actually publish, right?

parameter | description | required | default | type
-- | -- | -- | -- | --
docker-password | Name of environment variable storing your Docker password | - | DOCKER_PASSWORD | env_var_name
docker-username | Name of environment variable storing your Docker username | - | DOCKER_LOGIN | env_var_name

vsoch:
As long it's a contributor to the Docker repository (someone that can push) it will work.

.circleci/config.yml
```diff
@@ -51,7 +55,41 @@ jobs:
...
+          image: sourcecred/sourcecred
+          tag: latest
+          filters:
+            branches:
+             only: master
+          after_build:
+            - run:
+                name: Publish Docker Tag with Sourcecred Version
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
```
Beanow: Cool idea to use the just built container `--version` command. My inclination would be to use git tags for this though. It's pretty common for sourcecreds workflow to update master before creating releases.

Perhaps tag the master branch as `dev` and split a `v0.4.0` git tag into `latest`, `0`, `0.4` and `0.4.0`?
(The `0` one would not be too useful now, but will be after a 1.0.0 release)

.circleci/config.yml
```diff
@@ -51,7 +55,41 @@ jobs:
...
+          filters:
+            branches:
+              ignore: 
+                - master
+          after_build:
+            - run:
+                name: Preview Docker Tag for Build
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
+                   echo "Version for Docker tag is ${DOCKER_TAG}"
```
Beanow: Should we keep --version based tagging, I think the text echoed here should reflect it being a preview.
Suggesting `Version that would be used for Docker tag is ${DOCKER_TAG}`

.circleci/config.yml
vsoch: I'm not totally clear how you'd want this to work with CircleCI logic - I haven't done anything with tags [because it seems a little buggy](https://discuss.circleci.com/t/workflows-with-jobs-triggered-by-tag/18543/4) to not appear as a workflow. If you want to take a shot at editing the config, please go ahead! I wouldn't know the changes to make.

.circleci/config.yml
vsoch: See the title "Preview Docker Tag for Build"

vsoch: It's not obvious here, but it jumps out at you in the interface.

image

but if the echo is really bothering you please go ahead and change!

.circleci/config.yml
Beanow: The filter that I've used in past projects for tag only jobs:
```yml
filters:
  branches:
    ignore: /.*/
  tags:
    only: /^v.*/
```

It assumes you have a `v` prefix in the git tag for your semver releases.

vsoch: okay - I'll try this!

.circleci/config.yml
vsoch: Making this change, since I'm making the one above too.

.circleci/config.yml
decentralion: The workflow I take when publishing releases is:
- create a PR which updates the version within SourceCred, e.g. https://github.com/sourcecred/sourcecred/pull/1271
- merge that PR
- create a Release via the GitHub UI

creating the release is what sets the tag, I believe on the then-master commit. However, that commit will already have been through CI. so, do we have any guarantee that the publish workflow will actually trigger?

Beanow: That's right. Branches and tags are at a high level the same thing. A pointer to a commit ID.
Pushing a tag (just like a branch) to github triggers it's own webhook to CircleCI, that's what we're listening for here, so you're guaranteed this job is run as a result of pushing the tag. You don't have to coordinate it with pushing commits to a branch.

In fact, if you ever find you've messed it up. With a bit more work you can "update" tags by deleting them upstream `git push --delete origin tagname`, deleting it locally `git tag --delete tagname`, recreating the tag and pushing it. Even if the same commit is referenced, you can force trigger the webhook again. It's made more cumbersome because conceptually tags should be immutable and you probably want a minor semver update instead.

decentralion:
Cool, I'd like to defer to @Beanow for approval on this since I think he has more context; @Beanow if you approve it I'll merge it. :)

Beanow:
Some things left to resolve:

- We need to set `DOCKER_LOGIN` and `DOCKER_PASSWORD` in the CircleCI settings.
- I think we should have a `:dev` tag which publishes the current master.
- The parallelization issue may be applicable here as well https://github.com/sourcecred/widgets/pull/22#pullrequestreview-280398743.

Optional imo:
- Following the common tagging scheme of splitting a `v0.4.0` release into `latest`, `0`, `0.4`, `0.4.0` would be ideal. Although I feel like this is not urgent and will :+1: `v0.4.0` / `0.4.0` tags as well.

Beanow:
Workflow on https://github.com/sourcecred/widgets/pull/22 looks good, happy to set it up similarly here if you like.

Mentioned in https://github.com/sourcecred/widgets/pull/22#issuecomment-525494557 we probably should create a new user on docker hub to limit access of the (DOCKER_LOGIN and DOCKER_PASSWORD) credentials.

Beanow:
I am, but it's 0:50am, I will tap out for today as well.

Beanow:
Morning @vsoch :smile:

I believe https://github.com/sourcecred/sourcecred/commit/2c906190f708ee53b08466a8bf3ea26c5cbd8e4d has the same issue of running `test` multiple times from different workflows.
image


I don't know of any way to describe cross-workflow dependencies in Circle so the closest to the original test vs full test configuration I believe we can get is to use one workflow.

My suggestion would be having only the publish for `:dev` require `test`, `test_full` and `test_full_10` as this builds on master as well, like the full tests do. The others only requiring `test`.

Looking at the current behaviour, the 3 tests are parallel.
https://circleci.com/workflow-run/f8ca5d57-a10a-43ad-8506-79ddd391754c
image

.circleci/config.yml
```diff
@@ -51,22 +55,71 @@ jobs:
...
       - test_full_10:
           filters:
             branches:
               only:
                 - master
                 - /ci-.*/
+
+      - docker/publish:
+          image: sourcecred/sourcecred
+          requires:
```
Beanow: As `test` is run in parallel to the ones below, it should be included as well.

.circleci/config.yml
```diff
@@ -51,22 +55,71 @@ jobs:
...
+            tags:
+              only: /^v.*/
+          after_build:
+            - run:
+                name: Publish Docker Tag with Sourcecred Version
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2 | cut -c 2-) # without v
+                   echo "Version for Docker tag is ${DOCKER_TAG}"
+                   docker tag sourcecred/sourcecred:latest sourcecred/sourcecred:v${DOCKER_TAG}
+                   docker tag sourcecred/sourcecred:latest sourcecred/sourcecred:${DOCKER_TAG}
```
Beanow: Sorry I didn't explain clearly what I meant here. I think either with or without the `v` prefix is good. Let's pick one or the other.

Beanow:
I've added a deploy user on Docker hub and CircleCI which has permissions only for this repo.
The user I've verified by logging in with it locally and testing:

```
$ docker push sourcecred/sourcecred:dev
...
dev: digest: sha256:d2a7bedf3cc3c10896c816df928ce3b4e3202de087986d5c8a7a8eceaeb6b3a9 size: 3478

$ docker push sourcecred/widgets:dev
...
errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```

The security on this is still pretty spotty, for example it's also tied to an email address that can be used for password resetting. And these credentials not excluding account management operations, etc. There's not a lot more I think we can do until Docker Hub starts improving these things.

Beanow:
Looking over the final output of the build, I realized we're not leveraging `--cache-from`. Which I think should be an option for the publish job, so I've opened this https://github.com/CircleCI-Public/docker-orb/issues/15

But we can create a workaround I believe, pulling the images we want to use for caching in a new job and referencing them with the `extra_build_args` options.

I'd like to suggest addressing this in a new PR though as this is otherwise ready. Let's have it building now and see how it behaves.

Beanow:
Not required so I'm approving the current state, it's up to you whether you'd like to do include it now or separately.

Beanow:
Paging @decentralion, I believe we're ready to merge.
Widgets has served as a test for actually deploying builds, otherwise tested the dry-runs here.

Beanow:
It looks like this pushes images to the GitHub Package Registry `docker.pkg.github.com`. If moving to another registry is on the table there are more options and I believe we should include those as well.

PR #1335, opened 2019-08-29
Blacklist more problematic quasar interactions

wchargin:
Blacklist more problematic quasar interactions

Summary:
Context: <https://github.com/sourcecred/sourcecred/issues/1256#issuecomment-526252852>

Without also blacklisting the reaction, we hit an invariant violation in
the relational view (reactions are expected to have exactly one author).

Test Plan:
Running `node ./bin/sourcecred.js load quasarframework/quasar-cli` now
completes successfully (in about 2 minutes 40 seconds). It does emit a
warning:

```
Issue[MDU6SXNzdWUzNDg0NjUzNDg=].reactions: unexpected null value
```

…because one of the reactions was blacklisted. But the relational view
handles this correctly, it seems: timeline cred is still computed and
renders without obvious error.

wchargin-branch: blacklist-more-quasar


wchargin:
WANT_LGTM=any

PR #1337, opened 2019-08-30
mirror: guess typenames and warn on mismatch

### LINKED ISSUE #1336: Warn when we *suspect* bad typenames on GitHub objects
GitHub object IDs are opaque, and we must not rely on them to make any
decisions that influence our output. But in practice they have a very
consistent format, and we could rely on them to emit _warnings_ where we
suspect that the GitHub API is not fulfilling its contract. (See #996
for context.)

I sampled all objects from 270 repositories that I have loaded,
obtaining 139672 object IDs. Every one of these IDs was valid urlsafe
base64 that, when decoded as ASCII text, matched the Python regex
`^[0-9]*:([A-Za-z_-]*)`, where the capturing group contains the GitHub
type name.

```shell
for db in ~/tmp/sourcecred/cache/*.db; do
    sqlite3 "${db}" 'SELECT id FROM objects ORDER BY id'
done | sort | uniq >/tmp/ids
```

```python
>>> ids = open("/tmp/ids").read().splitlines()
>>> len(ids)
139672
>>> import base64
>>> dec = [base64.urlsafe_b64decode(x).decode("ascii") for x in ids]
>>> import re
>>> pat = re.compile('^[0-9]*:([A-Za-z_-]*)')
>>> tys = {x: pat.match(x).group(1) for x in dec}
>>> import collections
>>> cts = collections.Counter(tys.values())
>>> for ty in sorted(cts): print("%s %d" % (ty, cts[ty]))
... 
Bot 3
Commit 31614
Issue 6444
IssueComment 31412
Organization 6
PullRequest 10732
PullRequestReview 23818
PullRequestReviewComment 24957
Reaction 8252
Ref 267
Repository 270
User 1897
```

Cross-checking this count checksum against the actual recorded types
gives identical results:

```shell
for db in ~/tmp/sourcecred/cache/*.db; do
    sqlite3 "${db}" 'SELECT typename, id FROM objects'
done | sort | uniq |
awk -F '|' '{ c[$1]++ } END { for (x in c) print x, c[x] }' | sort
```

We can take advantage of this by adding a parameter to the `Mirror`
class:

```js
guessTypename?: (Schema.ObjectId) => Schema.Typename | null
```

When provided, this function will be invoked with the ID of each
`(object_id, typename)` pair that we receive from the server. If the
actual typename does not match the guessed typename, we emit a warning
to the console. In no case should the guessed typename actually
influence the results.

Thus, when we receive a result like the following:

```
$ ghquery '{ node(id: "MDg6UmVhY3Rpb24zNDUxNjA2MQ==") { ... on Reaction { user { __typename id } } } }'
{"data":{"node":{"user":{"__typename":"User","id":"MDEyOk9yZ2FuaXphdGlvbjQzMDkzODIw"}}}}
```

we can base64-decode `MDEyOk9yZ2FuaXphdGlvbjQzMDkzODIw` to get
`012:Organization43093820`, guess that the typename should be
`Organization`, and complain loudly that this doesn’t match the claimed
typename of `User`.

This is useful not only because we can fail fast, but also because it
enables us to detect the problem earlier in the pipeline. By failing
when we first see the bad ID/typename pair, we have the provenance
information handy, so we can print a message like “reaction `foo` has
user of claimed type `bar` but actual type `baz`”, rather than “this
node got into the database with a bad type somehow; dive into the tables
yourself to find out how”.

wchargin:
Summary:
The format of GitHub’s GraphQL object IDs is explicitly opaque, and so
we must not introspect them in any way that would influence our results.
But it seems reasonable to introspect these IDs solely for diagnostic
purposes, enabling us to proactively detect GitHub’s contract violations
while we still have useful information about the root cause.

This commit adds an optional `guessTypename` option to the Mirror
constructor, which accepts a function that attempts to guess an object’s
typename based on its ID. If the guess differs from what the server
claims, we continue on as before, but omit a console warning to help
diagnose the issue more quickly.

Resolves #1336. See that issue for details.

Test Plan:
Unit tests for `mirror.js` updated, retaining full coverage. To test
manually, revert #1335, then load `quasarframework/quasar-cli`. Note
that it emits the following warning before failing:

> Warning: when setting Reaction["MDg6UmVhY3Rpb24zNDUxNjA2MQ=="].user:
> object "MDEyOk9yZ2FuaXphdGlvbjQzMDkzODIw" looks like it should have
> type "Organization", but the server claims that it has type "User"

Unit tests for the GitHub typename guesser added as well.

Running `yarn test --full` passes.

wchargin-branch: mirror-guess-typenames

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -54,7 +54,10 @@ export default async function fetchGithubRepo(
   // equals signs in file names.
   const dbFilename = `mirror_${Buffer.from(resolvedId).toString("hex")}.db`;
   const db = new Database(path.join(cacheDirectory, dbFilename));
-  const mirror = new Mirror(db, schema(), {blacklistedIds: BLACKLISTED_IDS});
+  const mirror = new Mirror(db, schema(), {
+    blacklistedIds: BLACKLISTED_IDS,
+    guessTypename: _guessTypename,
+  });
   mirror.registerObject({typename: "Repository", id: resolvedId});
 
   // These are arbitrary tuning parameters.
@@ -74,6 +77,23 @@ export default async function fetchGithubRepo(
     connectionPageSize: 100,
   });
   return ((mirror.extract(resolvedId): any): Repository);
+}
+
+// GitHub object IDs are urlsafe-base64-encoded strings that decode to
+// ASCII strings of the form "123:Typename4567[...]", where the "123"
+// numbers are a function only of the typename and the "4567" numbers
+// are the object's database ID, and the "[...]" is either empty or a
+// further section like ":commithash" for commits.
+//
+// See tests for `_guessTypename` for some example object IDs.
+const GITHUB_ID_TYPENAME_PATTERN = /^[0-9]*:([A-Za-z0-9_-]*[A-Za-z_-])[0-9]+(?:[^A-Za-z0-9_-].*)?$/;
+
+export function _guessTypename(
+  objectId: Schema.ObjectId
+): Schema.Typename | null {
+  const decodedId = new Buffer(objectId, "base64").toString("utf-8");
+  const match = decodedId.match(GITHUB_ID_TYPENAME_PATTERN);
+  return match ? match[1] : null;
 }
 
 const GITHUB_GRAPHQL_SERVER = "https://api.github.com/graphql";
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -91,7 +91,7 @@ const GITHUB_ID_TYPENAME_PATTERN = /^[0-9]*:([A-Za-z0-9_-]*[A-Za-z_-])[0-9]+(?:[
 export function _guessTypename(
   objectId: Schema.ObjectId
 ): Schema.Typename | null {
-  const decodedId = new Buffer(objectId, "base64").toString("utf-8");
+  const decodedId = Buffer.from(objectId, "base64").toString("utf-8");
   const match = decodedId.match(GITHUB_ID_TYPENAME_PATTERN);
   return match ? match[1] : null;
 }
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -86,7 +86,7 @@ export default async function fetchGithubRepo(
 // further section like ":commithash" for commits.
 //
 // See tests for `_guessTypename` for some example object IDs.
-const GITHUB_ID_TYPENAME_PATTERN = /^[0-9]*:([A-Za-z0-9_-]*[A-Za-z_-])[0-9]+(?:[^A-Za-z0-9_-].*)?$/;
+const GITHUB_ID_TYPENAME_PATTERN = /^[0-9]*:([a-z0-9_-]*[a-z_-])[0-9]+(?:[^a-z0-9_-].*)?$/i;
 
 export function _guessTypename(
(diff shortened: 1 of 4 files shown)


src/plugins/github/fetchGithubRepo.js
```diff
@@ -76,6 +79,23 @@ export default async function fetchGithubRepo(
...
 }
 
+// GitHub object IDs are urlsafe-base64-encoded strings that decode to
+// ASCII strings of the form "123:Typename4567[...]", where the "123"
+// numbers are a function only of the typename and the "4567" numbers
+// are the object's database ID, and the "[...]" is either empty or a
+// further section like ":commithash" for commits.
+//
+// See tests for `_guessTypename` for some example object IDs.
+const GITHUB_ID_TYPENAME_PATTERN = /^[0-9]*:([A-Za-z0-9_-]*[A-Za-z_-])[0-9]+(?:[^A-Za-z0-9_-].*)?$/;
```
Beanow: Alt: can use `/i` regex flag to be case-insensitive and shorten the pattern. Since we're not using any uppercase or lowercase specific matching.

src/plugins/github/fetchGithubRepo.js
```diff
@@ -76,6 +79,23 @@ export default async function fetchGithubRepo(
...
+
+export function _guessTypename(
+  objectId: Schema.ObjectId
+): Schema.Typename | null {
+  const decodedId = new Buffer(objectId, "base64").toString("utf-8");
+  const match = decodedId.match(GITHUB_ID_TYPENAME_PATTERN);
```
Beanow: In theory the above binary to utf8 conversion would allow access to the full unicode range, including control characters. Do we know if [`String.prototype.match()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) is vulnerable to any [naughty strings](https://github.com/minimaxir/big-list-of-naughty-strings)?

Particularly during execution, as the regex used looks restrictive enough to avoid naughty strings in the output.

src/plugins/github/fetchGithubRepo.js
```diff
@@ -76,6 +79,23 @@ export default async function fetchGithubRepo(
...
+  return match ? match[1] : null;
```
Beanow: Should we check `match[1]` as an enum against expected types first? If one day this starts outputting `BigFurryMonster` as a type we know GitHub has possibly changed the opaque ID format or introduced a new type. Which just so happens to match our regex.

src/plugins/github/fetchGithubRepo.js
wchargin: It’d be quite surprising if the builtin `match` had inherent structural
vulnerabilities, as opposed to something like ReDoS (which this regex is
not vulnerable to). Even if there were a particular naughty string that
caused this regex to behave unexpectedly, it seems clear that the worst
that could happen would be that this function returns the wrong result.
Recall that we JSON-stringify guessed typenames before printing them to
the console, which neutralizes control characters.

src/plugins/github/fetchGithubRepo.js
wchargin: If GitHub starts sending a new type here, we’ll certainly want to know
about it, because it means that our local schema is wrong and we’ll do
the wrong thing with the resulting data. We could in principle check the
type against a set of _all_ GitHub’s declared types, but this is large—

```
$ ghquery '{ __schema { types { name } } }' |
> jq '.data.__schema.types[].name' | sort -u | wc -l
740
```

—and may change frequently, and I don’t see a strong reason to do so.

PR #1338, opened 2019-08-30
Adding small changes to use docker layers for cache

vsoch:
This pull request will pull sourcecred/sourcecred:latest first and then use the layers for cache as extra build args.  @Beanow figured out all the logistics for sourcecred/widgets, and as promised, I'm following up on the merge of #1320.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,11 @@ commands:
           paths:
             - node_modules
           key: v1-dependencies-{{ checksum "package.json" }}
+  pull_cache:
+    description: Pulls docker images usable for our cache
+    steps:
+      - run: docker pull sourcecred/sourcecred:dev
+      - run: docker pull node:12
 
 jobs:
   test:
@@ -69,6 +74,9 @@ workflows:
             branches:
               ignore: 
                 - master
+          extra_build_args: --cache-from=node:12,sourcecred/sourcecred:dev
+          before_build:
+            - pull_cache
           after_build:
             - run:
                 name: Preview Docker Tag for Build
@@ -86,6 +94,9 @@ workflows:
               ignore: /.*/
             tags:
               only: /^v.*/
+          extra_build_args: --cache-from=node:12,sourcecred/sourcecred:dev
+          before_build:
+            - pull_cache
           after_build:
             - run:
                 name: Publish Docker Tag with Sourcecred Version
@@ -111,6 +122,9 @@ workflows:
 
       - docker/publish:
           image: sourcecred/sourcecred
+          extra_build_args: --cache-from=node:12,sourcecred/sourcecred:dev
+          before_build:
+            - pull_cache
           requires:
             - test
             - test_full


.circleci/config.yml
```diff
@@ -69,6 +73,9 @@ workflows:
             branches:
               ignore: 
                 - master
+          extra_build_args: --cache-from=sourcecred/sourcecred:latest
```
Beanow: The current build https://circleci.com/gh/good-labs/sourcecred/50 is not using caching.
I believe this is because you need to include the base image that's in the Dockerfile `FROM`.
So `--cache-from=node:12,sourcecred/sourcecred:latest`.

Beanow: As well as pull this image. In the earlier commands.

vsoch: Gotcha! I wasn't sure if node would be included as a layer, since it's part of sourcecred/sourcecred. I'll add right away.

Beanow: Yeaah I think this is a pretty unintuitive aspect of `--cache-from`. My guess is this has to do with switching from implicit whatever we have locally to explicit cache _only_ from here. Which then fails on the very first layer because we don't have node:12 locally.

vsoch: It must use the entire image id. Anyway, green now! :green_apple: :green_book: :green_heart:

Beanow:
:thinking: still didn't cache. Is :latest that much out of date?

vsoch:
node:12 was updated a day ago - I can't tell how different it is though, at least from the Docker Hub interface. We'd have to have had a manifest with layers (from before) and now after.

Beanow:
> node:12 was updated a day ago

That would explain. The layer hash is used to see if it needs to be invalidated for caching. How big of a difference doesn't matter.

Still think it's interesting, should we use :dev instead of :latest?
Because :latest here should be actual releases. We'll likely lose out a lot of caching chances especially when package.json or yarn.lock are updated on master.

vsoch:
I think it we target a more specific version of node:12, we would be more likely to get a cache. That said, if we target a more specific version, we probably won't get the same security updates. So - it's a tradeoff.

Beanow:
Agreed, I rather build a little longer than miss out on security updates.

So thinking about `:dev` to cache from a little more, we have these 3 build filters:
- master
- tag releases
- other branches

master and tag releases should benefit from `:dev`

Other branches however runs most often and may sometimes not benefit.
Particularly when master has updated after the branch started. A rebase would fix that.
Or when the branch is not based off of master at all, for example a stable branch to maintain the previous major version. (Which we don't do)

So I think it would give overall better results.

vsoch:
okay so just to verify - you want to pull dev, to be added (used for cache) to build tag and master releases?

Beanow:
I think it should be all 3 on dev. So it can replace the :latest pull.

vsoch:
"latest" changed to "dev"

Beanow:
Updated my comment for the record to avoid confusion. I think the other branches benefits a lot from :dev.

decentralion:
merged from @beanow's approval. thanks! :)

PR #1341, opened 2019-09-01
mirror: initialize EAV primitives at registration

wchargin:
Summary:
This establishes the invariant that every object in the `objects` table
has all relevant rows in the `primitives` table, though those rows’
values are never yet set.

Test Plan:
Unit tests updated. Manually loading `sourcecred/example-github` and
running `.dump primitives` generates reasonable-looking output, with
lots of rows, including entries for nested fields and eggs. Verified
that the set of non-`id` columns on `Issue` equals the set of values for
the `fieldname` column of an `Issue` object, and likewise for `Commit`s,
thus covering each kind of field.

wchargin-branch: mirror-eav-init


wchargin:
Thanks for the reviews! If it’s not too much trouble, I’ll wait for
#1342 to be reviewed as well before merging these, so that we can get
all the breaking version changes in at once rather than forcing people
to clear caches multiple times.

decentralion:
Sounds good. I'll just keep working through them.

PR #1343, opened 2019-09-01
mirror: add EAV reading to `extract`, behind flag

wchargin:
Summary:
This completes the end-to-end EAV mode pipeline, but does not yet set it
as default or use it in production.

A note about indentation: we take care to avoid reindenting the entire
block of `extract` test cases, which is over 900 lines long. As to the
implementation code, reindenting the legacy type-specific primitives
branch is not easily avoidable, but when we remove that branch we won’t
have to reindent the EAV mode branch: we can replace its `if` block with
two scope blocks (which is the right thing to do, anyway).

Test Plan:
We reuse existing tests, which suffice for full coverage in both
implementation branches. Note that these tests cover the case of object
types with no primitive fields (the `Feline` and `Socket` types), which
are more likely to fail in a broken EAV implementation than in a broken
type-specific primitives implementation due to deletion anomalies.

To check that all relevant calls to `mirror.extract(…)` have been
properly replaced with `extract(mirror, …)`, run

    yarn coverage -f graphql/mirror -t 'EAV primitives'

and note that the “else” path of the `if (fullOptions.useEavPrimitives)`
branch is not taken; then, run

    yarn coverage -f graphql/mirror -t 'legacy type-specific primitives'

and note that the “if” path of the same branch is not taken.

To check that the table hiding logic is working, invert the branch that
checks `if (fullOptions.useEavPrimitives)`, and note that every test
case using the table hiding logic fails (except for some of the error
handling test cases, which do not actually need to read primitive data).

Finally, `yarn test --full` passes after flipping the `useEavPrimitives`
default to `true`.

wchargin-branch: mirror-eav-extract


wchargin:
Recommend reviewing with `-w`, due to the reindent of the legacy reading
logic discussed in the PR description.

wchargin:
Merged dependents and updated diffbase. Please take a look when you get
a chance.

wchargin:
@decentralion Friendly ping on this PR series?

decentralion:
@wchargin I'm intending to review it soon. If I don't review within one week, please consider me in "review default" and merge them sans review.

decentralion:
Is it intended that I review a stack of 6 commits here? or is the diffbase wrong?

wchargin:
The diffbase is correct, but this is just one commit’s worth of changes
(the commit is genuinely large in comparison to other commits in this
series, as noted in the comment on this PR—suggest ignoring whitespace).

Rendering it as “wants to merge 6 commits” is just what GitHub does when
you merge master into a PR branch.

wchargin:
Thanks! I actually think that the testing for this commit is pretty
cool, so you’re welcome to take a look if you’re interested.

Issue #1313, opened 2019-08-21
Switch to EAV tables in GraphQL mirror


wchargin:
Switch to EAV tables in GraphQL mirror

The GraphQL mirror currently creates one table per object type in the
GitHub schema, used to store primitive fields, with schemata like this
(formatted for readability):

```sql
/* 1 */ CREATE TABLE "primitives_Repository" (
/* 2 */     id TEXT NOT NULL PRIMARY KEY,
/* 3 */     "url",
/* 4 */     "name",
/* 5 */     "createdAt",
/* 6 */     FOREIGN KEY(id) REFERENCES objects(id)
/* 7 */ );
```

Lines 3–5 are dynamically generated, with one column per primitive field
on the GraphQL object type. The ID column and foreign key reference are
always present.

Because the columns are dynamically generated, most queries that alter,
read from, or write to these tables must be as well. But this violates
the First Rule of SQL: only create prepared statements whose bodies are
compile-time constant expressions. Dynamically generating SQL queries
opens us up to SQL injection, and while I’ve been quite careful in the
implementation and the data source is also trusted, it’s still a
dangerous property to have lurking around.

In retrospect, I think that the fact that these tables require
dynamically generated DDL and DML queries should have been a sign that
the table design itself leaves something to be desired. An obvious
alternative is to use a single table for storing all primitive values:
i.e., an [entity–attribute–value table][eav]. I considered this when
first designing the schema, but decided against it because I’d heard
vague cautions to avoid the EAV model when possible. After completing
the initial implementation, I don’t think that such criticisms apply
here, primarily because we’re writing a generic _family_ of schemata
(amusingly, mathematicians might call this a [_schema schema_][as])
rather than a schema for any fixed business application.

This refactoring should make the Mirror module code easier to understand
for newcomers. There will be a **constant** number of tables. There will
be **no** dynamically generated queries; all queries will be
compile-time constant expressions. There will be less side-condition
checking and validation logic, because less will be required. And the
code will be safer and more reusable.

[as]: https://en.wikipedia.org/wiki/Axiom_schema
[eav]: https://en.wikipedia.org/wiki/Entity%E2%80%93attribute%E2%80%93value_model#Data_structure

decentralion:
[...]

PR #1350, opened 2019-09-02
mirror: use fixed temp table for transitive deps

wchargin:
Summary:
The Mirror module extraction code calculates the set of transitive
dependencies and stores these results in a temporary table to avoid
unnecessary marshalling between JavaScript and C. We originally chose
the temporary table name dynamically, guaranteeing that it was unused.
However, this is unnecessary:

  - The temporary table namespace is unique to each database connection,
    so we need only consider possible conflicts in the same connection.
  - A `Mirror` instance exercises exclusive ownership of its database
    connection, per its constructor docs, so we need only consider
    conflicts within this module.
  - Temporary tables are only used in the `extract` method, so we need
    only consider conflicts in this method.
  - The `extract` method makes no open calls nor recursive calls, and
    does not yield control back to the event loop, so only one stack
    frame can be in `extract` at any time.
  - The `extract` method itself only creates the temporary table once.

Thus, the temporary table creation is safe. Furthermore, the failure
mode is simply that we raise an exception and fail cleanly; there is no
risk of data loss or corruption.

This patch replaces the dynamically generated table name with a fixed
name. On top of the work in #1313, this removes the last instance of SQL
queries that are not compile-time constant expressions.

Test Plan:
Running `yarn unit -f graphql/mirror` suffices.

wchargin-branch: mirror-fixed-temp-table


wchargin:
Good catch; thanks! Will fix before merging.

PR #1359, opened 2019-09-07
test: fix `example-github-load` snapshot test

wchargin:
Summary:
Generated with `./scripts/update_snapshots.sh` (with #1360 patched in).
This fixes failures introduced in #1358.

Test Plan:
Running `yarn test --full` now passes. Inspecting the diff (after piping
the old and new snapshots to `jq -S .`) shows that this includes only
additions, which seems appropriate given the precipitating change.

wchargin-branch: fix-1358-failures


wchargin:
That’s because you were hitting the converse side of #1360.

wchargin:
I generated this with `./scripts/update_snapshots.sh`; not sure that didn’t work for you.

PR #1374, opened 2019-09-11
Add `sourcecred discourse` command

decentralion:
This adds a new command, `discourse`, which makes it convenient to load
Discourse servers as standalone SourceCred projects.

For example, you could load the official SourceCred discourse via the
following:

```sh
export SOURCECRED_DISCOURSE_KEY=....
yarn backend
node bin/sourcecred.js discourse https://discourse.sourcecred.io credbot
yarn start
```

I've updated the README with instructions for using the plugin.

Test plan: No automated testing because I see this tool as a temporary
placeholder until we get the SourceCred instances setup. I manually
tested the error cases (e.g. providing an invalid server url) as well as
success cases like the one above.

### DIFF
--- a/src/cli/discourse.js
+++ b/src/cli/discourse.js
@@ -1,0 +1,140 @@
+// @flow
+// Implementation of `sourcecred discourse`
+// This is a (likely temporary command) to facilitate loading a single
+// discourse server.
+
+import dedent from "../util/dedent";
+import {LoggingTaskReporter} from "../util/taskReporter";
+import type {Command} from "./command";
+import * as Common from "./common";
+import {defaultWeights} from "../analysis/weights";
+import {load} from "../api/load";
+import {declaration as discourseDeclaration} from "../plugins/discourse/declaration";
+import {type Project} from "../core/project";
+
+function usage(print: (string) => void): void {
+  print(
+    dedent`\
+    usage: sourcecred discourse DISCOURSE_URL DISCOURSE_USERNAME
+                                [--weights WEIGHTS_FILE]
+           sourcecred discourse --help
+
+    Loads a target Discourse server, generating cred scores for it.
+
+    Arguments:
+        DISCOURSE_URL
+            The url to the Discourse server in question, for example
+            https://discourse.sourcecred.io
+
+        DISCOURSE_USERNAME
+            A user account on the Discourse server, to be used as the
+            "perspective" for the Discourse API calls. This user should not be
+            a privileged or admin user, otherwise hidden or deleted topics may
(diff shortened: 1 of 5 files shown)


src/cli/help.js
```diff
@@ -39,6 +41,8 @@ function usage(print: (string) => void): void {
     Commands:
       load          load repository data into SourceCred
       clear         clear SoucrceCred data
+      scores        print SourceCred scores to stdout
```
decentralion: (added scores as it was missing)

src/cli/discourse.js
```diff
@@ -0,0 +1,113 @@
...
+    die(std, "expected server url to start with 'https://' or 'http://'");
+  }
+
+  const project: Project = {
+    id: projectId,
+    repoIds: [],
+    discourseServer: {serverUrl, apiUsername},
+  };
+  const taskReporter = new LoggingTaskReporter();
+  const params = {alpha: 0.05, intervalDecay: 0.5, weights: defaultWeights()};
```
Beanow: Looks like `--weights` support is not yet implemented here and we have duplicated hard-coded values for `alpha` and `intervalDecay`.

Suggest extracting the weight file loading function as a common CLI function to add it here.
And moving the `alpha` and `intervalDecay` values to a constants file or functions, like `defaultWeights()`.

wchargin:
Why isn’t this done as part of `sourcecred load`?

decentralion:
> Why isn’t this done as part of `sourcecred load`?

I considered refactoring `sourcecred load` so that it takes an arg structure like:

- first positional arg is the project id
- `--github foo/bar` loads a github repo
- `--github @foo` loads an org
- `--discourse-url https://... --discourse-user credbot` loads a Discourse instance

This would have an advantage over this `sourcered discourse` command in that it enables mixing discourse and GitHub projects. However, it is still a short-term temporary solution (I want to migrate to a config-based approach rather than flags) and it breaks existing users. I figured there was no point breaking existing users just to put in a new temporary solution; I'd rather do a single break once we roll out the new instance system. At the same time, the general-purpose project config approach that I'm going to use for SourceCred's own needs (mixed Discourse+GitHub scope, along with node collapsing) will have a very unpolished UI initially, and I want to make it possible for users to try SourceCred on Discourse easily, while I think the GitHub/Discourse hybrid will be a power user (i.e. me) feature for a while.

Therefore, I felt that adding a new command that makes it easy to try out the Discourse plugin, without breaking any existing users, would be the best approach.\

wchargin:
SGTM

PR #1376, opened 2019-09-11
Share default TimelineCredParameters

decentralion:
At present, every place in the codebase that needs
TimelineCredParameters constructs them ad-hoc, meaning we don't have any
shared defaults across different consumers. This commit adds a
defaultParams method to `timeline/params.js`, which optionally allows
setting a weight override. It also re-factors every user of the params
to use this method instead.

I considered re-exporting defaultParams from the timelineCred module
(like I did in #1375 with TimelineCredParameters), but it was a bit
messy, and I decided to just re-route files to depend directly on the
params module.

Test plan: A simple unit test was added to ensure that weights overrides
work as intended. `git grep "intervalDecay: "` reveals that there are no
other explicit parameter constructions in the codebase.

Thanks to @Beanow; this is based on his [review feedback](https://github.com/sourcecred/sourcecred/pull/1374#discussion_r323149740).


src/analysis/timeline/params.js
```diff
@@ -56,3 +57,10 @@ export function paramsFromJSON(
     weights: weightsFromJSON(p.weights),
   };
 }
+
+export function defaultParams(weights?: Weights): TimelineCredParameters {
```
Beanow: While I see your point about only weights being used for overrides. I feel like this approach (selective overrides, and a default parameters function) is somewhat unconventional. Most commonly I see optional arguments to a constructor or factory, which are then merged with the default values.

In some cases the defaults are also exposed, for example in the [yaml package](https://www.npmjs.com/package/yaml), because you may want to reuse your options object for multiple function calls. Though in this case, the `TimelineCred` class already acts as the way to store and reuse your options.

The bottom line is, I wouldn't make it the responsibility of the caller to know about `defaultParams` when constructing. Instead make it the constructor / factory's concern to fill in any parameters not provided. And not to be selective based on what's already used in the current code-base. As this convention will make exposing features in a library-style easier later on.

wchargin:
> This commit adds a new type, `PartialTimelineCredParameters`, which
> is basically `TimelineCredParameters` with every field marked optional.

No need; use built-in `$Shape<…>`: see #1379.

PR #1380, opened 2019-09-16
add `contract` for collapsing nodes together

decentralion:
This commit adds Graph.contract, which allows collapsing certain
nodes in the graph into each other. This will enable the creation of a
SourceCred "identity" plugin, allowing identity resolution between users
different accounts on different services.

Test plan: Thorough unit tests have been added. `yarn test` passes.

Thanks to @wchargin for [review feedback][1] which significantly
improved this API.

[1]: https://github.com/sourcecred/sourcecred/pull/1380#discussion_r324958055

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -209,6 +209,15 @@ type CachedOrder = {|
   +modificationCount: number,
 |};
 
+/**
+ * Specifies how to contract a graph, collapsing several old nodes
+ * into a single new node, and re-writing edges for consistency.
+ */
+export type NodeContraction = {|
+  +old: $ReadOnlyArray<NodeAddressT>,
+  +replacement: Node,
+|};
+
 export class Graph {
   _nodes: Map<NodeAddressT, Node>;
   _edges: Map<EdgeAddressT, Edge>;
@@ -887,6 +896,63 @@ export class Graph {
       }
     }
     return result;
+  }
+
+  /**
+   * Create a new graph, in which some nodes have been contracted together.
+   *
+   * contractNodes takes a list of NodeContractions, each of which specifies a
+   * replacement node, and a list of old node addresses to map onto the new
+   * node. A new graph will be returned where the new node is added, none of
+   * the old nodes are present, and every edge incident to one of the old nodes
+   * has been re-written so that it is incident to the new node instead.
+   *
+   * If the same node addresses is "old" for several contractions, all incident
+   * edges will be re-written to connect to whichever contraction came last.
+   *
+   * If the replacement node is present in the graph, no error will be thrown,
+   * provided that the replacement node is consistent with the one in the graph.
+   *
+   * If there is a "chain" of remaps (i.e. a->b and b->c), then an error will
+   * be thrown, as support for chaining has not yet been implemented.
+   *
+   * The original Graph is not mutated.
+   *
+   * contractNodes runs in O(n+e+k), where `n` is the number of nodes, `e` is the
+   * number of edges, and `k` is the number of contractions. If needed, we can
+   * improve the peformance by mutating the original graph instead of creating
+   * a new one.
+   */
+  contractNodes(contractions: $ReadOnlyArray<NodeContraction>): Graph {
+    const remap = new Map();
+    const replacements = new Set();
+    const contracted = new Graph();
+    for (const {old, replacement} of contractions) {
+      for (const addr of old) {
+        if (replacements.has(addr)) {
+          throw new Error(
+            `Chained contractions are not supported: ${NodeAddress.toString(
+              addr
+            )}`
+          );
+        }
+        remap.set(addr, replacement.address);
+      }
+      replacements.add(replacement.address);
+      contracted.addNode(replacement);
+    }
+    for (const node of this.nodes()) {
+      if (!remap.has(node.address)) {
+        contracted.addNode(node);
+      }
+    }
+    for (const edge of this.edges({showDangling: true})) {
+      const src = NullUtil.orElse(remap.get(edge.src), edge.src);
+      const dst = NullUtil.orElse(remap.get(edge.dst), edge.dst);
+      const newEdge = {...edge, src, dst};
+      contracted.addEdge(newEdge);
+    }
+    return contracted;
   }
 
   checkInvariants() {
--- a/src/core/graph.test.js
+++ b/src/core/graph.test.js
@@ -1358,6 +1358,164 @@ describe("core/graph", () => {
     });
   });
 
+  describe("contractNodes", () => {
+    const a = node("a");
+    const b = node("b");
+    const c = node("c");
+    it("has no effect with no contractions", () => {
+      const g = simpleGraph();
+      const g_ = simpleGraph().contractNodes([]);
+      expect(g.equals(g_)).toBe(true);
+    });
+    it("adds the new node to the graph", () => {
+      const g = simpleGraph().contractNodes([{old: [], replacement: c}]);
+      const g_ = simpleGraph().addNode(c);
+      expect(g.equals(g_)).toBe(true);
+    });
+    it("filters old nodes from the graph", () => {
+      const g = new Graph()
+        .addNode(a)
+        .addNode(b)
+        .contractNodes([{old: [a.address, b.address], replacement: c}]);
+      const expected = new Graph().addNode(c);
+      expect(g.equals(expected)).toBe(true);
+    });
+    it("re-writes regular edges", () => {
+      const g = new Graph()
+        .addNode(a)
+        .addNode(b)
+        .addEdge(edge("forward", a, b))
+        .addEdge(edge("backward", b, a))
+        .contractNodes([{old: [a.address], replacement: c}]);
+      const expected = new Graph()
+        .addNode(c)
+        .addNode(b)
+        .addEdge(edge("forward", c, b))
+        .addEdge(edge("backward", b, c));
+      expect(g.equals(expected)).toBe(true);
+    });
+    it("re-writes edges, including dangling or loop edges", () => {
+      const g = new Graph()
+        .addNode(a)
+        .addEdge(edge("loop", a, a))
+        .addEdge(edge("dangle1", a, b))
+        .addEdge(edge("dangle2", b, a))
+        .contractNodes([{old: [a.address], replacement: c}]);
+      const expected = new Graph()
+        .addNode(c)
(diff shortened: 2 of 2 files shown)


src/core/graph.js
```diff
@@ -889,6 +889,71 @@ export class Graph {
...
+   * For example, if a user is represented on both GitHub and Discourse, we can create
+   * a new node that represents that identity, and collapse their GitHub and Discourse user
+   * nodes into it. (If we chose, we could also collapse the GitHub node into the Discourse
+   * node or vice-versa.)
+   *
+   * As its argument, combineNodes takes a map which maps a "src" node address
+   * (the address to be collapsed into another node) and a "dst" node address
+   * (the node to collapse into). As currently implemented, if a node is ever a
+   * dst, it can never be a src, although the code could be re-written to
+   * support this. Also, all "dst" nodes much actually be present in the graph
```
Beanow: Typo: `much` > `must`

src/core/graph.js
```diff
@@ -889,6 +889,71 @@ export class Graph {
...
+   * instead of creating a new one.
+   */
+  combineNodes(combinations: Map<NodeAddressT, NodeAddressT>): Graph {
+    for (const [src, dst] of combinations.entries()) {
+      if (combinations.has(dst)) {
+        // If A->B and B->C, I call this a "chained replacement".
+        // Supporting these transformations is not hard, but I don't yet see
+        // any need for them, so I'm deferring the implementation complexity
+        // until we actually need it.
+        throw new Error("Chained replacements not yet supported.");
```
Beanow: If I read the remainder of the code correctly. The end result of a combination is the `src` node being removed and all edges being moved over to the `dst`. In which case a `A > B > C` chain should have an identical result to `A > C` + `B > C` shouldn't it?

If so, the error and comments could suggest this alternative.
Best including the names of the nodes too, like below error.

src/core/graph.js
```diff
@@ -889,6 +889,71 @@ export class Graph {
...
+   * support this. Also, all "dst" nodes much actually be present in the graph
+   * (although this constraint could also be relaxed).
+   *
+   * The resultant Graph is distinct from the input graph, which is never
+   * modified. combineNodes runs in O(n+e+k), where `n` is the number of nodes,
+   * `e` is the number of edges, and `k` is the number of combinations. If
+   * needed, we can improve the peformance by mutating the original graph
+   * instead of creating a new one.
+   */
+  combineNodes(combinations: Map<NodeAddressT, NodeAddressT>): Graph {
```
wchargin: I’d prefer to call this `contractNodes` so that the meaning is
immediately obvious. If you object for some reason, the doc comment
should at least mention contraction for greppability.

src/core/graph.js
```diff
@@ -889,6 +889,71 @@ export class Graph {
...
 
+  /**
+   * Create a new graph, in which some nodes have been combined.
+   *
+   * combineNodes allows collapsing nodes together. For example, if we had a graph
+   * with nodes A, B and C, then we could collapse A into B, creating a new
+   * graph with just nodes B and C. The edges in the graph are re-written in concert;
+   * for example, an edge from A to B would become a loop edge from B to itself.
+   *
+   * combineNodes' motivating use cases is collapsing different user identities together.
```
wchargin: nit: s/cases/case/

src/core/graph.js
```diff
@@ -889,6 +889,71 @@ export class Graph {
...
+   * For example, if a user is represented on both GitHub and Discourse, we can create
+   * a new node that represents that identity, and collapse their GitHub and Discourse user
+   * nodes into it. (If we chose, we could also collapse the GitHub node into the Discourse
```
wchargin: The form that always contracts into a new node rather than choosing one
of the nodes arbitrarily is certainly nicer in some ways. For example,
suppose that the GitHub plugin expects its authorship edges to always be
connected to users. If we contract arbitrarily, then there’s a “50/50
chance” that the GitHub plugin will break, and so an error might slink
by; if we always create a new node, we’ll be sure to discover the
problem immediately.

Was there any consideration for the signature

```js
type Contraction = {|
  +old: $ReadOnlyArray<NodeAddressT>,
  +new: NodeAddressT,
|};

contractNodes(contractions: $ReadOnlyArray<Contraction>): Graph {
  // …
}
```

where `new` fresh?

src/core/graph.js
decentralion: Good point, but obviated by @wchargin's suggestion.

decentralion:
I've refactored the code to take @wchargin's suggestion into account. The implementation is now simpler, and we no longer have unhandled edge cases. Goes to show how much it can help to have a better interface!

src/core/graph.js
```diff
@@ -889,6 +898,55 @@ export class Graph {
...
+    }
+    for (const node of this.nodes()) {
+      if (!remap.has(node.address)) {
+        contracted.addNode(node);
+      }
+    }
+    for (const edge of this.edges({showDangling: true})) {
+      let {src, dst} = edge;
+      const srcRemap = remap.get(src);
+      if (srcRemap != null) {
```
wchargin:
```js
    for (const edge of this.edges({showDangling: true}) {
      const src = NullUtil.orElse(remap.get(src), edge.src);
      const dst = NullUtil.orElse(remap.get(dst), edge.dst);
      // …
    }
```

src/core/graph.js
```diff
@@ -889,6 +898,55 @@ export class Graph {
...
+   * edges will be re-written to connect to whichever contraction came last.
+
+   * The original Graph is not mutated.
+   *
+   * contract runs in O(n+e+k), where `n` is the number of nodes, `e` is the
+   * number of edges, and `k` is the number of contractions. If needed, we can
+   * improve the peformance by mutating the original graph instead of creating
+   * a new one.
+   */
+  contract(contractions: $ReadOnlyArray<GraphContraction>): Graph {
```
wchargin: As written, it looks like `new` nodes are perimtted to already exist in
the graph as long as the data is consistent; is that intentional? (It’s
not documented.)

wchargin: This doesn’t handle chained contractions, right? It looks like both

```js
graph.contract([
  {old: [a.address], new: b},
  {old: [b.address], new: c},
]);
```

and

```js
graph.contract([
  {old: [b.address], new: c},
  {old: [a.address], new: b},
]);
```

will turn `a` into two nodes `b` and `c`, with incident edges remapped
to `b`, and `c` an isolated vertex?

The semantics that you want, I think, are that you imagine applying the
contractions to the graph one at a time, but when you remap nodes and
edges in the graph you also remap later entries in the contractions
list. This conservatively extends the existing “take-last” semantics to
also handle chains.

src/core/graph.test.js
```diff
@@ -1358,6 +1358,58 @@ describe("core/graph", () => {
...
+      const expected = new Graph().addNode(c);
+      expect(g.equals(expected)).toBe(true);
+    });
+    it("re-writes edges, including dangling or loop edges", () => {
+      const g = new Graph()
+        .addNode(a)
+        .addEdge(edge("loop", a, a))
+        .addEdge(edge("dangle1", a, b))
+        .addEdge(edge("dangle2", b, a))
+        .contract([{old: [a.address], replacement: c}]);
```
wchargin: Maybe also test a normal (non-dangling, non-loop) edge, in- and out-?

src/core/graph.js
decentralion: Right. I originally had that construction, except for a little while I had `remap: Map<NodeAddressT, Node>` which necessitated a different pattern. But now I can go back to that.

src/core/graph.js
decentralion: Yes, I think it's fine. I've updated the documentation to reflect this, and added test cases.

src/core/graph.test.js
decentralion: Done.

src/core/graph.js
decentralion: I agree that the semantics you describe make sense. I spent at least 10 minutes thinking about how to get these semantics without either significantly complicating the implementation (keeping a backward map pointing every replacement to its origins), or making the runtime much worse `O(n+e+k) -> O(nk + ek)`, and didn't find any good solutions. Since we don't actually have any need for chained contractions, I don't think it's worth the effort to implement and test the more complicated behavior. Therefore, I instead implemented logic to detect any chained contractions and throw an error.

I left a (commented) block of code with a test for the chained contractions, so if we decide we need it later, we'll have a bit of a head start, test-driven-development style. If you see an elegant way to implement contraction chaining, feel free to do so in a followon PR. (I considered skipping the test instead of commenting it, which would decrease the probability of code rot, but I feel like there's some benefit to having 0 skipped tests--it means that if we start seeing test skips in jest, it indicates some new kind of technical debt that we're accumulating.)

wchargin: The approach that comes immediately to mind is to use a union-find
structure to accumulate the connected components, plus an extra forward
pass to resolve roots to their final node names. This should be nearly
linear time (up to α(·)), but I’ll grant that it’s not straightforward
given that there’s no union-find structure in the “standard library”.

Detecting and failing sounds good to me.

src/core/graph.js
```diff
@@ -889,6 +898,63 @@ export class Graph {
...
+   * be thrown, as support for chaining has not yet been implemented.
+   *
+   * The original Graph is not mutated.
+   *
+   * contract runs in O(n+e+k), where `n` is the number of nodes, `e` is the
+   * number of edges, and `k` is the number of contractions. If needed, we can
+   * improve the peformance by mutating the original graph instead of creating
+   * a new one.
+   */
+  contract(contractions: $ReadOnlyArray<GraphContraction>): Graph {
```
wchargin: nit: You might consider `contractNodes` and `NodeContraction`, as
without further qualification *contraction* generally refers to an
operation on a single edge, but it’s clear enough either way.

src/core/graph.js
```diff
@@ -209,6 +209,15 @@ type CachedOrder = {|
   +modificationCount: number,
 |};
 
+/**
+ * Specifies how to contract the graph, collapsing several old nodes
```
wchargin: s/the graph/a graph/: This is a type definition; there is no “`this`”
context.

src/core/graph.js
decentralion: Done.

src/core/graph.js
decentralion: Done.

PR #1382, opened 2019-09-17
GitHub plugin: Expose user addresses

decentralion:
Allow getting the node address for a user, given the user's login. This
will be needed by the upcoming identity plugin.

If the login in question corresponds to a bot, then a bot address will
be returned. When we make the bot-set configuration (rather than
hardcoded), we'll need to change the signature of this function; I think
that's fine.

Test plan: Unit tests added. (Also, it's really simple.)

### DIFF
--- a/src/plugins/github/nodes.js
+++ b/src/plugins/github/nodes.js
@@ -2,6 +2,7 @@
 
 import deepFreeze from "deep-freeze";
 import {NodeAddress, type NodeAddressT} from "../../core/graph";
+import {botSet} from "./bots";
 
 export opaque type RawAddress: NodeAddressT = NodeAddressT;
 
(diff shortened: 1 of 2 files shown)


src/plugins/github/nodes.js
```diff
@@ -36,6 +37,24 @@ export const Prefix = deepFreeze({
   pullComment: _githubAddress(COMMENT_TYPE, PULL_TYPE),
 });
 
+/**
+ * Return the address corresponding to a GitHub login.
```
wchargin: GitHub logins can also correspond to organizations (they can correspond
to any `Actor`). Presumably we don’t care about those because we don’t
have nodes for organizations, but that probably merits at least a
comment?

decentralion: Done.

PR #1384, opened 2019-09-18
Add the identity plugin

decentralion:
This commit adds the new SourceCred identity plugin. As described in the
README.md file:

This folder contains the Identity plugin. Unlike most other plugins, the
Identity plugin does not add any new contributions to the graph. Instead, it
allows collapsing different user accounts together into a shared 'identity'
node.

To see why this is valuable, imagine that a contributor has an account on both
GitHub and Discourse (potentially with a different username on each service).
We would like to combine these two identities together, so that we can
represent that user's combined cred properly. The Identity plugin enables this.

Specifically, the instance maintainer can provide a (locally unique) username
for the user, along with a list of aliases the user is known by, e.g.
`github/username` and `discourse/other_username`. The aliases are simple string
representations, that are intended to be easy to maintain by hand in a
configuration file. Then, the identity plugin will provide a list of
`NodeContraction`s that can be used by `Graph.contractNodes` to combine the
user identities as described.

The plugin is broken up into a few submoudles:
- `declaration.js` provides the PluginDeclaration. It has a single node
type (the identity node).
- `nodes.js` allows constructing identity nodes, and does some
validation on the username used to make the node.
- `alias.js` implements the logic for parsing aliases like
"github/decentralion" or "discourse/s_ben" into a node address.
- `nodeContractions.js` declares the `Identity` type (a username and a
list of aliases) and provides logic for turning a list of Identities
into a list of NodeContractions, suitable for use in
`Graph.contractNodes`.

The plugin is not yet integrated; that will come in a followon commit.

### DIFF
--- a/src/plugins/identity/alias.js
+++ b/src/plugins/identity/alias.js
@@ -1,0 +1,44 @@
+// @flow
+
+import {type NodeAddressT} from "../../core/graph";
+import {loginAddress as githubAddress} from "../github/nodes";
+import {userAddress as discourseAddress} from "../discourse/createGraph";
+
+/** An Alias is a string specification of an identity within another plugin.
+ *
+ * For now, the supported alias forms are `github/${githubUsername}` and
+ * `discourse/${discourseUsername}`. As a courtesty, if the user has put an @
+ * before the username, we strip it for them.
+ *
+ * This format has been chosen to be moderately extensible and easy to maintain
+ * by hand, as in the near future the alias files will be maintained by hand.
+ * This system will not scale well when user-provided plugins need to add to the
+ * aliasing scheme, so at that point we will rewrite this system.
+ */
+export type Alias = string;
+
+export function resolveAlias(
+  alias: Alias,
+  discourseUrl: string | null
+): NodeAddressT {
+  const re = /(\w+)\/@?(\w+)/g;
+  const match = re.exec(alias);
+  if (match == null) {
+    throw new Error(`Unable to parse alias: ${alias}`);
+  }
+  const prefix = match[1];
+  const name = match[2];
+  switch (prefix) {
+    case "github": {
+      return githubAddress(name);
+    }
+    case "discourse": {
+      if (discourseUrl == null) {
+        throw new Error(`Can't parse alias ${alias} without Discourse url`);
+      }
+      return discourseAddress(discourseUrl, name);
+    }
+    default:
+      throw new Error(`Unknown type for alias: ${alias}`);
+  }
+}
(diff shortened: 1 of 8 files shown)


src/plugins/identity/alias.js
```diff
@@ -0,0 +1,50 @@
...
+ * by hand, as in the near future the alias files will be maintained by hand.
+ * This system will not scale well when user-provided plugins need to add to the
+ * aliasing scheme, so at that point we will rewrite this system.
+ */
+export type Alias = string;
+
+export function resolveAlias(
+  alias: Alias,
+  discourseUrl: string | null
+): NodeAddressT {
```
Beanow: I believe the error handling may be significantly simplified with a regex.
For example `/^(github|discourse)/@?(\w+)$/`.
If explicit errors about unknown types are desired, removing the explicit options works too.
`/^(\w+)/@?(\w+)$/`

src/plugins/identity/alias.js
```diff
@@ -0,0 +1,50 @@
...
+ * This format has been chosen to be moderately extensible and easy to maintain
+ * by hand, as in the near future the alias files will be maintained by hand.
+ * This system will not scale well when user-provided plugins need to add to the
+ * aliasing scheme, so at that point we will rewrite this system.
+ */
+export type Alias = string;
+
+export function resolveAlias(
+  alias: Alias,
+  discourseUrl: string | null
```
Beanow: This API (along with the `githubAddress` and `discourseAddress` dependencies) isn't very extensible or neutral as a plugin. Instead, you might provide a map like `Map<string, function>` where string is your type/prefix, the function has a `UsernameT => NodeAddressT` signature.

Beanow: You might even consider turning this into a class or factory, to separate providing the functions from resolving an alias.

```
// factory style
const createAliasResolver = (mapping: Map<string, function>) => (alias: Alias) => { ... }

// OO style
class AliasResolver {
 ...
 constructor (mapping: Map<string, function>) { ... }
 resolve (alias: Alias) { ... }
}
```

decentralion: This is a very reasonable suggestion. However, I'm not going to do so in this PR. Right now, the Discourse and GitHub plugins are explicitly privileged all over the place (e.g. the project data format explicitly lists repo ids and a discourse server, the CLI explicitly looks for these keys, etc). I don't think there's much benefit to modularizing the plugin system piecemeal; instead, I'm planning to do a big refactor later, when we start to see external interest in adding plugins. At that point we can ensure the entire system has a clean and coherent design.

src/plugins/identity/alias.js
decentralion: Re-wrote it to use a regex per your suggestion.

PR #1402, opened 2019-10-08
Add post contents to the Discourse mirror

decentralion:
This modifies the Discourse fetcher and mirror so that we now keep post
contents around, thus enabling future reference detection (and other
things). The post contents are stored and provided as retrieved from the
API, which is in "cooked" HTML form.

Test plan: Unit tests and snapshots updated. Observe that the snapshots
now include Discourse post contents.

### DIFF
--- a/src/plugins/discourse/mirror.js
+++ b/src/plugins/discourse/mirror.js
@@ -15,7 +15,7 @@ import {
 
 // The version should be bumped any time the database schema is changed,
 // so that the cache will be properly invalidated.
-const VERSION = "discourse_mirror_v3";
+const VERSION = "discourse_mirror_v4";
 
 /**
  * An interface for retrieving all of the Discourse data at once.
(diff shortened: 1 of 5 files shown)


src/plugins/discourse/mirror.js
```diff
@@ -214,7 +215,8 @@ export class Mirror implements DiscourseData {
           author_username,
           topic_id,
           index_within_topic,
-          reply_to_post_index
+          reply_to_post_index,
+          cooked
         FROM posts`
       )
       .all()
```
Beanow: How does having `.all()` with the cooked HTML impact the memory footprint? Will all data be loaded into memory at once?

decentralion: Yeah, it loads all the posts into memory at once. When/if this becomes a memory issue we can address it by e.g. having an iterator over the posts so that they don't all need to be in memory at the same time. At the moment the full scraped db is only 1.9M so we're a long way from needing to worry about the memory footprint.

PR #1403, opened 2019-10-08
Add logic for plucking hyperlinks from cooked html

decentralion:
This commit adds a `parseLinks` method to a new module,
`plugins/discourse/references`. `parseLinks` allows us to extract the
hyperlinks from `<a>` tags in "cooked" html.

I added `htmlparser2` as a dependency to parse the html. There were a
lot of options to choose from; I chose htmlparser2 because it has a lot
of usage, reasonable performance, and suits our needs. We use this
dependency in a lightweight and local way, so we can always change it
later if needed.

One thing which was a bit odd: I wasn't able to import it using
`import`, and needed a `require` statement instead.

### DIFF
--- a/src/plugins/discourse/references.js
+++ b/src/plugins/discourse/references.js
@@ -1,0 +1,25 @@
+// @flow
+
+const htmlparser2 = require("htmlparser2");
+
+export type Hyperlink = string;
+
+export function parseLinks(cookedHtml: string): Hyperlink[] {
+  const links = [];
+  const httpRegex = /^https?:\/\//;
+  const parser = new htmlparser2.Parser({
+    onopentag(name, attribs) {
+      if (name === "a") {
+        const href = attribs.href;
+        if (href != null) {
+          if (href.match(httpRegex)) {
+            links.push(href);
+          }
+        }
+      }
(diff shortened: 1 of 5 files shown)


src/plugins/discourse/references.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+const htmlparser2 = require("htmlparser2");
```
Beanow: `htmlparser2` claims to be geared towards node. And needs browserify to work elsewhere.
Am I right in assuming web clients would never use this plugin directly because it's incorporated into the graph?

Beanow:
> One thing which was a bit odd: I wasn't able to import it using import, and needed a require statement instead.

From what I can see it transpiles from typescript into a `.js` without marking itself as `.mjs` or `type: module` in the package.json. So the output may be CommonJS?

src/plugins/discourse/references.js
decentralion: Yep, I don't expect web clients to depend on this directly. If we wind up needing it in the frontend, we can change deps (usage of htmlparser2 is very local)

PR #1404, opened 2019-10-08
discourse: factor out address module

decentralion:
This will make it possible to depend on addresses in the reference
module.


Beanow:
> the reference module.

You mean identity? Or am I missing a module?

decentralion:
I do mean the reference module--it gets added in the following PR. https://github.com/sourcecred/sourcecred/pull/1403

PR #1405, opened 2019-10-08
Parse Discourse references from hyperlinks

decentralion:
The `discourse/references` module now has a `linksToReferences` method
which extracts the parsed Discourse references from an array of
hyperlinks. The method is tested.

### DIFF
--- a/src/plugins/discourse/references.js
+++ b/src/plugins/discourse/references.js
@@ -2,9 +2,34 @@
 
 const htmlparser2 = require("htmlparser2");
 
-export type Hyperlink = string;
+import {type TopicId} from "./fetch";
 
-export function parseLinks(cookedHtml: string): Hyperlink[] {
+export type DiscoursePostReference = {|
+  +type: "POST",
+  +topicId: TopicId,
+  +postIndex: number,
+  +serverUrl: string,
+|};
+
+export type DiscourseTopicReference = {|
+  +type: "TOPIC",
+  +topicId: TopicId,
+  +serverUrl: string,
+|};
+
+export type DiscourseUserReference = {|
+  +type: "USER",
+  +username: string,
+  +serverUrl: string,
+|};
+
+export type DiscourseReference =
+  | DiscoursePostReference
+  | DiscourseTopicReference
+  | DiscourseUserReference;
+
+export type UrlString = string;
+export function parseLinks(cookedHtml: string): UrlString[] {
   const links = [];
   const httpRegex = /^https?:\/\//;
   const parser = new htmlparser2.Parser({
@@ -22,4 +47,39 @@ export function parseLinks(cookedHtml: string): Hyperlink[] {
   parser.write(cookedHtml);
   parser.end();
   return links;
+}
+
+export function linksToReferences(
+  links: $ReadOnlyArray<UrlString>
+): DiscourseReference[] {
+  const server = "(https://[\\w.-]+)";
+  const topic = `(?:${server})/t/[\\w-]+/(\\d+)`;
+  const post = `(?:${topic})/(\\d+)`;
+  const params = "(?:\\?[\\w-=]+)?";
+
+  const topicRegex = new RegExp(`^(?:${topic})(?:${params})/?$`);
+  const postRegex = new RegExp(`^(?:${post})(?:${params})/?$`);
+  const userRegex = new RegExp(`^(?:${server})/u/([\\w-]+)(?:${params})/?$`);
+  const references: DiscourseReference[] = [];
+  for (const link of links) {
+    let match = null;
+    const decoded = decodeURI(link);
+    if ((match = decoded.match(postRegex))) {
+      references.push({
+        type: "POST",
+        topicId: +match[2],
+        serverUrl: match[1],
+        postIndex: +match[3],
+      });
+    } else if ((match = decoded.match(topicRegex))) {
+      references.push({type: "TOPIC", topicId: +match[2], serverUrl: match[1]});
+    } else if ((match = decoded.match(userRegex))) {
+      references.push({
+        type: "USER",
+        username: match[2],
+        serverUrl: match[1],
+      });
+    }
+  }
+  return references;
 }
--- a/src/plugins/discourse/references.test.js
+++ b/src/plugins/discourse/references.test.js
@@ -1,6 +1,6 @@
 // @flow
 
-import {parseLinks} from "./references";
+import {parseLinks, linksToReferences} from "./references";
 
 describe("plugins/discourse/references", () => {
   describe("parseLinks", () => {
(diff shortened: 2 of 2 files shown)


src/plugins/discourse/references.js
```diff
@@ -23,3 +48,37 @@ export function parseLinks(cookedHtml: string): Hyperlink[] {
...
   return links;
 }
+
+export function linksToReferences(
+  links: $ReadOnlyArray<Hyperlink>
+): DiscourseReference[] {
+  const server = "(https://[\\w.-]+)";
+  const topic = `${server}/t/[\\w-]+/(\\d+)`;
+  const post = `${topic}/(\\d+)`;
+  const params = "(?:\\?[\\w-=]+)?";
```
Beanow: Perhaps this needs a snapshot test from a corpus of real-world links.
Especially interesting would be to know if this URL scheme has changed between discourse versions.

src/plugins/discourse/references.js
```diff
@@ -23,3 +48,37 @@ export function parseLinks(cookedHtml: string): Hyperlink[] {
   parser.end();
   return links;
 }
+
+export function linksToReferences(
+  links: $ReadOnlyArray<Hyperlink>
+): DiscourseReference[] {
+  const server = "(https://[\\w.-]+)";
```
Beanow: This character class may require preprocessing. For example with `punycode` to allow http://見.香港/. Also missing the optional `https?`.

src/plugins/discourse/references.js
Beanow: Params may need the `%` and `+` included for url encoding and the alias for spaces.

wchargin: Or just `decodeURI` on the whole URL first?

src/plugins/discourse/references.js
```diff
@@ -2,8 +2,33 @@
...
+  +username: string,
+  +serverUrl: string,
+|};
+
+export type DiscourseReference =
+  | DiscoursePostReference
+  | DiscourseTopicReference
+  | DiscourseUserReference;
 
+export type Hyperlink = string;
```
wchargin: This name suggests to me that the type represents the source for an HTML
or Markdown link, like `"<a>…</a>"` or `[…](…)`. But it seems from
context that these are actually just URL strings. What do you think of
using `UrlString` instead (to avoid confusion with builtin `URL`)?

src/plugins/discourse/references.js
Beanow: Generally agree with deep then wide. Though I would consider this an accessibility issue, and correctness of the proposed feature, more than preemptively adding features. http://xn--nw2a.xn--j6w193g/ isn't legible as far as I'm aware. Though I'm ok with either choice.

src/plugins/discourse/references.js
decentralion: Added a snapshot test with a few links from forums I know of. None of them are super weird... if you find weirder links somewhere, feel free to add them.

src/plugins/discourse/references.js
decentralion: Based on a bit of research, it looks like Discourse already doesn't include non-punycode characters in the topic slug anyway. Take a look at this for an example: https://forums.eveonline.com/t/topic/195153

Going to go with "deep before wide" and merge as-is.

src/plugins/discourse/references.js
decentralion: Sure, I added decodeURI. I don't know of any "real" examples (i.e. actual Discourse links) that need it, but can't hurt.

src/plugins/discourse/references.js
```diff
@@ -23,3 +48,37 @@ export function parseLinks(cookedHtml: string): Hyperlink[] {
...
+
+export function linksToReferences(
+  links: $ReadOnlyArray<Hyperlink>
+): DiscourseReference[] {
+  const server = "(https://[\\w.-]+)";
+  const topic = `${server}/t/[\\w-]+/(\\d+)`;
+  const post = `${topic}/(\\d+)`;
+  const params = "(?:\\?[\\w-=]+)?";
+
+  const topicRegex = new RegExp(`^${topic}${params}/?$`);
```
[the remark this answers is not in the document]
decentralion: Done.

src/plugins/discourse/references.js
```diff
@@ -23,3 +48,37 @@ export function parseLinks(cookedHtml: string): Hyperlink[] {
...
+  const postRegex = new RegExp(`^${post}${params}/?$`);
+  const userRegex = new RegExp(`^${server}/u/([\\w-]+)${params}/?$`);
+  const references: DiscourseReference[] = [];
+  let match = null;
```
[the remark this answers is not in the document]
decentralion: Done.

src/plugins/discourse/references.js
decentralion: Done.

src/plugins/discourse/references.js
wchargin: I don’t think that it’s an accessibility issue. Such strings are simply
not valid URIs: [RFC 3986] says that characters must be “US-ASCII” and
defines the set of valid characters within a path component as

```
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
```

and the [W3C URL spec] clearly specifies that each textual part of a URL
(e.g., the path, the query) is an ASCII string. Both of these standards
are referenced by the [HTML 5 spec].

Thus, if I write in a GitHub comment the Markdown

```markdown
[some link](http://見.香港/)
```

the rendered HTML that GitHub sends to the browser is actually

```html
<a href="http://%E8%A6%8B.%E9%A6%99%E6%B8%AF/" rel="nofollow">some link</a>
```

and, as @decentralion mentions, Discourse behaves similarly:

```html
<a href="http://xn--nw2a.xn--j6w193g/">some link</a>
```

This is because these servers are conforming to spec and sending valid
URIs. All that we need to do is properly parse percent signs (which we
might as well do). URIs are easy to parse _by design_; there is no need
to pull in Punycode conversions. To do otherwise wouldn’t be more
correct; it would be a non-standard extension.

There are lots of useful things to spend our limited energy on that will
probably have significantly higher impact. Let’s focus on those. :-)

[HTML 5 spec]: https://www.w3.org/TR/html50/infrastructure.html#valid-url
[RFC 3986]: https://tools.ietf.org/html/rfc3986#section-2
[W3C URL spec]: https://url.spec.whatwg.org/#concept-url

Beanow: Right, I was going to say we're not validating valid URIs but "cooked" html from a remote source. But if the HTML spec demands ASCII arguably it's the server doing it wrong if UTF stuff comes back from the server.

PR #1406, opened 2019-10-08
Add edge types for Discourse references

decentralion:
Test plan: It's just a declaration change. `yarn flow` passes.

### DIFF
--- a/src/plugins/discourse/declaration.js
+++ b/src/plugins/discourse/declaration.js
@@ -73,6 +73,30 @@ export const likesEdgeType: EdgeType = deepFreeze({
   description: "Connects a Discourse user to a post they liked.",
 });
 
+export const referencesPostEdgeType: EdgeType = deepFreeze({
+  forwardName: "references post",
+  backwardName: "post is referenced by",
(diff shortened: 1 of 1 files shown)


Beanow:
Can you touch on the though process behind the default weights?

decentralion:
Sure. 

I put the forward weight on the references (post, user) edges to 1/2 so that they would be half as strong as the edge from the containing post to its author. I figured it would be weird if a reference took "too much" cred from the post author, as it would disincentivize references. In the future, I want to do away with this calculus by making posts "hold" a certain amount of out weight for references, and if the post has no references it just goes to the seed vector, so that users don't get any cred benefit from eschewing references. I gave them a backwards 1/16 weight because I've observed that the algorithm is a little sensitive to "cred traps" and giving all edges slight backwards weights makes cred traps less likely. This is another thing I'd like to fix in a more principled fashion later, but giving a slight backwards weight is an easy hack.

I gave the mentions edges a weight of only 1/4 because:
- a mention adds less value than references, references produce a lot more context for readers (since they can follow the reference)
- they seem slightly easier to abuse than the references edges

PR #738, opened 2018-08-31
Fix crash on repos with underscores and dots

decentralion:
The GitHub regex in urlIdParse.js incorrectly disallowed repo names with
underscores and dots. Fixes #721.

Test plan:
Unit tests have been updated to include the failure case (they correctly
failed), and then code was updated so that the tests pass again.

PR #1412, opened 2019-10-17
Support dashes in alias usernames.

Beanow:
This should fix [this issue](https://discourse.sourcecred.io/t/credsperiment-week-3-distribution/282/2?u=beanow) where @s-ben's GitHub username isn't correctly collapsed into the identity.

Test plan: added a unit test with dashes and underscores, which failed before the fix.

### DIFF
--- a/src/plugins/identity/alias.js
+++ b/src/plugins/identity/alias.js
@@ -21,7 +21,7 @@ export function resolveAlias(
   alias: Alias,
   discourseUrl: string | null
 ): NodeAddressT {
-  const re = /(\w+)\/@?(\w+)/g;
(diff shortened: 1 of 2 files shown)


wchargin:
This is actually slightly over-permissive: GitHub very cleverly does not
allow underscores in usernames, because if they did then they wouldn’t
be able to assign `your_username.github.io` domains, because underscores
may not appear in DNS labels. Notably, the requirements for repositories
are more lenient, allowing underscores and periods both. Prior
discussion: #388, #738, et al. Fixed in #1414.

s-ben:
Thx for fixing!

PR #1413, opened 2019-10-18
identity: require aliases to be anchored

wchargin:
Summary:
All the documentation and tests seem to be assuming that aliases must be
anchored: `github/torvalds`, but not `some github/torvalds stuff`.
JavaScript regular expressions aren’t anchored by default; this commit
adds explicit anchoring and adds tests.

Test Plan:
Unit tests added.

wchargin-branch: alias-anchor

### DIFF
--- a/src/plugins/identity/alias.js
+++ b/src/plugins/identity/alias.js
@@ -17,26 +17,36 @@ import {userAddress as discourseAddress} from "../discourse/address";
  */
 export type Alias = string;
 
+const _VALID_ALIAS = /^(\w+)[/](.*)$/;
+const _VALID_GITHUB_NAME = /^@?([0-9a-z_-]+)$/i;
+const _VALID_DISCOURSE_NAME = /^@?([0-9a-z_-]+)$/i;
+
 export function resolveAlias(
   alias: Alias,
   discourseUrl: string | null
 ): NodeAddressT {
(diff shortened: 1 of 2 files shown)


src/plugins/identity/alias.js
```diff
@@ -17,25 +17,33 @@ import {userAddress as discourseAddress} from "../discourse/address";
  */
 export type Alias = string;
 
+const _VALID_ALIAS = /^(\w+)\/@?(.*)$/;
```
Beanow: Although this will be covered by the regex for each username. Why not use `(.+)` for the username part?

wchargin: Type-specific parsers should handle that. Maybe we’ll add an `self/`
alias that always refers to the main project node or something, and has
no need for a username. The `.*` pattern is simpler than `.+`, so we
should prefer it unless there’s a reason not to.

Beanow: Yeah that was what I was thinking. To have a bit of bias over what the username specific form can look like (similar to stripping @'s). But so long as we're exposing the function and not the pattern, either works for me :]

wchargin: Yep, agreed. You’re right that this is a bit inconsistent; we should
probably push the `@`-stripping down, too. Done.

PR #1414, opened 2019-10-18
identity: forbid underscores in GitHub logins

wchargin:
Summary:
GitHub logins may not have underscores, because underscores are not
valid characters in DNS labels. We already have a good-enough regular
expression for validating GitHub usernames; this commit updates the
alias parser to use that.

Discourse usernames are more permissive than what is listed here, but we
leave that unchanged for now.

Test Plan:
Unit tests updated.

wchargin-branch: alias-no-underscore

### DIFF
--- a/src/plugins/identity/alias.js
+++ b/src/plugins/identity/alias.js
@@ -1,6 +1,7 @@
 // @flow
 
 import {type NodeAddressT} from "../../core/graph";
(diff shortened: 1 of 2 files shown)


wchargin:
Yep! Hence the doc comment about this being a short-term implementation
that doesn’t need to scale. Thanks for the review!

PR #1416, opened 2019-10-18
Initiatives: create plugin declaration

Beanow:
Creates the declaration for the initiatives plugin and includes a readme.

If I understand [CONTRIBUTING.md](https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#while-writing-code) correctly, the idea is to have lots of small PRs.

Not sure if you prefer to merge the unused declaration into master, so right now I am targeting a feature branch that the semantically atomic commits can accumulate into.

I've also included rationale for the edge weights.

### DIFF
--- a/src/plugins/initiatives/declaration.js
+++ b/src/plugins/initiatives/declaration.js
@@ -1,0 +1,103 @@
+// @flow
+
+import deepFreeze from "deep-freeze";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {NodeType, EdgeType} from "../../analysis/types";
+import {NodeAddress, EdgeAddress} from "../../core/graph";
+
+export const nodePrefix = NodeAddress.fromParts(["sourcecred", "initiatives"]);
+export const edgePrefix = EdgeAddress.fromParts(["sourcecred", "initiatives"]);
+
+export const initiativeNodeType: NodeType = deepFreeze({
+  name: "Initiative",
+  pluralName: "Initiatives",
+  prefix: NodeAddress.append(nodePrefix, "initiative"),
+  defaultWeight: 1,
+  description:
+    "An initiative supernode, describing a scoped improvement to a project from proposal to completion.",
+});
+
+/*
+	Note on the forward and backward naming convention.
+	It follows the core/graph.js documentation to use
+	a <subject> <verb> <object> format to figure out
+	the directionality.
+*/
+
+/**
+ * An initiative (src) DEPENDS ON (verb) a dependency (dst).
+ * Forward: depending on something shows the value of the dependency.
+ * Backward: having a dependency does not endorse the iniative, but does flow
+ * some cred to incentivize reuse and attribution.
+ */
+export const dependsOnEdgeType: EdgeType = deepFreeze({
+  forwardName: "depends on",
+  backwardName: "is a dependency for",
+  prefix: EdgeAddress.append(edgePrefix, "dependsOn"),
+  defaultWeight: {forwards: 1, backwards: 1 / 16},
+  description: "Connects an initiative to it's dependencies.",
+});
+
+/**
+ * An initiative (src) REFERENCES (verb) a reference (dst).
+ * Forward: referencing from an initiative shows the value of the reference.
+ * But we assume a reference likely needs some refinement to be used by the initiative,
+ * so it flows less cred than to a dependency.
+ * Backward: having reference material does not endorse the iniative, but does flow
+ * some cred to incentivize using existing research and attribution.
+ */
+export const referencesEdgeType: EdgeType = deepFreeze({
+  forwardName: "references",
+  backwardName: "is referenced for",
+  prefix: EdgeAddress.append(edgePrefix, "references"),
+  defaultWeight: {forwards: 1 / 2, backwards: 1 / 16},
+  description: "Connects an initiative to it's references.",
+});
+
+/**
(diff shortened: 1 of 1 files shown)


Beanow:
Based on the missed eslint thing, made sure to fix my sublimelint extension.
Interesting we're also using prettier though. We can probably tighten the eslint configuration to closer match and include best practices. An example that come to mind: [`no-var`](https://eslint.org/docs/rules/no-var) with [`prefer-const`](https://eslint.org/docs/rules/prefer-const)

src/plugins/initiatives/declaration.js
```diff
@@ -0,0 +1,112 @@
...
+
+/*
+	Note on the forward and backward naming convention.
+	It follows the core/graph.js documentation to use
+	a <subject> <verb> <object> format to figure out
+	the directionality.
+*/
+
+/**
+ * A Discourse topic (src) TRACKS (verb) an initiative (dst).
```
decentralion: As discussed [here](https://discourse.sourcecred.io/t/write-the-initiatives-plugin/269/6?u=decentralion), I don't think we need a separate edge for this--we can consider the Discourse topic which defines an initiative to be an implicit contribution towards that initiative.

src/plugins/initiatives/declaration.js
```diff
@@ -0,0 +1,112 @@
...
+  forwardName: "references",
+  backwardName: "is referenced for",
+  prefix: EdgeAddress.append(edgePrefix, "references"),
+  defaultWeight: {forwards: 1 / 2, backwards: 1 / 16},
+  description: "Connects an initiative to it's references.",
+});
+
+/**
+ * A contribution (src) CONTRIBUTES TO (verb) an initiative (dst).
+ * Forward: receiving a contribution flows a medium amount of cred to the iniative.
```
decentralion: nit: i think this should be phrased with the contribution as the subject, e.g. "a contribution flows a medium amount of cred to the initiative"

src/plugins/initiatives/declaration.js
```diff
@@ -0,0 +1,112 @@
...
+ * An initiative (src) DEPENDS ON (verb) a dependency (dst).
+ * Forward: depending on something flows significant cred to the dependency.
+ * Backward: having a dependency does not flow much cred to the iniative,
+ * though does flow some to incentivize reuse and attribution.
+ */
+export const dependsOnEdgeType: EdgeType = deepFreeze({
+  forwardName: "depends on",
+  backwardName: "is a dependency for",
+  prefix: EdgeAddress.append(edgePrefix, "dependsOn"),
+  defaultWeight: {forwards: 1, backwards: 1 / 16},
```
decentralion: I notice we use 1 as "significant" and 1/2 as "medium".

Wondering if we should re-anchor on 1 as medium, and 2 or 4 as "significant".
(This might imply changes to other plugins, so its fine if we don't address it now.)

Beanow: I'm happy to adjust the comments in any way that is more clear. I used them as rational how I thought about value represented by edges. So long as the numerical weights are solid :D

Beanow:
> Yep, we do like having small semantic commits. Usually the way @wchargin and I do it is to build a stack of commits all pointing at master, and only merge commits as we become confident that we won't make more tweaks based on the yet-unmerged feature work. I'll let @wchargin comment on the idea of using feature branches.

So, for instance a later change like "remove the 'tracks' edge after all" instead would be to go back to these commits, modify it to look like it never had that edge type, rebase everything else, and only merge the whole thing until we're ready for the first iteration of the incentives system?

Beanow:
(Other than this PR being targeted to the feature branch)
I think this setup now has the right idea for commits right?

I've incorporated some feedback and included the "Missed the final declaration, tripping eslint." commit as well.
The new change is this: https://github.com/sourcecred/sourcecred/compare/master...dev-initiatives-declaration
Based off of that I have https://github.com/sourcecred/sourcecred/compare/dev-initiatives-declaration...dev-initiative-type

src/plugins/initiatives/declaration.js
Beanow: Done! :+1:

Beanow:
I've removed the readme from this PR, as it doesn't offer much value at this point anyway.

src/plugins/initiatives/declaration.js
Beanow: I've removed the edge type for trackers. Instead the idea is trackers will be added as a contribution. Using the editor to lower the weights in cases where the tracker doesn't provide much value in the form of discussion for example.

src/plugins/initiatives/declaration.js
Beanow: I've removed these quantifiers, and instead explained in more relative terms why some edges are assumed to be more important than others.

Beanow:
> Any particular reason to go with the pluralization as the plugin name?

Not really. Overall there seems to be inconsistency in this regard when looking at filenames.

There's plural
- src/analysis/types.js
- src/analysis/weights.js

There's singular
- src/core/address.js
- src/core/project.js

From the readability of an instance configuration file I feel like the plural fits slightly better.
Imagining something like this:

```yml
plugins:
  - github
  - initiatives
```

Could be turned into a question.
What optional features are we supporting? We're supporting github and initiatives.

Beanow:
Though the weights may still be subject to some balancing, pretty confident about this declaration.
Rebased against master to merge.

PR #1418, opened 2019-10-21
Implement Discourse initiative tracker parsing.

Beanow:
Continues from #1417 
This has a strict header format and crashes when it's not followed.

Initially I experimented with a very forgiving detection to get a best-effort result.
But considering the CredSperiment is the first and only target of this plugin right now, I believe crashing and forcing to update the wiki to match the expected format will produce more consistent results.

The expected pattern is:

```
All headers are case-insensitive and can be h1-h6.
Headers can appear in any order.
A matching header for each field must appear exactly once.

---

## Status: complete

Status value must be in the header, prefixed by "Status:".
Either "complete" or "completed". A missing status value,
or any other value is considered incomplete.

## Champions:

- [@Beanow](/u/beanow)

Any URLs that appear in the content below the "Champion" or "Champions" header.
No filters on user-like types applied here, that's left for after reference detection.

## Dependencies:

- [Dependency](/t/topic/123)

Any URLs that appear in the content below the "Dependency" or "Dependencies" header.

## References:

- [Reference](/t/topic/123)

Any URLs that appear in the content below the "Reference" or "References" header.

## Contributions:

- [Contribution](/t/topic/123)

Any URLs that appear in the content below the "Contribution" or "Contributions" header.
```

### DIFF
--- a/src/plugins/initiatives/discourse.js
+++ b/src/plugins/initiatives/discourse.js
@@ -1,0 +1,145 @@
+// @flow
+
+import {DomHandler, DomUtils, Parser} from "htmlparser2";
+
+import {type Topic, type Post} from "../discourse/fetch";
+import {type Initiative, type URL} from "./initiative";
+import {topicAddress} from "../discourse/address";
+
+type HeaderToURLsMap = Map<string, $ReadOnlyArray<URL>>;
+
+export async function initiativeFromDiscourseTracker(
+  serverUrl: string,
+  topic: Topic,
+  firstPost: Post
+): Promise<Initiative> {
+  if (firstPost.topicId !== topic.id) {
+    throw new Error("Post is from a different topic");
+  }
+
+  if (firstPost.indexWithinTopic !== 1) {
+    throw new Error("Post is not the first post in the topic");
+  }
+
+  const {title} = topic;
+  const {timestampMs} = firstPost;
+  const tracker = topicAddress(serverUrl, topic.id);
+
+  const htu: HeaderToURLsMap = await groupURLsByHeader(firstPost.cooked);
+  const completed = findCompletionStatus(htu);
+  const champions = singleMatch(htu, new RegExp(/^Champions?/i));
+  const contributions = singleMatch(htu, new RegExp(/^Contributions?/i));
+  const dependencies = singleMatch(htu, new RegExp(/^Dependenc(y|ies)/i));
+  const references = singleMatch(htu, new RegExp(/^References?/i));
+
+  const missing = [];
+  if (completed === null) missing.push("status");
+  if (!champions) missing.push("champions");
+  if (!contributions) missing.push("contributions");
+  if (!dependencies) missing.push("dependencies");
+  if (!references) missing.push("references");
+
+  if (
+    completed == null ||
+    champions == null ||
+    contributions == null ||
+    dependencies == null ||
+    references == null
+  ) {
+    missing.sort();
+    const missingStr = JSON.stringify(missing);
+    throw new Error(
+      `Missing or malformed headers ${missingStr} for initiative topic "${title}" (${topic.id})`
+    );
+  }
+
+  return {
+    title,
+    tracker,
+    timestampMs,
+    completed,
+    champions,
+    contributions,
+    dependencies,
+    references,
+  };
+}
+
+function findCompletionStatus(map: HeaderToURLsMap): boolean | null {
+  const pattern = new RegExp(/^Status:(.*)/i);
+  const headers = Array.from(map.keys())
+    .map((k) => k.trim())
+    .filter((k) => pattern.test(k));
+
+  if (headers.length !== 1) {
+    return null;
+  }
+
+  const matches = headers[0].match(pattern);
+  if (matches == null) {
+    return null;
+  }
+
+  const completedRE = new RegExp(/^completed?$/i);
+  return completedRE.test(matches[1].trim());
+}
+
+function singleMatch(
+  map: HeaderToURLsMap,
+  pattern: RegExp
+): $ReadOnlyArray<URL> | null {
+  const headers = Array.from(map.keys()).filter((k) => pattern.test(k.trim()));
+
+  if (headers.length !== 1) {
+    return null;
+  }
+
+  return map.get(headers[0]) || null;
+}
+
+export async function groupURLsByHeader(
+  cookedHTML: string
+): Promise<HeaderToURLsMap> {
+  const map: HeaderToURLsMap = new Map();
+  const dom = await promiseDOM(cookedHTML);
+
+  let currentHeader: string | null = null;
+  for (const rootEl of dom) {
+    switch (rootEl.name) {
+      case "h1":
+      case "h2":
+      case "h3":
+      case "h4":
+      case "h5":
+      case "h6":
+        currentHeader = DomUtils.getText(rootEl);
+        // We're also interested in just headers, so make sure an entry exists.
+        map.set(currentHeader, []);
+        break;
+      case "p":
+      case "ul":
+      case "ol":
+        if (currentHeader === null) break;
+        const existing = map.get(currentHeader) || [];
+        const anchors = DomUtils.findAll((el) => el.name === "a", [rootEl]).map(
+          (a) => a.attribs.href
+        );
+        map.set(currentHeader, [...existing, ...anchors]);
+        break;
+    }
+  }
+
+  return map;
+}
+
+function promiseDOM(cookedHTML: string): Promise<Object> {
+  return new Promise((res, rej) => {
+    const domHandler = new DomHandler((err, dom) => {
+      if (err) rej(err);
+      res(dom);
+    });
+    const htmlParser = new Parser(domHandler);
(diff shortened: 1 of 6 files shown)


src/plugins/initiatives/discourse.test.js
```diff
@@ -0,0 +1,271 @@
...
+        firstPost
+      );
+
+      // Then
+      expect(initiative.completed).toEqual(true);
+      expect(initiative).toMatchSnapshot();
+    });
+
+    it("considers blank status incomplete", async () => {
+      // Given
```
decentralion: The `// Given`, `// When` and `// Then` pattern is new to me. it seems a little redundant with just reading the code, can you explain what the motivation is?

src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,136 @@
...
+      case "h6":
+        currentHeader = DomUtils.getText(rootEl);
+        // We're also interested in just headers, so make sure an entry exists.
+        map.set(currentHeader, []);
+        break;
+      case "p":
+      case "ul":
+      case "ol":
+        if (currentHeader === null) break;
+        const existing = map.get(currentHeader) || [];
```
decentralion: could also use `MapUtil.pushValue` for each anchor.

src/plugins/initiatives/discourse.test.js
```diff
@@ -0,0 +1,271 @@
...
+        indexWithinTopic: 1,
+        replyToPostIndex: null,
+        timestampMs: 1571498171951,
+        authorUsername: "TestUser",
+        cooked: `
+            <h1>Example initiative</h1>
+            <h2>Status: Testing</h2>
+            <h2>Champion<a href="https://foo.bar/t/dont-include/10"><sup>?</sup></a>:</h2>
+            <p>
+              <a class="mention" href="/u/ChampUser">@ChampUser</a>
```
decentralion: interesting. i believe as written, the discourse plugin expects the links to look like `https://discourse.sourcecred.io/u/s_ben`. i wonder if the discourse plugin is actually not detecting any mentions at all right now?

src/plugins/initiatives/discourse.test.js
```diff
@@ -0,0 +1,271 @@
...
+      // When
+      const initiative = await initiativeFromDiscourseTracker(
+        serverUrl,
+        topic,
+        firstPost
+      );
+
+      // Then
+      expect(initiative.completed).toEqual(false);
+      expect(initiative).toMatchSnapshot();
```
decentralion: IMO, this would be a good fit for `toMatchInlineSnapshot` -- makes it much easier to check the output against the test case.

decentralion: also, note that right now the snapshot is interpreted as a binary file, which makes it quite hard to code review. this is likely because of the raw `NodeAddressT` in the initiative, which contains null bytes.

solutions that might work:
- `JSON.stringify` the initiative
- use NodeAddress.toString on the node address before snapshotting

src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,136 @@
...
+  const missing = [];
+  if (status === null) missing.push("status");
+  if (!champions) missing.push("champions");
+  if (!dependencies) missing.push("dependencies");
+  if (!references) missing.push("references");
+  if (!contributions) missing.push("contributions");
+
+  if (missing.length > 0) {
+    const missingStr = JSON.stringify(missing);
+    throw new Error(
```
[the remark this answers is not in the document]
Beanow: Doing a bunch of iterating based on top of this, I reached the same conclusion :]
That said, semantically I feel an error (rejected promise) is correct for this function. It's up to a higher level function to catch and turn it into a warning.

src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,136 @@
...
+
+  const completedRE = new RegExp(/^completed?$/i);
+  const completed = status ? completedRE.test(status.trim()) : false;
+
+  return {
+    title,
+    tracker,
+    timestampMs,
+    completed,
+    champions: champions || [],
```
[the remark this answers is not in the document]
Beanow: I disagree on this. There's a clear benefit here which flow does not handle in a pretty way.
I think having a list of things we did wrong is a lot more valuable than a single thing we did wrong and having to solve those problems one at a time because it failed to inform us all at once.

This is purely a coercion to convince flow it's never null at this point.
If there's a better way to cast the type to be not nullable here, I'd love to hear.
As we've checked for this with `if (missing.length > 0)`.

The same applies to `const completed = status ? completedRE.test(status.trim()) : false;` as status is not null here, this is a completely redundant if-statement when we know status is definitely a string here.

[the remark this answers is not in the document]
Beanow: Unless you meant in the same way that I did for the errors of this function.
Instead of `Thing | null` return types. Have it be `Thing` or throw. Catch those and throw the collection of errors.

src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,136 @@
...
+    contributions: contributions || [],
+    dependencies: dependencies || [],
+    references: references || [],
+  };
+}
+
+function findStatus(map: HeaderToURLsMap): string | null {
```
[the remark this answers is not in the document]
Beanow: Testing the value here against the regex makes sense.
That said, unspecified is ambiguous. Because an empty status value is considered incomplete. And the status header missing is considered an error.
Suggesting `boolean | null` return then.

src/plugins/initiatives/discourse.test.js
Beanow: The binary interpretation was why I didn't inline. I'll change it to the JSON version. :+1:

src/plugins/initiatives/discourse.test.js
Beanow: It's a convention I picked up from some BDD promotion blogs. It encourages you to separate input parameters (given) from the code being tested (when) and your assertion code (then), it's intended to make tests faster to read.

An example of where this influenced how I wrote tests.
Jest documentation has [example code](https://jestjs.io/docs/en/expect.html#tothrowerror) for toThrow like this:
```js
expect(() => {
  drinkFlavor('octopus');
}).toThrow();
```

Which conflates all of given, when and then, so I would have written it like:

```js

// Given
const flavor = 'octopus';

// When
const fn = () => drinkFlavor(flavor);

// Then
expect(fn).toThrow();
```

Examples of this being used with promise rejection:

```js
// When
const initiativeP = initiativeFromDiscourseTracker(
  serverUrl,
  topic,
  firstPost
);

// Then
await expect(initiativeP).rejects.toThrow(
  'Missing or malformed headers ["status","champions","dependencies","references","contributions"] for initiative topic "Example initiative" (123)'
);
```

src/plugins/initiatives/discourse.js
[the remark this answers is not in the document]
Beanow: I found the only other way to do this seems to be following flows' stricter type refinement approach. https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations

Replacing
```
if (missing.length > 0)
```

with 
```
if (
    completed == null ||
    champions == null ||
    dependencies == null ||
    references == null ||
    contributions == null
  )
```
even though it's redundant.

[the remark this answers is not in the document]
Beanow: I've implemented the later option, with the large if statement to do flow type refinement.

src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,136 @@
...
+
+  if (firstPost.indexWithinTopic !== 1) {
+    throw new Error("Post is not the first post in the topic");
+  }
+
+  const {title} = topic;
+  const {timestampMs} = firstPost;
+  const tracker = topicAddress(serverUrl, topic.id);
+
+  const map = await groupURLsByHeader(firstPost.cooked);
```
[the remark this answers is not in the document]
Beanow: I've included the type, though still use a shorthand name, otherwise prettier will indent some calls but not others. Making it much less readable imo.

src/plugins/initiatives/discourse.js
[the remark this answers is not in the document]
Beanow: Implemented as `boolean | null` with the completion regex moved to this function.

src/plugins/initiatives/discourse.test.js
Beanow: Tracking that in https://github.com/sourcecred/sourcecred/issues/1420, I think this will be up to the discourse data source or reference detection. When that's updated this test can be changed to reflect it.

src/plugins/initiatives/discourse.test.js
Beanow: The JSON version added a lot of escaping cruft, went for a `NodeAddress.toParts` approach.

Beanow:
I'm going to push this one back in development order. Starting with graph generation using mocks first, then working towards a Discourse source for Initiatives.

Beanow:
This PR is replaced by #1478 + #1480.
Separating the "cooked HTML" parsing from Discourse semantics.

PR #1423, opened 2019-10-23
discourse: factor out snapshotTestUtil

decentralion:
I want to have the reference tests depend on real snapshotted data.
Therefore, I'm factoring out the utilities for interacting with the
snapshot data out of fetch.test.js and into snapshotTestUtil.js

Test plan: `yarn test` still passes.

### DIFF
--- a/src/plugins/discourse/fetch.test.js
+++ b/src/plugins/discourse/fetch.test.js
@@ -1,34 +1,10 @@
 // @flow
 
-import deepFreeze from "deep-freeze";
-import {Fetcher, type DiscourseFetchOptions} from "./fetch";
-import base64url from "base64url";
-import path from "path";
-import fs from "fs-extra";
+import {Fetcher} from "./fetch";
+import {options, snapshotFetcher} from "./mockSnapshotFetcher";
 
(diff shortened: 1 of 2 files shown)


Beanow:
This is an example why I have a preference towards a `/src` folder separate from `/test`.

src/plugins/discourse/snapshotTestUtil.js
```diff
@@ -0,0 +1,26 @@
...
+  const filename = base64url(url);
+  const file = path.join(snapshotDir, filename);
+  if (await fs.exists(file)) {
+    const contents = await fs.readFile(file);
+    return new Response(contents, {status: 200});
+  } else {
+    throw new Error(`couldn't load snapshot for ${file}`);
+  }
+}
+export const snapshotFetcher = () => new Fetcher(options, snapshotFetch, 0);
```
Beanow: nit: the file contains only fetch related features. It should have fetch in the filename.
Suggesting `mockSnapshotFetcher`.

decentralion: Done.

PR #1424, opened 2019-10-23
Discourse plugin now properly detects mentions

### LINKED ISSUE #1420: Discourse plugin doesn't detect mentions
The Discourse plugin expects full absolute URLs leading to user identities, as seen here:
https://github.com/sourcecred/sourcecred/blob/01bdb2e94a1d8130f0cf1301e953408b4c788ff5/src/plugins/discourse/references.js#L62

However, as noted in [this review comment](https://github.com/sourcecred/sourcecred/pull/1418#discussion_r337319287), Discourse actually generates relative links for user mentions.

We should fix this, and ensure the references test code uses some real references from the test instance.

decentralion:
As suggested in #1420, heretofore the Discourse plugin wasn't actually
picking up mentions. The issue is that the (thoroughly tested) mention
detection logic assumed that mention urls took the form
`$SERVERURL/u/$USERNAME`, but actually they are encoded as a relative link,
as in `/u/$USERNAME`. As such, the logic was internally consistent but
never detected any actual mentions!

It's a good case study in the need for integration tests and not just
unit tests. I've updaded the code so we do have a proper integration
test: references.test.js validates that a topic reference, post
reference, and user mention are all properly detected in the real output
from a Discoures topic.

Test plan: `yarn test` passes; inspect updated snapshots and tests.

Fixes #1420.

### DIFF
--- a/src/plugins/discourse/references.js
+++ b/src/plugins/discourse/references.js
@@ -29,9 +29,23 @@ export type DiscourseReference =
   | DiscourseUserReference;
 
 export type UrlString = string;
-export function parseLinks(cookedHtml: string): UrlString[] {
+/**
+ * Parse the links from a Discourse post's cookedHtml, generating
+ * an array of UrlStrings. All of the UrlStrings will contain the full
+ * server URL (i.e. relative references are made absolute). The serverUrl
+ * is required so that we can do this.
+ */
+export function parseLinks(cookedHtml: string, serverUrl: string): UrlString[] {
   const links = [];
   const httpRegex = /^https?:\/\//;
+  if (serverUrl[serverUrl.length - 1] === "/") {
+    // Strip trailing slash if it was provided, so we can concatenate
+    // strings below.
+    serverUrl = serverUrl.slice(0, serverUrl.length - 1);
+  }
+  if (!serverUrl.match(httpRegex)) {
+    throw new Error(`Invalid server url ${serverUrl}`);
+  }
   const parser = new htmlparser2.Parser({
     onopentag(name, attribs) {
       if (name === "a") {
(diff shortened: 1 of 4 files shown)


src/plugins/discourse/references.test.js
```diff
@@ -106,42 +112,79 @@ describe("plugins/discourse/references", () => {
...
+                    "topicId": 287,
+                    "type": "POST",
+                  },
+                  "https://discourse.sourcecred.io/u/decentralion": undefined,
+                  "https://forums.eveonline.com/t/topic/195153": Object {
+                    "serverUrl": "https://forums.eveonline.com",
+                    "topicId": 195153,
+                    "type": "TOPIC",
+                  },
+                  "https://forums.eveonline.com/u/dorian_neil/summary": undefined,
```
Beanow: I feel like absolute URLs should still be supported.

src/plugins/discourse/references.js
```diff
@@ -59,7 +60,7 @@ export function linksToReferences(
 
   const topicRegex = new RegExp(`^(?:${topic})(?:${params})/?$`);
   const postRegex = new RegExp(`^(?:${post})(?:${params})/?$`);
-  const userRegex = new RegExp(`^(?:${server})/u/([\\w-]+)(?:${params})/?$`);
+  const userRegex = new RegExp(`^/u/([\\w-]+)(?:${params})?$`);
```
Beanow: Proposing instead of support replacing the server with `/` for all of these, we force normalize it's input URLs to be all absolute. Because in the future (unified) reference detection version, I would argue the core reference detector no longer has the context of the base URL that the URL watch fetched from.

src/plugins/discourse/references.test.js
decentralion: Fixed.

src/plugins/discourse/references.js
decentralion: Great suggestion, thank you. I've implemented it, and added a bevy of new integration tests that verify that the full reference detection pipeline works on both absolute and relative references.

decentralion:
I'm personally satisfied that I've addressed @Beanow's concerns, and I want to get mention detection in for this week's cred, so I'm going to merge this.

Issue #1426, opened 2019-10-25
Discourse mirror issues and possible enhancements


Beanow:
Digging into the Discourse plugin I found a whole bunch of things surrounding the mirror:

Enhancements:
- The API key is not required. All data is available through anonymous access.
- The rate limit does not apply to anonymous access allows for much faster and parallel requests.
  Adding parallel requests - discourse: 3m 22s
  Current implementation - discourse: 8m 28s
- The `Mirror` class has conflated SQL queries, read-only methods, internal write methods and mirroring logic. I would like to split this into a read-interface, read-write interface, an implementation of those and mirroring logic separately.

Issues:
- Resume correctness is poor. The very first topic / post that are saved are the latest ones.
  From there it uses the max ID and assumes this is the head of what's last been synced.
  Should you cancel the update soon after saving the latest topics / posts, you will be missing 99%   of all data and never attempt to load it.
- Edits are not handled at all. #1419 
- Most of the time, we prefer local data over remote data. Even though remote data is authoritative.

Notes:
- There's a very heavy reliance on the assumption topic and post ids are incremental.
- The (remote) API leaves a lot to be desired in terms of calls suitable for syncing. Making it difficult to be both fast and correct. For example, no likes sorted by date and no topics sorted by _any_ update, just by updates that constitute a "bump".

I would be interested in refactoring most if not all of the mirroring logic to improve these aspects.

Beanow:
I've gotten pretty far into a big overhaul: https://github.com/sourcecred/sourcecred/compare/master...wip/refactor-discourse-plugin

Before I restructure the commits and make another pass on handling TODO items, I'd like to discuss the changes on a high level and get your feedback on that.
cc @decentralion @wchargin 

- Now using anonymous API access.
- Removed API key and username all across the code.
- Changed project structure field to: `discourseServers: $ReadOnlyArray<string>`
  Supporting multiple servers per project, and removing the username.
- Updated load API and CLI to support multiple discourse servers.
  _(Currently multiple discourse servers in one project is incompatible with the `identities` plugin,
  as it uses ambiguous `discourse/@username` aliasing.)_
- Moved SQL operations into `MirrorRepository` class, which implements a read-only interface which the graph creation step accepts, having the syncing logic in `Mirror`.

On top of that, a major overhaul of how topics and posts are retrieved.
1. We walk through `/latest.json?order=activity&ascending=false&page=${page}`
  to find all the topics that were "bumped" since our last update.
[...]

PR #1431, opened 2019-10-30
Remove Discourse admin API key and user.

Beanow:
This removes all usage of and reference to the admin API key and username. Instead relying on anonymous access of the Discourse API.

This enables anyone to deploy an instance with discourse support, and is much safer, since the admin API key isn't used for this purpose anymore. Once merged I would encourage revoking any admin API keys used in the past.

The only notable remaining reference of the discourse username is in the project file.
Which goes from `0.3.0` to `0.3.1` in a backwards-compatible way here, simply ignoring the username if present. For #1426 I'm expecting a `0.4.0` version, so this is to prevent having to change project files twice.

Test plan: updated the snapshots to their latest anonymous versions. Ran yarn test and anonymous discourse loading from CLI numerous times.


src/core/project.js
```diff
@@ -42,7 +44,7 @@ export function projectToJSON(p: Project): ProjectJSON {
 }
 
 export function projectFromJSON(j: ProjectJSON): Project {
-  return fromCompat(COMPAT_INFO, j);
+  return fromCompat(COMPAT_INFO, j, upgrades);
```
decentralion: Nice! I think this is the first use of the `upgrades` API. @wchargin and I have gone back and forth a bit on the quality of this API, I'd be curious to hear what you thought of this interface.

decentralion:
Question: are any of the snapshot updates actually a result of the changes in this PR, or are they just drift on upstream? (The snapshot tests are never consistent from snapshot to snapshot because, among other things, the view count is returned by the server.)

In general I try to separate irrelevant / noisy snapshot updates from the PRs that touch DIscourse, for easier and clearer review. My ideal pattern would be:
- PR with pre-snapshot update (trivial approve)
- PR with the code changes, with another snapshot update included (but this one will only reveal differences due to the code change, as opposed to change on the upstream)

Beanow:
@decentralion
> Question: are any of the snapshot updates actually a result of the changes in this PR

There are actual changes. Because the requesting user (anonymous) has fewer privileges. It removes some "can do this action: true" style fields, "have you read this yet" and data like user creation dates. None of which we actively used internally, but they're different between using and not using the API key.

decentralion:
Great. I never wanted that user-specific data to begin with.

PR #1432, opened 2019-10-30
Discourse: split Mirror from MirrorRepository

Beanow:
Continues from #1431.
As suggested in #1426.
This splits the mirror class into a read and read-write repository interface. along with an implementation for the SQL side. And keeps updating logic in the mirror class.

Test plan: updated existing tests to reflect and ran those.

### DIFF
--- a/src/plugins/discourse/createGraph.js
+++ b/src/plugins/discourse/createGraph.js
@@ -16,7 +16,7 @@ import {
   type Topic,
   type LikeAction,
 } from "./fetch";
-import {type DiscourseData} from "./mirror";
+import {type ReadRepository} from "./mirrorRepository";
 import {
   authorsPostEdgeType,
   authorsTopicEdgeType,
@@ -143,7 +143,7 @@ export function likesEdge(serverUrl: string, like: LikeAction): Edge {
   };
 }
 
-export function createGraph(serverUrl: string, data: DiscourseData): Graph {
+export function createGraph(serverUrl: string, data: ReadRepository): Graph {
   const gc = new _GraphCreator(serverUrl, data);
   return gc.graph;
 }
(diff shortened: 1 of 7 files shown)


Beanow:
It draws in particular from [DDD](https://en.wikipedia.org/wiki/Domain-driven_design)

> Repository
>    Methods for retrieving domain objects should delegate to a specialized Repository object such that alternative storage implementations may be easily interchanged.

Which in turn is based a lot on [Multitier architecture](https://en.wikipedia.org/wiki/Multitier_architecture) where the data access and business logic layers are separated.

Using the trick of having a subset through the read-only interface is new for me _within javascript_ as well, since flow won't really enforce it at runtime. We could make a more strict separation where the write methods won't actually be exposed to parts that request the read interface.

decentralion:
> Using the trick of having a subset through the read-only interface is new for me within javascript as well, since flow won't really enforce it at runtime. We could make a more strict separation where the write methods won't actually be exposed to parts that request the read interface.

Getting an error from flow is good enough for me.

Beanow:
With #1443 and #1431 merged, rebased against master and fixed the conflicts.

decentralion:
@wchargin care to review? if not i'll merge on my own approval.

decentralion:
@Beanow you should be able to merge it (I changed # of approving reviews from 6 to 1)

Beanow:
> Merging is blocked
> The base branch restricts merging to authorized users. Learn more about protected branches.

For `master` looks like it would need ownership.

Also updated title based on @wchargin's suggestion about trailing dot in the first commit line and contributing guidelines. On squash I suggest using that as the commit message.

decentralion:
Great! Having there be an explicit list of people who are allowed to merge to master is exactly what I want, that way we can ensure that people who are merging to master understand our expectations around commit history, commit messages, test plans, etc.

I've added you to that list; you should now be able to merge this yourself. :)

Beanow:
Rebased and did a minor conflict resolution. #1446 added a new test case, which used the mirror for queries, which now should be the repo.

Issue #1415, opened 2019-10-18
Allow for combining data from multiple organizations


jakubgs:
In your `README` you state that:

>You can also combine data from multiple repositories into a single graph. To do so, pass multiple repositories to the load command, and specify an “output name” for the repository. For instance, the invocation
>```
>node bin/sourcecred.js load ipfs/js-ipfs ipfs/go-ipfs --output ipfs/meta-ipfs
>```
>will create a graph called ipfs/meta-ipfs in the cred explorer, containing the combined contents of the js-ipfs and go-ipfs repositories.

But this does not work for organizations:
```
+ nodejs /opt/sourcecred/repo/bin/sourcecred.js load @A @B @C @D --output combined                                                                                                                                                                                                           
Error: invalid spec: --output                                                                                                                                                                                                                                                                                                 
    at specToProject (webpack:///./src/plugins/github/specToProject.js?:20:804)
    at Promise.all.projectSpecs.map.s (webpack:///./src/cli/load.js?:17:1475)
    at Array.map (<anonymous>)
    at loadCommand (webpack:///./src/cli/load.js?:17:1385)
    at sourcecred (webpack:///./src/cli/sourcecred.js?:10:449)
    at eval (webpack:///./src/cli/command.js?:3:171)
    at main (webpack:///./src/cli/main.js?:5:206)
    at eval (webpack:///./src/cli/main.js?:6:84)
    at Module../src/cli/main.js (/opt/sourcecred/repo/bin/sourcecred.js:299:1)
    at __webpack_require__ (/opt/sourcecred/repo/bin/sourcecred.js:30:30)
```

Beanow:
Hey @jakubgs, good find!

`--output` used to be a hack to support github organizations by adding all the repositories you care about and pretending it was one big repository, this was before the `@org` notation was properly supported. Removing that option should allow you to load.

They will then be listed separately in the explorer, same as https://sourcecred.teamopen.dev/prototype/

If you need more fine-grained control. You can use a `--project <json file>`.
An example in this repository https://github.com/sourcecred/cred

I haven't tried using this for combining multiple organizations yet, going to have a look what that does.

Meanwhile, I'm curious what your use-case is for _combining_ the data from multiple organizations.

The working idea going forward is actually to not support this. Because we'd like to encourage each community to decided on their own valuations of what should earn cred and how much, right now that means setting the weights. That unit would be called an instance. Then use a higher-level interface (such as the `scores` format) to aggregate multiple instances.

[...]

PR #1433, opened 2019-10-30
Move to multiple servers with mirror options.

Beanow:
Continues from #1432 

As discussed in #1426, this adds multiple server support per project. Including loading in parallel and CLI updates. It anticipates on what the `MirrorOptions` object will be, as this is included in the project `0.4.0` format.

It creates an interesting compatibility issue with the identities plugin though, as multiple servers can't be interpreted by the `discourse/@username` aliases.

Test plan:
```
yarn install
yarn backend

node bin/sourcecred.js discourse \
  https://sourcecred-test.discourse.group \
  https://discourse.sustainoss.org \
  https://discourse.sourcecred.io
```


decentralion:
This PR adds a feature (multiple server support) which might introduce a number of subtle bugs (you pointed out one re: the identity plugin, and there may be more). Do we know of any use-cases or needs for this feature at the moment?

If we do want to merge it now, should also spend some time thinking about potential bugs it will introduce, and put some effort towards fixing them now while the context is fresh. But if we don't have any users for this feature, I don't think that's a good use of time.

Beanow:
Aside from the one with identity, there's several places in the explorer that assumes one discourse server. You're right that I don't know of any communities that have more than one forum, but we have had #1415 multiple orgs for github which I wouldn't have assumed for most communities either.

Not having this feature will make flowing cred between these forums, reference detection, instance config, etc. harder to do later as features stack up with the one-forum assumption. This would put pressure to break that assumption at the project level.

We could push this back as part of deep-than-wide. But I think breaking the assumption earlier will also help towards finding good APIs and fixing those assumption bugs. As the same would apply to Discord, where we have one, but I'm sure there will be communities with more.

decentralion:
Cross-community cred isn't something I'm focused on right now (it raises a lot of thorny issues around governance) but it is definitely something that warrants thought and experimentation. And you raise a good point around how making this change later will become more expensive, after we've had more time to assume that there's always one server.

However, at the moment we're focused on work that gives high leverage to SourceCred, by making SourceCred better at self-organizing the labor (e.g. via initiatives and champions) to direct coordinate engineering work. So in a sense the current "cost of engineering work" is very high, since we have little engineering bandwidth, but if we invest it well the cost will decrease a lot in the future.

Therefore, we should be willing to punt on features that don't give us much short term benefit, even if they will grow more expensive over time, if the rate at which they grow expensive is lower than the rate our resources will grow. and I don't think that this change will grow that much more expensive over time.

This was all a very complicated way of saying: "Yeah, deep then wide, I think we should punt on this for now."

decentralion:
@beanow shall we close this as wontfix for now?

Beanow:
> @Beanow shall we close this as wontfix for now?

#1451 is the alternative implementation. I'll indeed close this as wontfix.

PR #1434, opened 2019-10-30
Implement revised sync mechanism.

Beanow:
Continues from #1433 

As discussed in #1426, this adds the new approach to mirroring topics and posts with the Discourse API.

Since then I've also added support for deleting posts from a topic. And added the option to always recheck topics from a particular category. This is currently only exposed through project files.
For example, this would always recheck the initiatives subcategory.

```json
[
  {
    "type": "sourcecred/project",
    "version": "0.4.0"
  },
  {
    "discourseServers": [
      {
        "serverUrl": "https://discourse.sourcecred.io",
        "mirrorOptions": {
          "recheckTopicsInCategories": [16]
        }
      }
    ],
    "id": "@sourcecred",
    "identities": [],
    "repoIds": []
  }
]

```

There's one-or-two TODO's in there that I'll have another look at to revise. As well as a benchmark of parallel loading and writing, since it did add complexity, want to be sure it's worth keeping.

Test plan: expanded and revised the unit tests a lot to cover the new syncing method, as well as lots of manual loading through projects files and the discourse command.


src/plugins/discourse/fetch.js
```diff
@@ -149,23 +149,119 @@ export class Fetcher implements Discourse {
     return this._fetchImplementation(fullUrl, fetchOptions);
   }
 
-  async latestTopicId(): Promise<TopicId> {
-    const response = await this._fetch("/latest.json?order=created");
+  async categoryDefinitionTopicIds(): Promise<TopicId[]> {
+    const topicIdRE = new RegExp("/t/[\\w-]+/(\\d+)$");
+    const urls = [];
+    const withSubs = [];
```
decentralion: suggest change to `categoriesWithSubcategories: CategoryId[]`

src/plugins/discourse/fetch.js
```diff
@@ -149,23 +149,119 @@ export class Fetcher implements Discourse {
...
-    failForNotOk(response);
-    const json = await response.json();
-    return json.latest_posts.map(parsePost);
+  async topicsBumpedSince(sinceMs: number): Promise<Topic[]> {
+    const topics = [];
+    let pastSince = false;
+    let morePages = true;
+    let page = 0;
+
+    while (!pastSince && morePages) {
```
decentralion: idea: instead of `while (!pastSince && morePages)` how about `while (lastUnpinnedTimestamp <= sinceMs && morePages)` (with the obvious semantics for `lastUnpinnedTimestamp`).

As it currently is, the logic for the control loop is split between the while declaration, and the bottom of the function where `pastSince` is evaluated. With this change, the conceptual control logic is declared all in one place, with only the implementation of tracking `lastUnpinnedTimestamp` at the bottom.

src/plugins/discourse/fetch.js
```diff
@@ -149,23 +149,119 @@ export class Fetcher implements Discourse {
...
+    return uniqueIds;
   }
 
-  async latestPosts(): Promise<Post[]> {
-    const response = await this._fetch("/posts.json");
-    failIfMissing(response);
-    failForNotOk(response);
-    const json = await response.json();
-    return json.latest_posts.map(parsePost);
+  async topicsBumpedSince(sinceMs: number): Promise<Topic[]> {
```
decentralion: this should have a docstring, which clearly documents the answers to questions like: "does this include recently bumped about-category-x topics, or do they need to be fetched separately?"

src/plugins/discourse/fetch.js
decentralion: I'm generally a bit miffed with this code scope, it feels like there's a lot going on, including a lot of mutation of state outside the scope, and it's hard for me to feel confident that i understand it. 

what if we define helper functions like:

`processTopic`: validates an individual topic and returns the parsed Topic, along with any extra summary info
`addPage`: collects all the topics a whole response page, also returns the information needed to decide whether there are more pages 

then the outer block can just call `addPage` in a loop, and it will be easier for the reader to focus on individual pieces of functionality

src/plugins/discourse/fetch.js
```diff
@@ -178,26 +274,32 @@ export class Fetcher implements Discourse {
...
-    const posts = json.post_stream.posts.map(parsePost);
+    const {posts_count: postCount} = json;
+    let posts = json.post_stream.posts.map(parsePost);
     const topic: Topic = {
       id: json.id,
+      categoryId: json.category_id,
       title: json.title,
-      timestampMs: +new Date(json.created_at),
+      timestampMs: Date.parse(json.created_at),
+      bumpedMs: json.bumped_at
```
decentralion: nit: although in practice bumped_at should never have a timestamp of 0, as a matter of habit, I would write `json.bumped_at !== null`

decentralion:
Note: if https://github.com/sourcecred/sourcecred/pull/1443 merges first, we should make sure it gets re-incorporated into this change.

Beanow:
> could you break this PR into smaller pieces of functionality? E.g., introducing topicsBumpedSince and categoryDefinitionTopics each in their own PR.

I'll have a look what I can do at breaking these up smaller. @decentralion
Although these are some highly coupled changes. For instance, updating just the fetcher would completely break the mirror and all it's tests.

Regardless, the base branches have changed a fair bit, so I'll most likely abandon this PR and create some revised ones with the feedback included.

decentralion:
@Beanow I do prefer a multistage process which adds a new fetcher, builds a new mirror, swaps, and then deletes the old fetcher / old mirror. @wchargin did something similar with his EAV refactor.

src/plugins/discourse/fetch.js
Beanow: Not sure what you mean exactly here. Do you mean, to fail this method when bumped_at returned by the API is null?

decentralion: Hmm, I think I wrote that nit from a place of confusion... somehow I thought `bumpMs` was a boolean and we were doing an implicit type conversion or something. Please ignore that comment. :)

Beanow:
I've gone back, starting to break these up into parts. The first ones are there at #1452 and #1453.

Build new mirror and swap wouldn't be as applicable as the EAV refactor. As the same data could be inserted in different forms. Here the fetching itself is different, so it would fetch all data twice. :thinking: not sure that's a helpful stage.

In the meantime, closing this to indicate the other PRs will replace it.

src/plugins/discourse/fetch.js
Beanow: The nullable bumpedMs was really causing a lot of confusion around the codebase. In #1452 I've shifted this problem into the flow types. As some API calls have this field and others don't. I've given them different types instead of juggling with the nullable everywhere.

decentralion:
You could build the new fetcher, new mirror, (both unused), make the swap (use new fetcher and mirror, leave old code around unused), then delete the old code--I believe that's the pattern that @wchargin took for the EAV. That said, I'm not sure that all the effort is needed in this case.

PR #1441, opened 2019-11-11
test: fix `example-github-load` snaphshot test

wchargin:
Summary:
Generated with `./scripts/update_snapshots.sh`. This fixes failures
introduced in #1431.

Test Plan:
Running `yarn test --full` now passes. Inspecting the diff shows that
this only includes a compat version number change, which is appropriate.

wchargin-branch: fix-1431-failures


wchargin:
This change fixes exactly one bug by flipping exactly one bit.

Issue #1440, opened 2019-11-10
discourse: issue processing zcash forum


s-ben:
I tried running the discourse plugin on the [zcash forum](https://forum.zcashcommunity.com/),

`node bin/sourcecred.js discourse https://forum.zcashcommunity.com`

Got this error: 

```FetchError: request to https://forum.zcashcommunity.com/t/25608.json failed, reason: read ECONNRESET
    at ClientRequest.<anonymous> (/Users/sben/Documents/coins/sourcecred/sourcecred/node_modules/node-fetch/index.js:133:11)
    at ClientRequest.emit (events.js:197:13)
    at TLSSocket.socketErrorListener (_http_client.js:397:9)
    at TLSSocket.emit (events.js:197:13)
    at emitErrorNT (internal/streams/destroy.js:82:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
    at processTicksAndRejections (internal/process/next_tick.js:76:17)
```
Tried again later (just running the command again, not deleting data) and got some warnings along with this error:

```
Warning: Encountered error 'FOREIGN KEY constraint failed' on a like by t2yax on post id 29446.
Warning: Encountered error 'FOREIGN KEY constraint failed' on a like by t2yax on post id 25320.
Warning: Encountered error 'FOREIGN KEY constraint failed' on a like by t2yax on post id 22304.
Warning: Encountered error 'FOREIGN KEY constraint failed' on a like by t2yax on post id 22367.
Warning: Encountered error 'FOREIGN KEY constraint failed' on a like by t2yax on post id 20368.
Error: 404 Not Found on: https://forum.zcashcommunity.com/user_actions.json?username=kernelpanik23&filter=1&offset=0; maybe bad serverUrl?
    at failIfMissing (webpack:///./src/plugins/discourse/fetch.js?:35:433)
    at Fetcher.likesByUser (webpack:///./src/plugins/discourse/fetch.js?:35:246)
```

Note: the forum is a few years old and quite large. May take a couple hours to run.

decentralion:
https://github.com/sourcecred/sourcecred/pull/1443 should fix the second issue.
For the first one, we should add retry logic around the fetch to save the user from needing to re-launch themselves.

Beanow:
The 404 seems to indicate the user not existing (anymore?), rather than there being no likes.|
As https://forum.zcashcommunity.com/u/kernelpanik23/summary is also a 404.
[...]

PR #1443, opened 2019-11-12
Make Discourse robust to 404s on user actions

decentralion:
This fixes the non-recoverable error in #1440; namely SourceCred
crashing when the Discourse server returns 404 for a user's actions. I'm
not sure why this happens (maybe DB is in an inconsistent state?) but
missing the likes for a particular user is less frustrating than not
being able to load cred at all.

I've also added a unit test which verifies this behavior; I've confirmed
that before applying the fix, test test fails.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/discourse/mirror.js
+++ b/src/plugins/discourse/mirror.js
@@ -456,7 +456,7 @@ export class Mirror implements DiscourseData {
     })();
 
     reporter.start("discourse/likes");
-    for (const user of this.users()) {
+    const addUserLikes = async (user: string) => {
       let offset = 0;
       let upToDate = false;
       while (!upToDate) {
@@ -472,6 +472,16 @@ export class Mirror implements DiscourseData {
           upToDate = true;
         }
         offset += likeActions.length;
+      }
+    };
+    for (const user of this.users()) {
+      try {
+        await addUserLikes(user);
+      } catch (e) {
+        console.warn(
+          `Warning: Encountered error '${e.message}' ` +
+            `while processing likes for ${user}; skipping this user.`
+        );
       }
     }
     reporter.finish("discourse/likes");
--- a/src/plugins/discourse/mirror.test.js
+++ b/src/plugins/discourse/mirror.test.js
@@ -467,6 +467,24 @@ describe("plugins/discourse/mirror", () => {
     });
   });
 
+  it("warns if a user's likes are missing", async () => {
+    const {mirror, fetcher, reporter} = example();
+    const pid = fetcher.addPost(1, null, "credbot");
+    (fetcher: any).likesByUser = async () => {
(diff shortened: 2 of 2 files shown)


src/plugins/discourse/mirror.js
```diff
@@ -473,6 +473,16 @@ export class Mirror implements DiscourseData {
         }
         offset += likeActions.length;
       }
+    };
+    for (const user of this.users()) {
+      try {
+        await addUserLikes(user);
+      } catch (e) {
+        console.warn(
```
wchargin: OK—in general I’d prefer to avoid logspam for cases that we know will be
hit, like this one, but I’m willing to wait and see how common this is.
Ideally, we could save “info-logging” statements like this to a separate
log that’s displayed only on request.

src/plugins/discourse/mirror.test.js
```diff
@@ -467,6 +467,24 @@ describe("plugins/discourse/mirror", () => {
...
+    await mirror.update(reporter);
+    expect(mirror.topics()).toEqual([fetcher._topic(1)]);
+    expect(mirror.posts()).toEqual([fetcher._post(pid)]);
+    expect(mirror.likes()).toEqual([]);
+    expect(console.warn).toHaveBeenCalledWith(
+      "Warning: Encountered error '404 - Likes not found' " +
+        "while processing likes for credbot; skipping this user."
+    );
+    expect(console.warn).toHaveBeenCalledTimes(1);
+    jest.spyOn(console, "warn").mockImplementation(() => {});
```
wchargin: Willing to let this slide because I see that it’s following an existing
pattern, but this doesn’t feel right—this should be handled by
`beforeEach`/etc. callbacks. If one of the assertions in the test fails,
we’ll never effect this restoration. But I don’t really understand why
the `beforeEach`/etc. callbacks are set up the way that they are, so
willing to defer.

s-ben:
Was able to run with no errors this time and get timeline cred. 

image

When I switch to legacy mode though, I get the below error. Is this expected?
image

decentralion:
@s-ben legacy mode is barely supported at this point... but please open up the JavaScript debugging console and paste in whatever full error + stack trace it gave, and I can get a sense of how hard it would be to fix.

src/plugins/discourse/mirror.js
```diff
@@ -473,6 +473,16 @@ export class Mirror implements DiscourseData {
         }
         offset += likeActions.length;
       }
+    };
+    for (const user of this.users()) {
+      try {
+        await addUserLikes(user);
+      } catch (e) {
```
Beanow: I believe this is far too permissive of the problem we're trying to be robust to. For example, bug in the SQL query would cause ignores here, while I feel they should crash.

PR #1453, opened 2019-11-13
Discourse: extend repository

Beanow:
Continues from #1452

These are additions to the MirrorRepository, including feedback from #1434.
If it helps to review, each commit can be viewed in isolation.

Test plan: `yarn unit`


Beanow:
Will break these up into individual PRs, based on #1452 request.

decentralion:
Thanks!

PR #1454, opened 2019-11-15
Discourse: add categoryId and bumpedMs to Topic data

Beanow:
As not all API calls return bumpedMs, make a new type to show the distinction.

Test plan: `yarn unit`


Beanow:
Rebased due to squash merge of parent #1451, no further changes.

PR #1455, opened 2019-11-15
Discourse: make topicWithPosts fetch all posts

Beanow:
Previously it would only consider page 1.
Now we're walking through all pages, as this is a much more effective way of discovering all posts.

Test plan: `yarn unit`
Created a new paginated topic for this at https://sourcecred-test.discourse.group/t/big-filler-topic-1/26


Beanow:
Rebased due to squash merge of parent #1454.

PR #1456, opened 2019-11-15
Discourse: add categoryDefinitionTopicIds to fetcher

Beanow:

Test plan: `yarn unit`
Created the a category and subcategory for this at
https://sourcecred-test.discourse.group/c/filler/big-filler


src/plugins/discourse/fetch.js
```diff
@@ -164,6 +168,57 @@ export class Fetcher implements Discourse {
...
+      const match = topicIdRE.exec(url);
+      if (match == null) {
+        throw new Error(
+          `Encountered topic URL we failed to parse it's TopicId from: ${url}`
+        );
+      }
+      return Number(match[1]);
+    });
+
+    return new Set(ids);
```
Beanow: This is modified from the comment https://github.com/sourcecred/sourcecred/pull/1452#discussion_r346069716


> Why would they not be unique? Can multiple categories share a topic?

---

> Fair point. I've included it as a better safe than sorry, explicit semantic constraint. For efficiency, the syncing relies on those semantics. Perhaps it's better to just change the return type to be a Set<TopicId>.
>
> In terms of the API, I think it shouldn't currently be possible to share topics between categories. But I've not looked too far into it, as it's a pretty niche thing to want very elaborate setups for the "about x category" definition topics.

---

This now has the `Set<TopicId>` type.

Beanow:
Rebased due to squash merge of parent #1455.

PR #1457, opened 2019-11-15
Discourse: implement topicsBumpedSince in fetcher

Beanow:

Test plan: `yarn unit`
Created many new topics for this, to ensure it paginates at:
https://sourcecred-test.discourse.group/c/filler


src/plugins/discourse/fetch.js
```diff
@@ -323,6 +383,50 @@ function failForNotOk(response: Response) {
...
+ */
+function parseLatestTopic(
+  usernamesById: Map<number, string>,
+  json: any
+): Topic {
+  // TODO: this is a code smell.
+  // Currently we're using a quirk of the API code, that the first poster in the summary
+  // is always the original poster. The only other way to tell is by parsing it from the
+  // translated human description.
+  // See: https://github.com/discourse/discourse/blob/23367e79ea735598766ec6f507f6132e0bad3dba/lib/topic_query.rb#L443
```
Beanow: Has an unresolved comment https://github.com/sourcecred/sourcecred/pull/1452#discussion_r346061717

> Hmm, is there any way for us to write a test that verifies this assumption holds? (that the first poster is the original poster). If that test operates on the snapshotted real data from the server, then it would act as a canary in case Discourse changes the behavior.
>
>Also, what is the TODO here? Do we want to eventually change it to parse the other description?

---
An example of what the API returns here looks like this:
```json
"posters": [
  {
    "extras": "latest",
    "description": "Original Poster, Most Recent Poster",
    "user_id": 2,
    "primary_group_id": null
  },
  {
    "extras": null,
    "description": "Frequent Poster",
    "user_id": 65,
    "primary_group_id": null
  }
]
```

Looking at the code, the only way I found to identify the author of the topic (Original Poster / OP in Discourse terms), is 1. by being the first in the array. 2. by having the "Original Poster" in the description.

The problem with the description is, it's a translated and concatenated string for human reading. We may be able to use this *if* we can reliably force the language to be the one we do a pattern match against. *Or* we can detect the language of the returned descriptions and support the entire range of translations in our pattern matching.

I've opted for option 1 of making the assumption it's the first in the array. However I think both are rather flimsy assumptions and could easily be changed upstream.

A final alternative could be, to not use either detection. Remove the `authorUsername` field from this type, similar to how we removed `bumpedMs` from `TopicView` at the other call. And force ourselves to fetch this data by loading the `TopicView`. As this API call has a very explicit author field at the path `topic.details.created_by.username`.

decentralion: Hmm, interesting, thanks for the explanation. How much inconvenience would it be to take your final suggestion and remove the authorUsername field from this type?

I'm ok with going with this (possibly flimsy) approach for now, so long as we have at least some kind of canary. It could maybe be a snapshot test on our end that scans all of the descriptions for some sentinel topic on our discourse, and fails if the 1st entry is not the original author. But if we can do the robust thing instead, that's great.

Beanow: It should be a good test to see if my latest branching strategy lowers overhead with propagating changes or not. I'll go ahead with removing it from the type.

Beanow: Split up the types and removed `authorUsername` from the snapshot tests.

Beanow:
Rebased against master as parent #1456 was squash merged.
And now has the suggested removal of `authorUsername` from this API call return type.

PR #1458, opened 2019-11-15
Discourse: adds bumpedMsForTopic and topicsInCategories queries

Beanow:
These are to be used by the new mirror code.

Test plan: `yarn unit`
New test cases added.

### DIFF
--- a/src/plugins/discourse/mirrorRepository.js
+++ b/src/plugins/discourse/mirrorRepository.js
@@ -3,7 +3,14 @@
 import type {Database} from "better-sqlite3";
 import stringify from "json-stable-stringify";
 import dedent from "../../util/dedent";
-import type {TopicId, PostId, Topic, Post, LikeAction} from "./fetch";
+import type {
+  CategoryId,
+  TopicId,
+  PostId,
+  Topic,
+  Post,
+  LikeAction,
+} from "./fetch";
 
 // The version should be bumped any time the database schema is changed,
 // so that the cache will be properly invalidated.
@@ -45,6 +52,13 @@ export interface ReadRepository {
    * Gets all of the like actions in the history.
    */
   likes(): $ReadOnlyArray<LikeAction>;
+
+  /**
+   * Finds the TopicIds of topics that have one of the categoryIds as it's category.
+   */
+  topicsInCategories(
+    categoryIds: $ReadOnlyArray<CategoryId>
+  ): $ReadOnlyArray<TopicId>;
 }
 
 export type MaxIds = {|
@@ -63,6 +77,12 @@ export interface MirrorRepository extends ReadRepository {
   addTopic(topic: Topic): AddResult;
   addPost(post: Post): AddResult;
   addLike(like: LikeAction): AddResult;
+
+  /**
+   * For the given topic ID, retrieves the bumpedMs value.
+   * Returns null, when the topic wasn't found.
+   */
+  bumpedMsForTopic(id: TopicId): number | null;
 }
 
 function toAddResult({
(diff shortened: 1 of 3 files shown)


decentralion:
@wchargin, would you mind reviewing this? I don't feel as confident reviewing SQL code.

Beanow:
Rebased as it's dependency #1454 got merged.

Beanow:
`topicsInCategories` looks relevant for initiatives later on and want to move it to the ReadRepository interface for that.

src/plugins/discourse/mirrorRepository.test.js
```diff
@@ -24,4 +25,61 @@ describe("plugins/discourse/mirrorRepository", () => {
...
+    // Given
+    const db = new Database(":memory:");
+    const url = "http://example.com";
+    const repository = new SqliteMirrorRepository(db, url);
+    const topic: Topic = {
+      id: 123,
+      categoryId: 1,
+      title: "Sample topic",
+      timestampMs: 456789,
+      bumpedMs: 456789,
```
decentralion: given that bumpedMs and timestampMs are the same value, this test can't distinguish whether it's returning the bumpedMs or the timestampMs.

src/plugins/discourse/mirrorRepository.test.js
```diff
@@ -24,4 +25,61 @@ describe("plugins/discourse/mirrorRepository", () => {
...
+    const repository = new SqliteMirrorRepository(db, url);
+
+    // When
+    const bumpedMs = repository.bumpedMsForTopic(123);
+
+    // Then
+    expect(bumpedMs).toEqual(null);
+  });
+
+  it("topicsInCategories finds an an existing topic by categoryId", () => {
```
decentralion: would like to see a test where topicsInCategories has an empty list of categoryIds, and where it has more than one categoryId

src/plugins/discourse/mirrorRepository.js
```diff
@@ -45,6 +52,13 @@ export interface ReadRepository {
    * Gets all of the like actions in the history.
    */
   likes(): $ReadOnlyArray<LikeAction>;
+
+  /**
+   * Finds the TopicIds of topics that have one of the categoryIds as it's category.
+   */
+  topicsInCategories(
```
decentralion: Why do we have it takes a list of categoryIds rather than a single one? Intuitively, I would expect it to take a single categoryId and then have callers `map` over it. although having it take an array is fine too (and if we mostly want to select for groups of categories, it's a lot more convenient)

Beanow: The initial use-case is to support the mirror option
`recheckTopicsInCategories: $ReadOnlyArray<CategoryId>` which would save SQL round trips if there were multiple IDs configured. I've now made it part of ReadRepository to use for the initiatives plugin. But that is guaranteed to be just one category ID.
Neither are a hot loop though, so a singular approach would be fine too.

src/plugins/discourse/mirrorRepository.test.js
Beanow: I defaulted to it, as it's what a real world topic would look like with just 1 post. But fair point, so I've changed to having a bump in mind for this and the other tests with a similar issue.

src/plugins/discourse/mirrorRepository.js
Beanow: I'm inclined to leave this, as this results in the least amount of SQL round trips in all situations and the overhead of using it for 1 category is to add that category to an array.

src/plugins/discourse/mirrorRepository.test.js
Beanow: Added a none and multiple categories case.

src/plugins/discourse/mirrorRepository.js
```diff
@@ -293,6 +313,27 @@ export class SqliteMirrorRepository
...
   }
 
+  topicsInCategories(
+    categoryIds: $ReadOnlyArray<CategoryId>
+  ): $ReadOnlyArray<TopicId> {
+    return this._db
+      .prepare(
+        dedent`\
+          SELECT id FROM topics
+          WHERE category_id IN (${categoryIds.map((_) => "?").join(",")})
```
decentralion: This pattern feels a little funky to me--surely there should be a way to put an array in that doesn't require manually creating a query with array.length "?" characters? cc @wchargin. That said, I don't see an actual problem with it, so I'll approve

Beanow: I thought so too. But I've not seen any option to bind arrays with better-sqlite3 in their API.
The flow type also hints at this, being either a dictionary for named parameters or a positional value defined as `number | string | Buffer | null`.

PR #1459, opened 2019-11-15
Discourse: add replaceTopicTransaction method to repo

Beanow:
As the new mirroring will fetch all posts of a topic, inserting them through a DB transaction makes it atomic.

Test plan: `yarn unit`
New test cases added.

### DIFF
--- a/src/plugins/discourse/mirrorRepository.js
+++ b/src/plugins/discourse/mirrorRepository.js
@@ -83,6 +83,14 @@ export interface MirrorRepository extends ReadRepository {
    * Returns null, when the topic wasn't found.
    */
   bumpedMsForTopic(id: TopicId): number | null;
+
+  /**
+   * Idempotent insert/replace of a Topic, including all it's Posts.
+   *
+   * Note: this will insert new posts, update existing posts and delete old posts.
+   * As these are separate queries, we use a transaction here.
+   */
+  replaceTopicTransaction(topic: Topic, posts: $ReadOnlyArray<Post>): void;
 }
 
 function toAddResult({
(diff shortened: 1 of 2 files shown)


Beanow:
> This makes the mirror handle deletions, yes?

Well technically it doesn't change anything about the `Mirror` as it's just adding an unused method to the `MirrorRepository`.

But yes, the previous approach was to do `addTopic` and `addPost` over whatever the mirror update finds. Which never removes anything.

This method however does the `addTopic`, `addPost` and does the new `deleteUnexpectedPosts`. Meaning it will change two things: doing this as a transaction, and deleting posts.

Another side-effect of using this method, is that you can't avoid supporting post edits. As omitting posts would be a delete. The previous mirror update logic tried to omit as many posts as it can, on the assumption it will just add and not remove.

The previous code for a side-by-side.
https://github.com/sourcecred/sourcecred/blob/e79cca6c6c5e0df08357ebf154f4b8712dd19727/src/plugins/discourse/mirror.js#L120-L123

Beanow:
> (Sidenote: I would find it helpful to have more descriptive and comprehensive commit messages. The commit message currently doesn't mention anything about new logic for handling deleted posts)

Can you give me an example of what you'd like to see?
As this is technically a new method, the docstring does explain what it supports, but minus the comparison to the old method.
https://github.com/sourcecred/sourcecred/blob/49a8bb71a5135769bdb0731fa00bbee591c19ba5/src/plugins/discourse/mirrorRepository.js#L80-L86

decentralion:
I gave an example of a more descriptive commit message [here](https://github.com/sourcecred/sourcecred/pull/1458#pullrequestreview-320096805). I think [this](https://github.com/sourcecred/sourcecred/commit/b2943390dcdfa312442cc4de09d79765522b10e8) is also an example of a commit message that, while not particularly long, gives surrounding context for how the change works and how it's been tested.

Beanow:
Updated the commit message.

Beanow:
Rebased as parent commit got squash merged.

PR #1460, opened 2019-11-15
Discourse: add SyncHeads to the repository

Beanow:
This tracks the local state for new mirroring logic.

A bit of context what this `definitionCheckMs` is. In the `MirrorOptions` we provide an interval at which to check category definition topics for changes.

```js
export type MirrorOptions = {|
  // Category definition topics don't show up in the list of bumped topics.
  // We need to proactively check them. This sets the interval at which we
  // should check.
  +recheckCategoryDefinitionsAfterMs: number,

  ...
|};
```

As it doesn't make sense to add something like a `lastChecked` field to Topic, the new table is used as a global variable, when we've last checked these category definition topics.

Test plan: `yarn unit`
Added new test cases.

### DIFF
--- a/src/plugins/discourse/mirrorRepository.js
+++ b/src/plugins/discourse/mirrorRepository.js
@@ -14,7 +14,8 @@ import type {
 
 // The version should be bumped any time the database schema is changed,
 // so that the cache will be properly invalidated.
-const VERSION = "discourse_mirror_v5";
+const VERSION = "discourse_mirror_v6";
+const DEFINITION_CHECK_KEY = "definition_check";
 
 /**
  * An interface for reading the local Discourse data.
@@ -61,6 +62,21 @@ export interface ReadRepository {
   ): $ReadOnlyArray<TopicId>;
 }
 
+export type SyncHeads = {|
+  /**
+   * The timestamp of the last time we've loaded category definition topics.
+   * Used for determining whether we should reload them during an update.
+   */
+  +definitionCheckMs: number,
+
+  /**
+   * The most recent bumpedMs timestamp of all topics.
+   * Used for determining what the most recent topic changes we have stored in
+   * our local mirror, and which we should fetch from the API during update.
+   */
+  +topicBumpMs: number,
+|};
+
 export type MaxIds = {|
   +maxPostId: number,
   +maxTopicId: number,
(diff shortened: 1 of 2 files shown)


src/plugins/discourse/mirrorRepository.js
```diff
@@ -61,6 +62,11 @@ export interface ReadRepository {
   ): $ReadOnlyArray<TopicId>;
 }
 
+export type SyncHeads = {|
+  +definitionCheckMs: number,
```
decentralion: it would be helpful to have a comment explaining what definitionCheckMs signifies. (i assume it's the timestamp when we last did an explicit search for category definition topics, since those aren't returned like normal topics--but the next person reading this code may not have that context fresh in their minds)

src/plugins/discourse/mirrorRepository.js
```diff
@@ -167,6 +184,11 @@ export class SqliteMirrorRepository
     db.prepare("INSERT INTO meta (zero, config) VALUES (0, ?)").run(config);
 
     const tables = [
+      dedent`\
+        CREATE TABLE sync_heads (
+          key TEXT PRIMARY KEY,
```
decentralion: i believe we only ever use the DEFINITION_CHECK_KEY here. in which case, why not use 

`zero INTEGER PRIMARY KEY`

and then we can maintain the invariant that theres only a single record, so we always can get the first one. basically the same thing we do with the `meta` table a few lines above.

src/plugins/discourse/mirrorRepository.test.js
```diff
@@ -48,6 +48,49 @@ describe("plugins/discourse/mirrorRepository", () => {
...
+    // Given
+    const db = new Database(":memory:");
+    const url = "http://example.com";
+    const repository = new SqliteMirrorRepository(db, url);
+    const topic: Topic = {
+      id: 123,
+      categoryId: 1,
+      title: "Sample topic",
+      timestampMs: 456789,
+      bumpedMs: 456789,
```
decentralion: test can't differentiate between the method returning bumpedMs and returning timestampMs.

src/plugins/discourse/mirrorRepository.js
decentralion: I now realize this information is in the commit message. :) Would be nice to copy the same info here.

src/plugins/discourse/mirrorRepository.js
Beanow: It's possible we'll later find more timestamps to track for things the mirror should do periodically. Perhaps like checking for deleted topics. So semantically it makes sense to me to have a key-value store with room for more keys later on.

On the other hand, it would be weird to have multiple meta entries (where this zero primary key is from) as the meta entry has a 1:1 relationship with the full database.

src/plugins/discourse/mirrorRepository.js
Beanow: Added descriptions to the type comments.

src/plugins/discourse/mirrorRepository.test.js
Beanow: Fixed this, same as other tests.

Beanow:
Rebased as parent got squash merged.

PR #1463, opened 2019-11-20
Discourse: scope mirror tests as being "mode 1"

Beanow:
This is to prepare for mode 2 being tested side-by-side.

The normalizeMode1Topics function enforces bumpedMs is not updated for mode 1 tests.
Additionally describe "update semantics" is redundant, as the mirror has no other function than update.

Note, the diff doesn't look too great primarily due to indentation changes not being detected well.

Test plan: `yarn unit`
Existing code should pass the changes made to the tests.

### DIFF
--- a/src/plugins/discourse/mirror.test.js
+++ b/src/plugins/discourse/mirror.test.js
@@ -169,6 +169,14 @@ class MockFetcher implements Discourse {
   }
 }
 
+function normalizeMode1Topics(topics: Topic[]): Topic[] {
+  return topics.map((topic) => ({
+    ...topic,
+    // Mode 1 didn't support bumpMs handling.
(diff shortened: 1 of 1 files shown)


Beanow:
> Note: toggling "ignore whitespace diffs" in the GitHub review settings made the diff a lot cleaner

Oh, I hadn't seen that option before and wouldn't have thought it would fix the original diff having entire blocks misaligned. It definitely helps.

Beanow:
> I'm a little confused by this change, but since it's temporary / migration code and I don't see any concrete issues, going to give it approval.

Right. The alternative was that changing the test structure to this "mode 1 / mode 2" format would be mixed in with #1464. Which already has a substantial change to the test code, thinking this would be easier to review.

PR #1464, opened 2019-11-20
Discourse: add update mode 2

Beanow:
This extends the MockFetcher in the tests as well to provide all new semantics update mode 2 relies on.

It ties together the bulk of [proposed changes](https://github.com/sourcecred/sourcecred/issues/1426) and previous PRs.

- Incrementally (re)load topics based on their bump date.
- Category definition topics (re)loaded at a set interval, as they don't have bump dates.
- Support for `recheckTopicsInCategories` option to always reload categories' contents regardless of a new bump date.
- Load entire topics including posts, to support post editing and deleting. (see #1459)
- Increased efficiency over "mode 1" by not needing to guess topics and posts by ID.

Test plan: `yarn unit`
As a whole test suite is added for "mode 2", however the new mode is not exposed besides tests. Manual testing in a later PR.

### DIFF
--- a/src/plugins/discourse/mirror.js
+++ b/src/plugins/discourse/mirror.js
@@ -1,7 +1,7 @@
 // @flow
 
 import type {TaskReporter} from "../../util/taskReporter";
-import type {Discourse, CategoryId, Topic} from "./fetch";
+import type {Discourse, CategoryId, Topic, TopicLatest} from "./fetch";
 import {MirrorRepository} from "./mirrorRepository";
 
 export type MirrorOptions = {|
@@ -21,6 +21,12 @@ const defaultOptions: MirrorOptions = {
   recheckCategoryDefinitionsAfterMs: 24 * 3600 * 1000, // 24h
   recheckTopicsInCategories: [],
 };
+
+function sortTopicByBumpedMs(a: TopicLatest, b: TopicLatest): number {
+  return a.bumpedMs - b.bumpedMs;
+}
+
+export type MirrorMode = "MODE_1_ITERATE_BY_ID" | "MODE_2_LOAD_BUMPED_TOPICS";
 
 /**
  * Mirrors data from the Discourse API into a local sqlite db.
@@ -44,6 +50,7 @@ export class Mirror {
   +_repo: MirrorRepository;
   +_fetcher: Discourse;
   +_serverUrl: string;
+  +_mode: MirrorMode;
 
   /**
    * Construct a new Mirror instance.
@@ -58,7 +65,8 @@ export class Mirror {
     repo: MirrorRepository,
     fetcher: Discourse,
     serverUrl: string,
-    options?: $Shape<MirrorOptions>
+    options?: $Shape<MirrorOptions>,
+    mode?: MirrorMode
   ) {
     this._repo = repo;
     this._fetcher = fetcher;
@@ -67,13 +75,108 @@ export class Mirror {
       ...defaultOptions,
       ...(options || {}),
     };
+    this._mode = mode || "MODE_1_ITERATE_BY_ID";
   }
 
   async update(reporter: TaskReporter) {
     reporter.start("discourse");
-    await this._updateTopics(reporter);
-    await this._updateLikes(reporter);
+    switch (this._mode) {
+      case "MODE_1_ITERATE_BY_ID":
+        await this._updateTopics(reporter);
+        await this._updateLikes(reporter);
+        break;
+      case "MODE_2_LOAD_BUMPED_TOPICS":
+        await this._updateTopicsV2(reporter);
+        await this._updateLikes(reporter);
+        break;
+      default:
+        (this._mode: empty);
+        throw new Error(`Invalid mode ${this._mode}`);
+    }
     reporter.finish("discourse");
+  }
+
+  async _updateTopicsV2(reporter: TaskReporter) {
+    reporter.start("discourse/topics");
+
+    const {
+      topicBumpMs: lastLocalTopicBumpMs,
+      definitionCheckMs: lastDefinitionCheckMs,
+    } = this._repo.syncHeads();
+
+    const startTime = Date.now();
+    const shouldCheckDefinitions =
+      startTime - lastDefinitionCheckMs >=
+      this._options.recheckCategoryDefinitionsAfterMs;
+
+    const bumpedTopics: TopicLatest[] = await this._fetcher.topicsBumpedSince(
+      lastLocalTopicBumpMs
+    );
+    const topicBumpsById: Map<number, number> = new Map(
+      bumpedTopics.map((t) => [t.id, t.bumpedMs])
+    );
+
+    // Make sure we have oldest first in our load queue.
+    bumpedTopics.sort(sortTopicByBumpedMs);
+
+    // Create a uniqueness filter.
+    const existingTopics = new Set();
+    const once = (tid) => {
+      if (existingTopics.has(tid)) return false;
+      existingTopics.add(tid);
+      return true;
+    };
+
+    // Add definition topics if the flag to do so is set.
+    const definitionTopicIds = shouldCheckDefinitions
+      ? await this._fetcher.categoryDefinitionTopicIds()
+      : [];
+
+    // Add specific-category topics when it's set in our options.
+    const selectCategoryTopics = this._repo.topicsInCategories(
+      this._options.recheckTopicsInCategories
+    );
+
+    // Note: order is important here.
+    // Initial load should happen in order of bump date,
+    // reloads at the end are ok to be in random order.
+    const topicLoadQueue = [
+      ...bumpedTopics.map((t) => t.id),
+      ...definitionTopicIds,
+      ...selectCategoryTopics,
+    ].filter(once);
+
+    for (const topicId of topicLoadQueue) {
+      const topicWithPosts = await this._fetcher.topicWithPosts(topicId);
+      if (topicWithPosts != null) {
+        const {topic, posts} = topicWithPosts;
+
+        // We find the bump by:
+        // 1. What fetcher's topicsBumpedSince tells us.
+        // 2. What the local DB contains (probably force updated a topic that wasn't bumped).
+        // 3. Fall back on creation date (category definition topics don't normally have bump dates at all).
+        const bumpedMs =
+          topicBumpsById.get(topicId) ||
+          this._repo.bumpedMsForTopic(topicId) ||
+          topic.timestampMs;
+        if (bumpedMs == null) {
+          throw new Error(`Missing bump date for topic ID: ${topic.id}`);
+        }
+        const mergedTopic: Topic = {
+          ...topic,
+          bumpedMs,
+        };
+        this._repo.replaceTopicTransaction(mergedTopic, posts);
+      }
+    }
+
+    if (shouldCheckDefinitions) {
+      // Note: use the start time as to avoid missing any changes
+      // between when we queries upstream and when we completed all tasks.
+      this._repo.bumpDefinitionTopicCheck(startTime);
+    }
+
+    reporter.finish("discourse/topics");
   }
 
   async _updateTopics(reporter: TaskReporter) {
--- a/src/plugins/discourse/mirror.test.js
+++ b/src/plugins/discourse/mirror.test.js
@@ -3,7 +3,7 @@
 import {max} from "d3-array";
 import sortBy from "lodash.sortby";
 import Database from "better-sqlite3";
-import {Mirror, type MirrorOptions} from "./mirror";
+import {Mirror, type MirrorOptions, type MirrorMode} from "./mirror";
 import {SqliteMirrorRepository} from "./mirrorRepository";
 import {
   type Discourse,
@@ -376,7 +376,9 @@ describe("plugins/discourse/mirror", () => {
     }
   });
 
-  const example = (optionOverrides?: $Shape<MirrorOptions>) => {
+  const exampleMode = (mode: MirrorMode) => (
+    optionOverrides?: $Shape<MirrorOptions>
+  ) => {
     // Explicitly set all options, so we know what to expect in tests.
     const options: MirrorOptions = {
       recheckCategoryDefinitionsAfterMs: 3600000, // 1h
@@ -386,15 +388,32 @@ describe("plugins/discourse/mirror", () => {
     const db = new Database(":memory:");
     const url = "http://example.com";
     const repo = new SqliteMirrorRepository(db, url);
-    const mirror = new Mirror(repo, fetcher, url, {
-      ...options,
-      ...(optionOverrides || {}),
-    });
+    const mirror = new Mirror(
+      repo,
+      fetcher,
+      url,
+      {
+        ...options,
+        ...(optionOverrides || {}),
+      },
+      mode
+    );
     const reporter = new TestTaskReporter();
     return {fetcher, mirror, reporter, url, repo};
   };
 
+  function expectLikesSorted(as, bs) {
+    const s = (ls) =>
+      sortBy(
+        ls,
+        (x) => x.username,
+        (x) => x.postId
+      );
+    expect(s(as)).toEqual(s(bs));
+  }
+
   describe("mirror mode 1", () => {
+    const example = exampleMode("MODE_1_ITERATE_BY_ID");
     it("mirrors topics from the fetcher", async () => {
       const {mirror, fetcher, reporter, repo} = example();
       const t1 = fetcher.addTopic();
@@ -438,7 +457,7 @@ describe("plugins/discourse/mirror", () => {
       const {mirror, fetcher, reporter, repo} = example();
       const t1 = fetcher.addTopic();
       const t2 = fetcher.addTopic();
-      const [p3] = [fetcher.addPost({topicId: t2.topicId})];
+      const p3 = fetcher.addPost({topicId: t2.topicId});
       await mirror.update(reporter);
       const posts = [
         fetcher._post(t1.postId),
@@ -488,16 +507,6 @@ describe("plugins/discourse/mirror", () => {
       await mirror.update(reporter);
       expect([...repo.users()].sort()).toEqual(["alpha", "beta", "delta"]);
     });
-
-    function expectLikesSorted(as, bs) {
-      const s = (ls) =>
-        sortBy(
-          ls,
-          (x) => x.username,
-          (x) => x.postId
-        );
-      expect(s(as)).toEqual(s(bs));
-    }
 
     it("provides all the likes by users that have posted", async () => {
       const {mirror, fetcher, reporter, repo} = example();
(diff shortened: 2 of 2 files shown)


src/plugins/discourse/mirror.js
```diff
@@ -22,6 +22,10 @@ const defaultOptions: MirrorOptions = {
   recheckTopicsInCategories: [],
 };
 
+function sortTopicByBumpedMs(a: TopicLatest, b: TopicLatest): number {
```
decentralion: Can be removed if we use `lodash.sortby` (see comment below)

src/plugins/discourse/mirror.js
```diff
@@ -67,15 +73,109 @@ export class Mirror {
...
+      case 1:
+        await this._updateTopics(reporter);
+        await this._updateLikes(reporter);
+        break;
+      case 2:
+        await this._updateTopicsV2(reporter);
+        await this._updateLikes(reporter);
+        break;
+      default:
+        throw new Error(`Invalid mode ${this._mode}, expecting 1 or 2`);
```
decentralion: If we switch mode to an enum, then we can do ```throw new Error(`Invalid mode: ${this._mode: empty}`)``` and then flow will correctly type error if we ever add a new mode, but don't handle it.

src/plugins/discourse/mirror.test.js
```diff
@@ -88,27 +90,56 @@ class MockFetcher implements Discourse {
     return matchingLikes.slice(offset, offset + CHUNK_SIZE);
   }
 
+  async topicsBumpedSince(sinceMs: number): Promise<TopicLatest[]> {
+    return Array.from(this._topicToPostIds.keys())
+      .filter((tid) => this._topicCategories.get(tid) != tid)
```
decentralion: I'm confused by this. If topicCategory is a map from topicId to categoryId, why would the topicId and categoryId be the same?

src/plugins/discourse/mirror.test.js
```diff
@@ -132,6 +164,20 @@ class MockFetcher implements Discourse {
     };
   }
 
+  addCategory(topicId: TopicId): PostId {
```
decentralion: What does it mean to "add" a category? And why does it return a postId? I am quite confused.

src/plugins/discourse/mirror.test.js
```diff
@@ -132,6 +164,20 @@ class MockFetcher implements Discourse {
...
+    const postId = this.addPost(topicId, null);
+    this._topicCategories.set(topicId, topicId);
```
decentralion: The topicCategories is supposed to map to categoryIds...why does it get set to a topicId here?

src/plugins/discourse/mirror.test.js
```diff
@@ -154,11 +200,42 @@ class MockFetcher implements Discourse {
       topicId: topicId,
       authorUsername: NullUtil.orElse(username, "credbot"),
       cooked: NullUtil.orElse(cooked, "<h1>Hello World</h1>"),
+      timestampMs: 2000 + postId,
     };
     this._posts.set(postId, postInfo);
     return postId;
   }
 
+  editPost(postId: PostId, changes: $Shape<PostInfo>): void {
```
decentralion: Not asking for a change at this point, but would have preferred to have all these updates to test code in a separate commit. (Add the new test methods, then use them to test the new update mode.)

src/plugins/discourse/mirror.test.js
```diff
@@ -339,13 +426,12 @@ describe("plugins/discourse/mirror", () => {
 
       const postsFromTopic = NullUtil.get(await fetcher.topicWithPosts(1))
         .posts;
-      expect(postsFromTopic.map(getId)).toEqual([p1]);
+      expect(postsFromTopic.map(getId)).toEqual([p1, p2, p3]);
```
decentralion: why do the mode 1 tests change?

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+      await mirror.update(reporter);
+      const actualAfter = repo.posts();
+
+      expect(actualBefore).toEqual(expectedBefore);
+      expect(actualAfter).toEqual(expectedAfter);
+      // Sanity check, make sure the category post was edited in the mock fetcher.
+      expect(fetcher._post(p2)).not.toEqual(expectedAfter[1]);
+    });
+
+    it("provides usernames for all active users", async () => {
```
decentralion: what would happen if a user only posted in a category definition topic?

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+
+    it("doesn't find likes of users that never posted", async () => {
+      const {mirror, fetcher, reporter, repo} = example();
+      fetcher.addPost(1, null);
+      fetcher.addLike("nope", 1, 1);
+      await mirror.update(reporter);
+      expect(repo.likes()).toEqual([]);
+    });
+
+    it("mirrors category definition topics only after it's interval", async () => {
```
decentralion: this test description makes me think that the test verifies that the updated category definitions will actually get discovered (after enough time has passed), but that doesn't happen here.

decentralion: maybe we should mock Date.now() so we can fully test the interval logic

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+      const p5 = fetcher.addPost(5, null);
+      fetcher.setTopicCategory(5, 4);
+
+      // See we're trying to update #4 and #5.
+      const topicsCall = jest.spyOn(fetcher, "topicWithPosts");
+      await mirror.update(reporter);
+
+      fetcher.editPost(p1, {cooked});
+      fetcher.editPost(p5, {cooked});
+      const earlyCalls = [...topicsCall.mock.calls];
```
decentralion: (personally, I would use topicsCalls.mock.calls.slice(), but this is fine too)

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+      expect(laterTopics).toEqual([topic4, topic5]);
+    });
+
+    it("mirrors respects recheckTopicsInCategories option", async () => {
+      const {mirror, fetcher, reporter, repo} = example({
+        recheckTopicsInCategories: [4],
+      });
+      const cooked = "<p>New content, without bump</p>";
+
+      // Adds a category description #4.
```
decentralion: Should we also have a category description which is *not* getting rechecked?

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+      expect(fetcher._topic(4)).toMatchObject({id: 4, categoryId: 4});
+      expect(fetcher._topic(5)).toMatchObject({id: 5, categoryId: 4});
+      expect(fetcher._post(p5)).toMatchObject({id: 5, cooked});
+      // The initial load has strict order requirements.
+      expect(earlyCalls).toEqual([[1], [2], [3], [5], [4]]);
+      // The update does not.
+      expect(topicsCall).toHaveBeenCalledTimes(2);
+      expect(topicsCall).toHaveBeenCalledWith(4);
+      expect(topicsCall).toHaveBeenCalledWith(5);
+      expect(laterPosts).toMatchInlineSnapshot(`
```
decentralion: Why do we keep this inline snapshot? Is it to verify that the updated version was retrieved? If so, wouldn't it be cleaner to just test those posts directly -- easier for people making future modifications to understand what the intended behavior is.

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+      const topicsCall = jest.spyOn(fetcher, "topicWithPosts");
+      await mirror.update(reporter);
+      const earlyCalls = [...topicsCall.mock.calls];
+      topicsCall.mockClear();
+      await mirror.update(reporter);
+
+      // The initial load has strict order requirements.
+      expect(earlyCalls).toEqual([[1], [2]]);
+      // The update does not.
+      expect(topicsCall).toHaveBeenCalledTimes(2);
```
decentralion: Not readily clear to me what this test is checking.

src/plugins/discourse/mirror.test.js
```diff
@@ -568,4 +654,548 @@ describe("plugins/discourse/mirror", () => {
...
+      // #1 will be updated because of the new post bump.
+      // #4 will be updated because of the recheck of category.
+      // #5 will be updated because of the new post bump and recheck of category.
+      expect(topicsCall).toHaveBeenCalledTimes(3);
+      expect(topicsCall).toHaveBeenCalledWith(1);
+      expect(topicsCall).toHaveBeenCalledWith(4);
+      expect(topicsCall).toHaveBeenCalledWith(5);
+    });
+
+    it("only fetches new topics on `update`", async () => {
```
decentralion: You say it only fetches new topics... but if we had added a post to topic 1, the topic would not be "new", but it would get fetched, I think? So should the test description be "only fetches new or bumped topics"?

src/plugins/discourse/mirror.js
```diff
@@ -67,15 +73,109 @@ export class Mirror {
...
+
+    const bumpedTopics: TopicLatest[] = await this._fetcher.topicsBumpedSince(
+      lastLocalTopicBumpMs
+    );
+    const topicBumpsById: Map<number, number> = new Map(
+      bumpedTopics.map((t) => [t.id, t.bumpedMs])
+    );
+
+    // Make sure we have oldest first in our load queue.
+    bumpedTopics.sort(sortTopicByBumpedMs);
```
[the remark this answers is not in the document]
Beanow: Hm, reason I tend to avoid this is because it's a [2630 line dependency](https://github.com/lodash/lodash/blob/4.7.0-npm-packages/lodash.sortby/index.js) to abstract away a 3 line comparison function and shallow copy instead of mutate. Although I'll agree that comparison functions require a bit more working memory to reason about, I think that's much alleviated by not inlining the comparison function.

src/plugins/discourse/mirror.js
```diff
@@ -67,15 +73,109 @@ export class Mirror {
...
+    );
+
+    // Note: order is important here.
+    // Initial load should happen in order of bump date,
+    // reloads at the end are ok to be in random order.
+    const topicLoadQueue = [
+      ...bumpedTopics.map((t) => t.id),
+      ...definitionTopicIds,
+      ...selectCategoryTopics,
+    ].filter(once);
```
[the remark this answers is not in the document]
Beanow: Yes, I ended up with this because sorting is preserved, Set doesn't necessarily preserve order.

src/plugins/discourse/mirror.test.js
Beanow: Although the process of _discovering_ category definition topics is different. The underlying structure is still a topic with posts which have authors. So these users will be included. Keeping in mind the (re)load interval of category definition topics.

src/plugins/discourse/mirror.js
```diff
@@ -67,15 +73,109 @@ export class Mirror {
...
+
+    for (const topicId of topicLoadQueue) {
+      const topicWithPosts = await this._fetcher.topicWithPosts(topicId);
+      if (topicWithPosts != null) {
```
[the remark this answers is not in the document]
Beanow: The fetcher returns a null when the API returns a 403, 404 or 410 status.
This typically happens when topics are deleted or made private.
(In mode 1, it would also happen because we were ID guessing)

We might consider deleting the topic from the mirror when this happens, but as this wouldn't be enough to implement topic deletion correctly, we're ignoring it instead.

As you can never really exclude a race condition, I don't think we should warn or crash here.
For example, one might assume a topic with this ID must exist, because we found it in the list of recently bumped topics. But if we have 200 topics to fetch that still takes several minutes to do and topics can be deleted in that time.
More likely though, the `recheckTopicsInCategories` option finds it's topic IDs in the local database, so they could be weeks old and refer to a topic that's been deleted.

src/plugins/discourse/mirror.test.js
Beanow: Wouldn't hurt :man_shrugging:. But the expected behavior of those are tested elsewhere.
The category definition topics that aren't in this list should update when `recheckCategoryDefinitionsAfterMs` has elapsed.
The normal topics in a category should be updated when they're bumped.

src/plugins/discourse/mirror.test.js
Beanow: I find spreading into a new array literal a bit more obvious to be used as a shallow copy than `.slice()`. Especially because array slices are a concept in other languages that don't make shallow copies. Such as Go.

https://blog.golang.org/go-slices-usage-and-internals#TOC_4.
>  A slice is a descriptor of an array segment. It consists of a pointer to the array, the length of the segment, and its capacity (the maximum length of the segment).

src/plugins/discourse/mirror.test.js
Beanow: The fetcher should be called exactly twice in the second update.
Because there should be exactly two topics discovered in the first update that are in the category we're forcing updates of the second time.

Without this assert it could be more than 2, as the below checks only assert at least those two distinct calls have been made. If the actual calls were: `[1, 1, 2, 1, 2, 2, 44]` it would pass.
```
      expect(topicsCall).toHaveBeenCalledWith(1);
      expect(topicsCall).toHaveBeenCalledWith(2);
```

src/plugins/discourse/mirror.test.js
Beanow: You're right, this is a copy-pasted test from mode 1 and could be misinterpreted then as well, as new posts should be included too, which would be a bumped topic.

Reflecting what this test does, it("should not fetch other topics again when only a new topic was added between updates").

It isn't to say that new topics are the only type of topic the update function should look for. But that the scenario here is adding a new topic between updates and we're expecting update to detect it can leverage local data to only need to fetch that new topic.

src/plugins/discourse/mirror.test.js
Beanow: That the update was retrieved for one, but not the other. Yes, I can update to directly testing if I keep a copy of pre-edit P1.

src/plugins/discourse/mirror.test.js
Beanow: We could mock time to test that. I've done so before in another project that had several critically important functions that were time related. Such as requiring different behavior depending on the hour-of-day, timezone, date, day-of-week, etc.

Mocking this correctly can be tricky though. From that, I would recommend not mocking Date.now().
But instead adding a dependency to the Mirror constructor, **and all if it's dependencies** from which to source the current time. For example a `getDate: () => Date` function. Forcing ourselves to use that dependency as the source of time in reviews. Basically all across the code base. Reason being, there are other ways to get time that we might not have mocked. `new Date()`, `process.hrtime()`, timestamps created by the database and probably more. Not doing something this rigorously can lead to flaky tests due to the mock only covering part of the time sources, causing unexpected false-positives and false-negatives.

Additionally, bringing in a library like [luxon](https://www.npmjs.com/package/luxon) to make those tests legible, instead of having millisecond time math all over the place.

Right now, I'm dodging having to do that by making use of the fact that 0 ms is a valid interval and it will be executed on the next update even if the time between update calls is <1 ms, while the default test value of `recheckCategoryDefinitionsAfterMs: 3600000, // 1h` is safe to assume won't elapse in a single test case. Although we could use values up to `+Infinity` as well to guarantee this.

The only drawback being, we're not doing automated testing with likely real-world scenarios. Like 2 days between updates or 40 minutes between updates. Which seems like a smoke test to me. Not sure it's worth refactoring for that right now.

src/plugins/discourse/mirror.test.js
Beanow: Because the MockFetcher changed. It's implementation details of `topicWithPosts` are now consistent with #1455 behavior change to the real fetcher. Along with the other changes to MockFetcher it probably warrants still splitting this PR.

src/plugins/discourse/mirror.test.js
Beanow: Hah, right. I wrote up a reply of how that happened. The gist being it builds on the previous MockFetcher design where Topics were created almost entirely by convention. But this just makes it worse :sweat_smile:.

How about, I change the below function to be:
`setTopicCategory(topicId: TopicId, categoryId: CategoryId | null, isDefinition: boolean)`

So you can do:
```js
// Create Topic ID 4, with a post.
const p1 = addPost(4, null);
// Set Topic ID 4 as the category definition topic for Category ID 6.
setTopicCategory(4, 6, true);

// Create Topic ID 5 with a post.
const p2 = addPost(5, null);
// Set Topic ID 5 as a normal topic in Category ID 6.
setTopicCategory(4, 6, false);
```

Beanow: It is confusing. It comes from MockFetcher generating Topics implicitly, by calling `addPost` with a new TopicId.
This means, TopicIds are planned, but created through Posts. PostIds are incremental and returned by `addPost`.

The change here introduces new concepts. Categories and category definition topics.
Categories, like Topics are also generated implicitly.
And Topics are:
- Uncategorized when not in the `_topicCategories` mapping.
- Category definition topics when, TopicId == CategoryId.
- Normal topics in a category when, TopicId != CategoryId.

The way to create a topic is by adding it's first post.
```js
// Creates topic: 4, with post: p1.
const p1 = addPost(4, null);
```

The way to have a topic in a category normally:
```js
// Adds that topic to category ID 6.
setTopicCategory(4, 6);
```

But a category should have a category definition topic.
One way to do so, using just the convention of "Category definition topics when, TopicId == CategoryId"
```js
// Creates topic: 6, with post: p2.
const p2 = addPost(6, null);
// Adds that topic to category ID 6, which makes it a category definition topic.
setTopicCategory(6, 6);
```

The functional equivalent of that is:
```js
// Creates topic: 6, with post: p2.
// And adds the topic to category ID 6, which makes it a category definition topic.
const p2 = addCategory(6);
```

---

But that certainly doesn't make anything any clearer.
Perhaps the process should be inverted. So the signature is:
`defineCategory(topicId: TopicId): CategoryId`

Throwing if it's called with a topicId that doesn't exist in the MockFetcher yet.
Doing away with the TopicId == CategoryId convention.

It would look something like:
```js
// Create topic: 5, with post: p1.
const p1 = addPost(5, null);
// Create category: c1, with definition topic: 5.
const c1 = defineCategory(5);

// Create topic: 7, with post: p2.
const p2 = addPost(7, null);
// Add it to category: c1.
setTopicCategory(7, c1);
```

src/plugins/discourse/mirror.test.js
Beanow: I've renamed this test (and in mode 1 in #1473)

src/plugins/discourse/mirror.test.js
Beanow: Now uses `toEqual` with an array of posts.

src/plugins/discourse/mirror.js
Beanow: Changed to an enum.

src/plugins/discourse/mirror.test.js
Beanow: I've removed this logic, see #1473

src/plugins/discourse/mirror.js
```diff
@@ -44,6 +48,7 @@ export class Mirror {
   +_repo: MirrorRepository;
   +_fetcher: Discourse;
   +_serverUrl: string;
+  +_mode: number;
```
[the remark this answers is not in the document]
Beanow: Changed to an enum.

Beanow:
@decentralion I've moved the changed to MockFetcher to it's own PR #1473 
Hoped that most of the comments of this PR would be preserved by changing the base to #1473's branch. Thankfully that seems to be the case :smile:

src/plugins/discourse/mirror.test.js
```diff
@@ -868,4 +877,587 @@ describe("plugins/discourse/mirror", () => {
...
+        fetcher._topic(c1.topicId),
+        fetcher._topic(t2.topicId),
+      ]);
+    });
+
+    it("mirrors posts from the fetcher", async () => {
+      const {mirror, fetcher, reporter, repo} = example();
+      const t1 = fetcher.addTopic();
+      const t2 = fetcher.addTopic();
+      const [p3] = [fetcher.addPost({topicId: t2.topicId})];
```
decentralion: nit: the array wrapping on both sides can be removed

src/plugins/discourse/mirror.test.js
```diff
@@ -868,4 +877,587 @@ describe("plugins/discourse/mirror", () => {
...
+      expect(repo.posts()).toEqual(posts);
+    });
+
+    it("mirrors edited posts from the fetcher", async () => {
+      const {mirror, fetcher, reporter, repo} = example();
+      const t1 = fetcher.addTopic();
+      const t2 = fetcher.addTopic();
+      const p3 = fetcher.addPost({topicId: t2.topicId});
+      fetcher.editPost(p3, {cooked: "<p>Here's some edited content.</p>"});
+      await mirror.update(reporter);
```
decentralion: what would happen if the mirror updates, and then we edit the post, and then we update the mirror again? can we have a test for this case?

src/plugins/discourse/mirror.test.js
```diff
@@ -868,4 +877,587 @@ describe("plugins/discourse/mirror", () => {
...
+      const actualAfter = repo.posts();
+
+      expect(actualBefore).toEqual(expectedBefore);
+      expect(actualAfter).toEqual(expectedAfter);
+      // Sanity check, make sure the category post was edited in the mock fetcher.
+      expect(fetcher._post(c2.postId)).not.toEqual(expectedAfter[1]);
+    });
+
+    it("mirrors remainder when a post is deleted from the fetcher", async () => {
```
decentralion: "mirrors remainder" -- what does remainder mean in this case?

src/plugins/discourse/mirror.test.js
decentralion: makes sense.

src/plugins/discourse/mirror.js
```diff
@@ -67,15 +75,109 @@ export class Mirror {
...
+      case "MODE_1_ITERATE_BY_ID":
+        await this._updateTopics(reporter);
+        await this._updateLikes(reporter);
+        break;
+      case "MODE_2_LOAD_BUMPED_TOPICS":
+        await this._updateTopicsV2(reporter);
+        await this._updateLikes(reporter);
+        break;
+      default:
+        throw new Error(`Invalid mode ${this._mode}`);
```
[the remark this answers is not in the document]
Beanow: So, as I understand this is a type cast operation in Flow. How does this improve the errors?
My expectation would be that assigning the variable or calling the constructor with something not matching the enum would trip a type error already?

src/plugins/discourse/mirror.test.js
Beanow: Content exposed by the fetcher which was not deleted.
The remainder after deleting a post from the fetcher.

src/plugins/discourse/mirror.js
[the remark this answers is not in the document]
decentralion: Sorry, I miswrote. I meant to say: `this.mode: empty`. Asserting that `this.mode` is empty in the default case means that flow will error if we ever add another (uncovered) case to the enum.

Here's an in-depth explanation: https://medium.com/@ibosz/advance-flow-type-1-exhaustive-checking-with-empty-type-a02e503cd3a0

src/plugins/discourse/mirror.test.js
Beanow: Resolved by moving to #1473

src/plugins/discourse/mirror.js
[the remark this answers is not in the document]
Beanow: Got it :+1: it's to ensure all enum options have a case associated with it.
Have a slight preference for this being on it's own line like:
```js
      default:
        (this._mode: empty);
        throw new Error(`Invalid mode ${this._mode}`);
```

But yeah, it should be deleted pretty soon after this.

src/plugins/discourse/mirror.test.js
Beanow: Oops, yeah missed that.

src/plugins/discourse/mirror.test.js
Beanow: This is covered, for example in "mirrors updates normal post edits, not category definition post edits".

https://github.com/sourcecred/sourcecred/blob/c12b148ae5b4805a0d084b754fb71f79678db7a8/src/plugins/discourse/mirror.test.js#L991-L1029

decentralion:
:+1: Merge at your convenience.

src/plugins/discourse/mirror.js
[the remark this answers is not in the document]
Beanow: For `lodash.sortby` I'd like to have a go at replacing it with a small unit function. But since this is definitely not the only place where it would need to be used, I think that is best done as a separate PR. For now I feel this solution is way more comprehensible than the dependency.

PR #1473, opened 2019-11-29
Discourse: make MockFetcher API similar to real forum

Beanow:
This change came from #1464, which initially had a lot of these changes to the MockFetcher included. But should have been it's own commit. It's now split to here.

This extends the MockFetcher in the tests to provide new semantics update mode 2 relies on.
They're based on the below changes to the Fetcher:

- add categoryId and bumpedMs to Topic data #1454
- make topicWithPosts fetch all posts #1455
- add categoryDefinitionTopicIds to fetcher #1456
- implement topicsBumpedSince in fetcher #1457

Particularly because the addition of two new concepts (categories and category definition topics), the API of the MockFetcher got rather convoluted. This refactor makes it behave a lot more like you'd be familiar with within Discourse.

Such as, creating a topic creates it's opening post as a side effect. Instead of a post with an unknown topic ID creating a topic as a side effect.
And creating a category creates it's category definition topic as a side effect.

Also, we're being a lot more explicit, using objects instead of positional arguments.

Test plan: `yarn unit` as it only covers changes to unit testing code.

### DIFF
--- a/src/plugins/discourse/mirror.test.js
+++ b/src/plugins/discourse/mirror.test.js
@@ -1,11 +1,13 @@
 // @flow
 
+import {max} from "d3-array";
 import sortBy from "lodash.sortby";
 import Database from "better-sqlite3";
 import {Mirror, type MirrorOptions} from "./mirror";
 import {SqliteMirrorRepository} from "./mirrorRepository";
 import {
   type Discourse,
+  type CategoryId,
   type TopicId,
   type PostId,
   type Topic,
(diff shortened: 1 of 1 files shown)


src/plugins/discourse/mirror.test.js
```diff
@@ -89,19 +138,48 @@ class MockFetcher implements Discourse {
...
+    const view = this._topicView(id);
+    const posts = (this._topicToPostIds.get(id) || []).map((pid) =>
+      this._post(pid)
+    );
+    let bumpedMs = view.timestampMs;
+
+    // Category definition topics don't support bumping, as they are never
+    // in any API responses that contain this field.
+    if (!this._categoryDefinitionTopics.has(id)) {
+      bumpedMs = posts.reduce(
```
decentralion: Since we already depend on d3-array, I would personally go for:

```js
import {max} from "d3-array";

bumpedMs = max(posts, (p) => p.timestampMs)
```

Implementation is [quite reasonable](https://github.com/d3/d3-array/blob/master/src/max.js) at 20 LOC.

src/plugins/discourse/mirror.test.js
```diff
@@ -110,63 +188,158 @@ class MockFetcher implements Discourse {
...
-      return null;
+    return this._posts.get(id) || null;
+  }
+
+  /**
+   * Adds a new Category.
+   * This will create a category definition topic as well.
+   * The definition topic will create an opening post.
+   *
+   * Note: does not have any perception of subforums.
```
decentralion: By subforum do you mean sub-category relationships?

src/plugins/discourse/mirror.test.js
```diff
@@ -110,63 +188,158 @@ class MockFetcher implements Discourse {
...
+   */
+  addCategory(topic?: $Shape<TopicOptions>): CreatedCategory {
+    // Have the new Category exist.
+    const categoryId = this._nextCategoryId;
```
decentralion: `const categoryId = this._nextCategoryId++` is sufficient (though `const categoryId = ++this._nextCategoryId` would have the wrong behavior). that said, your way is unambiguously correct and the concise way requires a little more familiarity with the operator, so no change requested.

src/plugins/discourse/mirror.test.js
Beanow: I did mean sub-categories yes. For the current fetcher APIs there's no distinction between them. Only the _real_ `categoryDefinitionTopicIds()` implementation needs to know about it and abstracts the differences away for the mirror.

src/plugins/discourse/mirror.test.js
Beanow: Right, that was my intention here. I'm aware of `var++` and `++var` syntax. But putting them on separate lines makes it a lot easier to see these are two semantic operations and what the intended order is.

src/plugins/discourse/mirror.test.js
```diff
@@ -19,61 +20,109 @@ import * as MapUtil from "../../util/map";
 import * as NullUtil from "../../util/null";
 import {TestTaskReporter} from "../../util/taskReporter";
 
-type PostInfo = {|
-  +indexWithinTopic: number,
-  +replyToPostIndex: number | null,
-  +topicId: number,
-  +authorUsername: string,
-  +cooked: string,
+type TopicOptions = {|
```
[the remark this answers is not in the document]
Beanow: From a technical perspective it's a little odd to do this at the type vs the addX functions, but from the way you'd familiarize this makes sense. So I'll add it there.

src/plugins/discourse/mirror.test.js
Beanow: Updated the comment here.

src/plugins/discourse/mirror.test.js
Beanow: Using max now to do the max + pluck.
Checking for null is now done in the lines above.

src/plugins/discourse/mirror.test.js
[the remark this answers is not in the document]
Beanow: Added these comments.

decentralion:
:+1: Merge at your convenience.

PR #1477, opened 2019-12-07
Initiatives: implement basic createGraph

Beanow:
Depends on #1416 and #1417
Now depends on #1508 as well (reference detector interface).

With feedback from https://github.com/sourcecred/sourcecred/commit/8adce788c12145ab2396d82c4483ac600d021a33

Provides a core feature for [Initiatives](https://discourse.sourcecred.io/t/write-the-initiatives-plugin/269).
Along with tests to define it's behavior.

```js
function createGraph(repo: InitiativeRepository, refs: ReferenceDetector): Graph
```

### DIFF
--- a/src/plugins/initiatives/createGraph.js
+++ b/src/plugins/initiatives/createGraph.js
@@ -1,0 +1,98 @@
+// @flow
+
+import {
+  Graph,
+  EdgeAddress,
+  NodeAddress,
+  type Edge,
+  type Node,
+  type EdgeAddressT,
+  type NodeAddressT,
+} from "../../core/graph";
+import type {ReferenceDetector, URL} from "../../core/references";
+import type {Initiative, InitiativeRepository} from "./initiative";
+import {
+  initiativeNodeType,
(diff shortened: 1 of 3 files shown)


src/plugins/initiatives/createGraph.js
```diff
@@ -0,0 +1,110 @@
...
+ * A service which provides reference detection features.
+ */
+export interface ReferenceDetector {
+  /**
+   * Tries to infer the node address from an absolute URL.
+   * Returning undefined when the detector isn't aware of how to resolve this URL.
+   * Note: the detector shouldn't attempt to detect whether the node actually exists.
+   */
+  addressFromUrl(url: URL): ?NodeAddressT;
+}
```
Beanow: This TODO now has it's own PR #1508
So I plan to revise this PR to depend on it.

Beanow: Did exactly that, depending on the core reference detector interface now.

PR #1478, opened 2019-12-07
Initiatives: implement "cooked HTML" template parsing

Beanow:
Depends on #1417
 This plus the #1480 PR replace the previous #1418.

Will return a partial Initiative, or throw when the template isn't matched.

Based on the template that's currently used on the forums:
https://discourse.sourcecred.io/t/about-the-initiatives-category/242


Beanow:
Missed the "considers blank status incomplete" test from previous code.

Beanow:
@decentralion @wchargin something weird going on with sharness here.
The `":(exclude,top)src/plugins/initiatives/htmlTemplate.test.js"` exception is required to make a local `yarn sharness` or `yarn test --full` succeed. But the CI test passes without it.

Beanow:
Switched the DOM parsing to be synchronous. As the implementation is also synchronous, just supports streaming data if needed.

PR #1480, opened 2019-12-08
Initiatives: create Initiative from discourse topics

Beanow:
Depends on #1478.
This plus the #1478 PR replace the previous #1418.

This adds the Discourse specific semantics on top of the basic template parsing.

Test plan: `yarn unit`
No usage available for manual tests.

### DIFF
--- a/src/plugins/initiatives/discourse.js
+++ b/src/plugins/initiatives/discourse.js
@@ -1,0 +1,85 @@
+// @flow
+
+import type {Topic, Post, TopicId} from "../discourse/fetch";
+import type {Initiative, URL} from "./initiative";
+import type {HtmlTemplateInitiativePartial} from "./htmlTemplate";
+import {topicAddress} from "../discourse/address";
+
+/**
+ * Uses data from a Discourse Topic to create an Initiative representation.
+ *
+ * The Post should be the opening post of the Topic.
+ * The Post body should adhere to the `parseCookedHtml` expected template.
+ */
+export function initiativeFromDiscourseTracker(
+  serverUrl: string,
+  topic: Topic,
+  openingPost: Post,
+  parseCookedHtml: (cookedHTML: string) => HtmlTemplateInitiativePartial
+): Initiative {
+  if (serverUrl.endsWith("/")) {
+    throw new Error("serverUrl shouldn't end with trailing slash");
+  }
+
+  const {title} = topic;
+  const {timestampMs} = openingPost;
+  try {
+    if (openingPost.topicId !== topic.id) {
+      throw new Error(`Post ${openingPost.id} is from a different topic`);
+    }
+
+    if (openingPost.indexWithinTopic !== 1) {
+      throw new Error(
+        `Post ${openingPost.id} is not the first post in the topic`
+      );
+    }
+
+    const tracker = topicAddress(serverUrl, topic.id);
+    const partial = parseCookedHtml(openingPost.cooked);
+    return {
+      title,
+      tracker,
+      timestampMs,
+      completed: partial.completed,
+      dependencies: absoluteURLs(serverUrl, partial.dependencies),
+      references: absoluteURLs(serverUrl, partial.references),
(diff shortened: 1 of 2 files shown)


src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,72 @@
...
+      timestampMs,
+      completed: partial.completed,
+      dependencies: absoluteURLs(serverUrl, partial.dependencies),
+      references: absoluteURLs(serverUrl, partial.references),
+      contributions: absoluteURLs(serverUrl, partial.contributions),
+      champions: absoluteURLs(serverUrl, partial.champions),
+    };
+  } catch (e) {
+    // To make solving issues easier, add which initiative topic caused the problem.
+    e.message = `${e.message} for initiative topic "${title}" (${topic.id})`;
```
decentralion: i think putting the url of the topic is even more helpful than the title and id

src/plugins/initiatives/discourse.test.js
```diff
@@ -0,0 +1,263 @@
...
+  type HtmlTemplateInitiativePartial,
+} from "./htmlTemplate";
+import {initiativeFromDiscourseTracker} from "./discourse";
+import type {Topic, Post} from "../discourse/fetch";
+import {NodeAddress} from "../../core/graph";
+
+jest.mock("./htmlTemplate");
+const spyParseCookedHtml = ((parseCookedHtml: any): JestMockFn<any, any>);
+
+function givenParseError(message: string) {
```
decentralion: You could simplify the testing if we split `initiativeFromDiscourseTracker` into two functions, a wrapper and an internal. the wrapper does the most basic validation (check serverUrl), and calls `parseCookedHTML` on the post's cookedHtml, and passes the results into the internal method.

then, to test the internals, we don't need to mock parseCookedHTML, we can test the internal method directly on the expected inputs. and the wrapper, we only need to check that it calls parseCookedHTML faithfully (which IMO flow gets for us, i don't think we even need to unit test it), and we test the validation logic on the wrapper.

what do you think of this change? i don't think it's _necessary_, but it would be cleaner.

another approach could be: write a method that, given the intended initiative partial template, produces the raw HTML to put in the post. then the test would be more "end-to-end". i don't think this is the best approach because of the complexity it adds to the test code, but it's interesting to think about.

src/plugins/initiatives/discourse.js
Beanow: I did find myself wanting to open these up from the terminal when I went back to edit wikis that triggered errors. Would be good to have urls.

src/plugins/initiatives/discourse.test.js
Beanow: Sure, though instead of splitting the implementation, I would then pass `parseCookedHtml` as a function argument, curried or as a constructor argument for a class. We can create a flow type for the function signature of `parseCookedHtml`. I was wondering where the limit was for getting utility from interfaces vs just mocking the one usage. :smile:

decentralion: Yeah. For the record, I"m also OK with the current implementation, but I think making it part of the interface somehow would be cleaner. :)

Beanow: I've updated this to take `parseCookedHtml` as an argument instead of mocking the module.

src/plugins/initiatives/discourse.js
Beanow: Now using URLs

src/plugins/initiatives/discourse.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+function absoluteURLs(
+  serverUrl: string,
+  urls: $ReadOnlyArray<URL>
+): $ReadOnlyArray<URL> {
+  return urls.map((url) => {
+    if (url.startsWith("/")) {
+      return `${serverUrl}${url}`;
+    }
+
```
decentralion: teensy nit: i would remove the linebreak so the whole function block scans as a single paragraph. change not needed.

PR #1483, opened 2019-12-08
Initiatives: implement DiscourseInitiativeRepository

Beanow:
Depends on #1477 and #1480.

Uses the Discourse mirror data and #1480 parsing function to create an `InitiativeRepository` implementation. Which #1477 can use for `createGraph`.

Test plan: `yarn unit`
No usage other than in tests yet.


Beanow:
Based on feedback in https://github.com/sourcecred/sourcecred/pull/1480 this now has the option to override the `parseCookedHtml` implementation as an option. Defaulting to the `parseCookedHtml` implementation we previously had a hard-coded call to.

PR #1486, opened 2019-12-09
GitHub: assume installation token length of 40

### LINKED ISSUE #1474: GitHub installation access token length
With #1461 / #1471, the GitHub installation access token were added.
[The docs](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation) suggested a length of 16 `"token": "v1.1f699f1069f60xxx",` assuming the `xxx`s were just masking. But the actual length seems longer.

As seen [in this run](https://github.com/sourcecred/sourcecred-action/commit/a243e2e93c56fa08204115223900db1d28fa39da/checks?check_suite_id=335051037). `Warning: GitHub installation access token has an unexpected hexadecimal component length of 40.`

Beanow:
In the documentation 16 characters were displayed.
But testing showed we're typically seeing 40.

Fixes #1474

Test plan: `yarn unit`

cc @vsoch


vsoch:
If this is the case, then this should be good. Will this merge into the dev tag so that we can test it on the next Sunday's run?

Beanow:
> Will this merge into the dev tag so that we can test it on the next Sunday's run?

Yes the dev tag is automatically built from the master branch.

PR #1490, opened 2019-12-13
Initiatives: add queries to mirror to implement DiscourseQueries

Beanow:
Depends on #1483

We defined a DiscourseQueries interface, intended as a subset of the Discourse plugin's MirrorRepository methods. This subset is used by the Initiatives plugin to source Iniaitive data.

We're now adding the new methods it needed to the MirrorRepository.

Want to ask @wchargin (as @decentralion suggested in Discord) about the "flow assertion" I wanted to add as a unit test. I ended up wanting to test this type `ReadRepository` is actually a superset of `DiscourseQueries` by casting it. Because there's no actual implementation code that uses a `ReadRepository` as being a `DiscourseQueries` in this commit, but it's intended to be used that way later.

```js
  describe("DiscourseQueries", () => {
    it("can be casted from ReadRepository", () => {
      // Don't need to actually call it. Flow will detect issues.
      function _unused_castReadRepository(
        repo: ReadRepository
      ): DiscourseQueries {
        return (repo: DiscourseQueries);
      }
    });
  });
```


Beanow:
I've adopted the shorter syntax @wchargin suggested for the Flow cast.
The negative case's annotation `// $ExpectFlowError` I haven't found any documentation on. Is this a Flow feature?

@decentralion can you review?

src/plugins/discourse/mirrorRepository.js
```diff
@@ -60,6 +60,18 @@ export interface ReadRepository {
...
+
+  /**
+   * Gets a Topic by ID.
+   */
+  topicById(id: TopicId): ?Topic;
+
+  /**
+   * Gets a number of Posts in a given Topic.
+   *
+   * Note: starting with the first post, ordered by `indexWithinTopic`.
```
decentralion: Please update docString to explain what the `numberOfPosts` argument does.

src/plugins/discourse/mirrorRepository.test.js
```diff
@@ -285,4 +285,92 @@ describe("plugins/discourse/mirrorRepository", () => {
...
+      cooked: "<p>Follow up 2</p>",
+    };
+
+    // When
+    repository.addTopic(topic);
+    // Deliberately scramble the adding order, order of `indexWithinTopic` should matter.
+    repository.addPost(p3);
+    repository.addPost(p1);
+    repository.addPost(p2);
+    const posts = repository.postsInTopic(topic.id, 2);
```
decentralion: For completeness; please test with `numberOfPosts` greater than the actual number of posts (e.g. `numberOfPosts == 5`)

Perhaps `numberOfPosts == 0` as well.

(No need to make separate test cases, just add a few lines to this one.)

src/plugins/discourse/mirrorRepository.js
Beanow: Done, updated to

```js
  /**
   * Gets a number of Posts in a given Topic. Starting with the first post,
   * ordered by `indexWithinTopic`.
   *
   * numberOfPosts: the maximum number of posts to get (may return fewer).
   */
  postsInTopic(topicId: TopicId, numberOfPosts: number): $ReadOnlyArray<Post>;
```

src/plugins/discourse/mirrorRepository.test.js
Beanow: Done, updated to

```js
    // When
    repository.addTopic(topic);
    // Deliberately scramble the adding order, order of `indexWithinTopic` should matter.
    repository.addPost(p3);
    repository.addPost(p1);
    repository.addPost(p2);
    const posts0 = repository.postsInTopic(topic.id, 0);
    const posts2 = repository.postsInTopic(topic.id, 2);
    const posts5 = repository.postsInTopic(topic.id, 5);

    // Then
    // Note: these are in order, starting from the opening post.
    expect(posts0).toEqual([]);
    expect(posts2).toEqual([p1, p2]);
    expect(posts5).toEqual([p1, p2, p3]);
```

Beanow:
It was originally for Initiatives data, but #1530 also came to depend on it for reference detection.
Requesting this is still included for a merge :] @decentralion

PR #1491, opened 2019-12-14
Enable and add upgrade testing for projectFromJSON

Beanow:
Same as the underlying util `fromCompat`, the object isn't statically typed, but parsed from JSON at runtime. Hence it's marked as `Compatible<any>`, to become `Compatible<T>` after validation.

Recognizing this potential difference at runtime, the new signature allows differently structured objects to be given as input for tests without flow errors. Allowing us to test the upgrade functionality.

Test plan: `yarn test` (unit and flow in particular)

### DIFF
--- a/src/core/project.js
+++ b/src/core/project.js
@@ -44,13 +44,11 @@ const upgrades = {
   "0.3.1": upgradeFrom030,
 };
 
-export type ProjectJSON = Compatible<Project>;
-
-export function projectToJSON(p: Project): ProjectJSON {
+export function projectToJSON(p: Project): Compatible<Project> {
   return toCompat(COMPAT_INFO, p);
 }
 
-export function projectFromJSON(j: ProjectJSON): Project {
+export function projectFromJSON(j: Compatible<any>): Project {
   return fromCompat(COMPAT_INFO, j, upgrades);
 }
 
--- a/src/core/project.js
+++ b/src/core/project.js
@@ -24,16 +24,35 @@ export type ProjectId = string;
  * the future (e.g. showing the last update time for each of the project's data
  * dependencies).
  */
-export type Project = {|
+export type Project = Project_v040;
+export type SupportedProject = Project_v030 | Project_v031 | Project_v040;
+
+type Project_v040 = {|
   +id: ProjectId,
   +repoIds: $ReadOnlyArray<RepoId>,
   +discourseServer: DiscourseServer | null,
   +identities: $ReadOnlyArray<Identity>,
 |};
 
+type Project_v031 = {|
+  ...Project_v040,
+  +discourseServer: {|
+    +serverUrl: string,
+    +apiUsername?: string,
+  |} | null,
+|};
+
+type Project_v030 = {|
+  ...Project_v040,
+  +discourseServer: {|
+    +serverUrl: string,
+    +apiUsername: string,
+  |} | null,
+|};
+
 const COMPAT_INFO = {type: "sourcecred/project", version: "0.4.0"};
 
-const upgradeFrom030 = (p) => ({
+const upgradeFrom030 = (p: Project_v030 | Project_v031): Project_v040 => ({
   ...p,
   discourseServer:
     p.discourseServer != null ? {serverUrl: p.discourseServer.serverUrl} : null,
@@ -48,7 +67,7 @@ export function projectToJSON(p: Project): Compatible<Project> {
   return toCompat(COMPAT_INFO, p);
 }
 
-export function projectFromJSON(j: Compatible<any>): Project {
+export function projectFromJSON(j: Compatible<SupportedProject>): Project {
   return fromCompat(COMPAT_INFO, j, upgrades);
 }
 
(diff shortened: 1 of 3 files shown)


src/core/project.js
```diff
@@ -44,13 +44,11 @@ const upgrades = {
...
 
-export type ProjectJSON = Compatible<Project>;
-
-export function projectToJSON(p: Project): ProjectJSON {
+export function projectToJSON(p: Project): Compatible<Project> {
   return toCompat(COMPAT_INFO, p);
 }
 
-export function projectFromJSON(j: ProjectJSON): Project {
+export function projectFromJSON(j: Compatible<any>): Project {
```
decentralion: What if instead of throwing away the type information, we just enumerate all the supported Project types? E.g.

```js
type SupportedProject = Project_v_0_3_0 | Project_v_0_3_1 | Project_v_0_4_0;
type Project = Project_v_0_4_0 // latest;
export function projectToJSON(p: Project): Compatible<Project> {
export function projectFromJSON(j: Compatible<SupportedProject>): Project {
```

Beanow: I suppose that would help mostly for our upgrade functions. Reality is still we could get any json data as input at runtime. I'll have a look at doing this.

Beanow: @decentralion I've tried this approach, but an interesting issue with `opaque` came up.

https://github.com/sourcecred/sourcecred/commit/ffe296001439bdad746d801018d3ace47c161c70

decentralion: I really like it in ffe2960. Keeping all of the types around for the older serializations will make it **so much easier** for us to upgrade the types down the line and maintain backwards compatibility. Let's just remove opaque from compatible. I trust us to be responsible with the ability to construct arbitrary compatible objects in the codebase, so the opacity isn't really needed.

decentralion:
As a reviewer, I find the commit message pretty hard to interpret:

> Same as the underlying util fromCompat, **the object** isn't statically typed, but parsed from JSON at runtime. Hence it's marked as Compatible<any>, to become Compatible<T> after validation.

This section is confusing because it never tells me what "the object" that is being changed is. Is it the (wrapped) compatible object? Or is it the underlying object which is being put into a compatible wrapper? I can infer this, but only if I read to the end of the paragraph and then go back to the beginning to re-interpret, which is a lot more cognitive labor.

> Recognizing this potential difference at runtime, the new signature allows differently structured objects to be given as input for tests without flow errors. Allowing us to test the upgrade functionality.

Again, what is the potential difference? Is the potential difference a difference between the generic type `T` and the real type of the object? In this case, figuring it out requires referencing the code as well as reading the commit message carefully.

When writing commit messages, I try to first establish local context: what are the pieces present, how do the pieces inter-relate, and why are we trying to change them? Having that context makes it much easier to understand and orient on the actual change the code is making.

I think a good example of this is the followon PR to this one (https://github.com/sourcecred/sourcecred/pull/1492), which makes it really clear what the context is, and why a change is needed. :)

src/core/project.js
Beanow: I've included that commit :+1:

Beanow:
> As a reviewer, I find the commit message pretty hard to interpret:

Suggesting an alternative one for use when squashing.

```
`projectFromJSON` makes use of `fromCompat`'s functionality to upgrade
from previous versions of the format. That upgrade code did not have any
unit tests or flow type coverage.

To add these tests a few changes were necessary:

- Defining `SupportedProject` and previous version of the `Project` type.
  To correctly express the function signatures with Flow.

- Changing the `projectFromJSON` argument from `Compatible<Project>` to
  `Compatible<SupportedProject>`. As testing with previous version formats
  in unit tests would trigger flow errors before.

- Removing `opaque` from the `Compatible<T>` type.  To support casting
  `Compatible<Project>` to `Compatible<SupportedProject>` in our tests.
```

src/core/project.js
```diff
@@ -24,16 +24,35 @@ export type ProjectId = string;
...
 
+type Project_v031 = {|
+  ...Project_v040,
+  +discourseServer: {|
+    +serverUrl: string,
+    +apiUsername?: string,
+  |} | null,
+|};
+
+type Project_v030 = {|
```
decentralion: Looks like Project_v030 is the same as Project_v031? Any reason we don't just say `type Project_v030 = Project_v031`?

src/core/project.js
```diff
@@ -48,7 +67,7 @@ export function projectToJSON(p: Project): Compatible<Project> {
   return toCompat(COMPAT_INFO, p);
 }
 
-export function projectFromJSON(j: Compatible<any>): Project {
+export function projectFromJSON(j: Compatible<SupportedProject>): Project {
```
decentralion: I'm fine leaving this as `Compatible<any>` since as you noted, it could actually be anything. However, if it's not a project we'll get a runtime error. that's fine. and this should allow us to not need the cast in test code...

src/core/project.js
Beanow: Difference is `discourseServer.apiUsername` became optional.

Beanow:
Closing in favor of #1517, #1518 and #1519. Achieving the same thing in smaller commits.

PR #1385, opened 2019-09-18
Integrate the identity plugin

decentralion:
This commit integrates the identity plugin, which was created in #1384.
It does this by adding explicit identity fields to the project
configuration, which are then applied when loading the graph in
`api/load.js`.

The actual integration is quite straightforward.

Test plan: The underlying logic is thoroughly tested; I added one new
test case to verify that it is integrated properly. Since the project
compat has changed, I've updated all the snapshots. Prior to merging
this PR, I will produce one "integration test", using this code to do
identity resolution for a real project (i.e. on the SourceCred instance
itself).

PR #1492, opened 2019-12-14
Use createProject to set default values for Project

Beanow:
Creation of new Project instances is spread out across the code.
So whenever there's a change in it's format, the PR is cluttered
with adding a logical default value in many places. It means
our default values might be inconsistent as well.

For example #1385 adds many `identities: [],` lines.
A similar situation would happen with the planned Initiatives
plugin, adding many `initiatives: null,` lines.

Using this function we can manage what default values to add
from a central place. Avoiding noise and code churn.

### DIFF
--- a/src/api/load.test.js
+++ b/src/api/load.test.js
@@ -8,7 +8,7 @@ import fs from "fs-extra";
 import type {Options as LoadGraphOptions} from "../plugins/github/loadGraph";
 import type {Options as LoadDiscourseOptions} from "../plugins/discourse/loadDiscourse";
 import {nodeContractions} from "../plugins/identity/nodeContractions";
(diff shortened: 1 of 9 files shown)


Beanow:
Note that in #1493 this function avoided the need to modify these files:

	src/api/load.test.js
	src/cli/discourse.js
	src/cli/genProject.js
	src/cli/load.test.js
	src/core/project_io.test.js
	src/plugins/github/specToProject.js
	src/plugins/github/specToProject.test.js

Beanow:
I wonder if this new factory function should be enforced similar to #1520 in a followup PR.
cc @decentralion

Beanow:
Rebased against master and leaving the above enforce idea for a followup.

PR #1493, opened 2019-12-14
Initiatives: add InitiativeOptions to project spec

Beanow:
Depends on #1492, #1517, #1518 and #1519 for tests and proper Project formatting.

These will be used for example as arguments for #1483's DiscourseInitiativeRepositoryOptions.

Test plan: `yarn test`


Beanow:
Needs an update, as Project had type changes and a refactor on upgrades.

Beanow:
I've updated to match the Project refactors.

Beanow:
Agreed, this change is for the discourse-tracker route of Initiatives. Using file-based this won't be needed.

PR #1495, opened 2019-12-19
Allow `while(true)` loops

decentralion:
The eslint no-constant-condition rule disallows while(true) loops,
since the true is a constant condition. However, I find the allowed
alternative (`for (;;)`) less readable, so I am adding the sub-rule that
allows constant conditions for loops.

Test plan: Add a `while (true)` loop somewhere in the codebase.
Before this commit, it will generate a lint error; afterwards, it is fine.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,6 +29,7 @@ module.exports = {
         caughtErrorsIgnorePattern: "^_$|^_unused_",
       },
     ],
+    "no-constant-condition": ["warn", {checkLoops: false}],
     "no-use-before-define": ["off"],
     "no-useless-constructor": ["off"],
     "no-case-declarations": ["off"],


decentralion:
@wchargin: it appears that eslint allows while-true loops inside of generator functions. if you change that function in the markov chain from a generator to a regular function, then eslint will fail.

i assume this is a bug in eslint? i can't see why it would be intended behavior.

Beanow:
> i assume this is a bug in eslint? i can't see why it would be intended behavior.

Looked into it a little. Having a yield statement explicitly considers it not a _constant condition_ loop.
Documented in https://github.com/eslint/eslint/pull/8762 and https://github.com/eslint/eslint/issues/8566
(side note, the issue reveals eslint also has a notion of champions :smile:)

TL;DR an infinite generator is different from an infinite loop, as the control is handed back to the caller for each loop with a generator. You can break the loop by not calling `.next()` on the generator.

decentralion:
Thanks for doing the research @beanow and finding why eslint behaves this way. I find eslint's (implicit) API reasoning inconsistent and therefore frustrating. In [the docs](https://eslint.org/docs/rules/no-constant-condition), the rationale for not supporting `while (true)` loops was that you can use `for (;;)` for the same effect. However, the same logic suggests it should be blacklisted for for generators too, since one could use `for (;;)` around a yield statement too. So they should just make `while(true)` OK by default. (Could be interesting to try sending a pull request to this effect, I wonder if they would merge it?)

Anyway, merging this PR.

Beanow:
Right, I find the exception of `for (;;)` really weird. The lack of a test makes it an implied constant "always" condition. Compared to `while (true)` so long as `true` is truthy, aka "always" :man_shrugging:. Both are constant conditions imo.

The exception for generators makes sense to me though. Yield creates an inversion of control.
So instead, the loop should be found at the control level.

```js
// Don't want.
while (true) {
  myGen.next();
}
```

PR #1496, opened 2019-12-19
Add `sourcecred init` command

decentralion:
The `sourcecred init` command interactively sets up a `sourcecred.json`
file, which is currently just a serialized `Project`.

The command guides the user through common options like adding GitHub
repos or orgs, and adding a Discourse server.

This is intended as a first step towards having an instance system.

Test plan:
This is an interactive tool which has relatively little room for
subtle/hard to track down bugs (either it produces a valid project
config or it doesn't). As such, I've decided not to write automated
tests.

I manually tested its common cases (loading orgs, adding repos, adding a
Discourse server), and verified that it will conservatively refuse to
overwrite an existing `sourcecred.json` file.

### DIFF
--- a/src/cli/init.js
+++ b/src/cli/init.js
@@ -1,0 +1,158 @@
+// @flow
+// Implementation of `sourcecred init`
+
+import dedent from "../util/dedent";
+import {type RepoId, stringToRepoId} from "../core/repoId";
+import {type Project, projectToJSON} from "../core/project";
+import type {Command} from "./command";
+import * as Common from "./common";
+import fs from "fs-extra";
+import process from "process";
+import read from "read";
+import path from "path";
+import {fetchGithubOrg} from "../plugins/github/fetchGithubOrg";
+import {type DiscourseServer} from "../plugins/discourse/loadDiscourse";
+
+function usage(print: (string) => void): void {
+  print(
+    dedent`\
+    usage: sourcecred init
+           sourcecred init --help
+
+    Sets up a new SourceCred instance, by interactively creating a SourceCred
+    project configuration, and saving it to 'sourcecred.json'.
+
+    Arguments:
+        --help
+            Show this help message and exit, as 'sourcecred help init'.
+
+    Environment variables:
+        SOURCECRED_GITHUB_TOKEN
+            API token for GitHub. This should be a 40-character hex
+            string. Required if you want to load whole GitHub orgs.
+
+            To generate a token, create a "Personal access token" at
+            <https://github.com/settings/tokens>. When loading data for
+            public repositories, no special permissions are required.
+            For private repositories, the 'repo' scope is required.
+    `.trimRight()
+  );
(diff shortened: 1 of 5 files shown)


src/cli/init.js
```diff
@@ -0,0 +1,158 @@
...
+      }
+      default: {
+        return die(std, `Unexpected argument ${args[i]}`);
+      }
+    }
+  }
+  const dir = process.cwd();
+  const projectFilePath = path.join(dir, "sourcecred.json");
+  if (await fs.exists(projectFilePath)) {
+    return die(std, `Refusing to overwrite sourcecred.json file in ${dir}`);
```
Beanow: I think this sums up the cautionary tale of interactive scripts well.

For the majority of settings I'm inclined to use a config format which supports comments, set up the minimum with interaction, and rely on easy to grock, well documented config and examples.

In this case the absolute minimum would be the project name.

Though the utility of the most common settings I can also see. Particularly scanning an organization to add individual repo names. That might also be a good case to make this a separate utility we can use for editing our config as well. An "automatically add missing repositories" could be really handy for large multi-repo orgs.

Not blocking, but wonder what your thoughts are on this.

src/cli/init.js
```diff
@@ -0,0 +1,158 @@
...
+  );
+}
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  std.err("fatal: run 'sourcecred help init' for help");
+  return 1;
+}
+
+const initCommand: Command = async (args, std) => {
```
Beanow: I see 3 concerns complected this function.
- cli command handling
- a `(...) => Promise<Project>` dialog
- I/O

Suggest separating these. Which would allow testing selectively, such as just the I/O when the dialog is mocked.

src/cli/init.js
```diff
@@ -0,0 +1,158 @@
...
+  const name = await aread({
+    prompt: "Choose an instance name:",
+    default: basename,
+  });
+  std.out("");
+
+  let repoIds: RepoId[] = [];
+  const githubToken = Common.githubToken();
+  if (githubToken == null) {
+    std.out("Skipping organization loading, as no GitHub token available.");
```
Beanow: We might also fall back on using interaction for the token, as a new user might overlook the token requirement. It would also allow us to mix org and repo adding in a single loop, so people don't have to plan out the order ahead of time.

Beanow:
Noting the build error seems unrelated to the change. But suspect it could be related to the recently resolved security issue with npm and yarn now changing requirements for overwriting globally installed binaries. Will look into that separately.

src/cli/init.js
decentralion: I think you're right, the interactive approach is really cumbersome. I've generated a new PR at #1515 which is a non-interactive approach, kind of based on the gen-project command. Please take a look over there; I'm going to close this PR as obsolete.

PR #1507, opened 2019-12-21
Prototype link was broken

burrrata:


decentralion:
I squashed and merged with an updated commit message. Thanks for the fix @burrrata!

PR #1510, opened 2019-12-22
UniRef: add CascadingReferenceDetector implementation

Beanow:
Depends on #1508 (and #1509 for tests)

The `CascadingReferenceDetector` is an abstraction we can use for combining multiple `ReferenceDetector` instances and giving them a priority.

### DIFF
--- a/src/core/references/cascadingReferenceDetector.test.js
+++ b/src/core/references/cascadingReferenceDetector.test.js
@@ -1,0 +1,85 @@
+// @flow
+
+import {type NodeAddressT, NodeAddress} from "../graph";
+import type {URL, ReferenceDetector} from "./referenceDetector";
+import {MappedReferenceDetector} from "./mappedReferenceDetector";
+import {CascadingReferenceDetector} from "./cascadingReferenceDetector";
+
+const nodeA = NodeAddress.fromParts(["test", "A"]);
+const nodeB = NodeAddress.fromParts(["test", "B"]);
+const nodeC = NodeAddress.fromParts(["test", "C"]);
+
+function exampleDetector(pairs?: [URL, NodeAddressT][]): ReferenceDetector {
+  const map: Map<URL, NodeAddressT> = new Map(pairs);
+  const refs = new MappedReferenceDetector(map);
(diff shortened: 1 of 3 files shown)


src/core/references/cascadingReferenceDetector.test.js
```diff
@@ -0,0 +1,85 @@
...
+import {CascadingReferenceDetector} from "./cascadingReferenceDetector";
+
+const nodeA = NodeAddress.fromParts(["test", "A"]);
+const nodeB = NodeAddress.fromParts(["test", "B"]);
+const nodeC = NodeAddress.fromParts(["test", "C"]);
+
+function exampleDetector(pairs?: [URL, NodeAddressT][]): ReferenceDetector {
+  const map: Map<URL, NodeAddressT> = new Map(pairs);
+  const refs = new MappedReferenceDetector(map);
+  jest.spyOn(refs, "addressFromUrl");
```
decentralion: It seems like the spyOn allows us to test the implementation details as well as the output. But all we really care about is the output. So I think we could simplify the tests a fair bit by removing all the spy checking, and just verifying that the logic is implemented correctly.

Beanow: I'm not entirely sure what you mean, but interpreting it this way:
We can verify the correctness of the logic the "black box" way of looking at inputs and outputs. The spy on the implementation detail is probably unnecessary.

I agree it could be black box tested and this might be over-constraining the class. That said, I'm expecting that there's likely to be a performance penalty for calling the underlying reference detectors and expect the reference detector to be used in several O(n) loops.

The simplicity of this cascading pattern is intended firstly as a priority, but not incurring a great performance cost is part of that as well. For instance, we could make the whole system async and call the underlying detectors in parallel, then deal with the priorities afterwards. But that would be stricly worse in complexity and performance.

Whether these implementation tests are the best way to capture that I'm not sure, but that's what I had it in mind for :smile:

decentralion: Cool. Happy to merge with this test.

PR #1511, opened 2019-12-22
UniRef: adding TranslatingReferenceDetector implementation

Beanow:
Depends on #1508 (and #1509 for tests)

The `TranslatingReferenceDetector` is an abstraction particularly useful for the Initiatives reference detector. Which should use the Discourse reference detector as it's base and translate the node address of the returned discourse topic to the initiative's node address.

### DIFF
--- a/src/core/references/translatingReferenceDetector.test.js
+++ b/src/core/references/translatingReferenceDetector.test.js
@@ -1,0 +1,56 @@
+// @flow
+
+import {type NodeAddressT, NodeAddress} from "../graph";
+import type {URL, ReferenceDetector} from "./referenceDetector";
+import {MappedReferenceDetector} from "./mappedReferenceDetector";
+import {TranslatingReferenceDetector} from "./translatingReferenceDetector";
+
+const nodeA = NodeAddress.fromParts(["test", "A"]);
+const nodeB = NodeAddress.fromParts(["test", "B"]);
+
+function exampleDetector(pairs?: [URL, NodeAddressT][]): ReferenceDetector {
+  const map: Map<URL, NodeAddressT> = new Map(pairs);
+  const refs = new MappedReferenceDetector(map);
+  jest.spyOn(refs, "addressFromUrl");
+  return refs;
+}
+
+describe("core/references/translatingReferenceDetector", () => {
+  describe("TranslatingReferenceDetector", () => {
+    it("should implement the ReferenceDetector interface", () => {
+      const _unused_toReferenceDetector = (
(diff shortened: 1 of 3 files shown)


src/core/references/translatingReferenceDetector.test.js
```diff
@@ -0,0 +1,56 @@
...
+function exampleDetector(pairs?: [URL, NodeAddressT][]): ReferenceDetector {
+  const map: Map<URL, NodeAddressT> = new Map(pairs);
+  const refs = new MappedReferenceDetector(map);
+  jest.spyOn(refs, "addressFromUrl");
+  return refs;
+}
+
+describe("core/references/translatingReferenceDetector", () => {
+  describe("TranslatingReferenceDetector", () => {
+    it("should implement the ReferenceDetector interface", () => {
```
decentralion: Since the TRD already is declared as implementing ReferenceDetector, I think this test is superfluous.

src/core/references/translatingReferenceDetector.test.js
```diff
@@ -0,0 +1,56 @@
...
+      const _unused_toReferenceDetector = (
+        x: TranslatingReferenceDetector
+      ): ReferenceDetector => x;
+    });
+
+    it("should not call translate function when base has no hit", () => {
```
decentralion: This test also seems unnecessary, since we're testing the specific implementation details rather than the observable output. (I suppose it could matter if calling the translate function had side effects... but then the real problem is that the translate fn has side effects)

Beanow: The translate should normally be possible as a good old idempotent function without side effects. But there may be performance penalties involved such as hitting a large database on disk. Moreover there's a very specific scope for the translation function.

```js
 type TranslateFunction = (NodeAddressT) => ?NodeAddressT;
```

Note this does not allow a `null | undefined` input.
While it technically could (in fact my first prototype was `(?NodeAddressT) => ?NodeAddressT`).
The semantics of the ReferenceDetector interface makes this really pointless though. When you can't verify the underlying node exists, you're not supposed to return an address. So if the base returns `null | undefined`, that's it, no point in translating that because you have no verified-to-exist node.

This test makes that explicit. And while that is redundant now with the Flow function signature, I think the semantic may be lost for future reference by only relying on the signature.

decentralion: Cool, thanks for the explanation. Also, who am I to complain about having too much tested behavior. ;) Approval stands.

Issue #1512, opened 2019-12-22
CI config issue, tag deploy didn't work


Beanow:
There's a problem with the tagged version deployment [here](https://github.com/sourcecred/sourcecred/blob/5727a831a92418d9247d2bd7dffd0707e78f6d75/.circleci/config.yml#L94-L114)
Which has a structure of:
```yml
      - docker/publish:
          # ...
          requires:
            - test
          filters:
            branches:
              ignore: /.*/
            tags:
              only: /^v.*/
          # ...
```

This job is intended to run only on tag releases with a `v` prefix.
However it has a requirement on a previous `test` job.

Looking at: https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag

> ### Executing Workflows for a Git Tag
>
> CircleCI does not run workflows for tags unless you explicitly specify tag filters. Additionally, if a job requires any other jobs (directly or indirectly), you must use regular expressions to specify tag filters for those jobs. Both lightweight and annotated tags are supported.

In other words, our deployment job would never run. As we require `test` to run first.
And `test` has no explicit filter to run on tag pushes. So it implicitly **won't**.

![facepalm](https://media.giphy.com/media/SEp6Zq6ZkzUNW/giphy.gif) <sup>(Oh circle and your weird config)</sup>

To fix, we need to copy the test job and set it's filter so it will run on tag pushes.
At which point we might as well call this a different workflow.

```diff
workflows:
  version: 2.0
  commit:
    jobs:
      - test
      - deploy_dryrun:
          requires:
            - test
          filters:
            branches:
              ignore:
                - master
      - deploy_dev:
          requires:
            - test
          filters:
            branches:
              only: master
+
+  tagged-release:
+    jobs:
+      - test:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/
      - deploy_tag:
          requires:
            - test
          filters:
            branches:
              ignore: /.*/
            tags:
              only: /^v.*/
```

Beanow:
Created a test project to demonstrate.
1. [Before configuration](https://github.com/Beanow/circle-tag-filters/tree/v0.1.1), has no `deploy_tag` 
![image](https://user-images.githubusercontent.com/497556/71328391-e33aa100-2516-11ea-90d0-b0c3cd6bf688.png)
2. [After configuration](https://github.com/Beanow/circle-tag-filters/tree/v0.1.2), does have `deploy_tag`
[...]

PR #1513, opened 2019-12-22
CI: remove broken tag deploy

Beanow:
See #1512 for full context.

Short explanation:
Because the job wants to run only on tag pushes, but requires the `test` job (which doesn't run on tag pushes), the job will never run.

Test plan:
Created a test project to demonstrate.
1. [Before configuration](https://github.com/Beanow/circle-tag-filters/tree/v0.1.1), has no `deploy_tag` 
image
2. [After configuration](https://github.com/Beanow/circle-tag-filters/tree/v0.1.2), does have `deploy_tag`
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,28 +93,6 @@ workflows:
 
       - docker/publish:
           image: sourcecred/sourcecred
-          tag: latest
-          requires:
-            - test
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              only: /^v.*/
-          extra_build_args: --cache-from=node:12,sourcecred/sourcecred:dev
-          before_build:
-            - pull_cache
-          after_build:
-            - run:
-                name: Publish Docker Tag with Sourcecred Version
-                command: |
-                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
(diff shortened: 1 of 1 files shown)


decentralion:
Would love it if @vsoch  can approve this -- I haven't been paying close attention to the docker CI stuff.

vsoch:
Ah I remember those screen shots! I think we discussed this in #1512 - it's good with me to remove the section that will never run. @Beanow are we okay with just publishing a `dev` tag? Arguably the tagging and then push (of two images, dev and the tag) could happen in one block, the current / remaining docker/publish. If we don't want to publish the additional tag, I think we probably want to remove the tag preview section (it wouldn't make sense to derive it).

https://github.com/sourcecred/sourcecred/blob/45d10ea455fb1ea736e9ea4a6bf8524126a9ce11/.circleci/config.yml#L87-L92

Beanow:
> @Beanow are we okay with just publishing a `dev` tag?

No :smile:, it's just that I've broken the PR up in two smaller commits.

This removes the job that never runs. And #1514 follows up on this by adding back a new workflow which should do the tagged release correctly.

The way I would suggest reviewing is:
If the tag publishing job I'm removing is indeed broken and the removal is done correctly. This PR can be merged.

Moving on to #1514 we can discuss whether the replacement workflow makes sense, is correct, and the tagging preview is still set up correctly.

PR #1514, opened 2019-12-22
CI: add tagged-release workflow

### LINKED ISSUE #1512: CI config issue, tag deploy didn't work

Beanow:
Continues from #1513
Fixes #1512

In contrast with our previous "tags only" deploy job, this configuration makes sure all it's preceding jobs are also set to a "tags only" filter in order to run.

Test plan:
Created a test project to demonstrate.
1. [Before configuration](https://github.com/Beanow/circle-tag-filters/tree/v0.1.1), has no `deploy_tag` 
image
2. [After configuration](https://github.com/Beanow/circle-tag-filters/tree/v0.1.2), does have `deploy_tag`
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,6 +111,43 @@ workflows:
             branches:
               only: master
 
+  # Separate workflow just for version tag releases.
+  tagged-release:
+    jobs:
+      - test:
+          filters: &version-tag-only
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v.*/
+
+      - test_full:
+          filters: *version-tag-only
+
+      - test_full_10:
+          filters: *version-tag-only
+
+      - docker/publish:
+          filters: *version-tag-only
+          requires:
+            - test
+            - test_full
+            - test_full_10
+          deploy: false
+          image: sourcecred/sourcecred
+          tag: latest
+          extra_build_args: --cache-from=node:12,sourcecred/sourcecred:dev
+          before_build:
+            - pull_cache
+          after_build:
+            - run:
+                name: Publish Docker Tag with Sourcecred Version
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
+                   echo "Version for Docker tag is ${DOCKER_TAG}"
+                   docker tag sourcecred/sourcecred:latest sourcecred/sourcecred:${DOCKER_TAG}
+                   docker push sourcecred/sourcecred:${DOCKER_TAG}
+                   docker push sourcecred/sourcecred:latest
 
   nightly:
     triggers:


Beanow:
Rebased against master, @vsoch can you help review this? :]

.circleci/config.yml
```diff
@@ -111,6 +111,41 @@ workflows:
...
+          before_build:
+            - pull_cache
+          after_build:
+            - run:
+                name: Publish Docker Tag with Sourcecred Version
+                command: |
+                   DOCKER_TAG=$(docker run sourcecred/sourcecred --version | cut -d' ' -f2)
+                   echo "Version for Docker tag is ${DOCKER_TAG}"
+                   docker tag sourcecred/sourcecred:latest sourcecred/sourcecred:${DOCKER_TAG}
+                   docker push sourcecred/sourcecred
```
vsoch: This is just a personal preference, but for pushes I think it's worth being explicit about the tags being pushed, both for the developer to see as documentation, and in the case that there is any tiny chance that a sourcecred/sourcecred:other or sourcecred/sourcecred:dev container could be pushed by accident.

.circleci/config.yml
```diff
@@ -111,6 +111,41 @@ workflows:
...
+
+      - docker/publish:
+          filters: *version-tag-only
+          requires:
+            - test
+            - test_full
+            - test_full_10
+          image: sourcecred/sourcecred
+          tag: latest
+          extra_build_args: --cache-from=node:12,sourcecred/sourcecred:dev
```
vsoch: could this ever be node 10? And just to double check - the dev tag is going to be the latest (and likely most useful for cache, moreso than latest?)

.circleci/config.yml
```diff
@@ -111,6 +111,41 @@ workflows:
             branches:
               only: master
 
+  # Separate workflow just for version tag releases.
+  tagged-release:
+    jobs:
+      - test:
+          filters: &version-tag-only
```
vsoch: nicely done!

.circleci/config.yml
Beanow: Currently we only ship `node:12` based images (See [`Dockerfile`](https://github.com/sourcecred/sourcecred/blob/279bbdd451d4fec23b44baf16231b65d3ed7db36/Dockerfile#L1)).
We do use `node:10` for tests, but that's a Circle executor, not our base image.

.circleci/config.yml
Beanow: :smile:  anchors are so helpful in reducing noise

.circleci/config.yml
Beanow: Agree with that :smiley:.
I've made them explicit, but that created an interesting situation.

`docker push sourcecred/sourcecred:${DOCKER_TAG}` would have been enough, because the `:latest` tag would be pushed by the Orb.

But that seems likely to catch someone off guard when they're also seeing an explicit push here. With the https://en.wikipedia.org/wiki/Principle_of_least_astonishment I decided to:
- add `deploy: false` so the Orb doesn't push `:latest`.
- add both push tags explicitly.

What do you think about this?

.circleci/config.yml
vsoch: okay doke, this sounds good to me.

PR #1515, opened 2019-12-28
Add non-interactive `sourcecred init` command

decentralion:
The `sourcecred init` command sets up a `sourcecred.json`
file, which is currently just a serialized `Project`. This is based on
work in #1496, but is non-interactive instead.

This is a first step towards implementing the [instance system].

[instance system]: https://discourse.sourcecred.io/t/sourcecred-instance-system/244

Test plan:

This command is thoroughly tested by sharness tests; see
`sharness/test_cli_init.t`.

Thanks to @wchargin for help with the sharness testing.

### DIFF
--- a/src/cli/init.js
+++ b/src/cli/init.js
@@ -3,7 +3,6 @@
 
 import stringify from "json-stable-stringify";
 import dedent from "../util/dedent";
-import {type RepoId} from "../core/repoId";
 import {type Project, projectToJSON, createProject} from "../core/project";
 import type {Command} from "./command";
 import * as Common from "./common";
@@ -11,6 +10,7 @@ import fs from "fs-extra";
 import process from "process";
 import path from "path";
 import {type DiscourseServer} from "../plugins/discourse/loadDiscourse";
+import {type RepoId} from "../plugins/github/repoId";
 import {specToProject} from "../plugins/github/specToProject";
 import * as NullUtil from "../util/null";
 
@@ -106,17 +106,6 @@ const initCommand: Command = async (args, std) => {
         if (++i >= args.length)
           return die(std, "--discourse given without value");
         discourseUrl = args[i];
-        /*
-        if (!discourseUrl.match(new RegExp("^https?://"))) {
-          return die(
-            std,
-            "invalid discourse url: must start with http:// or https://"
-          );
-        }
-        if (discourseUrl.endsWith("/")) {
-          discourseUrl = discourseUrl.slice(0, discourseUrl.length - 1);
-        }
-        */
         break;
       }
       case "--force": {
--- a/src/cli/init.js
+++ b/src/cli/init.js
@@ -86,7 +86,7 @@ function die(std, message) {
 const initCommand: Command = async (args, std) => {
   let withForce = false;
   let printToStdOut = false;
-  let discourseUrl: ?string;
+  let discourseUrl: string | null = null;
   let githubSpecs: string[] = [];
 
   for (let i = 0; i < args.length; i++) {
@@ -106,6 +106,7 @@ const initCommand: Command = async (args, std) => {
         if (++i >= args.length)
           return die(std, "--discourse given without value");
         discourseUrl = args[i];
+        /*
         if (!discourseUrl.match(new RegExp("^https?://"))) {
           return die(
             std,
@@ -115,6 +116,7 @@ const initCommand: Command = async (args, std) => {
         if (discourseUrl.endsWith("/")) {
           discourseUrl = discourseUrl.slice(0, discourseUrl.length - 1);
         }
+        */
         break;
       }
       case "--force": {
@@ -138,26 +140,13 @@ const initCommand: Command = async (args, std) => {
   }
 
   const githubToken = Common.githubToken();
-  if (!githubToken && githubSpecs.length > 0) {
-    return die(std, "tried to load GitHub specs, but no GitHub token provided");
-  }
-  let repoIds: RepoId[] = [];
-  for (const spec of githubSpecs) {
-    const subproject = await specToProject(spec, NullUtil.get(githubToken));
-    repoIds = [...repoIds, ...subproject.repoIds];
-  }
 
-  const discourseServer: DiscourseServer | null = discourseUrl
-    ? {serverUrl: discourseUrl}
-    : null;
-  const project: Project = createProject({
-    // the id field is obsolete in the instance system, and will be
-    // removed once we fully migrate to sourcecred instances.
-    id: "obsolete-id",
-    repoIds,
-    discourseServer,
-  });
-
+  if (githubToken == null && githubSpecs.length) {
+    die(std, `provided GitHub specs, but no GitHub token present. try --help`);
+  }
+  const specsWithToken =
+    githubToken == null ? null : {specs: githubSpecs, token: githubToken};
+  const project = await genProject(discourseUrl, specsWithToken);
   const projectJson = projectToJSON(project);
   const stringified = stringify(projectJson, {space: 4});
   if (printToStdOut) {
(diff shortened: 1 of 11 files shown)


src/cli/init.js
```diff
@@ -0,0 +1,183 @@
...
+  );
+}
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  std.err("fatal: run 'sourcecred help init' for help");
+  return 1;
+}
+
+const initCommand: Command = async (args, std) => {
```
Beanow: Like with previous PR, suggest separating the:
- cli command handling
- `Project` generating logic
- I/O or printing of a `Project`

src/cli/init.js
```diff
@@ -0,0 +1,183 @@
...
+  let githubSpecs: string[] = [];
+
+  for (let i = 0; i < args.length; i++) {
+    switch (args[i]) {
+      case "--help": {
+        usage(std.out);
+        return 0;
+      }
+      case "--github": {
+        if (++i >= args.length) return die(std, "--github given without value");
```
Beanow: The `(++i >= args.length)` check seems off to me. Aside from code golfing, what happens with:
`sourcecred init --github --github sourcecred/sourcecred`?

src/cli/init.js
```diff
@@ -0,0 +1,183 @@
...
+  }
+  const dir = process.cwd();
+  const projectFilePath = path.join(dir, "sourcecred.json");
+  const fileAlreadyExists = await fs.exists(projectFilePath);
+  if (fileAlreadyExists && !(withForce || printToStdOut)) {
+    return die(std, `refusing to overwrite sourcecred.json without --force`);
+  }
+
+  const githubToken = Common.githubToken();
+  if (githubToken == null && githubSpecs.length > 0) {
```
Beanow: Nit: I think we can solve this with Flow and a small check.

```js
const githubToken: ?string = Common.githubToken();
if (githubToken) { /* ... */ }
```

As returning an empty string would both be a semantic error (should be `null | undefined` instead).
And `"" == false` anyway so it would retain the current check.

yarn.lock
```diff
@@ -7180,6 +7180,13 @@ read-pkg@^3.0.0:
     normalize-package-data "^2.3.2"
     path-type "^3.0.0"
 
+read@^1.0.7:
```
Beanow: I don't think we're using read anymore. May want to revert the lock file.

src/cli/init.js
```diff
@@ -0,0 +1,183 @@
...
+  const discourseServer: DiscourseServer | null = discourseUrl
+    ? {serverUrl: discourseUrl}
+    : null;
+  const project: Project = {
+    // the id field is obsolete in the instance system, and will be
+    // removed once we fully migrate to sourcecred instances.
+    id: "obsolete-id",
+    repoIds,
+    discourseServer,
+    identities: [],
```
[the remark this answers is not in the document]
decentralion: Done

src/cli/init.js
decentralion: What seems off to you about the check?

`sourcecred init --github --github sourcecred/sourcecred` will error:

`fatal: unexpected argument sourcecred/sourcecred`

Providing the github flag twice in a row is an invalid input, so that it errors is fine. If we wanted an even more helpful error message, we could do validation on each argument to `--github` to check that it looks like a valid github spec, at which point it would say `fatal: "--github" is not a valid github spec`, which is perhaps more helpful.

src/cli/init.js
decentralion: When you say "solve this", what problem are you proposing solving? Do you think that comparing the GitHub token to null is a problem? Also, how is flow involved in solving this problem?

While I'm not sure if this is what you're proposing: switching the githubToken check to just evaluate truthiness is a good idea, since we should treat an explicitly empty github token (which you can generate, e.g. via `SOURCECRED_GITHUB_TOKEN="" node bin/sourcecred.js init`) the same as a null github token. So I've made this change.

yarn.lock
decentralion: Good call.

src/cli/init.js
Beanow: Have no problem with comparing to null, I think the string length check is not a case we should consider here.

My thinking is that:
- `""` is an unexpected, invalid token.
- `null` is an expected case from the `?string` type.
- `undefined` is an expected case from the `?string` type.

Should empty strings cause an issue, but considered equivalent to `null | undefined` `Common.githubToken()` should have returned `null | undefined` instead.

Either way `""` is falsy, so we don't need the `.length > 0` check.

src/cli/init.js
decentralion: So, I tried this, and the resultant code felt much more complicated, difficult to reason about and maintain. As an example, I tried to factor the argument parsing logic out of the main IO flow of the project, but then found myself dealing with issues like that the command should print usage and exit as soon as it finds the `--help` flag, rather than potentially error-ing due to malformed flags.

As I went down this (and other rabbit holes), I felt that I was solving a puzzle rather than a problem (to borrow a quote from [a Hickey talk you linked](https://www.youtube.com/watch?v=2V1FtfBDsLU)). IMO, having a complected 100 line CLI script written in an imperative mood is not a problem.

Here is what I came up with when I tried to pull it apart. Let me know if you would have preferred this version.

```js
const initCommand: Command = async (args, std) => {
  const maybeParsedArgs = parseArguments(args);
  if (maybeParsedArguments.type === "FAILURE") {
    return die(maybeParsedArgs.failure);
  }
  const parsedArgs = maybeParsedArgs.result;
  if (parsedArgs.wantsHelp) {
    usage(std.out);
    return 0;
  }

  const maybeProject = await generateProject(parsedArgs, Common.githubToken());
  if (maybeProject.type === "FAILURE") {
    return die(maybeProject.failure);
  }
  const project = maybeProject.result;

  const dir = process.cwd();
  const projectFilePath = path.join(dir, "sourcecred.json");
  if ((await fs.exists(projectFilePath)) && !withForce) {
    return die(std, `refusing to overwrite sourcecred.json without --force`);
  }
  const basename = path.basename(dir);

  const githubToken = Common.githubToken();
  if (githubToken == null && githubSpecs.length > 0) {
    return die(
      std,
      "tried to load GitHub specs, but no GitHub token provided."
    );
  }
  let repoIds: RepoId[] = [];
  for (const spec of githubSpecs) {
    const subproject = await specToProject(spec, NullUtil.get(githubToken));
    repoIds = [...repoIds, ...subproject.repoIds];
  }

  const discourseServer: DiscourseServer | null = discourseUrl
    ? {serverUrl: discourseUrl}
    : null;
  const project: Project = {
    id: basename,
    repoIds,
    discourseServer,
    identities: [],
  };

  const projectJson = projectToJSON(project);
  await fs.writeFile(projectFilePath, JSON.stringify(projectJson, null, 2));

  std.out("Wrote project file to 'sourcecred.json'");

  return 0;
};
```

You can find the [implementation here](https://github.com/sourcecred/sourcecred/blob/monadic-init/src/cli/init.js). (Sidenote: how do you push a single commit to GitHub without making a vestigial branch for it?)

decentralion:
@Beanow: Yesterday I replied to most of your comments, but never finished and never actually pushed the updated version. I've done that now. :)

src/cli/init.js
Beanow: > (Sidenote: how do you push a single commit to GitHub without making a vestigial branch for it?)

Afaik, you can't. Typically I link to commits after using a branch to push them.

Beanow:
The dependency #1492 has been merged. Can you rebase?

src/cli/init.js
Beanow: I've added an example of what I have in mind: https://github.com/sourcecred/sourcecred/commit/9830bb4e896d2ba34e7ab5a1898880b49aef5250

Notice how this makes the functions `generateProject` and `outputProject` correct even outside the CLI context. And how much less complex `initCommand` is.

wchargin: > Sidenote: how do you push a single commit to GitHub without making a
> vestigial branch for it?

Use `git push origin HEAD:refs/heads/new-branch-name`, or replace `HEAD`
with any ref.

This obviously creates a branch on GitHub’s side, but I presume that
you’re asking how to avoid creating a local branch. This is how.

I use this dozens of times per day. I don’t use branches for individual
pull requests (instead, I use them for actual *branches* of sequences of
changes when this is a helpful thing to do), so this is my standard
workflow.

nothingismagick: Didn't come here for this, but it made my day. Thanks @wchargin

decentralion:
Decided to prioritize cleaning up the sc algorithm a bit (to unblock other contributors) over setting up instances. Temporarily closing this PR until I give it my attention again.

PR #1516, opened 2019-12-28
Remove deprecated gen-project command

decentralion:
The (untested, explicitly experimental) gen-project command is redundant
with the (tested, and intended for long-term support) init command.
Thus, I am removing it.

Test plan: `yarn flow` passes, and `git grep gen-project` has no hits.


decentralion:
Closing until I prioritize #1515

PR #1517, opened 2020-01-01
Define previous Project type versions

Beanow:
These are important to accurately add types to function signatures of validating and upgrading logic.

This is a replacement for #1491, splitting it into smaller PRs.

Test plan:
Manually validate previous types.
Relevant PRs where it changed: #1451 (v0.4.0), #1431 (v0.3.1, and v0.3.0 in the before state of this diff)

### DIFF
--- a/src/core/project.js
+++ b/src/core/project.js
@@ -54,7 +54,7 @@ export function encodeProjectId(id: ProjectId): string {
   return base64url.encode(id);
 }
 
-const upgradeFrom030 = (p) => ({
+const upgradeFrom030 = (p: Project_v031 | Project_v030) => ({
   ...p,
   discourseServer:
     p.discourseServer != null ? {serverUrl: p.discourseServer.serverUrl} : null,
--- a/src/core/project.js
+++ b/src/core/project.js
@@ -24,7 +24,10 @@ export type ProjectId = string;
  * the future (e.g. showing the last update time for each of the project's data
  * dependencies).
  */
-export type Project = {|
+export type Project = Project_v040;
+export type SupportedProject = Project_v030 | Project_v031 | Project_v040;
+
+type Project_v040 = {|
   +id: ProjectId,
   +repoIds: $ReadOnlyArray<RepoId>,
   +discourseServer: DiscourseServer | null,
@@ -32,17 +35,6 @@ export type Project = {|
 |};
 
 const COMPAT_INFO = {type: "sourcecred/project", version: "0.4.0"};
-
-const upgradeFrom030 = (p) => ({
-  ...p,
-  discourseServer:
-    p.discourseServer != null ? {serverUrl: p.discourseServer.serverUrl} : null,
-});
-
-const upgrades = {
-  "0.3.0": upgradeFrom030,
-  "0.3.1": upgradeFrom030,
-};
 
 export type ProjectJSON = Compatible<Project>;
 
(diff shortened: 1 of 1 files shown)


src/core/project.js
```diff
@@ -24,13 +24,32 @@ export type ProjectId = string;
  * the future (e.g. showing the last update time for each of the project's data
  * dependencies).
  */
-export type Project = {|
+export type Project = Project_v040;
+export type SupportedProject = Project_v030 | Project_v031 | Project_v040;
+
+type Project_v040 = {|
```
decentralion: I notice that you define the current version as raw/full object, and then define the older versions as modifications to the current version.

It seems there are three options here:

1. Current version is raw, older versions are modifications to current
2. Oldest version is raw, newer versions are modifications to older
3. All versions are raw

Where by "raw" I mean specifying the full object rather than using tricks like `...Project_v040`.

Is there a particular reason for doing it this way? Is the expectation that when we make a change and have Project_v041, we'll modify Project_v040 to be defined relative to Project_v041?

What do you think of taking option 2 instead? I think it will be less confusing if every historical project type just has a full declaration. Will take more space though.

Beanow: Good point. My preferences are 1 or 3. As I think it's important to make the current version human readable. The cognitive load of option 2 would be to do shallow merges in your head to figure out what the current signature is.

3 would be most readable with this in mind, however I think it's mostly the current version for which reading as-is is important. The others should primarily be correct for the upgrading feature. And it might actually be a more common question "what's the difference" for example when trying to figure out if the upgrade function is correct.

decentralion: Heh, I got confused with my own numbering system--I actually meant to endorse option 3. I think option 2 is the worst because, indeed, the most important thing is to be able to quickly reference the current signature.

I think you're right that readability matters the most for the current version, and for older versions they just need to be technically correct for the typechecker and validating upgrade functions. However, option 1 does involve more cognitive load when setting up the versions, so there's a higher probability that we'll make mistakes when writing the types. (Hopefully we would catch these quickly due to type errors, though.)

Here's a question. Suppose we remove a field. Is there a concise, inline way to write "Project_v040 with X field removed"? If not, then every time we do a field removal we'll need to write the full raw type anyway.

If we can't remove fields, then I think we should just keep it simple and consistent and enumerate the full spec for each version. However, to cut down on clutter, we could put all the older version specs at the bottom of the file.

I'd like to make a decision here synchronously and merge today, but cc @wchargin who may have interesting thoughts on the matter.

Beanow: > Here's a question. Suppose we remove a field. Is there a concise, inline way to write "Project_v040 with X field removed"? If not, then every time we do a field removal we'll need to write the full raw type anyway.

I'm not sure, as this is a flow type spread it has a different implementation from object spreads in actual JS. That said JS doesn't have a functional-style delete in this style either other than:

```js
const without = {
  ...base,
  removeMe: undefined
};
```

Converting from a "raw" to a difference-to-newer definition does sound error prone. So I'm happy to switch to all raws. :]

Beanow: Done :+1:

Beanow:
Sorted it to have this at the bottom.

- upgrade from v3
- v3

As it would make more sense for our case:

- upgrade from v0.3.0 | v0.3.1
- v0.3.1
- v0.3.0

Beanow:
@decentralion the function signature added in the nit commit is also present in the followup PR #1518
https://github.com/sourcecred/sourcecred/pull/1518/files#diff-6afd9097cae739340b1fef26d4c040edR59

Which I think is slightly better as it includes the return type :smile:

```js
const upgradeFrom030 = (p: Project_v030 | Project_v031): Project_v040 => ({
```

Anyhow, I meant to split this into smaller commits:
- Adding new type definitions
- Improving signatures of existing code with them

Since the second includes the change you're requesting and is approved, I'll merge them as separate commits and revert the change here.

PR #1518, opened 2020-01-01
Update projectFromJSON & upgrade function signatures

Beanow:
This creates better flow type coverage for the upgrading from older Project types feature.

Note `projectFromJSON`'s function signature changes like this:

```diff
- (Compatible<Project>) => Project
+ (Compatible<any>) => Project
```

And that makes sense, because we use this function to validate an object we parsed from JSON at runtime. It could actually be anything.

Added benefit is that is makes writing unit tests possible. Because now will flow not throw a type error when we provide something other than `Compatible<Project>` as input, to test upgrading or validation functionality.

Note that the underlying utility `fromCompat` already uses `Compatible<any>` for the same object.

This is a replacement for #1491, splitting it into smaller PRs.

Test plan: `yarn flow`

### DIFF
--- a/src/core/project.js
+++ b/src/core/project.js
@@ -42,7 +42,7 @@ export function projectToJSON(p: Project): ProjectJSON {
   return toCompat(COMPAT_INFO, p);
 }
 
-export function projectFromJSON(j: ProjectJSON): Project {
+export function projectFromJSON(j: Compatible<any>): Project {
   return fromCompat(COMPAT_INFO, j, upgrades);
 }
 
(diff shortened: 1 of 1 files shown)


decentralion:
Can you update the commit /  PR description to give an explanation of why switching from taking ProjectJSON to taking Compat<any> is semantically appropriate? (We've discussed this on the other PR / offline, and I understand the reason, but I want it to be the case that someone running `git blame` on this line would also be able to find the justification.)

Beanow:
Yes, `projectFromJSON`'s function signature changes like this:

```diff
- (Compatible<Project>) => Project
+ (Compatible<any>) => Project
```

And that makes sense, because we use this function to validate an object we parsed from JSON at runtime. It could actually be anything.

Added benefit is that is makes writing unit tests possible. Because now flow not throw a type error when we provide something other than `Compatible<Project>` as input to test upgrading or validation functionality.

Note that the underlying utility `fromCompat` already uses `Compatible<any>` for the same object.

decentralion:
Cool, please add that to the PR / commit description and I will approve. :)

Beanow:
Rebased to target master. Also merged a conflict as #1517 changed the order so older versions and upgrade are at the bottom of the file.

PR #1521, opened 2020-01-03
Add instructions to set up the GitHub API Token

burrrata:
The README explains how to set the SOURCECRED_GITHUB_TOKEN, but later in the Docker section. People who aren't using Docker will follow the initial installation instructions. Added the instructions to set that up when users first install and setup SourceCred.

PR #1522, opened 2020-01-03
ci: fail fast on forked PRs with no credentials

wchargin:
Summary:
PRs created from forks don’t have credentials when running CI. This
commit causes the `test-full` job (which requires credentials) to fail
fast with a helpful error message.

Test Plan:
Push distinct versions of this commit to a fork and to the main
repository, and open pull requests for each. Note that the tests pass
from the main repository, but fail with a nice message from the fork:

![Screenshot of expected fast-fail behavior][ss]

The “team member pushes to trusted branch” workflow has already been
successfully exercised for #1521.

[ss]: https://user-images.githubusercontent.com/4317806/71707839-b782ab00-2da1-11ea-8aa9-7d8720538a87.png

wchargin-branch: forked-pr-fail-fast

### DIFF
--- a/config/test.js
+++ b/config/test.js
@@ -10,6 +10,11 @@ main();
 
 function main() {
   const options = parseArgs();
+  if (isForkedPrFullRun(options)) {
+    printForkedPrFullRunErrorMessage();
+    process.exitCode = 1;
+    return;
+  }
   const printVerboseResults = options.mode === "FULL";
   const runOptions = {printVerboseResults};
   const tasks = makeTasks(options.mode, options.limitMemoryUsage);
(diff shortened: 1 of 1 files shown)


config/test.js
```diff
@@ -33,6 +36,36 @@ function parseArgs() {
...
+  }
+  console.error(
+    [
+      "fatal: cannot run full test suite: missing credentials",
+      "Tests on forked PRs run without credentials by default. A core team ",
+      "member will sanity-check your PR and push its head commit to a branch ",
+      "on the main SourceCred repository, which will re-run these tests.",
+    ].join("\n")
+  );
+  process.exitCode = 1;
```
decentralion: Why does checking if we are on a fork trigger these side effects? It seems much clearer to move this above to line 13, and let this check just be a check.

wchargin: This function is intended to conceptually mean “ensure that we’re not on
a fork, printing a message and exiting the process if we are (and
wtherwise returning nothing)”. E.g., in Python you’d `sys.exit(1)` here.
But in Node you’re not supposed to use `process.exit` because it
hard-exits Node without waiting for other tasks to finish, which can
lead to your stdout/stderr not being flushed, among other bad things. So
the recommended practice is this contorted mess:
<https://nodejs.org/api/process.html#process_process_exit_code>

It’s a pretty terrible API, to be sure.

I can inline the `console.error` and `process.exitCode` bit into `main`
if you like, but I think that it’s valuable that `main` be short and
easily skimmable. If you prefer, we could restructure it to

```javascript
function main() {
  const options = parseArgs();
  if (checkForkedPrFullRun(options)) {
    printForkedPrFullRunErrorMessage();
    process.exitCode = 1;
    return;
  }
  // ...
}
```

but that just feels like extra code with little value to me.

No strong feelings here; what do you think?

decentralion: I have a slight preference for the refactored version you put forward, but hearing that you considered alternative approaches is good enough for me. Since this is code that will mostly be touched only by maintainers (and hopefully accumulates complexity only slowly) I'm not as concerned with how it is factored, especially in more "stylistic" cases like this. Approved.

wchargin: Okay, changed to your preferred way.

PR #1526, opened 2020-01-07
Expose some internal APIs in a new bundle

wchargin:
Summary:
For convenient import by scripts and Observable notebooks that want to
use SourceCred code outside its normal build system. We export a subset
of the codebase, including some core data structures and algorithms and
also some plugin metadata, but no plugin loading code.

To build, run `yarn backend` (or `yarn backend --watch`), then grab the
new `bin/api.js` file.

Test Plan:
Sample usage, with normal Node:

```javascript
const {
  core: {
    graph: {Graph, NodeAddress, EdgeAddress},
  },
} = require("./api").default;

function node(address) {
  return {
    address,
    description: "blurgh",
    timestampMs: -1,
  };
}

const g = new Graph();
g.addNode(node(NodeAddress.fromParts(["people", "alice"])));
g.addNode(node(NodeAddress.fromParts(["people", "bob"])));
g.addEdge({
  address: EdgeAddress.fromParts(["friendship"]),
  src: NodeAddress.fromParts(["people", "alice"]),
  dst: NodeAddress.fromParts(["people", "bob"]),
  timestampMs: 0,
});

console.log(require("json-stable-stringify")(g));
```

This prints a valid graph JSON object.

wchargin-branch: api-bundle

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -1,0 +1,44 @@
+// @flow
+
+import deepFreeze from "deep-freeze";
+
+// Exports for calling SourceCred code programmatically. Both the
+// structure and the contents of this API are experimental and subject
+// to change.
+import * as address from "../core/address";
+import * as discourseAddress from "../plugins/discourse/address";
+import * as discourseDeclaration from "../plugins/discourse/declaration";
+import * as githubDeclaration from "../plugins/github/declaration";
+import * as githubEdges from "../plugins/github/edges";
+import * as githubNodes from "../plugins/github/nodes";
+import * as graph from "../core/graph";
+import * as graphToMarkovChain from "../core/attribution/graphToMarkovChain";
+import * as markovChain from "../core/attribution/markovChain";
+import * as timelineCred from "../analysis/timeline/timelineCred";
+
+export default deepFreeze({
+  analysis: {
+    timeline: {
+      timelineCred,
+    },
+  },
+  core: {
+    address,
+    attribution: {
+      markovChain,
+      graphToMarkovChain,
+    },
+    graph,
+  },
+  plugins: {
+    github: {
+      declaration: githubDeclaration,
+      edges: githubEdges,
+      nodes: githubNodes,
+    },
+    discourse: {
+      address: discourseAddress,
+      declaration: discourseDeclaration,
+    },
+  },
+});
--- a/config/paths.js
+++ b/config/paths.js
@@ -27,6 +27,7 @@ module.exports = {
   // point within the build directory.
   backendEntryPoints: {
     sourcecred: resolveApp("src/cli/main.js"),
+    api: resolveApp("src/api/index.js"),
     //
     generateGithubGraphqlFlowTypes: resolveApp(
       "src/plugins/github/bin/generateGraphqlFlowTypes.js"


wchargin:
This should help unblock some Observable notebook work. In the wise
words of @decentralion, “oooOOOOooo”!

wchargin:
As written, this can’t be required from Observable, because it doesn’t
vendor its dependencies, and the dynamic `require`s inside the notebook
context don’t resolve with Observable’s magic logic. A fix for this is
to build without `externals: [nodeExternals()]`, which will require
splitting out a build config because we only want to do that for this
target.

I’ll make that build config change in (per @decentralion’s suggestion) a
separate PR.

wchargin:
Feel free to request additional API entry points on this PR as you need
them.

src/api/index.js
```diff
@@ -0,0 +1,44 @@
...
+    github: {
+      declaration: githubDeclaration,
+      edges: githubEdges,
+      nodes: githubNodes,
+    },
+    discourse: {
+      address: discourseAddress,
+      declaration: discourseDeclaration,
+    },
+  },
```
Beanow: Wondering if it makes sense to let this be a mixin step.
So the default export would look like:
```js
{
  ...core,
  plugins,
}
```

Beanow:
> To build, run yarn backend (or yarn backend --watch), then grab the
new bin/api.js file.

The convention I'm used to for bundled APIs it to be in `/dist` while `/bin` has executables for the CLI. Some smaller packages that are intended to be used just as a library, I've also seen bundled files go to the root `/index.js`.

Can we separate the output paths? When adding extra config changes like `nodeExternals` I think it will also make expectations clearer.

src/api/index.js
wchargin: I’m not sure what you’re suggesting. What’s a “mixin step”? Why would we
want to spread `core` into the export but store `plugins` on the
attribute intact?

This whole structure is mainly intended to serve as a bridge to analysis
in Observable notebooks, and is explicitly not under any API guarantees,
so I’m not super inclined to bikeshed about the exact shape of the
export.

Beanow: Well, from the core side of things the use-case of exposing the essentials as a library is something we're going to want for the long term. Plugins on the other hand are kind of hitching a ride on this API while they're still baked into the monolith.

So to me it would make sense we declare an API for _just_ the core components so we can refine this over time to be an API we're happy with. The plugin API doesn't need that treatment for the foreseeable future.

Hence I'm suggesting we make that difference explicit with separate declarations, and letting the plugins hitch a ride at the final step. Call it mixing in, or composing. It's typically the only thing an `index.js` file does.

wchargin: I still don’t understand what change is actually being proposed. We do
need to export both the core and the plugins for now, because plugin
declarations are required to get the weights, which is required for
analysis. As written, the default export here exactly mirrors the
directory structure, which seems like a simple and uncomplicated way of
exposing our existing code without introducing additional friction or
cognitive overhead.

The one person who has used this for actual work is @decentralion, so
I’ll let them weigh in on whether they think that the plugin exports
need to be somehow separated from the rest.

PR #1531, opened 2020-01-07
UniRef: implement Initiatives reference detector

Beanow:
Giving this detector priority, would allow us to consider a Discourse Topic URL (or any other tracker) a reference to an Initiative node, when there is a valid Initiative being linked.

Test plan: `yarn unit`

### DIFF
--- a/src/plugins/initiatives/createGraph.js
+++ b/src/plugins/initiatives/createGraph.js
@@ -19,7 +19,7 @@ import {
   championsEdgeType,
 } from "./declaration";
 
-function initiativeAddress(initiative: Initiative): NodeAddressT {
+export function initiativeAddress(initiative: Initiative): NodeAddressT {
   return NodeAddress.append(
     initiativeNodeType.prefix,
     ...NodeAddress.toParts(initiative.tracker)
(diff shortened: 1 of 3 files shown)


Beanow:
> now that we're moving away from initiatives-on-discourse, do we still want this?

Good question. The design for the file-based loading is still unclear on reference detection.

In theory this detector will work for any "tracker" that we can have references for (meaning, canonical URLs and nodes in the graph).

Having a custom UI and detecting references to that, we wouldn't need this. As we can directly map without a "base".

We could also decide that every initiative config entry must have a discussion URL.
That discussion URL could be used as the "tracker", so referencing the Discourse thread would be a reference to the Initiative.

decentralion:
I'm disinclined to require that every initiative have a discussion URL. At the very least, the discussion urls won't be unique to a particular initiative, since I expect we may have threads that spawn multiple initiatives.

For now, I say let's treat the discussion thread as a regular contribution to the initiative, and not assume that every initiative will have a standalone discussion thread. In which case, I think we can close this for now.

Beanow:
Agreed, we'll settle on an implementation when the file-based design is more clear.

PR #1532, opened 2020-01-07
WIP: create GitHub reference detector

Beanow:
Want to get your feedback @decentralion @wchargin, as I'm not too familiar with the thoughts behind `RelationalView`. Not sure yet whether `urlReferenceMap()` makes sense here.

What I've done in a prototype with this, is to add all loaded GitHub repositories into a single "super view" to create this urlReferenceMap and use that for `fromRelationalView` to make the reference detection map.

### DIFF
--- a/src/plugins/github/referenceDetector.js
+++ b/src/plugins/github/referenceDetector.js
@@ -1,0 +1,12 @@
+// @flow
+
+import {MappedReferenceDetector} from "../../core/references";
+import {RelationalView} from "./relationalView";
+
+export function fromRelationalView(
+  view: RelationalView
+): GithubReferenceDetector {
+  return new GithubReferenceDetector(view.urlReferenceMap());
+}
+
+export const GithubReferenceDetector = MappedReferenceDetector;
--- a/src/plugins/github/referenceDetector.test.js
+++ b/src/plugins/github/referenceDetector.test.js
@@ -1,0 +1,35 @@
+// @flow
+
+import {RelationalView} from "./relationalView";
+import {GithubReferenceDetector, fromRelationalView} from "./referenceDetector";
+import {MappedReferenceDetector} from "../../core/references";
+
+describe("plugins/github/referenceDetector", () => {
+  describe("GithubReferenceDetector", () => {
+    it("should be a MappedReferenceDetector", () => {
+      // Given
+      const map = new Map();
+
+      // When
+      const refs = new GithubReferenceDetector(map);
+
+      // Then
+      expect(refs).toBeInstanceOf(MappedReferenceDetector);
+    });
+  });
+
+  describe("fromRelationalView", () => {
+    it("should use urlReferenceMap to create the instance", () => {
+      // Given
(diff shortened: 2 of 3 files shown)


decentralion:
The code in this commit looks fine to me. However:

> What I've done in a prototype with this, is to add all loaded GitHub repositories into a single "super view" to create this urlReferenceMap and use that for fromRelationalView to make the reference detection map.

This is not safe for relational views, see documentation here:

https://github.com/sourcecred/sourcecred/blob/fa00751cd47cada17a2699cb9b54a91b16af3d2b/src/plugins/github/relationalView.js#L59-L67

I wonder if we can build a reference pipeline that hits the underlying sqlite db directly? Then we could start migrating away from RelationalView (which should be deprecated now that we have an actual relational database).

Beanow:
Great point, that's probably a reason to write something else.
In the end I'm only interested in a full list of `ReferentEntity` and their address and url accessor methods.

A repository class to query the DB for this makes a lot of sense.

Beanow:
I've messed around with using DB queries to get the references without using `RelationalView`.

https://github.com/sourcecred/sourcecred/blob/a368002bc185355fedc6467e81d9856b41aa55fb/src/plugins/github/referenceDetector.js

It does a "cross section" of the database, one typename at a time. So by ordering these correctly, you can be sure you've seen all the parents as well.

To be able to create a `NodeAddress`, we sometimes need to store information about the parent.
For example an `Issue` needs to know the repo owner and repo name.
An `IssueComment` needs to know the issue number, repo owner and repo name, etc.

I'm using an "infix" map for this, because the addresses follow a specific pattern:
`[owner, name]` becomes `[owner, name, number]` becomes `[owner, name, number, id]`
So instead of storing structured objects, I'm storing `string[]` and append values to that for each child.

---

I've stopped there, because I'm wondering whether we'll have much use for adding this approach vs using `RelationalView`.

One reason being, the mirror DBs have separate files for each repository. So much like `RelationalView`, we would need to merge the results from several repos.

The other being, I'm not sure these queries will see any other use-case than what I'm doing here. The time investment to do the actual consistency checking, typing and error handling for these queries may be wasteful.

So I'm inclined to `RelationalView` per repo, as it was intended and let `Map` deduplicate for me. As the only overlap should be a handful of `User` types.

decentralion:
Thanks for the investigation, @Beanow. Based on the prototype in the linked commit, I agree that we should continue using relationalView--it would take a lot more effort to get confident in the new approach, and since we aren't currently planning to invest the effort to wholly migrate off RelationalView, we'd end up with two very different but overlapping ways of accessing the same data, which is a lot of technical debt.

Using one relationalView per repo and building an explicit structure for resolving references across them makes sense to me. We should probably change relationalView so that it doesn't have an addRepository method and instead just takes the repository once in the constructor... it would discourage the unsupported behavior at the type level. Would be a good first issue: see #1539

Beanow:
Closing in favor of #1542 and #1543.
Thanks for the feedback :]

PR #1533, opened 2020-01-07
Remove core/pagerankGraph

decentralion:
The `pagerankGraph` module was an attempt to do a better job of
co-ordinating the data needed to run Pagerank, by wrapping the Graph
class alongside context on edge weights, etc. However, it was obsoleted
by work on TimelineCred. Thus, we can remove it entirely. I intend to
make another attempt at collecting all the data needed for cred analysis
in a way that doesn't couple with plugin code, and this time it will be
timeline-aware.

Test plan: `yarn test`


Beanow:
I haven't got a lot of context on this so just going to ask. For public record :]

This isn't currently used by either the legacy or timeline approach, but was intended as an abstraction to annotate the graph?

decentralion:
One thing that has frustrated me for a long time is that we don't have a clean data structure that packages a Graph alongside the context needed to compute scores for it. E.g. the node and edge weights don't get stored cleanly alongside the graph.

PageRankGraph was my first attempt to fix this, by adding a class which bundles a Graph with associated weights and scores. However, it was obsoleted by switching to TimelineCred, which has a different type signature.

decentralion:
> ```js
> export type WeightedEdge = {|
>   +edge: Edge,
>   +weight: EdgeWeight,
> |};
> ```

It was a reasonable type signature before timeline cred. Now that cred has a temporal component, the same edge may have different weights in different times. Thus, we can confidently retire this type.

PR #1534, opened 2020-01-07
Remove the repoIdRegistry

decentralion:
Before we added the concept of "SourceCred Projects", we tracked cred
instances via their GitHub repostiory id. The replacement for this
system was added in #1238, I missed the RepoIdRegistry in the cleanup.

This commit removes all code pertaining to the now-obsolete
RepoIdRegistry.

Test plan:
- `yarn test --full` passes
- manual inspection of `yarn start`; it still loads properly
- manual inspection of the output for build_static_site.sh
- `git grep repoIdRegistry` returns no hits


src/core/repoIdRegistry.js
```diff
@@ -1,115 +0,0 @@
-// @flow
-
-/**
- * The RepoIdRegistry is a small JSON file that SourceCred uses to track which
- * RepoIds have been loaded.
- *
- * The registry consists of Entries. The entry contains the RepoId of the
- * loaded repository, and may contain additional metadata (e.g. the last-loaded
- * timestamp).
- */
```
Beanow: This summary makes this review really easy :smile:

PR #1536, opened 2020-01-08
api: extract browser-friendly build target

wchargin:
Summary:
This re-packages the build for the internal APIs exposed under #1526 to
be more browser-friendly. Removing `target: "node"` (and adding an
explicit `globalObject: "this"` for best-effort cross-compatibility) is
the biggest change from the backend build; removing all the extra
loaders and static site generation is the biggest change from the
frontend build.

This build configuration is forked from `webpack.config.backend.js`.

Test Plan:
Run `yarn api`, then upload the contents of `dist/api.js` to an
Observable notebook and require it as an ES module. Verify that the
SourceCred APIs are exposed: e.g., `sourcecred.core.graph.Graph` should
be a valid constructor.

wchargin-branch: api-build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@
 # backend
 /bin
 
+# api
+/dist
+
 # misc
 data/
 .token
--- a/config/webpack.config.api.js
+++ b/config/webpack.config.api.js
@@ -1,0 +1,75 @@
+// @flow
+
+// Do this as the first thing so that any code reading it knows the right env.
+process.env.NODE_ENV = process.env.NODE_ENV || "development";
+process.env.BABEL_ENV = process.env.NODE_ENV;
+
+const webpack = require("webpack");
+const ModuleScopePlugin = require("react-dev-utils/ModuleScopePlugin");
+const RemoveBuildDirectoryPlugin = require("./RemoveBuildDirectoryPlugin");
+const paths = require("./paths");
+const getClientEnvironment = require("./env");
+
+const env = getClientEnvironment(null);
+
+module.exports = {
+  // Don't attempt to continue if there are any errors.
+  bail: true,
+  node: {
+    // Don't munge `__dirname` and `__filename`.
+    // https://github.com/webpack/webpack/issues/1599#issuecomment-186841345
+    __dirname: false,
+    __filename: false,
+  },
+  entry: {
+    api: paths.apiIndexJs,
+  },
+  output: {
+    path: paths.apiBuild,
+    // Generated JS file names (with nested folders).
+    // There will be one main bundle, and one file per asynchronous chunk.
+    // We don't currently advertise code splitting but Webpack supports it.
+    filename: "[name].js",
+    chunkFilename: "[name].[chunkhash:8].chunk.js",
+    libraryTarget: "umd",
+    library: "sourcecred",
+    // Use `this` for compatibility with both Node and browser.
+    globalObject: "this",
+  },
+  resolve: {
+    extensions: [".js", ".json"],
+    plugins: [
+      // Prevents users from importing files from outside of src/ (or node_modules/).
+      // This often causes confusion because we only process files within src/ with babel.
+      // To fix this, we prevent you from importing files out of src/ -- if you'd like to,
+      // please link the files into your node_modules/ and let module-resolution kick in.
+      // Make sure your source files are compiled, as they will not be processed in any way.
+      new ModuleScopePlugin(paths.appSrc, [paths.appPackageJson]),
+    ],
+  },
+  module: {
+    strictExportPresence: true,
+    rules: [
+      {
+        // "oneOf" will traverse all following loaders until one will
+        // match the requirements. If no loader matches, it will fail.
+        oneOf: [
+          // Process JS with Babel.
+          {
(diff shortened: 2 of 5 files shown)


config/webpack.config.api.js
```diff
@@ -0,0 +1,74 @@
...
+  module: {
+    strictExportPresence: true,
+    rules: [
+      {
+        // "oneOf" will traverse all following loaders until one will
+        // match the requirements. If no loader matches, it will fail.
+        oneOf: [
+          // Process JS with Babel.
+          {
+            test: /\.(js|jsx|mjs)$/,
```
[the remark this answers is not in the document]
wchargin: If you wish to review this code, you may find it simplest to look at the
diff from the forked version:

```
diff -u config/webpack.config.backend.js config/webpack.config.api.js
```

This loader is uniform across all our loaders, which I think is good for
the principle of least surprise. It would be surprising if a utility
file should compilation errors in one build but not the others because
it’s being passed through Babel only in some configs.

That said, this exact kind of thing is why (at my suggestion) this
repository uses `.js` extensions for all files. Whether a file “should”
be `.js` or `.jsx` is often unclear, and the distinction doesn’t really
provide any value. So all of our stuff is `.js`, and this loader test is
just copied all the way back from the standard configs.

package.json
```diff
@@ -92,6 +92,7 @@
     "start": "NODE_ENV=development webpack-dev-server --config config/webpack.config.web.js",
     "build": "NODE_ENV=production webpack --config config/webpack.config.web.js",
     "backend": "NODE_ENV=development webpack --config config/webpack.config.backend.js",
+    "api": "webpack --config config/webpack.config.api.js",
```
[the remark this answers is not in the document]
wchargin: Yeah, I was also noting that the naming is a bit non-uniform; we might
prefer something more like `yarn build-site`, `yarn build-bin`, and
`yarn build-api`/`yarn build-library`. But, as you noted, now isn’t the
right time to rename all the rest.

config/webpack.config.api.js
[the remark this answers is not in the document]
Beanow: > which I think is good for the principle of least surprise.

Makes sense :]

> Whether a file “should” be .js or .jsx is often unclear,

Yeah the double edged sword of JSX. The functional components style with state reduction has been my favorite way to separate concerns.

Anyway, since this is less surprising for the current setup, removing JSX support would also be a thing for when the frontend has moved it's own module.

config/webpack.config.api.js
```diff
@@ -0,0 +1,74 @@
...
+    api: paths.apiIndexJs,
+  },
+  output: {
+    path: paths.apiBuild,
+    // Generated JS file names (with nested folders).
+    // There will be one main bundle, and one file per asynchronous chunk.
+    // We don't currently advertise code splitting but Webpack supports it.
+    filename: "[name].js",
+    chunkFilename: "[name].[chunkhash:8].chunk.js",
+    libraryTarget: "umd",
```
Beanow: Using the `umd` library target, when not using a require function like Observables does but including it through a `<script>` tag, this will set a global variable. Currently this global is `window.default`. Because `index.js` exports `default`.

We can define `library: "sourcecred",` to namespace this. Though we will still have the default in between.
For example `window.sourcecred.default.core.graph.Graph`.

A way around that would be export several const in `index.js`. Which would also address my comment in #1526

```js

export const analysis = deepFreeze({
  timeline: {
    timelineCred,
  },
});

export const core = deepFreeze({
  address,
  attribution: {
    markovChain,
    graphToMarkovChain,
  },
  graph,
});

export const plugins = deepFreeze({
  github: {
    declaration: githubDeclaration,
    edges: githubEdges,
    nodes: githubNodes,
  },
  discourse: {
    address: discourseAddress,
    declaration: discourseDeclaration,
  },
});
```

Combined with `library: "sourcecred",` and a tweak to the observable example script that should be nice and consistent :]

.gitignore
```diff
@@ -12,6 +12,9 @@
 # backend
 /bin
 
+# api
+/dist
```
Beanow: I think the `/dist` path should also be in the `.prettierignore`. (And probably `package.json`'s `files` array.)

wchargin: Added to `.prettierignore`; thanks! I had deliberately omitted this from
`"files"` in `package.json` because this should not be included when
this package is installed as a dependency, because it is unstable and
only used for ad hoc testing and exploration in Observable notebooks.

config/webpack.config.api.js
wchargin: Noted. This doesn’t bother me because it’s not actually intended to be
used directly in a browser, so setting `default` isn’t a problem, and
splitting this out into separate exports would heterogenize the shape in
a way that really doesn’t align with the semantics and also introduces
extra work and typing, both here and in the Observable script. I think
that it’s fine as is.

Beanow: In that case, only setting `library: "sourcecred",` doesn't change the current behavior for Observables, but prevents loading as `window.default`.

wchargin: Okay, sure: added `library: "sourcecred"`.

PR #1537, opened 2020-01-08
api: add server for easy Observable integration

wchargin:
Summary:
This commit adds a simple Python server for connecting the output of
`yarn api` (or `yarn api --watch`) to an observable notebook. We need a
custom server rather than just `python3 -m http.server` to send CORS
headers properly. This server enables a very tight loop from editing
SourceCred core code on your local filesystem to seeing live updates in
an Observable notebook, with latency on the order of one second.

Test Plan:
Run `yarn api --watch` in the background. Launch the new API server.
Navigate to <https://observablehq.com/demo>. Copy the two paragraphs of
Observable code from `scripts/serve_api.py` into _separate_ Observable
cells, and execute them. Note that `myGraph` becomes a valid SourceCred
graph. Modify `src/core/graph.js` to add `this._aaa = 123;` to the top
of the `Graph` constructor. Re-execute the first Observable cell (the
one that loads the SourceCred module), and note that `myGraph` updates
to include the new `_aaa` attribute:

![Screenshot of Observable notebook after test plan][ss]

[ss]: https://user-images.githubusercontent.com/4317806/71958748-dddf8680-31a5-11ea-9016-5df76ceeea46.png

wchargin-branch: api-server

### DIFF
--- a/scripts/serve_api.py
+++ b/scripts/serve_api.py
@@ -1,0 +1,71 @@
+"""Serve SourceCred API, for consumption by Observable notebooks.
+
+Usage:
+
+    python3 ./scripts/serve_api.py [PORT]
+
+from the SourceCred repository root. Omit the port to try a default, or
+specify port 0 to pick a free port automatically.
+
+This server binds to `localhost`, which is a loopback address only
+accessible from your machine, not the local network.
+
+Corresponding Observable code:
+
+    sourcecred = {
+      const server = "http://localhost:9009";
+      const blob = await fetch(server).then((r) => r.blob());
+      const esModule = await require(URL.createObjectURL(blob));
+      return esModule.default;
+    }
+
+    myGraph = new sourcecred.core.graph.Graph()  // e.g.
+
+This server serves the same response for all requests.
+
+For best results, run concurrently with `yarn api --watch`. After
+changing SourceCred code, just re-execute the definition of the
+`sourcecred` module in the Observable notebook to live-load the new code
+and update your notebook.
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+import sys
+from wsgiref import simple_server
+
+
+_API_FILE = os.path.join("dist", "api.js")
+_DEFAULT_PORT = 9009
+
+
+def _app(environ, start_response):
+    # This will 500 if the file can't be read, which is fine.
+    with open(_API_FILE, "rb") as infile:
+        # Re-read the file every time to account for updates.
+        contents = infile.read()
+    headers = [
+        ("Access-Control-Allow-Origin", "*"),
+        ("Content-Type", "application/javascript; charset=UTF-8"),
+        ("X-Content-Type-Options", "nosniff"),
+    ]
+    start_response("200 OK", headers)
+    return (contents,)
+
+
+def _main():
+    port = int(sys.argv[1]) if len(sys.argv) > 1 else _DEFAULT_PORT
+    server = simple_server.make_server("localhost", port, _app)
+    print("Serving on port %d" % server.server_port)
+    try:
+        server.serve_forever()
+    except KeyboardInterrupt:
+        print()
+        pass
+
+
+if __name__ == "__main__":
+    _main()


scripts/serve_api.py
```diff
@@ -0,0 +1,71 @@
...
+accessible from your machine, not the local network.
+
+Corresponding Observable code:
+
+    sourcecred = {
+      const server = "http://localhost:7007";
+      const blob = await fetch(server).then((r) => r.blob());
+      const esModule = await require(URL.createObjectURL(blob));
+      return esModule.default;
+    }
```
Beanow: Can you add a note why are we using the blob URL?
I'm guessing it's to force we're not caching the file?
Similar to something like 
```js
require(`http://localhost:7070/api.js?t=${Date.now()}`)
```

scripts/serve_api.py
```diff
@@ -0,0 +1,71 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import os
+import sys
+from wsgiref import simple_server
+
+
+_API_FILE = os.path.join("dist", "api.js")
+_DEFAULT_PORT = 7007
```
decentralion: nit: mind changing the default port? maybe 9009 or 7008?

I already use 7007 a lot.

scripts/serve_api.py
wchargin: The primary reason is to prevent caching. A secondary reason is that it
hints at what to do if you’re serving from a CDN that doesn’t serve the
proper MIME types, like GitHub’s raw file view. (Namely, just wrap this
in `new Blob([b], {type: "application/javascript"})`. You can’t do that
if it’s a direct `require("http://...")`. I thought that this was worth
including because @decentralion _was_ trying to require directly from
GitHub—it’s a natural thing to do—and was running into errors.)

scripts/serve_api.py
wchargin: Sure, will go with 9009. (I picked 7007 because it’s what you used when
running my first prototype of this server, but of course fine to amend!)

scripts/serve_api.py
Beanow: Good tip, I hadn't tried executing js files from GitHub in this way, only json. So I hadn't run into the mime type issues.

scripts/serve_api.py
```diff
@@ -0,0 +1,71 @@
...
+specify port 0 to pick a free port automatically.
+
+This server binds to `localhost`, which is a loopback address only
+accessible from your machine, not the local network.
+
+Corresponding Observable code:
+
+    sourcecred = {
+      const server = "http://localhost:9009";
+      const blob = await fetch(server).then((r) => r.blob());
```
nothingismagick: Just chiming in here: don't forget to catch the await for an error!
```
await fetch(server)
  .then((r) => r.blob())
  .catch((e) => {
    throw new Error(e);
  })
```

Beanow:
> I’ve used the Webpack dev server for things like this in the past, but
> run into issues where the automatically injected frame-based reloading
> mechanisms break things down the pipeline. (Adding `inline: false`, as
> in your example, seems like it _should_ fix this, but from experience it
> sometimes doesn’t!)
> 
> If we agree that this Python script is more trouble than it’s worth, and
> the Webpack dev server appears to be behaving, then I’m open to dropping
> this.

Well webpack-dev-server is the solution that doesn't require any additional dependencies.
If it sprinkles in too much magic to be well suited, we can add a simpler file serving dependency.

```js
// yarn add -D st
"api:serve": "st --no-cache --cors --port 9009 --dir dist",

sourcecred = {
  const server = "http://localhost:9009/api.js";
  const blob = await fetch(server).then((r) => r.blob());
  const esModule = await require(URL.createObjectURL(blob));
  return esModule.default;
}
```

The reason I'm suggesting not to use Python is developer familiarity. Since this is a node project, it would be less friction to stay with it's ecosystem I think.

scripts/serve_api.py
Beanow: I'm all for error handling, but is there a benefit to throwing the error again vs letting it go as an unhandled error?

nothingismagick: Unhandled errors will at some point be deprecated by Node. Potentially with 14. And since this is just a suggestion, my feeling is that its best to lead by example.

Further, it allows you to swallow the error and do something else than merely throwing an unhandled error as an exception.

nothingismagick:
I have to agree with avoiding using python if at all possible. My experience in polyglot systems is that they are dastardly hard to debug, and it does, like @Beanow said, introduce friction and cognitive load.

scripts/serve_api.py
Beanow: > Unhandled errors will at some point be deprecated by Node. Potentially with 14.

Interesting! Hopefully one day the stock Promises will be as useful as something like https://github.com/fluture-js/Fluture.

Had forced error handling since inception. And also makes a distinction between expected (a rejection value) and unexpected errors (divide by zero in your map function).

wchargin: Isn’t `p.catch((e) => { throw new Error(e); })` exactly the same as `p`?
Errors thrown from within promise callbacks become rejections. I’m not
sure what you’re suggesting.

Recall also that this code is executed by Observable, not Node. If there
is an error, it will be clearly displayed in the UI, and the user will
debug it. There is nothing useful to do by handling the error.

scripts/serve_api.py
wchargin: I will note that if one doesn’t care about these issues, one can write:

```
sourcecred = (await require("http:localhost:9009")).default
```

This works as long as the MIME type served is correct, and is a bit
shorter. So if @decentralion prefers I can use that as the recommended
incantation.

Beanow:
> I’m fairly strongly disinclined to add any dependencies for this task.
> The `st` module has 15 transitive dependencies.

I've looked at several options. Including writing a script nearly indentical to the python one.

This one looks good to me because:
- It correctly sets the javascript `Content-Type`.
- Supports a CORS wildcard.
- There's 1 line of code we need to maintain.

I would be OK with taking on the maintenance of a hand-crafted static file server to reduce the dependency footprint _if it wasn't for development purposes_.

Currently we're requiring end-users to install all the dev dependencies in order to build the frontend. When we can avoid this I think the `st` dependency should be a non-issue.

scripts/serve_api.py
Beanow: Just bumped into a service that promises the solve the mime issue.
https://raw.githack.com/

Beanow:
Sorry for my bikeshedding and ["Not how I would have done it"](https://github.com/sourcecred/docs/blob/fdb78c28bca31aabffd7700ee9a78bfd79a79ccc/culture/review_culture.md#avoid-not-how-i-would-have-done-it) on this PR @wchargin.

If there's an interest for one of the alternatives I'm happy to write up a PR for that.

wchargin:
> Sorry for my bikeshedding and ["Not how I would have done it"](https://github.com/sourcecred/docs/blob/fdb78c28bca31aabffd7700ee9a78bfd79a79ccc/culture/review_culture.md#avoid-not-how-i-would-have-done-it) on this PR @wchargin.

(I missed your comment earlier, but) thank you! I appreciate it.

PR #1538, opened 2020-01-09
Refactor Weights into a class

decentralion:
Currently, Weights are tracked as a raw object, and there are a lot of
'floating' functions in the weights module which allow copying or
serializing them. This commit reorganizes the module around a Weights
class, which has attached methods instead of floating functions.

The resulting APIs are isomorphic, but consumers of Weights need only
import a single class rather than a smattering of functions.

This commit is a prelude to future work around packaging Weights with
Graphs, both to make it possible to compute scores without needing
plugin declarations, and so that plugins can specify weights for
specific nodes (e.g. initiatives).

Test plan: `yarn test` is sufficient for this refactor.

### DIFF
--- a/src/analysis/timeline/params.js
+++ b/src/analysis/timeline/params.js
@@ -1,12 +1,6 @@
 // @flow
 
-import {
-  type Weights,
-  type WeightsJSON,
-  toJSON as weightsToJSON,
-  fromJSON as weightsFromJSON,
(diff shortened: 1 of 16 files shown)


decentralion:
Thanks for the feedback @Beanow and @wchargin. Retiring this PR based on your comments.

PR #1541, opened 2020-01-10
CI: make sure we log in for tagged-releases

Beanow:
Over at https://github.com/sourcecred/sourcecred/pull/1514#discussion_r365443659
I suggested adding `deploy: false` to be less surprising with the tags we're pushing to DockerHub.
As mentioned in #1512, we actually need `publish: true` to log in with DockerHub.

Test plan:
Compare with a push to `master`, to see our credentials are correct for pushing `:dev`.
https://circleci.com/gh/sourcecred/sourcecred/4422

With the tagged deploy.
https://circleci.com/gh/sourcecred/sourcecred/4426

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,6 @@ workflows:
             - test
             - test_full
             - test_full_10
-          deploy: false
           image: sourcecred/sourcecred
           tag: latest
           cache_from: node:12,sourcecred/sourcecred:dev


Beanow:
@vsoch from https://github.com/sourcecred/sourcecred/issues/1512#issuecomment-573236521

> Indeed if we use the orb and put deploy to false it won't log us in, but we need to do that to have more explicit control about the pushes. Why not just add:
> 
> ```shell
> docker login -u "${DOCKER_LOGIN}" -p "${DOCKER_PASSWORD}" docker.io
> ```

> If deploy is true, then we would just be pushing sourcecred/sourcecred:latest, so I think you could set deploy to true instead and then manually just push the custom tag. If you think it's unclear for the developer, we could add a note (comment) about that.

You're right we can use either option there.
In this case I'm pushing `:latest` twice. Once explicitly and once from the orb.

Open to suggestions from @decentralion if one would be preferable / less confusing. :smiley:
For now just want to make sure it works.

vsoch:
It's redundant to have it push twice (with deploy as true and then the manual line) but I think it's actually okay because it's more transparent for the developer. There isn't actually any issue with doing the push twice because the layers will already be found to exist and won't be pushed again. Solid and quick work, @Beanow !

PR #1542, opened 2020-01-11
UniRef: add urlReferenceMap to RelationalView

Beanow:
`RelationalView` provides easy access to `ReferentEntities`, which we can use for reference detection. The map produced by `urlReferenceMap` can be used easily by a `MappedReferenceDetector`.

As discussed in #1532 we'll use `RelationalView` for this before deprecating it.

Test plan: `yarn unit`

### DIFF
--- a/src/plugins/github/relationalView.js
+++ b/src/plugins/github/relationalView.js
@@ -1,5 +1,7 @@
 // @flow
 
+import {type URL} from "../../core/references";
+import {NodeAddress, type NodeAddressT} from "../../core/graph";
 import {toCompat, fromCompat, type Compatible} from "../../util/compat";
 import stringify from "json-stable-stringify";
 import {parseReferences} from "./parseReferences";
@@ -513,6 +515,30 @@ export class RelationalView {
       this._userlikes.set(N.toRaw(address), entry);
       return [address];
     }
+  }
+
+  /**
+   * Creates a Map<URL, NodeAddressT> for each ReferentEntity in this view.
+   * Note: duplicates are accepted within one view. However for any URL, the
+   * corresponding N.RawAddress should be the same, or we'll throw an error.
+   */
+  urlReferenceMap(): Map<URL, NodeAddressT> {
+    const refToAddress: Map<URL, NodeAddressT> = new Map();
+    for (const e: ReferentEntity of this.referentEntities()) {
+      const addr = N.toRaw(e.address());
+      const url = e.url();
+      const existing = refToAddress.get(url);
+      if (existing && existing != addr) {
+        throw new Error(dedent`\
+          An entry for ${url} already existed, but with a different NodeAddressT.
+          This is probably a bug with SourceCred. Please report it on GitHub.
+          Old: ${NodeAddress.toString(existing)}
+          New: ${NodeAddress.toString(addr)}
+        `);
+      }
+      refToAddress.set(url, addr);
+    }
+    return refToAddress;
   }
 
   _addReferences() {
--- a/src/plugins/github/relationalView.test.js
+++ b/src/plugins/github/relationalView.test.js
@@ -1,5 +1,6 @@
 // @flow
 
+import {NodeAddress} from "../../core/graph";
 import * as R from "./relationalView";
 import * as N from "./nodes";
 import {exampleRepository, exampleRelationalView} from "./example/example";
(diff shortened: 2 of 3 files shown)


decentralion:
One more thing: in the RelationalView test file, we have individual examples of each entity type, e.g. here:

https://github.com/sourcecred/sourcecred/blob/adec1b0b5d40ca364873b0b725015b82ae6e515d/src/plugins/github/relationalView.test.js#L116

rather than having a snapshot test for the entire urlReferenceMap, lets just check that these entities are present in the map as expected. cuts down on the snapshot size (huge snapshots are an antipattern, as a reviewer my eyes glaze over). it feels like we shouldn't really need the snapshot at all.

Beanow:
> One more thing: in the RelationalView test file, we have individual examples of each entity type, e.g. here:

Good point. I'm not sure how to add a smaller test repository, without using "private" methods on RelationalView (which is our test subject, so I'm hesitant to do that), other than hand-crafting a new repository json.

src/plugins/github/relationalView.js
```diff
@@ -515,6 +515,30 @@ export class RelationalView {
     }
   }
 
+  /**
+   * Creates a Map<string, N.RawAddress> for each ReferentEntity in this view.
+   * Note: duplicates are accepted within one view. However for any URL, the
+   * corresponding N.RawAddress should be the same, or we'll throw an error.
+   */
+  urlReferenceMap() {
```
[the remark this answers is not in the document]
Beanow: Done, switched the aliases over to `<URL, NodeAddressT>` as well to be more in line with the rest of reference detection.

src/plugins/github/relationalView.test.js
```diff
@@ -390,6 +391,49 @@ describe("plugins/github/relationalView", () => {
...
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"REPO\\",\\"sourcecred-test\\",\\"example-github\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"ISSUE\\",\\"sourcecred-test\\",\\"example-github\\",\\"1\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"PULL\\",\\"sourcecred-test\\",\\"example-github\\",\\"3\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"REVIEW\\",\\"sourcecred-test\\",\\"example-github\\",\\"5\\",\\"100313899\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"COMMENT\\",\\"ISSUE\\",\\"sourcecred-test\\",\\"example-github\\",\\"2\\",\\"373768703\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"COMMENT\\",\\"PULL\\",\\"sourcecred-test\\",\\"example-github\\",\\"3\\",\\"369162222\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"COMMENT\\",\\"REVIEW\\",\\"sourcecred-test\\",\\"example-github\\",\\"5\\",\\"100313899\\",\\"171460198\\"]",
+          "NodeAddress[\\"sourcecred\\",\\"github\\",\\"COMMIT\\",\\"MDY6Q29tbWl0MTIzMjU1MDA2OjBhMjIzMzQ2YjRlNmRlYzAxMjdiMWU2YWE4OTJjNGVlMDQyNGI2NmE=\\"]",
+        ]
+      `);
```
Beanow: @decentralion How is this as an alternative to the mega snapshot?

decentralion: This is basically what I had in mind. However, to make it easier to read, I recommend creating an array of {url, address} objects, and snapshotting that array. that way the reader can directly correlate the address to the url without needing to mentally track an index in both arrays.

Beanow: I tried a few options.

```js
type MapBased = Map<string, ?string>;
type ObjectBased = {url: string, address: ?string}[];
type TupleBased = [string, ?string][];
```

And found the tuple version most easy to read in the snapshot (how prettier stores it) and test error message context. All of them were either a lot larger than the inline snapshot, so I've moved that back to the snapshot file.

Beanow:
Requesting review for the test readability change anyway, as I don't see a need to rush merge this :]

PR #1543, opened 2020-01-11
UniRef: add GithubReferenceDetector

Beanow:
Depends on #1542

As a `RelationalView` is not designed for multiple repositories, we should implement our own merging of mappings obtained from `RelationalViews`.

`fromRelationalViews` is a factory function which does this for us. And by accepting an array of `RelationalViews` it's more apparent it should be used this way.

Test plan: `yarn unit`

### DIFF
--- a/src/plugins/github/referenceDetector.js
+++ b/src/plugins/github/referenceDetector.js
@@ -1,0 +1,37 @@
+// @flow
+
+import dedent from "../../util/dedent";
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
+import {MappedReferenceDetector, type URL} from "../../core/references";
+import {RelationalView} from "./relationalView";
+
+/**
+ * Builds a GithubReferenceDetector using multiple RelationalView.
+ * As RelationalView should only be used for one repository at a time, you will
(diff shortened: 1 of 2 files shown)


src/plugins/github/referenceDetector.js
```diff
@@ -0,0 +1,37 @@
...
+): GithubReferenceDetector {
+  const map: Map<URL, NodeAddressT> = new Map();
+  for (const view of views) {
+    for (const [url, addr] of view.urlReferenceMap().entries()) {
+      const existing = map.get(url);
+      if (existing && existing != addr) {
+        throw new Error(dedent`\
+          An entry for ${url} already existed, but with a different NodeAddressT.
+          This is probably a bug with SourceCred. Please report it on GitHub.
+          Old: ${existing}
```
decentralion: if you want the address to be readable for humans, it's best to use this method which pretty prints so we dont put nul bytes in the error message.

https://github.com/sourcecred/sourcecred/blob/c1f123f3522e8cab4c83beb74e775119fd93a049/src/core/address.js#L43-L48

Beanow: Applied for both PRs

PR #1560, opened 2020-01-14
Update distributionToCred contract

### LINKED ISSUE #1317: Blank weeks at the start of timeline cred
I've noticed this loading `@sfosc` with the `v0.4.0` release. There are several blank weeks included before the first contributions.

Contributions start in "The week of November 25" yet the first interval is:
```js
{
  "endTimeMs": 1540080000000,
  "startTimeMs": 1539475200000
}

new Date(1539475200000).toISOString()
// "2018-10-14T00:00:00.000Z"
new Date(1540080000000).toISOString()
// "2018-10-21T00:00:00.000Z"
```

All users, including adamhjk who created the org and repository, have nulls for their scores in these intervals.

```js
{
  "id": "adamhjk",
  "intervalCred": [
    null,
    null,
    null,
    null,
    null,
    null,
    8.03125,
    14.953116717673753,
    28.867859281699292,
    61.841989977282196,
    // ...
    0.6651734167260511
  ],
  "totalCred": 245.39729499182977
},
```

Something similar seems to be going on for https://sourcecred.io/cred/timeline/@libp2p/

BrianLitwin:
Resolves #1317

Updates timeline cred to handle the case where the scoring nodes' total
cred sums to zero in an interval. In practice, we've encountered this
circumstance when a github.io repository contains timestamps that
predates any User's contributions by several weeks, such as
sfosc.github.io.

Test Plan:

- Added a test case to handle this circumstance
- Updated a test case per discussion on #1317 to return an empty array
when there are no scoring nodes passed to the function, so that we handle
these cases consistently.

Also loaded sfosc.github.io and observed that the cred output appeared
to match expectations and didn't contain any `NaN` or `Infinity` values
as it did before.

### DIFF
--- a/src/analysis/timeline/distributionToCred.js
+++ b/src/analysis/timeline/distributionToCred.js
@@ -30,6 +30,13 @@ export type FullTimelineCred = $ReadOnlyArray<{|
  * This implementation normalizes the scores so that in each interval, the
  * total score of every node matching scoringNodePrefix is equal to the
  * interval's weight.
+ *
+ * Edge cases:
+ *     - If in an interval the sum of the scoring nodes' distribution is 0,
+ *       return a total cred score of 0 for all nodes in the interval.
+ *     - If none of the nodes match a scoring node prefix, return
+ *       a total cred score of 0 for all nodes in all intervals.
+ *
  */
 export function distributionToCred(
   ds: TimelineDistributions,
@@ -49,15 +56,14 @@ export function distributionToCred(
     }
     cred[i] = new Array(intervals.length);
   }
-  if (scoringNodeIndices.length === 0) {
-    throw new Error("no nodes matched scoringNodePrefix");
-  }
 
   return ds.map(({interval, distribution, intervalWeight}) => {
     const intervalTotalScore = sum(
       scoringNodeIndices.map((x) => distribution[x])
     );
-    const intervalNormalizer = intervalWeight / intervalTotalScore;
+
+    const intervalNormalizer =
+      intervalTotalScore == 0 ? 0 : intervalWeight / intervalTotalScore;
     const cred = distribution.map((x) => x * intervalNormalizer);
     return {interval, cred};
   });
--- a/src/analysis/timeline/distributionToCred.test.js
+++ b/src/analysis/timeline/distributionToCred.test.js
@@ -87,7 +87,7 @@ describe("src/analysis/timeline/distributionToCred", () => {
       ];
       expect(expected).toEqual(actual);
     });
-    it("errors when no nodes are scoring", () => {
+    it("handles the case where no nodes are scoring", () => {
       const ds = [
         {
           interval: {startTimeMs: 0, endTimeMs: 10},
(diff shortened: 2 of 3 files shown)


BrianLitwin:
@decentralion this is updated. Your suggestions make sense to me, and they simplify the logic of the function.  

Going to follow this up with an Issue about filtering intervals where scoring nodes have a total cred scroe of 0.

src/analysis/timeline/distributionToCred.js
```diff
@@ -49,14 +56,17 @@ export function distributionToCred(
...
-    throw new Error("no nodes matched scoringNodePrefix");
-  }
 
   return ds.map(({interval, distribution, intervalWeight}) => {
     const intervalTotalScore = sum(
       scoringNodeIndices.map((x) => distribution[x])
     );
+
+    if (intervalTotalScore === 0) {
+      const cred = distribution.map((x) => x * 0);
```
decentralion: `((x) => x * 0)` is the same as `(() => 0)` for finite values

however, rather than special casing the return, I think it would be cleaner to say:

`const intervalNormalizer = intervalTotalScore == 0 ? 0 : intervalWeight / intervalTotalScore`

BrianLitwin: Updated

src/analysis/timeline/distributionToCred.js
```diff
@@ -49,16 +56,16 @@ export function distributionToCred(
...
-  if (scoringNodeIndices.length === 0) {
-    throw new Error("no nodes matched scoringNodePrefix");
-  }
 
   return ds.map(({interval, distribution, intervalWeight}) => {
     const intervalTotalScore = sum(
       scoringNodeIndices.map((x) => distribution[x])
     );
+
     const intervalNormalizer = intervalWeight / intervalTotalScore;
```
decentralion: Rather than checking if the intervalTotalScore is 0 for every single cred entry, we should do it once when we compute the normalizer.

`const intervalNormalizer = intervalTotalScore == 0 ? 0 : intervalWeight/intervalTotalScore`

BrianLitwin: re-updated 😬

src/analysis/timeline/distributionToCred.js
```diff
@@ -49,15 +56,14 @@ export function distributionToCred(
...
-  }
 
   return ds.map(({interval, distribution, intervalWeight}) => {
     const intervalTotalScore = sum(
       scoringNodeIndices.map((x) => distribution[x])
     );
-    const intervalNormalizer = intervalWeight / intervalTotalScore;
+
+    const intervalNormalizer =
+      intervalTotalScore == 0 ? 0 : intervalWeight / intervalTotalScore;
```
decentralion: Now that I'm reading this on review, I see that the code I suggested to you doesn't quite match our best practices. In general we use the type-coercing equality operator (`==`) only when comparing to `null`, since in that case we really do want to take advantage of the fact that `null == undefined`. In many other cases, the type coercion just creates opportunities for new unexpected behavior, e.g. the fact that `[] == 0` is likely to be surprising.

However, I won't block this review on the nit; this is the kind of thing that we should have a [lint rule](https://eslint.org/docs/rules/eqeqeq) for. I'll follow up with a lint rule.

decentralion: cf https://github.com/sourcecred/sourcecred/pull/1579

PR #1562, opened 2020-01-16
mirror: add network request logging

### LINKED ISSUE #1450: Keep a network log in the GraphQL mirror
When debugging issues relating to loading data from GitHub, it is often
helpful to insert logging statements to print every outgoing request and
incoming response. This is of course extremely noisy in general, which
is why we don’t have it on by default. But we _could_ modify the Mirror
module to save each outgoing request and incoming response to a new
table in its database. Users encountering trouble could then just send
us their database, and we’d be able to inspect the logs to look for
trouble. They wouldn’t have to build a special debugging build of
SourceCred, and even transient, hard-to-reproduce errors would have
logged useful information as soon as the problem is hit once.

We’d want to save the following properties:

  - Request data:
      - The request timestamp.
      - The stringified GraphQL query.
      - The JSON-stringified query parameters (even though they’re
        always `{}` for now).
  - Response data (once the response is received):
      - The response timestamp.
      - The raw response from the server, which should be a JSON object.
      - Whether the promise resolved or rejected.
  - Processing data (once ingestion completes):
      - The associated update ID: a primary key into the `updates` table
        that can be used to link a mutation to a database object with
        the request/response that generated it.

Note that authentication is handled by the provider of the `postQuery`
callback; the Mirror module knows nothing about it, so we don’t need to
strip out any tokens.

This would enable us to answer simple questions like “What was the query
that caused this ‘nonexistent node: undefined’ error?” as well as deeper
after-the-fact questions like “What query most recently updated the
`reactions` connection of this issue?” by cross-referencing against the
Mirror module’s existing update IDs.

(Context: I originally floated an idea like this in #998.)

BrianLitwin:
Logs network requests and responses into a table in the db. Also logs
an `UpdateId` that links the mutation to the network request/response
that generated it.

This is a breaking change to the Mirror database format; existing caches
will need to be regenerated.

Resolves #1450.

Test Plan:
Unit tests added, covering the core functionality and integration into
the public APIs. After loading a project and manually inspecting it in
an sqlite shell, the results in the `network_log` table looked
reasonable.

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -322,7 +322,7 @@ export class Mirror {
           CREATE TABLE network_log (
               rowid INTEGER PRIMARY KEY,
               query TEXT,
-              query_parameters TEXT, -- stringified JSON object
+              query_parameters TEXT,  -- stringified JSON object
               request_time_epoch_millis INTEGER,
               response TEXT,
               response_time_epoch_millis INTEGER,
@@ -757,9 +757,11 @@ export class Mirror {
     return this._db
       .prepare(
         dedent`\
-        INSERT INTO network_log (query, query_parameters, request_time_epoch_millis)
-        VALUES (:query, :queryParameters, :requestTimestamp)
-      `
+          INSERT INTO network_log (
+              query, query_parameters, request_time_epoch_millis
+          )
+          VALUES (:query, :queryParameters, :requestTimestamp)
+        `
       )
       .run({
         query,
@@ -781,11 +783,11 @@ export class Mirror {
 
     const stmt = this._db.prepare(
       dedent`\
-          UPDATE network_log
-          SET
-              response = :response,
-              response_time_epoch_millis = :responseTimestamp
-          WHERE rowid = :rowid
+        UPDATE network_log
+        SET
+            response = :response,
+            response_time_epoch_millis = :responseTimestamp
+        WHERE rowid = :rowid
       `
     );
     const saveResponse = _makeSingleUpdateFunction(stmt);
@@ -799,11 +801,11 @@ export class Mirror {
   _logRequestUpdateId(rowid: NetworkLogId, updateId: UpdateId): void {
     const stmt = this._db.prepare(
       dedent`\
-          UPDATE network_log
-          SET
-              update_id = :updateId
-          WHERE rowid = :rowid
-        `
+        UPDATE network_log
+        SET
+            update_id = :updateId
+        WHERE rowid = :rowid
+      `
     );
     const saveUpdateId = _makeSingleUpdateFunction(stmt);
     saveUpdateId({rowid, updateId});
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -188,7 +188,7 @@ export class Mirror {
     // it requires bumping the version, bump it: requiring some extra
     // one-time cache resets is okay; doing the wrong thing is not.
     const blob = stringify({
-      version: "MIRROR_v7",
+      version: "MIRROR_v8",
       schema: this._schema,
       options: {
         blacklistedIds: this._blacklistedIds,
@@ -317,6 +317,21 @@ export class Mirror {
         dedent`\
           CREATE INDEX idx_connection_entries__connection_id
           ON connection_entries (connection_id)
+        `,
+        dedent`\
+          CREATE TABLE network_log (
+              rowid INTEGER PRIMARY KEY,
+              query TEXT,
+              query_parameters TEXT, -- stringified JSON object
+              request_time_epoch_millis INTEGER,
+              response TEXT,
+              response_time_epoch_millis INTEGER,
+              update_id INTEGER,
+              CHECK((query IS NULL) = (query_parameters IS NULL)),
+              CHECK((request_time_epoch_millis IS NULL) = (query IS NULL)),
+              CHECK((response_time_epoch_millis IS NULL) = (response IS NULL)),
+              FOREIGN KEY(update_id) REFERENCES updates(rowid)
+          )
         `,
       ];
       for (const sql of tables) {
@@ -729,6 +744,72 @@ export class Mirror {
   }
 
   /**
+   * Save the request query and query parameters to the database.
+   */
+  _logRequest(
+    body: Queries.Body,
+    variables: {+[string]: any},
+    timestamp: Date
+  ): NetworkLogId {
+    const query = Queries.stringify.body(body, Queries.inlineLayout());
+    const queryParameters = stringify(variables);
+
+    return this._db
+      .prepare(
+        dedent`\
+        INSERT INTO network_log (query, query_parameters, request_time_epoch_millis)
+        VALUES (:query, :queryParameters, :requestTimestamp)
+      `
+      )
+      .run({
+        query,
+        queryParameters,
+        requestTimestamp: +timestamp,
+      }).lastInsertRowid;
+  }
+
+  /**
+   * Save the network response and response timestamp to the table row
+   * corresponding to the request that generated it.
+   */
+  _logResponse(
+    rowid: NetworkLogId,
+    jsonResponse: UpdateResult,
+    timestamp: Date
+  ): void {
+    const response = stringify(jsonResponse);
+
+    const stmt = this._db.prepare(
+      dedent`\
+          UPDATE network_log
+          SET
+              response = :response,
+              response_time_epoch_millis = :responseTimestamp
+          WHERE rowid = :rowid
+      `
+    );
+    const saveResponse = _makeSingleUpdateFunction(stmt);
+    saveResponse({response, responseTimestamp: +timestamp, rowid});
+  }
+
+  /**
+   * Save the UpdateId in the table row corresponding to the network request
+   * that generated it.
+   */
+  _logRequestUpdateId(rowid: NetworkLogId, updateId: UpdateId): void {
+    const stmt = this._db.prepare(
+      dedent`\
+          UPDATE network_log
+          SET
+              update_id = :updateId
+          WHERE rowid = :rowid
+        `
+    );
+    const saveUpdateId = _makeSingleUpdateFunction(stmt);
+    saveUpdateId({rowid, updateId});
+  }
+
+  /**
    * Perform one step of the update loop: find outdated entities, fetch
    * their updates, and feed the results back into the database.
    *
@@ -759,9 +840,14 @@ export class Mirror {
       connectionLimit: options.connectionLimit,
     });
     const body = [Queries.build.query("MirrorUpdate", [], querySelections)];
-    const result: UpdateResult = await postQuery({body, variables: {}});
+    const variables = {};
+    const requestRowId = this._logRequest(body, variables, options.now());
+    const result: UpdateResult = await postQuery({body, variables});
+    this._logResponse(requestRowId, result, options.now());
+
     _inTransaction(this._db, () => {
       const updateId = this._createUpdate(options.now());
+      this._logRequestUpdateId(requestRowId, updateId);
       this._nontransactionallyUpdateData(updateId, result);
     });
     return Promise.resolve(true);
@@ -1947,6 +2033,7 @@ export function _buildSchemaInfo(schema: Schema.Schema): SchemaInfo {
 }
 
 type UpdateId = number;
+type NetworkLogId = number;
 
 /**
  * A set of objects and connections that should be updated.
--- a/src/graphql/mirror.test.js
+++ b/src/graphql/mirror.test.js
@@ -14,6 +14,7 @@ import {
   _makeSingleUpdateFunction,
   Mirror,
 } from "./mirror";
+import stringify from "json-stable-stringify";
 
 describe("graphql/mirror", () => {
   function issueTimelineItemClauses() {
@@ -148,6 +149,7 @@ describe("graphql/mirror", () => {
               "links",
               "connections",
               "connection_entries",
+              "network_log",
             ])
           ).sort()
         );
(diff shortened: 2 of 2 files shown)


wchargin:
Awesome! I’ll take a look.

src/graphql/mirror.test.js
```diff
@@ -3627,6 +3628,58 @@ describe("graphql/mirror", () => {
...
+        now: () => new Date(),
+      });
+
+      const results = await db
+        .prepare(
+          "SELECT query, query_parameters, update_id, response FROM network_log"
+        )
+        .get();
+
+      expect(results).toMatchSnapshot();
```
wchargin: Snapshot testing is a bit of a brute force implement. It only works for
literal equality checks, and the ease with which one can write snapshot
tests makes it tempting to make tests that are brittle. This in turn
gets folks in the habit of giving snapshots only cursory inspections and
even updating them blindly. When a snapshot test fails, it’s often hard
to tell whether the change is a legitimate error, a flake, or a legacy
value that needs to be updated. Instead of snapshot tests, a targeted
test that checks _properties_ of the relevant value is often clearer and
more useful.

Thus, we try to use snapshot tests only for things that are really
unwieldy and large, *and* that can’t reasonably be reduced to something
more targeted. Note that the module that you’re currently editing has
3500 lines of tests, and yet only two snapshot tests among them, and
those two snapshots are explicitly for integration tests that are meant
to be copy-pastable from the snapshot test file the GitHub GraphQL UI
(as documented in the comment on the test case).

This newly added snapshot test, for instance, would spuriously fail if
the system under test changed the query name `MirrorUpdate` to `MU`.
(e.g., as a refactor to reduce network bandwidth).

Please replace these snapshot tests with something more precise. It may
be helpful to follow the pattern used by the rest of the tests: we test
the helper methods carefully and precisely, and then simply check that
`update` invokes the helper methods correctly. If you follow this
pattern, then you can control exactly what gets passed to `_logRequest`
and friends. This will also give you the opportunity to test that the
query variables are serialized correctly even when non-empty. You can
update the existing test, “update > delegates and updates the database”
to add the checks that the helpers are being invoked correctly.

src/graphql/mirror.js
```diff
@@ -318,6 +318,18 @@ export class Mirror {
           CREATE INDEX idx_connection_entries__connection_id
           ON connection_entries (connection_id)
         `,
+        dedent`\
+          CREATE TABLE network_log (
```
wchargin: Table schema looks good! You are creating a new table, so you must bump
the Mirror schema version (currently `MIRROR_v7`).

src/graphql/mirror.js
```diff
@@ -318,6 +318,18 @@ export class Mirror {
...
+              rowid INTEGER PRIMARY KEY,
+              query TEXT NOT NULL,
+              query_parameters TEXT NOT NULL,
```
wchargin: The term `query_parameters` is apt, but in other contexts can refer to
URL query parameters, like `page?foo=bar&baz=quux`. For clarity, please
specify that this is a JSON string:

```javascript
              query_parameters TEXT NOT NULL,  -- stringified JSON object
```

src/graphql/mirror.js
```diff
@@ -318,6 +318,18 @@ export class Mirror {
...
+              request_timestamp INTEGER NOT NULL,
```
wchargin: Optional, but consider making these nullable for symmetry? There’s no
critical application logic that relies on them being required.

src/graphql/mirror.js
```diff
@@ -318,6 +318,18 @@ export class Mirror {
...
+              response TEXT,
+              response_timestamp INTEGER,
+              update_id INTEGER,
+              FOREIGN KEY(update_id) REFERENCES updates(rowid)
```
wchargin: OK: You could add `CHECK` constraints here that enforce that the
relevant groups of parameters are all null or not null together (e.g.,
`response` and `response_timestamp` should be either both absent or both
present), but it’d also be fine to leave them out.

(IMHO, the underlying ontology here is that there are three relations:
say, `request_log`, `response_log`, and `update_log`, linked by foreign
keys, and within each relation all columns are non-null. Using this
compact single-relation form is fine because it’s probably more
convenient for interactive use, which is a primary design constraint.)

src/graphql/mirror.js
```diff
@@ -759,9 +833,14 @@ export class Mirror {
       connectionLimit: options.connectionLimit,
     });
     const body = [Queries.build.query("MirrorUpdate", [], querySelections)];
-    const result: UpdateResult = await postQuery({body, variables: {}});
+    const variables = {};
+    const requestRowId = await this._log_request(body, variables);
+    const result: UpdateResult = await postQuery({body, variables: variables});
```
wchargin: nit: Consider field-punning `{body, variables}` for consistency.

src/graphql/mirror.js
```diff
@@ -728,6 +740,68 @@ export class Mirror {
...
+    const response = JSON.stringify(jsonResponse);
+    const response_timestamp = new Date();
+    this._db
+      .prepare(
+        dedent`\
+        UPDATE network_log
+        SET
+            response = ?,
+            response_timestamp = ?
+        WHERE rowid = ?
```
wchargin: If the provided row ID does not actually match any rows, this code as
written will silently succeed. The `_makeSingleUpdateFunction` helper
can handle the relevant checks and errors for you; you should use it.

src/graphql/mirror.js
wchargin: Consider calling this `request_time_epoch_millis` (and likewise for the
response), as in the `updates` table, to make the units evident?

src/graphql/mirror.js
```diff
@@ -728,6 +740,68 @@ export class Mirror {
     }
   }
 
+  /**
+   * Save the request query and query parameters to the database.
+   */
+
+  async _log_request(
```
wchargin: nit: No blank line between doc comment and function declaration.

src/graphql/mirror.js
```diff
@@ -728,6 +740,68 @@ export class Mirror {
...
+    body: Queries.Body,
+    variables: {+[string]: any}
+  ): Promise<number> {
```
wchargin: Consider defining an internal type alias `type NetworkLogId = number`
for readability, like the existing `type UpdateId`? It’s not obvious
from the type signature what number the `Promise<number>` represents.

src/graphql/mirror.js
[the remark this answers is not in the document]
BrianLitwin: updated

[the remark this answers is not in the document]
BrianLitwin: updated - the new linter helped!

src/graphql/mirror.js
```diff
@@ -728,6 +740,68 @@ export class Mirror {
...
+    const query = Queries.stringify.body(body, Queries.inlineLayout());
+    const query_parameters = JSON.stringify(variables);
```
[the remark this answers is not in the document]
BrianLitwin: updated

src/graphql/mirror.js
```diff
@@ -728,6 +740,68 @@ export class Mirror {
...
+    const request_timestamp = new Date();
```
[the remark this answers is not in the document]
BrianLitwin: updated

src/graphql/mirror.js
```diff
@@ -728,6 +740,68 @@ export class Mirror {
...
+    const response_timestamp = new Date();
+    this._db
+      .prepare(
+        dedent`\
+        UPDATE network_log
+        SET
+            response = ?,
+            response_timestamp = ?
+        WHERE rowid = ?
+      `
```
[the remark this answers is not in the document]
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: added the timestamp field

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.test.js
```diff
@@ -3627,6 +3628,58 @@ describe("graphql/mirror", () => {
...
+        })
+      );
+
+      await mirror.update(postQuery, {
+        nodesOfTypeLimit: 2,
+        nodesLimit: 3,
+        connectionLimit: 4,
+        connectionPageSize: 5,
+        since: new Date(123),
+        now: () => new Date(),
```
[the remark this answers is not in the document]
BrianLitwin: updated - thanks for explaining this!

src/graphql/mirror.test.js
BrianLitwin: Thanks again for a great explanation! Updated.

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
```diff
@@ -318,6 +318,18 @@ export class Mirror {
           CREATE INDEX idx_connection_entries__connection_id
           ON connection_entries (connection_id)
         `,
+        dedent`\
+          CREATE TABLE network_log (
+              rowid INTEGER PRIMARY KEY,
+              query TEXT NOT NULL,
+              query_parameters TEXT NOT NULL, -- stringified JSON object
+              request_timestamp INTEGER,
```
Beanow: I think `request_timestamp` should be `NOT NULL`?

src/graphql/mirror.test.js
```diff
@@ -987,13 +989,30 @@ describe("graphql/mirror", () => {
...
         const postQuery = jest.fn();
         const now = jest
           .fn()
           .mockImplementationOnce(() => new Date(456))
-          .mockImplementationOnce(() => new Date(789));
+          .mockImplementationOnce(() => new Date(457))
+          .mockImplementationOnce(() => new Date(458))
+          .mockImplementationOnce(() => new Date(789))
+          .mockImplementationOnce(() => new Date(790))
+          .mockImplementationOnce(() => new Date(791));
```
Beanow: nit: there's a fair bit of coupling in this testing method
From a reviewers' pov, this amount of dates is difficult to keep in mind.

_Oh the joys of testing time! :sweat_smile:_

This seems to tread into "[solving puzzles, not problems](https://www.youtube.com/watch?v=2V1FtfBDsLU&t=4419)" territory. So I'm leaving this as a nit.

That said. The issue I see with coupling the test with implementation details this tightly, is that it becomes harder to change an implementation detail and know the program is still doing what it's supposed to. (i.e. the tests pass)

The easiest way to work around this I think, is to pass the `now: () => Date` function as an argument to the log functions, instead of the `Date` instances. This way, in a higher-level test, you don't need to call the function (only verify that it was passed).

src/graphql/mirror.test.js
```diff
@@ -3627,6 +3689,79 @@ describe("graphql/mirror", () => {
...
+        const results = await db
+          .prepare(
+            "SELECT query, query_parameters, request_timestamp FROM network_log"
+          )
+          .get();
+        const expected = Queries.stringify.body(query, Queries.inlineLayout());
+        expect(results.query).toEqual(expected);
+        expect(results.query_parameters).toEqual(stringify({foo: "bar"}));
+        expect(results.request_timestamp).toEqual(
+          new Date(123).getMilliseconds()
```
Beanow: nit: I think it's ok to just use `123` here without going through a `Date` object.

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
Beanow: Just noticed https://github.com/sourcecred/sourcecred/pull/1562#discussion_r367784945 this was already discussed. Go ahead and ignore. :smile:

src/graphql/mirror.test.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: Updated. I had trouble testing this however, I believe b/c of the `+` attached to the `timestamp` parameter passed to `run`. Passing `null` to `_logRequest` for the `timestamp` parameter didn't trigger a constraint violation, but passing `null` directly to the `run` function in place of `+timestamp` did.

src/graphql/mirror.test.js
BrianLitwin: Do you think we need to update `_createUpdate` as well?

Beanow: I would go with the smallest PR. So leaving that as-is.
You could create a follow up for that if you like though.
Though do check with @decentralion / @wchargin, since they have more context on this.

BrianLitwin: I'm going to leave this as-is unless we update the `_createUpdate` method as well.  I agree that the code you highlighted is onerous, but don't love the idea of using `options.now` on one line and `options.now()` a few lines later; the rationale of keeping the test code easier to read is kinda non-obvious if you were reading the module.

BrianLitwin:
@wchargin @Beanow This is RFAL; Thanks for the first reviews!

decentralion:
@BrianLitwin I recommend using the "re-request review" button which you'll find next to @wchargin and @Beanow's names in the review box at the top of the PR.

src/graphql/mirror.js
```diff
@@ -1947,6 +2033,7 @@ export function _buildSchemaInfo(schema: Schema.Schema): SchemaInfo {
 }
 
 type UpdateId = number;
+type NetworkLogid = number;
```
wchargin: Can we call this `NetworkLogId` for consistency with `UpdateId`,
`ObjectId`, `RepoId`, `TaskId`, etc. elsewhere in SourceCred? (It’s true
that this differs from SQLite’s `rowid`, but the in-project convention
is stronger than the cross-project convention, and also SQLite’s choice
is motivated by SQL’s case-insensitivity, which doesn’t apply to us.)

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.js
BrianLitwin: updated

src/graphql/mirror.test.js
```diff
@@ -1096,8 +1115,52 @@ describe("graphql/mirror", () => {
...
+        ]);
+
+        //We should have called logRequestUpdateId with updateId
+        expect(spyLogRequestUpdateId).toHaveBeenCalledTimes(2);
+        expect(spyLogRequestUpdateId.mock.calls[0]).toEqual([
+          spyLogRequest.mock.results[0].value,
+          spyCreateUpdate.mock.results[0].value,
+        ]);
+
+        expect(spyLogRequestUpdateId).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
BrianLitwin: updated

src/graphql/mirror.test.js
```diff
@@ -1096,8 +1115,52 @@ describe("graphql/mirror", () => {
...
+          new Date(457),
+        ]);
+
+        expect(spyLogResponse.mock.calls[1]).toEqual([
+          spyLogRequest.mock.results[1].value,
+          await postQuery.mock.results[1].value,
+          new Date(790),
+        ]);
+
+        //We should have called logRequestUpdateId with updateId
```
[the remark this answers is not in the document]
BrianLitwin: updated

wchargin:
Thanks! Merged master for you, and adjusted the SQL spacing a wee bit
just per local conventions (no big deal). Will merge once CI passes.

decentralion:
Nice work @BrianLitwin!

PR #1563, opened 2020-01-17
lint: require camel case variables

wchargin:
Summary:
We’ve had this policy unspoken since the beginning; making it explicit
as a lint rule makes it one less thing to think about. The few existing
offenders can almost all be changed with no value lost; one is an extern
that needs a lint suppression.

Test Plan:
That Flow and tests still pass suffices.

wchargin-branch: lint-camelcase


BrianLitwin:
I'm a fan!

PR #1564, opened 2020-01-19
Backend: create DataDirectory class

Beanow:
Currently the responsibility for the SourceCred directory
is spread out in different places. Some in core/project_io
some in api/load, some in the plugins.

This class is intended to centralize that IO using simple
interfaces we can depend on (and mock) instead.

### DIFF
--- a/src/backend/dataDirectory.js
+++ b/src/backend/dataDirectory.js
@@ -1,0 +1,50 @@
+// @flow
+
+import path from "path";
+import fs from "fs-extra";
+import Database from "better-sqlite3";
+import stringify from "json-stable-stringify";
+import {type Project, projectToJSON} from "../core/project";
+import {directoryForProjectId} from "../core/project_io";
+import {type CacheProvider} from "./cache";
+import type {
+  ProjectStorageProvider,
+  ProjectStorageExtras,
+} from "./projectStorage";
+
+/**
+ * Represents a SourceCred data directory.
+ */
+export class DataDirectory implements CacheProvider, ProjectStorageProvider {
+  +_sourcecredDirectory: string;
+  +_cacheDirectory: string;
+
+  constructor(sourcecredDirectory: string) {
+    this._sourcecredDirectory = sourcecredDirectory;
+    this._cacheDirectory = path.join(sourcecredDirectory, "cache");
+  }
+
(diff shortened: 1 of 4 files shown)


src/backend/dataDirectory.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Represents a SourceCred data directory.
+ */
+export class DataDirectory implements CacheProvider, ProjectStorageProvider {
+  sourcecredDirectory: string;
+  cacheDirectory: string;
+
+  constructor(sourcecredDirectory: string) {
+    this.sourcecredDirectory = sourcecredDirectory;
+    this.cacheDirectory = path.join(sourcecredDirectory, "cache");
```
decentralion: The `cacheDirectory` is a public property, however there doesn't seem to be any way for a client to know if the cacheDirectory has actually been created. So, to be safe, every client needs to do `await fs.mkdirp(dataDirectory.cacheDirectory)`. Seems messy. Can we give the `DataDirectory` and `initialize()` method, or just have it ensure directories exist in the constructor?

src/backend/project.js
```diff
@@ -0,0 +1,14 @@
+//@flow
+
+import {Graph} from "../core/graph";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+import {type Project} from "../core/project";
+
+export type ProjectStorageExtras = {
```
decentralion: Why is this file called `project.js` and not `projectStorage.js`?

src/backend/dataDirectory.js
Beanow: It should probably be private. Since you're not supposed to do really any IO there when this class is responsible for that :]

src/backend/project.js
Beanow: Had "project backend" vs "core project" in mind. You're right though, not having anything else in it it would make more sense as `projectStorage.js`.

Beanow: Renamed

src/backend/dataDirectory.js
Beanow: Made both private and immutable.
(Wasn't aware underscore properties were an actually supported Flow convention)

Beanow:
Tests added

Issue #1559, opened 2020-01-13
Reduce load complexity


Beanow:
In #1506 we settled on using static typing and hard-coded loading order of plugins.
To do that I think it's necessary to reduce complexity of [`/src/api/load.js`](https://github.com/sourcecred/sourcecred/blob/063c08d97435400dd9715b79b6afa3875ec3c3c0/src/api/load.js#L43-L46).

Currently it's hard to change the loading code for new features like UniRef. And the tests are brittle, relying on very specific module mocking. Meaning the tests can fail by changing the API of an external module.

Future changes (such as refactoring into modules, and the instance system) should benefit from simple functions as well. So I'm planning to pick this up as a first step to landing the UniRef feature.

Beanow:
So you might have noticed by the number of PRs, this is quite challenging to actually satisfy. This approach is what I came up with after several complete refactors of load.

The issue that _kept coming back_ was the conflating of arguments that were environment related, with the more pure arguments. As well as constantly needing to rely on module mocking when it came to the plugins.

It's worth comparing the previous boilerplate just to mock.
https://github.com/sourcecred/sourcecred/blob/abf06c075f683293235d1c75affedc4a3a9054ed/src/api/load.test.js#L29-L87

With the tests found in the finished result.
https://github.com/sourcecred/sourcecred/tree/ab843da3b6f3f72df58ea2e4debcf030e79412ed/src/backend

Such as: https://github.com/sourcecred/sourcecred/blob/ab843da3b6f3f72df58ea2e4debcf030e79412ed/src/backend/pluginPlan.test.js#L15-L31

Beanow:
In terms of how this makes UniRef easier to implement and how this preserves static hard-coded loading discussed in #1506 

The `pluginGraphPlan` is where we will start depending on `ReferenceDetector`. Each plugin load interface would expose one and we can compose a unified detector, before creating graphs.

https://github.com/sourcecred/sourcecred/blob/ab843da3b6f3f72df58ea2e4debcf030e79412ed/src/backend/pluginPlan.js#L45-L72

Beanow:
As designed and referenced in #1586, this is now resolved.

Though the previous two comments referred to the LoadPlan concept. This was rejected and replaced with LoadContext + PluginLoaders (see #1586) instead, having this issue in mind.

PR #1565, opened 2020-01-19
Backend: introduce LoadPlan

Beanow:
Part of #1559 suggested refactor.

LoadPlan is the key concept which allows us to simplify APIs,
the smaller argument lists make change (and testing) easier.
Additionally since it's DI based, should remove the need for
mocking external modules.

### DIFF
--- a/src/backend/loadPlan.js
+++ b/src/backend/loadPlan.js
@@ -1,0 +1,50 @@
+//@flow
+
+import {Graph} from "../core/graph";
+import {type Project} from "../core/project";
+import {type TimelineCredParameters} from "../analysis/timeline/params";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+
+/**
+ * This file introduces a "plan" concept.
+ *
+ * A plan allows us to separate the large amount of dependency injection that's
+ * necessary to load a SourceCred instance. It makes a hard distinction between
+ * which arguments are environment-bound (like the TaskReporter, CacheProvider,
+ * and GithubToken) and which are parameters for the load (like Project).
+ *
+ * A <Something>Plan type only includes the later *parameters for loading*,
+ * because this Plan is typically the output of a factory function which curried
+ * the environment-bound arguments for us.
+ */
+
+export type MirrorPlan = (project: Project) => Promise<void>;
+
+export type GraphPlan = (project: Project) => Promise<Graph>;
+
+export type CredComputationPlan = (
+  graph: Graph,
+  params: ?$Shape<TimelineCredParameters>
+) => Promise<TimelineCred>;
+
+export type LoadPlan = {|
+  +mirror: MirrorPlan,
+  +createGraph: GraphPlan,
+  +computeCred: CredComputationPlan,
+|};
+
+export type LoadResult = {|
+  +graph: Graph,
+  +cred: TimelineCred,
+|};
+
+export async function executePlan(
+  plan: LoadPlan,
+  project: Project,
+  params: ?$Shape<TimelineCredParameters>
+): Promise<LoadResult> {
+  await plan.mirror(project);
+  const graph = await plan.createGraph(project);
+  const cred = await plan.computeCred(graph, params);
+  return {graph, cred};
+}
--- a/src/backend/cache.js
+++ b/src/backend/cache.js
@@ -1,0 +1,17 @@
+//@flow
+
+import Database from "better-sqlite3";
+
+/**
+ * A cache abstraction which plugins can use to store mirroring data in.
+ */
+export interface CacheProvider {
+  /**
+   * Opens a new Database handle.
+   * Given the same `id`, an existing Database from the cache *may* be provided.
+   *
+   * Make sure the `id` you're using is both _deterministic_ and doesn't
+   * conflict with other plugins.
+   */
+  database(id: string): Promise<Database>;
+}
--- a/src/backend/dataDirectory.js
+++ b/src/backend/dataDirectory.js
@@ -1,0 +1,47 @@
+// @flow
+
+import path from "path";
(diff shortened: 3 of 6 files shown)


src/backend/loadPlan.js
```diff
@@ -0,0 +1,50 @@
+//@flow
+
+import {Graph} from "../core/graph";
+import {type Project} from "../core/project";
+import {type TimelineCredParameters} from "../analysis/timeline/params";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+
+/**
+ * This file introduces a "plan" concept.
```
decentralion: IMO, this docstring provides the "why" (do we have a plan) but not "what" (is a plan).

I suggest something like:

---

This file introduces `Plans`. A `Plan` represents an execution context for SourceCred code that needs access to environmental variables, hidden credentials, disk, and IO with the outside world. Functionally, a `Plan` allows us to separate providing access to these environmental resources from providing parameters to control SourceCred's specific behavior.

Each `Plan` is a function which takes parameters and returns a Promise. The plan constructor is responsible for wrapping in any environmental variables; Plan consumers need only provide specific parameters and arguments.

Having Plans allows us to cleanly test SourceCred loading behavior without needing extensive dependency injection along the boundary with external data and IO...

---

Also, apropos of this discussion: why the name `Plan`? I don't find it super descriptive and wonder if we could find something more specific. Like `EnvironmentExecutor` or something? (I admit that's not great either.)

wchargin: +1 to addressing “what” first and foremost. Certainly include “why” if
that’s not obvious from the “what”.

+1 that *plan* is a strange name here. When I think of a plan, I think
of a piece of data describing how something should happen, like a query
plan (SQL) or a test plan. I don’t think of an action that executes
asynchronously and then returns nothing or something after having some
side effects.

At a glance, it looks like you’re basically trying to define monad
transformers for various `ReaderT`? …but then it’s not clear what
`MirrorPlan` is or how it fits in.

src/backend/loadPlan.js
```diff
@@ -0,0 +1,50 @@
...
+ * necessary to load a SourceCred instance. It makes a hard distinction between
+ * which arguments are environment-bound (like the TaskReporter, CacheProvider,
+ * and GithubToken) and which are parameters for the load (like Project).
+ *
+ * A <Something>Plan type only includes the later *parameters for loading*,
+ * because this Plan is typically the output of a factory function which curried
+ * the environment-bound arguments for us.
+ */
+
+export type MirrorPlan = (project: Project) => Promise<void>;
```
wchargin: It looks like this is supposed to silently have some side effect that is
depended on implicitly by some set of later processes. This kind of
spooky action at a distance tends to make systems harder to understand.

Instead, why doesn’t the `MirrorPlan` evaluate to a `Promise<Mirror>`?
Encode the information into the data; don’t suffer from [Boolean
blindness][bb].

[bb]: https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/

decentralion: +1, I would find this a lot cleaner / easier to reason about.

src/backend/loadPlan.js
Beanow: LoadPlan would be a struct of closures, or partially applied functions.

Perhaps the components as `MirrorClosure`, `GraphClosure`, `CredComputationClosure` with `LoadPlan` make sense?

src/backend/loadPlan.js
Beanow: The main goal here is:
Both the concrete environment that was supplied *and* how the environment should be effected are abstracted at this level.

You're right, that makes the "Mirroring" operation rather abstract.
Because it *is* intended to side-effect, as the `Promise<void>` would typically communicate. But we're intentionally not aware of the environment, nor how it should be affected. The `Promise<void>` will simply tell us "it's done".

However, that's not the case for _concrete_ implementations. A concrete implementation also reveals what the environment is (the function wasn't partially applied yet), and how it should be affected.

When a higher level function constructs a `LoadPlan` instance. It references concrete implementations. So this function does is aware whether it would make sense to create dependencies through a shared environment here.

This is why `Promise<Mirror>` would be the environment leaking out, and a breach of this concept. The side-effects should remain in the environment. And a `LoadPlan` constructor is responsible for it's relationship.

wchargin: I have read your comment a few times but am having trouble following.
Specifically, I don’t understand the bits about “higher level”
functions, “affecting environments”, and “concrete implementations”.

Conceptually, I don’t see why the Mirror interface is “intended to
side-effect” any more than the others. Loading a mirror _may_ require
fetching from the network, or may fetch from cache, or may fetch from a
test database. It doesn’t matter. What matters are which dependencies we
need (here: a `Project`) and what outputs we provide (here: a `Mirror`).
The whole point of dependency injection is establish the coupling at
this semantic layer rather than the implementation layer.

If other parameters are needed—for instance, a GitHub token—shouldn’t
they also be in the dependency set rather than being hard-coded to
derive from environment variables?

src/backend/loadPlan.js
wchargin: “Closure” tells me how it’s represented but not what it is. (It’s still
better than “plan”, because “plan” purports to tell me what it is but
actually conveys a misleading message.) These look like _providers_ or
_producers_, which are also terms familiar from dependency injection
frameworks.

The distinction in form between `LoadPlan` and the other “plans” is also
confusing on a second read. Why is `LoadPlan` a product while all the
others are functions?

It is hard for me to see the bigger picture here, partly because these
PRs are split so finely that I cannot see how the pieces fit together.

Beanow:
Closing in favor of #1586 LoadContext style.

PR #1566, opened 2020-01-19
Discourse: implement DiscourseLoader interface

Beanow:
This separates updating the mirror from creating the graph.
Necessary for LoadPlan as well as later UniRef support.

We're duplicating the DiscourseServer type, as the other
loadDiscourse API will be removed later.


Beanow:
Closing this PR for now.

PR #1567, opened 2020-01-19
Identity: implement IdentityLoader interface

Beanow:
We'll use this for LoadPlan.


decentralion:
Why does this have no test plan? It seems like having a unit test or two for this method would be nice.

Beanow:
Closing this PR for now.

PR #1568, opened 2020-01-19
Github: add CacheProvider support for fetchGithubRepo

Beanow:
Depends on #1564

Here's one I'd like some feedback on.

I've avoided making a breaking change to the function signature here. But that leaves the signature in a clearly weird state. Would you rather have `fetchGithubRepo` switch over to using `CacheProvider` only and update all usage?

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -18,6 +18,18 @@ import {BLACKLISTED_IDS} from "./blacklistedObjectIds";
 import type {Repository} from "./graphqlTypes";
 import schema from "./schema";
 import {validateToken} from "./token";
+import {type CacheProvider} from "../../backend/cache";
+
+function cacheID(resolvedId: Schema.ObjectId): string {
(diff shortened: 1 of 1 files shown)


src/plugins/github/fetchGithubRepo.js
```diff
@@ -34,7 +46,7 @@ import {validateToken} from "./token";
  */
 export default async function fetchGithubRepo(
   repoId: RepoId,
-  options: {|+token: string, +cacheDirectory: string|}
+  options: {|+token: string, +cacheDirectory: string, +cache?: CacheProvider|}
```
decentralion: This signature where we provide a (required) cacheDirectory, an (optional) cacheProvider, and then ignore the required argument if the optional one is provided, is really weird.

If you want to make a minimal change to support both use cases, you could set up a union type, right?

```js
type CacheInfo = {|type: "DIRECTORY", directory: string|} 
  | {|type: "PROVIDER", provider: CacheProvider|}
```

Beanow:
Closing in favor of #1614

PR #1569, opened 2020-01-19
Github: add fetchGithubRepoFromCache

Beanow:
Depends on #1614

Updating the mirror and using the mirror data should be separated. As unified reference detection requires the mirror of all plugins are updated. Upcoming refactors to the load system will solidify this.

While this refactor is ongoing, we will ignore the fetchGithubRepo return value, using it as a mirror update only. And use this new function to extract the data as needed in other loading steps.

Test plan: `yarn test --full`
As the existing `fetchGithubRepo` didn't have unit testing (and would require a lot of module mocking), instead I'm using the new function in `loadGraph`, which will be invoked by a full test.

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -20,6 +20,57 @@ import {validateToken} from "./token";
 import {cacheIdForRepoId} from "./cacheId";
 import {type CacheProvider} from "../../backend/cache";
 
+type FetchRepoOptions = {|
+  +token: string,
+  +cache: CacheProvider,
+|};
+
+/**
+ * Retrieve previously scraped data for a GitHub repo from cache.
+ *
+ * Note: the GithubToken requirement is planned to be removed.
+ * See https://github.com/sourcecred/sourcecred/issues/1580
+ *
+ * @param {RepoId} repoId
+ *    the GitHub repository to retrieve from cache
+ * @param {GithubToken} token
+ *    authentication token to be used for the GitHub API; generate a
+ *    token at: https://github.com/settings/tokens
+ * @return {Promise<Repository>}
+ *    a promise that resolves to a JSON object containing the data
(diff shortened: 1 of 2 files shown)


Beanow:
> Why do
> you need the `registerObject`?

Just not clear on the Mirror API. :smile:
Will remove that.

Also including #1580 changes.

Beanow:
Updated on multiple points.
- Rebased against master.
- Now depends on #1614 (as we'll use a common CacheProvider argument).
- No longer doing the `mirror.registerObject`.
- Update docstring and commit message.
- Include usage so we have test coverage.

Beanow:
Rebased as dependency got merged.

src/plugins/github/fetchGithubRepo.js
```diff
@@ -20,6 +20,57 @@ import {validateToken} from "./token";
 import {cacheIdForRepoId} from "./cacheId";
 import {type CacheProvider} from "../../backend/cache";
 
+type FetchRepoOptions = {|
+  +token: string,
```
decentralion: The token should be a validated GitHub token, not a string. It's the cleaner / more correct type signature.

I recognize this becomes stricter in that we'll now error if GitHub creates tokens that don't match our validator... but if that happens I'd rather just update the validation logic.

Beanow: Seems better as a separate PR to me. I'm keeping compatibility with the `fetchGithubRepo` arguments, which had an explicit reason to warn instead of error. Changing the API would break in a few places so this PR would complect.

My invocations of this function after the load refactor would be from `GithubTokens` already, casting back to strings.

decentralion: OK

Beanow: Tracking this issue here: https://github.com/sourcecred/sourcecred/issues/1626

PR #1570, opened 2020-01-19
Github: implement GithubLoader interface

Beanow:
This separates updating the mirror from creating the graph.
Necessary for LoadPlan as well as later UniRef support.

### DIFF
--- a/src/plugins/github/loader.js
+++ b/src/plugins/github/loader.js
@@ -1,0 +1,64 @@
+// @flow
+
+import {TaskReporter} from "../../util/taskReporter";
+import {fetchGithubRepoFromCache} from "./fetchGithubRepo";
+import fetchGithubRepo from "./fetchGithubRepo";
+import {RelationalView} from "./relationalView";
+import {createGraph as _createGraph} from "./createGraph";
+import {Graph} from "../../core/graph";
+import {type RepoId, repoIdToString} from "./repoId";
+import {type GithubToken} from "./token";
+import {type CacheProvider} from "../../backend/cache";
+
+export interface GithubLoader {
+  updateMirror(
+    repoIds: $ReadOnlyArray<RepoId>,
+    token: GithubToken,
+    cache: CacheProvider,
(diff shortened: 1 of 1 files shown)


src/plugins/github/loader.js
```diff
@@ -0,0 +1,64 @@
...
+  token: GithubToken,
+  cache: CacheProvider,
+  reporter: TaskReporter
+): Promise<void> {
+  for (const repoId of repoIds) {
+    const taskId = `github/${repoIdToString(repoId)}`;
+    reporter.start(taskId);
+    await fetchGithubRepo(repoId, {
+      token: token,
+      cacheDirectory: "unused_legacy",
```
decentralion: Let's change the preceding PR and remove this legacy line.

src/plugins/github/loader.js
```diff
@@ -0,0 +1,64 @@
...
+
+export interface GithubLoader {
+  updateMirror(
+    repoIds: $ReadOnlyArray<RepoId>,
+    token: GithubToken,
+    cache: CacheProvider,
+    reporter: TaskReporter
+  ): Promise<void>;
+
+  createGraph(
```
decentralion: Why do we need a token for `createGraph`-- it should only be needed for updating the mirror, no?

Beanow: The end result of a load, is a data directory with caches updated, it's project, graph, and cred.json files written out and a bunch of task reporting. So graph building and cred computation are part of it.

This interface includes the entirety of what's required to achieve that.

decentralion: Yeah, and for that reason it makes sense that the `updateMirror` step needs a GitHub token. But why does the `createGraph` step (which should be able to directly retrieve data from the mirror) need the token?

Beanow: Since this came up a few times, the reason and proposed solution here:
https://github.com/sourcecred/sourcecred/issues/1580

Beanow:
Closing this PR for now.

PR #1571, opened 2020-01-19
Backend: implement pluginMirrorPlan

Beanow:
Provides the Plan abstraction to mirror Github and Discourse.

### DIFF
--- a/src/backend/pluginPlan.test.js
+++ b/src/backend/pluginPlan.test.js
@@ -1,0 +1,119 @@
+// @flow
+
+import {Graph} from "../core/graph";
+import {node as graphNode} from "../core/graphTestUtil";
+import {createProject} from "../core/project";
+import {TestTaskReporter} from "../util/taskReporter";
+import {pluginMirrorPlan} from "./pluginPlan";
+
+const githubSentinel = graphNode("github-sentinel");
+const githubGraph = () => new Graph().addNode(githubSentinel);
+const discourseSentinel = graphNode("discourse-sentinel");
+const discourseGraph = () => new Graph().addNode(discourseSentinel);
+const combinedGraph = () => Graph.merge([githubGraph(), discourseGraph()]);
+
+const mockCacheProvider = () => ({
+  database: jest.fn(),
+});
+
+const mockPluginLoaders = () => ({
+  github: {
+    updateMirror: jest.fn(),
+    createGraph: jest.fn().mockResolvedValue(githubGraph()),
+  },
+  discourse: {
+    updateMirror: jest.fn(),
+    createGraph: jest.fn().mockResolvedValue(discourseGraph()),
+  },
+  identity: {
+    contractGraph: jest.fn().mockResolvedValue(combinedGraph()),
+  },
+});
+
+describe("src/backend/pluginPlan", () => {
+  describe("pluginMirrorPlan", () => {
+    it("should update discourse mirror", async () => {
+      // Given
+      const loaders = mockPluginLoaders();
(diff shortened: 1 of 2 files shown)


src/backend/pluginPlan.test.js
```diff
@@ -0,0 +1,119 @@
...
+        id: "has-discourse",
+        discourseServer: {serverUrl: "http://foo.bar"},
+      });
+
+      // When
+      const mirror = pluginMirrorPlan(loaders, ghToken, cache, reporter);
+      await mirror(project);
+
+      // Then
+      const {discourse, github, identity} = loaders;
```
decentralion: This test is currently comprehensive in that it verifies none of the other loaders were called inappropriately. However, it strikes me as quite unlikely that the next person to add a loader will update this particular test case, at which point the test is verifying that some (old) subset of the loaders are not being called inappropriately.

I think we can make things a bit simpler by having the test only verify that the discourse mirror was updated correctly, without verifying that other methods weren't called.

src/backend/pluginPlan.test.js
```diff
@@ -0,0 +1,119 @@
...
+        cache,
+        reporter
+      );
+      expect(discourse.createGraph).toBeCalledTimes(0);
+      expect(github.updateMirror).toBeCalledTimes(0);
+      expect(github.createGraph).toBeCalledTimes(0);
+      expect(identity.contractGraph).toBeCalledTimes(0);
+    });
+
+    it("should fail when missing GithubToken", async () => {
```
decentralion: It should fail when missing GH token, *and* repoIds are present in the project. If there are no repoIds, then missing GH token should not fail. Please rename this test case, and add another one in the case with no GH token and no repoIds.

src/backend/pluginPlan.test.js
```diff
@@ -0,0 +1,119 @@
...
+      // Given
+      const loaders = mockPluginLoaders();
+      const cache = mockCacheProvider();
+      const ghToken = null;
+      const reporter = new TestTaskReporter();
+      const project = createProject({
+        id: "has-github",
+        repoIds: ([{owner: "sourcecred-test", name: "example-github"}]: any),
```
decentralion: Rather than any-casting, use `makeRepoId("sourcecred-test", "example-github")` or `stringToRepoId("sourcecred-test/example-github")`.

src/backend/pluginPlan.test.js
```diff
@@ -0,0 +1,119 @@
...
+      expect(github.updateMirror).toBeCalledTimes(0);
+      expect(github.createGraph).toBeCalledTimes(0);
+      expect(identity.contractGraph).toBeCalledTimes(0);
+    });
+
+    it("should update github mirror", async () => {
+      // Given
+      const loaders = mockPluginLoaders();
+      const cache = mockCacheProvider();
+      const ghToken = ("t000ken": any);
```
decentralion: Rather than any casting, please use `validateToken("0".repeat(40))`

src/backend/pluginPlan.test.js
```diff
@@ -0,0 +1,119 @@
...
+      const reporter = new TestTaskReporter();
+      const project = createProject({
+        id: "has-github",
+        repoIds: ([{owner: "sourcecred-test", name: "example-github"}]: any),
```
decentralion: As above, please construct an actual repoId without any casting. Since repoIds are immutable, you can define the example repoId once in the outermost describe block

Beanow:
Closing this PR for now.

PR #1572, opened 2020-01-19
Backend: implement pluginGraphPlan

Beanow:
Provides the Plan abstraction to support Github, Discourse and
Identities in an equivalent way to the current load.

### DIFF
--- a/src/backend/pluginPlan.js
+++ b/src/backend/pluginPlan.js
@@ -1,7 +1,8 @@
 //@flow
 
 import {type CacheProvider} from "./cache";
-import {type MirrorPlan} from "./loadPlan";
+import {type MirrorPlan, type GraphPlan} from "./loadPlan";
+import {Graph} from "../core/graph";
 import {type Project} from "../core/project";
 import {TaskReporter} from "../util/taskReporter";
 import {type GithubToken} from "../plugins/github/token";
@@ -38,5 +39,34 @@ export function pluginMirrorPlan(
     }
 
     await Promise.all(tasks);
+  };
+}
+
+export function pluginGraphPlan(
+  {github, discourse, identity}: PluginLoaders,
+  githubToken: ?GithubToken,
+  cache: CacheProvider
+): GraphPlan {
+  return async (project: Project): Promise<Graph> => {
+    const tasks: Promise<Graph>[] = [];
+
+    if (project.discourseServer) {
+      tasks.push(discourse.createGraph(project.discourseServer, cache));
+    }
+
+    if (project.repoIds.length) {
+      if (!githubToken) throw new Error("GithubToken not set");
+      tasks.push(github.createGraph(project.repoIds, githubToken, cache));
+    }
+
+    const pluginGraphs = await Promise.all(tasks);
+    let graph = Graph.merge(pluginGraphs);
+
+    if (project.identities.length) {
+      const {serverUrl} = project.discourseServer || {serverUrl: null};
+      graph = identity.contractGraph(graph, project.identities, serverUrl);
+    }
+
+    return graph;
   };
 }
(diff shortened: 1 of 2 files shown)


src/backend/pluginPlan.test.js
```diff
@@ -116,4 +116,87 @@ describe("src/backend/pluginPlan", () => {
...
+      // Then
+      const {discourse, github, identity} = loaders;
+      expect(graph).toEqual(discourseGraph());
+      expect(discourse.updateMirror).toBeCalledTimes(0);
+      expect(discourse.createGraph).toBeCalledTimes(1);
+      expect(discourse.createGraph).toBeCalledWith(
+        project.discourseServer,
+        cache
+      );
+      expect(github.updateMirror).toBeCalledTimes(0);
```
decentralion: As discussed here (https://github.com/sourcecred/sourcecred/pull/1571#discussion_r368328923), I think that verifying that unrelated methods were not called is not robust to future changes, so we may as well not try to verify it.

src/backend/pluginPlan.test.js
```diff
@@ -116,4 +116,87 @@ describe("src/backend/pluginPlan", () => {
...
+      expect(github.createGraph).toBeCalledTimes(0);
+      expect(identity.contractGraph).toBeCalledTimes(0);
+    });
+
+    it("fail when missing GithubToken", async () => {
```
decentralion: As discussed here (https://github.com/sourcecred/sourcecred/pull/1571#discussion_r368329017), the test description does not match the implementation or intended contract. I'd like it if we can silo GitHub token requirements to mirror acquisition, anyway.

src/backend/pluginPlan.test.js
```diff
@@ -116,4 +116,87 @@ describe("src/backend/pluginPlan", () => {
...
+      expect(github.updateMirror).toBeCalledTimes(0);
+      expect(github.createGraph).toBeCalledTimes(0);
+      expect(identity.contractGraph).toBeCalledTimes(0);
+    });
+
+    it("should update github graph", async () => {
+      // Given
+      const loaders = mockPluginLoaders();
+      const cache = mockCacheProvider();
+      const ghToken = ("t000ken": any);
```
decentralion: As discussed here (https://github.com/sourcecred/sourcecred/pull/1571#discussion_r368329145), I think the any-cast is unnecessary code smell.

src/backend/pluginPlan.js
```diff
@@ -40,3 +41,32 @@ export function pluginMirrorPlan(
...
+  cache: CacheProvider
+): GraphPlan {
+  return async (project: Project): Promise<Graph> => {
+    const tasks: Promise<Graph>[] = [];
+
+    if (project.discourseServer) {
+      tasks.push(discourse.createGraph(project.discourseServer, cache));
+    }
+
+    if (project.repoIds.length) {
```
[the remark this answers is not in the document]
Beanow: The reason it needs a token, is because the *name* of the mirror database is based on the GraphQL node ID of the repository. We don't know this ID without querying the remote API.

In the current `master` you can see this here:
https://github.com/sourcecred/sourcecred/blob/dea45fb8c1168addc4667be51d81bc40cc28a292/src/plugins/github/fetchGithubRepo.js#L61-L62

src/backend/pluginPlan.test.js
```diff
@@ -116,4 +116,87 @@ describe("src/backend/pluginPlan", () => {
       expect(identity.contractGraph).toBeCalledTimes(0);
     });
   });
+
+  describe("pluginGraphPlan", () => {
+    it("should update discourse graph", async () => {
```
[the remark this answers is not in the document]
Beanow: :laughing: Good find, this was from a previous type signature.

Beanow:
Closing this PR for now.

PR #1573, opened 2020-01-19
Backend: implement timelineCredPlan

Beanow:
Provides the Plan abstraction for `TimelineCred.compute`.

### DIFF
--- a/src/backend/timelineCredPlan.js
+++ b/src/backend/timelineCredPlan.js
@@ -1,0 +1,33 @@
+//@flow
+
+import {Graph} from "../core/graph";
+import {TaskReporter} from "../util/taskReporter";
+import {type TimelineCredParameters} from "../analysis/timeline/params";
+import {type PluginDeclaration} from "../analysis/pluginDeclaration";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+
+type ComputeFn = (opts: {|
+  graph: Graph,
+  params?: $Shape<TimelineCredParameters>,
+  plugins: $ReadOnlyArray<PluginDeclaration>,
+|}) => Promise<TimelineCred>;
+
+export function timelineCredPlan(
+  pluginDeclarations: $ReadOnlyArray<PluginDeclaration>,
+  reporter: TaskReporter,
+  compute: ComputeFn
+) {
+  return async (
+    graph: Graph,
+    params: ?$Shape<TimelineCredParameters>
+  ): Promise<TimelineCred> => {
+    reporter.start("compute-cred");
+    const cred = await compute({
+      graph,
+      params: params || {},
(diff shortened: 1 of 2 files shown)


src/backend/timelineCredPlan.js
```diff
@@ -0,0 +1,33 @@
...
+  compute: ComputeFn
+) {
+  return async (
+    graph: Graph,
+    params: ?$Shape<TimelineCredParameters>
+  ): Promise<TimelineCred> => {
+    reporter.start("compute-cred");
+    const cred = await compute({
+      graph,
+      params: params || {},
```
decentralion: just pass thru params directly, `TimelineCred.compute` allows null/undefined params and then sets defaults.

src/backend/timelineCredPlan.test.js
```diff
@@ -0,0 +1,60 @@
...
+      expect(cred).toEqual(fakeCred);
+      expect(compute).toBeCalledTimes(1);
+      expect(compute).toBeCalledWith({
+        graph: fakeGraph,
+        params,
+        plugins,
+      });
+    });
+
+    it("should give the right tasks to the TaskReporter", async () => {
```
decentralion: If you wanna make this test more robust, you could verify that the task starts when compute starts, and task finishes after compute finishes. I.e. set up mock compute promise that will hang until you (synchronously) call a callback, and then:

- verify tasks are empty
- start computeCred
- verify there's one active task and one entry
- finish computeCred
- verify two entries and no active tasks

src/backend/timelineCredPlan.js
Beanow: Not actually. What it allows is the `params` key to be missing from the object. Using null/undefined would cause a flow error here.

`params?: $Shape<TimelineCredParameters>` (from `TimelineCred.compute`)
versus
`params: ?$Shape<TimelineCredParameters>` (this function)

src/backend/timelineCredPlan.test.js
Beanow: I've not seen this before in the codebase, do we have an example for this?

To be honest though. I think the time and ordering accuracy of task reporter feedback doesn't warrant this type of unit tests. If with manual testing it provides enough feedback to know whether the program froze and/or which step it's at, that will suffice.

If we really care *that* much about whether a task starts and finishes at the right time, we should create a middleware for it and just test it once.

```js
function reportTask<T>(
	reporter: TaskReporter,
	name: string,
	yourTask: () => Promise<T>
): Promise<T> {
	reporter.start("name");
	const result = await yourTask();
	reporter.finish("name");
	return result;
}
```

This looks very familiar to the current implementation :smile:

Beanow:
Closing this PR for now.

PR #1574, opened 2020-01-19
Backend: implement createPlan

Beanow:
Ties together the underlying plans and APIs exposed by preceding PRs and uses them to create our full `LoadPlan`.

### DIFF
--- a/src/backend/loadPlan.js
+++ b/src/backend/loadPlan.js
@@ -4,6 +4,15 @@ import {Graph} from "../core/graph";
 import {type Project} from "../core/project";
 import {type TimelineCredParameters} from "../analysis/timeline/params";
 import {TimelineCred} from "../analysis/timeline/timelineCred";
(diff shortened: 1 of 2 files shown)


src/backend/loadPlan.js
```diff
@@ -38,6 +47,27 @@ export type LoadResult = {|
   +cred: TimelineCred,
 |};
 
+export function createPlan(
+  cache: CacheProvider,
+  githubToken: ?GithubToken,
+  pluginDeclarations: $ReadOnlyArray<PluginDeclaration>,
```
decentralion: Why does it take an array of plugin declarations, but hardcode the loaders? What happens if the user provides declarations that don't correspond to the hardcoded loader set?

Beanow: Noted for next attempt. Will use the same import-style dependency as the plugin loader itself uses.

Beanow:
Closing this PR for now.

PR #1575, opened 2020-01-19
API: implement loadDefaultPlan

Beanow:
This matches the current `load` function signature for easy transitioning.


Beanow:
Closing this PR for now.

PR #1576, opened 2020-01-19
Use stable stringify for cred.json & graph.json in load

Beanow:
The upcoming DataDirectory class will use stable stringify too.
But since that will affect the snapshots, make sure those are
updated before we switch to the new load implementation.


Beanow:
Going ahead with that rebase against master :]

Although the motivation of "a future change will do the same", is a little weak by moving this forward.
There are also no drawbacks that I could think of.

PR #1577, opened 2020-01-19
Use loadDefaultPlan instead of load

### LINKED ISSUE #1559: Reduce load complexity

Beanow:
:tada: When we reach this point, it fixes #1559


Beanow:
Closing this PR for now.

PR #1579, opened 2020-01-19
Enable eslint eqeqeq rule

decentralion:
The [eqeqeq rule][1] allows us to ban potentially confusing usage of the
type-coercing `==` equality operator. I've set it to "smart", so it
allows `== null` checks, as well as directly comparing literals or the
output of `typeof`.

Alongside activating the rule, I've fixed 15 violations. Inspired by
[review of #1560][2].

[1]: https://eslint.org/docs/rules/eqeqeq
[2]: https://github.com/sourcecred/sourcecred/pull/1560#discussion_r368331081

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
   ],
   rules: {
     camelcase: ["error", {properties: "never", allow: ["^_unused_.*"]}],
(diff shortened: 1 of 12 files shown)


decentralion:
@wchargin, I agree that taking the simpler exception makes sense. Will merge with the proposed change.

Issue #1580, opened 2020-01-20
GithubToken dependency to create graph from cache


Beanow:
Currently, to produce a Github graph from a populated mirror there is an unexpected dependency on a GithubToken. This is because:

1. The database name is based on the repository GraphQL ID.
2. We need to call `mirror.extract()` with the repository GraphQL ID.

Resolving that ID requires an upstream query.

I'm suggesting we remove this dependency by:
1. Deriving the database name from the `RepoId` instead.
2. Implementing a query on `Mirror` which allows us to resolve the repository GraphQL ID based on a `RepoId`.

Beanow:
cc @wchargin @decentralion

decentralion:
This makes sense to me; I'd support the change.  I think we could use base64 encoding like for [project ids](https://github.com/sourcecred/sourcecred/blob/f5a44a5a4fa93c810c8889e4574d75423ded4773/src/core/project.js#L71-L73) for the database names.

 I'm also willing to not block the load refactor sequence on this if it feels too much like a yak's yak, if you know what I mean.

Beanow:
Base64 is case-sensitive, so isn't suitable for all filesystems, we're using hex currently because of that.

Either way, step 1 is easy I should have that done in a minute.

Step 2 is a bit more complex. @wchargin would probably more about the possible pitfalls to this.
Afaik it would look something like: Find a `Repository` with a matching `name` primitive, join it over it's `owner` link, to validate the owner has a `login` primitive that matches. And return the `Repository` object ID.

wchargin:
In retrospect, I wish that I had used base32 instead of hex, because it
can be converted with `base32(1)` at the command line rather than having
to fiddle with a Node incantation that I can never remember. (I might
still change that.)

To be clear, your (1) and (2) are alternatives, not sequential steps,
right?

Your (2) is possible with direct DB access—something like

```sql
WITH repos AS (
    SELECT id FROM objects WHERE typename = 'Repository'
)
SELECT repo_id, repo_name, repo_owner
FROM (
    SELECT repos.id AS repo_id, primitives.value AS repo_name
    FROM repos
    JOIN primitives ON repos.id = primitives.object_id
    WHERE primitives.fieldname = 'name'
) NATURAL JOIN (
    SELECT repos.id AS repo_id, primitives.value AS repo_owner
    FROM repos
    JOIN links ON repos.id = links.parent_id
    JOIN primitives ON links.child_id = primitives.object_id
[...]

PR #1581, opened 2020-01-20
Github: derive mirror database name from repoId

Beanow:
Currently, to produce a Github graph from a populated mirror
there is an unexpected dependency on a GithubToken. See #1580.

This is step 1 to remove the dependency. It will allow us to
locate the Database without a GithubToken.

Test plan:
- `yarn test` will test the properties of `CacheId`.
- Manual testing to verify `fetchGithubRepo` has the desired side-effects on disk.

### DIFF
--- a/src/plugins/github/cacheId.js
+++ b/src/plugins/github/cacheId.js
@@ -1,0 +1,22 @@
+// @flow
+
+import {type RepoId, repoIdToString} from "./repoId";
+
+/**
+ * A derived ID to reference a cache layer.
+ */
+export opaque type CacheId: string = string;
+
+/**
+ * Derives the CacheId for a RepoId.
+ * Returned CacheId's will be:
+ * - Deterministic
+ * - Unique for this plugin
+ * - Lowercase
+ * - Safe to use for filenames
+ */
+export function cacheIdForRepoId(repoId: RepoId): CacheId {
+  const repoString = repoIdToString(repoId);
+  const repoStringHex = Buffer.from(repoString).toString("hex");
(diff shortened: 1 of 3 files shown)


src/plugins/github/cacheId.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import {type RepoId, repoIdToString} from "./repoId";
+
+/**
+ * A derived ID to reference a cache layer.
+ * Instances of this type will be:
+ * - Deterministic
+ * - Unique for this plugin
```
decentralion: Unique per RepoId

wchargin: These first two list items don’t really make sense. An instance of
`CacheId` is just a string, which isn’t deterministic or
nondeterministic, nor unique or shared.

It sounds like you want to say two separate things: first, that values
of this type (nit: “instances” suggests a class relationship, which
isn’t the case here) are lowercase and free of special characters, so
they can be portably used as file names; second, that `cacheIdForRepoId`
is a deterministic injective function. The second part is good to
document, but should be on the function, not the type.

Beanow: Currently interchangeable. Though should we ever decide to add a second type to derive from, they mustn't conflict. Subtle implementation note for the future :]

Beanow: Moved comment. And added "Safe to use for filenames".

Beanow: Also replaced "instances" with returned CacheId's.

PR #1583, opened 2020-01-21
Rename `Weights.defaultWeights` to `Weights.empty`

decentralion:
`empty` is a more descriptive name for a `Weights` object that has no
weights set, rather than `defaultWeights`.

In every case where we were importing `defaultWeights` as a direct
symbol, I switched to importing the whole module, as usage of
`Weights.empty` makes it clear that the empty object returned will be an
empty weights (as opposed to an empty list or some other empty type).
This is as proposed in the reviews from #1538.

Test plan: It's just a rename, so `yarn test` is sufficient.

### DIFF
--- a/src/analysis/timeline/params.js
+++ b/src/analysis/timeline/params.js
@@ -1,12 +1,7 @@
 // @flow
 
-import {
-  type Weights,
-  type WeightsJSON,
-  toJSON as weightsToJSON,
-  fromJSON as weightsFromJSON,
-  defaultWeights,
-} from "../../core/weights";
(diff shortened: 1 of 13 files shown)


decentralion:
> Are you open to separating out the whole-module imports into a prior change so that the rename can really be just a rename?

I'd rather not. It feels like considerably more work for no real benefit, since I don't see a plausible error that is introduced by this change, and would be caught by making that change first. I'm going to merge with an updated rationale for why `yarn test` is sufficient, per @Beanow's suggestion.

Issue #1557, opened 2020-01-13
Refactor 


decentralion:

Beanow:
Sounds like the way to go to me :]
Clean data types in the core and more flexibility everywhere in the system I think are indicators of that.

Aside from Initiatives, I'm imagining plugins could now create "# of downloads" types of edges with automatic weights based on the underlying statistic that weren't possible before.

And with a solid type, the operations should be easier to define too. An idempotent assign (think Object.assign) like we're doing for default weights overridden by the weights file, or multiplications done more internally.

PR #1584, opened 2020-01-21
Add `Weights.merge`

decentralion:
This adds a `merge` method to the weights module, which allows combining
multiple weights together. If any of the weights overlap (i.e. the same
address is specified in multiple Weights), then an error will be thrown.

In the future, we will likely extend the API so that the client can
specify how to resolve cases where the same address is present in
multiple weights.

The method is thoroughly unit tested.

This is part of work on #1557 (we'll want to be able to merge
`WeightedGraph`s.)

Test plan: Run `yarn test`

### DIFF
--- a/src/core/weights.js
+++ b/src/core/weights.js
@@ -48,6 +48,21 @@ export function copy(w: Weights): Weights {
   };
 }
 
+/**
+ * Merge multiple Weights together.
+ *
+ * The resultant Weights will have every weight specified by each of the
+ * input weights. If there are any overlaps (i.e. the same address is present
+ * in two or more of the input weights), an error will be thrown. In the future,
+ * we will likely modify this function to add a resolver that determines how to
+ * combine multiple overlapping weights.
+ */
+export function merge(ws: $ReadOnlyArray<Weights>): Weights {
+  const nodeWeights = MapUtil.merge(ws.map((x) => x.nodeWeights));
+  const edgeWeights = MapUtil.merge(ws.map((x) => x.edgeWeights));
+  return {nodeWeights, edgeWeights};
+}
+
 export type WeightsJSON = Compatible<{|
   +nodeWeights: {[NodeAddressT]: NodeWeight},
   +edgeWeights: {[EdgeAddressT]: EdgeWeight},
--- a/src/core/weights.test.js
+++ b/src/core/weights.test.js
@@ -2,6 +2,7 @@
 
 import stringify from "json-stable-stringify";
 import {NodeAddress, EdgeAddress} from "../core/graph";
+import {type Weights as WeightsT} from "./weights";
 import * as Weights from "./weights";
 
 describe("core/weights", () => {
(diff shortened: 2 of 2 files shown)


src/core/weights.js
```diff
@@ -48,6 +53,42 @@ export function copy(w: Weights): Weights {
...
+ * The resultant Weights will have every weight specified by each of the
+ * input weights. If the inputs provide contradictory weights for any node
+ * or edge address, an error will be thrown.
+ */
+export function merge(ws: $ReadOnlyArray<Weights>): Weights {
+  const weights = empty();
+  for (const {nodeWeights, edgeWeights} of ws) {
+    for (const [addr, weight] of nodeWeights.entries()) {
+      const existing = weights.nodeWeights.get(addr);
+      if (existing != null && existing != weight) {
```
wchargin: Why does the `eqeqeq` lint rule allow `existing != weight`?

decentralion: Probably an outdated base branch. Willfix.

decentralion:
@wchargin: See #1557 for context. Basically, I want core to have a type called `WeightedGraph`, which includes a graph along with node and edge weights. Then, when plugins create the `WeightedGraph`, they include both node-specific weights (e.g. if an initiative has a hardcoded weight) and also the type level weights. Type level weights are encoded via the prefix. So hypothetically if we generated two GitHub `WeightedGraph`s and merged them, then the merge would succeed because the (overlapping) type prefix weights would be the same.

In practice, each Plugin will produce its own `WeightedGraph`, and I expect plugins not to be providing overlapping weights, so this behavior might not be needed. If so, we could do the simpler thing and use `MapUtil.merge` (which is more restrictive). However, when multiple `WeightedGraph`s do have overlapping but identical weights, merging them together is clearly reasonable, so I decided I'd write a contract here which biases towards being permissive, given that we can do the right thing.

Beanow:
> but there isn’t an natural way to define a
contraction operation in node-weighted graphs.
>
> Another perspective: When contracting nodes u and v with weights p
and q, why is it that the final weight should be p (or q, assuming
equal weights), rather than p + q or p · q or max(p, q)?

Good point. And we will explicitly need a node contraction. The identities plugin comes to mind.
What about requiring a function to resolve conflicts? Which could be to throw because they should be disjoint, or any of the mentioned ones, add, multiply, max, ...

src/core/weights.js
```diff
@@ -48,6 +53,42 @@ export function copy(w: Weights): Weights {
...
 
+/**
+ * Merge multiple Weights together.
+ *
+ * The resultant Weights will have every weight specified by each of the
+ * input weights. If the inputs provide contradictory weights for any node
+ * or edge address, an error will be thrown.
+ */
+export function merge(ws: $ReadOnlyArray<Weights>): Weights {
+  const weights = empty();
```
Beanow: Nit: because `empty()` here doesn't mention weights. Would like if the variable was typed.
`const weights: Weights = empty();`

wchargin: Up to @decentralion, but this looks fine to me as is. We know that it’s
a `Weights` because the variable is called `weights`, the module is
`weights.js`, and the function takes a bunch of `Weights` and returns a
`Weights`. I don’t think that this suggestion makes the code materially
better.

Beanow: Yeah this is a nit, cognitive load preference.
> the module is
weights.js, and the function takes a bunch of Weights and returns a
Weights

Having the typed variable doesn't have me thinking "did we import `empty` from somewhere else?", "being named `weights`, I'm assuming it's `Weights` typed... right?".

decentralion:
You raise good points about how it's not obvious that my (implicit) choice of resolver function is appropriate, and that we should allow arbitrary resolvers. For now, since I don't actually need resolution for my plans, I am going to leave that out, and simply error whenever there is an ambiguity; once these become an issue in practice, we can go ahead and add resolvers.

This has the added advantage of massively simplifying the implementation. PTAL.

src/core/weights.test.js
```diff
@@ -69,4 +70,71 @@ describe("core/weights", () => {
...
+      const w1 = simpleWeights([["foo", 3]], [["bar", 2, 3]]);
+      const w2 = simpleWeights([["foo", 3]], [["bar", 2, 3]]);
+      expect(() => Weights.merge([w1, w2])).toThrowError("duplicate key");
+    });
+    it("errors on conflicting node weights", () => {
+      const w1 = simpleWeights([["foo", 3]], []);
+      const w2 = simpleWeights([["foo", 4]], []);
+      expect(() => Weights.merge([w1, w2])).toThrowError("duplicate key");
+    });
+    it("errors on conflicting edge weights (forwards)", () => {
```
Beanow: Nit: the overlapping edge weights without conflicts is missing.

decentralion: good point. merged before seeing this comment. going to ignore it for now.

PR #1585, opened 2020-01-21
Add `Weights.equals`

decentralion:
This adds an `equals` method to the Weights module, which reports
whether two weights are equal. This is really just a wrapper around
calling `deepEqual` (since the Weights are just an object consisting of
two maps). However, this makes the module more consistent with `Graph`
(which provides an `equals` method), and ensures that clients can
confidently check equality without needing to reason about how Weights
are implemented.

Test plan: This commit adds an extremely simple wrapper function, with
added unit tests. `yarn test` passes.

### DIFF
--- a/src/core/weights.js
+++ b/src/core/weights.js
@@ -1,6 +1,7 @@
 // @flow
 
 import * as MapUtil from "../util/map";
+import deepEqual from "lodash.isequal";
 import {
   type NodeAddressT,
   type EdgeAddressT,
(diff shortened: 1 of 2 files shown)


decentralion:
Much like `Graph.equals`, it will mostly be used for test code. (Having `WeightedGraph.equals` is convenient for testing `WeightedGraph.merge`). I'm also fine with just having a helper function in `weightedGraph.test.js`.

PR #1587, opened 2020-01-22
Create a node for each Discourse like

decentralion:
*Let's use the syntax `(node)` to represent some node, and `> edge >` to
represent some edge.*

In the past, for every like, we would create the following graph
structure:

`(user) > likes > (post)`

As of this commit, we instead create:

`(user) > createsLike > (like) > likes > (post)`

We make this change because we want to mint cred for likes. Arguably,
this is more robust than minting cred for activity: something being
liked signals that at least one person in the community found a post
valuable, so you can think of moving cred minting away from raw activity
and towards likes as a sort of implicit "cred review".

Create a node for each like is a somewhat hacky way to do it--in
principle, we should have a heuristic which increases the cred weight of
a post based on the number of likes it has received--but it is expedient
so we can prototype this quickly.

Obviously, this is not robust to Sibyll attacks. If we decide to adopt
this, in the medium term we can add some filtering logic so that e.g. a
user must be whitelisted for their likes to mint cred. (And, in a nice
recursive step, the whitelist can be auto-generated from the last week's
cred scores, so that e.g. every user with at least 50 cred can mint more
cred.) I think it's OK to put in a Sibyll-vulnerable mechanism here
because SourceCred is still being designed for high trust-level
communities, and the existing system of minting cred for raw activity is
also vulnerable to Sibyll and spam attacks.

Test plan: Unit tests updated; also @s-ben can report back on whether
this is useful to him in demo-ing SourceCred [on MakerDAO][1].

If we merge this, we should simultaneously explicitly set the weight to
like nodes to 0 in our cred instance, so that we can separate merging
this feature from actually changing our own cred (which should go
through a separate review).

[1]: https://forum.makerdao.com/t/possible-data-source-for-determining-compensation

### DIFF
--- a/src/plugins/discourse/createGraph.js
+++ b/src/plugins/discourse/createGraph.js
@@ -23,11 +23,12 @@ import {
   postRepliesEdgeType,
   topicContainsPostEdgeType,
   likesEdgeType,
+  createsLikeEdgeType,
   referencesTopicEdgeType,
   referencesUserEdgeType,
   referencesPostEdgeType,
 } from "./declaration";
-import {userAddress, topicAddress, postAddress} from "./address";
+import {userAddress, topicAddress, postAddress, likeAddress} from "./address";
 import {
   type DiscourseReference,
   parseLinks,
@@ -54,14 +55,20 @@ export function topicNode(serverUrl: string, topic: Topic): Node {
 export function postNode(
   serverUrl: string,
   post: Post,
-  topicTitle: string
+  description: string
 ): Node {
-  const url = `${serverUrl}/t/${String(post.topicId)}/${String(
-    post.indexWithinTopic
-  )}`;
-  const descr = `[post #${post.indexWithinTopic} on ${topicTitle}](${url})`;
   const address = postAddress(serverUrl, post.id);
-  return {timestampMs: post.timestampMs, address, description: descr};
+  return {timestampMs: post.timestampMs, address, description};
+}
+
+export function likeNode(
+  serverUrl: string,
+  like: LikeAction,
+  postDescription: string
+): Node {
+  const address = likeAddress(serverUrl, like);
+  const description = `❤️ by ${like.username} on post ${postDescription}`;
+  return {timestampMs: like.timestampMs, address, description};
 }
 
 export function authorsTopicEdge(serverUrl: string, topic: Topic): Edge {
@@ -128,6 +135,21 @@ export function postRepliesEdge(
   };
 }
 
+export function createsLikeEdge(serverUrl: string, like: LikeAction): Edge {
+  const address = EdgeAddress.append(
+    createsLikeEdgeType.prefix,
+    serverUrl,
+    like.username,
+    String(like.postId)
+  );
+  return {
+    address,
+    timestampMs: like.timestampMs,
+    src: userAddress(serverUrl, like.username),
+    dst: likeAddress(serverUrl, like),
+  };
+}
+
 export function likesEdge(serverUrl: string, like: LikeAction): Edge {
   const address = EdgeAddress.append(
     likesEdgeType.prefix,
@@ -138,7 +160,7 @@ export function likesEdge(serverUrl: string, like: LikeAction): Edge {
   return {
     address,
     timestampMs: like.timestampMs,
-    src: userAddress(serverUrl, like.username),
+    src: likeAddress(serverUrl, like),
     dst: postAddress(serverUrl, like.postId),
   };
 }
@@ -153,6 +175,7 @@ class _GraphCreator {
   serverUrl: string;
   data: ReadRepository;
   topicIdToTitle: Map<TopicId, string>;
+  postIdToDescription: Map<PostId, string>;
 
   constructor(serverUrl: string, data: ReadRepository) {
     if (serverUrl.endsWith("/")) {
@@ -162,6 +185,7 @@ class _GraphCreator {
     this.data = data;
     this.graph = new Graph();
     this.topicIdToTitle = new Map();
+    this.postIdToDescription = new Map();
 
     for (const username of data.users()) {
       this.graph.addNode(userNode(serverUrl, username));
@@ -174,18 +198,23 @@ class _GraphCreator {
     }
 
     for (const post of data.posts()) {
-      this.addPost(post);
+      const topicTitle =
+        this.topicIdToTitle.get(post.topicId) || `[unknown topic]`;
+      const url = `${this.serverUrl}/t/${String(post.topicId)}/${String(
+        post.indexWithinTopic
+      )}`;
+      const description = `[post #${post.indexWithinTopic} on ${topicTitle}](${url})`;
+      this.addPost(post, description);
+      this.postIdToDescription.set(post.id, description);
     }
 
     for (const like of data.likes()) {
-      this.graph.addEdge(likesEdge(serverUrl, like));
+      this.addLike(like);
     }
   }
 
-  addPost(post: Post) {
-    const topicTitle =
-      this.topicIdToTitle.get(post.topicId) || "[unknown topic]";
-    this.graph.addNode(postNode(this.serverUrl, post, topicTitle));
+  addPost(post: Post, description: string) {
+    this.graph.addNode(postNode(this.serverUrl, post, description));
     this.graph.addEdge(authorsPostEdge(this.serverUrl, post));
     this.graph.addEdge(topicContainsPostEdge(this.serverUrl, post));
     this.maybeAddPostRepliesEdge(post);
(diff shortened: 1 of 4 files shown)


Beanow:
> If we merge this, we should simultaneously explicitly set the weight to
like nodes to 0 in our cred instance, so that we can separate merging
this feature from actually changing our own cred (which should go
through a separate review).

I'm not sure this:
`(user) > likes > (post)`
Is fully equivalent to this:
`(user) > createsLike > (like: 0 weight) > likes > (post)`
I'm expecting parameters like `alpha` to change the impact of a like this way.

And doing a node contraction instead to be sure they are equivalent seems like needless extra work.

So I would have a strong preference for bringing the _community consensus_ forward. Discussing the proposed changes before merging. Perhaps including a prototype to show the new behavior with example weights.

vsoch:
For those that need a refresher on [Sybil Attacks](https://www.geeksforgeeks.org/sybil-attack/).

I'd want to ask what is the current standard for the entities that make up nodes. Intuitively it would be (somewhat) tangible things like people, topics, pull requests, etc. Adding a like as a full fledged entity first makes me think about how this fits into a long term graph. If we think of many different kinds of reactions to posts, there are many different ones, and so representing just a "like" doesn't do the future landscape of reactions full justice. An alternative could be to have a "reaction" node that then has an attribute to describe the kind (does the graph support nodes with attributes?).

My gut reaction is that it would be more appropriate to adjust the weight on the link - the links are relationships, the nodes are entities, and to model a relationship as a node seems off.

Beanow:
> I'd want to ask what is the current standard for the entities that make up nodes. Intuitively it would be (somewhat) tangible things like people, topics, pull requests, etc.

Don't know how much sense this makes for SC nodes, but here's a way of thinking about it from DDD.
https://en.wikipedia.org/wiki/Domain-driven_design#Building_blocks
And the difference it makes between a value object and entities.

Entities are **not** defined by their attributes.
_A Topic title is changd, it's still the same Topic._

Value objects **are** defined by their attributes.
_Change the zipcode of an address, it's not the same address._

The commit may be in a grey area, but all other nodes so far are _entities_.
I would argue a like is a _value object_.

Likes only have (same as edges):
- a source
- a destination

If you change the source or destination.
```diff
Source:
- Beanow likes Post-123
+ Dandelion likes Post-123

Destination:
- Beanow likes Post-123
+ Beanow likes Post-555
```
It's not the same like.

decentralion:
> I'm not sure this:
> `(user) > likes > (post)`
> Is fully equivalent to this:
> `(user) > createsLike > (like: 0 weight) > likes > (post)`
> I'm expecting parameters like `alpha` to change the impact of a like this way.

This is a good point. If we set the node weight of the like nodes to `0`, then this PR would actually _decrease_ the cred-significance of likes, because `alpha` fraction of the cred that would flow through the like will instead "evaporate" as it passes through the like node.


> Given that we're interested in allowing plugins to set weights with their own logic (WeightedGraph).
And we're interested in more versatile ways to mint cred (supernodes). It seems like a very temporary solution.

Yeah, it definitely is a temporary solution. It's actually motivated less by SC's own use case than by something we saw on another project's Discourse instance. Basically, there was a user who had made a very large number of relatively low-effort (and low-like) posts, and was receiving a disproportionate share of the cred, while other higher-signal users had low cred. On a hunch, I explored the forum data, and found that the higher signal posters had fewer posts, but far more likes received. 

This gave me the idea that we could mint cred based on likes (which represent implicit review of activity) rather than on raw activity. Since we're in active dialogue with that other project, I wanted to prototype something super quickly to show them. The feedback from that project was that this version seems better than the original. 

I sent a PR for this branch mostly to prompt discussion -- I don't feel it's super important that it merges in it's current form, esp. since we are pretty close (via WeightedGraph) to being able to implement it much more cleanly. 

> Perhaps including a prototype to show the new behavior with example weights.

I've done an analysis of the new behavior. I'm going to post it on Discourse instead of here, so that everyone in the community can take a look and chime in.

> I'd want to ask what is the current standard for the entities that make up nodes. Intuitively it would be (somewhat) tangible things like people, topics, pull requests, etc. Adding a like as a full fledged entity first makes me think about how this fits into a long term graph.

I would go with a minimalist description of a node, and say: "a node is anything which **should** accumulate cred". (One will note that this is somewhat subjective, which is fine.) I think it's clear that contributors, pull requests, etc, should accumulate cred. Likes probably should not. So I agree that likes should not be nodes.

>  If we think of many different kinds of reactions to posts, there are many different ones, and so representing just a "like" doesn't do the future landscape of reactions full justice. 

On the GitHub side, we have distinct edge types for each positive reaction (we ignore the 👎 reaction for now)

> (does the graph support nodes with attributes?).

Not really. You can see the types here:

https://github.com/sourcecred/sourcecred/blob/c7f386933d3415595b1a27037cc4d882725b71d4/src/core/graph.js#L121-L144

decentralion:
Update: I've posted an analysis of how this changes the Discourse cred here: https://discourse.sourcecred.io/t/minting-discourse-cred-on-likes-not-posts/603

decentralion:
I've come to believe we should merge this change as-is. By having likes and posts be separate nodes (which both mint cred), we can express cred minting formulae like `nLikes * likesConstant + nPosts * postConstant`

If we don't make likes correspond to nodes, we could still set custom weights on each post based on the number of likes... but we're constrained to expressing things like `nPosts * postConstant + nLikes * nPosts * likeConstant`

Because for any individual post, right now we can mint it `postConstant` cred directly, and `nLikeForThisPost * likeConstant` implicitly, if we conceptualize each post as "owning" its likes.

However, without this change, then the minted cred would be `postConstant * (nLikes * likeConstant + 1)`, because `postConstant`, applied at the type level, would be multiplicative with the specific node weight.

Feel free to ask questions if this is unclear; hopefully as we start to document the algorithm more cleanly it will be easier to communicate this.

As for why this makes sense conceptually: it basically means that Likes are contributions. It's counter-intuitive, but it makes sense if you think about it. People are doing some work by reading posts, and expressing which posts they found valuable. That information is itself valuable (which is why we are consuming it in the algorithm). Right now the amount of cred we mint per like is definitely in excess of the like's intrinsic value, which is a (temporary, acceptable) hack since we're still in an alpha state.

decentralion:
Note: If we were using semver to signify cred-changing changes, we would need to bump version numbers here. We're not doing that right now, but it's interesting to think about.

wchargin:
(Semver for cred-changing changes seems a long way off to me, for the
same reason that Prettier takes a semver exemption for code formatting
changes.)

src/plugins/discourse/createGraph.js
```diff
@@ -173,19 +199,26 @@ class _GraphCreator {
...
+      this.graph.addNode(likeNode(serverUrl, like, postDescription));
       this.graph.addEdge(likesEdge(serverUrl, like));
+      this.graph.addEdge(createsLikeEdge(serverUrl, like));
     }
   }
 
-  addPost(post: Post) {
+  addPost(post: Post): string {
     const topicTitle =
       this.topicIdToTitle.get(post.topicId) || "[unknown topic]";
```
Beanow: Nit: returning just the post description here seems weird.
Prefer to factor out a `postDescription(post: Post): string` method, which both `addPost()` use, and the constructor's loop uses.

src/plugins/discourse/createGraph.js
```diff
@@ -173,19 +199,26 @@ class _GraphCreator {
       this.graph.addEdge(authorsTopicEdge(serverUrl, topic));
     }
 
+    const postDescriptions: Map<PostId, string> = new Map();
```
Beanow: Nit: inconsistent with `topicIdToTitle` which sets the map as a class property.

src/plugins/discourse/createGraph.js
```diff
@@ -173,19 +199,26 @@ class _GraphCreator {
...
+      const description = this.addPost(post);
+      postDescriptions.set(post.id, description);
     }
 
     for (const like of data.likes()) {
+      const postDescription =
+        postDescriptions.get(like.postId) || "[unknown post]";
+      this.graph.addNode(likeNode(serverUrl, like, postDescription));
       this.graph.addEdge(likesEdge(serverUrl, like));
+      this.graph.addEdge(createsLikeEdge(serverUrl, like));
```
Beanow: Nit: inconsistent with other node types, which use an `addLike` style method.

src/plugins/discourse/createGraph.js
decentralion: I refactored it to be consistent with `topicIdToTitle`; the weird return is removed.

src/plugins/discourse/createGraph.js
decentralion: Refactored for consistency.

src/plugins/discourse/createGraph.js
decentralion: Refactored for consistency.

PR #1589, opened 2020-01-23
Rename `Weights.defaultWeights` to `Weights.empty`

decentralion:
`empty` is a more descriptive name for a `Weights` object that has no
weights set, rather than `defaultWeights`.

In every case where we were importing `defaultWeights` as a direct
symbol, I switched to importing the whole module, as usage of
`Weights.empty` makes it clear that the empty object returned will be an
empty weights (as opposed to an empty list or some other empty type).
This is as proposed in the reviews from #1538.

Test plan: It's just a rename and change to imports, so `yarn flow`
would catch any issues. `yarn test` passes.

### DIFF
--- a/src/analysis/timeline/params.js
+++ b/src/analysis/timeline/params.js
@@ -1,12 +1,7 @@
 // @flow
 
-import {
-  type Weights,
-  type WeightsJSON,
-  toJSON as weightsToJSON,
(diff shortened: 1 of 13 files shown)


decentralion:
When tests pass, I will merge with carry-over approval from #1583, which I merged into the wrong branch.

PR #1590, opened 2020-01-24
Add core/weightedGraph

decentralion:
This module is a simple data type which contains a graph and associated
weights. It provides methods for JSON (de)serialization, constructing
new WeightedGraphs, and for merging them.

Test plan: See included unit tests. The unit tests are simple because
the data type and associated methods are quite simple; the underlying
functions for Graphs and Weights have more extensive testing.

Progress towards #1557.

### DIFF
--- a/src/core/weightedGraph.js
+++ b/src/core/weightedGraph.js
@@ -1,0 +1,54 @@
+// @flow
+
+import {Graph, type GraphJSON} from "./graph";
+import {type Weights as WeightsT, type WeightsJSON} from "./weights";
+import * as Weights from "./weights";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
+
+/** The WeightedGraph a Graph alongside associated Weights
+ *
+ * Any combination of Weights and Graph can make a valid WeightedGraph. If the
+ * Weights contains weights for node or edge addresses that are not present in
+ * the graph, then those weights will be ignored. If the graph contains nodes
+ * or edges which do not correspond to any weights, then default weights will
+ * be inferred.
+ */
+export type WeightedGraph = {|+graph: Graph, +weights: WeightsT|};
+
+const COMPAT_INFO = {type: "sourcecred/weightedGraph", version: "0.1.0"};
+
+export type WeightedGraphJSON = Compatible<{|
+  +graphJSON: GraphJSON,
+  +weightsJSON: WeightsJSON,
+|}>;
+
+/**
+ * Create a new, empty WeightedGraph.
+ */
+export function empty(): WeightedGraph {
+  return {graph: new Graph(), weights: Weights.empty()};
+}
+
+export function toJSON(w: WeightedGraph): WeightedGraphJSON {
+  const graphJSON = w.graph.toJSON();
+  const weightsJSON = Weights.toJSON(w.weights);
+  return toCompat(COMPAT_INFO, {graphJSON, weightsJSON});
+}
+
+export function fromJSON(j: WeightedGraphJSON): WeightedGraph {
+  const {graphJSON, weightsJSON} = fromCompat(COMPAT_INFO, j);
+  const graph = Graph.fromJSON(graphJSON);
+  const weights = Weights.fromJSON(weightsJSON);
+  return {graph, weights};
+}
+
+/**
+ * Merge multiple WeightedGraphs together.
+ *
+ * This delegates to the semantics of Graph.merge and Weights.merge.
+ */
+export function merge(ws: $ReadOnlyArray<WeightedGraph>): WeightedGraph {
+  const graph = Graph.merge(ws.map((w) => w.graph));
+  const weights = Weights.merge(ws.map((w) => w.weights));
+  return {graph, weights};
+}
--- a/src/core/weightedGraph.test.js
+++ b/src/core/weightedGraph.test.js
@@ -1,0 +1,63 @@
+// @flow
+
+import * as Weights from "./weights";
+import {Graph} from "./graph";
+import * as WeightedGraph from "./weightedGraph";
+import * as GraphTest from "./graphTestUtil";
+
+describe("core/weightedGraph", () => {
+  function expectEqual(wg1, wg2) {
+    expect(wg1.graph.equals(wg2.graph)).toBe(true);
+    expect(wg1.weights).toEqual(wg2.weights);
+  }
+
+  describe("empty", () => {
+    it("empty produces an empty WeightedGraph", () => {
+      const {weights, graph} = WeightedGraph.empty();
+      expect(graph.equals(new Graph())).toBe(true);
+      expect(weights).toEqual(Weights.empty());
+    });
+  });
+
+  describe("toJSON/fromJSON", () => {
+    it("works for an empty WeightedGraph", () => {
+      const wg = WeightedGraph.empty();
+      const wgJSON = WeightedGraph.toJSON(wg);
+      const wg_ = WeightedGraph.fromJSON(wgJSON);
+      expectEqual(wg, wg_);
+    });
+    it("works for a non-empty WeightedGraph", () => {
+      const node = GraphTest.node("foo");
+      const graph = new Graph().addNode(node);
+      const weights = Weights.empty();
+      weights.nodeWeights.set(node.address, 5);
+      const wg = {graph, weights};
+      const wgJSON = WeightedGraph.toJSON(wg);
+      const wg_ = WeightedGraph.fromJSON(wgJSON);
+      expectEqual(wg, wg_);
+    });
+  });
+
+  describe("merge", () => {
+    it("merge works on nontrivial graph", () => {
+      // Not attempting to validate edge case semantics here, since this is just
+      // a wrapper around Graph.merge and WeightedGraph.merge; those functions
+      // are tested more thoroughly.
+      const n1 = GraphTest.node("foo");
+      const n2 = GraphTest.node("bar");
+      const g1 = new Graph().addNode(n1);
+      const g2 = new Graph().addNode(n2);
+      const w1 = Weights.empty();
+      w1.nodeWeights.set(n1.address, 1);
+      const w2 = Weights.empty();
+      w2.nodeWeights.set(n2.address, 2);
+      const wg1 = {graph: g1, weights: w1};
+      const wg2 = {graph: g2, weights: w2};
+      const g = Graph.merge([g1, g2]);
+      const w = Weights.merge([w1, w2]);
+      const wg = WeightedGraph.merge([wg1, wg2]);
+      const wg_ = {weights: w, graph: g};
+      expectEqual(wg, wg_);
+    });
+  });
+});


src/core/weightedGraph.js
```diff
@@ -0,0 +1,49 @@
...
+  +graph: Graph,
+  +weights: WeightsT,
+|};
+
+const COMPAT_INFO = {type: "sourcecred/weightedGraph", version: "0.1.0"};
+
+export type WeightedGraphJSON = Compatible<{|
+  +graphJSON: GraphJSON,
+  +weightsJSON: WeightsJSON,
+|}>;
```
Beanow: I notice we're nesting `Compatible<T>` types here. How does this affect versioning and/or upgrade path of these types?

For example, currently we have:
- Graph v0.8.0
- Weights v0.2.0

Should we create a Graph 0.9.0, which has a sensible upgrade function from v0.8.0.
Would we be able to express that `WeightedGraphJSON` v0.1.0 is the composition of:
- Graph v0.8.0 | v0.9.0
- Weights v0.2.0

src/core/weightedGraph.test.js
```diff
@@ -0,0 +1,56 @@
...
+import {Graph} from "./graph";
+import * as WeightedGraph from "./weightedGraph";
+import * as GraphTest from "./graphTestUtil";
+
+describe("core/weightedGraph", () => {
+  function expectEqual(wg1, wg2) {
+    expect(wg1.graph.equals(wg2.graph)).toBe(true);
+    expect(wg1.weights).toEqual(wg2.weights);
+  }
+  it("empty produces an empty WeightedGraph", () => {
```
Beanow: Nit: to prevent noise when adding tests, should use `describe()` consistently.
```
describe "empty"
  it "produces ____"
```

src/core/weightedGraph.test.js
```diff
@@ -0,0 +1,56 @@
...
+    const {weights, graph} = WeightedGraph.empty();
+    expect(graph.equals(new Graph())).toBe(true);
+    expect(weights).toEqual(Weights.empty());
+  });
+  describe("toJSON/fromJSON", () => {
```
Beanow: Depending on the nested `Compatible<T>` question above.
Would appreciate some test coverage for the top-level and underlying versions we expect to support.
Similar to the Project tests
https://github.com/sourcecred/sourcecred/blob/325487a3ea569e1a0d176b60f39245fb2243804d/src/core/project.test.js#L46

src/core/weightedGraph.test.js
```diff
@@ -0,0 +1,56 @@
...
+      const graph = new Graph().addNode(node);
+      const weights = Weights.empty();
+      weights.nodeWeights.set(node.address, 5);
+      const wg = {graph, weights};
+      const wgJSON = WeightedGraph.toJSON(wg);
+      const wg_ = WeightedGraph.fromJSON(wgJSON);
+      expectEqual(wg, wg_);
+    });
+  });
+  it("merge works on nontrivial graph", () => {
```
Beanow: Nit: same as above
```
describe "merge"
  it "works on ____"
```

src/core/weightedGraph.test.js
decentralion: Are you saying I should add a test case that validates that it upgrades from 0.1.0 (ie. the current version)? 

Personally, I don't think this is needed. Graph (and WeightedGraph) is re-generated ephemerally, but not persisted, so we don't actually need upgrade handling yet. (For project, which is actually config and not data, it's reasonable that we invest more effort in handling upgrades.)

src/core/weightedGraph.test.js
decentralion: I think it's fine, since we don't have any other tests for empty, nor is it clear why we would add them. If/when we do add another test case, indenting this one test is not that noisy.

src/core/weightedGraph.test.js
decentralion: yeah, in this case it's easier for me to imagine adding more tests, so i'm more open to adding a describe block.

i'll even do it for empty too, i don't think it's very necessary, but it is more consistent.

src/core/weightedGraph.js
decentralion: Interesting question. I suppose that we'd need to bump WGJ to v0.2.0 concurrently with bumping Graph to v0.9.0. Outside the scope of this PR (and not a current priority), but it'd be interesting to see some facilities for handling these nested versioning cases.

decentralion:
cc @vsoch and @BrianLitwin for general interest. also feel free to add a review.

src/core/weightedGraph.js
```diff
@@ -0,0 +1,49 @@
...
+
+import {Graph, type GraphJSON} from "./graph";
+import {type Weights as WeightsT, type WeightsJSON} from "./weights";
+import * as Weights from "./weights";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
+
+export type WeightedGraph = {|
+  +graph: Graph,
+  +weights: WeightsT,
+|};
```
wchargin: What are the semantics of a `WeightedGraph`?

For instance, is it the case that each address in `weights` corresponds
to a node/edge in the graph (say, that the weights are *valid*)? Is it
also the case that each node/edge in the graph has a weight (say, that
the weights are *complete*)?

Is there any meaning associated with weights, or any natural domain for
them? For instance, for some kinds of weighted graphs, negative and zero
weights make sense, and combination along paths is done additively. For
others, only positive weights make sense, and combination is done
multiplicatively. In others, weights may be any real or ±∞, with
combination given by max(—, —). (The underlying theme here is usually
that [the weights have the structure of a semiring][app].) Does the type
defined in this module have an opinion?

[app]: https://www.cs.colostate.edu/~cs575dl/Sp2015/Lectures/APP.pdf

src/core/weightedGraph.js
wchargin: If `Graph.fromJSON` supports converting v0.8.0 to v0.9.0 transparently
in its `fromJSON` (perhaps via `compat` resolvers), then `WeightedGraph`
won’t need to bump. Otherwise, the `Graph` change is “breaking”, and
`WeightedGraph` will need to bump. As far as I can tell, it’s the same
as normal semver. (No action required.)

src/core/weightedGraph.test.js
```diff
@@ -0,0 +1,63 @@
...
+      expect(graph.equals(new Graph())).toBe(true);
+      expect(weights).toEqual(Weights.empty());
+    });
+  });
+
+  describe("toJSON/fromJSON", () => {
+    it("works for an empty WeightedGraph", () => {
+      const wg = WeightedGraph.empty();
+      const wgJSON = WeightedGraph.toJSON(wg);
+      const wg_ = WeightedGraph.fromJSON(wgJSON);
```
wchargin: nit: We generally call these `wg2` rather than `wg_`\*, which is
significantly easier to read, especially in contexts where the code is
underlined (e.g., stock Vim with `set cursorline`). Could you change
these accordingly? (In cases where “`2`” is ambiguous, we also have
precedent for `newWg`.)

\* `git grep '[^ ]_ '` shows that there is no precedent for the form
used here.

src/core/weightedGraph.test.js
Beanow: Not the upgrade path per say. But `Compatible<T>` adds flow coverage for the shape of the body. And having the compat info header. Not which version that compat info should have.

Testing we have composed the correct versions seems useful for this nesting case.

vsoch:
I'm taking my first real look at the code base this weekend, I might have something intelligent to say after that (but no promises!)

src/core/weightedGraph.test.js
decentralion: I'm sorry, I still don't understand what you're asking for. Would you be willing to add this as a follow-on PR? This is PR is upstream of a lot of work I want to do in a timely fashion, so I'm feeling some impatience to not block this on non-structural issues.

src/core/weightedGraph.js
decentralion: The type defined in this module is unopinionated with respect to all of these questions. All it does is provide a way to package weights alongside a graph. I've added a docstring to this effect.

The Weights type itself is unopinionated about how weight composition works. The semantics are defined in weightEvaluator.js:

https://github.com/sourcecred/sourcecred/blob/9784927402bf4fde8716949b71092d6dcd42d72d/src/analysis/weightEvaluator.js#L16-L28

However, it's likely that we will create more evaluators in the future, which apply different interpretations to the weights.

src/core/weightedGraph.test.js
decentralion: Not sure how you're coming to that conclusion that there's no precedent for this form. Here are some pre-existing examples in the codebase.

https://github.com/sourcecred/sourcecred/blob/9784927402bf4fde8716949b71092d6dcd42d72d/src/analysis/timeline/params.test.js#L28-L32

https://github.com/sourcecred/sourcecred/blob/9784927402bf4fde8716949b71092d6dcd42d72d/src/cli/common.test.js#L78-L79

https://github.com/sourcecred/sourcecred/blob/9784927402bf4fde8716949b71092d6dcd42d72d/src/core/graph.test.js#L1371-L1379

Since this is pattern is already widespread, I don't think there's value in avoiding this particular instance in this PR. However, I'd be fine accepting a change which removes them en masse and enforces compliance automatically.

wchargin: Ah—I was at an old checkout due to some archaeology; sorry. Feel free to
defer given precedent.

src/core/weightedGraph.test.js
Beanow: What I mean is, there is no test coverage for the contents of COMPAT_INFO. Neither through flow nor unit tests. Since we're composing them I think it would make sense to test we composed them correctly.

```diff
[
- COMPAT_INFO,
  {
    graphJSON: [
-     COMPAT_INFO,
      graph
    ],
    weightsJSON: [
-     COMPAT_INFO,
      weights
    ]
  }
]
```

Or to put it differently, I would like to see a test *fail* when we accidentally make an incompatible version upgrade one one of the parts.

```diff
Expected:
[
  {type: "sourcecred/weightedGraph", version: "0.1.0"},
+ {type: "sourcecred/graph", version: "0.8.0"},
  {type: "sourcecred/weights", version: "0.2.0"},
]

Actual:
[
  {type: "sourcecred/weightedGraph", version: "0.1.0"},
- {type: "sourcecred/graph", version: "0.9.0"},
  {type: "sourcecred/weights", version: "0.2.0"},
]
```

decentralion: Sorry, I still don't understand your ask well enough to faithfully add this test. Are you suggesting that we have a snapshot test that verifies that the composed compats were as expected? Or an upgrade test that validates that we can correctly load a compat with the current version?

decentralion: I'm going to merge this, but please feel free to send a PR adding this. I'm genuinely curious to see what you had in mind, and suspect it will be a valuable improvement.

PR #1591, opened 2020-01-25
Add `weightedGraph.contractNodes`

decentralion:
This is needed so we can actually use `WeightedGraph`s as a return type
from plugins. I wrote a very simple implementation that doesn't try to
contract weights together. The docstring explains why this is reasonable
(at least for now).

### DIFF
--- a/src/core/weightedGraph.js
+++ b/src/core/weightedGraph.js
@@ -1,0 +1,49 @@
+// @flow
+
+import {Graph, type GraphJSON} from "./graph";
+import {type Weights as WeightsT, type WeightsJSON} from "./weights";
+import * as Weights from "./weights";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
+
+export type WeightedGraph = {|
+  +graph: Graph,
+  +weights: WeightsT,
+|};
+
+const COMPAT_INFO = {type: "sourcecred/weightedGraph", version: "0.1.0"};
+
+export type WeightedGraphJSON = Compatible<{|
+  +graphJSON: GraphJSON,
+  +weightsJSON: WeightsJSON,
+|}>;
+
+/**
+ * Create a new, empty WeightedGraph.
+ */
+export function empty(): WeightedGraph {
+  return {graph: new Graph(), weights: Weights.empty()};
+}
+
+export function toJSON(w: WeightedGraph): WeightedGraphJSON {
+  const graphJSON = w.graph.toJSON();
+  const weightsJSON = Weights.toJSON(w.weights);
+  return toCompat(COMPAT_INFO, {graphJSON, weightsJSON});
+}
+
+export function fromJSON(j: WeightedGraphJSON): WeightedGraph {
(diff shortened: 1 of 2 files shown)


src/core/weightedGraph.js
```diff
@@ -47,3 +47,27 @@ export function merge(ws: $ReadOnlyArray<WeightedGraph>): WeightedGraph {
...
+ *
+ * This is a wrapper over `Graph.contractNodes`. No effort is made to contract
+ * the weights; thus, it is entirely possible that nodes will "lose" their
+ * associated weight during a contraction.
+ *
+ * These semantics are currently acceptable becuase `contractNodes` is used by
+ * the identity plugin to collapse user identities together, and we currently
+ * always give users a weight of 0 (as they do not mint cred). If we want to
+ * start giving users specific weights, or if we begin using contractNodes for
+ * other types of nodes in the graph, we should revisit this assumption.
```
Beanow: I believe this assumption doesn't hold up.

Currently this assumption is about a convention in the weights file we use.
In fact overriding the defaults:
https://github.com/sourcecred/sourcecred/blob/325487a3ea569e1a0d176b60f39245fb2243804d/src/plugins/github/declaration.js#L58-L64
https://github.com/sourcecred/sourcecred/blob/325487a3ea569e1a0d176b60f39245fb2243804d/src/plugins/discourse/declaration.js#L28-L34

Part of the `WeightedGraph` goal is to allow plugins to set weights. So as far as the core is concerned these weights should be treated as something we don't know the use-cases for. Each plugin will be free to come up with reasons why they _would_ use weights, maybe even dynamic weights, for users or other nodes.

Definitely believe weight contraction should be supported.

decentralion: Hmm, you're right that we can technically give users a non-zero weight, although that weight never has any effect (try changing the weight for user nodes in a cred instance, recompute, and see if scores change). From an implementation standpoint, the reason is that users always have a null timestamp, and nodes with a null timestamp never mint cred. But you're right that this is an un(der)documented detail of the algorithm that we shouldn't depend on.

However, I think this behavior is actually quite principled: minting cred to users doesn't make sense. Giving all users cred by default just encourages Sibyll attacks and mass account creation; having more user accounts is not valuable for the project in itself. Giving individual users direct cred minting violates the principle of neutrality, and shifts the project towards arguing about which users should directly have cred by fiat, rather than having users earn cred based on the value of their contributions. So one approach would be to document that user types always have a weight of 0.

Also, I don't see how to do weight contraction in a principled way. Suppose that you are contracting node A (of type AType) and node B (of type BType) into node C (of type CType). Suppose that A, AType, B, BType, C, and CType all have explicit weights. How would you contract them?

decentralion:
> As you may expect, I am not eager to add this functionality to the
weighted graph module. Will this only actually be used by the identity
plugin? If so, could it live as a helper there instead of tarnishing the
core?

This is a very good point. I believe we can do this, by having the identity plugin explicitly expose a `transformWeightedGraph` method which does this under the hood. I will rework accordingly.

decentralion:
Replaced by #1594.

PR #1592, opened 2020-01-25
Remove the legacy SourceCred UI

decentralion:
This commit removes the legacy (non-timeline) SourceCred UI. While it
does have advantages for cred inspection (it gives an insight into the
edge structure), it gives totally different scores from mainline
SourceCred, which is a potential source of confusion. AFAIK, we haven't
been using it, and keeping all this legacy code around creates
friction/refactoring costs. I'd rather just get rid of it, and
acknowledge that we need to invest more into cred introspection tools in
the future.

Test plan: `yarn test --full` passes; inspection of the UI (via `yarn
start`) shows that things are working as intended, and the legacy cred
link is gone.


Beanow:
> AFAIK, we haven't
> been using it, and keeping all this legacy code around creates
> friction/refactoring costs. I'd rather just get rid of it, and
> acknowledge that we need to invest more into cred introspection tools in
> the future.

I *do* use it extensively because of the easier edge structure insights.
My preference would be to replace both timeline and legacy explorer into a new UI in line with https://discourse.sourcecred.io/t/sourcecred-beta-roadmap/422/2

decentralion:
Got it. Based on the feedback that you use it extensively, I'm fine to keep it around until we are ready to replace them both.

PR #1593, opened 2020-01-25
Stop setting weights on user node types

decentralion:
As discussed in [this GitHub comment][1], it doesn't make sense for user
node types (or user nodes) to have non-zero weight. The reason is that
we use weights for minting cred. Minting cred to users in general
doesn't make sense (having more user accounts is not intrinsically
valuable to a project) and minting cred to specific users is
inappropriate (it means that users' cred is being determined by their
power to influence the weights, rather than because of the value of
their contributions).

This commit makes two changes:
- It sets the default weight for all user types to 0. This has no
implications for cred, since the user weights were already (implicitly)
discarded because users all have null timestamps.
- It filters user node types from the weight config, so the UI no longer
incorrectly suggests that user node weights can be meaningfully changed.

As a result of the second change, the identity plugin now displays in
the weight change UI but has no node or edge types associated. As a
followon commit, we may want to add a bit of filtering logic to clean
that up.

Test plan:
Setting the default weights to 0 for the user types has no effect on
cred, as can be manually ascertained by taking an existing cred
instance, changing the user type weights, and re-calculating.

Filtering the user node types from the WeightConfig is validated through
manual inspection testing.

I've found that frontend unit testing of changes like this has limited
value; since there aren't subtle edge cases to validate, and regressions
are unlikely, I don't think we need a unit test at this time. Therefore,
I haven't added formal tests.

[1]: https://github.com/sourcecred/sourcecred/pull/1591#discussion_r370951707


src/plugins/github/declaration.js
```diff
@@ -59,7 +59,7 @@ export const userNodeType = deepFreeze({
   name: "User",
   pluralName: "Users",
   prefix: N.Prefix.user,
-  defaultWeight: 1,
+  defaultWeight: 0,
   description: "NodeType for a GitHub user",
```
Beanow: Seems like we should apply this for GitHub bots as well.

decentralion: Done, thanks.

PR #1594, opened 2020-01-26
Add `identity/contractIdentities`

decentralion:
This commit adds a `contractIdentities` method to the Identity plugin,
which allows contracting a WeightedGraph using the provided identities.
The method does not attempt to contract weights together, although as a
safety check it will error if weights have been explicitly provided for
any of the contracted nodes.

This PR replaces #1591; see that pull for some context on why this
method is defined on the identity plugin rather than as part of the
WeightedGraph module.

Test plan: For ease of testing, `contractIdentities` is a thin wrapper
around `nodeContractions` (which is already tested) and a new private
`_contractWeightedGraph` method (for which tests have been added). Since
`contractIdentities` is a trivial oneline composition, it does not need
any additional explicit testing.

`yarn test` passes.

### DIFF
--- a/src/plugins/identity/contractIdentities.js
+++ b/src/plugins/identity/contractIdentities.js
@@ -1,0 +1,67 @@
+// @flow
+
+import * as Weights from "../../core/weights";
+import {type WeightedGraph as WeightedGraphT} from "../../core/weightedGraph";
+import {type NodeContraction, NodeAddress} from "../../core/graph";
+import {nodeContractions} from "./nodeContractions";
+import {type Identity} from "./identity";
+
+/**
+ * Applies nodeContractions to a WeightedGraph.
+ *
+ * This functionality is defined as part of the identity plugin rather than as
+ * a feature of WeightedGraph because it doesn't attempt to contract weights
+ * (it's not clear how to do this in a general principled way). Within the
+ * identity plugin, we do not expect user identities to have weights
+ * associated, so we can contract the graph without attempting to contract the
+ * weights.
+ *
+ * As a safety measure, this method will error if any of the node addresses
+ * being contracted has an explicitly set weight. It will not error if there
+ * are matching type weights, so that it is still possible (e.g.) to apply
+ * a weighting to an entire plugin.
+ *
+ * For more context on this decision, see discussion in #1591.
+ */
+export function _contractWeightedGraph(
+  wg: WeightedGraphT,
+  contractions: $ReadOnlyArray<NodeContraction>
+): WeightedGraphT {
+  const {graph, weights} = wg;
+  for (const {old} of contractions) {
+    for (const address of old) {
+      const weight = weights.nodeWeights.get(address) || 0;
+      if (weight !== 0) {
+        throw new Error(
+          `Explicit weight ${weight} on contracted node ${NodeAddress.toString(
+            address
+          )}`
+        );
+      }
+    }
+  }
+  return {
+    graph: graph.contractNodes(contractions),
+    weights: Weights.copy(weights),
+  };
+}
+
+/**
+ * Given a WeightedGraph and identity information, produce a contracted
+ * WeightedGraph where all of an identitiy's aliases have been contracted into
+ * a unified identity.
+ *
+ * An error will be thrown if any of the aliases had an explicitly set weight,
+ * since we don't currently support weight contraction.
+ *
+ * Note: This function has no unit tests, because it is a trivial composition
+ * of two tested functions. Thus, flow typechecking in sufficient. If adding
+ * any complexity to this function, please also add tests.
+ */
+export function contractIdentities(
+  wg: WeightedGraphT,
+  identities: $ReadOnlyArray<Identity>,
+  discourseUrl: string | null
+): WeightedGraphT {
+  return _contractWeightedGraph(wg, nodeContractions(identities, discourseUrl));
+}
--- a/src/plugins/identity/contractIdentities.test.js
+++ b/src/plugins/identity/contractIdentities.test.js
@@ -1,0 +1,45 @@
+// @flow
+
+import {node} from "../../core/graphTestUtil";
+import {Graph, NodeAddress} from "../../core/graph";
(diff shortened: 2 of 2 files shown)


Beanow:
> The method does not attempt to contract weights together

How should plugins set weights for identities?
Or more likely, an edge weight to a user before contracting.

decentralion:
> How should plugins set weights for identities?

No one should set weights for identities; cf #1593.

> Or more likely, an edge weight to a user before contracting.
Edge weights are assigned based on address (not src or dst), and while contraction may rewrite the src or dst of an edge, it does not rewrite the address. Therefore, edge weights are all unaffected by contraction.

Beanow:
Makes sense,

> Edge weights are assigned based on address (not src or dst), and while contraction may rewrite the src or dst of an edge, it does not rewrite the address. Therefore, edge weights are all unaffected by contraction.

That said, _graph_ contraction does perform the rewrite of src/dst addresses.
So I find it surprising if this is not done for edge weights.

https://github.com/sourcecred/sourcecred/blob/3a962af0aff46e18025de7861c458c4b4fa7d8ba/src/core/graph.js#L904-L908

An example scenario would be, Initiatives wants to define the champions edge weight as a percentage. Like "20% of contributions & references". So a dynamic weight would be set towards a GitHub or Discourse user. Contraction would need to rewrite the user address to an identity address.

decentralion:
> So a dynamic weight would be set towards a GitHub or Discourse user. Contraction would need to rewrite the user address to an identity address.

I still don't understand what you're saying. If the dynamic weight were set on the (Initiative -> User) edge, that weight would still apply after contraction, since the weight was set on the edge address (which is unchanged). So no change to the contraction logic is required.

> Initiatives wants to define the champions edge weight as a percentage. Like "20% of contributions & references".

A good use case for [edge tranches](https://discourse.sourcecred.io/t/edge-tranches-fixing-an-incentive-misalignment/418).

Beanow:
Aah, no I was just confused.

The edge src/dst is stored in the graph, so those are updated to the new address. The edge weight doesn't have a src/dst, just the address of the (already updated) graph edge, meaning it won't need updating. :+1:

src/plugins/identity/contractIdentities.test.js
```diff
@@ -0,0 +1,45 @@
...
+    const w = Weights.empty();
+    w.nodeWeights.set(NodeAddress.empty, 3);
+    return w;
+  };
+  describe("_contractWeightedGraph", () => {
+    it("contracts the graph", () => {
+      const wg = {graph: graph(), weights: weights()};
+      const contracted = _contractWeightedGraph(wg, contractions());
+      const expected = graph().contractNodes(contractions());
+      expect(expected.equals(contracted.graph)).toBe(true);
```
Beanow: Nit: prefer a deep equal to see differences if possible, maybe using toJSON for this.

src/plugins/identity/contractIdentities.test.js
```diff
@@ -0,0 +1,45 @@
...
+import {node} from "../../core/graphTestUtil";
+import {Graph, NodeAddress} from "../../core/graph";
+import * as Weights from "../../core/weights";
+import {_contractWeightedGraph} from "./contractIdentities";
+
+describe("plugins/identity/contractIdentities", () => {
+  const a = node("a");
+  const b = node("b");
+  const c = node("c");
+  const graph = () => new Graph().addNode(a).addNode(b);
```
Beanow: Nit: prefer including an edge contraction, though OK with accepting as-is to delegate this to `Graph.contractNodes()` tests.

src/plugins/identity/contractIdentities.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Applies nodeContractions to a WeightedGraph.
+ *
+ * This functionality is defined as part of the identity plugin rather than as
+ * a feature of WeightedGraph because it doesn't attempt to contract weights
+ * (it's not clear how to do this in a general principled way). Within the
+ * identity plugin, we do not expect user identities to have weights
+ * associated, so we can contract the graph without attempting to contract the
+ * weights.
```
Beanow: Side-note: speaking of unexpected things. Maybe we should document a plugin developers warning that creating user-to-user edges between plugins could create self-loops when contracting identities?

src/plugins/identity/contractIdentities.test.js
decentralion: expect.equals is deep equals, vs. expect.toBe which checks value equality.

https://jestjs.io/docs/en/expect#toequalvalue

Beanow: We're using `graph.equals` though, not `expect.toEqual`.
The current pattern is `expect(boolean).toBe(true)`.

decentralion: Oh, true. But graph.equals is defined as a proper equality checker for graphs.

Beanow: It is a solid test.
My nit here is that it would be useful feedback during testing to see *how* it differs.
Probably should have formulated it that way. I would like it if a failed test would present me with a diff.

decentralion: Ah, now I understand, thanks.

Not going to change it. (We use this pattern all over the codebase--if we think it's important to provide more useful diffs when checking graph equality, I'd rather try to do it consistently than in just one place.)

src/plugins/identity/contractIdentities.test.js
decentralion: Yeah, this is just a sanity check that we've delegated to Graph.contractNodes--which has thorough testing of edge contractions.

src/plugins/identity/contractIdentities.js
decentralion: Heh, yeah, add it to the list of stuff we should document whenever we document how to add a new plugin ;)

decentralion:
... oops, I merged this into the wrong branch :/

PR #1595, opened 2020-01-26
Generate Weights from PluginDeclarations

decentralion:
This adds a simple method, `weightsForDeclaration`, which generates
weights from a plugin declaration. This is a small but important piece
of #1557, as it allows us to create appropriate Weights cleanly from the
plugin.


src/analysis/pluginDeclaration.js
```diff
@@ -25,3 +27,15 @@ export function combineTypes(
   const edgeTypes = [].concat(...decs.map((x) => x.edgeTypes));
   return {nodeTypes, edgeTypes};
 }
+
+export function weightsForDeclaration(dec: PluginDeclaration): WeightsT {
```
Beanow: Nit: slight preference for `declaration` over `dec`.

decentralion: Ack(nowledged); not making changes.

PR #1597, opened 2020-01-26
Weights.merge: add support for resolvers

decentralion:
This commit adds support for resolvers to `Weights.merge`. The change is
documented and unit tested. Another step towards #1557.

Test plan: Inspect included tests; `yarn test` passes.

Note: I re-indented some test code, recommend reviewing with 
whitespace diffs toggled off.

### DIFF
--- a/src/core/weights.js
+++ b/src/core/weights.js
@@ -1,7 +1,12 @@
 // @flow
 
 import * as MapUtil from "../util/map";
-import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
+import {
+  type NodeAddressT,
+  type EdgeAddressT,
+  NodeAddress,
+  EdgeAddress,
+} from "../core/graph";
 import {toCompat, fromCompat, type Compatible} from "../util/compat";
 
 /**
@@ -11,6 +16,8 @@ import {toCompat, fromCompat, type Compatible} from "../util/compat";
  */
 export type NodeWeight = number;
 
+export type NodeOperator = (NodeWeight, NodeWeight) => NodeWeight;
+
 /**
  * Represents the forwards and backwards weights for a particular Edge (or
  * edge address prefix).
@@ -18,6 +25,8 @@ export type NodeWeight = number;
  * Weights are linear, so 2 is twice as important as 1.
  */
 export type EdgeWeight = {|+forwards: number, +backwards: number|};
+
+export type EdgeOperator = (EdgeWeight, EdgeWeight) => EdgeWeight;
 
 /**
  * Represents the weights for nodes and edges.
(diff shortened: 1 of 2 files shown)


src/core/weights.js
```diff
@@ -48,19 +52,58 @@ export function copy(w: Weights): Weights {
...
+      );
+    };
+  }
+  for (const w of ws) {
+    for (const [addr, val] of w.nodeWeights.entries()) {
+      const existing = nodeWeights.get(addr);
+      if (existing == null) {
+        nodeWeights.set(addr, val);
+      } else {
+        nodeWeights.set(addr, nodeResolver(existing, val));
```
Beanow: Suggest a try-catch around the resolver, to add the `NodeAddress` to the error message.
In case of a bug where we're expecting disjoint weights, this seems like important debug info :]

src/core/weights.js
```diff
@@ -48,19 +52,58 @@ export function copy(w: Weights): Weights {
...
- * input weights. If there are any overlaps (i.e. the same address is present
- * in two or more of the input weights), an error will be thrown. In the future,
- * we will likely modify this function to add a resolver that determines how to
- * combine multiple overlapping weights.
+ * When there are overlaps (i.e. the same address is present in two or more
+ * input weights), the appropriate resolver will be invoked to resolve the
+ * conflict. The resolver will receive, as its first argument, the result of
+ * resolving the last conflict (or the first weight if there have been no
+ * conflicts), and its second argument will be the latest conflicting weight.
+ * Whatever it returns will be used as the weight (modulo future conflicts).
```
Beanow: The process is very similar to a [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) function.
Perhaps the terms make sense to use as well. Accumulator and current value.

Adding these semantics would be nice as comments on the NodeOperator and EdgeOperator types as well.

src/core/weights.js
decentralion: Good idea. I updated it, and checked the behavior with some new tests.

wchargin:
(Apologies for delay—I plan to review this tonight.)

src/core/weights.js
decentralion: Sure, I can update the docs to use this vocabulary; PTAL.

Regarding adding comments directly to the types-- the merge function uses them as reducers, but the types themselves could be used for other purposes too. So I don't think we should add comments directly at the type level.

decentralion:
Sounds good @wchargin . I've updated the PR to reflect @Beanow's suggestions.

src/core/weights.js
wchargin: I think that “reducers” in the JavaScript sense of the word give the
wrong impression, precisely because “accumulator and current value” is
the wrong formulation. The reduction operator should be associative, and
should be able to combine values in any order: `f(f(x, y), z)` should be
the same as `f(x, f(y, z))`. (Note that I’m not saying that it needs to
be commutative.)

(I recommend watching [“Four Solutions to a Trivial Problem”][4s] if you
haven’t seen it, but parallelism isn’t the only goal. Reductions in the
OCaml sense are far simpler to think about than “folds”.)

[4s]: https://www.youtube.com/watch?v=ftcIcn8AmSY

wchargin:
(quick comment now since I see that changes are being made)

src/core/weights.js
```diff
@@ -48,19 +57,72 @@ export function copy(w: Weights): Weights {
...
+ * The resultant Weights will have every weight specified by each of the input
+ * weights.
+ *
+ * When there are overlaps (i.e. the same address is present in two or more of
+ * the Weights), then the appropriate resolver will be invoked to resolve the
+ * conflict. The semantics are similar to Array.reduce, where the accumulated
+ * weight and the current weight are both passed as arguments to the resolver.
+ * For the first conflict, the "accumulated weight" is equal to the first
+ * weight matching that address. The final accumulated weight will be included
+ * in the resultant Weights.
```
wchargin: As [mentioned earlier][1] (and also in personal communication before
this PR was sent), I think that left-to-right accumulator semantics are
confusing here. By saying that “the accumulated weight and the current
weight are both passed as arguments to the resolver”, the client must
consider two distinct concepts: a notion of _time_ (“current” weight)
and a notion of _state_ (“accumulated” weight). This is inessential.

It would be simpler to say:

> The reducer takes two weights and combines them to return a new
> weight.

Then, we don’t have to think about whether we’re folding left- or
right-associatively, and don’t have to think about any notion of
currency.

[1]: https://github.com/sourcecred/sourcecred/pull/1597#discussion_r372164566

src/core/weights.js
```diff
@@ -48,19 +57,72 @@ export function copy(w: Weights): Weights {
...
+ * conservative "error on conflict" resolvers.
  */
-export function merge(ws: $ReadOnlyArray<Weights>): Weights {
-  const nodeWeights = MapUtil.merge(ws.map((x) => x.nodeWeights));
-  const edgeWeights = MapUtil.merge(ws.map((x) => x.edgeWeights));
-  return {nodeWeights, edgeWeights};
+export function merge(
+  ws: $ReadOnlyArray<Weights>,
+  nodeResolver: ?NodeOperator,
+  edgeResolver: ?EdgeOperator
```
wchargin: Optional: Consider passing these as an optional object with optional
keys, so that the parameters can be named? `merge(ws, null, null)` is a
tad awkward.

src/core/weights.js
Beanow: Interesting talk, thanks :smile: 
I did notice this awkwardness with javascript's "reduce function" signature. Given JS's history, can picture why going for a fold instead of reduce makes sense.

Anyway, I didn't consider suggesting an associative function. The docstring explicitly described an accumulator and current value, just in different terms. Agreed that in this case, we may not need these concepts and could require the function to be associative instead.

src/core/weights.js
Beanow: Agree :] and I would ask this a bit more strongly than optional.
Would argue not doing so adds a bit of API technical debt.

Over time a practice of multiple optional positional arguments makes the invocations brittle and really hard to glean the function signature. Named parameters help out a lot there.

`createReverseIndex(articles, null, 3);`
:thinking: what do you mean 3? How do I know this 3 shouldn't be in the 2nd position? _(Answer: need to look up the definition)_.

`createReverseIndex(articles, {minTokenLength: 3});`
Ah!

src/core/weights.js
Beanow: Marking this as resolved as the terms are more clear now.
Will defer to @wchargin whether it should be associative instead.

src/core/weights.js
decentralion: Done.

src/core/weights.js
decentralion: Done.

src/core/weights.js
wchargin: @Beanow: Yes; `JSON.stringify(obj, null, 4)` is a commonplace example.

src/core/weights.js
```diff
@@ -48,19 +57,68 @@ export function copy(w: Weights): Weights {
...
+ * When no resolvers are explicitly provided, merge defaults to
+ * conservative "error on conflict" resolvers.
  */
-export function merge(ws: $ReadOnlyArray<Weights>): Weights {
-  const nodeWeights = MapUtil.merge(ws.map((x) => x.nodeWeights));
-  const edgeWeights = MapUtil.merge(ws.map((x) => x.edgeWeights));
-  return {nodeWeights, edgeWeights};
+export function merge(
+  ws: $ReadOnlyArray<Weights>,
+  resolvers: ?{|+nodeResolver: NodeOperator, +edgeResolver: EdgeOperator|}
```
wchargin: This is fine; it’d also be fine to allow explicitly setting a node
resolver but not an edge resolver (or vice versa), which imho is a bit
more natural.

Beanow: The pattern that we use in some places that I think would make sense here:

Have a type where the options are all required (such as `TimelineCredParameters`).
And take optional arguments as `key?: $Shape<T>`.

Because it allows a Flow check to know we have a default value for everything.

```js
(overrides?: $Shape<T>) => {
  const defaults: T = {/* ... */};
  const guaranteedCompleteOptions: T = {...defaults, ...overrides};
};
```

PR #1601, opened 2020-01-29
Add `identity/contractIdentities`

decentralion:
This commit adds a `contractIdentities` method to the Identity plugin,
which allows contracting a WeightedGraph using the provided identities.
The method does not attempt to contract weights together, although as a
safety check it will error if weights have been explicitly provided for
any of the contracted nodes.

This PR replaces #1591; see that pull for some context on why this
method is defined on the identity plugin rather than as part of the
WeightedGraph module.

Test plan: For ease of testing, `contractIdentities` is a thin wrapper
around `nodeContractions` (which is already tested) and a new private
`_contractWeightedGraph` method (for which tests have been added). Since
`contractIdentities` is a trivial oneline composition, it does not need
any additional explicit testing.

`yarn test` passes.

### DIFF
--- a/src/plugins/identity/contractIdentities.js
+++ b/src/plugins/identity/contractIdentities.js
@@ -1,0 +1,67 @@
+// @flow
+
+import * as Weights from "../../core/weights";
+import {type WeightedGraph as WeightedGraphT} from "../../core/weightedGraph";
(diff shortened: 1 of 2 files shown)


decentralion:
Merging sans review because this is a recreation of #1594, which I merged into the wrong branch.

PR #1605, opened 2020-01-30
Convert load pipeline to pass WeightedGraphs

decentralion:
This commit changes `api/load` and downstream consumers to use
WeightedGraphs instead of regular Graphs. In addition to `api/load`, we
also modify the frontends and the timeline cred calculation module.

However, we don't yet _use_ the weights from the WeightedGraph. So as to
make this commit easier to review, it only changes the data type being
passed around; however in practice the consumers ignore the weights and
simply use the underlying graph. A followon commit will modify the
consumers so that they properly retrieve weights from within the
WeightedGraph.

This is a major step towards #1557.

Test plan:

`yarn test --full` passes; manual testing verifies that the frontend
still displays cred properly, and that modifying the weights and
re-calculating shows that the weights are being used properly.


src/core/weightedGraph.js
```diff
@@ -48,7 +48,8 @@ export function fromJSON(j: WeightedGraphJSON): WeightedGraph {
  * This delegates to the semantics of Graph.merge and Weights.merge.
  */
 export function merge(ws: $ReadOnlyArray<WeightedGraph>): WeightedGraph {
-  const graph = Graph.merge(ws.map((w) => w.graph));
+  const graphs = ws.map((w) => w.graph);
+  const graph = Graph.merge(graphs);
```
Beanow: Nit: looks like a functionally equivalent, unrelated change

decentralion: Good catch -- added it while debugging something. Will revert.

PR #1606, opened 2020-01-30
Add WeightedGraph.overrideWeights

decentralion:
In a few occasions in the codebase, we need the ability to take a
WeightedGraph and apply manual user overrides to its weights (keeping
the base weights wherever non-conflicting). It's actually a fairly
simple application of Weights.merge, but since it's of general utility
I'm adding it to the WeightedGraph API.

Test plan: I've added unit tests that validate its behavior; take a
look. `yarn test` passes.

### DIFF
--- a/src/core/weightedGraph.test.js
+++ b/src/core/weightedGraph.test.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as Weights from "./weights";
-import {Graph} from "./graph";
+import {Graph, NodeAddress, EdgeAddress} from "./graph";
 import * as WeightedGraph from "./weightedGraph";
 import * as GraphTest from "./graphTestUtil";
 
@@ -10,6 +10,9 @@ describe("core/weightedGraph", () => {
     expect(wg1.graph.equals(wg2.graph)).toBe(true);
     expect(wg1.weights).toEqual(wg2.weights);
   }
+  const foo = GraphTest.node("foo");
+  const bar = GraphTest.node("bar");
+  const foobar = GraphTest.edge("foobar", foo, bar);
 
   describe("empty", () => {
     it("empty produces an empty WeightedGraph", () => {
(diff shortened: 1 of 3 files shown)


src/core/weightedGraph.test.js
```diff
@@ -60,4 +61,37 @@ describe("core/weightedGraph", () => {
...
     });
   });
+
+  describe("overrideWeights", () => {
+    const example = () => {
+      const graph = new Graph().addNode(foo).addNode(bar);
+      const weights = Weights.empty();
+      weights.nodeWeights.set(foo.address, 1);
+      weights.edgeWeights.set(foobar.address, {forwards: 2, backwards: 2});
+      weights.edgeWeights.set(EdgeAddress.empty, {forwards: 2, backwards: 2});
```
Beanow: Nit: think it would be better to have all the weight numbers used be distinct.
To make sure that they didn't for whatever reason get mixed up with another node/edge.

src/core/weightedGraph.test.js
```diff
@@ -60,4 +61,37 @@ describe("core/weightedGraph", () => {
...
+      expectEqual(g1, g2);
+    });
+    it("takes weights from base and overrides, choosing overrides on conflicts", () => {
+      const overrides = Weights.empty();
+      overrides.nodeWeights.set(foo.address, 100);
+      overrides.nodeWeights.set(bar.address, 100);
+      overrides.edgeWeights.set(foobar.address, {
+        forwards: 100,
+        backwards: 100,
+      });
```
Beanow: Nit: same here, think it would be better to have all the weight numbers used be distinct.

src/core/weightedGraph.test.js
decentralion: Done.

src/core/weightedGraph.test.js
decentralion: Done.

PR #1607, opened 2020-01-30
Move weights out of TimelineCredParams

decentralion:
This commit moves weights out of the "parameters" to TimelineCred. This
makes sense, because the Weights are now passed to TimelineCred via the
included WeightedGraph. As such, we now have the `api/load` options
include explicit Weights that are used as overrides, rather than having
them be included in the TimelineCred parameters.

Test plan: I've manually tested this commit by:
- Changing weights in the explorer, and verifying that the `recalculate
cred` button activates as expected, and the new weights are used
correctly in the resultant distribution.
- Verifying that downloading weights form the UI still works.
- Verified that uploading weights to the UI still works.
- Verifying that passing command-line weights files still works.

Also, `yarn test` passes.


src/cli/discourse.js
```diff
@@ -97,7 +97,8 @@ const command: Command = async (args, std) => {
   await load(
     {
       project,
-      params: {weights},
+      params: null,
```
Beanow: Side-note: I feel like this is an example of why
`key?: $Shape<T>`
is better than the current
`key: ?$Shape<T>`
for an Options type.

But let's not change that in this PR.

src/cli/load.test.js
```diff
@@ -76,7 +75,8 @@ describe("cli/load", () => {
           id: "foo/bar",
           repoIds: [makeRepoId("foo", "bar")],
         }),
-        params: defaultParams(),
+        params: null,
```
Beanow: This changes the test more than is needed, and should probably keep the `defaultParams()`.

Because `defaultParams()` sets a value for all options (unlike `partialParams()`), setting it to `null` now removes the alpha and intervalDecay options.

src/cli/load.test.js
```diff
@@ -99,7 +99,8 @@ describe("cli/load", () => {
           id: projectId,
           repoIds: [stringToRepoId(projectId)],
         }),
-        params: defaultParams(),
+        params: null,
```
Beanow: Same comment here, better keep `defaultParams()`.

src/explorer/TimelineExplorer.js
```diff
@@ -41,31 +42,38 @@ export class TimelineExplorer extends React.Component<Props, State> {
...
-    const {alpha, intervalDecay, weights} = this.state;
+    const {alpha, intervalDecay} = this.state;
+    return {alpha, intervalDecay};
+  }
+
+  weights(): WeightsT {
     // Set the weights to a copy, to ensure that the weights we pass into e.g.
     // analyzeCred are a distinct reference from the one we keep in our state.
-    return {alpha, intervalDecay, weights: weightsCopy(weights)};
+    return Weights.copy(this.state.weights);
```
Beanow: Side-note: noticing we're probably copying the weights more than strictly needed.
But since that was the existing behavior, no change needed for this PR.

decentralion: I suspect we aren't. But happy to not change it. :)

src/cli/discourse.js
decentralion: Ack.

src/cli/load.test.js
decentralion: Done.

src/cli/load.test.js
decentralion: Done.

PR #1609, opened 2020-02-02
mirror: remove unnecessary shadowing

wchargin:
Summary:
A common table expression shadows a (main or temporary) table name for
`SELECT` statements, but doing so is confusing and makes the code harder
to read.

Test Plan:
Existing unit tests suffice.

wchargin-branch: mirror-no-cte-shadow

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -1649,7 +1649,7 @@ export class Mirror {
         db.prepare(
           dedent`\
             WITH RECURSIVE
-            direct_dependencies (parent_id, child_id) AS (
+            cte_direct_deps (parent_id, child_id) AS (
                 SELECT parent_id, child_id FROM links
                 WHERE child_id IS NOT NULL
                 UNION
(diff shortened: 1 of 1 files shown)


wchargin:
FWIW, I’m especially inclined to patch this because I wasn’t able to
find an official source on whether it’s well defined behavior. I asked
on the sqlite-users list, but haven’t heard back yet:
<http://sqlite.1065341.n5.nabble.com/Shadowing-a-table-name-with-a-common-table-expression-td110832.html>

Beanow:
Needed to look a bit closer for what shadowing you meant.

https://github.com/sourcecred/sourcecred/blob/5de027e83d085a956a94fb404efad97d006299d7/src/graphql/mirror.js#L1639-L1646
Is created for that same `extract` method and used as the target table for a `INSERT INTO ... FROM` query. Meanwhile we're using the same name of `transitive_dependencies` as a [CTE](https://www.sqlite.org/lang_with.html) created from the `WITH` statement.

https://github.com/sourcecred/sourcecred/blob/5de027e83d085a956a94fb404efad97d006299d7/src/graphql/mirror.js#L1663-L1668

wchargin:
Yes, that is correct.

PR #1610, opened 2020-02-02
mirror: add default typename-mismatch error message

wchargin:
Summary:
Previously, `_nontransactionallyRegisterObject` differed from its
counterpart `registerObject` in two ways: the former does not enter a
transaction, but it also requires a function to format an error message.
The value provided by `registerObject` to the helper is a suitable
default, so we can use it to decouple the transaction semantics from the
error message.

Test Plan:
Existing tests suffice, retaining full coverage.

wchargin-branch: mirror-default-typename-message

### DIFF
--- a/src/graphql/mirror.js
+++ b/src/graphql/mirror.js
@@ -365,14 +365,7 @@ export class Mirror {
     +id: Schema.ObjectId,
   |}): void {
     _inTransaction(this._db, () => {
-      this._nontransactionallyRegisterObject(object, (guess) => {
-        const s = JSON.stringify;
-        const message =
-          `object ${s(object.id)} ` +
-          `looks like it should have type ${s(guess)}, ` +
-          `not ${s(object.typename)}`;
-        return message;
-      });
+      this._nontransactionallyRegisterObject(object);
     });
   }
 
@@ -386,10 +379,21 @@ export class Mirror {
       +typename: Schema.Typename,
       +id: Schema.ObjectId,
     |},
-    guessMismatchMessage: (guess: Schema.Typename) => string
+    guessMismatchMessage?: (guess: Schema.Typename) => string
   ): void {
     const db = this._db;
     const {typename, id} = object;
+
+    if (guessMismatchMessage == null) {
+      guessMismatchMessage = (guess) => {
+        const s = JSON.stringify;
+        const message =
+          `object ${s(object.id)} ` +
+          `looks like it should have type ${s(guess)}, ` +
+          `not ${s(object.typename)}`;
+        return message;
+      };
+    }
 
     const existingTypename = db
       .prepare("SELECT typename FROM objects WHERE id = ?")


Beanow:
The API here is a little curious.
The only other usage of this message builder function seems to be:
https://github.com/sourcecred/sourcecred/blob/d825c9a1a2629df2e3a525f95dc390a9447510f0/src/graphql/mirror.js#L494-L497

Which takes a `context`, we're using to prefix the message in a: "when doing ___: warning message" style. Otherwise the suffix is near identical. Would it make sense to have the same `context` function / string here and not allow overriding the message generation?

wchargin:
Sure, that could also make sense, but I don’t see that it’s meaningfully
better than this in any way.

As to “only other usage”, I’m making this change because I’m shortly
going to add another usage, and this change makes that more convenient.

Beanow:
> Sure, that could also make sense, but I don’t see that it’s meaningfully
> better than this in any way.

Just minimizing to the functionality actually required and to prevent re-implementing the suffix part.
`prefix: string` is much simpler than `messageBuilder: (T) => string`

That said. Just a comment. If you plan to use it, not going to block on this :]

wchargin:
The “suffix parts” are different, though; one is for a server response
and one is for an explicitly registered object.

A fixed string is structurally “smaller” than a supplier, but it’s not
simpler, because the fixed string is at the wrong level of abstraction.

> That said. Just a comment. If you plan to use it, not going to block
> on this :]

Right, I’m not super interested in debating the minutiae of this private
helper function.

Beanow:
Private or not, this PR changed the signature, so I included a review of the overall signature and this stood out to me. I didn't nit-and-approve because I thought a reply could clear up what I noticed. Would you recommend a different review approach?

wchargin:
Reviewing the signature is perfectly fine, and asking for clarification
rather than nitting and approving is perfectly fine (I certainly don’t
mean to lower the bar to “approve immediately unless the change is
obviously broken”!). I do think that the comment came across as “not how
I would have done it”.

> Before doing this, you should ask: in what ways is doing it my way
> materially better? What benefits come from the changed approach? If
> you don't have clear answers to these questions, don't ask for the
> change.

I thought about the suggestion for a few minutes and wasn’t able to
identify answers to these questions.

The bar for this is especially high when the PR author hasn’t requested
review from the person offering their opinion. It’s really easy for such
drive-by review comments to pile on, and so it’s especially important
that they be clearly beneficial. Time and mental energy spent reading,
considering, and implementing and/or responding to these comments is
time and mental not spent on primary work. For instance, I’ve spent
about half an hour on this review so far across three rounds of comments
and therefore three context switches. This is totally worth it when the
discussion improves the code or helps someone learn something, so we
just need to make sure that we’re hitting that goal.

I appreciate your openness to feedback. :-)

Beanow:
Thank you for the feedback :]

I had "not how I would have done it" in mind for why I wouldn't block. And regularly review without request for decentralization reasons. But I'll dial down the meticulousness in the future. The time vs benefits, esp for unrequested reviews is great feedback! I've gotten a better sense of "materially better" in this context too.

wchargin:
No problem—this has been something that took me a long time to
internalize, too; I’m still getting better at it. :-) thanks!

PR #1614, opened 2020-02-02
Github: switch to CacheProvider for fetchGithubRepo

Beanow:
Depends on #1613 
Replaces #1568

Delegating to a `CacheProvider` instance, will limit the number of places where we need to handle filesystem details. It will also allow a mock, or in-memory cache to be provided.

Test plan:
1. `yarn test --full`
Existing flow/eslint coverage should include ` src/plugins/github/bin/*` for correct api usage.
2. Optional: `scripts/update_snapshots.sh` with a `SOURCECRED_GITHUB_TOKEN`.
The updated snapshots should be unchanged.

### DIFF
--- a/src/api/load.test.js
+++ b/src/api/load.test.js
@@ -26,6 +26,7 @@ import {
   partialParams,
 } from "../analysis/timeline/params";
 import * as WeightedGraph from "../core/weightedGraph";
+import {DataDirectory} from "../backend/dataDirectory";
 
 type JestMockFn = $Call<typeof jest.fn>;
 jest.mock("../plugins/github/loadWeightedGraph", () => ({
@@ -102,11 +103,11 @@ describe("api/load", () => {
   it("calls github githubWeightedGraph with the right options", async () => {
     const {options, taskReporter, sourcecredDirectory} = example();
     await load(options, taskReporter);
-    const cacheDirectory = path.join(sourcecredDirectory, "cache");
+    const cache = new DataDirectory(sourcecredDirectory);
     const expectedLoadGraphOptions: LoadGraphOptions = {
       repoIds: project.repoIds,
       token: exampleGithubToken,
-      cacheDirectory,
+      cache,
     };
     expect(githubWeightedGraph).toHaveBeenCalledWith(
       expectedLoadGraphOptions,
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -6,7 +6,6 @@
 
 import Database from "better-sqlite3";
 import fetch from "isomorphic-fetch";
-import path from "path";
 import retry from "retry";
 
 import {type RepoId, repoIdToString} from "./repoId";
@@ -19,6 +18,7 @@ import type {Repository} from "./graphqlTypes";
 import schema from "./schema";
 import {validateToken} from "./token";
 import {cacheIdForRepoId} from "./cacheId";
+import {type CacheProvider} from "../../backend/cache";
 
 /**
  * Scrape data from a GitHub repo using the GitHub API.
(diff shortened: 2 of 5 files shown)


src/plugins/github/loadGraph.js
```diff
@@ -15,11 +15,12 @@ import {RelationalView} from "./relationalView";
 import {type RepoId, repoIdToString} from "./repoId";
 import {Graph} from "../../core/graph";
 import {type GithubToken} from "./token";
+import {type CacheProvider} from "../../backend/cache";
 
 export type Options = {|
   +repoIds: $ReadOnlyArray<RepoId>,
   +token: GithubToken,
-  +cacheDirectory: string,
+  +cache: CacheProvider,
```
decentralion: could we call this argument `cacheProvider` rather than just `cache`? When reading the code in `api/load`, i found it confusing that we passed `dataDirectory` as `cache`-- passing it as `cacheProvider` would be less confusing.

src/api/load.test.js
```diff
@@ -102,11 +103,11 @@ describe("api/load", () => {
   it("calls github githubWeightedGraph with the right options", async () => {
     const {options, taskReporter, sourcecredDirectory} = example();
     await load(options, taskReporter);
-    const cacheDirectory = path.join(sourcecredDirectory, "cache");
+    const cache = new DataDirectory(sourcecredDirectory);
```
decentralion: I'm a little surprised this test still works -- in the past the cacheDirectory was a string (value equality) but now we are passing in an object, and the expected DataDirectory will be a different instance than the actual one.

docs are unclear... but I guess this test passing means that `toHaveBeenCalledWith` uses deep equality checking?

src/plugins/github/fetchGithubRepo.js
```diff
@@ -35,9 +35,9 @@ import {cacheIdForRepoId} from "./cacheId";
  */
 export default async function fetchGithubRepo(
   repoId: RepoId,
-  options: {|+token: string, +cacheDirectory: string|}
+  options: {|+token: string, +cache: CacheProvider|}
```
decentralion: as elsewhere, i'd prefer that `cache` be called `cacheProvider`

src/api/load.test.js
Beanow: Yes. It needs to have used deep equality to have worked before this change as well.
`expectedLoadGraphOptions` would not be object equal to the actual object passed.

But indeed the docs are not really clear about it. You can also use any of the matchers with it.
For example, from [`expect.any(constructor)`](https://jestjs.io/docs/en/expect#expectanyconstructor):
> You can use it inside toEqual or toBeCalledWith instead of a literal value.
```js
expect(mock).toBeCalledWith(expect.any(Number));
```

src/plugins/github/loadGraph.js
Beanow: I'm using cache as the variable name consistently.
Reason is, I think -Provider seems superfluous in variable names.
```js
// Looks interchangeable
cache.database(...);
cacheProvider.database(...);
```

> i found it confusing that we passed dataDirectory as cache

Instead, would it help if I casted dataDirectory to a CacheProvider instead?
```js
{
  cache: (dataDirectory: CacheProvider),
}
```

decentralion: That works too. Since this is a nit and you're the author, I'm fine with either approach.

Beanow: Cast added.

src/plugins/github/fetchGithubRepo.js
Beanow: Using above cast instead.

Beanow:
Rebased as dependency was merged. Included the `CacheProvider` cast from the comment.

Issue #1586, opened 2020-01-21
Load refactor design doc


Beanow:
Putting this down as a tracker. Based on feedback on "LoadPlan", will revise it and propose a doc first.

From feedback on "LoadPlan" the term Plan and the API's weren't clear enough.
So I'll iterate on the concept and add a document for this first.

Beanow:
Taken another stab at this and here's the preview of what I came up with:
https://github.com/sourcecred/sourcecred/tree/wip/load-context/src/backend

## Goals

- Divide and conquer the complexity of load (#1559).
- Allow more robust mocking in tests (#1559).
- Avoid hiding function arguments (#1565 feedback).
- Communitate through the signature of individual steps how we're expected to compose them (#1565 feedback).

## At a glance

* [LoadContext]
  - [PluginLoaders]
    * [GithubLoader]
      - declaration
      - updateMirror
      - createGraph
    * [DiscourseLoader]
      - ...
    * [IdentityLoader]
      - ...
  - [ComputeFunction]
    * TimelineCred.compute

### LoadContext

Our spider-in-the-web.
Has a single responsiblity: composition.

> This class is responsible composing all the variables and concrete functions of the loading process.
>
> Specifically it composes:
> - The loading environment (through the constructor).
> - Concrete functions of the loading process (internally).
> - Parameters for a load (Project and TimelineCredParameters).
>
> You can think of LoadContext as an instance where the environment and functions have been composed so it's ready to perform a load with.

It uses a [new trick](https://github.com/sourcecred/sourcecred/blob/wip/load-context/src/backend/loadContext.js#L48-L61) of aliasing external module functions as private properties.
This makes the spyOn / mock tests more robust, while fitting in the "composition" responsiblity.

### PluginLoaders

An abstract handle for the combination of known plugin loaders.
In other words, the "omniplugin" level of abstraction.

It's responsiblity is also composition.
Taking individual plugin interfaces, and composing them into "all plugins" semantics.

The `pluginLoaders.js` module exposes such "all plugins" semantics as functions like `mirrorAll()` and `combinedGraph()`.
It accepts the underlying plugin's loaders as interfaces for easy mocking.
The responsiblity to feed this the concrete plugin loaders lies with LoadContext.

### ComputeFunction

Similar to PluginLoaders, this accepts an interface like `TimelineCred.compute`.
During a load, we have the added responsiblity of using the TaskReporter.

[...]

Issue #1506, opened 2019-12-21
Static or dynamic loading refactor


Beanow:
## UniRef Context

In order to support [Unified Reference Detection](https://discourse.sourcecred.io/t/unified-reference-detection/301) we will have a dependency on our plugin's mirrored data in the createGraph stage.

Specifically, the reference detector should only return a "hit" for URLs that exists in our mirror. And to make it unified, every plugin will depend on every other plugin. E.g. GitHub's createGraph will depend on Discourse, GitHub and Initiatives data.

## Splitting load logic

Currently in `/src/api/load.js` we're treating each plugin's load as a single step which combines the mirroring and createGraph. Based on the new requirement the idea was to split this into two stages.

1. Populating the data layer of plugins (mirroring / syncing with APIs).
2. Graph creation.

In between those two steps, we can create our reference detector with data dependencies and pass that on to the graph creation functions. There's one exception to this, node contractions should always be last, so we'll consider this another stage.

```js
export type PopulationInput = {|
  +githubToken: string | null,
  +project: Project,
  +cacheDirectory: string,
  +taskReporter: TaskReporter,
|};

export type PopulationResult = {|
  +referenceDetector?: ReferenceDetector,
  +discourseData?: ReadRepository,
  +createGraph: GraphCreationFunction,
|};

export type GraphCreationInput = {|
  +referenceDetector: ReferenceDetector,
  +discourseData: ?ReadRepository,
|};

export type PopulateFunction = (PopulationInput) => ?Promise<PopulationResult>;

// Note: we're not providing the PopulationInput again. We're expecting necessary data
// to be provided through for example closures.
export type GraphCreationFunction = (GraphCreationInput) => Promise<Graph>;

async function stagedLoading(
  populateFns: $ReadOnlyArray<PopulateFunction>,
  input: PopulationInput
): Promise<Graph> {
  // First stage: parallel population

  // Reducing stage 1 results to stage 2 input

  // Second stage: graph creation.

  // Final stage: identity node contractions.
}
```

## Scaling flaw in this abstraction

So this sort of works in a generic way, but the Initiatives plugin wouldn't fit these stages.
The way the Initiatives reference detector would work is:
1. Use the Discourse data to create our set of `Initiative` objects and build a
`trackerToInitiative: Map<NodeAddressT, NodeAddressT>` based on this.
1. Use the Discourse reference detector to not need any knowledge about URL format and instead handle the `NodeAddressT` coming from the Discourse reference detector.
[...]

PR #1616, opened 2020-02-03
Backend: implement PluginLoaders.declarations

Beanow:

This is the first of several commits to create the PluginLoaders abstraction. Using this allows us to define "for all plugins" semantics, while keeping each underlying plugin interface flexible.

Test plan: `yarn test`

### DIFF
--- a/src/backend/pluginLoaders.js
+++ b/src/backend/pluginLoaders.js
@@ -1,0 +1,39 @@
+//@flow
+
+import {type Project} from "../core/project";
+import {type PluginDeclaration} from "../analysis/pluginDeclaration";
+import {type Loader as GithubLoader} from "../plugins/github/loader";
+import {type Loader as IdentityLoader} from "../plugins/identity/loader";
+import {type Loader as DiscourseLoader} from "../plugins/discourse/loader";
+
+/**
+ * A type combining all known plugin Loader interfaces.
+ *
+ * Using this allows us to define "for all plugins" semantics, while keeping
+ * each underlying plugin's interface flexible.
+ */
+export type PluginLoaders = {|
+  +github: GithubLoader,
+  +discourse: DiscourseLoader,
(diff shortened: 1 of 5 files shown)


src/backend/pluginLoaders.js
```diff
@@ -0,0 +1,39 @@
...
+ * Using this allows us to define "for all plugins" semantics, while keeping
+ * each underlying plugin's interface flexible.
+ */
+export type PluginLoaders = {|
+  +github: GithubLoader,
+  +discourse: DiscourseLoader,
+  +identity: IdentityLoader,
+|};
+
+/**
```
decentralion: love it... i will have use for this method in some of my own refactors. thanks.

src/plugins/identity/loader.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import {type PluginDeclaration} from "../../analysis/pluginDeclaration";
+import {declaration} from "./declaration";
+
+export interface Loader {
+  declaration(): PluginDeclaration;
```
decentralion: Hmm, why have a method that returns the declaration, rather than just having the declaration as a (read-only) property?

Note that having the function wrapper doesn't provide any additional safety because we are returning the same object either way, but it's fine because the object is deep-frozen by convention.

Beanow: No strong reason for this. Because in Flow the interfaces can do all sorts of wild things, it's almost the same as an object type. You can even turn them into object maps https://flow.org/en/docs/types/interfaces/#toc-interfaces-as-maps :scream: I find usages like these surprising for interfaces, so I arbitrarily stuck with just functions.

Beanow:
> it would be nice to have a base Loader interface that the specific loaders implement/inherit from.

This is explicitly avoided here, because of this discussion https://github.com/sourcecred/sourcecred/issues/1506

PR #1617, opened 2020-02-03
Backend: implement PluginLoaders.updateMirror

Beanow:
Depends on #1616 and #1614

Note, the return type is a CachedProject. See #1586 for discussion. Having this type allows us to create new functions with a semantic of requiring the project is mirrored into cache. It is opaque, because only the "all plugins" semantic which PluginsLoaders has could know when mirroring of a Project has been completed.

Additionally MirrorEnv is not a strict type. We're expecting this to be a subset of parameters. We'll use Flow to ensure we only use the ones we need from it.

Test plan: `yarn test`


src/plugins/discourse/loader.js
```diff
@@ -1,12 +1,50 @@
...
+  const mirror = new Mirror(repo, fetcher, serverUrl, mirrorOptions);
+  await mirror.update(reporter);
+}
+
+async function repository(
+  cache: CacheProvider,
+  serverUrl: string
+): Promise<SqliteMirrorRepository> {
+  // TODO: should replace base64url with hex, to be case insensitive.
+  const db = await cache.database(base64url.encode(serverUrl));
```
Beanow: Note: leaving this as it matches existing behavior. Plan to remove it in a future PR.

Beanow:
TL;DR, gave this some thought, and imo it doesn't compose well, and makes the interfaces of the plugins more dangerous.

---

Mirroring a plugin, such as Discourse, doesn't mirror the full Project, nor accepts it as an argument. So the type shouldn't contain the Project. It should at best have it's own options, like `DiscourseServer` > `CachedDiscourseServer`.

The resulting type composition would be a weird reinvention of Project. And we'd miss data including only the options for the plugins that need mirroring. So we'd repeat the Project as well.

```js
opaque type CachedProject = {|
  +project: Project, // Not to lose information
  +github: {|
    +cache: CacheProvider, // Shared "dangerous" cache
  |} | null,
  +discourse: {|
    +cache: CacheProvider, // Shared "dangerous" cache
    +discourseServer: DiscourseServer
  |} | null,
|};
```

There's no such thing as a Discourse / Github cache. The cache is expected to be shared between plugins and mutated. So after a `CachedDiscourseServer` would be returned, the cache it contains is being mutated, in parallel, to produce the `CachedGithubRepos`.
_Plugins should never assume they have exclusive ownership of the cache._

This is different from when `PluginLoaders.updateMirror` resolves, we _know_ we don't need to do these mutations of the cache anymore during this load pipeline, and it has cached everything it needs to, for this particular Project. Which is why we can add the context of: "we've completed our best-effort caching of this whole Project".

Finally, making strong assumptions about the cache state is "dangerous" in the same sense that assuming a TCP socket means guaranteed networking is. So I think composing types with a cache as if they're neat immutable data objects doesn't seem like a good idea.

While `PluginLoaders` does make this dangerous assumption. It's at least more reasonable to assume we have exclusive cache access here. In reality we would need to hold some type of filesystem and table write-lock for the duration of the load. But that's probably overkill and outside of scope here.

Beanow:
Rebased against master since dependencies were merged.

PR #1618, opened 2020-02-03
Backend: implement PluginLoaders.createPluginGraphs

Beanow:
Depends on #1617 and #1569

Similar to CachedProject, we're using an opaque PluginGraphs return type. Because only PluginLoaders can add the semantic of this being all plugin graphs, and to use this semantic in future functions.

Test plan: `yarn test`

### DIFF
--- a/src/backend/pluginLoaders.js
+++ b/src/backend/pluginLoaders.js
@@ -2,6 +2,7 @@
 
 import {TaskReporter} from "../util/taskReporter";
 import {type Project} from "../core/project";
+import {type WeightedGraph as WeightedGraphT} from "../core/weightedGraph";
 import {type PluginDeclaration} from "../analysis/pluginDeclaration";
 import {type CacheProvider} from "./cache";
 import {type GithubToken} from "../plugins/github/token";
@@ -31,10 +32,22 @@ opaque type CachedProject = {|
   +project: Project,
 |};
 
+/**
+ * Represents all disjoint WeightedGraphs for a CachedProject.
+ */
+opaque type PluginGraphs = {|
+  +graphs: $ReadOnlyArray<WeightedGraphT>,
+  +cachedProject: CachedProject,
+|};
+
 type MirrorEnv = {
   +githubToken: ?GithubToken,
   +reporter: TaskReporter,
   +cache: CacheProvider,
+};
+
+type GraphEnv = {
+  +githubToken: ?GithubToken,
 };
 
 /**
@@ -81,4 +94,32 @@ export async function updateMirror(
   }
   await Promise.all(tasks);
   return {project, cache};
+}
+
+/**
+ * Creates PluginGraphs containing all plugins requested by the Project.
+ */
+export async function createPluginGraphs(
+  {github, discourse}: PluginLoaders,
+  {githubToken}: GraphEnv,
+  {cache, project}: CachedProject
+): Promise<PluginGraphs> {
+  const tasks: Promise<WeightedGraphT>[] = [];
+  if (project.discourseServer) {
+    tasks.push(discourse.createGraph(project.discourseServer, cache));
+  }
+  if (project.repoIds.length) {
+    if (!githubToken) {
+      throw new Error("Tried to load GitHub, but no GitHub token set");
+    }
+    tasks.push(github.createGraph(project.repoIds, githubToken, cache));
+  }
+
+  // It's important to use Promise.all so that we can load the plugins in
+  // parallel -- since loading is often IO-bound, this can be a big performance
+  // improvement.
+  return {
+    graphs: await Promise.all(tasks),
+    cachedProject: {cache, project},
+  };
 }
--- a/src/backend/pluginLoaders.test.js
+++ b/src/backend/pluginLoaders.test.js
@@ -1,10 +1,27 @@
 // @flow
 
+import * as WeightedGraph from "../core/weightedGraph";
+import {node as graphNode} from "../core/graphTestUtil";
 import {createProject} from "../core/project";
 import {TestTaskReporter} from "../util/taskReporter";
 import {validateToken} from "../plugins/github/token";
 import {makeRepoId} from "../plugins/github/repoId";
 import * as PluginLoaders from "./pluginLoaders";
+
+const githubSentinel = graphNode("github-sentinel");
+const discourseSentinel = graphNode("discourse-sentinel");
+
+const mockGithubGraph = () => {
+  const wg = WeightedGraph.empty();
+  wg.graph.addNode(githubSentinel);
+  return wg;
+};
+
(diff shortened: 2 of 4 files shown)


src/backend/pluginLoaders.js
```diff
@@ -31,12 +32,24 @@ opaque type CachedProject = {|
   +project: Project,
 |};
 
+/**
+ * Represents all disjoint WeightedGraphs for a CachedProject.
```
decentralion: "all disjoint" -- are we guaranteeing that PluginGraphs need to be disjoint now?

src/backend/pluginLoaders.js
```diff
@@ -31,12 +32,24 @@ opaque type CachedProject = {|
...
+ */
+opaque type PluginGraphs = {|
+  +graphs: $ReadOnlyArray<WeightedGraphT>,
+  +cachedProject: CachedProject,
```
decentralion: why does PluginGraphs hold onto the cachedProject?

src/backend/pluginLoaders.js
```diff
@@ -82,3 +95,31 @@ export async function updateMirror(
...
+  {github, discourse}: PluginLoaders,
+  {githubToken}: GraphEnv,
+  {cache, project}: CachedProject
+): Promise<PluginGraphs> {
+  const tasks: Promise<WeightedGraphT>[] = [];
+  if (project.discourseServer) {
+    tasks.push(discourse.createGraph(project.discourseServer, cache));
+  }
+  if (project.repoIds.length) {
+    if (!githubToken) {
```
decentralion: We shouldn't need this check now -- we already have the cachedProject.

src/backend/pluginLoaders.js
decentralion: More generally, I don't see why we have this type at all. Just return a read-only array of graphs.

src/backend/pluginLoaders.js
Beanow: Flow disagrees :sweat_smile:, the `GraphEnv` accepts the same `?GithubToken` as elsewhere, meaning we need to refine it if we want to use it as a `GithubToken`.

There's a good reason it should be optional here. Such as loading a Project without any Github repos. Whether it's a CachedProject or not, it might not specify Github repos, thus not needing a token. Which is why the check happens after
` if (project.repoIds.length)`.

src/backend/pluginLoaders.js
Beanow: Well, specifically we require that `WeightedGraph.merge(graphs)` doesn't crash it, or it would be a bug. So not it's not a type guarantee, more an assumption.

src/backend/pluginLoaders.js
Beanow: The type `$ReadOnlyArray<WeightedGraphT>` does not allow the assumption we can do:
`WeightedGraph.merge(graphs)` without any conflict resolution.
Or rather, that it's a plugin bug if it does cause an error to be thrown.

The opaque `PluginGraphs` does add this extra context, allowing the assumption.

The _only_ other context in which we can make this assumption is the `createPluginGraphs` function itself. So if we're ditching this type. I suggest we don't return `$ReadOnlyArray<WeightedGraphT>`, but a merged `WeightedGraphT`.

As to containing the `CachedProject`, it's similar to why `CachedProject` has the `Project`.
The following `contractIdentities` step doesn't make sense to apply to any _other_ Project than the one that created the `PluginGraphs`. Though removing the `CacheProvider` from this type would be an assumption about what the expected operations are.

Beanow: If we want to remove the `PluginGraphs` type I would suggest:

We should return a `WeightedGraphT`. Because only in this context do we know it's correct to let `WeightedGraph.merge(graphs)` crash if there's unresolved conflicts.

The new `contractPluginGraphs` signature then would be:
```js
// Before
(PluginLoaders, PluginGraphs) => WeightedGraphT
// After
(PluginLoaders, WeightedGraphT, Project) => WeightedGraphT
```

Moving the responsibility to *use* this function to offer identities to `LoadContext`, as well as to ensure the `Project` is the same as the one that produced the input `WeightedGraphT`.

Which is the opposite approach of what we settled on for `updateMirror` and `CachedProject`.

src/backend/pluginLoaders.js
Beanow: Though, we can remove this token entirely when we've fixed this issue #1580

Beanow:
Rebased as 1 of 2 dependencies got merged.

src/backend/pluginLoaders.js
Beanow: Another alternative would be to make both `createPluginGraphs` and `contractPluginGraphs`  (#1619) private functions. So they're called by a public function which combines both. Same as
https://github.com/sourcecred/sourcecred/blob/806919a6ccd009d96b6645c2e8b6d6b68dd8ec91/src/api/loadWeightedGraph.js#L21-L41

src/backend/pluginLoaders.js
decentralion: Yeah, I meant that I wanted it to get removed per #1580. Fine with that being a different PR though.

src/backend/pluginLoaders.js
decentralion: Sure, but `WeightedGraph.merge` works on non-disjoint graphs. As a silly example, given that `g` is the WeightedGraph for a GitHub project, `WeightedGraph.merge([g,g])` returns a copy of `g`.

So let's fix the comment. Just saying it represents the plugin-generated WeightedGraphs for a CachedProject is likely fine.

src/backend/pluginLoaders.js
Beanow: There's some complexity involved that's blocking #1580, so I think that would be good as it's own PR.

src/backend/pluginLoaders.js
decentralion: OK. I'm actually fine with either approach. Feels like an implementation detail to me so I'd rather go for velocity... approving.

src/backend/pluginLoaders.js
Beanow: Wouldn't the Github WeightedGraph have the plugin's `weightsForDeclaration`? Meaning, same address weights, meaning it would throw without a resolver?

decentralion: Oh, you're right, it would.

src/backend/pluginLoaders.js
Beanow: Since the current implementation is fairly local to the PluginLoaders and LoadContext. Changing it should be relatively easy. I'll go with the solution as-is, but would be happy to revisit in a follow-up if needed :]

PR #1619, opened 2020-02-03
Backend: implement PluginLoaders.contractPluginGraphs

Beanow:
Depends on #1618 

Note: this doesn't include a `WeightedGraph.overrideWeights` step. Because overriding weights isn't related to plugins, this will be handled as a separate feature, later in the load pipeline.

Test plan: `yarn test`


Beanow:
> although since I requested changes on the preceding PRs it will need some tweaks too

After discussing the previous PRs, the feedback there didn't end up affecting this PR.

PR #1620, opened 2020-02-03
Backend: add ComputeFunction

Beanow:

> ### ComputeFunction
> 
> Similar to PluginLoaders, this accepts an interface like `TimelineCred.compute`.
> During a load, we have the added responsiblity of using the TaskReporter.
> 
> This lets us mock the concrete `TimelineCred.compute`, while testing just the extra functionality.
> The responsiblity to compose this with the concrete `TimelineCred.compute` lies with LoadContext.

Test plan: `yarn test`

### DIFF
--- a/src/backend/computeFunction.js
+++ b/src/backend/computeFunction.js
@@ -1,0 +1,35 @@
+//@flow
+
+import {type WeightedGraph} from "../core/weightedGraph";
+import {TaskReporter} from "../util/taskReporter";
+import {type TimelineCredParameters} from "../analysis/timeline/params";
+import {type PluginDeclaration} from "../analysis/pluginDeclaration";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+
+/**
+ * An abstract handle for TimelineCred.compute-like functions.
+ */
+export type ComputeFunction = (opts: ComputeOpts) => Promise<TimelineCred>;
+
+// Note: type should allow extra properties, it's expected to be a subset.
+type ComputeEnv = {
+  +reporter: TaskReporter,
+};
+
+type ComputeOpts = {|
+  weightedGraph: WeightedGraph,
+  params?: $Shape<TimelineCredParameters>,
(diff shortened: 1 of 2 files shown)


src/backend/computeFunction.js
```diff
@@ -0,0 +1,34 @@
...
+import {type TimelineCredParameters} from "../analysis/timeline/params";
+import {type PluginDeclaration} from "../analysis/pluginDeclaration";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+
+/**
+ * An abstract handle for TimelineCred.compute-like functions.
+ */
+export type ComputeFunction = (opts: ComputeOpts) => Promise<TimelineCred>;
+
+// Note: type should not be strict, it's expected to be a subset.
```
decentralion: I don't understand this comment.

src/backend/computeFunction.js
```diff
@@ -0,0 +1,34 @@
...
+type ComputeEnv = {
+  +reporter: TaskReporter,
+};
+
+type ComputeOpts = {|
+  weightedGraph: WeightedGraph,
+  params?: $Shape<TimelineCredParameters>,
+  plugins: $ReadOnlyArray<PluginDeclaration>,
```
decentralion: You can put
`// TODO(@decentralion, #1557): remove plugins arg`
here if you want

src/backend/computeFunction.js
Beanow: My bad, Flow calls this "exact" instead of "strict".
https://flow.org/en/docs/types/objects/#toc-exact-object-types
In other words, disallowing extra properties.

In this case we _do_ want extra properties to be allowed.

src/backend/computeFunction.js
Beanow: Added this.

src/backend/computeFunction.js
Beanow: Rephrased the comment.

PR #1621, opened 2020-02-03
Backend: implement LoadContext

Beanow:
Depends on #1619 and #1620

It uses a new trick of aliasing external module functions as private properties. This makes the spyOn / mock tests more robust, while fitting in the composition responsibility.

Test plan: `yarn test`

Edit: **Conflicts with #1624**
`LoadResult` needs to include the new PluginDeclarations type.

### DIFF
--- a/src/backend/loadContext.js
+++ b/src/backend/loadContext.js
@@ -1,0 +1,125 @@
+//@flow
+
+import {type Project} from "../core/project";
+import {type Weights as WeightsT} from "../core/weights";
+import {type WeightedGraph as WeightedGraphT} from "../core/weightedGraph";
+import * as WeightedGraph from "../core/weightedGraph";
+import {type TimelineCredParameters} from "../analysis/timeline/params";
+import {type GithubToken} from "../plugins/github/token";
+import {type CacheProvider} from "./cache";
+import {TaskReporter} from "../util/taskReporter";
+import {TimelineCred} from "../analysis/timeline/timelineCred";
+import {type ComputeFunction as ComputeFunctionT} from "./computeFunction";
+import {type PluginLoaders as PluginLoadersT} from "./pluginLoaders";
+import * as ComputeFunction from "./computeFunction";
+import * as PluginLoaders from "./pluginLoaders";
+import {default as githubLoader} from "../plugins/github/loader";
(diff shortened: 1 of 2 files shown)


src/backend/loadContext.js
```diff
@@ -0,0 +1,115 @@
...
+      this._pluginLoaders,
+      pluginGraphs
+    );
+    const weightedGraph = this._overrideWeights(
+      contractedGraph,
+      weightsOverrides
+    );
+    const plugins = this._declarations(this._pluginLoaders, project);
+    const cred = await this._computeTask(this._compute, this._options, {
+      params: params || {},
```
decentralion: we shouldn't need `|| {}` -- the type signature for computTask already allows null-ish params

Beanow:
Rebased as 1 of 2 dependencies got merged.

src/backend/loadContext.js
Beanow: I'm afraid not. This is another case where we're mismatching:
`key?: $Shape<T>` (what `.compute()` uses)
`key: ?$Shape<T>` (what `load` uses)

However. It's a good point, I want to fix that at this invocation and move this problem to the `api/load.js` file.

Beanow:
1) I've updated the LoadResult to include `PluginDeclarations` which addresses the conflict with #1624.
https://github.com/sourcecred/sourcecred/blob/f7c92c9ada14dbc1c0321c9ff6b2e4cf879e58ec/src/backend/loadContext.js#L21-L25

2) To address a nit about optional arguments. I've updated the `LoadContext.load` signature to use named optional parameters. Not only making `TimelineCredParameters` optional, but also the override weights.

https://github.com/sourcecred/sourcecred/blob/f7c92c9ada14dbc1c0321c9ff6b2e4cf879e58ec/src/backend/loadContext.js#L33-L36
https://github.com/sourcecred/sourcecred/blob/f7c92c9ada14dbc1c0321c9ff6b2e4cf879e58ec/src/backend/loadContext.js#L91-L94

3) Finally I've updated tests to cover both changes.

src/backend/loadContext.js
Beanow: Addressed by updating signature here. See https://github.com/sourcecred/sourcecred/pull/1621#issuecomment-582116351

PR #1622, opened 2020-02-03
Switch to LoadContext based loading

Beanow:
Depends on #1621

This naming is temporary, as the old loading code is removed it will be named load and replace the existing function.

Test plan:
- `yarn test --full`
- Manually `bin/sourcecred.js load` a project

### DIFF
--- a/src/api/load.js
+++ b/src/api/load.js
@@ -23,6 +23,7 @@ import {type Weights as WeightsT} from "../core/weights";
 import {loadWeightedGraph} from "./loadWeightedGraph";
 import {DataDirectory} from "../backend/dataDirectory";
 import {type CacheProvider} from "../backend/cache";
+import {LoadContext} from "../backend/loadContext";
 
 export type LoadOptions = {|
   +project: Project,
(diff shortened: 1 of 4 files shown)


Beanow:
Made a change to the `LoadContext.load` call, because it's signature changed to take a named options object.

https://github.com/sourcecred/sourcecred/blob/c61d356dc2bed3f44e2d006349992e05b0ab8bc5/src/api/load.js#L50-L53

src/api/load.js
```diff
@@ -33,6 +34,26 @@ export type LoadOptions = {|
...
+    sourcecredDirectory,
+    githubToken,
+    project,
+    params,
+    weightsOverrides,
+  } = options;
+  const data = new DataDirectory(sourcecredDirectory);
+  const context = new LoadContext({cache: data, githubToken, reporter});
+  const result = await context.load(project, {
+    params: params || {},
```
decentralion: (nit): shouldn't params be pass thru?

Beanow: This time we have, the problem:
https://github.com/sourcecred/sourcecred/blob/c61d356dc2bed3f44e2d006349992e05b0ab8bc5/src/api/load.js#L28-L30
And the workaround in the same file :sweat_smile:
A direct pass through would cause a Flow error.

Currently I'm using this `loadContext` function as a drop-in replacement of the legacy `load`. After all legacy stuff is removed we could update the API to be more natural here.

PR #1623, opened 2020-02-03
Add to/fromJSON for plugin declarations

decentralion:
This commit adds a simple method for (de-)serializing arrays of
PluginDeclarations. This will allow us to save PluginDeclarations for
consumption by the frontend, without having them bundled with cred
results in TimelineCred. Thus, we can simplify and clean up TimelineCred
as described in #1557.

Test plan: Inspect unit tests and snapshots; `yarn test` passes.

### DIFF
--- a/src/analysis/pluginDeclaration.js
+++ b/src/analysis/pluginDeclaration.js
@@ -4,6 +4,9 @@ import {type NodeAddressT, type EdgeAddressT} from "../core/graph";
 import type {EdgeType, NodeType, NodeAndEdgeTypes} from "./types";
 import * as Weights from "../core/weights";
 import {type Weights as WeightsT} from "../core/weights";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
+
+const COMPAT_INFO = {type: "sourcecred/pluginDeclarations", version: "0.1.0"};
 
 // TODO(@decentralion): Maybe merge this file with analysis/types
 
@@ -22,6 +25,18 @@ export type PluginDeclaration = {|
   // user type weights.
   +userTypes: $ReadOnlyArray<NodeType>,
 |};
+
+export type PluginDeclarations = $ReadOnlyArray<PluginDeclaration>;
+
+export function toJSON(decs: PluginDeclarations): PluginDeclarationsJSON {
+  return toCompat(COMPAT_INFO, decs);
+}
+
+export function fromJSON(json: PluginDeclarationsJSON): PluginDeclarations {
+  return fromCompat(COMPAT_INFO, json);
+}
+
+export type PluginDeclarationsJSON = Compatible<PluginDeclarations>;
 
 export function combineTypes(
   decs: $ReadOnlyArray<PluginDeclaration>
--- a/src/analysis/pluginDeclaration.test.js
+++ b/src/analysis/pluginDeclaration.test.js
@@ -1,11 +1,14 @@
 // @flow
 
+import stringify from "json-stable-stringify";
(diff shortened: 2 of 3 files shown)


src/analysis/pluginDeclaration.js
```diff
@@ -23,6 +26,18 @@ export type PluginDeclaration = {|
...
+
+export function toJSON(decs: PluginDeclarations): PluginDeclarationsJSON {
+  return toCompat(COMPAT_INFO, decs);
+}
+
+export function fromJSON(json: PluginDeclarationsJSON): PluginDeclarations {
+  return fromCompat(COMPAT_INFO, json);
+}
+
+export type PluginDeclarationsJSON = Compatible<PluginDeclarations>;
```
Beanow: Nit: I don't think `PluginDeclarationsJSON` needs to be exported.

src/analysis/pluginDeclaration.test.js
```diff
@@ -52,4 +55,29 @@ describe("analysis/pluginDeclaration", () => {
...
+    });
+    it("works round-trip on an non-empty declaration", () => {
+      const json = toJSON([nonEmptyDeclaration]);
+      const result = fromJSON(json);
+      expect(result).toEqual([nonEmptyDeclaration]);
+    });
+    it("snapshots on an non-empty declaration", () => {
+      // stringify to avoid having literal NUL bytes in our source.
+      expect(stringify(toJSON([nonEmptyDeclaration]))).toMatchInlineSnapshot(
+        `"[{\\"type\\":\\"sourcecred/pluginDeclarations\\",\\"version\\":\\"0.1.0\\"},[{\\"edgePrefix\\":\\"E\\\\u0000\\",\\"edgeTypes\\":[{\\"backwardName\\":\\"is pointed to\\",\\"defaultWeight\\":{\\"backwards\\":3,\\"forwards\\":2},\\"description\\":\\"a type\\",\\"forwardName\\":\\"points\\",\\"prefix\\":\\"E\\\\u0000edge\\\\u0000\\"}],\\"name\\":\\"non-empty\\",\\"nodePrefix\\":\\"N\\\\u0000\\",\\"nodeTypes\\":[{\\"defaultWeight\\":2,\\"description\\":\\"a type\\",\\"name\\":\\"node\\",\\"pluralName\\":\\"nodes\\",\\"prefix\\":\\"N\\\\u0000node\\\\u0000\\"}],\\"userTypes\\":[]}]]"`
```
Beanow: Nit: suggest not using an inline snapshot.

Inline snapshots are something I interpret as, "I should read these to validate they're testing expected behavior". Since we already have the round-trip for these I think the contents don't need to be visible, as long as they don't change.

decentralion: Changed.

src/analysis/pluginDeclaration.js
decentralion: We don't need it yet, but for consistency with our other data formats, I will export it. Since we're writing it to disk, it makes sense that we'd allow others to know about the type.

Beanow: *Weeelll*, as we found out with Projects here https://github.com/sourcecred/sourcecred/pull/1518
The real signature for the fromJSON inputs are `Compatible<any>` (maybe even just `any`).

Because they're just an object we parsed from JSON and still needs runtime validation. With that in mind, I think there's rarely any case where we would use the type. For a return type it makes sense, but in practice it tends to be a one-liner which serializes it immediately after.

```js
stringify(toJSON(thing));
```

PR #1624, opened 2020-02-03
api/load: save plugin declarations to disk

decentralion:
This builds on #1623 and is another step towards separating cred
computation from plugin declarations, as described in #1557. Basically,
this will allow the frontend to get plugin declarations even if the
TimelineCred computation never saw them.

Test plan: See included unit tests, also try loading a project and
inspect the newlys saved file.


src/api/load.js
```diff
@@ -107,6 +110,10 @@ export async function load(
   const graphJSON = WeightedGraph.toJSON(weightedGraph);
   await fs.writeFile(graphFile, stringify(graphJSON));
 
+  const pluginsFile = path.join(projectDirectory, "pluginDeclarations.json");
+  const pluginsJSON = pluginsToJSON(plugins);
+  await fs.writeFile(pluginsFile, stringify(pluginsJSON));
```
Beanow: Directly managing the project filesystem from load is going to be replaced, since `DataDirectory` / `ProjectStorageProvider` is responsible for this. https://github.com/sourcecred/sourcecred/blob/master/src/backend/dataDirectory.js

Would you be open to updating `DataDirectory` / `ProjectStorageProvider` to accept a `PluginDeclarations` type instead, and switch to `DataDirectory` for writing the file instead?

decentralion: I updated the DataDirectory and ProjectStorage classes to accept the PluginDeclarations. I didn't refactor load to actually use the DataDirectory, but it should make it trivial for you to include this change in your refactored pipeline (if this merges first), or trivial for me to incorporate it in yours (if yours merges first).

Beanow:
I've dismissed the approval of #1621, so I need to resolve the conflict there. Feel free to merge this one first!

PR #1625, opened 2020-02-04
Frontend gets plugins from disk, not TimelineCred

decentralion:
This commit modifies the frontend so that it now pulls plugin
declarations from disk, rather than from the TimelineCred. This will
allow us to decouple the TimelineCred from the PluginDeclarations, which
is another step towards #1557.

As proof that the frontend no longer gets plugins from the TimelineCred,
I removed the public `plugins()` method on TimelineCred.

Test plan: The frontend is somewhat sketchily tested. `yarn test`
passing is good, manual inspection of the frontend is also necessary;
I've done this.

### DIFF
--- a/src/analysis/timeline/timelineCred.js
+++ b/src/analysis/timeline/timelineCred.js
@@ -76,10 +76,6 @@ export class TimelineCred {
     return this._params;
   }
 
-  plugins(): $ReadOnlyArray<PluginDeclaration> {
-    return this._plugins;
-  }
-
   /**
    * Creates a new TimelineCred based on the new Parameters.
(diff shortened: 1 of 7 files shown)


decentralion:
> I think if we required declarations to be an argument for reanalyze and userNodes, we could remove them from the constructor, private properties and JSON. Though that would make sense to me as a follow-up as well. I'll defer to you if it should be in this PR or not.

Yeah, the goal here is to decouple so that the pure frontend code doesn't depend on grabbing the plugins from `TimelineCred`. I'm going to remove the plugins from TimelineCred entirely in a future PR. One step at a time. :)

PR #1627, opened 2020-02-04
Remove types from TimelinePagerank

decentralion:
This commit modifies the timelinePagerank module so that it no longer
takes in node/edge types. Instead, the timelinePagerank just takes a
WeightedGraph and uses weights from that WeightedGraph. This is a key
part of decoupling the core cred computation logic from the plugin
logic, as described in #1557.

I also modified the timelinePagerank module's immediate dependencies
(the weightEvaluator module) to do the same. Since the weight evaluators
now have a simpler contract (no overriding, etc), the unit tests have
been simplified.

Test plan: It's a simple refactor, so `yarn test` should be sufficient.
As a bit of added caution, I manually tested changing weights in the
frontend, and verified that cred updates as expected.

### DIFF
--- a/src/analysis/weightsToEdgeEvaluator.test.js
+++ b/src/analysis/weightsToEdgeEvaluator.test.js
@@ -1,6 +1,5 @@
 // @flow
 
-import deepFreeze from "deep-freeze";
 import {NodeAddress, EdgeAddress} from "../core/graph";
 import {type Weights as WeightsT} from "../core/weights";
 import * as Weights from "../core/weights";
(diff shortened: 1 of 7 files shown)


src/analysis/weightsToEdgeEvaluator.test.js
```diff
@@ -16,57 +15,19 @@ describe("analysis/weightsToEdgeEvaluator", () => {
...
-      nodeTypes: [fallbackNodeType, srcNodeType],
-      edgeTypes: [fallbackEdgeType],
-    });
+    const evaluator = weightsToEdgeEvaluator(weights);
     return evaluator(edge);
   }
 
   it("applies default weights when none are specified", () => {
-    expect(evaluateEdge(Weights.empty())).toEqual({forwards: 1, backwards: 2});
+    expect(evaluateEdge(Weights.empty())).toEqual({forwards: 1, backwards: 1});
```
Beanow: This change had me scratching my head for a bit. Assuming the previous `defaultWeight: 2` src node weight caused backwards to be `1 (default edge weight) * 2 (src node weight) * 1 (default node weight)`.

PR #1631, opened 2020-02-05
Rename core/attribution to core/algorithm

decentralion:
As part of my cleanup to make it easy to document and re-implement the
SourceCred algorithm, I want a place in core where we can consolidate
the js implementation. I'm renaming `core/attribution` to
`core/algorithm` to make this clearer.

Test plan: It's just a rename. `yarn test` passing is sufficient to
assure us of correctness.


decentralion:
@vsoch: This one is kind of pro-forma as far as reviews are concerned (you can just approve it) -- I added you as the reviewer also as an FYI that `core/algorithm` is now a thing, and what its intent is.

PR #1633, opened 2020-02-05
Remove TimelineCred.reduceSize

decentralion:
TimelineCred has a `reduceSize` method which discards cred for most
nodes, keeping only cred for the top nodes of each type across all time.
I've wanted to remove this for a while, because it is a bad fit for the
kind of experimentation we're starting to do with showing the top nodes
for recent activity periods. Since the recent nodes haven't had much
time to accumulate cred, they are almost all discarded by reduceSize. As
an added inducement, I want to get rid of reduceSize for #1557 because
it requires type information.

`reduceSize` still serves one function, which is enabling the frontend
to load faster because it only loads a smaller amount of data which is
discoverable in the UI. However, it doesn't make sense to discard most
of the data just for a fast UI load--we can later make another data
structure which is tuned for the needs of the frontend, and have that
data structure include only summary statistics.

This will make the cred.json file much larger for large repos, e.g. I
expect that loading tensorflow/tensorflow would now go over the 100MB
hard cap for GitHub pages. However, right now we're prioritizing using
SourceCred for medium-small projects (e.g. SourceCred itself, and maybe
Maker), so this isn't a current concern.

### DIFF
--- a/src/analysis/timeline/timelineCred.js
+++ b/src/analysis/timeline/timelineCred.js
@@ -151,59 +151,6 @@ export class TimelineCred {
     return this.credSortedNodes(userTypes.map((x) => x.prefix));
   }
 
-  /**
-   * Create a new, filtered TimelineCred, by removing low-scored nodes.
-   *
-   * Cred Graphs may have a huge number of small contributions, like comments,
-   * in which end users are not particularly interested. However, the size of
-   * the TimelineCred offered to the frontend matters quite a bit. Therefore,
-   * we can use this method to discard almost all nodes in the graph.
-   *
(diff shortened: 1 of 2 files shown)


Beanow:
For reference. I did a manual test run for the impact on our canonical cred.

```
before 13,8 MB (13764591 bytes)
after  30,0 MB (29979854 bytes)
```

It's worth looking into compression in the explorer. Even on the fastest gzip setting:
```
$ gzip -1 cred.json
cred.json.gz 5,2 MB (5239655 bytes)
```

decentralion:
Cool, thanks. We do have precedent for compressing things before sending them to the frontend, so we can do that without too much difficulty in a followon PR.

PR #1637, opened 2020-02-05
Create compatIO utility functions

Beanow:
Currently similar code to read/write Compatible JSON files is copy
pasted across the code. This takes some common practices and provides
a generic utility for it.

Correct Flow type usage can't be detected if the JSON type is opaque
though. GraphJSON is an example of this, so removed the opaque for a
smoke test.

Test plan: `yarn test`

### DIFF
--- a/src/backend/compatIO.js
+++ b/src/backend/compatIO.js
@@ -1,0 +1,60 @@
+// @flow
+
+import fs from "fs-extra";
+import stringify from "json-stable-stringify";
+import {type Compatible} from "../util/compat";
+
+/**
+ * Utility functions, which allow us to read/write a Compatible of any
+ * type to a JSON file. They should automatically detect the Flow types
+ * and produce errors when misused.
+ *
+ * Note, the Compatible<T> type used cannot be opaque.
+ */
+
+/**
+ * Creates a Compatible JSON file writer.
+ * The writer's Flow type for `data` will automatically match the provided
+ * `toJSON`'s argument type.
+ * Optionally takes a `typeName` to improve error messages.
+ */
+export function compatWriter<T, J>(
+  toJSON: (T) => Compatible<J>,
+  typeName?: string
+): (path: string, data: T) => Promise<void> {
+  return async (path: string, data: T): Promise<void> => {
+    try {
+      await fs.writeFile(path, stringify(toJSON(data)));
+    } catch (e) {
+      throw new Error(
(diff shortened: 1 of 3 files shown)


src/backend/compatIO.js
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Creates a Compatible JSON file writer.
+ * The writer's Flow type for `data` will automatically match the provided
+ * `toJSON`'s argument type.
+ * Optionally takes a `typeName` to improve error messages.
+ */
+export function compatWriter<T, J>(
+  toJSON: (T) => Compatible<J>,
+  typeName?: string
```
decentralion: Rather than take the `typeName` as a string, how about taking the corresponding `COMPAT_INFO`?

src/backend/compatIO.test.js
```diff
@@ -0,0 +1,212 @@
...
+
+      // Then
+      const actualData = (await fs.readFile(filePath)).toString("utf-8");
+      expect(actualData).toEqual(
+        `[{"type":"mock","version":"1"},{"aaa":"first","bbb":"second"}]`
+      );
+    });
+  });
+
+  describe("compatReader", () => {
```
decentralion: I'd like to see a test that shows that we get a helpful error message on compat version mismatch.

src/backend/compatIO.js
Beanow: Well, I haven't found a good way to get Flow types as strings at runtime. So we could get the `COMPAT_INFO` by any-casting the opaque type _after_ `toJSON` succeeded.

src/backend/compatIO.test.js
Beanow: Added one for type and version mismatch.

src/backend/compatIO.js
Beanow: Oh, by taking you meant manually as an argument? That could work, though I think it's nice to add your own names in free form. Since it's usage is to enhance error messages. It should be something clear to the user.

Beanow:
I've added some examples in this branch. @decentralion 
https://github.com/sourcecred/sourcecred/compare/bean/compatio...bean/compatio-examples

Particularly here, I think it shows why this saves copy pasting and headaches.
https://github.com/sourcecred/sourcecred/commit/471990069cbc09dfd05c89ac9e6cc2e5b6c71e75?diff=unified#diff-d34cd625d39ce6c0eb7d568a073db5afL155-R156

PR #1638, opened 2020-02-05
Initiatives: add initiativesMap type

Beanow:
Based on [forum discussion][1], Initiatives should be tracked as in files.

There's one main issue with storing the existing Initiative type as
JSON in a file, there's no natural NodeAddress for a file-based tracker.
This type resolves that by using the keys of a JSON object as a unique
reference within that file.

Test plan: `yarn test`

[1]: https://discourse.sourcecred.io/t/576

### DIFF
--- a/src/plugins/initiatives/initiativesMap.js
+++ b/src/plugins/initiatives/initiativesMap.js
@@ -1,0 +1,32 @@
+// @flow
+
+import {type Compatible, fromCompat, toCompat} from "../../util/compat";
+import {type URL} from "./initiative";
+
+const COMPAT_INFO = {type: "sourcecred/initiatives", version: "0.1.0"};
+
+/**
+ * JSON (compat) representation of an Initiative.
+ *
+ * Major difference is that instead of using tracker addresses, we use object
+ * keys as an ID, and derive the tracker address from that.
+ */
+export type InitiativesMap = {[entryKey: string]: InitiativeEntry};
+
+type InitiativeEntry = {|
+  +title: string,
+  +timestampMs: number,
+  +completed: boolean,
+  +dependencies: $ReadOnlyArray<URL>,
+  +references: $ReadOnlyArray<URL>,
+  +contributions: $ReadOnlyArray<URL>,
+  +champions: $ReadOnlyArray<URL>,
+|};
+
+export function fromJSON(j: Compatible<any>): InitiativesMap {
+  return fromCompat(COMPAT_INFO, j);
+}
+
+export function toJSON(m: InitiativesMap): Compatible<InitiativesMap> {
+  return toCompat(COMPAT_INFO, m);
+}
--- a/src/plugins/initiatives/initiativesMap.test.js
+++ b/src/plugins/initiatives/initiativesMap.test.js
@@ -1,0 +1,29 @@
+// @flow
+
+import {fromJSON, toJSON} from "./initiativesMap";
+
+describe("plugins/initiatives/initiativesMap", () => {
+  const exampleMap = {
+    "2020-01-08_sample-initiative": {
+      title: "Sample initiative",
+      timestampMs: 1578520917766,
+      completed: false,
+      champions: [],
+      contributions: [],
+      dependencies: [],
+      references: [],
+    },
+  };
+
+  describe("toJSON/fromJSON", () => {
+    it("should handle an example map round-trip", () => {
+      // Given
+
+      // When
+      const actual = fromJSON(toJSON(exampleMap));
+
+      // Then
+      expect(actual).toEqual(exampleMap);
+    });
+  });
+});


src/plugins/initiatives/initiativesMap.js
```diff
@@ -0,0 +1,32 @@
...
+export type InitiativesMap = {[entryKey: string]: InitiativeEntry};
+
+type InitiativeEntry = {|
+  +title: string,
+  +timestampMs: number,
+  +completed: boolean,
+  +dependencies: $ReadOnlyArray<URL>,
+  +references: $ReadOnlyArray<URL>,
+  +contributions: $ReadOnlyArray<URL>,
+  +champions: $ReadOnlyArray<URL>,
```
decentralion: I think it should be `$ReadOnlyArray<Alias>` instead of URL. Since we already have a concise sublanguage that's intended for describing user identities.

src/plugins/initiatives/initiativesMap.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+export type InitiativesMap = {[entryKey: string]: InitiativeEntry};
+
+type InitiativeEntry = {|
+  +title: string,
+  +timestampMs: number,
+  +completed: boolean,
+  +dependencies: $ReadOnlyArray<URL>,
+  +references: $ReadOnlyArray<URL>,
+  +contributions: $ReadOnlyArray<URL>,
```
decentralion: I suspect we have more edge types than we need here -- it's not super clear, for example, when something is a dependency vs a contribution. Since it's easier to add complexity than to remove it, I propose that we remove the dependency type until we're sure what we need it for.

Beanow:
> The description says that this pull solves the problem of assigning NodeAddressT to initiatives, but I don't see any code for that.

https://github.com/sourcecred/sourcecred/blob/5780c5c8ef7c53b41c594d0b189dfff6ae5de51b/src/plugins/initiatives/initiativesMap.js#L14

We declare the type as-stored-in-files. Which has `entryKey: string` to derive a `NodeAddressT` from.
Indeed this PR (no longer) includes the conversion between `InitiativesMap` and `Initiative[]` (which includes a tracker `NodeAddressT` each).

src/plugins/initiatives/initiativesMap.js
Beanow: These types are compatible with the main `Initiative` type, which we will need for conversion.
https://github.com/sourcecred/sourcecred/pull/1417/files

Using aliases would increase complexity, as we would need two types of resolvers. One for aliases and another for URLs.

src/plugins/initiatives/initiativesMap.js
Beanow: These types are compatible with the main `Initiative` type, which we will need for conversion.
https://github.com/sourcecred/sourcecred/pull/1417/files

We can remove these in it's own PR, though I feel that would be a loss of fidelity.
Because the directional flows of these edges are different.
https://github.com/sourcecred/sourcecred/blob/c9958e346ee789f41e1e406acc864c241245e70e/src/plugins/initiatives/declaration.js

decentralion: Sounds good.

decentralion:
Also, I recommend adding a link to the forum discussion that the PR body mentions:
https://discourse.sourcecred.io/t/rfc-change-initiatives-plugin-to-be-file-based-instead-of-discourse-based/576

Beanow:
Good point. I remember in a previous discussion giving up on reference detection of Initiatives from files. Right now I propose we add a `+permalink: URL` field. We can generate an arbitrary link, and replace those when we have a new explorer which offers node permalinks.

For example `https://github.com/sourcecred/cred/blob/master/initiatives.json#entry_78192739`

src/plugins/initiatives/initiativesMap.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * JSON (compat) representation of an Initiative.
+ *
+ * Major difference is that instead of using tracker addresses, we use object
+ * keys as an ID, and derive the tracker address from that.
+ */
+export type InitiativesMap = {[entryKey: string]: InitiativeEntry};
+
+type InitiativeEntry = {|
```
decentralion: propose adding `initiativeWeight: number`

then if completed === true, then we set a custom weight of `initiativeWeight`, otherwise `0`

i'm also fine with this being another PR.

Beanow: Yeah this should also be updated in the base type and createGraph, so suggest a different PR for that.

Beanow:
Resolving the linking and unique key issue through https://github.com/sourcecred/sourcecred/issues/1640

PR #1639, opened 2020-02-05
Initiatives: add conversion functions for InitiativesMap

Beanow:
Depends on #1638 

These functions provide a general purpose conversion between
InitiativesMap and $ReadOnlyArray<Initiative>.

By normalizing the format of the `entryKey` we produce, it allows for a
lot of operations, without needing to come up with your own unique IDs.
Such as merging several InitiativesMaps or loading Initiatives from a
different source, such as Discourse, to store them in files instead.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/initiativesMap.js
+++ b/src/plugins/initiatives/initiativesMap.js
@@ -1,7 +1,10 @@
 // @flow
 
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
 import {type Compatible, fromCompat, toCompat} from "../../util/compat";
-import {type URL} from "./initiative";
+import type {Initiative, URL} from "./initiative";
+
+const trackerPrefix = NodeAddress.fromParts(["sourcecred", "initiativesMap"]);
 
 const COMPAT_INFO = {type: "sourcecred/initiatives", version: "0.1.0"};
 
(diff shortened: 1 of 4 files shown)


src/plugins/initiatives/initiativesMap.js
```diff
@@ -1,7 +1,10 @@
 // @flow
 
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
 import {type Compatible, fromCompat, toCompat} from "../../util/compat";
-import {type URL} from "./initiative";
+import type {Initiative, URL} from "./initiative";
+
+const trackerPrefix = NodeAddress.fromParts(["sourcecred", "initiativesMap"]);
```
decentralion: by convention, the first two entries should be `[organizationPublishingPlugin, pluginName]`, and subsequent pieces are plugin-defined. Since this is our initiatives plugin, it should start with `["sourcecred", "initiatives"]`

src/plugins/initiatives/initiativesMap.js
```diff
@@ -30,3 +33,90 @@ export function fromJSON(j: Compatible<any>): InitiativesMap {
...
+  return initiatives;
+}
+
+export function _trackerAddress(entryKey: string): NodeAddressT {
+  return NodeAddress.append(trackerPrefix, entryKey);
+}
+
+/**
+ * Creates an InitiativesMap from Initiatives. Warning: this is a lossy
+ * operation, as the original tracker address will be dropped!
```
decentralion: I don't understand what this means. What original tracker address is being dropped? It appears that what you really mean is that this method (which I propose calling `createInitiativesMap`) does not faithfully include the input initiatives, but rather re-writes their keys in attempt to avoid collissions. is this correct?

Beanow: That's correct.

The tracker is the node address of the source which produced the initiative.
Such as a Discourse topic.
https://github.com/sourcecred/sourcecred/blob/c9958e346ee789f41e1e406acc864c241245e70e/src/plugins/initiatives/initiative.js#L25

The node address of the initiative node itself is a prefix + the tracker
https://github.com/sourcecred/sourcecred/blob/c9958e346ee789f41e1e406acc864c241245e70e/src/plugins/initiatives/createGraph.js#L22-L27

Beanow:
> This seems very non-robust and it means that cross-initiative references may arbitrarily break due to e.g. re-ordering the list of initiatives.

My understanding of the requirements were a _predictable_ address given the same file. I can see how these functions would set the expectation of a _GUID_ you can permalink instead, which isn't helpful.

That said: "defined only once (single source of truth)" we don't _have_ a single source of truth here.

But I have a simpler solution though. Adding an optional `generateKey` function to `initiativesToMap` 
when it can't faithfully convert to the same key. When it's not set (e.g. we're not adopting them from a different source or merging files), it will throw when a faithful conversion isn't possible.

Beanow:
Resolving the linking and unique key issue through https://github.com/sourcecred/sourcecred/issues/1640

Issue #1640, opened 2020-02-06
Initiatives: filenames as unique key (InitiativesDirectory)


Beanow:
So we ran into the issue of having keys which don't conflict and we can link to.
The simple solution for now would be to use filenames as the key and only allow 1 Initiative per file.

The concept of an `InitiativesDirectory`.
To coop with filesystem scaling issues we could upgrade this concept to have a "glob" like `./**/*.json`. However for now we'll only use a single level. `./*.json`

decentralion:
The glob doesn't really solve all the filesystem scaling issues; there's still a limit on the number of file descriptors a process can read (some docs [here](https://docs.oracle.com/cd/E19623-01/820-6168/file-descriptor-requirements.html)). That said, I am thinking of this as a temporary solution, before long we'll want to build an actual database for these initiatives (and have a proper CRUD type app for writing, rather than doing it all via literal pull request). So I'm fine with this for now.

cc @wchargin

PR #1641, opened 2020-02-06
Add InitiativesDirectory and InitiativeFile types

Beanow:
_Depends on #1646 #1647_

Based on [forum discussion][1], Initiatives should be tracked in files.

The main issue with storing the existing Initiative type as JSON in a
file, is there's no natural NodeAddress or URL for a file-based tracker.
This type resolves that by using the file name within a directory as a
unique reference and requiring a remoteUrl for referencing. (See #1640)

[1]: https://discourse.sourcecred.io/t/576

Test plan: `yarn test`

Further note: I've added more types that I normally would in a single PR.
I would have started with just `InitiativeFile`, but adding this type in a `initiativesDirectory.js` file warranted some more context. As noted in the comments, `InitiativeFile` is not intended to be used separately from an `InitiativesDirectory`.

### DIFF
--- a/src/plugins/initiatives/initiativesDirectory.js
+++ b/src/plugins/initiatives/initiativesDirectory.js
@@ -1,0 +1,63 @@
+// @flow
+
+import {type ReferenceDetector} from "../../core/references";
+import {type Compatible, fromCompat, toCompat} from "../../util/compat";
+import {type InitiativeRepository, type URL} from "./initiative";
+
+/**
+ * Represents an Initiatives directory.
+ *
+ * Initiative directories contain a set of InitiativeFiles in a `*.json` pattern.
+ * Where the file name is the ID of that Initiative.
+ * Additionally we require a `remoteUrl` for this directory. We expect this directory
+ * to be something you can browse online. This allows us to create a ReferenceDetector.
+ */
+export type InitiativesDirectory = {|
+  +localPath: string,
+  +remoteUrl: string,
+|};
+
+/**
+ * Opaque because we only want this file's functions to create these load results.
+ * However we do allow anyone to consume it's properties.
+ */
+export opaque type LoadedInitiativesDirectory: {|
+  +referenceDetector: ReferenceDetector,
+  +initiatives: InitiativeRepository,
+|} = {|
+  +referenceDetector: ReferenceDetector,
+  +initiatives: InitiativeRepository,
+|};
+
+// Adding below signature to help clarity of this commit.
+// TODO: @beanow will implement this in a follow-up.
+type _unused_loadDirectoryFunction = (InitiativesDirectory) => Promise<LoadedInitiativesDirectory>;
+
+/**
+ * Represents a single Initiative using a file as source.
+ *
+ * Note: The file name will be used to derive the InitiativeId. So it doesn't
+ * make sense to use this outside of the context of an InitiativesDirectory.
+ */
+export type InitiativeFile = {|
+  +title: string,
+  +timestampIso: ISOTimestamp,
+  +completed: boolean,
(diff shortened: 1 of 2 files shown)


src/plugins/initiatives/initiativesDirectory.test.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import {fromJSON, toJSON} from "./initiativesDirectory";
+
+const exampleInitiativeFile = () => ({
```
decentralion: nit: personally, I would make this a const and `deepFreeze` it. Change not requested (since it's not particularly better either way).

Also, I would add the explicit type signature, just as an added safety in case we add/remove fields. (Although there'd still be a type error at the toJSON call, that error would be less 'direct').

Beanow:
Updated to depend on #1646 / #1647 In this case not a big change, just updates the comments to remove tracker concept and refer to InitiativeIds instead.

src/plugins/initiatives/initiativesDirectory.test.js
Beanow: Added the return type. But will keep the generator function.
In a follow-up we'll have a generator besides this which takes arguments :]
Also find that generator functions make this very pleasant to read.
```js
// Given
const initiativeFile = exampleInitiativeFile();
```

Beanow:
Based on a synchronous chat with @decentralion, changed the timestamp to be an ISO date string instead. This will make files much easier to edit by hand.

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -0,0 +1,63 @@
...
+  +timestampIso: ISOTimestamp,
+  +completed: boolean,
+  +dependencies: $ReadOnlyArray<URL>,
+  +references: $ReadOnlyArray<URL>,
+  +contributions: $ReadOnlyArray<URL>,
+  +champions: $ReadOnlyArray<URL>,
+|};
+
+// Note: setting this to opaque forces us to convert it to timestampMs.
+opaque type ISOTimestamp = string;
```
decentralion: we may want to factor this out to a util module, along with typesafe methods for converting to-from a `MsSinceEpoch` representation.

src/plugins/initiatives/initiativesDirectory.test.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import {type InitiativeFile, fromJSON, toJSON} from "./initiativesDirectory";
+
+const exampleInitiativeFile = (): InitiativeFile => ({
+  title: "Sample initiative",
+  timestampIso: ("2020-01-08T22:01:57.766Z": any),
```
decentralion: the `any` cast here is a smell that suggests we should have a typechecked module for working with ISO timestamps

src/plugins/initiatives/initiativesDirectory.js
decentralion: also, this should probably be an opaque subtype of string.

`opaque type ISOTimestamp: string = string;`

Beanow: Part of the reason why I didn't subtype or make a util, is because I wanted to add additional pressure to convert to the internal convention of using `timestampMs` instead. This way the type is pretty unwieldy outside of this file. As the any cast in the tests show.

src/plugins/initiatives/initiativesDirectory.test.js
Beanow: Yeah, as mentioned it shows the type is not meant to be used outside the one file :]
For the test file we make an any-cast exception to match.

PR #1643, opened 2020-02-06
Initiatives: create and parse InitiativeFile addresses

Beanow:
Depends on #1646, #1647, #1641
Relates to #1640

The private functions `_initiativeFileId` will be used as a helper to
load a directory. The public function `initiativeFileURL` will be used
to add links to the remote file in the node description.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/initiativesDirectory.js
+++ b/src/plugins/initiatives/initiativesDirectory.js
@@ -1,8 +1,17 @@
 // @flow
 
 import {type ReferenceDetector} from "../../core/references";
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
 import {type Compatible, fromCompat, toCompat} from "../../util/compat";
-import {type InitiativeRepository, type URL} from "./initiative";
+import {initiativeNodeType} from "./declaration";
+import {
+  type InitiativeId,
+  type InitiativeRepository,
+  type URL,
+  createId,
+} from "./initiative";
+
+export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
 
 /**
  * Represents an Initiatives directory.
@@ -60,4 +69,32 @@ export function fromJSON(j: Compatible<any>): InitiativeFile {
 
 export function toJSON(m: InitiativeFile): Compatible<InitiativeFile> {
   return toCompat(COMPAT_INFO, m);
+}
+
+/**
+ * When provided with the initiative NodeAddressT of an InitiativeFile this extracts
+ * the URL from it. Or null when the address is not for an InitiativeFile.
+ */
+export function initiativeFileURL(address: NodeAddressT): string | null {
+  const initiativeFilePrefix = NodeAddress.append(
+    initiativeNodeType.prefix,
+    INITIATIVE_FILE_SUBTYPE
+  );
+
+  if (!NodeAddress.hasPrefix(address, initiativeFilePrefix)) {
+    return null;
+  }
+
+  const parts = NodeAddress.toParts(address);
+  const remoteUrl = parts[4];
+  const fileName = parts[5];
+  return `${remoteUrl}/${fileName}`;
+}
+
+// Creates the InitiativeId for an InitiativeFile.
+export function _initiativeFileId(
+  {remoteUrl}: InitiativesDirectory,
+  fileName: string
+): InitiativeId {
+  return createId(INITIATIVE_FILE_SUBTYPE, remoteUrl, fileName);
 }
--- a/src/plugins/initiatives/initiativesDirectory.test.js
+++ b/src/plugins/initiatives/initiativesDirectory.test.js
@@ -1,6 +1,15 @@
 // @flow
 
-import {type InitiativeFile, fromJSON, toJSON} from "./initiativesDirectory";
+import {NodeAddress} from "../../core/graph";
+import {createId, addressFromId} from "./initiative";
+import {
+  type InitiativeFile,
+  type InitiativesDirectory,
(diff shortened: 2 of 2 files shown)


src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -1,8 +1,17 @@
...
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
 import {type Compatible, fromCompat, toCompat} from "../../util/compat";
 import {type InitiativeRepository, type URL} from "./initiative";
+import {initiativeNodeType} from "./declaration";
+
+// Note: fictional tracker address, we won't create a node for the files.
+const initiativeFilePrefix = NodeAddress.fromParts([
+  "sourcecred",
+  "initiatives",
+  "directory",
```
decentralion: the convention we're using in other plugins is that the third component is the node type (e.g. `ISSUE` or `PULL_REQUEST`). since the node types with this prefix are individual initiatives, maybe we should be consistent with `INITIATIVE`. it's certainly not a directory node type.

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -60,3 +69,37 @@ export function fromJSON(j: Compatible<any>): InitiativeFile {
...
+// Creates the Initiative NodeAddressT for an InitiativeFile.
+export function _initiativeAddress(
+  remoteUrl: string,
+  fileName: string
+): NodeAddressT {
+  return NodeAddress.append(
+    initiativeNodeType.prefix,
+    ...NodeAddress.toParts(_trackerAddress(remoteUrl, fileName))
+  );
+}
```
decentralion: hmm, i'm confused. what is the difference between an initiative address and a tracker address, and why do we have both? If we do keep both around, the rationale should be documented in the code here, but my preference is to only have one address per initiative.

src/plugins/initiatives/initiativesDirectory.js
Beanow: Good idea, does make the address clearer. In this case it would be an `INITIATIVE_FILE`.

Beanow:
Refactored to depend on #1646, #1647. This removes the tracker concept, switching to InitiativeIds, so this also changes the helper functions.

src/plugins/initiatives/initiativesDirectory.js
Beanow: Now depending on #1647 to remove trackers.

src/plugins/initiatives/initiativesDirectory.js
Beanow: Defined the subtype now.
https://github.com/sourcecred/sourcecred/blob/c48c924a9f5fcf667d8219af80713d5c54d112fc/src/plugins/initiatives/initiativesDirectory.js#L14

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -61,3 +70,31 @@ export function fromJSON(j: Compatible<any>): InitiativeFile {
...
+ */
+export function initiativeFileURL(address: NodeAddressT): string | null {
+  const initiativeFilePrefix = NodeAddress.append(
+    initiativeNodeType.prefix,
+    INITIATIVE_FILE_SUBTYPE
+  );
+
+  if (!NodeAddress.hasPrefix(address, initiativeFilePrefix)) {
+    return null;
+  }
```
decentralion: Can you explain why we want this behavior, rather than throwing an error when we got a non-initiative address?

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -61,3 +70,31 @@ export function fromJSON(j: Compatible<any>): InitiativeFile {
...
+
+  const parts = NodeAddress.toParts(address);
+  const remoteUrl = parts[4];
+  const fileName = parts[5];
+  return `${remoteUrl}/${fileName}`;
+}
+
+// Creates the InitiativeId for an InitiativeFile.
+export function _initiativeFileId(
+  remoteUrl: string,
```
decentralion: nit: might be nice to take `InitiativeDirectory` object as the first argument, for more type safety / clarity

src/plugins/initiatives/initiativesDirectory.js
Beanow: Not throwing makes it really easy to compose this function should you want to.
An example usage is in our own `initiatives/createGraph` https://github.com/sourcecred/sourcecred/commit/3749a422bda92a2d631b7fc48773079e9f8118f5 (prototype)

As this is kind of a "reverse" reference detector, that might be interesting outside this module too.
https://github.com/sourcecred/sourcecred/blob/f924521fdd0761f448bc7463fcf91bf1ec59b444/src/core/references/referenceDetector.js#L18

However you're right in the context of purely converting InitiativeFile to Initiative, this should throw. And we do, in the follow-up PR #1648.
https://github.com/sourcecred/sourcecred/blob/f867f5d9d8d55fcc567286cd67c0d13c555abf0e/src/plugins/initiatives/initiativesDirectory.js#L195-L198

src/plugins/initiatives/initiativesDirectory.js
Beanow: No strong preference here. Done!

src/plugins/initiatives/initiativesDirectory.js
decentralion: Cool, thanks. I love that we have Flow to ensure that all clients will need to handle the case where it doesn't correspond to an initiative.

PR #1644, opened 2020-02-06
Initiatives: validate and read files from local directory

Beanow:
Depends on #1643 
Relates to #1640 

Helper functions intended to be used in succession by `loadDirectory`. Only `_validatePath` provides helfpul error messages. It's the caller's responsiblity to do this first.

Introduces dependency `globby` for globbing with a Promises API.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/initiativesDirectory.js
+++ b/src/plugins/initiatives/initiativesDirectory.js
@@ -1,8 +1,12 @@
 // @flow
 
+import path from "path";
+import fs from "fs-extra";
+import globby from "globby";
 import {type ReferenceDetector} from "../../core/references";
 import {type NodeAddressT, NodeAddress} from "../../core/graph";
 import {type Compatible, fromCompat, toCompat} from "../../util/compat";
+import {compatReader} from "../../backend/compatIO";
 import {initiativeNodeType} from "./declaration";
 import {
   type InitiativeId,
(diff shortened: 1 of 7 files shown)


src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -103,3 +106,47 @@ export function _initiativeAddress(
...
+    );
+  }
+  return absPath;
+}
+
+// Gets all *.json filenames in the given directory.
+export async function _findFiles(
+  localPath: string
+): Promise<$ReadOnlyArray<string>> {
+  const absoluteFileNames = await globby(path.join(localPath, "*.json"));
```
decentralion: globby supports `**` expansion. Why not do `**/*.json` and get subdirectory support from the start?

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -103,3 +106,47 @@ export function _initiativeAddress(
...
+
+// Reads all given filenames in the given directory, validating them as compat.
+export async function _readFiles(
+  localPath: string,
+  fileNames: $ReadOnlyArray<string>
+): Promise<NamesToInitiativeFiles> {
+  const map: NamesToInitiativeFiles = new Map();
+
+  // TODO(perf): can do this as a stream or in parallel.
+  const sortedFileNames = [...fileNames].sort();
```
decentralion: not that I mind, but why do we sort them?

Beanow:
Rebased so the added dependency won't conflict with #1645 changing the dependencies as well.

src/plugins/initiatives/initiativesDirectory.js
Beanow: The eventual type we'll arrive at is: `$ReadOnlyArray<Initiative>`.
This means it's an ordered type.

During prototyping I had several test flakes. So I thought, it would be good for reproducibility of loads as well to side with caution and sort them. Rather than sort them in tests only. As this order will be relevant in all parts of the system up to the WeightedGraph. Once in a graph it can be sorted by it's address.

src/plugins/initiatives/initiativesDirectory.js
Beanow: For now this is less complex. As technically we would need to represent this as an array in the ID / NodeAddress and let the separator be handled by `path.join` and `[].join("/")` for URLs.

Beanow:
Updated example files to use ISO date string, as added to #1641

src/plugins/initiatives/initiativesDirectory.js
decentralion: I don't think we need to represent it as an array in the id / NodeAddress. We could just have the full subpath be a component, as in `["sourcecred", "initiative", "INITIATIVE_FILE", "https://github.com/sourcecred/cred/master/", "path/to/my/file.json"]`.

src/plugins/initiatives/initiativesDirectory.js
decentralion: > The eventual type we'll arrive at is: $ReadOnlyArray<Initiative>.
This means it's an ordered type.

[] and ROA are both ordered. :)

> During prototyping I had several test flakes. So I thought, it would be good for reproducibility of loads as well to side with caution and sort them. Rather than sort them in tests only. As this order will be relevant in all parts of the system up to the WeightedGraph. Once in a graph it can be sorted by it's address.

Makes sense!

src/plugins/initiatives/initiativesDirectory.js
Beanow: And when your windows filesystem returns `path\to\my\file.json` instead? We're now responsible for normalizing filesystem paths and URLs :D

It can be done! But I'm just punting on it for now to make this a later improvement.

src/plugins/initiatives/initiativesDirectory.js
Beanow: Update: I've left a comment about this in the file.

Beanow:
Updated to make use of CompatIO's `compatReader` #1637

src/plugins/initiatives/initiativesDirectory.js
decentralion: Thanks, I appreciate the comment.

src/plugins/initiatives/initiativesDirectory.js
decentralion: Oh, I didn't realize that Windows has a different file path representation. Which is kind of embarrassing as I now see that I've [written bugs around backslash handling in the past](https://github.com/tensorflow/tensorboard/issues/51). Cool, let's revisit this later. :)

PR #1646, opened 2020-02-07
Initiatives: create InitiativeId concept

Beanow:
Previously we were relying on the `Initiative.tracker` to define the
address of an Initiative. Based on feedback at #1643 we want to remove
trackers. So we'll need a replacement ID.

This will enforce a new InitiativeId convention. As well as how to derive
a NodeAddressT from it. In a follow-up we'll remove the tracker concept.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/initiative.js
+++ b/src/plugins/initiatives/initiative.js
@@ -1,8 +1,25 @@
 // @flow
 
-import {type NodeAddressT} from "../../core/graph";
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
+import {initiativeNodeType} from "./declaration";
 
 export type URL = string;
+
+// Composite ID, used as input for NodeAddressT.
+export opaque type InitiativeId = string[];
+
+// Enforce that each ID has at least a sub-type and 1..n components.
+export function createId(
+  subType: string,
+  requiredId: string,
+  ...rest: string[]
+): InitiativeId {
+  return [subType, requiredId, ...rest];
+}
+
+export function addressFromId(id: InitiativeId): NodeAddressT {
+  return NodeAddress.append(initiativeNodeType.prefix, ...id);
+}
 
 /**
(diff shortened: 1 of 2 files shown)


src/plugins/initiatives/initiative.js
```diff
@@ -1,9 +1,26 @@
 // @flow
 
-import {type NodeAddressT} from "../../core/graph";
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
+import {initiativeNodeType} from "./declaration";
 
 export type URL = string;
 
+// Composite ID, used as input for NodeAddressT.
+export opaque type InitiativeId = string[];
```
decentralion: why not `$ReadOnlyArray<string>`? Though I admit it doesn't matter much when the type is opaque.

We might later want it to be an opaque subtype of $ROA<string>, but I don't know that we need to do that now.

src/plugins/initiatives/initiative.js
```diff
@@ -1,9 +1,26 @@
...
+
+// Enforce that each ID has at least a sub-type and 1..n components.
+export function createId(
```
decentralion: Per [this review](https://github.com/sourcecred/sourcecred/pull/1648#pullrequestreview-355519335), I'd like to see a method that constructs an Id from an InitiativesDirectory and a InitiativeMapEntry -- maybe it could go into this PR? (Could use this function as a helper.)

Beanow: I'll add it to the other PR :]

src/plugins/initiatives/initiative.js
Beanow: As you mentioned, it didn't occur to me because the opaque type doesn't make it necessary.
We have a grand total of 2 lines of code reading this array. :smile:
Happy to take a follow-up to change it though.

PR #1647, opened 2020-02-07
Initiatives: replace trackers with InitiativeIds

Beanow:
Depends on #1646 

"Trackers" were an idea to let Initiatives be aware of the medium that
declares it. Such as a Discourse topic or file.

With Discourse in mind this was really useful. We could add an automatic
contribution edge, enhance reference detection to "upgrade" a URL pointing
to that Topic to resolve to the Initiative instead, etc.

Using files as the only source of Initiatives this becomes less relevant.
So in the interest of reducing complexity, we'll remove tracker awareness.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/createGraph.test.js
+++ b/src/plugins/initiatives/createGraph.test.js
@@ -8,7 +8,7 @@ import {
 } from "../../core/graph";
 import type {ReferenceDetector, URL} from "../../core/references";
 import type {Initiative, InitiativeRepository} from "./initiative";
-import {topicAddress} from "../discourse/address";
+import {createId} from "./initiative";
 import {createGraph} from "./createGraph";
 import {
   initiativeNodeType,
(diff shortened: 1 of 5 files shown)


src/plugins/initiatives/createGraph.test.js
```diff
@@ -80,10 +80,11 @@ function exampleNodeAddress(id: number): NodeAddressT {
   return NodeAddress.fromParts(["example", String(id)]);
 }
 
-function discourseInitiativeAddress(id: number): NodeAddressT {
+function testInitiativeAddress(num: number): NodeAddressT {
   return NodeAddress.append(
     initiativeNodeType.prefix,
-    ...NodeAddress.toParts(topicAddress("https://example.com", id))
+    "TEST_SUBTYPE",
+    num.toString()
```
decentralion: this is a total nit so no change requested -- i personally like to use `String(x)` because it works more consistently across types ... consider that if `x == null` then `String(x)` is fine but `x.toString()` isn't.

Beanow: Agreed, generally better. Updated.
I got a bit complacent here because the flow signature was a non-nullable `number`.

PR #1648, opened 2020-02-07
Initiatives: implement conversion from InitiativeFiles

Beanow:
Depends on #1644

Helper functions intended to be used in succession by loadDirectory.
Only `_validateUrl` provides helfpul error messages. It's the caller's
responsiblity to do this first.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/initiativesDirectory.js
+++ b/src/plugins/initiatives/initiativesDirectory.js
@@ -9,10 +9,12 @@ import {type Compatible, fromCompat, toCompat} from "../../util/compat";
 import {compatReader} from "../../backend/compatIO";
 import {initiativeNodeType} from "./declaration";
 import {
+  type Initiative,
   type InitiativeId,
   type InitiativeRepository,
   type URL,
   createId,
+  addressFromId,
 } from "./initiative";
 
 export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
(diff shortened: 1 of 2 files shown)


src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -145,3 +147,56 @@ export async function _readFiles(
...
+    }
+    return url.toString().replace(/\/+$/, "");
+  } catch (e) {
+    throw new Error(
+      `Provided initiatives directory URL was invalid: ${remoteUrl}\n${e}`
+    );
+  }
+}
+
+// Takes a fileName, InitiativeFile map and converts it to Initiatives.
```
decentralion: nit: comment and argument names do not agree (fileName vs remoteUrl).
how about having it take a `InitiativesDirectory`? More descriptive at the type level, and then the reader can very clearly see what property we grab off it.

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -145,3 +147,56 @@ export async function _readFiles(
...
+export function _convertToInitiatives(
+  remoteUrl: string,
+  map: NamesToInitiativeFiles
+): $ReadOnlyArray<Initiative> {
+  const initiatives = [];
+  for (const [fileName, initiativeFile] of map.entries()) {
+    const {timestampIso, ...partialInitiativeFile} = initiativeFile;
+    const initiative: Initiative = {
+      ...partialInitiativeFile,
+      id: createId(INITIATIVE_FILE_SUBTYPE, remoteUrl, fileName),
```
decentralion: Hmm, I'm not very comfortable with this -- it buries the details on how we address the initiatives inside the implementation details of a helper function.

Please factor out a `initiativeId(initiativesDirectory, initiativeMapEntry)` style function that returns the id.

decentralion:
(Usage of the "request changes" button intended to be very gentle in this case -- it's a lightweight but important refactor.)

src/plugins/initiatives/initiativesDirectory.js
Beanow: Comment here wasn't intended to add a correct list of arguments. Updated it.
Also taking the `InitiativesDirectory` as argument.

src/plugins/initiatives/initiativesDirectory.js
Beanow: Good find. I had already factored this function out in #1643, now updated to use it.

src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -150,3 +152,56 @@ export async function _readFiles(
...
+    );
+  }
+}
+
+// Converts the InitiativeFiles we've read to Initiatives.
+export function _convertToInitiatives(
+  directory: InitiativesDirectory,
+  map: NamesToInitiativeFiles
+): $ReadOnlyArray<Initiative> {
+  const initiatives = [];
```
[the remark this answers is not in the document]
Beanow: For next time :]

PR #1649, opened 2020-02-07
Initiatives: implement loadDirectory

Beanow:
Depends on #1648

It's tests are primarily smoke tests, as the underlying helper functions
have been tested more extensively.

Test plan: `yarn test`


Beanow:
Rebased and updated, because `_convertToInitiatives` now takes an `InitiativesDirectory` instead of `remoteUrl: string` (see https://github.com/sourcecred/sourcecred/pull/1648#pullrequestreview-355519335).

PR #1654, opened 2020-02-09
Improve distributionToCred

decentralion:
This commit makes several small improvements to the distributionToCred
module:

- We rename the output `FullTimelineCred` data structure to
`TimelineCredScores`, which is more descriptive
- We re-organize that data structure so that rather than being an array
of `{interval, cred}` objects, it has an `intervals` property and a
`intervalCredScores` property, both of which are arrays. This will make
downstream usage cleaner.
- An unused variable is removed.
- We document invariants about the TimelineCredScores data type.
- We mark the TimelineCredScores data type opaque, so that clients
recieving a TimelineCredScores can trust that the invariants are
maintained.

Test plan:
- The rename is robustly tested by `yarn flow`.
- That the refactor lands without changing existing semantics is
robustly tested by `yarn test --full`, since we snapshot a full cred
load; thus we know that cred scores haven't changed. (Also, we have
existing unit tests).
- The newly documented invariants aren't robustly tested by the test
code, but it's easy to see that they hold by reading the algorithm.

### DIFF
--- a/src/core/algorithm/distributionToCred.js
+++ b/src/core/algorithm/distributionToCred.js
@@ -9,26 +9,41 @@ import {type Interval} from "../interval";
 import {type TimelineDistributions} from "./timelinePagerank";
 import {NodeAddress, type NodeAddressT} from "../../core/graph";
 
-/**
- * Represents the full timeline cred for a graph.
- */
-export type FullTimelineCred = $ReadOnlyArray<{|
-  // The interval for this slice.
-  +interval: Interval,
-  // The cred for each node.
-  // (Uses the graph's canonical node ordering.)
-  +cred: Float64Array,
-|}>;
+export opaque type NodeOrderedCredScores: Float64Array = Float64Array;
 
 /**
- * Convert a TimelineDistribution into TimelineCred.
+ * Represents cred scores over time.
+ *
+ * It contains an array of intervals, which give timing information, and an
+ * array of CredTimeSlices, which are Float64Arrays. Each CredTimeSlice
+ * contains cred scores for an interval. The cred scores are included in
+ * node-address-sorted order, and as such the CredScores can only be
+ * interpreted in the context of an associated Graph.
+ *
+ * As invariants, it is guaranteed that:
+ * - intervals and intervalCredScores will always have the same length
+ * - all of the intervalCredScores will have a consistent implicit node ordering
+ *
+ * The type is marked opaque so that no-one else can construct instances that
+ * don't conform to these invariants.
+ */
+export opaque type TimelineCredScores: {|
+  +intervals: $ReadOnlyArray<Interval>,
+  +intervalCredScores: $ReadOnlyArray<NodeOrderedCredScores>,
+|} = {|
+  +intervals: $ReadOnlyArray<Interval>,
+  +intervalCredScores: $ReadOnlyArray<NodeOrderedCredScores>,
+|};
+
+/**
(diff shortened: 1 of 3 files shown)


decentralion:
cc @wchargin; i'm channeling you by starting to write explicit invariants for joined data types
cc @vsoch because it's part of my push to cleanup the cred computation pipeline

Beanow:
>     * We re-organize that data structure so that rather than being an array
>       of `{interval, cred}` objects, it has an `intervals` property and a
>       `intervalCredScores` property, both of which are arrays. This will make
>       downstream usage cleaner.

Most of the time I find tuple/object arrays to be a lot less complex to iterate over. So the diff I can see in the tests don't look like an improvement to me. Can you point to which downstream usage you're talking about to get an idea of why we want to switch?

decentralion:
It matches the APIs we're already actually using. When we get the cred, we tend to want just cred score without the intervals:

https://github.com/sourcecred/sourcecred/blob/4a4c35bfdc4fadc0c9fb90489ae2386712773375/src/analysis/timeline/timelineCred.js#L33-L40

https://github.com/sourcecred/sourcecred/blob/4a4c35bfdc4fadc0c9fb90489ae2386712773375/src/analysis/timeline/timelineCred.js#L106-L122

Meanwhile when we get the intervals, we tend to just want the intervals on their own. 
https://github.com/sourcecred/sourcecred/blob/4a4c35bfdc4fadc0c9fb90489ae2386712773375/src/analysis/timeline/timelineCred.js#L99-L104

Beanow:
https://github.com/sourcecred/sourcecred/blob/4a4c35bfdc4fadc0c9fb90489ae2386712773375/src/analysis/timeline/timelineCred.js#L33-L40

Is a good example of this imo :]

As an (exported) type, it would be clearer to use:

```js
// The timeline sequence of cred (one score per interval). 
+intervalCred: $ReadOnlyArray<{score: number, interval: Interval}>,
```

However, it would be very space-inefficient to copy the intervals for every node. So :+1:

src/core/algorithm/distributionToCred.js
```diff
@@ -10,25 +10,38 @@ import {type TimelineDistributions} from "./timelinePagerank";
...
-export type FullTimelineCred = $ReadOnlyArray<{|
-  // The interval for this slice.
-  +interval: Interval,
-  // The cred for each node.
-  // (Uses the graph's canonical node ordering.)
-  +cred: Float64Array,
-|}>;
+export opaque type TimelineCredScores: {|
+  +intervals: $ReadOnlyArray<Interval>,
+  +intervalCredScores: $ReadOnlyArray<Float64Array>,
```
Beanow: Rather than explaining what the `$ReadOnlyArray<Float64Array>` is in the comments, I prefer to separate it to another alias.

```js
opaque type NodeOrderedCredScores = Float64Array;

{+intervalCredScores: $ReadOnlyArray<NodeOrderedCredScores>, ...}
```

_Note: would have nit-and-approved, but since this PR's purpose is to improve the API I'll add as comment._

decentralion: Would have preferred a nit-and-approve still so I can merge off that approval if I adopt your comment. 

I think adding a comment makes sense when asking a question, or when you would need to see the changed approach to know if you'd want to approve it. In this case, since you're proposing an explicit change outright, it feels sensible to just approve it.

PR #1658, opened 2020-02-09
Remove plugins from TimelineCred

### LINKED ISSUE #1557: Refactor

decentralion:
This commit removes any dependency on plugins from the TimelineCred
module. Instead, TimelineCred just needs the array of "scoring node
prefixes", i.e. the prefix addresses for all nodes that we will use to
normalize cred totals.

This is not the end of my cred-pipeline refactors (and I actually intend
to eventually remove the TimelineCred module entirely), but it is a nice
waypoint, since we can now close #1557.


src/backend/loadContext.js
```diff
@@ -111,9 +111,11 @@ export class LoadContext {
...
     }
     const plugins = this._declarations(this._pluginLoaders, project);
+    const userTypes = [].concat(...plugins.map((x) => x.userTypes));
+    const scoringNodePrefixes = userTypes.map((x) => x.prefix);
     const cred = await this._computeTask(this._compute, this._options, {
       params,
-      plugins,
+      scoringNodePrefixes,
       weightedGraph,
     });
```
Beanow: This functionality is not part of LoadContext's responsibility of composition. It's shouldn't "know" how to convert between plugin declarations and `scoringNodePrefixes`, it's just passing variables around and calling things in order.

We should either defer this to a new function `(PluginDeclaration[]) => NodeAddressT[]` which LoadContext can call, or change the signature of some existing one. For example, we're not using declarations for anything else. So `PluginLoaders.declarations()` could be replaced by `PluginLoaders.scoringNodePrefixes()`.

src/backend/loadContext.test.js
```diff
@@ -1,19 +1,46 @@
...
+  description: "a foo",
+};
+const plugin: PluginDeclaration = deepFreeze({
+  name: "foo",
+  nodePrefix: NodeAddress.empty,
+  edgePrefix: EdgeAddress.empty,
+  nodeTypes: [userType, fooType],
+  edgeTypes: [],
+  userTypes: [userType],
+});
```
Beanow: Likewise, having specific values instead of fakes, suggest we're doing logic.
LoadContext isn't supposed to know how to do these things, so it should be enough to be given fakes.

Beanow:
With a small change I think this is still worth merging, in spite of #1686
Only needs to factor out a `scoringNodePrefixes: (PluginDeclaration[]) => NodeAddressT[]`

Want to take a stab at that? @decentralion

Beanow:
@decentralion as you're currently refactoring other TimelineCred parts. Can you check whether you want to finish or close this?

decentralion:
@Beanow this got blocked on all the LoadContext stuff, I found it difficult / tedious to make the changes there. This isn't high priority for me (and is unrelated to the other TimelineCred changes I'm making), so I'll just close it for now-- if you want to take it over the finish line by fixing up the LoadContext, I'd be happy to see it merged.

PR #1407, opened 2019-10-11
Expand the blacklist, found new type inconsistencies

Beanow:
- Bots being Users as a commit author
- Orgs being Users on a reaction

Repositories affected:
- prettier/prettier
- lovell/sharp
- facebook/jest
- babel/babel-eslint
- recharts/recharts
- webpack-contrib/css-loader
- yannickcr/eslint-plugin-react
- vuejs/vuex
- chimurai/http-proxy-middleware
- sass/node-sass
- lodash/lodash
- vuejs/vue
- reacttraining/react-router
- axios/axios
- webpack/webpack-dev-middleware
- eslint/eslint
- webpack/webpack
- webpack/webpack-cli
- sinonjs/sinon
- neutrinojs/webpack-chain
- webpack/webpack-dev-server

Found as part of https://github.com/teamopen-dev/sourcecred-stack-lookup

Test after this patch: pending, it's a lot of data after the cache invalidated :sweat_smile:

PR #1660, opened 2020-02-13
github: standardize blacklist URL comments

wchargin:
Summary:
Most of the blacklisted reactions helpfully link to their original
source, but some don’t. This patch adds the missing links.

Test Plan:
The following command now prints a URL on every line:

```
<src/plugins/github/blacklistedObjectIds.js \
awk '/^[^ ]/ { p = 0 }; p { gsub(".*// ", ""); print }; /reactions/ { p = 1 }'
```

wchargin-branch: blacklist-urls


Beanow:
> This patch adds the missing links.

See what you did there :3

Beanow:
Good point. I have some of those orgs associated with a checklist...
https://github.com/sourcecred/sourcecred/pull/1407

But it'll take a bit of patience to get those 1:1.
Can run that sometime if useful.

PR #1661, opened 2020-02-13
adding jq and updates to container base

vsoch:
This pull request will serve two purposes, both related to updating the sourcecred/sourcecred:dev image:

 - add an apt-get update so that we install security updates. While there is a tradeoff between running updates and build time, I think the 20 seconds or so are worth it in favor of security
 - adding jq to the base container. This is a relatively small package, and will allow for speeding up of the [cred-action](https://github.com/vsoch/cred-action) as it won't be required to run apt-get update and then install jq - we can just grab the base container and it's ready to go!

I mentioned this in https://github.com/sourcecred/cred/pull/17

Signed-off-by: vsoch <vsochat@stanford.edu>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,9 @@ FROM node:12
 RUN mkdir -p /code
 WORKDIR /code
 
+# Updates and install jq for downstream actions
+RUN apt-get update && apt-get install -y jq
+
 # Install global and local dependencies first so they can be cached.
 RUN npm install -gf yarn@^1.21.1
 COPY package.json yarn.lock /code/


vsoch:
For some reason the tests are running on my good-labs branch (and they obviously will fail without the GitHub token, and that's also why the test isn't linked or easy to find).

https://app.circleci.com/jobs/github/good-labs/sourcecred/389

How is this managed given that someone is likely to open a PR from a different username / org? If the tests are going to fail given running on someone else's branch, I think the circle setting should be unchecked to (disable) running tests on PR forks.

vsoch:
To review this, we really only need to note that the container builds. The sourcecred code isn't touched, so I think the publish-1 test is sufficient for passing.

wchargin:
> How is this managed given that someone is likely to open a PR from a
> different username / org?

From <https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#running-full-tests-on-ci>:

> As soon as you open your pull request, our CI will start running the
> basic test suite (`yarn test --ci`). It will also try to run the full
> test suite (`yarn test --ci --full`), but this will fail for PRs created
> from forks because it depends on credentials that aren’t exposed to
> untrusted PRs by default. This is expected—don’t worry!
> 
> Once a core team member sanity-checks your PR to make sure that it’s not
> accidentally leaking credentials into logs, they can “bless” your commit
> by pushing it to any branch on the main SourceCred repository. This will
> restart the full test suite, which will now actually run. Once your
> tests pass and the PR is approved, we’ll delete the extra branch.

I’ll go ahead and do this for you.

vsoch:
Ah interesting - I wonder if there is a way to alert the contributor about this, given that the PR is from another place? What about a github workflow that can comment back to the PR with some "never fear, help is on the way!" message?

wchargin:
We should have that via #1522; let me know if it isn’t working.

vsoch:
I think the issue is that it fails but it’s not able to report back a status update. From my end the status was yellow so I was waiting around for it, and finally after what seemed like too long I investigated (and note that there was no direct link from here I had to open up circle separately) and only then did I see the failure. The check still was awaiting status, so the typical contributor would likely continue waiting or just be confused. I think in addition to needing a fast fail there also needs to be a direct and clear message in a comment here to the contributor.

vsoch:
Oh, and thank you for the blessing! That's a fun way to put it :)

vsoch:
Definitely - do you have some recommended reading for the jq issues (I've seen it used widely around and wasn't aware). I'll test on the cred-action repository and then follow up here.

Beanow:
>     * add an apt-get update so that we install security updates

I feel like I might be misinterpreting this. `apt-get update` doesn't in itself install things. But would include security updates of jq-and-friends when installing that. Is that what you meant?
Or did you mean to include an `apt-get upgrade` here?

It makes sense to me to update/upgrade in the Dockerfile. It's also why we're using `node:12` and not `node:x.y.z`.

---

My familiarity with `jq` is pretty low so I have no idea how to interpret `jq '.'`.
What does it do, and why are we doing that?

`jq --raw-output '.[] | .head.ref` seems straightforward, plucking a particular value because we need that in the shell script.

Should we want to replicate that with other tools, I do suggest using the node equivalent of the suggested python scrips, as it will save a significant amount in image size to not include python :]

vsoch:
I'm testing a PR now to remove jq and use the python3 json.tool and vanilla python - will update here and close the PR here when I can confirm it's good :)

wchargin:
> My familiarity with `jq` is pretty low so I have no idea how to interpret `jq '.'`.

The `.` argument is the identity filter, so it just passes the input through unchanged, except for pretty-printing it:

```
$ printf '{"one":2,"three":4}' | jq .
{
  "one": 2,
  "three": 4
}
```

> Should we want to replicate that with other tools, I do suggest using
> the node equivalent of the suggested python scrips, as it will save a
> significant amount in image size to not include python :]

I looked into this briefly and couldn’t see a simple way to read stdin
without an external package (…). Not attached to Python at all; would be
happy to replace this with Node if there’s a simple way to do so.

Beanow:
> > 
> 
> I looked into this briefly and couldn’t see a simple way to read stdin
> without an external package (…). Not attached to Python at all; would be
> happy to replace this with Node if there’s a simple way to do so.

One way is
https://github.com/sourcecred/widgets/blob/805c59b6b2d850a131e0d1db687ebabef1f6786e/bin/contributor-wall-svg.js#L42

Had some bugs with that though: https://github.com/sourcecred/widgets/issues/14

vsoch:
Arg it almost finished, but I forgot that the action ran over night (so the branch already existed for the day and the push was rejected.) So... need to run again!

vsoch:
okay need to run it again - I should have checked the payload, the formatting is incorrect. Be back in another long time, lol.

vsoch:
okay we're good:

 - https://github.com/vsoch/cred-action/pull/2
 - https://github.com/vsoch/sourcecred-cred/pull/5

Closing here, since we don't need jq.

PR #1672, opened 2020-02-22
Initiatives: pass SOURCECRED_INITIATIVES_DIRECTORY to LoadContext

Beanow:
Depends on: #1671

This represent the local path which we didn't include in the Project.

Test plan: `yarn test`


Beanow:
> As we work on the instance system we may want to change / clean this up... but fine for now.

Agreed, the distinctions between canonical config, tweaking config, env dependent config and weaving them through the load pipeline is overall pretty messy :sweat_smile: 

But the instance system could allow a fixed directory structure as well making this particular parameter unnecessary.

PR #1673, opened 2020-02-22
Expose a ReferenceDetector to createPluginGraphs

Beanow:
Eventually all plugins are expected to use the ReferenceDetector. This commit composes the Github and Discourse detectors we can create, for now exposing it as unused to `createPluginGraphs`.

Test plan: `yarn test`

### DIFF
--- a/src/backend/pluginLoaders.js
+++ b/src/backend/pluginLoaders.js
@@ -10,6 +10,10 @@ import {type GithubToken} from "../plugins/github/token";
 import {type Loader as GithubLoader} from "../plugins/github/loader";
 import {type Loader as IdentityLoader} from "../plugins/identity/loader";
 import {type Loader as DiscourseLoader} from "../plugins/discourse/loader";
+import {
+  type ReferenceDetector,
+  CascadingReferenceDetector,
+} from "../core/references";
 
 /**
  * A type combining all known plugin Loader interfaces.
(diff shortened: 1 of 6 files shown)


src/backend/pluginLoaders.js
```diff
@@ -97,13 +101,41 @@ export async function updateMirror(
...
+/**
+ * Creates a ReferenceDetector composing all plugin reference detectors
+ * requested by the project.
+ */
+export async function createReferenceDetector(
+  {github, discourse}: $Shape<PluginLoaders>,
+  {githubToken}: GraphEnv,
+  {cache, project}: CachedProject
+): Promise<ReferenceDetector> {
+  const refs = [];
```
decentralion: Nit: would be nice to have an explicit typesig on refs

src/backend/pluginLoaders.test.js
```diff
@@ -271,6 +289,35 @@ describe("src/backend/pluginLoaders", () => {
...
+        {githubToken},
+        cachedProject
+      );
+
+      // Then
+      expect(references).toBeInstanceOf(CascadingReferenceDetector);
+      expect(((references: any): CascadingReferenceDetector).refs).toEqual([
+        fakes.githubReferences,
+        fakes.discourseReferences,
+      ]);
```
decentralion: aside: i'm skeptical that tests like this add any value. given that we aren't using the underlying logic, just essentially checking that pieces were piped around correctly, what does this get us that flow doesn't?

Beanow: Flow would not validate whether we've correctly interpreted the input Project to compose the detectors. Because these are all valid type-wise, but only the last is correct logic-wise (given the Project of this test):
- `new CascadingReferenceDetector([])`
- `new CascadingReferenceDetector([githubRefs])`
- `new CascadingReferenceDetector([discourseRefs])`
- `new CascadingReferenceDetector([githubRefs, discourseRefs])`

Will agree that it's not a durable test, but a smoke test is better than no test. As it would not cover any new plugins with reference detectors as they're added. Happy to improve if you have suggestions.

src/backend/pluginLoaders.js
Beanow: Done!

PR #1674, opened 2020-02-22
Initiatives: define weights to allow Cred minting

Beanow:
Using a required tuple of before and after completion weight is a simple
way to start minting Cred on Initiatives. It sets expectations by having
both states defined in a version controlled file.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/createGraph.js
+++ b/src/plugins/initiatives/createGraph.js
@@ -1,7 +1,6 @@
 // @flow
 
 import {
-  Graph,
   EdgeAddress,
   NodeAddress,
   type Edge,
@@ -9,6 +8,9 @@ import {
   type EdgeAddressT,
   type NodeAddressT,
 } from "../../core/graph";
+import {type WeightedGraph as WeightedGraphT} from "../../core/weightedGraph";
+import * as WeightedGraph from "../../core/weightedGraph";
+import type {NodeWeight} from "../../core/weights";
 import type {ReferenceDetector, URL} from "../../core/references";
 import type {Initiative, InitiativeRepository} from "./initiative";
 import {addressFromId} from "./initiative";
@@ -33,6 +35,13 @@ function initiativeNode(initiative: Initiative): Node {
     description:
       url == null ? initiative.title : `[${initiative.title}](${url})`,
   };
+}
+
+export function initiativeWeight(initiative: Initiative): ?NodeWeight {
+  if (!initiative.weight) return;
+  return initiative.completed
+    ? initiative.weight.complete
+    : initiative.weight.incomplete;
 }
 
 type EdgeFactoryT = (initiative: Initiative, other: NodeAddressT) => Edge;
@@ -63,15 +72,21 @@ const referenceEdge = edgeFactory(referencesEdgeType.prefix, true);
 const contributionEdge = edgeFactory(contributesToEdgeType.prefix, false);
 const championEdge = edgeFactory(championsEdgeType.prefix, false);
 
-export function createGraph(
+export function createWeightedGraph(
   repo: InitiativeRepository,
   refs: ReferenceDetector
-): Graph {
-  const graph = new Graph();
+): WeightedGraphT {
+  const wg = WeightedGraph.empty();
+  const {graph, weights} = wg;
 
   for (const initiative of repo.initiatives()) {
     // Adds the Initiative node.
-    graph.addNode(initiativeNode(initiative));
+    const node = initiativeNode(initiative);
+    const weight = initiativeWeight(initiative);
+    graph.addNode(node);
+    if (weight) {
+      weights.nodeWeights.set(node.address, weight);
+    }
 
     // Generic approach to adding edges when the reference detector has a hit.
     const edgeHandler = (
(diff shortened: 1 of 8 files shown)


src/plugins/initiatives/createGraph.js
```diff
@@ -31,6 +33,11 @@ function initiativeNode(initiative: Initiative): Node {
   };
 }
 
+export function initiativeWeight(initiative: Initiative): ?NodeWeight {
+  if (!initiative.weight) return;
+  return !initiative.completed ? initiative.weight[0] : initiative.weight[1];
+}
```
decentralion: Why do we want this to be nullable? Why not just require that all initiatives have weights?

src/plugins/initiatives/initiative.js
```diff
@@ -39,6 +40,8 @@ export type Initiative = {|
   +id: InitiativeId,
   +title: string,
   +timestampMs: number,
+  // This weight tuple indicates a before completion and after completion weight.
+  +weight?: [NodeWeight, NodeWeight],
```
decentralion: rather than using a tuple, let's use a type:

`type InitiativeWeight = {whenCompleted: NodeWeight, whenNotCompleted: NodeWeight}`

It's much more self documenting. Also, let's make it required rather than optional.

src/plugins/initiatives/createGraph.js
Beanow: To me this seems like the most expressive "internal type". A default value exists that is the equivalent of `null`, namely: `[1, 1]`, but it's surprising, adds needless data and 1x multiplications.

For sake of helping the user "do the right thing©" we do require weights to be set in the `InitiativeFile` type.

Or you could say,  I consider `Initiative` to be correct in this regard, vs `InitiativeFile` which requires weights due to being strongly coupled to our community usage patterns.

src/plugins/initiatives/initiative.js
Beanow: Sure! Using `{incomplete, complete}` then :]

As mentioned above, my preference would be to leave requiring at the `InitiativeFile` type but not on `Initiative`.

Beanow: Done!

src/plugins/initiatives/createGraph.js
```diff
@@ -35,6 +37,12 @@ function initiativeNode(initiative: Initiative): Node {
   };
 }
 
+export function initiativeWeight(initiative: Initiative): ?NodeWeight {
+  if (!initiative.weight) return;
+  const {incomplete, complete} = initiative.weight;
+  return !initiative.completed ? incomplete : complete;
+}
+
```
decentralion: Ok, I can see the case for letting the initiative.weight be null (in which case default); in that case, shouldn't this function contain the "default-setting" behavior?

I think the default weights should be `{incomplete: 0, complete: 1}`, which we could easily implement in this function.

src/plugins/initiatives/createGraph.js
```diff
@@ -35,6 +37,12 @@ function initiativeNode(initiative: Initiative): Node {
   };
 }
 
+export function initiativeWeight(initiative: Initiative): ?NodeWeight {
+  if (!initiative.weight) return;
+  const {incomplete, complete} = initiative.weight;
+  return !initiative.completed ? incomplete : complete;
```
decentralion: nit: can be simplified as `return initiative.completed ? complete : incomplete`

src/plugins/initiatives/createGraph.js
Beanow: Imo, default setting does not belong (at all) in this function. This is simply a conversion from `?InitiativeWeight` to `?NodeWeight`.

The result of `null` from an algorithm point of view should be to fall back on the node type weight. Which you can achieve in one of two ways:
A) With a noop multiplication `{incomplete: 1, complete: 1}`.
B) Not setting a weight at the individual node address at all.
This type simply allows you to choose both approaches, and prefers B as it's less expensive.

The value you're suggesting should be a configuration option that we don't currently support. What you really want is for the PluginDeclaration to set `{incomplete: 0, complete: 1}` as the default and allow overriding it at the node type level. I say that because "we don't give cred to incomplete initiatives" is a sane default, but definitely debatable. Other communities might want `{incomplete: 0.25, complete: 8}`.

For now, the _workaround_ for not having this flexibility is to require all the JSON files have weights.

src/plugins/initiatives/createGraph.js
Beanow: Ah that just happened from trying to keep with the convention of using the chronological order, `incomplete` before `complete`. Which doesn't provide any cognitive load benefit here so I'll update this.

Beanow: Same applies to shortening the variables so going with:
```js
return initiative.completed
  ? initiative.weight.complete
  : initiative.weight.incomplete;
```

Beanow: Done!

src/plugins/initiatives/createGraph.js
Beanow: As I feel this discussion is more a configuration discussion than a technical issue with this PR, going to merge as-is :slightly_smiling_face:

PR #1681, opened 2020-02-29
Modify default Discourse weights

decentralion:
This commit modifies the default weights in the Discourse plugin. The
overall theme is to make the plugin flow less cred to "raw activity", in
favor of only flowing cred to posts where there is some explicit signal
that they were valuable.

Most significantly, we move over to fully [like-minted cred][1], instead
of minting cred directly to posts and topics. Also, we remove the edges
that tend to flow cred to posts indiscriminately. For example, topics no
longer flow cred to every post within the topic.

Based on local testing on a few forums I'm familiar with, I feel
confident that these cred scores are an improvement over the current
defaults, as we now have a few "real life test cases" of high-noise,
high-activity users, and these weights reduce the amount of cred that
accrues to such "stress testers". With these changes, I feel that we can
start cautiously using the Discourse plugin in [Trust Level 2][2]
communities.

Test plan: `yarn flow` and code inspection are sufficient to verify that
the new weights are technically valid. Because my calibration process
for validating these changes involves subjective judgements about
contributions from real people, I'm declining to publicly post any
specifics; reviewers are welcome to reach out to my offline for further
discussion.

[1]: https://discourse.sourcecred.io/t/minting-discourse-cred-on-likes-not-posts/603
[2]: https://github.com/sourcecred/docs/pull/24

### DIFF
--- a/src/plugins/discourse/declaration.js
+++ b/src/plugins/discourse/declaration.js
@@ -12,7 +12,7 @@ export const topicNodeType: NodeType = deepFreeze({
   name: "Topic",
   pluralName: "Topics",
   prefix: NodeAddress.append(nodePrefix, "topic"),
-  defaultWeight: 2,
+  defaultWeight: 0,
   description:
     "A topic (or post-container) in a Discourse instance. Every topic has at least one post.",
 });
@@ -21,7 +21,7 @@ export const postNodeType: NodeType = deepFreeze({
   name: "Post",
   pluralName: "Posts",
   prefix: NodeAddress.append(nodePrefix, "post"),
-  defaultWeight: 1,
+  defaultWeight: 0,
   description: "A post in some topic in a Discourse instance.",
 });
 
@@ -45,7 +45,7 @@ export const topicContainsPostEdgeType: EdgeType = deepFreeze({
   forwardName: "contains post",
   backwardName: "is contained by topic",
   prefix: EdgeAddress.append(edgePrefix, "topicContainsPost"),
-  defaultWeight: {forwards: 1 / 16, backwards: 1 / 4},
+  defaultWeight: {forwards: 0, backwards: 1 / 8},
   description: "Connects a topic to the posts that it contains.",
 });
 
@@ -53,7 +53,7 @@ export const postRepliesEdgeType: EdgeType = deepFreeze({
   forwardName: "post is reply to",
   backwardName: "post replied to by",
   prefix: EdgeAddress.append(edgePrefix, "replyTo"),
-  defaultWeight: {forwards: 1, backwards: 1 / 16},
+  defaultWeight: {forwards: 1 / 2, backwards: 0},
   description: "Connects a post to the post that it is a reply to.",
 });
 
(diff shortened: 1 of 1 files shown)


decentralion:
> These aren’t exactly the same as the weights used for SourceCred itself
(we set node weight for likes to 16 rather than 4), but that’s okay with
me.

Yeah, my main goal here was to remove all the activity-minted cred, and make the edge weights more spam-resilient. If Discourse is used in isolation, then the absolute cred minting per like is irrelevant; and I plan to come up with "cross plugin relative weight norms" later (after #1682).

Beanow:
> Most of the edge weight changes mitigate posts accruing “un-reviewed”
> cred by linking to lots of high-cred things. It sounds enough reasonable
> to me that I don’t mind trying it out, though I do like the philosophy
> of the backward-edge behavior. Would be interested in thinking about how
> to get the best of all worlds.

TL;DR
I understood the intention, but to review I would need to have some indication of whether it would do what it's intended to do. It's too complex for me to judge the implications as-is. Too much uncertainty for me to even estimate whether "it looks like an improvement to me" from the author is causal or coincidental.

If I was asked to review: should we run this as an experiment? I could have given an answer.
(It would be: an experiment would make sense to me in our own instance rather than master.)

Since that's not the question, I'm going with the PR is under-explained in this regard.

---

For example:
- A lot of learning algorithms require sufficient entropy to avoid local maxima/minima. Does SourceCred benefit from a similar notion, esp. when taking a holistic view, with the contributors' behavior as part of the algorithm? And would flowing cred back to a larger context (like post -> topic -> forum -> project) provide this entropy?
- Does the lack of back-edges give more control to contributors to shape their graph, opening up to new gaming options? Perhaps making it actually less robust for TL2?
- Would having a lot more one-way edges increase the average amount of "hops" needed for cred to arrive at valuable nodes? And would the alpha parameter dilute those valuable nodes? Or would it simply take more iterations to arrive at a similar stable distribution?
- Does this change include a trade-off on the axis of [social validation <-> being "discovered" by the algorithm]? And have we got at least a theory where the sweet spot would be?
- Would this change be extra detrimental to "leaf nodes" and cause new contributors to be stuck in a cred-quarantine state until their first social validation?

etc. etc. Worst of all, a lot of this seems like it can't be considered in isolation.

decentralion:
@Beanow, I appreciate your pushback and think you raise reasonable points. The underlying issue here is that right now, we don't have tools or processes for assessing the quality of cred scores, and thus we're flying blind when making changes to them. 

This means that we should invest in cred analysis tooling, in repeatable experiments we can run on cred quality, etc. However, I don't think we should block changes to weights on developing this tooling. 

The current discourse weights were chosen with very little forethought (the PR that added them (https://github.com/sourcecred/sourcecred/pull/1292) had 0 discussion of weight choices, and was just focused on [unblocking the CredSperiment](https://discourse.sourcecred.io/t/discourse-from-day-1/185). These updated weights are more grounded, since they came about from a practical need to mitigate certain issues we're seeing in practice. We should avoid status quo bias, especially in cases where the status quo came about from a mostly random/arbitrary process.

Therefore, I'm merging this change. Thanks @wchargin and @Beanow for reviews.

PR #1685, opened 2020-03-01
github: use plaintext owner and name in cache key

wchargin:
Summary:
This is mostly a QoL improvement for maintainers. GraphQL mirrors are
stored in the SourceCred cache directory, but until now have been hard
to tell apart. All IDs looked like long, similar hex strings; e.g.:

```
github_736f75726365637265642d746573742f6578616d706c652d676974687562.db
github_736f75726365637265642f646f6373.db
github_736f75726365637265642f736f7572636563726564.db
```

It’s too hard to tell that these are `sourcecred-test/example-github`,
`sourcecred/docs`, and `sourcecred/sourcecred` in that order. (Yes, you
can work it out if you try; that’s not good enough.) With dozens of
caches loaded, finding the right one to poke at for debugging or
progress-checking takes way too much scripting.

The purpose of this abstruse encoding was portable filename safety, but
we can actually achieve this with human-readable names like these:

```
github_sourcecred-test_example_github.db
github_sourcecred_docs.db
github_sourcecred_sourcecred.db
```

This suffices because (a) login names cannot contain underscores, so we
can safely use that as a separator, and (b) GitHub disallows collisions
on names that are equal ignoring case, so we can convert all names to
lowercase without introducing collisions.

This change orphans existing caches. Running `sourcecred clear --cache`
will clean those up.

Test Plan:
Try to create a GitHub repository with the same name as a repository
that you already have, but with inverted case; note that this is
disallowed. Then, note that loading `sourcecred-test/example-github`
still works and produces a database with nicely readable name.

wchargin-branch: github-readable-db-names

### DIFF
--- a/src/plugins/github/cacheId.js
+++ b/src/plugins/github/cacheId.js
@@ -31,7 +31,7 @@ export function cacheIdForRepoId(repoId: RepoId): CacheId {
   // only `[0-9A-Za-z_.-]` and so is valid on all major filesystems.
   const owner = repoId.owner.toLowerCase();
   const name = repoId.name.toLowerCase();
-  if (owner.indexOf("_") !== -1) {
+  if (owner.includes("_")) {
     throw new Error(
       "unexpected underscore in GitHub owner name would be ambiguous: " +
         repoIdToString(repoId)
--- a/src/plugins/github/cacheId.js
+++ b/src/plugins/github/cacheId.js
@@ -9,14 +9,33 @@ export opaque type CacheId: string = string;
 
 /**
  * Derives the CacheId for a RepoId.
- * Returned CacheId's will be:
+ *
+ * Returned `CacheId`s will be:
  * - Deterministic
  * - Unique for this plugin
  * - Lowercase
  * - Safe to use for filenames
+ * - Distinct for semantically distinct inputs (input IDs that differ
+ *   only in case may map to the same output, because GitHub does not
+ *   permit collisions-modulo-case)
  */
 export function cacheIdForRepoId(repoId: RepoId): CacheId {
-  const repoString = repoIdToString(repoId);
-  const repoStringHex = Buffer.from(repoString).toString("hex");
-  return `github_${repoStringHex}`.toLowerCase();
+  // GitHub owner (user/organization) and repository names may be in
+  // mixed case, but GitHub prevents mixed-case collisions: e.g., if
+  // `foo` is a user then `Foo` cannot also be a user, and likewise for
+  // repositories. GitHub login names are DNS-safe (`[0-9A-Za-z-]` only)
+  // and so are filename-safe. Repository names have a slightly larger
+  // character set, including underscore, but are also filename-safe.
+  // Because login may not contain an underscore, it thus suffices to
+  // use an underscore as the delimiter. The resulting filename uses
+  // only `[0-9A-Za-z_.-]` and so is valid on all major filesystems.
+  const owner = repoId.owner.toLowerCase();
+  const name = repoId.name.toLowerCase();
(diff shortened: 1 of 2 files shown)


Beanow:
I agree with the rationale, but will add one property for the record.

> The purpose of this abstruse encoding was portable filename safety, but
> we can actually achieve this with human-readable names like these:

You've payed attention to collisions, but want to emphasize that's a requirement.
No combination should have different owner+name inputs, but the same cache key.

Relying on the fact that underscores are not allowed in the owner field is one way to approach this.
It may be unexpected that this is a character that doesn't cause collisions though. In the same boat would be `.` as that's also not allowed in the owner field.

I'm really not satisfied with any particular option in this regard. All the other special character are either hard to read or somehow have usability issues in CLIs. And `_` and `.` will rely on this property of not being allowed in owner.

- `github_foo bar_baz` :neutral_face:
- `github_foo;bar_baz` :neutral_face:
- `github_foo,bar_baz` :neutral_face:
- `github_foo+bar_baz` :neutral_face:
- `github_foo|bar_baz` :neutral_face:
- `github_foo.bar.baz` :neutral_face: (owner: foo, name: bar.baz)
- `github_foo_bar_baz` :neutral_face: (owner: foo, name: bar_baz)

So with all that said, I'm ok with choosing `_` as the lesser evil here.

src/plugins/github/cacheId.js
```diff
@@ -9,14 +9,33 @@ export opaque type CacheId: string = string;
...
+  // `foo` is a user then `Foo` cannot also be a user, and likewise for
+  // repositories. GitHub login names are DNS-safe (`[0-9A-Za-z-]` only)
+  // and so are filename-safe. Repository names have a slightly larger
+  // character set, including underscore, but are also filename-safe.
+  // Because login may not contain an underscore, it thus suffices to
+  // use an underscore as the delimiter. The resulting filename uses
+  // only `[0-9A-Za-z_.-]` and so is valid on all major filesystems.
+  const owner = repoId.owner.toLowerCase();
+  const name = repoId.name.toLowerCase();
+  if (owner.indexOf("_") !== -1) {
```
Beanow: Nit: prefer [`String.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)

src/plugins/github/cacheId.js
```diff
@@ -9,14 +9,33 @@ export opaque type CacheId: string = string;
...
+    throw new Error(
+      "unexpected underscore in GitHub owner name would be ambiguous: " +
+        repoIdToString(repoId)
+    );
+  }
```
Beanow: Nit: This is technically redundant, as `RepoId` is an opaque type which already tests for the absence of underscores in it's factory function. In other instances we also rely on this to enforce certain properties that aren't available in Flow types.

For example: minimum array length.
https://github.com/sourcecred/sourcecred/blob/09739512ab14dc7a7a6825ffe069f671a02ad7d3/src/plugins/initiatives/initiative.js#L9-L19

src/plugins/github/cacheId.js
wchargin: Done; thanks!

src/plugins/github/cacheId.js
wchargin: Yep; I commented as much on the test case (lines 69–71). It pays to be
extra defensive here because we’re constructing a cache key, so silent
collisions would be quite bad. This protects against the case where we
change the `RepoId` invariants and forget to update this caller. The
check is basically free since this code is rarely called, so I think
that it’s worth it.

Issue #1679, opened 2020-02-28
Release v0.5.0 🚀


Beanow:
![SourceCred logo](https://raw.githubusercontent.com/sourcecred/sourcecred/master/src/assets/logo/rasterized/logo_128.png) _versioned release tracker_

# Summary

Drawing from https://discourse.sourcecred.io/t/summary-of-planning-chat/300
v0.5.0 is sliced when the Initiatives plugin completes.

# Deliverables

- [x] [Load refactor](https://discourse.sourcecred.io/t/refactoring-the-load-system/629)
- [x] Initiatives plugin can use [Unified reference detection](https://discourse.sourcecred.io/t/unified-reference-detection/301)
- [x] [Initiatives plugin](https://discourse.sourcecred.io/t/write-the-initiatives-plugin/269)
- [x] [Support entries in Initiatives](https://github.com/sourcecred/sourcecred/issues/1739)

# Highlights since v0.4.0

New features:
- [Added Discourse support](https://discourse.sourcecred.io/t/create-discourse-plugin/660) (+[reference detection](https://discourse.sourcecred.io/t/discourse-reference-mention-detection/270), +[anonymous access](https://github.com/sourcecred/sourcecred/pull/1431), +[mirror revision](https://discourse.sourcecred.io/t/discourse-mirror-revision/365))
- [Added Identity support](https://discourse.sourcecred.io/t/create-identity-plugin/662)
- [Added Initiatives support](https://discourse.sourcecred.io/t/write-the-initiatives-plugin/269)
- [Published as a Docker image](https://discourse.sourcecred.io/t/publish-sourcecred-as-a-docker-image/661)
- [New data format available: `soucrecred output`](https://github.com/sourcecred/sourcecred/issues/1773)

Enhancements:
- [Minting Discourse Cred on Likes](https://discourse.sourcecred.io/t/discourse-like-minted-cred/663)
- [GraphQL Mirror EAV refactor](https://discourse.sourcecred.io/t/graphql-mirror-eav-refactor/272)
- [GraphQL Mirror fidelity awareness](https://discourse.sourcecred.io/t/graphql-mirror-fidelity-awareness/275)
- [Keep a network log in the GraphQL mirror](https://github.com/sourcecred/sourcecred/issues/1450) (PR #1562)
- [WeightedGraph; Enable plugins to set weights](https://discourse.sourcecred.io/t/weightedgraph-enable-plugins-to-set-weights/664)
[...]

PR #1692, opened 2020-03-06
Chore: make all minor-version package upgrades

Beanow:
Preparing for release #1679

Note: due to a regression, not upgrading eslint-plugin-react
See yannickcr/eslint-plugin-react#2570

Also updated package.json to latest semver in-range versions.
Note, this changes all packages (other than eslint-plugin-react)
to ^x.x.x format.

Test plan:
- `yarn test --full` to see nothing broke
- `yarn outdated` to see only major version upgrades are remaining


Beanow:
I ran into this regression again: https://github.com/yannickcr/eslint-plugin-react/issues/2570
As such, reverting back to `"eslint-plugin-react": "7.16.0"` for now.

decentralion:
Why don't we upgrade the versions in package.json too? I think greenkeeper has a tool to help do this.

Beanow:
That is the default behavior for in-range semver updates for both yarn and npm.
I'll have a look to force update them all instead, though the lock file should work.

PR #1694, opened 2020-03-06
Chore: update major versions that aren't breaking for us

Beanow:
I'm taking a pass here on all things frontend. Because they're not in test coverage, they _do_ break the build, and don't belong in this repository to begin with :]

Not updated:
- history
- react-dev-utils
- react-router
- recharts
- whatwg-fetch
- tmp


Beanow:
Pinning tmp, as it seems to have an undocumented breaking change or regression in v0.1.0.

PR #1695, opened 2020-03-07
Util: implement sortBy

### LINKED ISSUE #1488: Smaller alternative to lodash.sortBy
We had chat about `lodash.sortby` https://github.com/sourcecred/sourcecred/pull/1464#discussion_r348660981

> Hm, reason I tend to avoid this is because it's a [2630 line dependency](https://github.com/lodash/lodash/blob/4.7.0-npm-packages/lodash.sortby/index.js) to abstract away a 3 line comparison function and shallow copy instead of mutate.

> For `lodash.sortby` I'd like to have a go at replacing it with a small unit function. But since this is definitely not the only place where it would need to be used, I think that is best done as a separate PR. For now I feel this solution is way more comprehensible than the dependency.

There's only a few properties that we need from this function.
- The readability of having a "pluck" function `(x) => x.fieldToSortBy`.
- Creating a shallow copy so the input array isn't mutated.
- [Being simple](https://www.youtube.com/watch?v=rI8tNMsozo0) in implementation.

Beanow:
This replaces the complex `lodash.sortby` dependency.
Doing what we need it for in just a few lines.

Fixes #1488

Note: one major difference here is we're now more aware of the types being used for comparison.
I'll make some inline comments to point out cases where I needed to filter out nulls for example.

Test plan: `yarn test --full`

Suggested entry point for review: `src/util/sortBy.js`

### DIFF
--- a/src/analysis/timeline/timelineCred.js
+++ b/src/analysis/timeline/timelineCred.js
@@ -1,7 +1,7 @@
 // @flow
 
 import {sum} from "d3-array";
-import sortBy from "lodash.sortby";
+import sortBy from "../../util/sortBy";
 import * as NullUtil from "../../util/null";
 import * as MapUtil from "../../util/map";
 import {toCompat, fromCompat, type Compatible} from "../../util/compat";
@@ -136,7 +136,9 @@ export class TimelineCred {
       const match = (a) => prefixes.some((p) => NodeAddress.hasPrefix(a, p));
       addresses = addresses.filter(match);
     }
-    const credNodes = addresses.map((a) => this.credNode(a));
+    const credNodes = NullUtil.filterList(
+      addresses.map((a) => this.credNode(a))
+    );
     return sortBy(credNodes, (x: CredNode) => -x.total);
   }
 
--- a/src/core/interval.js
+++ b/src/core/interval.js
@@ -1,7 +1,7 @@
 // @flow
 
 import {max, min} from "d3-array";
-import sortBy from "lodash.sortby";
+import sortBy from "../util/sortBy";
 import {utcWeek} from "d3-time";
 import * as NullUtil from "../util/null";
 import type {Node, Edge, Graph} from "./graph";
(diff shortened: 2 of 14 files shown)


src/analysis/timeline/timelineCred.js
```diff
@@ -136,7 +136,9 @@ export class TimelineCred {
       const match = (a) => prefixes.some((p) => NodeAddress.hasPrefix(a, p));
       addresses = addresses.filter(match);
     }
-    const credNodes = addresses.map((a) => this.credNode(a));
+    const credNodes = NullUtil.filterList(
+      addresses.map((a) => this.credNode(a))
+    );
```
Beanow: Example that needed a null filter.

src/core/interval.js
```diff
@@ -40,7 +42,9 @@ export type GraphIntervalPartition = $ReadOnlyArray<GraphInterval>;
  */
 export function partitionGraph(graph: Graph): GraphIntervalPartition {
   const nodes = Array.from(graph.nodes());
-  const timefulNodes = nodes.filter((x) => x.timestampMs != null);
+  const timefulNodes: $ReadOnlyArray<TimefulNode> = (nodes.filter(
+    (x) => x.timestampMs != null
+  ): any);
```
Beanow: This one was tricky, as the base type of `Node[]` has a `timestampMs: number | null`. And null isn't supported for comparisons.

We're aware of this, which is why we're doing this `nodes.filter()`... but flow doesn't automatically enrich the type here to not have a non-null `timestampMs`.

src/util/sortBy.test.js
```diff
@@ -0,0 +1,138 @@
+// @flow
+
+import sortBy from "./sortBy";
+
+describe("util/sortBy", () => {
+  const numericSample = [4, 2, 1, 8, 16];
```
decentralion: nit: I would deepFreeze these examples so they can't be modified by test code.

Beanow: done

PR #1697, opened 2020-03-08
add initial Grain events and ledger

decentralion:
This commit adds a GrainEvent type, which represents any valid event in
the Grain ledger. Currently, distributions (minting grain) and transfers
are supported. We expect to add more events later (like
boosts/sponsorship).

Given a sequence of GrainEvents, we can construct a Ledger, which keeps
track of the balances and earnings for each actor, and allows querying
over the events. Currently, we just have an InMemoryLedger, but we may
expect more implementations in the future.

For now, I've kept the interface simple and synchronous, since in the
near-term we will be loading all events in memory. In the future, we may
refactor to an async/generator approach.

We track all grainholders by their node address.

Test plan: Inspect unit tests; `yarn test` passes.

### DIFF
--- a/src/grain/grain.js
+++ b/src/grain/grain.js
@@ -139,3 +139,15 @@ export function multiplyFloat(grain: Grain, num: number): Grain {
 export function fromApproximateFloat(f: number): Grain {
   return multiplyFloat(ONE, f);
 }
+
+/**
+ * Approximates the division of two grain values
+ *
+ * This naive implementation of grain division converts the given values
+ * to floats and performs simple floating point division.
+ *
+ * Do not assume this will be precise!
+ */
+export function toFloatRatio(numerator: Grain, denominator: Grain): number {
+  return Number(numerator) / Number(denominator);
+}
--- a/src/grain/ledger.js
+++ b/src/grain/ledger.js
@@ -22,6 +22,7 @@ export type TransferV1 = {|
   +recipient: NodeAddressT,
   +amount: Grain,
   +timestampMs: number,
+  +memo: string,
 |};
 
 export type LedgerEvent = DistributionV1 | TransferV1;
--- a/src/grain/ledger.js
+++ b/src/grain/ledger.js
@@ -50,7 +50,11 @@ export interface Ledger {
   events(): $ReadOnlyArray<LedgerEvent>;
 }
 
-export class InMemoryLedger implements Ledger {
+export function inMemoryLedger(events: $ReadOnlyArray<LedgerEvent>): Ledger {
+  return new _InMemoryLedger(events);
+}
+
+class _InMemoryLedger implements Ledger {
   _balances: Map<NodeAddressT, Grain>;
   _earnings: Map<NodeAddressT, Grain>;
   _events: $ReadOnlyArray<LedgerEvent>;
--- a/src/grain/ledger.js
+++ b/src/grain/ledger.js
@@ -1,0 +1,146 @@
+// @flow
+
+import {type NodeAddressT, NodeAddress} from "../core/graph";
+import stringify from "json-stable-stringify";
+import * as NullUtil from "../util/null";
+import {
+  type Grain,
+  format as formatGrain,
+  ZERO,
+  DECIMAL_PRECISION,
+} from "./grain";
+import {type DistributionV1} from "./distribution";
+
+/**
+ * Models a transfer of grain from a sender to recipient.
+ * A grainholder may transfer grain to themself (which is a no-op).
+ */
+export type TransferV1 = {|
+  +type: "TRANSFER",
+  +version: number,
+  +sender: NodeAddressT,
+  +recipient: NodeAddressT,
+  +amount: Grain,
+  +timestampMs: number,
+|};
+
+export type LedgerEvent = DistributionV1 | TransferV1;
+
+/**
+ * A ledger tracks balances, earnings, and the event history of participants.
+ *
+ * We may refactor the `events` to be an async generator in the future,
+ * when we stop storing all the events in memory.
+ */
+export interface Ledger {
+  /**
+   * Stores the current grain balance of each address.
+   */
+  balances(): Map<NodeAddressT, Grain>;
+  /**
+   * Stores the lifetime earnings of each address.
+   *
+   * Necessary for computing which participants have been underpaid.
+   */
+  earnings(): Map<NodeAddressT, Grain>;
+  /**
+   * Retrieve the events.
+   */
+  events(): $ReadOnlyArray<LedgerEvent>;
+}
+
+export class InMemoryLedger implements Ledger {
+  _balances: Map<NodeAddressT, Grain>;
+  _earnings: Map<NodeAddressT, Grain>;
+  _events: $ReadOnlyArray<LedgerEvent>;
+
+  /**
+   * Construct a Ledger from an array of in-memory events.
+   *
+   * The events must be in timestamp sorted order.
+   */
+  constructor(events: $ReadOnlyArray<LedgerEvent>) {
+    this._balances = new Map();
+    this._earnings = new Map();
+    this._events = events;
+    let lastTimestamp = -Infinity;
+    for (const e of events) {
+      if (e.timestampMs < lastTimestamp) {
+        throw new Error(
+          `event timestamps out of order: encountered ${e.timestampMs} after ${lastTimestamp}`
+        );
+      }
+      lastTimestamp = e.timestampMs;
+      switch (e.type) {
+        case "DISTRIBUTION":
+          this._processDistribution((e: DistributionV1));
+          break;
+        case "TRANSFER":
+          this._processTransfer((e: TransferV1));
+          break;
+        default:
+          throw new Error(`Unsupported event type: ${(e.type: empty)}`);
+      }
+    }
+  }
+
+  events(): $ReadOnlyArray<LedgerEvent> {
+    return [...this._events];
+  }
+
+  balances(): Map<NodeAddressT, Grain> {
+    return new Map(this._balances);
+  }
+
+  _balance(a: NodeAddressT): Grain {
+    return NullUtil.orElse(this._balances.get(a), ZERO);
+  }
+
+  _earning(a: NodeAddressT): Grain {
+    return NullUtil.orElse(this._earnings.get(a), ZERO);
+  }
+
+  earnings(): Map<NodeAddressT, Grain> {
+    return new Map(this._earnings);
+  }
+
+  _processDistribution(d: DistributionV1) {
+    const {version, receipts} = d;
+    if (version !== 1) {
+      throw new Error(`Unsupported distribution version: ${version}`);
+    }
+    for (const {amount, address} of receipts) {
+      const balance = this._balance(address) + amount;
+      this._balances.set(address, balance);
+      const earned = this._earning(address) + amount;
+      this._earnings.set(address, earned);
+    }
+  }
+
+  _processTransfer(t: TransferV1) {
+    const {recipient, sender, amount, version, timestampMs} = t;
+    if (version !== 1) {
+      throw new Error(`Unsupported transfer version: ${t.version}`);
+    }
+    const recipientBalance = this._balance(recipient);
+    const senderBalance = this._balance(sender);
+    if (senderBalance < amount) {
+      const forDisplay = {
+        amount: formatGrain(amount, DECIMAL_PRECISION),
+        recipient: NodeAddress.toString(recipient),
+        sender: NodeAddress.toString(sender),
+        timestampMs: timestampMs,
+      };
+      throw new Error(
+        `Invalid transfer (sender can't afford): ${stringify(forDisplay)}`
+      );
+    }
+    if (sender === recipient) {
+      // No need to do any update if the sender and recipient are the same.
+      // (And using the logic below would cause grain to "vanish")
+      return;
+    }
+    this._balances.set(recipient, recipientBalance + amount);
+    this._balances.set(sender, senderBalance - amount);
+  }
+}
--- a/src/grain/distribution.js
+++ b/src/grain/distribution.js
@@ -1,0 +1,226 @@
+// @flow
+
+/**
+ * In SourceCred, projects regularly distribute Grain to contributors based on
+ * their Cred scores. This is called a "Distribution".
+ *
+ * This module contains the logic for calculating distribution amounts for
+ * contributors. A distribution contains "receipts" showing how much Grain each
+ * contributor will receive, the timestamp of the distribution in question, and the
+ * "strategy" used to distribute grain.
+ *
+ * Currently we support two strategies:
+ * - IMMEDIATE, which distributes a fixed budget of grain based on the cred scores
+ * in the most recent completed time interval
+ * - LIFETIME, which distributes a fixed budget of grain based on cred scores
+ * across
+ * all time, prioritizing paying people who were under-paid historically (i.e.
+ * their lifetime earnings are lower than we would expect given their current
+ * cred score)
+ *
+ * In all cases, the timestamp for the distribution is used to determine which cred
+ * scores are in scope. For example, if you create a immediate distribution with a
+ * timestamp in the past, it will reward people with cred in the past time
+ * period, not the current one.
+ */
+import {sum} from "d3-array";
+import {mapToArray} from "../util/map";
+import {type NodeAddressT} from "../core/graph";
+import {type Grain, multiplyFloat, ZERO} from "./grain";
+
+export const DISTRIBUTION_VERSION_1 = 1;
+
+export type DistributionStrategy = ImmediateV1 | LifetimeV1;
+
+export type ImmediateV1 = {|
+  +type: "IMMEDIATE",
+  +version: number,
+  +budget: Grain,
+|};
+
+export type LifetimeV1 = {|
+  +type: "LIFETIME",
+  +version: number,
+  +budget: Grain,
+|};
+
+export type GrainReceipt = {|
+  +address: NodeAddressT,
+  +amount: Grain,
+|};
+
+export type DistributionV1 = {|
+  +type: "DISTRIBUTION",
+  +timestampMs: number,
+  +version: number,
+  +strategy: DistributionStrategy,
+  +receipts: $ReadOnlyArray<GrainReceipt>,
+|};
+
+export type CredTimeSlice = {|
+  +intervalEndMs: number,
+  +cred: Map<NodeAddressT, number>,
+|};
+
+export type CredHistory = $ReadOnlyArray<CredTimeSlice>;
+
+/**
+ * Compute a full Distribution given:
+ * - the strategy we're using
+ * - the full cred history for all users
+ * - the lifetime earnings of all users
+ * - the timestamp for the distribution
+ */
+export function distribution(
+  strategy: DistributionStrategy,
+  credHistory: CredHistory,
+  earnings: Map<NodeAddressT, Grain>,
+  timestampMs: number
+): DistributionV1 {
+  const filteredSlices = credHistory.filter(
+    (s) => s.intervalEndMs <= timestampMs
+  );
+  if (!filteredSlices.length) {
+    return {
+      type: "DISTRIBUTION",
+      timestampMs,
+      version: DISTRIBUTION_VERSION_1,
+      strategy,
+      receipts: [],
+    };
+  }
+
+  const receipts: $ReadOnlyArray<GrainReceipt> = (function () {
+    switch (strategy.type) {
+      case "IMMEDIATE":
+        if (strategy.version !== 1) {
+          throw new Error(
+            `Unsupported IMMEDIATE strategy: ${strategy.version}`
+          );
+        }
+        const lastSlice = filteredSlices[filteredSlices.length - 1];
+        return computeImmediateReceipts(strategy.budget, lastSlice.cred);
+      case "LIFETIME":
+        if (strategy.version !== 1) {
+          throw new Error(`Unsupported LIFETIME strategy: ${strategy.version}`);
+        }
+        const totalCred = new Map();
+        for (const {cred} of filteredSlices) {
+          for (const [address, ownCred] of cred.entries()) {
+            const existingCred = totalCred.get(address) || 0;
+            totalCred.set(address, existingCred + ownCred);
+          }
+        }
+        return computeLifetimeReceipts(strategy.budget, totalCred, earnings);
+      default:
+        throw new Error(`Unexpected type ${(strategy.type: empty)}`);
+    }
+  })();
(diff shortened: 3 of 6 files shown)


src/grain/ledger.js
```diff
@@ -0,0 +1,109 @@
...
+  constructor(history: LedgerHistory) {
+    this._history = history;
+    this._balances = new Map();
+    this._earnings = new Map();
+    for (const d of history.distributions) {
+      this._processDistribution(d);
+    }
+    for (const t of history.transfers) {
+      this._processTransfer(t);
+    }
```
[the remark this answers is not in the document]
decentralion: The history (in particular transactions) needs to be in timestamp order, I can add an invariant checker to validate that. However, this code (where we process all distributions and then all transfers) is fine. Note the problem case you showed is one where we processed some transfers before all the distributions were processed.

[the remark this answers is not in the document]
Beanow: Correct, I'm claiming the current code would produce a false-positive if this was the provided history. It *should* fail on this illegal transfer.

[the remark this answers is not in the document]
decentralion: Oh, you're right. Consider me persuaded, I'll re-write it per your suggestions.

src/grain/ledger.js
```diff
@@ -0,0 +1,161 @@
...
+
+  canonicalAddress(a: NodeAddressT): NodeAddressT {
+    return NullUtil.orElse(this._history.aliases.get(a), a);
+  }
+
+  _processDistribution(d: Distribution) {
+    const recipient = this.canonicalAddress(d.recipient);
+    const balance = this._balance(recipient) + d.amount;
+    this._balances.set(recipient, balance);
+    const earned = NullUtil.orElse(this._earnings.get(recipient), zero);
```
Beanow: Nit: earnings also has this `get() || 0` helper function. Should use for consistency.

src/grain/ledger.js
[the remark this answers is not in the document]
Beanow: This part is addressed.

src/grain/ledger.js
```diff
@@ -0,0 +1,109 @@
...
+|};
+
+export interface Ledger {
+  balances(): Map<NodeAddressT, Grain>;
+  earnings(): Map<NodeAddressT, Grain>;
+  history(): LedgerHistory;
+}
+
+// $ExpectFlowError
+const zero: Grain = 0n;
```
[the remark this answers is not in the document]
decentralion: Done.

src/grain/ledger.js
```diff
@@ -0,0 +1,109 @@
...
+export type AliasSpec = {|
+  +canonical: NodeAddressT,
+  +alias: NodeAddressT,
+|};
+
+export type LedgerHistory = {|
+  +transfers: $ReadOnlyArray<GrainTransfer>,
+  +distributions: $ReadOnlyArray<GrainDistribution>,
+  +aliases: Map<NodeAddressT, NodeAddressT>,
+|};
```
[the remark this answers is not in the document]
decentralion: Marking resolved since I did something similar (`LedgerEvent = Distribution | Transfer`)

src/grain/ledger.js
```diff
@@ -0,0 +1,161 @@
+// @flow
+
+import deepFreeze from "deep-freeze";
+import {type NodeAddressT, NodeAddress} from "../core/graph";
+import stringify from "json-stable-stringify";
+import * as NullUtil from "../util/null";
+import {type Grain, format as formatGrain} from "./grain";
+
+export type LedgerEvent = Distribution | Transfer;
```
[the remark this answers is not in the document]
decentralion: Done.

src/grain/ledger.js
```diff
@@ -0,0 +1,161 @@
...
+  +amount: Grain,
+  +timestampMs: number,
+|};
+
+export type Distribution = {|
+  +type: "DISTRIBUTION",
+  +recipient: NodeAddressT,
+  +amount: Grain,
+  +timestampMs: number,
+|};
```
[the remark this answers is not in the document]
decentralion: I tried to find a way to represent the "common ground" but couldn't get anything to work. Maybe @wchargin would have some ideas.

src/grain/ledger.js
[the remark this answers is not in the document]
decentralion: Done. I've added a string version (e.g. `0.1.0`) for consistency with versioning across the rest of the codebase.

src/grain/ledger.js
decentralion: Can't find it. I assume this comment is outdated.

src/grain/ledger.js
```diff
@@ -0,0 +1,109 @@
...
+
+export type LedgerHistory = {|
+  +transfers: $ReadOnlyArray<GrainTransfer>,
+  +distributions: $ReadOnlyArray<GrainDistribution>,
+  +aliases: Map<NodeAddressT, NodeAddressT>,
+|};
+
+export interface Ledger {
+  balances(): Map<NodeAddressT, Grain>;
+  earnings(): Map<NodeAddressT, Grain>;
```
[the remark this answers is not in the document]
decentralion: Earnings is a particularly useful thing to cache because when computing distributions, we need to know the lifetime earnings for every contributor--that way we can compare their lifetime payout to their total cred.

src/grain/ledger.js
Beanow: I mean
```js
  _earning(a: NodeAddressT): Grain {
    return NullUtil.orElse(this._earnings.get(a), zero);
  }
```

So this line can be `const earned = this._earning(recipient);`
Same as the balance above.

src/grain/ledger.js
```diff
@@ -0,0 +1,109 @@
...
+  history(): LedgerHistory;
+}
```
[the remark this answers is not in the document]
decentralion: I think this makes sense, but for now the only implementation is one that loads all the events into memory. I'd like to punt on making this async until we actually have a need for it. I don't think the refactoring costs will be too high, so I want to opt for velocity instead.

src/grain/ledger.js
decentralion: Got it. Fixed.

decentralion:
@Beanow: I've addressed your comments. I also simplified the design by taking the `aliases` out of the LedgerHistory (and removing LedgerHistory entirely), so Ledger now just operates on directly on the event stream. We can add support for aliases by modeling it as a transformer over the underlying events, thus separating concerns.

src/grain/ledger.js
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Construct a Ledger from an array of in-memory events.
+   *
+   * The events must be in timestamp sorted order.
+   */
+  constructor(events: $ReadOnlyArray<LedgerEvent>) {
+    this._balances = new Map();
+    this._earnings = new Map();
+    this._events = events;
+    let lastTimestamp = -Infinity;
```
Beanow: When are we updating this timestamp?

src/grain/ledger.js
```diff
@@ -0,0 +1,220 @@
...
+      switch (e.type) {
+        case "DISTRIBUTION":
+          this._processDistribution((e: Distribution));
+          break;
+        case "TRANSFER":
+          this._processTransfer((e: Transfer));
+          break;
+        default:
+          throw new Error(`Unsupported event type: ${(e.type: empty)}`);
+      }
```
Beanow: Nit: generally think logic heavy constructors are a smell.
Here I would think it's a small improvement to have one of
```js
// Static (preferably pure) factory function
InMemoryLedger.from = ($ReadOnlyArray<LedgerEvent>) => InMemoryLedger

// Append / replay method to process events
const ledger = new InMemoryLedger();
ledger.replay(events);
```

src/grain/ledger.js
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }
+
+  events(options: ?$Shape<EventsOptions>): $ReadOnlyArray<LedgerEvent> {
```
Beanow: Please use an optional argument instead of nullable argument `(options?: $Shape<EventsOptions>)`.

I think `thing.method()` communicates "default options" better than `thing.method(null)` or `thing.method({})`.

src/grain/ledger.js
```diff
@@ -0,0 +1,220 @@
...
+    const fullOptions = {
+      ...DEFAULT_EVENTS_OPTIONS,
+      ...NullUtil.orElse(options, {}),
+    };
+    const {address, eventOrder, eventSort} = fullOptions;
+    let filter = (_) => true;
```
Beanow: Nit: for perf we could make this nullable instead. So we can do.
```js
const events = filter ? this._events.filter(filter) : this._events;
```

I'm assuming the compiler doesn't detect noop filters on $ReadOnlyArrays to skip them.

src/grain/ledger.js
```diff
@@ -0,0 +1,220 @@
...
+    if (address) {
+      filter = (x: LedgerEvent) => {
+        switch (x.type) {
+          case "DISTRIBUTION":
+            return x.recipient === address;
+          case "TRANSFER":
+            return x.recipient === address || x.sender === address;
```
Beanow: Nit: looking at this check, do we need the switch statement?
The filter seem also a correct for distributions, `x.sender` would be `undefined` and therefore not equal to the not-null `address`. Not sure how exactly Flow wants you to refine union types though.

src/grain/ledger.js
```diff
@@ -0,0 +1,220 @@
...
+          default:
+            throw new Error((x.type: empty));
+        }
+      };
+    }
+    const events = this._events.filter(filter);
```
Beanow: Nit: I think it's worth commenting that this variable is expected to be a shallow copy.
In the case we're not calling `sortBy()`, but do call `.reverse()` it would be updated in-place.

src/grain/ledger.js
Beanow: Actually, if this turns out to be a bug, I think that would be an argument for
https://eslint.org/docs/rules/prefer-const#suggest-using-const-prefer-const

Beanow:
@decentralion updating the identities docs, realized we should probably support a kind of "account contraction" even type for the grain ledger as well, as rewriting history (updating addresses in-place, pretending it was one account) is an absolute no-go for ledgers. I think would be best as a follow-up?

src/grain/ledger.js
```diff
@@ -0,0 +1,146 @@
...
+ * Models a transfer of grain from a sender to recipient.
+ * A grainholder may transfer grain to themself (which is a no-op).
+ */
+export type TransferV1 = {|
+  +type: "TRANSFER",
+  +version: number,
+  +sender: NodeAddressT,
+  +recipient: NodeAddressT,
+  +amount: Grain,
+  +timestampMs: number,
```
hammadj, line 24: I think it would be beneficial to have an optional field for arbitrary data / memo. Will be helpful if we want to add a related link or on-chain txID or any other message for context on what that transfer was for. Can just be a string.

hammadj: Thinking of AraCred, one way to do the on-chain bridge would be to have an "Ethereum" node address, and when someone wants to take some Grain on chain, it would just be modelled as a Transfer from their grain balance to the "Ethereum" node's grain balance. That data field would then contain the txID of minting that grain on chain.

The grain is then free to travel anywhere on chain, be bought and sold, etc, and then whoever wants to bring it back in the grain ledger, it would just be modelled as a Transfer from the "Ethereum" node address to that users node address, with the data field containing the txID of burning that token on chain.

src/grain/ledger.js
```diff
@@ -0,0 +1,146 @@
...
+|};
+
+export type LedgerEvent = DistributionV1 | TransferV1;
```
hammadj, line 27: I'm wondering if "Distribution" actually belongs as a ledger event vs having only the lower level functions of "Mint" and "Transfer", which is a lot more generic and composable. The "Distribution" would essentially be a set of "Mint" events for specific node addresses. Not sure if the ledger needs to / should "know" about distributions, it seems like the only thing that would matter would be the list of "Receipts" aka minting grain.

Conceptually as well, "Distribution" is not a "ledger event" as much as it is a "collection of ledger events", vs its very obvious that a "Transfer" is a ledger event.

src/grain/ledger.js
decentralion: I've removed the query methods on the ledger, to save them for when they're actually motivated by a use case.

src/grain/ledger.js
decentralion: I've removed the query methods on the ledger, I figure it will be simpler to write them when we know exactly what API we need.

src/grain/ledger.js
decentralion: I've removed the query methods on the ledger, I figure it will be simpler to write them when we know exactly what API we need.

src/grain/ledger.js
decentralion: I've removed the query methods on the ledger, I figure it will be simpler to write them when we know exactly what API we need.

src/grain/ledger.js
decentralion: Yeah, it was a bug. Fixed, and I think that lint rule is a fine idea.

src/grain/grain.js
```diff
@@ -0,0 +1,110 @@
...
+ * around this by lying to Flow and claiming that the BigInts are actually
+ * numbers. Whenever we actually construct a BigInt, we will
+ * suppress the flow error at that declaration. Mixing BigInts with other types
+ * (e.g. 5n + 3) produces a runtime error, so even though Flow will not save
+ * us from these, they will be easy to detect. See [facebook/flow#6639][flow]
+ *
+ * [BigInt]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+ * [flow]: https://github.com/facebook/flow/issues/6639
+ */
+export type Grain = number;
```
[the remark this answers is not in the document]
decentralion: No longer included in this PR.

src/grain/ledger.js
decentralion: Yeah, I agree. This would serve the same need as [references](https://github.com/sourcecred/cred/blob/aed8b172d3e3b4fa7a61a6cad13100513e5ca9ff/transfers.json#L32) in our current ad-hoc system

src/grain/ledger.js
decentralion: I thought about this. The issue is for the UI that we're going to build this for, we're going to want to be able to find all the distributions for a user that gave them grain. So if we don't keep that info in the ledger, then we need to track two ledgers: the grain balance ledger, and the distribution history ledger. At which point we risk them de-synchronizing, so we might want to bind them more tightly by giving every LedgerEvent an ID and referencing the distribution IDs from the distribution history. But adding IDs adds even more complexity.

Given that this is a "private" data format to SourceCred, and that that the simpler ledger is derivable from a transformation on this ledger version, we also could always upgrade into that format later (and just provide a migration script). 

Thus, I'm inclined to keep it as is. Happy to hear alternative perspectives, though.

decentralion:
> we should probably support a kind of "account contraction" event type for the grain ledger as well, as rewriting history (updating addresses in-place, pretending it was one account) is an absolute no-go for ledgers. I think would be best as a follow-up?

Yeah, I originally put it in the ledger logic itself, but I realized it would be a lot simpler to decouple it by having the identity transformation occur as a transformation on the ledger events before they are seen by the ledger itself.

src/grain/ledger.js
decentralion: I've heard this a lot, I'm not really sure what the problem is with having logic in the constructor though. From an implementation perspective, I imagine re-writing the Ledger so the constructor takes an opaque bundle of fully-computed arguments, or doing your replay approach. But the replay approach invites questions about whether it's OK to replay multiple sequences of events on top of each other, which needs more testing, so overall the change feels a bit annoying.

If you feel strongly about it I can make one of these changes, but I'm kind of inclined to just merge as-is.

Beanow:
> Yeah, I originally put it in the ledger logic itself, but I realized it would be a lot simpler to decouple it by having the identity transformation occur as a transformation on the ledger events before they are seen by the ledger itself.

"Transforming ledger events" is just a fancy term for an illegal operation in a ledger: rewriting history.

```
# Before rewriting history
discourse/abc receives +100g
github/abc receives +50g
discourse/abc tries to send 120g <-- insufficient funds

# After rewriting history
identity/abc receives +100g
identity/abc receives +50g
identity/abc tries to send 120g <-- sufficient funds
```

The time when an identity was merged is important for validation rules of other events. Thus I think it the best way to model this is with an event.

src/grain/ledger.js
Beanow: Good question @hammadj. And I'm inclined to agree with @decentralion, but with a slightly different rationale.

---

First, after having another look. I want to get some terminology straight.

The `Ledger` interface and `InMemoryLedger` class. **Are not ledgers**. The `Ledger` interface is actually a _projection_, the result of aggregating a stream of events into a "current state".

It includes data a ledger does not have: reduced state. And it's missing the only, but required mutation a ledger has: append.

I think a ledger is an append-only log of _account mutations_. This is a specific variant of the append-only log of events. The term borrows from the (much more mature than IT) field of accounting.

---

> So if we don't keep that info in the ledger, then we need to track two ledgers: the grain balance ledger, and the distribution history ledger. At which point we risk them de-synchronizing,

This is very much on point. You can definitely create working systems with multiple event logs. I've built a couple of these in the past. But the main concern when deciding whether it's best to have one or multiple is _consistency guarantees_.

Consistency in an event log all boils down to _the order of events_.

Using one log is simple, because a log is ordered. The sequence number / array index is what gives you consistency.

When you're using multiple logs, the only approximation of order you have is the timestamp. Timestamps open you up to all kinds of race conditions and edge cases. Referencing IDs does not solve this. And things get very messy when considering a distributed system.

That said, I think @hammadj's suggested approach is feasible, but comes at a cost. Representing a distribution as a sequence of "Grain Minted" events would work. Without the second event log, this is a lossy representation.

You can _reconstruct_ the distribution by projecting this sequence of minting events back into what the actual effect was of this sequence. And you can answer the question of whether these events were part of the same distribution using reference IDs. But what can't be reconstructed is whether the actual effect matches the intended effect. So preserving this intent is what we'd need the second event log for.

By modeling the distribution as a single event, you don't lose the context or intent.

I think we _want to_ provide consistency guarantees about the distribution as a whole and it's intent. Using a single event and one log, is actually a less complex way to achieve that than requiring the second log and dealing with inconsistency.

---

As an aside. Computational, implementation or event model complexity are not very important. Taking time to carefully model your events is definitely worthwhile though, because changing events after the fact is illegal.

For better or worse you will be supporting these event models _until the end of time_. I disagree with @decentralion here, with regards to "private data" and upgrade paths. These events are your _source of truth_ and updating them after the fact is a breach of contract. Even re-encoding the same data with the same meaning into a new format should be frowned upon. If this were on-chain data, this would rightfully be made infeasible to do. The only thing you can upgrade is how you interpret this data, and fix mistakes by appending corrections.

If this "simpler ledger" you had in mind was actually not a ledger but a projection though. That's something you can do. That's an upgrade of how you interpret the data. Discarding data and making use case specific projections is totally OK (and a feature in many cases).

Finding those caching and optimization strategies is something you can do later and as needed. But reproducing data that was lost because the model was incomplete when you recorded it is something you want to avoid.

So as a rule of thumb, I would side with caution and include data you can't easily reproduce and seems like it would have a future use case.

For example, something I like about the current model: it includes the starting budget and (versioned reference to) the strategy used. But doesn't include easy to reproduce data such as the sum of actually minted grain and the difference from the starting budget.
But I'll get into that in other comments.

src/grain/ledger.js
decentralion: Update: I decided to make the `InMemoryLedger` class a non-exported class, and just have a pure function which produces a fully constructed one. I believe this obviates concerns about having the logic-heavy constructor (it's now a pure implementation detail, we don't have to worry about others trying to inherit from it or something), without needing a rewrite.

decentralion:
You've given an example where merging identities can make an invalid chain into a valid chain. However, I assert that it would never make a valid chain invalid (it may turn valid transfers into self-transfers, but we explicitly test to ensure that self-transfers are supported).

From an implementation perspective, re-writing the event history is much more convenient. Suppose that we want to be able to query to find all of the distributions when user Foo (aliases Bar and Zod) got grain. If we re-write the history before computing the ledger state, we can just query for Foo (the identity re-write is locally scoped to the history transformer). However, if we don't, then every Ledger implementation needs to implement logic for tracking different identities and aggregating across them.

However, you seem to have priors on working with ledgers and account histories (that I dont have), so if you feel strongly that we should be doing this your way, I'm open to it.

Also, either way, we will definitely never allow un-merging identities, since that can transform a valid history into an invalid history.

Beanow:
> You've given an example where merging identities can make an invalid chain into a valid chain. 

The rewritten history from my example was not a valid chain. It incorrectly considers a transaction to be valid that was invalid at the time. What it shows is that applying validation rules to an alternate history doesn't prove anything.

> From an implementation perspective, re-writing the event history is much more convenient.

Especially in a financial record, we can't sacrifice correctness for convenience.
So convenience is going to have to come from elsewhere ;]
Careful and disciplined event modeling being the recommended starting point.

First I would like to point out there's a distinction between technical correctness and having proof for that on record. Rewriting your event history, may be a technically correct representation. But loses it's authenticity if you delete the originals it was based on. In the same way that an accountant can't simply rewrite company records and burn the original books by showing how the rewriting process was correct.

The "Ledger" class, which is actually a projection, is responsible for validating the sequence of events. Why must it be responsible for that? Because we're depending on it:
- How would you create new distributions if you don't know you're basing them on a valid history?
- And what's the point of showing users a balance or transaction history we can't prove is valid?

So as I showed in that example, if we're giving this projection an alternative history, the reduce function's validation checks provide no evidence that the original events were ever valid.

Of course this is assuming we do a "blind" transformation of the history, which transforms without validating the complete sequence of events first. If we can prove that the alternative history accurately represents the original history, we could build some chain of proofs.

What would a "smart" transformation that does that look like? A smart transformation is itself a projection. It needs to apply _all the validation rules_ to assume the responsibility of ensuring the original input was a valid sequence of events. Note how this implies we need to support processing and validating the _original events before identity contracting_ which is exactly what we were trying to avoid having to implement, so we've gained no convenience here.

And this only handles the technical correctness. If we don't keep the original events on record we're still the accountant that burned the books. We'll have to be able to run this validating transformation with the original events till the end of time.

So doing this rewriting of history in a sound way is actually a lot more complex than designing a neat "AccountsMerged" event and implementing the command and reducer for it.

It makes intuitive sense too.
Normally all we ever do in an event log is to append facts (events are facts).
If something needs to be amended, you record the fact you corrected it.
Rewriting history is overall dangerous territory.

decentralion:
Superseded by #1743, #1774

PR #1699, opened 2020-03-09
Discourse: enable mirror schema upgrading

Beanow:
There's a small change I have in mind to upgrade to a v7 database.
However it doesn't warrant throwing out your existing cache, so I've
created a way to upgrade versions.

For this PR I've dug up the v5 (previous) schema and added an upgrade
to the current v6 as an example. This so the upgrading PR can be reviewed
separately from the v7 schema I'll suggest in a follow-up.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/discourse/mirrorSchema.js
+++ b/src/plugins/discourse/mirrorSchema.js
@@ -1,0 +1,91 @@
+// @flow
+
+import stringify from "json-stable-stringify";
+import dedent from "../../util/dedent";
+
+// The version should be bumped any time the database schema is changed,
+// so that the cache will be properly invalidated.
+export const SCHEMA_VERSION = "discourse_mirror_v6";
+
+type Upgrade = {|
+  +target: string,
+  +changes: $ReadOnlyArray<string>,
+|};
+
+type VersionConfig = {|
+  +version: string,
+  +serverUrl: string,
+|};
+
+export const stringifyConfig = (c: VersionConfig): string => stringify(c);
+export const parseConfig = (json: string): VersionConfig => JSON.parse(json);
+
+// Queries to upgrade from a previous version to the target version.
+export const upgrades: {[current: string]: Upgrade} = {
+  discourse_mirror_v5: {
+    target: "discourse_mirror_v6",
+    changes: [
+      dedent`\
+        CREATE TABLE sync_heads (
+          key TEXT PRIMARY KEY,
+          timestamp_ms INTEGER NOT NULL
+        )`,
+    ],
+  },
+};
+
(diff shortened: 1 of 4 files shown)


src/plugins/discourse/mirrorSchema.js
```diff
@@ -0,0 +1,91 @@
...
+    target: "discourse_mirror_v6",
+    changes: [
+      dedent`\
+        CREATE TABLE sync_heads (
+          key TEXT PRIMARY KEY,
+          timestamp_ms INTEGER NOT NULL
+        )`,
+    ],
+  },
+};
```
decentralion: Invariants required:
- for every version in the upgrades, the target should be a higher version
- for every target, it should either be the current latest version, or the target should also have an upgrade available

This way we know we can get from every version included to the current verison. Maybe worth documenting / testing?

Beanow: Agreeing with those invariants.
Should the 2nd not be true, because of a breaking change that isn't upgradable for instance. I think it's OK to remove all the upgrades preceding and start from a new "create" point.

I'm assuming we don't have particular tools for this, so I'd write this as unit tests right?

decentralion: > Should the 2nd not be true, because of a breaking change that isn't upgradable for instance. I think it's OK to remove all the upgrades preceding and start from a new "create" point.

Yeah, that seems reasonable.

> I'm assuming we don't have particular tools for this, so I'd write this as unit tests right?

Seems reasonable. A test that basically does a graph traversal over the upgrades object, and verifies that each upgrade path arrives at the current version?

Beanow: I've added a commit to this PR which has the invariant tests.
https://github.com/sourcecred/sourcecred/pull/1699/commits/efde87398055fe8487f3eb22f45b00d4bc356011

The diff is growing to be fairly large with that included. So if the general approach seems right, I'm inclined to break this out to a util and several PRs. What do you think?

Beanow:
Having sat on this for a while. I'm inclined to no go with migrations for the mirror. Primarily because I don't want to commit to the extra complexity for disposable cache data.

Suggesting we limit the number of cache clears required by end-users by recommending them to use versioned releases instead.

What do you think @decentralion?

decentralion:
> Having sat on this for a while. I'm inclined to not go with migrations for the mirror. Primarily because I don't want to commit to the extra complexity for disposable cache data.
>
> Suggesting we limit the number of cache clears required for end-users by recommending them to use versioned releases instead.

Sounds good to me. Closing since we have consensus.

PR #1701, opened 2020-03-09
Discourse: use plaintext cache keys

Beanow:
In a similar vain to #1685, make the cache keys human readable.

Previously we had base64 and no prefix, which is case-sensitive.
So we were not scoping to our plugin, nor supporting case-insensitive
filesystems.

The fields preserved from a URL are the protocol, hostname and optional
custom port. Using `_` as the separator for those.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/discourse/cacheKey.js
+++ b/src/plugins/discourse/cacheKey.js
@@ -1,0 +1,14 @@
+// @flow
+
+import {type ServerUrl} from "./models";
+
+/**
+ * Creates a cache key that's unique and safe for filesystem use.
+ */
+export function cacheKey(serverUrl: ServerUrl): string {
+  const url = new global.URL(serverUrl);
+  // Remove trailing colon from protol.
+  const protocol = url.protocol.replace(/:$/, "");
+  const portSuffix = url.port ? `_${url.port}` : "";
+  return `discourse_${protocol}_${url.hostname}${portSuffix}`.toLowerCase();
+}
--- a/src/plugins/discourse/cacheKey.test.js
+++ b/src/plugins/discourse/cacheKey.test.js
@@ -1,0 +1,45 @@
+// @flow
+
+import {parseServerUrl} from "./models";
+import {cacheKey} from "./cacheKey";
+
+describe("plugins/discourse/cacheKey", () => {
+  describe("cacheKey", () => {
+    it("should work with valid server URLs", () => {
+      const expected = [
+        {serverUrl: "https://foo.bar", cacheKey: "discourse_https_foo.bar"},
+        {
+          serverUrl: "http://example.com/",
+          cacheKey: "discourse_http_example.com",
+        },
+        {
+          serverUrl: "HTTPS://Casing.WTF",
+          cacheKey: "discourse_https_casing.wtf",
+        },
+        {
+          serverUrl: "https://custom.port.wtf:9001",
+          cacheKey: "discourse_https_custom.port.wtf_9001",
+        },
+        {
+          serverUrl: "http://default.port.wtf:80",
+          cacheKey: "discourse_http_default.port.wtf",
+        },
+        {
+          serverUrl: "https://default.port.wtf:443",
+          cacheKey: "discourse_https_default.port.wtf",
+        },
+        {
+          serverUrl: "https://discourse.sourcecred.io/",
+          cacheKey: "discourse_https_discourse.sourcecred.io",
+        },
+      ];
+
+      const actual = expected.map(({serverUrl}) => ({
+        serverUrl,
+        cacheKey: cacheKey(parseServerUrl(serverUrl)),
+      }));
+
+      expect(expected).toEqual(actual);
+    });
+  });
+});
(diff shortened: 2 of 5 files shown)


src/plugins/discourse/cacheKey.test.js
```diff
@@ -0,0 +1,22 @@
...
+        "https://foo.bar",
+        "http://example.com/",
+        "HTTPS://Casing.WTF",
+        "https://custom.port.wtf:9001",
+        "https://discourse.sourcecred.io/",
+      ];
+      const results = corpus
+        .map(validateServerUrl)
+        .map((url) => [url, cacheKey(url)]);
+      expect(results).toMatchSnapshot();
```
wchargin: I would really prefer if we could avoid using snapshots for tests like
this. Snapshots have one advantage: when the test output is large, can
only reasonably be machine-generated, and is expected to be updated and
audited frequently, the snapshot can be helpful. But they have a bunch
of disadvantages. They move the expected output away from the input.
They move the expected output away from the code under test. They make
it harder to understand the repository history. They make it really easy
to write brittle tests, and they encourage cavalierly committing
snapshot updates.

I often read tests first. For instance, I cam to this test file to check
whether the outputs are expected to be lowercased. But after reading the
test file I still have no idea. Okay, I’ll check the snapshots—but there
are *two* snapshot files added in this change. Now I have to figure out
which one is which. All of this is extra friction. The expected output
should be right here with the inputs and test code.

Snapshots make it easy to write bad tests; resist the temptation. :-)

(You can take a look at #1685 for one clean way to do this.)

Beanow: Thanks, updated as request :]

src/plugins/discourse/models.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
+export opaque type ServerUrl: string = string;
+
+/**
+ * Does a strict validation of ServerUrl inputs and returns an opaque type
+ * indicating it matches our assumptions elsewhere.
```
wchargin: These assumptions (which?) need to be documented on the opaque type
itself. Otherwise, the type is not useful. Some questions that I have:

  - The code restricts some portions of the URL (scheme, fragment, etc.)
    but not others (username, password). Is it legal for `ServerUrl`s to
    include these components?
  - If two inputs to `validateServerUrl` return the same output, are the
    inputs to be considered the same server URL? If so, this function
    should be called `parse*`, not `validate*` (as it probably should
    be in any case, just due to its type signature). Specifically:
      - Are `http://example.com` and `http://example.com:80` the same
        server URL? [The `URL` constructor elides canonical ports][cp].
      - Are input URLs permitted to have an optional trailing slash?
      - Are input URLs case-sensitive? Many portions of the URL are
        lowercased by the URL constructor, including protocol and host,
        but username and password are not.

The semantics of this type affect how I’m supposed to read, for
instance, the contract of the `cacheKey` function.

[cp]: https://nodejs.org/api/url.html#url_url_port

src/plugins/discourse/cacheKey.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import {type ServerUrl} from "./models";
+
+/**
+ * Creates a cache key that's unique and safe for filesystem use.
+ */
+export function cacheKey(serverUrl: ServerUrl): string {
+  const url = new global.URL(serverUrl);
+  const cleanProtocol = url.protocol.replace(/[^a-z]/i, "");
```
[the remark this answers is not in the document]
Beanow: Works for me. Updated!

src/plugins/discourse/models.js
Beanow: You're right, so I changed this from a blacklisting to a whitelisting strategy. Sacrificing some error message clarity for easier checking what the assumptions are.

Also updated the naming and comments to use `parse*` and consider the opaque version a normalized URL.

src/plugins/discourse/models.js
```diff
@@ -0,0 +1,43 @@
...
+export function parseServerUrl(serverUrl: string): ServerUrl {
+  try {
+    const url = new global.URL(serverUrl);
+
+    if (!new RegExp(/^https?/i).test(url.protocol)) {
+      throw "URL should have a http/https protocol";
+    }
+
+    const rebuiltUrl = new global.URL(
+      `${url.protocol}//${url.hostname}:${url.port}`
```
wchargin, line 29: FWIW, this matches any protocol _starting_ with `http[s]`:

```javascript
> new RegExp(/^https?/i).test(new global.URL("httpscary://wat").protocol)
true
```

Suggest either `/^https?:$/i` or just `["http:", "https:"].includes`;
the protocol always gets downcased by the URL constructor, anyway, as
your tests correctly show.

decentralion:
I don't think it makes sense to merge this anymore -- cli2 is fast approaching, and in it we don't need cache keys for the Discourse plugin, since we currently only support a single Discourse server per plugin, we can just go with a canonical name for the instance's discourse cache. I'm going to close this, if anyone sees a flaw in my reasoning feel free to re-open.

PR #1702, opened 2020-03-10
Add a mock middleware module for the grain ui

decentralion:
This contains a skeleton class (interface really) for the data provider
for the upcoming grain user interface. By agreeing on the interface, we
allow work on the backend and frontend to proceed in parallel.

Paired with @ianjdarrow

Test plan: Only mock tests were added, so really `yarn flow` and `yarn
lint` are doing all the work. They pass. :)

### DIFF
--- a/src/grain-explorer/middleware.js
+++ b/src/grain-explorer/middleware.js
@@ -1,0 +1,72 @@
+// @flow
+
+import {type NodeAddressT} from "../core/graph";
+import {type Grain} from "../grain/grain";
+import {type LedgerEvent} from "../grain/ledger";
+
+export type Contributor = {|
+  +address: NodeAddressT,
+  // Markdown name (may contain hyperlinks)
+  +name: string,
+  +balance: Grain,
+  +lastPeriodIssuance: Grain,
+|};
+
+export type MatchOptions = {|
+  +exact: boolean,
+|};
+
+/**
+ * The GrainMiddleware class provides data that's been preprocessed and
+ * organized to closely match the needs of the grain frontend.
+ */
+export class GrainMiddleware {
+  /*
+   * Lists every contributor in the project, along with summary information
+   * like their grain balance. In the future, we may add pagination.
+   */
+  contributors(): $ReadOnlyArray<Contributor> {
+    throw new Error("Not yet implemented");
+  }
+
+  /**
+   * Given the address of a contributor, return the summary information.
+   */
+  contributor(address: NodeAddressT): ?Contributor {
+    throw new Error("Not yet implemented" + address);
+  }
+
+  /**
+   * Returns all of the LedgerEvents corresponding to a particular contributor.
+   * In the future, we will add options (e.g. on how the events should be sorted);
+   * for now it will provide them in descending timestamp order.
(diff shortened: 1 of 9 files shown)


src/grain-explorer/middleware.js
```diff
@@ -0,0 +1,72 @@
...
+   * Fuzzily searches over the contributors to find ones that match the
+   * search string. Assigns a score to each match based on its closeness.
+   *
+   * In the future, we may add options (e.g. for exactness).
+   */
+  autocomplete(
+    search: string
+  ): $ReadOnlyArray<{|+score: number, +contributor: Contributor|}> {
+    throw new Error("Not yet implemented" + search);
+  }
```
Beanow: So, doing this type of prefix/freetext search on a dataset where 1000s of items (contributors) is considered large, the actual performance impact of a search is likely to be fairly small, especially if we query a well optimized search index for this.

However doing these queries in-browser, multiple times in rapid succession as you type, on the UI thread, is likely to cause noticeable frame dropping / hickups from a UX perspective.

With that in mind I would suggest 2 changes for performance:
- Making this an async function (even if the actual implementation fakes it with a `nextTick`).
- Setting a limit to the # of results to return.

src/grain-explorer/middleware.js
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Returns timeseries data of how much total supply of Grain there's been at
+   * every "issuance period". For now, we'll consider any timestamp which has
+   * at least one Grain distribution to be an issuance period.
+   */
+  supplyTimeseries(): $ReadOnlyArray<{|
+    +timestampMs: number,
+    +supply: Grain,
```
Beanow: Nit: would prefer `totalGrainSupply`. As that's more descriptive when viewed without the comments.

src/grain-explorer/middleware.js
```diff
@@ -0,0 +1,72 @@
...
+    throw new Error("Not yet implemented" + address);
+  }
+
+  /**
+   * Returns all of the LedgerEvents corresponding to a particular contributor.
+   * In the future, we will add options (e.g. on how the events should be sorted);
+   * for now it will provide them in descending timestamp order.
+   * In the future, we may change this to a paginated or asynchronous API.
+   */
+  events(address: NodeAddressT): $ReadOnlyArray<LedgerEvent> {
```
Beanow: Nit: prefer `accountLedgerEvents` to match the type name and reflect we're filtering by account.

src/grain-explorer/middleware.js
```diff
@@ -0,0 +1,72 @@
...
+  contributors(): $ReadOnlyArray<Contributor> {
+    throw new Error("Not yet implemented");
+  }
+
+  /**
+   * Given the address of a contributor, return the summary information.
+   */
+  contributor(address: NodeAddressT): ?Contributor {
+    throw new Error("Not yet implemented" + address);
+  }
```
Beanow: Matching earlier comment, these would be `accountSummary` / `accountSummaries`.

Beanow:
I believe this has gone stale. Let's take another stab at this after we've implemented the output format, and when UI work comes up again.

PR #1703, opened 2020-03-10
Add Discord Fetch Plan

BrianLitwin:
Adds a Fetcher class that takes fetch options and fetches against
the correct endpoints, returning re-formatted request data
and pagination info.

Adds a models.js file that defines Discord types.

Test Plan:

Unit tests added for fetcher class and helper functions in
models.js file.

Paired with @Beanow

### DIFF
--- a/src/plugins/discord/fetcher.js
+++ b/src/plugins/discord/fetcher.js
@@ -1,0 +1,141 @@
+// @flow
+
+import * as Model from "./models";
+import {type Snowflake} from "./models";
+
+type FetchEndpoint = (endpoint: string) => Promise<any>;
+
+/**
+ * Provide the Guild ID to fetch against, and the 'limit'
+ * parameter when fetching GuildMembers, Messages, and Reactions.
+ *
+ */
+type FetchOptions = {|
+  membersLimit: number,
+  messagesLimit: number,
+  reactionsLimit: number,
+|};
+
+export type PageInfo = {|
+  +hasNextPage: boolean,
+  +endCursor: Snowflake | null,
+|};
+
+export type ResultPage<T> = {|
+  +pageInfo: PageInfo,
+  +results: $ReadOnlyArray<T>,
+|};
+
+/**
+ * Fetcher is responsible for:
+ * - Returning the correct endpoint to fetch against for Guilds, Channels,
+ *   Members, and Reactions.
+ * - Formatting the returned results into the correct Typed objects
+ * - Returning pagination info in a PageInfo object, containing hasNextPage
+ *   and endCursor properties.
+ *   The endCursor property is calculated as the Id of the last object
+ *   returned in the response results. We are assuming Discord provides
+ *   consistent, ordered results.
+ *   The hasNextPage property is a boolean calculated as whether the number of
+ *   results recieved is equal to the `limit` property provided in the
+ *   fetch request.
+ *
+ *   Note that Discord doesn't support pagination for Channels, so we're
+ *   returning an array of Channel objects in the corresponding method.
+ *   See: https://discordapp.com/developers/docs/resources/guild#get-guild-channels
+ *
+ */
+export class Fetcher {
+  +_fetch: FetchEndpoint;
+  +_options: FetchOptions;
+
+  constructor(fetchEndpoint: FetchEndpoint, options: FetchOptions) {
+    this._fetch = fetchEndpoint;
+    this._options = options;
+  }
+
+  async guild(guildId: Snowflake): Promise<Model.Guild> {
+    const {id, name, permissions} = await this._fetch(`guilds/${guildId}`);
+    return {id: id, name: name, permissions: permissions};
+  }
+
+  async channels(guildId: Snowflake): Promise<$ReadOnlyArray<Model.Channel>> {
+    const response = await this._fetch(`/guilds/${guildId}/channels`);
+    return response.map((x) => ({
+      id: x.id,
+      name: x.name,
+      type: Model.channelTypeFromId(x.type),
+    }));
+  }
+
+  async members(
+    guildId: Snowflake,
+    after: Snowflake
+  ): Promise<ResultPage<Model.GuildMember>> {
+    const {membersLimit} = this._options;
+    const endpoint = `/guilds/${guildId}/members?after=${after}&limit=${membersLimit}`;
+    const response = await this._fetch(endpoint);
+    const results = response.map((x) => ({
+      user: {
+        id: x.user.id,
+        username: x.user.username,
+        discriminator: x.user.discriminator,
+        bot: x.user.bot || x.user.system || false,
+      },
+      nick: x.nick || null,
+      roles: x.roles,
+    }));
+    const hasNextPage = results.length === membersLimit;
+    const endCursor =
+      response.length > 0 ? response[response.length - 1].user.id : null;
+    const pageInfo = {hasNextPage, endCursor};
+    return {results, pageInfo};
+  }
+
+  async messages(
+    channel: Snowflake,
+    after: Snowflake
+  ): Promise<ResultPage<Model.Message>> {
+    const {messagesLimit} = this._options;
+    const endpoint = `/channels/${channel}/messages?after=${after}&limit=${messagesLimit}`;
+    const response = await this._fetch(endpoint);
+    const results = response.map((x) => ({
+      id: x.id,
+      channelId: channel,
+      authorId: x.author.id,
+      timestampMs: Date.parse(x.timestamp),
+      content: x.content,
+      reactionEmoji: (x.reactions || []).map((r) => r.emoji),
+      nonUserAuthor: Model.isAuthoredByNonUser(x),
+      mentions: (x.mentions || []).map((user) => user.id),
+    }));
+    const hasNextPage = results.length === messagesLimit;
+    const endCursor =
+      response.length > 0 ? response[response.length - 1].id : null;
+    const pageInfo = {hasNextPage, endCursor};
+    return {results, pageInfo};
+  }
+
+  async reactions(
+    channel: Snowflake,
+    message: Snowflake,
+    emoji: Model.Emoji,
+    after: Snowflake
+  ): Promise<ResultPage<Model.Reaction>> {
+    const {reactionsLimit} = this._options;
+    const emojiRef = Model.emojiToRef(emoji);
+    const endpoint = `/channels/${channel}/messages/${message}/reactions/${emojiRef}?after=${after}&limit=${reactionsLimit}`;
+    const response = await this._fetch(endpoint);
+    const results = response.map((x) => ({
+      emoji: x.emoji,
+      channelId: channel,
+      messageId: message,
+      authorId: x.id,
+    }));
+    const hasNextPage = results.length === reactionsLimit;
+    const endCursor =
+      response.length > 0 ? response[response.length - 1].id : null;
+    const pageInfo = {hasNextPage, endCursor};
+    return {results, pageInfo};
+  }
+}
--- a/src/plugins/discord/fetcher.test.js
+++ b/src/plugins/discord/fetcher.test.js
@@ -1,0 +1,299 @@
+// @flow
+
+import {Fetcher} from "./fetcher";
+import {
+  type Channel,
+  type Guild,
+  type Reaction,
+  type Message,
+  type GuildMember,
+  type Snowflake,
+  type Emoji,
+} from "./models";
+
+/**
+ * Note: the 'any' type signature is assigned to server response objects
+ * to make explicit that they are untyped prior to transformation by the
+ * appropriate handlers
+ */
+
+describe("plugins/discord/fetcher", () => {
+  const defaultOptions = () => ({
+    membersLimit: 100,
+    messagesLimit: 100,
+    reactionsLimit: 100,
+  });
+
+  describe("fetch guild", () => {
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      await fetcher.guild("1");
+      expect(fetch.mock.calls[0]).toEqual(["guilds/1"]);
+    });
+
+    it("handles response", async () => {
+      const expected: Guild = {id: "1", name: "guildname", permissions: 0};
+      const fetch = jest.fn(() => Promise.resolve(expected));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      const guild = await fetcher.guild("1");
+      expect(guild).toEqual(expected);
+    });
+  });
+
+  describe("fetch channels", () => {
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      await fetcher.channels("1");
+      expect(fetch.mock.calls[0]).toEqual(["/guilds/1/channels"]);
+    });
+
+    it("handles response", async () => {
+      const testChannel = (id: string): any => ({
+        id: id,
+        name: "testChannelName",
+        type: 0,
+      });
+
+      const testChannelObj = (id: Snowflake): Channel => ({
+        id: id,
+        name: "testChannelName",
+        type: "GUILD_TEXT",
+      });
+
+      const response: any[] = [testChannel("1")];
+      const expected: Channel[] = [testChannelObj("1")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      const data = await fetcher.channels("0");
+      expect(data).toEqual(expected);
+    });
+  });
+
+  describe("fetch members", () => {
+    const testMember = (userId: string): any => ({
+      user: {
+        id: userId,
+        username: "username",
+        discriminator: "disc",
+        bot: true,
+      },
+      nick: "nickname",
+      roles: ["test role"],
+    });
+
+    const options = () => ({
+      membersLimit: 2,
+      messagesLimit: 100,
+      reactionsLimit: 100,
+    });
+
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, options());
+      await fetcher.members("1", "0");
+      expect(fetch.mock.calls[0]).toEqual([
+        "/guilds/1/members?after=0&limit=2",
+      ]);
+    });
+
+    it("handles response", async () => {
+      const response: GuildMember[] = [testMember("1")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {results} = await fetcher.members("1", "0");
+      expect(results).toEqual(response);
+    });
+
+    it("returns correct endCursor", async () => {
+      const response: any[] = [testMember("1"), testMember("2")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {endCursor} = (await fetcher.members("1", "0")).pageInfo;
+      expect(endCursor).toEqual("2");
+    });
+
+    describe("next page", () => {
+      it("next page = true", async () => {
+        const response: any[] = [testMember("1")];
+        const fetch = jest.fn(() => Promise.resolve(response));
+        const fetcher = new Fetcher(fetch, {
+          membersLimit: 1,
+          messagesLimit: 100,
+          reactionsLimit: 100,
+        });
+        const {hasNextPage} = (await fetcher.members("1", "0")).pageInfo;
+        expect(hasNextPage).toBe(true);
+      });
+
+      it("next page = false", async () => {
+        const response: any[] = [testMember("1")];
+        const fetch = jest.fn(() => Promise.resolve(response));
+        const fetcher = new Fetcher(fetch, {
+          membersLimit: 2,
+          messagesLimit: 100,
+          reactionsLimit: 100,
+        });
+        const {hasNextPage} = (await fetcher.members("1", "0")).pageInfo;
+        expect(hasNextPage).toBe(false);
+      });
+    });
+  });
+
+  describe("fetch reactions", () => {
+    const emoji: Emoji = {id: "1", name: "emojiname"};
+
+    const options = () => ({
+      membersLimit: 100,
+      messagesLimit: 100,
+      reactionsLimit: 2,
+    });
+
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, options());
+      await fetcher.reactions("1", "2", emoji, "0");
+      expect(fetch.mock.calls[0]).toEqual([
+        `/channels/1/messages/2/reactions/emojiname:1?after=0&limit=2`,
+      ]);
+    });
+
+    it("handles response", async () => {
+      // response object from server, prior to type transformation
+      const response: any[] = [{id: "123", emoji}];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {results} = await fetcher.reactions("3", "2", emoji, "0");
+      const expected: Reaction[] = [
+        {emoji, channelId: "3", messageId: "2", authorId: "123"},
+      ];
+      expect(results).toEqual(expected);
+    });
+
+    it("returns correct endCursor", async () => {
+      const response: any[] = [
+        {id: "1", emoji},
+        {id: "2", emoji},
+      ];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {endCursor} = (
+        await fetcher.reactions("1", "2", emoji, "0")
+      ).pageInfo;
+      expect(endCursor).toBe("2");
+    });
+
+    describe("next page", () => {
+      it("next page = true", async () => {
+        const response: any[] = [{id: 1, emoji}];
+        const fetch = jest.fn(() => Promise.resolve(response));
+        const fetcher = new Fetcher(fetch, {
+          membersLimit: 100,
+          messagesLimit: 100,
+          reactionsLimit: 1,
+        });
+        const {hasNextPage} = (
+          await fetcher.reactions("1", "2", emoji, "0")
+        ).pageInfo;
+        expect(hasNextPage).toBe(true);
+      });
+
+      it("next page = false", async () => {
+        const response: any[] = [{id: 1, emoji}];
+        const fetch = jest.fn(() => Promise.resolve(response));
+        const fetcher = new Fetcher(fetch, {
+          membersLimit: 100,
+          messagesLimit: 100,
+          reactionsLimit: 2,
+        });
+        const {hasNextPage} = (
+          await fetcher.reactions("1", "2", emoji, "0")
+        ).pageInfo;
+        expect(hasNextPage).toBe(false);
+      });
+    });
+  });
+
+  describe("fetch messages", () => {
+    const testMessage = (id: string): any => ({
+      id: id,
+      author: {id: "2"},
+      timestamp: "2020-03-03T23:35:10.615000+00:00",
+      content: "Just going to drop this here",
+      reactions: [{emoji: {id: "1", name: "testemoji"}}],
+      mentions: [{id: "4", username: "testuser"}],
+    });
+
+    const options = () => ({
+      membersLimit: 100,
+      messagesLimit: 2,
+      reactionsLimit: 100,
+    });
+
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, options());
+      await fetcher.messages("1", "0");
+      expect(fetch.mock.calls[0]).toEqual([
+        "/channels/1/messages?after=0&limit=2",
+      ]);
+    });
+
+    it("handles response", async () => {
+      const response: any[] = [testMessage("1")];
+
+      const expected: Message[] = [
+        {
+          id: "1",
+          channelId: "123",
+          authorId: "2",
+          timestampMs: Date.parse("2020-03-03T23:35:10.615000+00:00"),
+          content: "Just going to drop this here",
+          reactionEmoji: [{id: "1", name: "testemoji"}],
+          nonUserAuthor: false,
+          mentions: ["4"],
+        },
+      ];
+
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {results} = await fetcher.messages("123", "0");
+      expect(results).toEqual(expected);
+    });
+
+    it("returns correct endCursor", async () => {
+      const response: any[] = [testMessage("1"), testMessage("2")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {endCursor} = (await fetcher.messages("123", "0")).pageInfo;
+      expect(endCursor).toBe("2");
+    });
+
+    describe("next page", () => {
+      it("next page = true", async () => {
+        const response: any[] = [testMessage("1")];
+        const fetch = jest.fn(() => Promise.resolve(response));
+        const fetcher = new Fetcher(fetch, {
+          membersLimit: 100,
+          messagesLimit: 1,
+          reactionsLimit: 2,
+        });
+        const {hasNextPage} = (await fetcher.messages("123", "0")).pageInfo;
+        expect(hasNextPage).toBe(true);
+      });
+
+      it("next page = false", async () => {
+        const response: any[] = [testMessage("1")];
+        const fetch = jest.fn(() => Promise.resolve(response));
+        const fetcher = new Fetcher(fetch, {
+          membersLimit: 100,
+          messagesLimit: 2,
+          reactionsLimit: 2,
+        });
+        const {hasNextPage} = (await fetcher.messages("123", "0")).pageInfo;
+        expect(hasNextPage).toBe(false);
+      });
+    });
+  });
+});
--- a/src/plugins/discord/models.js
+++ b/src/plugins/discord/models.js
@@ -1,0 +1,134 @@
+// @flow
+
+// https://discordapp.com/developers/docs/reference#snowflakes
+export type Snowflake = string;
+export const ZeroSnowflake: Snowflake = "0";
+
+export type BotToken = string;
+
+/**
+ * Discord Channels can be of various types, defined below.
+ * See: https://discordapp.com/developers/docs/resources/channel#channel-object-channel-types
+ *
+ */
+export type ChannelType =
+  | "GUILD_TEXT"
+  | "DM"
+  | "GUILD_VOICE"
+  | "GROUP_DM"
+  | "GUILD_CATEGORY"
+  | "GUILD_NEWS"
+  | "GUILD_STORE";
+
+/**
+ * The Discord server returns an Id field of type number to represent
+ * the type of Channel. Here we convert that to a text representation,
+ * also based on: https://discordapp.com/developers/docs/resources/channel#channel-object-channel-types
+ *
+ */
+export function channelTypeFromId(id: number): ChannelType {
+  switch (id) {
+    case 0:
+      return "GUILD_TEXT";
+    case 1:
+      return "DM";
+    case 2:
+      return "GUILD_VOICE";
+    case 3:
+      return "GROUP_DM";
+    case 4:
+      return "GUILD_CATEGORY";
+    case 5:
+      return "GUILD_NEWS";
+    case 6:
+      return "GUILD_STORE";
+    default: {
+      throw new Error(`Unknown channel type ID: ${id}`);
+    }
+  }
+}
+
+// https://discordapp.com/developers/docs/resources/guild#guild-object
+export type Guild = {|
+  +id: Snowflake,
+  +name: string,
+  +permissions: number,
+|};
+
+// https://discordapp.com/developers/docs/resources/channel#channel-object
+export type Channel = {|
+  +id: Snowflake,
+  +type: ChannelType,
+  +name: string,
+|};
+
+// https://discordapp.com/developers/docs/resources/user#user-object
+export type User = {|
+  +id: Snowflake,
+  +username: string,
+  +discriminator: string,
+  +bot: boolean,
+|};
+
+// https://discordapp.com/developers/docs/resources/guild#guild-member-object
+export type GuildMember = {|
+  +user: User,
+  +nick: string | null,
+  +roles: $ReadOnlyArray<Snowflake>,
+|};
+
+// From the Discord docs: "emoji takes the form of name:id for
+// custom guild emoji, or Unicode characters."
+// https://discordapp.com/developers/docs/resources/channel#create-message
+export type Emoji = {|
+  +id: ?Snowflake,
+  +name: string,
+|};
+
+export type EmojiRef = string;
+
+/**
+ * Custom emoji returned from the Discord server are defined
+ * by an `id` and `name` property. Generic emoji are defined by
+ * a unicode name property only, with a null id property.
+ * This function returns the appropriate reference based on this
+ * generic vs custom distinction.
+ */
+export function emojiToRef({id, name}: Emoji): EmojiRef {
+  // Built-in emoji, unicode names.
+  if (!id) return name;
+
+  // Custom emoji.
+  return `${name}:${id}`;
+}
+
+// Determines whether the message was created by a webhook or a Discord User
+export function isAuthoredByNonUser(rawMessage: {
+  +webhook_id?: Snowflake,
+}): boolean {
+  return rawMessage.webhook_id != null || false;
+}
+
+// https://discordapp.com/developers/docs/resources/channel#message-object
+export type Message = {|
+  +id: Snowflake,
+  +channelId: Snowflake,
+  +authorId: Snowflake,
+  // Could be a message from a webhook, meaning the authorId isn't a user.
+  +nonUserAuthor: boolean,
+  +timestampMs: number,
+  +content: string,
+  // Normally includes reaction counters, but we don't care about counters.
+  // We could filter based on which types of emoji have been added though.
+  +reactionEmoji: $ReadOnlyArray<Emoji>,
+  // Snowflake of user IDs.
+  +mentions: $ReadOnlyArray<Snowflake>,
+|};
+
+// https://discordapp.com/developers/docs/resources/channel#get-reactions
+export type Reaction = {|
+  +channelId: Snowflake,
+  +messageId: Snowflake,
+  +authorId: Snowflake,
+  +emoji: Emoji,
+|};
--- a/src/plugins/discord/models.test.js
+++ b/src/plugins/discord/models.test.js
@@ -1,0 +1,28 @@
+// @flow
+
+import {emojiToRef, isAuthoredByNonUser} from "./models";
+
+describe("plugins/discord/models", () => {
+  describe("model helper functions", () => {
+    describe("emojiToRef", () => {
+      it("returns name if id is null", () => {
+        expect(emojiToRef({id: null, name: "testEmojiName"})).toBe(
+          "testEmojiName"
+        );
+      });
+      it("returns name and id if id is not null", () => {
+        expect(emojiToRef({id: "testEmojiId", name: "testEmojiName"})).toBe(
+          "testEmojiName:testEmojiId"
+        );
+      });
+    });
+    describe("isAuthoredByNonUser", () => {
+      it("returns true if webhook_id property is provided in message", () => {
+        expect(isAuthoredByNonUser({webhook_id: "0"})).toBe(true);
+      });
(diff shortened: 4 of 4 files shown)


src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
...
+    const endpoint = `/channels/${channel}/messages?after=${after}&limit=${messagesLimit}`;
+    const response = await this._fetch(endpoint);
+    const messages = response.map((x) => ({
+      id: x.id,
+      channelId: channel,
+      authorId: x.author.id,
+      timestampMs: Date.parse(x.timestamp),
+      content: x.content,
+      reactionEmoji: (x.reactions || []).map((r) => r.emoji),
+      nonUserAuthor: x.webhook_id != null || false,
```
Beanow: Nit: for easier reading / maintainability this could use it's own named function.

```js
isAuthoredByNonUser(rawMessage: {+webhook_id?: Snowflake}): boolean
```

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,202 @@
...
+): Promise<void> {
+  let after = endCursor;
+  let cont = true;
+
+  while (cont) {
+    const {hasNextPage, endCursor} = await fetch(after);
+    after = endCursor;
+    cont = hasNextPage;
+  }
+}
```
Beanow: I think this function to hide pagination creates confusing / difficult to understand side-effects code.
Why not use an explicit loop?

```js
while (hasNextPage) {
  // ... have side-effects here
}
```

Bonus points if you're able to figure out and use async generators to convert paginated APIs into a stream :smile: (I'll have some examples of this soon©)

```js
class Fetcher {
  // All messages in a channel, using AsyncGenerator as an abstraction over pagination.
  async *messages(
    channel: Model.Snowflake
  ): AsyncGenerator<Model.Message, void, void> {
    /* ... */
  }
}

// When consuming, use the "new" for await ... of
for await (const message of fetcher.messages(channel)) {
  // ... have side-effects here
}
```

BrianLitwin:
Thanks for the feedback @Beanow !

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
...
+
+export type ReactionResponse = {|
+  +reactions: $ReadOnlyArray<Model.Reaction>,
+  ...pageInfo,
+|};
+
+export type MessageResponse = {|
+  +messages: $ReadOnlyArray<Model.Message>,
+  ...pageInfo,
+|};
```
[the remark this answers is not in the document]
BrianLitwin: Updated to use Generics. 

How would you have done the `hasNextPage` / `endCursor` part? I don't see a reason  _not_ to implement this in a testable, typed format. And I don't see a reason to include information we aren't going to use, like the page size.

src/plugins/discord/fetchDiscord.js
BrianLitwin: Can you elaborate on your suggestion? 

The idea of this function was to 1) have a testable pagination helper, and 2) remove the pagination logic from the site of the fetch implementation, so that the side-effects of calling the network and saving the results are emphasized in the fetching logic. 

Writing a unique while loop at the site of every paginated request is hard to test, has the potential to fail silently, and adds cognitive overhead to the code at those sites. Feels like low-hanging fruit to abstract away into a testable format.

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  const response = await fetcher.guild();
+  if (response.status === 200) {
+    return Promise.reject(
+      `Bot does not have access to Guild with id: ${guildId}`
+    );
+  } else {
+    return Promise.reject(`Guild does not exist with id: ${guildId}`);
+  }
```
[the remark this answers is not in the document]
BrianLitwin: Updated

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,198 @@
...
+}
+
+export async function validateGuildId(fetcher: Fetcher) {
+  try {
+    return await rejectInvalidGuildId(fetcher);
+  } catch (error) {
+    throw new Error(error);
+  }
+}
```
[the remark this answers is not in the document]
BrianLitwin: Thanks for this suggestion and the one above!! Updated.

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,198 @@
...
+
+  while (cont) {
+    const {hasNextPage, endCursor} = await fetch(after);
+    after = endCursor;
+    cont = hasNextPage;
+  }
+}
+
+// Verifies that the provided bot credentials have access to the provided Guild
+async function rejectInvalidGuildId(fetcher: Fetcher) {
```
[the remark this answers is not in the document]
BrianLitwin: This bit adds a more granular error message for clients; the tradeoff in complexity seems minimal. It verifies whether the guild id provided is linked to a guild that actually exists in 3-4 lines. It feels like a plausible error to mis-type or copy/paste a token. 

That said, I don't love this function. I'm don't feel airtight that the logic is correct. I.e, I'm not sure a non-200 response from the Discord server for a guild id means that the Guild doesn't exist.  There also may be additional checks we can perform, perhaps for token-length .

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,198 @@
...
+      default:
+        return Promise.resolve(response.json());
+    }
+  });
+}
+
+// If retry = true in tryDiscordFetch reject, will re-attempt the fetch.
+function retryDiscordFetch(fetch, url, fetchOptions) {
+  return new Promise((resolve, reject) => {
+    const operation = retry.operation();
```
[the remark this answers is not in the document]
Beanow: Noticed I didn't mention why randomized is important. Randomizing is very useful in avoiding deadlocks in distributed / parallel cases. And since we're already distributed + parallel simply from running multiple plugins at once, this should be useful in all cases :]

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,190 @@
...
+// Verifies that the provided bot credentials have access to the provided Guild
+export async function validateGuildId(fetcher: Fetcher) {
+  const guildId = fetcher._options.guildId;
+  const guilds = await fetcher.guilds();
+  const ids = guilds.map((x) => x.id);
+  for (const id of ids) {
+    if (id === guildId) {
+      return guildId;
+    }
+  }
```
Beanow: Nit: Looking at this bit again there's a couple of built-in functions this could be simplified with.
[`.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) seems to be what you need.
`.map().includes()`
`.map().find()`

`if(guilds.some((x) => x.id === guildId)`

src/plugins/discord/fetchDiscord.js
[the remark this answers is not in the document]
Beanow: Looking again I see what the intention is now.

1. You're looking if the guild is listed in `/users/@me/guilds`.
2. You're checking if the guild even exists with `/guilds/${guildId}`.

The second one could improve the error message if it's not a permission error but it's the wrong ID to begin with.

I've tested the 2nd step with Postman. And it turns out that we always receive a 403 response regardless of whether it exists.

For example:
- `/guilds/12345` is not a valid Snowflake and returns 403.
- `/guilds/9223372036854775807` is the maximum allowed Snowflake and returns 403.
- `/guilds/616186924390023171` does exist (server for [Tauri](https://github.com/tauri-apps/)), but no access returns 403.

(Note, I'm reasonably confident the maximum Snowflake value doesn't exist, given the timestamp it contains when parsing: [see docs](https://discordapp.com/developers/docs/reference#snowflakes) and [parser I used](https://github.com/dirigeants/snowflake/blob/master/src/index.ts))
```js
{
    timestamp: "2084-09-06T15:47:35.551Z",
    workerID: 31,
    processID: 31,
    increment: 4095
}
```

So adding the extra step of querying the specific guild doesn't refine the error.

---

On a different note though, it _is_ a useful URL to query.
Because I just noticed that a `/guild/${guildId}` query already includes roles and custom emoji in the response! Which we are interested in. Definitely the custom emoji.

So we could make this a lot simpler:

```js
// This doubles as a validation we have access.
try {
  const {name, emoji} = await fetch.guild(guildId);
} catch (e) {
  throw new Error(
    `Couldn't fetch Server information, maybe the ID is wrong, ` +
      `or the bot doesn't have access?\n${e.message}`
  );
}
```

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
+// @flow
+
+import * as Model from "./models";
+import sortBy from "lodash.sortby";
+
+type fetchEndpoint = (endpoint: string) => Promise<any>;
+type FetchOptions = {|
+  guildId: number,
```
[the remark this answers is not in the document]
Beanow: On second though. I think it would be better if the `guildId` is not part of the Fetcher's options.
Instead, let's keep this at the `discordFetchPlan` level. Because the Fetcher should simply wrap API calls and accept it as an argument when needed. It's part of the mirroring process to know which guildId we've set as "our server".

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
...
+  membersLimit: number,
+  messagesLimit: number,
+  reactionsLimit: number,
+|};
+
+type hasNextPage = boolean;
+export type endCursor = string;
```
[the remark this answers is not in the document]
Beanow: Looks to be fixed!

src/plugins/discord/fetcher.js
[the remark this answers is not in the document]
Beanow: Well aside from it being the value we use as input for `hasNextPage`, I found while making performance improvements in the Discourse plugin that even outside of the Fetcher context, it's a useful parameter for batching SQL queries. So I rather not throw away this information :]

src/plugins/discord/fetchDiscord.js
Beanow: I agree, because Discord is consistent with how it implements pagination, it's a good idea to implement pagination once in a testable function. Definitely support that. :+1:

What I don't like about the current approach:
- It forces you to use side-effects.
- It assumes we want to use chronological order (feed `after` in a loop).

The assumption part is OK, but should just be explicit through the name for example.
`async function fetchAllPagesChronologically(...)` just spitballing, but something explicit like that.

---

> the side-effects of calling the network

I don't think network calls necessarily are a side-effect. The async nature of it is wrapped with Promises, and we can have a clear input and output type for our functions.

`async fetchGuild(guildId: Snowflake): Promise<Guild>` shouldn't have side-effects.

---

So with that in mind, and now having generics for the pagination. Here's an idea how we could generalize pagination in a single function.

```js

// Given we have these types now.
export type EndCursor = Model.Snowflake;
export type PageInfo = {|
  +hasNextPage: boolean,
  +endCursor: EndCursor,
|};
export type ResultPage<T> = {|
  +pageInfo: PageInfo,
  +results: $ReadOnlyArray<T>,
|};
```

We can now define these.

```js
// A function that accepts `after` and returns a generic ResultPage.
export type FetchChronologicalPage<T> = (
  after: Model.Snowflake
) => Promise<ResultPage<T>>;

// A function to loop these, and return just an array of T.
export async function fetchAllPagesChronologically<T>(
  fetchPage: FetchChronologicalPage<T>
): Promise<$ReadOnlyArray<T>> {
  // ... while (nextPage), etc.
}

// Usage example:
async function getAllMessages(
  channel: Model.Snowflake
): Promise<$ReadOnlyArray<Model.Message>> {
  // This is our concrete FetchChronologicalPage<Model.Message>
  // We're basically partially applying the method so we don't
  // need to provide `channel` as an argument anymore.
  const fetchPage = (after: Model.Snowflake) =>
    this.fetcher.messages(channel, after);

  return fetchAllPagesChronologically(fetchPage);
}

// End user example:
async function processMessages(channel) {
  const messages = await getAllMessages(channel);
  for (const message of messages) {
    // ... ingest the message
  }
}
```

I wouldn't be awfully excited to see this implementation though.
For a modestly large server fetching _all_ messages in a single channel could mean you're spending half an hour and need to keep 100.000s of messages in-memory.

Because it's now wrapped in a single `fetchAllPagesChronologically()` you're also giving up control over the execution. You'll just have to wait till it's done.

Here's where async generators have significant advantages. In terms of memory footprint and control. They basically allow you to solve pagination once, while streaming the data, rather than keeping everything in memory. And giving control to the caller.

```js

// Alternative AsyncGenerator
async function* chronologicalStream<T>(
  fetchPage: FetchChronologicalPage<T>
): AsyncGenerator<T, void, void> {
  // Start at "after: 0".
  let currentPageInfo = {
    hasNextPage: true,
    endCursor: Models.ZeroSnowflake,
  };

  // Keep looping so long as there's a next page.
  while (currentPageInfo.hasNextPage) {
    // Gets the next page (with await).
    const {results, pageInfo} = await fetchPage(currentPageInfo.endCursor);

    // Give control back to the caller by yielding the results.
    yield* ensureAsync(results);

    // Update our pageInfo.
    currentPageInfo = pageInfo;
  }
}

/**
 * Takes any Iterable (sync/async) and turns it to an AsyncGenerator.
 * Note: this is mostly used just to deal with a Flow bug.
 */
export async function* ensureAsync<T>(
  it: Iterable<T> | AsyncIterable<T>
): AsyncGenerator<T, void, void> {
  // Flow doesn't correctly support delegating to a sync iterator in async context.
  yield* (it: any);
}

// Usage example:
function streamMessages(
  channel: Model.Snowflake
): AsyncGenerator<Model.Message, void, void> {
  // Same as before, this is our concrete FetchChronologicalPage<Model.Message>
  const fetchPage = (after: Model.Snowflake) =>
    this.fetcher.messages(channel, after);

  return chronologicalStream(fetchPage);
}

// End user example:
async function processMessages(channel) {
  const messages = streamMessages(channel);
  for await (const message of messages) {
    // ... ingest the message
  }
}
```

Beanow: Note, I'm an `AsyncGenerator` shill. I'm not requiring you use it. If you would like to, I'm happy to help and write some utility functions, like the `ensureAsync` one.

If you prefer using Promises of arrays, that's ok with me as well.

Beanow:
Already looking much better :]
Gave some more feedback.

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
...
+|};
+
+/**
+ * Convenience function to find the endCursor value in a response
+ * for Channels, Members, and Reactions.
+ *
+ */
+function getLastId(items: $ReadOnlyArray<any>): endCursor {
+  if (items.length > 0) {
+    return sortBy(items, (x) => x.id)[items.length - 1].id;
```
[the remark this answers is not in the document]
BrianLitwin: Updated to rely on the last item to provide the end cursor.

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
+// @flow
+
+import * as Model from "./models";
+import sortBy from "lodash.sortby";
```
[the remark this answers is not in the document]
BrianLitwin: Obsoleted, as I removed the sorting logic to find the end cursor.

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Convenience function to find the endCursor value in a response
+ * for Channels, Members, and Reactions.
+ *
+ */
+function getLastId(items: $ReadOnlyArray<any>): endCursor {
+  if (items.length > 0) {
+    return sortBy(items, (x) => x.id)[items.length - 1].id;
+  } else {
+    return "";
```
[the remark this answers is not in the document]
BrianLitwin: Updated to use `null`

src/plugins/discord/models.js
```diff
@@ -0,0 +1,112 @@
...
+
+  // Custom emoji.
+  return `${name}:${id}`;
+}
+
+export function refToEmoji(ref: EmojiRef): Emoji {
+  const [name, id] = ref.split(":");
+  if (!id) return {id: null, name};
+  return {id, name};
+}
```
decentralion: This code seems to lack tests

BrianLitwin: Added tests for helper functions in `models.js`

src/plugins/discord/fetchDiscord.js
BrianLitwin: Obsoleted

src/plugins/discord/fetcher.js
BrianLitwin: Updated

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,147 @@
...
+      id: x.id,
+      name: x.name,
+      permissions: x.permissions,
+    }));
+  }
+
+  async guild(): Promise<any> {
+    const {guildId} = this._options;
+    return this._fetch(`guilds/${guildId}`);
+  }
```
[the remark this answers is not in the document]
BrianLitwin: Deleted this method. As you pointed out, it's not effective in checking whether a guid exists.

src/plugins/discord/fetcher.js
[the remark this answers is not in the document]
BrianLitwin: My rationale for this `options` object: 

```js
type FetchOptions = {|
  guildId: number,
  membersLimit: number,
  messagesLimit: number,
  reactionsLimit: number,
|};

```

These will be constants for the duration of a `Fetchers` use-case. I don't think we should parameterize variables that won't change, and it makes it easier to reason about the function parameters that _will_ change by keeping constants out of function signatures. E.g., we're not going to use the same `Fetcher` to fetch different `guildId`s.  It also reduces the surface area where mistakes could occur by providing these constants in one place (at the classes' initialization) instead of several (whenever we need to call a method that requires `guildId`). 

Not a strong opinion however. If we wanted to remove the `guildId` from the `options` object, I'd think we want to do the same with the `limit` variables as well. 

What do you think?

src/plugins/discord/fetcher.js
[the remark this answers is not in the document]
BrianLitwin: Removed this function, as its use case (checking whether a `Guild` exists) is based on a flawed assumption.

BrianLitwin:
@decentralion @Beanow This is RFAL; all of @Beanow 's comments relative to the code in this PR have been addressed.

src/plugins/discord/models.js
```diff
@@ -0,0 +1,125 @@
...
+  // Custom emoji.
+  return `${name}:${id}`;
+}
+
+// Determines whether the message was created by a webhook or a Discord User
+export function isAuthoredByNonUser(rawMessage: {
+  +webhook_id?: Snowflake,
+}): boolean {
+  return rawMessage.webhook_id != null || false;
+}
```
decentralion: Hmm, why do we need this?

src/plugins/discord/models.js
```diff
@@ -0,0 +1,125 @@
...
+
+export type Message = {|
```
decentralion: This is just a representation of the type that comes over the wire from Discord, right? Can we link directly to the right section of the docs?

src/plugins/discord/models.js
```diff
@@ -0,0 +1,125 @@
...
+  +timestampMs: number,
+  +content: string,
+  // Normally includes reaction counters, but we don't care about counters.
+  // We could filter based on which types of emoji have been added though.
+  +reactionEmoji: $ReadOnlyArray<Emoji>,
+  // Snowflake of user IDs.
+  +mentions: $ReadOnlyArray<Snowflake>,
+|};
+
+export type Reaction = {|
```
decentralion: Link to docs?

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,149 @@
...
+ *
+ */
+type FetchOptions = {|
+  guildId: Model.Snowflake,
+  membersLimit: number,
+  messagesLimit: number,
+  reactionsLimit: number,
+|};
+
+export type EndCursor = Model.Snowflake | null;
```
decentralion: The way EndCursor is handled seems inconsistently. We declare that the type is either Snowflake or null, but the methods which take `EndCursor`s error on the null case. 

Therefore I propose either:

- We make the EndCursor type just a snowflake, and update the PageInfo so that it has `endCursor | null`
or:
- We make the methods that take EndCursor cleanly accept the null case (by just not appending the `after` flag)

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,149 @@
...
+      name: x.name,
+      type: Model.channelTypeFromId(x.type),
+    }));
+  }
+
+  async members(after: EndCursor): Promise<ResultPage<Model.GuildMember>> {
+    const {membersLimit, guildId} = this._options;
+    const endpoint = `/guilds/${guildId}/members?after=${NullUtil.get(
+      after
+    )}&limit=${membersLimit}`;
```
decentralion: Rather than erroring when `after==null`, we could just only append the `after` query slug when the cursor is not null, right?

BrianLitwin: I think we'd always want an after parameter; but this was obsoleted by updating the `EndCursor`. These endpoints are easier on the eyes now.

src/plugins/discord/fetcher.js
BrianLitwin: Updated, good suggestion

src/plugins/discord/models.js
BrianLitwin: Updated

src/plugins/discord/models.js
BrianLitwin: I added doc links to all the types we've added that are representations of Discord objects. The only one that's a little fuzzy is `Reaction`, because really it's just an Emoji with the channel, message, and author ids attached.

src/plugins/discord/models.js
BrianLitwin: Discord messages can be created with [webhooks](https://discordapp.com/developers/docs/resources/webhook), and [messages](https://discordapp.com/developers/docs/resources/channel#message-object) have a field that specifies whether the messages was authored by a webhook.  We can use this info to decide how to create edges and assign authorship of messages. Is that right @Beanow ?

src/plugins/discord/models.js
```diff
@@ -0,0 +1,125 @@
...
+    default: {
+      throw new Error(`Unknown channel type ID: ${id}`);
+    }
+  }
+}
+
+export type Guild = {|
+  +id: Snowflake,
+  +name: string,
+  +permissions: number,
```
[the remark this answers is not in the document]
BrianLitwin: Good Q. I'm not sure. @Beanow ?

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,149 @@
...
+
+import * as Model from "./models";
+import * as NullUtil from "../../util/null";
+
+type FetchEndpoint = (endpoint: string) => Promise<any>;
+
+/**
+ * Provide the Guild ID to fetch against, and the 'limit'
+ * parameter when fetching GuildMembers, Messages, and Reactions.
+ *
```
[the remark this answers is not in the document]
BrianLitwin: Do you mean the empty line after the comment or that the whole comment is redundant?

src/plugins/discord/models.js
Beanow: That's right :]
It changes the meaning of the `authorId: Snowflake` field. When a message is created by a webhook, the `authorId` is not a user ID, but the webhook's ID. It's similar to why we would need to know when a commit was by a bot account or actual user on GitHub.

Beanow: That's right :]
It changes the meaning of the `authorId: Snowflake` field. When a message is created by a webhook, the `authorId` is not a user ID, but the webhook's ID. It's similar to why we would need to know when a commit was by a bot account or actual user on GitHub.

src/plugins/discord/models.js
```diff
@@ -0,0 +1,125 @@
...
+ * This function returns the appropriate reference based on this
+ * generic vs custom distinction.
+ */
+export function emojiToRef({id, name}: Emoji): EmojiRef {
+  // Built-in emoji, unicode names.
+  if (!id) return name;
+
+  // Custom emoji.
+  return `${name}:${id}`;
+}
```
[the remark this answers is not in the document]
Beanow: We've not come up with either of the two representations. Both are Discord's.

The collapsed string variant is used in a URL path to query reactions.
https://github.com/sourcecred/sourcecred/pull/1703/files#diff-8047b9df542fc6804d91afa152e26c87R127-R128

The object variant is the shape the API returns. For instance in the reaction summaries on messages, or the list of custom emoji in a guild.

src/plugins/discord/models.js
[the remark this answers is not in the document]
Beanow: Because we require permissions :smile:

https://github.com/sourcecred/sourcecred/blob/discord/src/plugins/discord/Readme.md

> Permissions bot need:
>
> -  View Channels
> -  Read Message History
>
> `66560` as integer

src/plugins/discord/fetchDiscord.js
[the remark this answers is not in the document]
Beanow: This fell out of the PR scope, resolving.

src/plugins/discord/fetchDiscord.js
```diff
@@ -0,0 +1,198 @@
...
+        return Promise.resolve(response.json());
+    }
+  });
+}
+
+// If retry = true in tryDiscordFetch reject, will re-attempt the fetch.
+function retryDiscordFetch(fetch, url, fetchOptions) {
+  return new Promise((resolve, reject) => {
+    const operation = retry.operation();
+    operation.attempt(() => {
```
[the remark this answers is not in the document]
Beanow: Fell out of scope, resolving.

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,142 @@
...
+ */
+type FetchOptions = {|
+  guildId: Model.Snowflake,
+  membersLimit: number,
+  messagesLimit: number,
+  reactionsLimit: number,
+|};
+
+export type EndCursor = Model.Snowflake;
```
Beanow: Nit: I think the type alias doesn't make the function signatures easier to understand.
Let's just call them snowflakes?

src/plugins/discord/fetchDiscord.js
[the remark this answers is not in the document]
Beanow: Fell out of scope.

src/plugins/discord/testUtil.js
```diff
@@ -0,0 +1,33 @@
...
+export const testMember = (userId: number) => ({
+  user: {
+    id: userId,
+    username: "username",
+    discriminator: "disc",
+    bot: true,
+  },
+  nick: "nickname",
+  roles: ["test role"],
+});
```
Beanow: Only one test file is currently using this. So I think it would be better to have this in `fetcher.test.js`. It will make the test file easier to read.

Conventionally such helper functions or example data go at the top of the file unless they are very large snapshots or generic utils.

Beanow: It's important to use Flow types for example data like this. It will help you validate the example data is actually compliant with the models we're expecting. And will help prevent mistakes if we update these models.

In this case I would suggest using the function return type for that. E.g.
```diff
- const testChannel = (id: number) => ({ ... });
+ const testChannel = (id: number): Model.Channel => ({ ... });
```

In this case, it should actually fail for every function here, because a `Snowflake` is a string, and we're using `number` instead.

Beanow:
> For the Discourse plugin, we have code that generates "real" snapshots of actual API return data from our test Discourse server. That way, we can test our code against realistic data, and if Discourse ever changes its data format, it'd be easy for us to regenerate snapshots and then validate our code against the format.
> 
> You can see the machinery for generating these snapshots here: https://github.com/sourcecred/sourcecred/blob/master/src/plugins/discourse/update_discourse_api_snapshots.sh
> 
> Do you think it's feasible to do this for Discord too?

This should be feasible :]
The `fetchEndpoint` constructor argument can be switched for a mock variant.
Now that we've shrunk the scope of this PR, I think it's better not to add more to it. Would prefer to see this in a follow up.

src/plugins/discord/testUtil.js
BrianLitwin: Updated; was originally using this in multiple files, but good suggestion. I move them into the smallest required scope.

BrianLitwin: Great point! Updated.

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,142 @@
...
+
+type FetchEndpoint = (endpoint: string) => Promise<any>;
+
+/**
+ * Provide the Guild ID to fetch against, and the 'limit'
+ * parameter when fetching GuildMembers, Messages, and Reactions.
+ *
+ */
+type FetchOptions = {|
+  guildId: Model.Snowflake,
```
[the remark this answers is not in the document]
BrianLitwin: My inclination was that `guildId` is a constant, and there's no reason to parameterize a constant; just increases the surface area on which we could error. 

That caveat aside, I updated according to your suggestion.

src/plugins/discord/fetcher.js
BrianLitwin: Updated

src/plugins/discord/fetcher.js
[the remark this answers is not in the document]
Beanow: Thanks for the change!

I found the previous thread now: https://github.com/sourcecred/sourcecred/pull/1703#discussion_r392710220

> My rationale for this `options` object:
> 
> ```js
> type FetchOptions = {|
>   guildId: number,
>   membersLimit: number,
>   messagesLimit: number,
>   reactionsLimit: number,
> |};
> ```
> 
> These will be constants for the duration of a `Fetchers` use-case. I don't think we should parameterize variables that won't change, and it makes it easier to reason about the function parameters that _will_ change by keeping constants out of function signatures. E.g., we're not going to use the same `Fetcher` to fetch different `guildId`s. It also reduces the surface area where mistakes could occur by providing these constants in one place (at the classes' initialization) instead of several (whenever we need to call a method that requires `guildId`).
> 
> Not a strong opinion however. If we wanted to remove the `guildId` from the `options` object, I'd think we want to do the same with the `limit` variables as well.
> 
> What do you think?

Gotcha! Yeah I think there's no technical reason why Fetcher should require that `guildId` doesn't change. By doing that we're making Fetcher responsible for the use-case of Fetcher. That responsibility should be with what's using Fetcher.

If the use-case is mirroring a guild's data. Then the guild mirroring function should add constraints like `guildId` is constant for the whole of this process.

An example of a different use-case is to answer the question "with this bot token, which guilds do I have permissions for?". We might use such a query for validating the token, or to help users configure the plugin. It wouldn't make sense if this required a constant `guildId` be given to the constructor.

src/plugins/discord/testUtil.js
Beanow: I think this might have been undone in some places by moving the functions? :smile:
I'll comment inline on the ones I can spot that aren't typed yet.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+  describe("fetch guild", () => {
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      await fetcher.guild("1");
+      expect(fetch.mock.calls[0]).toEqual(["guilds/1"]);
+    });
+
+    it("handles response", async () => {
+      const expected = {id: 1, name: "guildname", permissions: []};
```
Beanow: Missing type `const exptected: Guild =`

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+  describe("fetch channels", () => {
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      await fetcher.channels("1");
+      expect(fetch.mock.calls[0]).toEqual(["/guilds/1/channels"]);
+    });
+
+    it("handles response", async () => {
+      const testChannel = (id: string) => ({
```
Beanow: Missing type `(id: Snowflake): Channel =>`

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+      const expected = [testChannelObj("1")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      const data = await fetcher.channels("0");
+      expect(data).toEqual(expected);
+    });
+  });
+
+  describe("fetch members", () => {
+    const testMember = (userId: number) => ({
```
Beanow: Missing type `(userId: Snowflake): Member`

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+        return hasNextPage;
+      };
+
+      it("next page = true", async () => {
+        expect(await testHasNextPage(true)).toBe(true);
+      });
+
+      it("next page = false", async () => {
+        expect(await testHasNextPage(false)).toBe(false);
+      });
```
Beanow: For readability, I believe it would be better to side with copy pasted code if it makes your test more straightforward. If you've seen examples of how I write tests with [Given, When, Then](https://en.wikipedia.org/wiki/Given-When-Then) it would work out something like this:

Given membersLimit=1 and 1 member
When fetching members
Then hasNextPage should be true

Given membersLimit=2 and 1 member
When fetching members
Then hasNextPage should be false

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+    });
+  });
+
+  describe("fetch reactions", () => {
+    const emoji = {id: "1", name: "emojiname"};
```
Beanow: Missing type `const emoji: Emoji =`

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, options());
+      await fetcher.reactions("1", "2", emoji, "0");
+      expect(fetch.mock.calls[0]).toEqual([
+        `/channels/1/messages/2/reactions/emojiname:1?after=0&limit=2`,
+      ]);
+    });
+
+    it("handles response", async () => {
+      const response = [{id: 123, emoji}];
```
Beanow: Missing type: `const response: Emoji[]`

Note: I think this would show a bug, because I think there's a missing spread operator.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+      const {results} = await fetcher.reactions("3", "2", emoji, "0");
+      const expected = [{emoji, channelId: "3", messageId: "2", authorId: 123}];
+      expect(results).toEqual(expected);
+    });
+
+    it("returns correct endCursor", async () => {
+      const response = [
+        {id: 1, emoji},
+        {id: 2, emoji},
+      ];
```
Beanow: Missing type: `const response: Emoji[]`

Note: I think this would show a bug, because I think there's a missing spread operator.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+      const fetcher = new Fetcher(fetch, options());
+      const {endCursor} = (
+        await fetcher.reactions("1", "2", emoji, "0")
+      ).pageInfo;
+      expect(endCursor).toBe(2);
+    });
+
+    describe("next page", () => {
+      const testHasNextPage = async (nextPage) => {
+        const response = [{id: 1, emoji}];
```
Beanow: Missing type: `const response: Emoji[]`

Note: I think this would show a bug, because I think there's a missing spread operator.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+      });
+
+      it("next page = false", async () => {
+        expect(await testHasNextPage(false)).toBe(false);
+      });
+    });
+  });
+
+  describe("fetch messages", () => {
+    const testMessage = (id: number) => ({
```
Beanow: Missing type `(id: Snowflake): Message =>`

Note: I think would fail because of numbers being used for Snowflakes/

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+      await fetcher.messages("1", "0");
+      expect(fetch.mock.calls[0]).toEqual([
+        "/channels/1/messages?after=0&limit=2",
+      ]);
+    });
+
+    it("handles response", async () => {
+      const response = [testMessage(1)];
+
+      const expected = [
```
Beanow: Missing type `const expected: Message[]`
Would probably error.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,268 @@
...
+        return hasNextPage;
+      };
+
+      it("next page = true", async () => {
+        expect(await testHasNextPage(true)).toBe(true);
+      });
+
+      it("next page = false", async () => {
+        expect(await testHasNextPage(false)).toBe(false);
+      });
```
Beanow: Same as before, better to copy-paste for readability and explicit tests.

BrianLitwin:
@Beanow Addressed your comments. Thanks for the review! I'm on board with using @decentralion testing strategy here or in a follow-up; working on it either way.

BrianLitwin:
@Beanow , I think I've addressed your comments. Where the tests mock responses from the server, I assigned a type of `any`, and added a note at the top of the test file indicating as much.

src/plugins/discord/fetcher.js
Beanow: Ah, I didn't mean to rename `Model.Snowflake` to `Snowflake`. I meant to remove the `EndCursor` type and use `(Model.)Snowflake` instead.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+    it("passes correct endpoint", async () => {
+      const fetch = jest.fn(() => Promise.resolve([]));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      await fetcher.channels("1");
+      expect(fetch.mock.calls[0]).toEqual(["/guilds/1/channels"]);
+    });
+
+    it("handles response", async () => {
+      // response object from server, prior to type transformation
+      const testChannel: any = (id: string) => ({
```
Beanow: There's an issue with this type pattern.
Originally this was:
```
const testChannel = (id: string) => //...
```
Will be typed as `const testChannel` is a function with `(id: string)` as args and an implied `any` return type.
Now this would be cast from a Function to an `any`. This is less strict than it should be.

The intention was to make the **return type** an explicit `any`. With a comment to mention, because we're mocking a raw JSON server response (not a mapped model) the type is intentionally `any`.

So the pattern should be changed like:
```diff
- // Turns a function into any
- const testChannel: any = (id: string) => // ...

+ // Sets an explicit any return type
+ const testChannel = (id: string): any => // ...
```

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+      const expected: Channel[] = [testChannelObj("1")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, defaultOptions());
+      const data = await fetcher.channels("0");
+      expect(data).toEqual(expected);
+    });
+  });
+
+  describe("fetch members", () => {
+    const testMember: any = (userId: string) => ({
```
Beanow: Same as previous comment

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+    it("handles response", async () => {
+      const response: GuildMember[] = [testMember("1")];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {results} = await fetcher.members("1", "0");
+      expect(results).toEqual(response);
+    });
+
+    it("returns correct endCursor", async () => {
+      const response: any = [testMember("1"), testMember("2")];
```
Beanow: We are losing type information here as well.
"Downgrading" from `any[]` to `any`.

For consistency, let's use `any[]`.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+          membersLimit: 1,
+          messagesLimit: 100,
+          reactionsLimit: 100,
+        });
+        const {hasNextPage} = (await fetcher.members("1", "0")).pageInfo;
+        expect(hasNextPage).toBe(true);
+      });
+
+      it("next page = false", async () => {
+        const response: any = [testMember("1")];
```
Beanow: Same as above.

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+      const fetcher = new Fetcher(fetch, options());
+      await fetcher.reactions("1", "2", emoji, "0");
+      expect(fetch.mock.calls[0]).toEqual([
+        `/channels/1/messages/2/reactions/emojiname:1?after=0&limit=2`,
+      ]);
+    });
+
+    it("handles response", async () => {
+      // response object from server, prior to type transformation
+      const response: any = [{id: "123", emoji}];
```
Beanow: Same as above.

src/plugins/discord/fetcher.test.js
BrianLitwin: updated

src/plugins/discord/fetcher.test.js
BrianLitwin: updated

src/plugins/discord/fetcher.test.js
BrianLitwin: updated

src/plugins/discord/fetcher.test.js
BrianLitwin: Updated, good catch, this was an oversight.

src/plugins/discord/fetcher.test.js
BrianLitwin: Updated

BrianLitwin:
RFAL @Beanow

src/plugins/discord/fetcher.js
```diff
@@ -0,0 +1,143 @@
...
+  messagesLimit: number,
+  reactionsLimit: number,
+|};
+
+export type EndCursor = Snowflake;
+
+export type PageInfo = {|
+  +hasNextPage: boolean,
+  +endCursor: EndCursor | null,
+|};
```
Beanow: Would like to stick with `Snowflake`, rather than alias it to `EndCursor`.

src/plugins/discord/fetcher.test.js
Beanow: Can this be typed as `const emoji: Emoji` or does it need to be a server response type?

src/plugins/discord/fetcher.js
BrianLitwin: Good point; you mentioned that earlier, I didn't quite understand what you meant. Updated.

src/plugins/discord/models.test.js
```diff
@@ -0,0 +1,28 @@
...
+        expect(emojiToRef({id: "testEmojiId", name: "testEmojiName"})).toBe(
+          "testEmojiName:testEmojiId"
+        );
+      });
+    });
+    describe("isAuthoredByNonUser", () => {
+      it("returns true if webhook_id property is provided in message", () => {
+        expect(isAuthoredByNonUser({webhook_id: "0"})).toBe(true);
+      });
+      it("returns true if webhook_id property is not provided in message", () => {
```
Beanow: Nit: returns true -> returns false

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+      const expected: Reaction[] = [
+        {emoji, channelId: "3", messageId: "2", authorId: "123"},
+      ];
+      expect(results).toEqual(expected);
+    });
+
+    it("returns correct endCursor", async () => {
+      const response: any[] = [
+        {id: 1, emoji},
+        {id: 2, emoji},
```
Beanow: Ids should be Snowflakes (strings).

src/plugins/discord/fetcher.test.js
```diff
@@ -0,0 +1,299 @@
...
+      ];
+      const fetch = jest.fn(() => Promise.resolve(response));
+      const fetcher = new Fetcher(fetch, options());
+      const {endCursor} = (
+        await fetcher.reactions("1", "2", emoji, "0")
+      ).pageInfo;
+      expect(endCursor).toBe(2);
```
Beanow: This `2` is ambiguous. It could be the "2" argument of the function or `id: "2"` from the API response.
Can you make sure it's a distinct value?

src/plugins/discord/models.test.js
BrianLitwin: good catch!

src/plugins/discord/fetcher.test.js
BrianLitwin: Good point, updated

src/plugins/discord/fetcher.test.js
BrianLitwin: Indeed, updated

Issue #1708, opened 2020-03-18
Attack Vector - any team member can edit PR descriptions


hammadj:
Right now, I could go into all the old PRs and tag myself in the description with "Paired-with @hammadj" to steal cred from all the contributors. Not sure what the best solution to this would be, but these are the permission levels available in GitHub: 
![image](https://user-images.githubusercontent.com/7143583/76917310-a1589700-6888-11ea-8f13-593fcf9d0df4.png)

decentralion:
@wchargin You're right that this attack is not part of our current threat model. That said, I think it's good for us to track attacks like this regardless--would be good to have some docs that track all the attacks we know about. (Or, we could make an "attack" label on GitHub?")

PR #1709, opened 2020-03-18
Add grain payout strategies

decentralion:
This module adds "payout strategies" for computing how to divide grain
between participants based on their scores and their lifetime earnings.

We have three payout strategies:

- fixedAmount: Given a fixed amount of Grain to distribute, split it
  proportional to cred scores.

- fixedRatio: Given a target level of grainPerCred, mint Grain so that
  every contributors' lifetime earnings is at least grainPerCred *
  theirCred. Takes past earnings into account.

- underpayment: Given a fixed amount of Grain, distribute it according
  to cred scores, prioritizing paying users who were "underpaid" in the
  past (i.e. they have a smaller proportion of lifetime earnings than is
  suggested by their cred score).

Test plan: Included somewhat thorough unit tests. I'd like to have
reviewers carefully review the algorithms.

### DIFF
--- a/src/grain/payouts.js
+++ b/src/grain/payouts.js
@@ -1,0 +1,169 @@
+// @flow
+/* global BigInt */
+
+/**
+ * This module contains "payout strategies" for distributing Grain based on scores.
+ *
+ * We currently have three payout strategies:
+ * - fixedAmount: Given a fixed amount of Grain to distribute, split it proportional
+ *   to cred scores.
+ *
+ * - fixedRatio: Given a target level of grainPerCred, mint Grain so that every
+ *   contributors' lifetime earnings is at least grainPerCred * theirCred. Takes past
+ *   earnings into account.
+ *
+ * - underpayment: Given a fixed amount of Grain, distribute it according to cred scores,
+ *   prioritizing paying users who were "underpaid" in the past (i.e. they have a smaller
+ *   proportion of lifetime earnings than is suggested by their cred score).
+ */
+import {sum} from "d3-array";
+import * as MapUtil from "../util/map";
+import {type NodeAddressT} from "../core/graph";
+import {type Grain, ZERO, ONE, multiplyFloat} from "./grain";
+
+/**
+ * Distribute an amount of Grain, strictly in proportion to the scores.
+ *
+ * This might be a good fit for distributing a fixed grain per week directly
+ * in proportion to the weekly cred score.
+ */
+export function fixedAmount(
+  harvestAmount: Grain,
+  scores: Map<NodeAddressT, number>
+): Map<NodeAddressT, Grain> {
+  if (harvestAmount < ZERO) {
+    throw new Error(`invalid harvestAmount: ${String(harvestAmount)}`);
+  }
+  if (scores.size === 0) {
+    return new Map();
+  }
+  const totalScore = sum(Array.from(scores.values()));
+  if (totalScore === 0) {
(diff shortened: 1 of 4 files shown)


src/grain/payouts.js
```diff
@@ -0,0 +1,180 @@
...
+  grainPerScore: number,
+  scores: Map<NodeAddressT, number>,
+  earnings: Map<NodeAddressT, Grain>
+): Map<NodeAddressT, Grain> {
+  if (!isFinite(grainPerScore) || grainPerScore < 0) {
+    throw new Error(`invalid grainPerScore: ${grainPerScore}`);
+  }
+  return MapUtil.mapValues(scores, (addr, score) => {
+    const earned = earnings.get(addr) || zero;
+    const target = _multiply(one, grainPerScore * score);
```
hammadj: Would be nice if the `zero` and `one` constants were all caps for the sake of readability to make it clear that they are constants instead of parameters or variables
```suggestion
    const earned = earnings.get(addr) || ZERO;
    const target = _multiply(ONE, grainPerScore * score);
```

src/grain/payouts.js
```diff
@@ -0,0 +1,180 @@
...
+  grainPerScore: number,
+  scores: Map<NodeAddressT, number>,
+  earnings: Map<NodeAddressT, Grain>
+): Map<NodeAddressT, Grain> {
+  if (!isFinite(grainPerScore) || grainPerScore < 0) {
+    throw new Error(`invalid grainPerScore: ${grainPerScore}`);
+  }
+  return MapUtil.mapValues(scores, (addr, score) => {
+    const earned = earnings.get(addr) || ZERO;
+    const target = _multiply(ONE, grainPerScore * score);
```
hammadj: When I first read "multiply" it makes it seem like thats the function actually doing the multiplication to calculate the target grain, when in reality the actual calculation is just the `grainPerScore * score`. 

Seems like the intent of this is to "convert" the float value to the BigInt value for grain. Would make it more readable if you wrapped this in a helper function, something like:
```ts
const toGrain = (score: number): Grain => _multiply(ONE, score);
```

I imagine it would get re-used in other places as well. Seems like you already had something like this in the tests

decentralion: Hmm, the "conversion" is actually a multiplication though. Since `ONE === 10n**18n`. If we ignore the type conversions, the line is something like `10n**18n * grainPerScore * score`. That said, I can see how a `floatToGrain` type method could be useful; you're right that I define such a method in the test code. I'll sleep on it and maybe add it in the morning.

Also, generally speaking, we use "request changes" when the reviewer wants to block approval pending fixing a serious issue; here's a good example of [@wchargin blocking a PR that introduced a security flaw](https://github.com/sourcecred/sourcecred/pull/1265#pullrequestreview-273471576). When there's just nits, we'll usually either just comment, or [nit-and-approve](https://github.com/sourcecred/docs/blob/master/culture/review_culture.md#consider-nit-and-approve).

decentralion: @hammadj I factored your suggestion into another PR: https://github.com/sourcecred/sourcecred/pull/1715

src/grain/payouts.js
decentralion: Was resolved in https://github.com/sourcecred/sourcecred/pull/1710.

wchargin:
Is this meant to be diffbased against #1708?

decentralion:
Superseded by #1743

PR #1710, opened 2020-03-18
Change lowercase const names to all caps in grain.js

hammadj:
Improves readability of the code and makes it easy to understand that these are constants instead of parameters/variables

Test plan: This is a rename; verify that CI passes.

### DIFF
--- a/src/grain/grain.js
+++ b/src/grain/grain.js
@@ -27,14 +27,14 @@
 export type Grain = number;
 
 // $ExpectFlowError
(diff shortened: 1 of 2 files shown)


decentralion:
Per our [contributing guidelines](https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#test-plan), please add a test plan. In this case, the trivial test plan suffices:

> Test plan: This is a rename; verify that CI passes.

PR #1715, opened 2020-03-21
Add math helper methods to the `grain` module

decentralion:
This commit adds two math helpers to the `grain` module:

`multiplyFloat` has a canonical way to multiply a grain amount by a
floating point number, which is useful e.g. if we need to multiply grain
amounts by cred ratios. For now, we have a naive approach which coerces
the grain into a float, does float multiplication, and then converts
back to grain.

This is imprecise, but in testing (see [notebook]) the absolute error
amounts are very small, i.e. in the atto-grain to femto-grain range.
Because we will be using this in conjunction with payout strategies that
implement explicit payment-error-correction over time (i.e. we calculate
who has been historically under-paid and give them more reward), these
insignificant errors will tend to disappear rather than accumulate over
time.

`fromFloat` is a helper method for creating a grain balance directly
from a float. It's basically an alias for `multiply(ONE, float)` and
is useful for test code.

Test plan: Unit tests included (and re-factored to use the new helpers).
See attached notebook for some exploration of error values.

[notebook]: https://observablehq.com/@decentralion/grain-arithmetic

Inspired by [this pull request review](https://github.com/sourcecred/sourcecred/pull/1709#pullrequestreview-376583943) by @hammadj.

Thanks to @wchargin for discussing grain multiplication with me at depth.

### DIFF
--- a/src/grain/grain.js
+++ b/src/grain/grain.js
@@ -108,3 +108,36 @@ export function format(
   }
   return digits.join("") + suffix;
 }
+
+/**
+ * Multiply a grain amount by a floating point number.
+ *
+ * Use this method when you need to multiply a grain balance by a floating
+ * point number, e.g. a ratio.
+ *
+ * Note that this method is imprecise. It is not safe to assume, for example,
+ * that `multiply(g, 1/3) + multiply(g, 2/3) === g` due to loss of precision.
+ * However, the errors will be small in absolute terms (i.e. tiny compared to
+ * one full grain).
+ *
+ * See some messy analysis of the numerical errors here:
+ * https://observablehq.com/@decentralion/grain-arithmetic
+ */
+export function multiplyFloat(grain: Grain, num: number): Grain {
+  return BigInt(Math.floor(Number(grain) * num));
+}
+
+/**
+ * Aproximately create a grain balance from a float.
+ *
+ * This method tries to convert the floating point `amt` into a grain
+ * balance. For example, `grain(1)` approximately equals `ONE`.
+ *
+ * Do not assume this will be precise! For example, `grain(0.1337)` results in
+ * `133700000000000016n`. This method is intended for test code.
+ *
+ * This is a shorthand for `multiplyFloat(ONE, amt)`.
+ */
+export function fromFloat(f: number): Grain {
+  return multiplyFloat(ONE, f);
+}
--- a/src/grain/grain.test.js
+++ b/src/grain/grain.test.js
@@ -1,68 +1,64 @@
 // @flow
 
-import {format, ONE, DECIMAL_PRECISION, ZERO} from "./grain";
+import {
+  format,
+  ONE,
+  DECIMAL_PRECISION,
+  ZERO,
+  fromFloat,
+  multiplyFloat,
+} from "./grain";
 
 describe("src/grain/grain", () => {
   describe("format", () => {
     // $ExpectFlowError
-    const pointOne = ONE / 10n;
-    // $ExpectFlowError
-    const onePointFive = pointOne * 15n;
-    // $ExpectFlowError
     const almostOne = ONE - 1n;
-    // $ExpectFlowError
-    const fortyTwo = ONE * 42n;
-    // $ExpectFlowError
-    const negative = -1n;
-    // $ExpectFlowError
-    const leet = ONE * 1337n;
-    // $ExpectFlowError
-    const leetAndSpecial = leet * 1000n + fortyTwo + fortyTwo / 100n;
 
     it("correctly rounds to smallest integer when decimals==0", () => {
       expect(format(ZERO)).toEqual("0g");
-      expect(format(pointOne)).toEqual("0g");
+      expect(format(fromFloat(0.1))).toEqual("0g");
       expect(format(almostOne)).toEqual("0g");
       expect(format(ONE)).toEqual("1g");
-      expect(format(onePointFive)).toEqual("1g");
-      expect(format(fortyTwo)).toEqual("42g");
+      expect(format(fromFloat(1.5))).toEqual("1g");
+      expect(format(fromFloat(42))).toEqual("42g");
     });
     it("correctly adds comma formatting for large numbers", () => {
-      expect(format(leet)).toEqual("1,337g");
-      expect(format(leet, 1)).toEqual("1,337.0g");
-      expect(format(leet + pointOne)).toEqual("1,337g");
-      expect(format(leet + pointOne, 1)).toEqual("1,337.1g");
-      expect(format(leetAndSpecial, 0)).toEqual("1,337,042g");
-      expect(format(leetAndSpecial, 2)).toEqual("1,337,042.42g");
+      expect(format(fromFloat(1337))).toEqual("1,337g");
+      expect(format(fromFloat(1337), 1)).toEqual("1,337.0g");
+      expect(format(fromFloat(1337.11))).toEqual("1,337g");
+      expect(format(fromFloat(1337.11), 1)).toEqual("1,337.1g");
+      expect(format(fromFloat(1337042.42), 0)).toEqual("1,337,042g");
+      expect(format(fromFloat(1337042.42), 2)).toEqual("1,337,042.42g");
     });
     it("correctly handles negative numbers", () => {
-      expect(format(negative * pointOne)).toEqual("-0g");
-      expect(format(negative * onePointFive)).toEqual("-1g");
-      expect(format(negative * fortyTwo)).toEqual("-42g");
-      expect(format(negative * onePointFive, 1)).toEqual("-1.5g");
-      expect(format(negative * onePointFive, 1)).toEqual("-1.5g");
-      expect(format(negative * leetAndSpecial, 0)).toEqual("-1,337,042g");
-      expect(format(negative * leetAndSpecial, 2)).toEqual("-1,337,042.42g");
+      expect(format(fromFloat(-0.1))).toEqual("-0g");
+      expect(format(fromFloat(-1.5))).toEqual("-1g");
+      expect(format(fromFloat(-42))).toEqual("-42g");
+      expect(format(fromFloat(-1.5), 1)).toEqual("-1.5g");
+      expect(format(fromFloat(-1.5), 1)).toEqual("-1.5g");
+      expect(format(fromFloat(-1337042.42), 0)).toEqual("-1,337,042g");
+      expect(format(fromFloat(-1337042.42), 2)).toEqual("-1,337,042.42g");
     });
     it("handles full precision", () => {
       expect(format(ZERO, DECIMAL_PRECISION)).toEqual("0.000000000000000000g");
       expect(format(ONE, DECIMAL_PRECISION)).toEqual("1.000000000000000000g");
-      expect(format(pointOne, DECIMAL_PRECISION)).toEqual(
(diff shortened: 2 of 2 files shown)


src/grain/grain.js
```diff
@@ -108,3 +108,36 @@ export function format(
...
+ * Note that this method is imprecise. It is not safe to assume, for example,
+ * that `multiply(g, 1/3) + multiply(g, 2/3) === g` due to loss of precision.
+ * However, the errors will be small in absolute terms (i.e. tiny compared to
+ * one full grain).
+ *
+ * See some messy analysis of the numerical errors here:
+ * https://observablehq.com/@decentralion/grain-arithmetic
+ */
+export function multiplyFloat(grain: Grain, num: number): Grain {
+  return BigInt(Math.floor(Number(grain) * num));
```
wchargin: Hmm… this conversion is a bit awkward considering that `grain` will
exceed `Number.MAX_SAFE_INTEGER` (2^53 − 1 ≈ 10^16 < 10^18). This means
that it’s not even the case that `multiplyFloat(g, 1) === g`, even for
such pedestrian values of `g` as `ONE + 1n`:

```javascript
> g = (10n ** 18n) + 1n;
1000000000000000001n
> BigInt(Math.floor(Number(g) * 1))
1000000000000000000n
```

(And, in fact, all values between `ONE - 64n` and `ONE + 64n` inclusive
have the same image under `(g) => multiplyFloat(g, 1)`.)

That is, this function is approximate not just in the floating point
argument but in the grain argument itself.

This deficit is not fundamental. We can do the calculation entirely in
bigint space, destructuring the float into its integer components; quick
sketch:

```javascript
function mul(grain: Grain, fac: number) {
  const {sgn, exponent, mantissa} = decompose(fac);
  // all small integers:
  //   - sgn: either -1 or +1
  //   - exponent: between -1023 and 1024 (assuming normal)
  //   - mantissa: between 0 and 2^51 - 1 (safe by construction)
  // such that fac = sgn * 2^exponent * (1 + mantissa / 2^52)
  // so
  //     fac * 2^1023 * 2^52 = sgn * 2^(exponent+1023) * (2^52 + mantissa)
  // is a product of integers---
  const bigFac =
    BigInt(sgn) *
    2n ** (BigInt(exponent) + 1023n) *
    (2n ** 52n + BigInt(mantissa));
  // ---and thus the bigint arithmetic is correct to within 1ulp (could
  // be 0.5ulp, but preserving floor semantics per your code):
  return (grain * bigFac) / 2n ** (1023n + 52n);
}
```

You need a bit of care for subnormals, but honestly it would just be
fine to throw on them and not support multiplying by `fac`s of magnitude
smaller than `1 / Number.MAX_VALUE` ≈ 5.5e-309. (Lots of floating point
code doesn’t support subnormals.)

Note that in this implementation we only ever convert from floating
point to bigint—a lossless conversion, even if the floating point input
is itself approximate. We never convert from bigint to floating point,
which is inherently lossy. Thus, we have sane properties at least in the
first argument, like:

  - `mul(g, 1) === g`
  - `mul(g1, f) + mul(g2, f) === mul(g1 + g2, f)`
  - `mul(k * g, f) === k * mul(g, f)` (for `k` a `BigInt`)

…and moreover it is immediate that you have those properties because
`grain` is only used in `mul` as a factor in a bigint product.

src/grain/grain.js
```diff
@@ -108,3 +108,36 @@ export function format(
...
+}
+
+/**
+ * Aproximately create a grain balance from a float.
```
wchargin: sp.: *Aproximately* → *Approximately*

src/grain/grain.test.js
```diff
@@ -81,4 +77,36 @@ describe("src/grain/grain", () => {
...
+      // To compare with arbitrary precision results, see:
+      // https://observablehq.com/@decentralion/grain-arithmetic
+
+      // Within 1 attoGrain of "true" value
+      // $ExpectFlowError
+      expect(multiplyFloat(ONE, 1 / 1337)).toEqual(747943156320119n);
+
+      // Within 300 attoGrain of "true" value
+      // $ExpectFlowError
+      expect(multiplyFloat(ONE, Math.PI)).toEqual(3141592653589793280n);
```
wchargin: It’s a bit weird that if we were to *improve* the error bounds on
`multiplyFloat` (per my comment above) that these tests could fail.
Maybe

```javascript
const actual = multiplyFloat(ONE, 1 / 1337);
const expected = 3141592653589793000n;
const tolerance = 300000n;
expect(actual).toBeGreaterThanOrEqual(expected - tolerance);
expect(actual).toBeLessThanOrEqual(expected + tolerance);
```

such that they pass as written and we can tighten if desired?

src/grain/grain.js
wchargin: A high-level summary of the second half of this comment, starting from
“This deficit is not fundamental”, is: “Every (finite) floating point
number exactly represents a ratio of two integers. If `fac = p / q`,
then just take `g * BigInt(p) / BigInt(q)` to compute the exact result.”

wchargin: Maybe an even higher-level rule of thumb: 64-bit floats only have about
16 significant figures (decimal) of precision. You’re trying to create a
fixed-point type whose scaling factor is 10^18. Thus, you can *never*
take `Number(grain)` without losing data.

wchargin:
I’ve taken a stab at this myself:
<https://github.com/sourcecred/sourcecred/commits/wchargin-grain-multiplyfloat-exact>

src/grain/grain.test.js
Beanow: Second this, and would suggest resembling the number counterpart jest has.
`.toBeCloseTo(number, numDigits?)`
https://jestjs.io/docs/en/expect#tobeclosetonumber-numdigits

Maybe
`expectBigintToBeCloseTo(actual, expected, tolerance)`

src/grain/grain.test.js
```diff
@@ -81,4 +77,36 @@ describe("src/grain/grain", () => {
...
+      // $ExpectFlowError
+      expect(multiplyFloat(1n, 5)).toEqual(5n);
+    });
+    it("behaves reasonably for larger grain values", () => {
+      // $ExpectFlowError
+      expect(multiplyFloat(ONE, 2)).toEqual(2n * ONE);
+    });
+    it("has small error on large grain values", () => {
+      // To compare with arbitrary precision results, see:
+      // https://observablehq.com/@decentralion/grain-arithmetic
```
Beanow: I think this is an example that asks for a smoke test and a full test separation.
This being the inconclusive smoke test with a couple of fast-to-test examples.

The full test would fuzz this function with a substantial range of possible arguments and tests whether our specified margin of error holds true for all of these possibly unforeseen cases.

wchargin: I’m certainly open to something like that, with two caveats:

  - It’s not clear to me how to generate the expected values.
  - With the [exact implementation][1], we’re more worried about logic
	errors than precision errors.

On the first point: assuming that we don’t want to shell out to Python
or Java, I haven’t yet seen an “obviously correct” exact implementation
in JavaScript. Unfortunately, the results given by `decimal` in
@decentralion’s notebook are not actually correct. The `Decimal`
constructor, given a float, returns a decimal that is within an
f64-halfulp of the correct value but not within a decimal-halfulp of the
correct value. For instance,

```javascript
const decimal = require("decimal.js-light");
decimal(Math.PI).toPrecision(24, decimal.ROUND_HALF_DOWN)
```

is 3.14159265358979300000000, but the correct value is:

> 3.141592653589793115997963468544185161590576171875

The first three missing digits (“…115”) matter at attointeger scale. We
could resolve this concern if we could find a way to tell `decimal` to
give the correct answers.

On the second point: with the inexact `Number(grain)` implementation,
we’re worried about the worst-case error within an arbitrary interval of
floats—say, floats between `0.01` and `5.0`. This interval contains both
`1/1337` (pretty low error) and `Math.PI` (notably higher error), so a
sweep/smoke test over the interval would plausibly be quite valuable.

With the exact implementation, the edge cases to worry about aren’t
arbitrary float values within a given interval, but rather weird edge
cases of floats themselves: subnormals, zero mantissas, NaN payloads,
etc. The failure modes are more likely to have to do with the bit
manipulation than the algebra, because the algebra is uniform across all
floats. Thus, we’re worried about logic errors rather than precision
errors, and it’s not clear to me that fuzzing is an effective way to
address that.

[1]: https://github.com/sourcecred/sourcecred/commit/173be6b923d049895489a803bc353c47efdfff09#diff-14fdfc9807975908ca7d0b18bc2f3b76R174-R189

src/grain/grain.test.js
wchargin: I saw `toBeCloseTo`, but it’s documented as taking only `number`s, while
other assertions on the page take `number | bigint`, so I wouldn’t trust
`toBeCloseTo` to be lossless. That’s why I suggested what I did, which
I think works fine:
<https://github.com/sourcecred/sourcecred/commit/6fc1a8efe9550b266aa570347e12c2ecdd492e63#diff-dac25656aec5c30832a9dadf8c52ff54R104-R108>

src/grain/grain.test.js
Beanow: > On the first point: assuming that we don’t want to shell out to Python
> or Java, I haven’t yet seen an “obviously correct” exact implementation
> in JavaScript.

Should we not have a good option available in JS, rather than shell out I would go for a static corpus, generating expected values for as strings in a JSON file for instance, parsing those as bigint.

> With the exact implementation, the edge cases to worry about aren’t
> arbitrary float values within a given interval, but rather weird edge
> cases of floats themselves: subnormals, zero mantissas, NaN payloads,
> etc.

That makes sense, my suggestion did not consider the exact implementation, just what was there in the PR commit so far. If we're going with lossy and want to be sure about our margin of error, we'll need more test coverage than 2 cases :] If we're ditching the lossy approach testing float edge cases makes more sense.

src/grain/grain.js
```diff
@@ -108,3 +108,36 @@ export function format(
...
+ *
+ * This method tries to convert the floating point `amt` into a grain
+ * balance. For example, `grain(1)` approximately equals `ONE`.
+ *
+ * Do not assume this will be precise! For example, `grain(0.1337)` results in
+ * `133700000000000016n`. This method is intended for test code.
+ *
+ * This is a shorthand for `multiplyFloat(ONE, amt)`.
+ */
+export function fromFloat(f: number): Grain {
```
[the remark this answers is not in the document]
decentralion: OK, done.

src/grain/grain.js
decentralion: Got it. As discussed offline, let's use your fixes. I'm going to merge this one to unblock development, please send a followon PR which fixes the implementation.

src/grain/grain.js
decentralion: Done.

src/grain/grain.test.js
decentralion: I'll let you fix this when we fix the precision.

decentralion:
I'm going to merge this sans-approval so as to unblock making some rapid progress on grain ledger with @hammadj. @wchargin, as discussed offline, please feel free to replace this with your superior implementation; I just need something reasonable for now to unblock developing the ledger.

Beanow:
@decentralion ok with that. Let's add a tracking issue though :]

PR #1718, opened 2020-03-25
blame: ignore reformattings due to Prettier 2.0.1

wchargin:
Summary:
This commit introduces an `ignoreRevsFile` for use with Git blame. See
the header of the new file for details.

Test Plan:
Run `git blame -L183,183 config/test.js`. Before this change, note that
it blames to the reformatting commit. After this change, and after
configuring `git config blame.ignoreRevsFile .git-blame-ignore-revs`,
note that it blames to the last real change.

wchargin-branch: prettier-2.0.1-unblame

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -10,10 +10,11 @@
 # replacing `COMMIT_HASH` with the appropriate hash, or pass similar
 # formatting to `git log`, etc.
 #
-# To teach your Git client to read this file, run
+# To use this file, ensure that you have Git 2.23 or later installed,
+# then configure your repository to teach it about this file:
 #
 #     git config blame.ignoreRevsFile .git-blame-ignore-revs
 #
-# (with Git 2.23+). See `git help blame` `git help config` for details.
+# See `git help blame` or `git help config` for details.
 
 421aded196eb40c1dd3ec103bedc48d13111fa98  # prettier: reformat code for 2.0.1
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,19 @@
+# List of commits that `git blame` should ignore.
+#
+# Include ONLY commits that are known to be no-ops, like automated
+# reformattings.
+#
+# To properly format an entry, run
+#
+#     git show --no-patch --format='%H  # %s' COMMIT_HASH
+#
+# replacing `COMMIT_HASH` with the appropriate hash, or pass similar
+# formatting to `git log`, etc.
+#
+# To teach your Git client to read this file, run
+#
+#     git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# (with Git 2.23+). See `git help blame` `git help config` for details.
+
+421aded196eb40c1dd3ec103bedc48d13111fa98  # prettier: reformat code for 2.0.1


Beanow:
Today I learned, I don't really know git :smile:

wchargin:
Heh, yep, there are some hidden gems. :-)

(WANT_LGTM=any here; feel free but not obligated to drive-by approve)

Beanow:
So I did as mentioned in the test plan but not seeing the expected result.

```
$ git status
On branch wchargin-prettier-2.0.1-unblame
Your branch is up to date with 'origin/wchargin-prettier-2.0.1-unblame'.
...
nothing added to commit but untracked files present (use "git add" to track)

$ git config blame.ignoreRevsFile .git-blame-ignore-revs

$ git blame -L183,183 config/test.js
421aded1 (William Chargin 2020-03-21 22:26:53 -0700 183)   const tasks = (function () {
```

Curiously though, running git blame from sublime (which just uses the same git cli) does seem to work?
image

Both of which are not on the version mentioned in the file.
`git version 2.17.1`

Both confirm the setting is set:
```
$ git config blame.ignoreRevsFile
.git-blame-ignore-revs
```

wchargin:
> Both of which are not on the version mentioned in the file.
> `git version 2.17.1`

Well, that’s the problem, right? As noted, you need Git 2.23 or higher,
because that’s the version of Git in which this feature was added:
<https://github.com/git/git/blob/9fadedd637b312089337d73c3ed8447e9f0aa775/Documentation/RelNotes/2.23.0.txt#L83-L84>

When I build Git at `v2.17.1`, I can also repro that it doesn’t work;
when I build at `v2.23.0` or `v2.26.0`, it works, as expected.

Beanow:
Cool. I was just thrown off by the plugin somehow seeming to support it while reporting the same version. Probably something weird going on there.

Worth noting Ubuntu currently doesn't ship `2.23+` in any stable repos.
Having a look at https://launchpad.net/~git-core/+archive/ubuntu/ppa gives the expected result.

```
$ git --version
git version 2.26.0

$ git blame -L183,183 config/test.js
78dc571c (Dandelion Mané 2019-07-16 01:51:14 +0100 183)   const tasks = (function () {
```

.git-blame-ignore-revs
```diff
@@ -0,0 +1,19 @@
...
+#     git show --no-patch --format='%H  # %s' COMMIT_HASH
+#
+# replacing `COMMIT_HASH` with the appropriate hash, or pass similar
+# formatting to `git log`, etc.
+#
+# To teach your Git client to read this file, run
+#
+#     git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# (with Git 2.23+). See `git help blame` `git help config` for details.
```
Beanow: Nit: suggest rephrasing / adding it's on paragraph to make clear this means a version requirement and it's not at all supported in earlier versions.

wchargin: Done.

Beanow:
@wchargin agree, it's not hampering lower versions. Even if you have the required version this is an opt-in with the config step needed.

And since there's nothing critical or required about this, it's just a sidenote.

Noting I think it gives some indication of expected install base. There's a good chance people who would like to use this feature, should upgrade through an alternative channel to do that :]

PR #1719, opened 2020-03-26
Eslint, warn on prefer-const rule

Beanow:
Based on an idea from https://github.com/sourcecred/sourcecred/pull/1697#discussion_r397734818

"This rule is aimed at flagging variables that are declared using let
keyword, but never reassigned after the initial assignment."
https://eslint.org/docs/rules/prefer-const

This is a rule that helps you spot mistakes while writing code with a
let statement. When thinking that you will need to reassign it, eslint
will point out you're not actually doing so.

This could be because you've forgotten something, assigned the wrong
variable, or no longer need to reassign.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,7 @@ module.exports = {
     "plugin:flowtype/recommended",
   ],
   rules: {
+    "prefer-const": ["warn"],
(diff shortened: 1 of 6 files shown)


Beanow:
Judging by how there's only one instance of unnecessary let in the core. The implicit standard is already to prefer const. And since it can prevent bugs to test for that, going to merge this on one approval :]

PR #1720, opened 2020-03-26
Add a README.md for the GitHub plugin

decentralion:
This mostly documents at a conceputal level that would be useful for
users; it's not trying to document it for contributors to the GitHub
plugin, although that could be useful as a followup.

Sending this out for initial review, although arguably we should merge
it to sourcecred/docs rather than directly alongside the code. I welcome
feedback on content, structure, and location.

Test plan: No tests, just review.

PR #1735, opened 2020-04-03
Identity: add user documentation to README

### LINKED ISSUE #1725: User docs: Identities
While identities don't really have a CLI. Knowing how to set it up with the project file and the effect it will have should be explained.

Beanow:
By adding information about configuration and what identity contractions
do, it provides more reference for instance maintainers and community
members to understand the plugin.

Closes #1725

### DIFF
--- a/src/plugins/identity/README.md
+++ b/src/plugins/identity/README.md
@@ -10,10 +10,136 @@ GitHub and Discourse (potentially with a different username on each service).
 We would like to combine these two identities together, so that we can
 represent that user's combined cred properly. The Identity plugin enables this.
 
-Specifically, the instance maintainer can provide a (locally unique) username
-for the user, along with a list of aliases the user is known by, e.g.
-`github/username` and `discourse/other_username`. The aliases are simple string
-representations, that are intended to be easy to maintain by hand in a
-configuration file. Then, the identity plugin will provide a list of
-`NodeContraction`s that can be used by `Graph.contractNodes` to combine the
-user identities as described.
+## Status and Caveats
+
+The Identity plugin currently only applies to new (Weighted)Graphs and Cred
+scores being generated. When including historical Graphs/Cred scores, for
+example to distribute Grain, you may need to contract the identities again.
+This can be tricky to do correctly, as they may have already received Grain
+on their GitHub/Discourse addresses before contraction.
+
+Code-reviewed support for Grain built into SourceCred should handle this in
+the future. But for now, requires the instance maintainer to handle Grain
+account contraction.
+
+## Configuration
+
+Identities can be configured using the `project.json` file.
+By adding entries to the the `identities` array.
+
+As an example of what this would look like:
+
+```js
+[
+  {
+    "type": "sourcecred/project",
+    "version": "0.4.0",
+  },
+  {
+    // The identities we should collapse.
+    "identities": [
+      {
+        "username": "User-A",
+        "aliases": ["github/user-a", "discourse/user-a"],
+      },
+      {
+        "username": "User-B",
+        "aliases": ["github/user-b", "discourse/user-b"],
+      },
+    ],
+
+    "id": "@example-project",
+    "discourseServer": {
+      // ...
+    },
+    "repoIds": [
+      // ...
+    ],
+  },
+]
+```
+
+### Username
+
+The `username` field allows you to choose a new username for the collapsed
+identity.
+
+It's case-sensitive, must be _unique in this project_ and can contain `A-Z`,
+`a-z`, `0-9`, `-` or `_` as valid characters.
+
+For display purposes the username is prefixed with an `@`.
+If a username is prefixed with `@` in the project file, it will be ignored.
+Meaning `"username": "User-A"` and `"username": "@User-A"` are equivalent.
+
+### Aliases
+
+Aliases are strings with a format of: `{plugin}/{user ID for this plugin}`
+Currently supported plugins are:
+
+| Plugin    | Example alias            | Notes                                                              |
+|:----------|:-------------------------|:-------------------------------------------------------------------|
+| GitHub    | `github/example-user`    | A user account on GitHub's platform.                               |
+| Discourse | `discourse/example-user` | A user account on the Discourse server configured in this project. |
+
+## How a contraction works
+
+Firstly, contractions are done _after_ the plugins have created their respective
+parts of the graph. But _before_ any Cred scores are calculated.
+
+The GitHub and Discourse plugins create edges for their respective user nodes
+before contraction:
+
+```
+GitHub user @abc
+├── authored Pull Request #123
+└── authored Issue #456
+
+Discourse user /u/abc
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+A new node is created, representing the Identity, with the username
+provided in the project file. While removing the alias nodes.
+
+```
+Identity @abc
+
+[REMOVED]
+├── authored Pull Request #123
+└── authored Issue #456
+
+[REMOVED]
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+And we connect the edges to the new identity node.
+
+```
+Identity @abc
+├── authored Pull Request #123
+├── authored Issue #456
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+Now we're ready to calculate Cred scores.
+
+### Reference detection
+
+Plugins often support reference detection. This analyses a URL found in a piece
+of content, to create a reference edge in the graph. The "detection" here means
+to find out, whether the URL we found represents a node in the graph or not.
+
+There isn't currently a URL scheme which represents an identity.
+
+But because reference detection is done _before_ contraction, this means we can
+use the URL for any of the aliases, and they'll point to the same identity.
+For example:
+
+- `https://github.com/abc`
+- `https://example.discourse/u/abc`
+
+These are effectively equivalent in this example, both creating edges to
+`Identity @abc`.
--- a/src/plugins/identity/README.md
+++ b/src/plugins/identity/README.md
@@ -1,22 +1,28 @@
 # SourceCred Identity Plugin
 
-This folder contains the Identity plugin. Unlike most other plugins, the
-Identity plugin does not add any new contributions to the graph. Instead, it
-allows collapsing different user accounts together into a shared 'identity'
-node.
+Unlike most other plugins, the Identity plugin does not add any new
+contributions to the graph. Instead, it's goal is to combine different user
+accounts into a single identity. For example, if a contributor has an account
+on both GitHub and Discourse (potentially with different usernames on each),
+the identity plugin can combine these two accounts into a single identity,
+allowing SourceCred to properly compute the Cred for this identity.
 
-To see why this is valuable, imagine that a contributor has an account on both
-GitHub and Discourse (potentially with a different username on each service).
-We would like to combine these two identities together, so that we can
-represent that user's combined cred properly. The Identity plugin enables this.
+The technical term for how we do such combining is a _contraction_. Which we'll
+use for the rest of this document.
 
 ## Status and Caveats
 
-The Identity plugin currently only applies to new (Weighted)Graphs and Cred
-scores being generated. When including historical Graphs/Cred scores, for
-example to distribute Grain, you may need to contract the identities again.
-This can be tricky to do correctly, as they may have already received Grain
-on their GitHub/Discourse addresses before contraction.
+The Identity plugin currently only applies to new WeightedGraphs and Cred
+scores being generated. However, if you're using this data to distribute
+Grain or other tokens, you may need to contract the identities again.
+
+For example, someone might have received Grain in the past for their GitHub and
+Discourse user respectively, creating 2 separate accounts in the Grain ledger.
+After these GitHub and Discourse nodes were contracted to one identity node,
+the ledger would now have 3 separate accounts. The GitHub and Discourse ones
+holding past distributions, but no longer receiving more. And a new identity
+one which will receive distributions going forward. Normally you'd want to
+have just 1 account instead which holds all distributions including past ones.
 
 Code-reviewed support for Grain built into SourceCred should handle this in
 the future. But for now, requires the instance maintainer to handle Grain
@@ -27,7 +33,8 @@ account contraction.
 Identities can be configured using the `project.json` file.
 By adding entries to the the `identities` array.
 
-As an example of what this would look like:
+Below is an example `project.json` file which defines two identities, "User-A"
+and "User-B", contrating their respective GitHub and Discourse accounts.
 
 ```js
 [
@@ -36,7 +43,7 @@ As an example of what this would look like:
     "version": "0.4.0",
   },
   {
-    // The identities we should collapse.
+    // The identities we should contract.
     "identities": [
       {
         "username": "User-A",
@@ -61,7 +68,7 @@ As an example of what this would look like:
 
 ### Username
 
-The `username` field allows you to choose a new username for the collapsed
+The `username` field allows you to choose a new username for the contracted
 identity.
 
 It's case-sensitive, must be _unique in this project_ and can contain `A-Z`,
@@ -83,11 +90,18 @@ Currently supported plugins are:
 
 ## How a contraction works
 
-Firstly, contractions are done _after_ the plugins have created their respective
-parts of the graph. But _before_ any Cred scores are calculated.
+Here is a simplified illustration of how contracting works under the hood.
 
-The GitHub and Discourse plugins create edges for their respective user nodes
-before contraction:
+Note that contractions are done _after_ the plugins have created their
+respective parts of the graph. But _before_ any Cred scores are calculated.
+
+Let’s assume we have a user `abc`, who has both GitHub and Discourse accounts
+which we want to contract.
+
+The GitHub and Discourse plugins have mapped out the contributions, users and
+edges as they exist on those services. Even though we merge this into a single
+graph, the GitHub and Discourse contributions of the _same person_ are still
+disconnected.
 
 ```
 GitHub user @abc
@@ -99,8 +113,8 @@ Discourse user /u/abc
 └── authored Post /t/123/2
 ```
 
-A new node is created, representing the Identity, with the username
-provided in the project file. While removing the alias nodes.
+The identity plugin creates a new node, representing the Identity, with the
+username provided in the project file. While removing the alias nodes.
 
 ```
 Identity @abc
@@ -114,7 +128,8 @@ Identity @abc
 └── authored Post /t/123/2
 ```
 
-And we connect the edges to the new identity node.
+Then the edges that were connected to the alias nodes are connected to the new
+identity node instead.
 
 ```
 Identity @abc
@@ -124,22 +139,6 @@ Identity @abc
 └── authored Post /t/123/2
 ```
 
-Now we're ready to calculate Cred scores.
-
-### Reference detection
-
-Plugins often support reference detection. This analyses a URL found in a piece
-of content, to create a reference edge in the graph. The "detection" here means
-to find out, whether the URL we found represents a node in the graph or not.
-
-There isn't currently a URL scheme which represents an identity.
-
-But because reference detection is done _before_ contraction, this means we can
-use the URL for any of the aliases, and they'll point to the same identity.
-For example:
-
-- `https://github.com/abc`
-- `https://example.discourse/u/abc`
-
-These are effectively equivalent in this example, both creating edges to
-`Identity @abc`.
+The disconnect between the GitHub and Discourse service is now gone. If we now
+run the SourceCred algorithm, user `@abc`'s Cred will flow to a single node and
+produce a much more accurate Cred score than before.
--- a/src/plugins/identity/README.md
+++ b/src/plugins/identity/README.md
@@ -34,7 +34,7 @@ Identities can be configured using the `project.json` file.
 By adding entries to the the `identities` array.
 
 Below is an example `project.json` file which defines two identities, "User-A"
-and "User-B", contrating their respective GitHub and Discourse accounts.
+and "User-B", contracting their respective GitHub and Discourse accounts.
 
 ```js
 [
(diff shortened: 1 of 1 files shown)


Beanow:
cc @s-ben as well, might want to review.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
-`github/username` and `discourse/other_username`. The aliases are simple string
-representations, that are intended to be easy to maintain by hand in a
-configuration file. Then, the identity plugin will provide a list of
-`NodeContraction`s that can be used by `Graph.contractNodes` to combine the
-user identities as described.
+## Status and Caveats
+
+The Identity plugin currently only applies to new (Weighted)Graphs and Cred
+scores being generated. When including historical Graphs/Cred scores, for
+example to distribute Grain, you may need to contract the identities again.
```
vsoch, line 17: "for example to distribute Grain" I think should be in () instead?

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+Identities can be configured using the `project.json` file.
+By adding entries to the the `identities` array.
+
+As an example of what this would look like:
+
+```js
+[
+  {
+    "type": "sourcecred/project",
+    "version": "0.4.0",
```
vsoch, line 36: Should there be a note that version is important? It might be the case that someone is editing an older file and just passes it over.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+The `username` field allows you to choose a new username for the collapsed
+identity.
+
+It must be _unique in this project_ and can contain `A-Z`, `a-z`, `0-9`, `-` or
+`_` as valid characters.
+
+For display purposes the username is prefixed with an `@`.
+If a username is prefixed with `@` in the project file, it will be ignored.
+Meaning `"username": "User-A"` and `"username": "@User-A"` are equivalent.
+
```
vsoch, line 73: What about if someone uses different capitalizations (which might be intentional but more likely a typo) e.g., "User-A" and "user-a"

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+### Aliases
+
+Aliases have a format of:
+`{plugin}/{user ID for this plugin}`
+
+Currently supported plugins are:
+
+| Plugin | Example alias | Notes |
+|---|---|---|
```
vsoch, line 82: Nit: can we make the table rows equally spaced so it looks nice in the markdown as well? Discourse is the largest in terms of cell size so we'd match to that.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+| GitHub | `github/example-user` | A user account on GitHub's platform. |
+| Discourse | `discourse/example-user` | A user account on the Discourse server configured in this project. |
+
+
+## How a contraction works
+
+Firstly, contractions are done _after_ the plugins have created their repective
```
vsoch, line 89: repective-> respective

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+parts of the graph. But _before_ any Cred scores are calculated.
+
+The GitHub and Discourse plugin create edges for their respective user nodes
```
vsoch, line 92: plugin -> plugins

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+└── authored Post /t/123/2
+```
+
+A new node is created, represeting the Identity, with the username
+provided in the project file. While removing the alias nodes.
+
+```
+Identity @abc
+
+???
```
vsoch, line 111: I'm confused about what the ??? means. I understand the idea that the identity @abc is now combining these two, but would it be incorrect to represent all the contributions as one list under that alias? Is the goal to show that the sub-usernames are removed, but the graph structure is still maintained?

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+And we connect the edges to the new identity node.
+
+```
+Identity @abc
```
vsoch, line 123: Ahh here we go! This is what I expected to see.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+├── authored Pull Request #123
+├── authored Issue #456
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+Now we're ready to calculate Cred scores.
+
+### Reference detection
```
vsoch, line 132: This is the first time we are talking about "the URL" - maybe we can have a sentence like "A sourcecred prototype creates a URL to represent each identity. For example, for identity `abc` we would have:
```
<show url>
```
However the urls are maintained for the previous alias for reference. For example, we can refer to any of these URLS

```
<show urls>
```
and they will take us to the page for `<previous url for abc>`

src/plugins/identity/README.md
Beanow: Actually version is not important for Identity. Current master is compatible with v0.3.0 through to v0.5.0 of the project file. For which the way to configure identities has not changed.

Since the doc is committed alongside the code, this covers the full supported range :]

vsoch: Great! Just wanted to check.

src/plugins/identity/README.md
Beanow: Usernames are case-sensitive. I'll mention that explicitly.

src/plugins/identity/README.md
Beanow: Meant to show this like an in-between step where the nodes are updated, but the edges still point to the no-longer-existing nodes. The ??? is probably not very clear. I've replaced it with `[REMOVED]` does that make it a bit more clear?

src/plugins/identity/README.md
Beanow: > A sourcecred prototype creates a URL to represent each identity.

It does not actually do that though :]
There's currently no URL to refer to the identity.

src/plugins/identity/README.md
Beanow: That seems more like style than a grammar rule. The example seems important enough to take it from abstract "historical Graphs/Cred scores" to a concrete use-case, that I picked commas over parens :]

src/plugins/identity/README.md
vsoch: Yes, that works better. The ??? makes me think "I don't know what is going on here?" but "REMOVED" is explicit.

src/plugins/identity/README.md
Beanow: Mentioned case-sensitivity.

src/plugins/identity/README.md
Beanow: Thanks, forgot to run a spell check. Fixed.

src/plugins/identity/README.md
Beanow: Fixed.

src/plugins/identity/README.md
vsoch: okay, whatever it does, it would be helpful to explain the space of these URLs because this is the first time it's being mentioned. My lack of understanding is a reflection of that missing bit.

src/plugins/identity/README.md
Beanow: Resolved with the [REMOVED] variant I think :]

src/plugins/identity/README.md
vsoch: Yes it's just against my preference, can't argue that it's wrong.

src/plugins/identity/README.md
Beanow: What do you mean by "the space of these URLs"? Do you mean where the URLs occur?

Since this is about reference detection, any place the other plugins apply reference detection. Like Discourse posts, GH comments or commit messages.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,132 @@ GitHub and Discourse (potentially with a different username on each service).
...
+  },
+]
+```
+
+### Username
+
+The `username` field allows you to choose a new username for the collapsed
+identity.
+
+It must be _unique in this project_ and can contain `A-Z`, `a-z`, `0-9`, `-` or
```
vsoch, line 67: Nit: if we are still talking about "the username field" I think it should state "as valid characters and is case sensitive?" If the reference is to the characters you could say "as valid characters that are case sensitive."

src/plugins/identity/README.md
vsoch: This section is still not clear to me, I am not familiar with the name space of URLs.

src/plugins/identity/README.md
Beanow: Moved this to the front, to avoid confusion whether we're referring to characters or usernames. But it should have been `is` instead of `are` yeah :]

src/plugins/identity/README.md
vsoch: The section "Reference Detection" opens and I need a little more description for what reference detection is, and why those URLs are relevant to it. This is written from the point of someone that understands these concepts, it doesn't take account for the reader that does not.

decentralion: AFAIK the identity plugin doesn't interact with reference detection at all. Basically the rule is that the identity plugin will re-write edges to identities, which will cover reference edges that came from individual plugins too. Given this, is there a need to mention reference detection explicitly here at all?

vsoch: @decentralion that might be the best solution! If we keep some mention here, the description that you just wrote is very clear and could be used.

Beanow: Yes, technically it only does the contraction. So it doesn't interact with reference detection directly.

Reason I'm mentioning it is because I'm anticipating a question. For example when creating an initiative json file where URLs are currently the only accepted input.

Q: Now `@abc` has an identity, and identities don't have URLs, how should I refer to that identity as a URL?
A: Use the URL of any of the underlying aliases, as the contraction happens after reference detection, it will refer to the identity.

I could add this example as a caveat for initiatives' docs instead. But sided with caution to ensure there's something on the subject rather than nothing :]

src/plugins/identity/README.md
Beanow: Ran a [table formatter](https://packagecontrol.io/packages/Markdown%20Table%20Formatter).

src/plugins/identity/README.md
vsoch: Maybe that should go with documentation for creating the json file (and then be linked from here?)

s-ben: Agree with @vsoch. Was confused as to why reference detection was discussed here. 

It's good to have this topic documented somewhere, but I think there's a better place for it. It feels like more developer-centric doco that should live in more low-level technical documentation, or even just in code comments for now. Perhaps the Initiatives plugin README if that helps explain why you can use aliases to point to contributors.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,136 @@ GitHub and Discourse (potentially with a different username on each service).
...
+
+Aliases are strings with a format of: `{plugin}/{user ID for this plugin}`
+Currently supported plugins are:
+
+| Plugin    | Example alias            | Notes                                                              |
+|:----------|:-------------------------|:-------------------------------------------------------------------|
+| GitHub    | `github/example-user`    | A user account on GitHub's platform.                               |
+| Discourse | `discourse/example-user` | A user account on the Discourse server configured in this project. |
+
+## How a contraction works
```
s-ben, line 84: Took at stab at rewording the below section for flow.

Suggested: 

## Combining Entities

Below is a simplified illustration of how combining aliases works under the hood. 

Note that combining of aliases is performed after the plugins have created their respective part of the graph, but before any cred scores are calculated. 

Let’s assume we have a user, `abc`, who has both GitHub and Discourse accounts which we want to combine.

Running the GitHub plugin creates an identity node with two edges: a Pull Request (PR) and an Issue.

```
GitHub user @abc
├── authored Pull Request #123
└── authored Issue #456
```

Running the Discourse plugin creates an identity node with two edges, both posts `abc` has authored. 

```
Discourse user /u/abc
├── authored Post /t/123/1
└── authored Post /t/123/2
```

Note that the GitHub and Discourse plugin have created separate graphs, which have not yet been merged. 

First, the identity plugin creates a new identity node, with the username provided in the `project.json` file, while removing the alias nodes.

```
Identity @abc

[REMOVED]
├── authored Pull Request #123
└── authored Issue #456

[REMOVED]
├── authored Post /t/123/1
└── authored Post /t/123/2
```

Then the edges are connected to the new identity node. 

```
Identity @abc
This conversation was marked as resolved by vsoch
├── authored Pull Request #123
├── authored Issue #456
├── authored Post /t/123/1
└── authored Post /t/123/2
```

The SourceCred core algorithm can then next calculate cred scores.

src/plugins/identity/README.md
Beanow: Got it, I'll remove this section and leave it for the Initiatives plugin to explain.

Beanow:
@s-ben The term "contraction" I'm guessing draws from the formal graph theory terms. https://en.wikipedia.org/wiki/Edge_contraction though it doesn't seem to exactly do what's explained here, it's definitely related. @decentralion may have an answer to that.

In code we're using the term contraction everywhere though, so the time being I'm inclined to have the documentation match the code's terms to avoid creating a new source of confusion.

Realizing this might make the docs less accessible to the end-user audience. However I would rather this doc is an accurate source of information for (aspiring) developers and cred-technicians, so they can grasp the implications of a change like https://github.com/sourcecred/cred/pull/25/commits/38d84500393fec209249228d53158464d2998b58 and create / review these, than to prioritize the end-user audience.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,136 @@ GitHub and Discourse (potentially with a different username on each service).
 We would like to combine these two identities together, so that we can
 represent that user's combined cred properly. The Identity plugin enables this.
```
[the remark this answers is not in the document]
Beanow: So I'm not completely sure on what aspects you'd like to see consistency with #1720.
But here's how I'm looking at the suggestions.

Of course I'll check for any technical inaccuracies I can find.
If it seems more concise while having a good (subjective) flow, I'll take it as an improvement.

I'll prefer to keep little side-notes that could lead to "oh, I didn't know that" insights that aren't obvious. For example I would like to keep:
> Unlike most other plugins, the Identity plugin does not add any new contributions to the graph.

As that could lead to: "Oh, I didn't know the expectation was for most plugins to add contributions to the graph." And could help understand why there have been discussions to move Identity to the core, rather than a plugin.

By contrast, this could be removed:
> This folder contains the Identity plugin.

In the `plugins/identity` folder, containing a README titled `SourceCred Identity Plugin`, I'm sure the intended audience has enough info to connect the dots.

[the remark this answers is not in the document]
Beanow: Keeping the "unlike other plugins", capitalizing "Cred" and rephrased this part:
```diff
- allowing SourceCred to properly compute the user’s cred.
+ allowing SourceCred to properly compute the Cred for this identity.
```

src/plugins/identity/README.md
```diff
@@ -10,10 +10,136 @@ GitHub and Discourse (potentially with a different username on each service).
...
-Specifically, the instance maintainer can provide a (locally unique) username
-for the user, along with a list of aliases the user is known by, e.g.
-`github/username` and `discourse/other_username`. The aliases are simple string
-representations, that are intended to be easy to maintain by hand in a
-configuration file. Then, the identity plugin will provide a list of
-`NodeContraction`s that can be used by `Graph.contractNodes` to combine the
-user identities as described.
+## Status and Caveats
+
+The Identity plugin currently only applies to new (Weighted)Graphs and Cred
```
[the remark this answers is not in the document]
Beanow: > Therefore, if you re-run SourceCred after merging new identities, the scores will shift. This can be tricky,

This isn't really the issue I'm trying to highlight. Cred scores can shift around for a number of reasons (and this is normal because retroactive Cred is a feature). Contracting identities being one, weight changes being another common case.

The problem that identity contraction has which a weight change doesn't, is that Grain might have been distributed to your GitHub and Discourse node in previous weeks, while post-contraction it would be distributed to your identity node going forward. Not dealing with this you're going from having 2 accounts in the Grain ledger, to 3! While you'd want this to be 1, just the identity one.

I'll take a stab at clarifying this section.

src/plugins/identity/README.md
```diff
@@ -10,10 +10,136 @@ GitHub and Discourse (potentially with a different username on each service).
...
+This can be tricky to do correctly, as they may have already received Grain
+on their GitHub/Discourse addresses before contraction.
+
+Code-reviewed support for Grain built into SourceCred should handle this in
+the future. But for now, requires the instance maintainer to handle Grain
+account contraction.
+
+## Configuration
+
+Identities can be configured using the `project.json` file.
```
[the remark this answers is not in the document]
Beanow: > located in the instance’s root directory. For instance, here is SourceCred’s project.json.

The file's location isn't fixed. You need to specify the path using the CLI's `--project <file>` argument. `sourcecred load --help` would explain this in more detail.
https://github.com/sourcecred/sourcecred/blob/f2403ce33febd02a0b11f93e9f6fed560e1a21f1/src/cli/load.js#L22-L70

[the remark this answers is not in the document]
Beanow: Adopting just the "Below is an example ___ which" style for context on how to read the json file, in my own words though :]

src/plugins/identity/README.md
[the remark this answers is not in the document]
Beanow: Have replaced instances of `collapsing` with `contracting` for consistency. Realizing that I'd like to keep `combining` in the introduction, I've added an introduction to the term there:

> The technical term for how we do such combining is a _contraction_. Which we'll
use for the rest of this document.

Beanow: > Running the GitHub plugin creates an identity node with ...
> Running the Discourse plugin creates an identity node with ...

They do not create identity nodes. They create GitHub user nodes and Discourse user nodes resp. The identity plugin creates identity nodes.

However all 3 of these nodes are marked as being "user node" types. Which we treat differently from other nodes in a few cases. For example, we would distribute Grain to user node types, rather than every node in the graph. `scores.json` files also filter to only include user node types.

---

> Note that the GitHub and Discourse plugin have created separate graphs, which have not yet been merged.

This seems a bit misleading to me in both terminology and visualization. Technically at this point we've already merged data from the different plugins into a single graph (why I visualized it as a single code block). However these sections of the graph are disconnected from each other, because they don't have any edges between them. After the contraction, they would be connected because they share the same author.

This disconnect while being in the same graph is relevant for seeing why Cred _flow_ also changes when contracting identities. For example, in the disconnected situation, Cred you've earned through contributions on GitHub, would not impact how much Cred can flow through your Discourse likes. While in the connected situation this "path" in the graph now exists and Cred can flow between services.

I'll elaborate more on this in the example to point this out as I explain the contraction itself.

Beanow:
I've updated based on feedback. Since it's a fairly chunky difference I've added this as a new commit, showing the differences for review. https://github.com/sourcecred/sourcecred/pull/1735/commits/721515f4e825eabaeb2809c0f51bd04e9e2c3985

Will squash them before merging :]

src/plugins/identity/README.md
```diff
@@ -1,19 +1,144 @@
...
+For display purposes the username is prefixed with an `@`.
+If a username is prefixed with `@` in the project file, it will be ignored.
+Meaning `"username": "User-A"` and `"username": "@User-A"` are equivalent.
+
+### Aliases
+
+Aliases are strings with a format of: `{plugin}/{user ID for this plugin}`
+Currently supported plugins are:
+
+| Plugin    | Example alias            | Notes                                                              |
```
vsoch, line 86: Beautiful spacing of this table @Beanow :)

src/plugins/identity/README.md
```diff
@@ -1,19 +1,144 @@
...
+
+[REMOVED]
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+Then the edges that were connected to the alias nodes are connected to the new
+identity node instead.
+
+```
```
vsoch, line 134: Good job refactoring this explanation, it's much more clear now.

src/plugins/identity/README.md
Beanow: Yeah that was all the work of the plugin! Select and press fix.
In my own notes I typically don't bother and just render as HTML.
But it's a good suggestion, this is easier to review and maintain this way.

(Explicit left text alignment colons? :eyes: Well if we're going to have 67 dashes anyway, why not?)

Beanow:
I think this is about ready for merge, let me know what you think @s-ben :]

src/plugins/identity/README.md
```diff
@@ -1,19 +1,144 @@
...
+├── authored Pull Request #123
+└── authored Issue #456
+
+Discourse user /u/abc
+├── authored Post /t/123/1
+└── authored Post /t/123/2
+```
+
+The identity plugin creates a new node, representing the Identity, with the
+username provided in the project file. While removing the alias nodes.
```
s-ben, line 117: The grammar nazi in me cannot let `. While` go. I think it works with `, while`, but that leaves a lot of commas... I think we can make this sentence less choppy by removing a couple of the other commas, i.e.,

> The identity plugin creates a new node representing the Identity with the
> username provided in the project file, while removing the alias nodes.

Beanow: I see what you mean. Mostly this part about the project file is throwing a wrench into this. What about:

> Using the information from the project file, the plugin creates a new Identity node and removes the alias nodes.

s-ben: Works for me

Beanow: Added this and now merged :D

PR #1736, opened 2020-04-04
Discord: create snapshots from test instance

BrianLitwin:
This adds a bash script that fetches data from our test
Discord instance. We will be able to test against this
data and easily update the data if Discord's api changes.

A SOURCECRED_DISCORD_BOT_TOKEN environment variable is 
required to run the script.

Test plan:

After running bash script, inspected snapshot files
and verified that the data appears reasonable.

Verified that the check for `jq` and `SOURCECRED_DISCORD_TOKEN` both
fail and exit if `jq` isn't installed or the Discord bot token
hasn't been set.


src/plugins/discord/update_discord_api_snapshots.sh
```diff
@@ -0,0 +1,39 @@
...
+rm -r "${snapshots_dir}"
+mkdir "${snapshots_dir}"
+fetch "guilds" "/guilds/678348980639498428"
+fetch "channels" "/guilds/678348980639498428/channels"
+fetch "members" "/guilds/678348980639498428/members?after=0&limit=5"
+fetch "messages" "/channels/678348980849213472/messages?after=0&limit=5"
+fetch "messages_pagination_0" "/channels/678394406507905129/messages?after=0&limit=5"
+fetch "messages_pagination_1" "/channels/678394406507905129/messages?after=678394436757094410&limit=5"
+fetch "messages_pagination_2" "/channels/678394406507905129/messages?after=678394455849566208&limit=5"
+fetch "reaction" "/channels/678348980849213472/messages/678697350385893404/reactions/sourcecred:678399364418502669?after=0&limit=5"
```
Beanow, line 39: I notice you're manually giving the snapshots names. In the Discourse counterpart of this we encode the URL into the filename, so making the same request from the MockFetcher would find this same file.

Since this is a different approach, I wonder how you'd want something like MockFetcher to resolve these files?

BrianLitwin: Updated

PR #1749, opened 2020-04-14
Initiatives: use common timestamp types and utils

Beanow:
Assumes #1748 is merged

In PR #1746 we've added `src/util/timestamp.js`. This commit replaces the
plugin's internal opaque string and number types with well tested types
and conversion functions.

Test plan:
- Verify all types (`Initiative`, `InitiativeFile`) use the new timestamp types.
- `yarn flow` should enforce using the util functions to parse and convert.
- `yarn unit` still passes (which includes identical snapshots of timestamps).

### DIFF
--- a/src/plugins/initiatives/initiativeFile.test.js
+++ b/src/plugins/initiatives/initiativeFile.test.js
@@ -1,6 +1,7 @@
 // @flow
 
 import {NodeAddress} from "../../core/graph";
+import * as Timestamp from "../../util/timestamp";
 import {createId, addressFromId} from "./initiative";
 import {type InitiativesDirectory} from "./initiativesDirectory";
 import {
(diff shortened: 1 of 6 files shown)


src/plugins/initiatives/initiativeFile.test.js
```diff
@@ -13,7 +14,7 @@ import {
 
 const exampleInitiativeFile = (): InitiativeFile => ({
   title: "Sample initiative",
-  timestampIso: ("2020-01-08T22:01:57.766Z": any),
+  timestampIso: Timestamp.toISO(Date.parse("2020-01-08T22:01:57.766Z")),
```
Beanow, line 17: Note: as I found this, I considered adding another validationg function like `(string) => TimestampISO`. But for tests this is actually a validated pattern. And not having this function discourages using ISO's as the internal representation.

`Date.parse(string)` would return `NaN` if the timestamp is invalid. And `Timestamp.toISO` would throw because `Number.isInteger(NaN) === false`.

PR #1750, opened 2020-04-16
Add "toFloatRatio" helper method

decentralion:
Add a numerically-naive method for calculating the floating point ratio
between grain values, as it is needed in #1743.

Following discussion in [this review], we hope that @wchargin will
re-write this method later to have better precision.

Test plan: Attached unit tests should pass.

[this review]: https://github.com/sourcecred/sourcecred/pull/1715#discussion_r396909459

Paired with @hammadj

### DIFF
--- a/src/grain/grain.test.js
+++ b/src/grain/grain.test.js
@@ -7,6 +7,7 @@ import {
   ZERO,
   fromApproximateFloat,
   multiplyFloat,
+  toFloatRatio,
 } from "./grain";
 
 describe("src/grain/grain", () => {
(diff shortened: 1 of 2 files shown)


src/grain/grain.test.js
```diff
@@ -117,4 +118,32 @@ describe("src/grain/grain", () => {
...
+    });
+    it("calculates repeating decimal ratios", () => {
+      // $ExpectFlowError
+      expect(toFloatRatio(ONE * 5n, ONE * 3n)).toEqual(5 / 3);
+    });
+    it("approximates correctly when Grain values are not exactly equal", () => {
+      // $ExpectFlowError
+      const almostOne = ONE - 1n;
+      expect(toFloatRatio(ONE, almostOne)).toEqual(1);
+    });
```
Beanow, line 142: This depends on what you consider "correct". One interpretation of "correct" means to floor in case of precision errors. Which #1743 currently depends on to not exceed it's budget.

For instance: `toFloatRatio(almostOne, ONE)` should result in `0.999999999999999999`.
But since Number can't represent that answer, it defaults to `1`. In JS `1 === 0.999999999999999999`.
The closest that doesn't default to `1` is `0.9999999999999999` (16 rather than 18 decimals).

Handling this correctly and consistently is tricky, so I'm ok with making a concession on the math for now.
If we do, we should take additional measures in #1743 to ensure we're not violating the rule `sum(receipts) <= budget`.

decentralion: Good point. I added a manual warning in #1743 so we can be aware if this happens. (If it does happen, it's likely to be at the level of atto-grains, which is not a huge issue.)

wchargin: Any reasonable definition of “correct” here must surely include “within
0.5ulp of exact”, in which case specs only differ in rounding behavior.
The rounding behavior matters, but I wouldn’t say that it’s tricky. You
just pick round-half-down and move on.

If

    x = 0_999_999_999_999_999_999
    y = 1_000_000_000_000_000_000

then `toFloatRatio(x, y)` must be `1.0`, because that is the only float
that is within 0.5ulp of the exact value.

PR #1751, opened 2020-04-16
Add `legacyConverter` module for the grain ledger

decentralion:
As part of the CredSperiment, we currently track grain distributions and
transfers in a very ad-hoc format in [sourcecred/cred][1]. This commit
adds a formal specification of that ad-hoc type, and tested logic for
converting from that type to our new event format. Once we complete this
conversion, we can throw this code away.

Test plan: `yarn test`, since I added nice unit tests.

### DIFF
--- a/src/grain/legacyConverter.js
+++ b/src/grain/legacyConverter.js
@@ -1,0 +1,145 @@
+// @flow
+
+import {
+  type DistributionV1,
+  type ImmediateV1,
+  type LifetimeV1,
+} from "./distribution";
+import {type TransferV1} from "./ledger";
+import {type Alias, resolveAlias} from "../plugins/identity/alias";
+import type {LedgerEvent} from "./ledger";
+import {type Grain, ONE, ZERO} from "./grain";
+
+type LegacyInterval = {|+startTimeMs: number, +endTimeMs: number|};
+type LegacyPayment = {|+alias: string, +fast: number, +slow: number|};
+type LegacyDistribution = {|
+  +interval: LegacyInterval,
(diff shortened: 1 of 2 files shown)


src/grain/legacyConverter.js
```diff
@@ -0,0 +1,145 @@
...
+    receipts: immediateReceipts,
+  };
+  const lifetimeDistribution: DistributionV1 = {
+    type: "DISTRIBUTION",
+    timestampMs,
+    version: 1,
+    strategy: lifetimeStrategy,
+    receipts: lifetimeReceipts,
+  };
+  return [immediateDistribution, lifetimeDistribution];
```
Beanow, line 125: I haven't reviewed this yet, since we're depending on the event models, and they are still being debated.
Given that we need to support event models forever, I think they need to be finalized first before we can merge this :]

src/grain/legacyConverter.js
```diff
@@ -0,0 +1,145 @@
...
+    version: 1,
+    budget: lifetimeBudget,
+  };
+  const immediateDistribution: DistributionV1 = {
+    type: "DISTRIBUTION",
+    timestampMs,
+    version: 1,
+    strategy: immediateStrategy,
+    receipts: immediateReceipts,
+  };
```
hammadj, line 117: These types have changed, so probably need to get #1743 and #1774 finalized and merged first so this can be refactored to conform to the types in those PRs

Beanow:
Please reopen after the grain ledger events have been settled :]

PR #1752, opened 2020-04-16
resolveAlias: add support for sourcecred identity

decentralion:
This commit modifies the resolveAlias function in the identity plugin's
alias module so that it now allows you to convert sourcecred identity
aliases (e.g. "sourcecred/user") back into NodeAddresses. This will be
necessary so that we can convert our ad-hoc distributions and transfers
(which use aliases, including SourceCred identity aliases) into the more
robust formats for the productionized grain ledger, which use full node
addresses.

I did a slight refactor on the identity module to expose a method for
constructing the address without the rest of the node.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/identity/alias.test.js
+++ b/src/plugins/identity/alias.test.js
@@ -3,6 +3,7 @@
 import {resolveAlias} from "./alias";
 import {loginAddress as githubAddress} from "../github/nodes";
 import {userAddress as discourseAddress} from "../discourse/address";
+import {identityAddress} from "./identity";
 
 describe("src/plugins/identity/alias", () => {
   describe("resolveAlias", () => {
@@ -75,6 +76,16 @@ describe("src/plugins/identity/alias", () => {
         const url = "https://example.com";
         const a = resolveAlias("discourse/login", url);
         const b = resolveAlias("discourse/@login", url);
+        expect(a).toEqual(b);
+      });
+      it("a sourcecred identity", () => {
+        const actual = resolveAlias("sourcecred/example", null);
+        const expected = identityAddress("example");
+        expect(actual).toEqual(expected);
+      });
+      it("a sourcecred identity with prefixed @", () => {
+        const a = resolveAlias("sourcecred/example", null);
+        const b = resolveAlias("sourcecred/@example", null);
         expect(a).toEqual(b);
       });
     });
--- a/src/plugins/identity/identity.js
+++ b/src/plugins/identity/identity.js
@@ -3,6 +3,7 @@
 import {NodeAddress, type Node} from "../../core/graph";
 import {nodePrefix} from "./declaration";
 import {type Alias} from "./alias";
+import type {NodeAddressT} from "../../core/graph";
 
 /**
  * A Username is a locally (within-instance) unique identifier for a user of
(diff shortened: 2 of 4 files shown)


src/plugins/identity/identity.js
```diff
@@ -31,16 +32,30 @@ export type IdentitySpec = {|
...
+ * Create a new node representing an identity.
+ */
+export function identityNode(identity: Identity): Node {
+  const username = validateUsername(identity.username);
+  const address = identityAddress(username);
   const description = `@${username}`;
   return {address, timestampMs: null, description};
 }
+
+export function identityAddress(username: Username): NodeAddressT {
```
hammadj, line 59: nit: since its a function, would be more readable as "getIdentityAddress" or "getAddressForUsername". "identityAddress" sounds more like a name for a variable.

src/plugins/identity/identity.js
```diff
@@ -31,16 +32,30 @@ export type IdentitySpec = {|
...
   }
-  const username = match[1];
-  const address = NodeAddress.append(nodePrefix, username);
+  return match[1];
+}
+
+/**
+ * Create a new node representing an identity.
+ */
+export function identityNode(identity: Identity): Node {
```
hammadj, line 52: Not sure if this is the standard naming pattern elsewhere, but it would be more intuitive if it was named "createNodeForIdentity" or "createIdentityNode"

wchargin: re: both naming questions: We don’t generally use `get*` prefixes except
where they are informative (usually, `get` suggests that an operation is
expensive or effectful, like reading from disk or network).

Yes, the pattern is standard (see, e.g., `Graph.node`). These functions
just return values. They don’t “retrieve” or “get” anything from
anywhere. Changing `identityNode` to `createNodeForIdentity` just takes
up space and energy without communicating more. These are fine as is.

src/plugins/identity/identity.js
wchargin: See above.

decentralion:
> Curious to know what would happen if there's naming collisions with identities, I imagine there's checks somewhere to prevent duplicate usernames?

Not sure off the top of my head, we'd need to look at the client code consuming the identities. It's out of scope for this PR.

src/plugins/identity/alias.test.js
```diff
@@ -77,6 +78,16 @@ describe("src/plugins/identity/alias", () => {
         const b = resolveAlias("discourse/@login", url);
         expect(a).toEqual(b);
       });
+      it("a sourcecred identity", () => {
+        const actual = resolveAlias("sourcecred/example", null);
+        const expected = identityAddress("example");
+        expect(actual).toEqual(expected);
+      });
+      it("a discourse login with prefixed @", () => {
```
Beanow, line 86: Nit: "discourse login" should be "sourcecred identity"

src/plugins/identity/identity.test.js
```diff
@@ -1,33 +1,41 @@
...
+      expect(() => identityAddress(identity.username)).toThrowError(
+        "Invalid username"
+      );
     });
     it("errors for a bad username", () => {
       const identity = {username: "$foo$bar", aliases: ["github/foo"]};
       expect(() => identityNode(identity)).toThrowError("Invalid username");
+      expect(() => identityAddress(identity.username)).toThrowError(
+        "Invalid username"
+      );
```
Beanow, line 30: Nit: have some preference to split up the tests for `validateUsername`, `identityNode` and `identityAddress`.

The pattern I usually go for is to:
- Test `validateUsername` with a number of good and bad example inputs.
- Test `identityAddress` does username validation at all, testing one bad input throws.
- Test `identityNode` does username validation at all, testing one bad input throws.

src/plugins/identity/alias.test.js
decentralion: Done.

src/plugins/identity/identity.test.js
decentralion: Exported `validateUsername` and added explicit tests.
Left the existing tests as is--I'm ok with having a bit of redundant testing.

PR #1754, opened 2020-04-20
Initiatives: add normalizeNodeEntry implementation

Beanow:
Depends on #1753

This is where most flexibility when hand-writing JSON files is expected
to come from. As it makes few assumptions about the formatting but the
internal normalized type is still consistent.

Test plan: `yarn test`


Beanow:
Tests currently failing because of GitHub server issues. CI is failing to do a git checkout.

Beanow:
The current rerun of CI seems to not have the checkout issue, so moving ahead.

PR #1755, opened 2020-04-20
Util: create findDuplicates

Beanow:
Often we can use representations like `Set` to avoid duplicates in the first
place. However when duplicates are not allowed in some kind of user input
we may want to present them with a useful error message.

This util will find the duplicate elements, so they can be highlighted in an error.

Test plan: `yarn test`

### DIFF
--- a/src/util/findDuplicates.js
+++ b/src/util/findDuplicates.js
@@ -1,0 +1,18 @@
+// @flow
+
+/**
+ * Finds elements in the array which are included twice or more.
+ * Uses a === comparison, not deep equality.
+ */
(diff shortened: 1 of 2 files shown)


Beanow:
> I'm worried people might mis-use this function, expecting it to do deep inspection and not shallow. But there's not any convenient way to add support for deep equality (since we can't assume serializability, we'd need to do n^2 deep equality checks) and it's documented in the comment, so approved.

Agreed yeah. I don't think the use-case has shown up much at all. We need uniqueness regularly, but rarely need to know what the duplicates are. So I think the comment should work for now, and handle a deep comparison if the need arises.

PR #1756, opened 2020-04-20
Initiatives: create EdgeSpec type and normalization

Beanow:
Depends on #1753, #1754, #1755

This type will replace the current `$ReadOnlyArray<URL>` on the Initiative
fields to support `NodeEntry` arrays as well. Like with the NodeEntryJson
they will have a manual-entry convenience flavor and an internally
normalized one.

The normalization function aims to throw errors that help users notice any
input mistakes.

Test plan: `yarn test`

You can view a common example of this intended format at:
https://github.com/sourcecred/sourcecred/blob/b10ce1548a449a0ece9e0876ba34041c867ccd57/src/plugins/initiatives/example/initiative-C.json#L15-L23


Beanow:
> We should think about how we will add manual edge-weights per connection, since it's a really intuitive feature to want.

I have thought about this, quite a lot. And is part of why this feature took me so many iterations of different formats and prototypes. It's a rabbit hole I could talk about for hours, but let me jump straight to why I suggest not supporting it now.

I would consider the value of this feature in terms of how much it enables you to improve the **quality of cred**. And a secondary is we don't want to sacrifice too much in terms of **ease of entry**.

I think there's no low-hanging fruit here. Sticking with the "dependencies / references / contributions" categories is the low-hanging fruit to improve cred quality and having ease of entry too.

---

If you want a more detailed breakdown:
The current situation is, we have introduced concepts to distinguish contributions:
- dependencies
- references
- contributions

They offer much convenience when entering data, because we can define the dynamics once by setting type-level edge weights for them, and simply categorize contributions from there. The main downsides I think are having to learn concepts, and some contributions will be awkward fits. Used consistently though I think we can get some great leverage on a simple construct. It does allow you to model different cred flows for different types of contributions and relationships.

The way I feel you could improve cred from that baseline is: having _a more powerful descriptive language_ than using these contribution categories. To me `+manualEdgeWeight: EdgeWeight` does not qualify as more powerful. We've lost control over the backward edge weight, and we've lost contextual information about why it has this edge weight. It loses out on the secondary too, as it's more difficult to enter this data in terms of less intuitive JSON format, and in terms of needing to decide on an extra set of weights.

There's 2 main routes I see to having a substantially more powerful descriptive language:
1. Percentage edge weights support. For example setting the champions edge to 20% of outbound weight. Not really unique to initiatives, but definitely opens a lot of options. Currently that means a dependency on CredRank.
2. A "Manual graph DSL" level of total graph control. Which I would love to have, and let Initiatives implement, possibly making the dependencies/references/contributions terms obsolete. However this is a big effort, not low hanging fruit. _And_ without a very neatly designed DSL will be significantly worse on our secondary objective of ease of entry.

That's where I get to the conclusion of, I think there's no low hanging fruit here.

We probably can improve quality of cred, while not sacrificing ease of entry much. But not in a matter of weeks. We're talking about building CredRank or the Manual graph DSL.

Well out of scope for beta1 and not in line with urgently offering more recognition for contributions outside the GitHub/Discourse/Discord plugins.

Beanow:
I'm interpreting the approval as, manual edge weights are a future want-to-have but not a blocker for this particular PR. So merging this for now, and can continue the feature discussion separately :]

PR #1758, opened 2020-04-20
Initiatives: add addressForNodeEntry

Beanow:
Depends on #1753, #1757

Defines the NodeAddress format we want to use for a NodeEntry.

Because we want to use the parent `InitiativeId` as parts of the address,
we'll need to read the underlying `string[]`, changing the opaque type.

Test plan: `yarn test`


Beanow:
Rebasing caused conflicts with #1754. They both created a new tests file and appended functions to nodeEntry. So they just happened to be on the same lines. Nothing changed implementation wise, but re-requesting review anyway.

```
# Conflicts:
#	src/plugins/initiatives/nodeEntry.js
#	src/plugins/initiatives/nodeEntry.test.js
```

PR #1763, opened 2020-04-23
Initiatives: add EdgeSpecJson support to InitativeFile v0.2.0

Beanow:
An additional change here is we're allowing more keys to be omitted in
the JSON format. This is both intuitive for data entry, and safer in
terms of Flow types (as JSON.parse returns any).

The test examples now cover a v0.1.0 (initiative-A), v0.2.0 with just
URLs (initiative-B) and one with just entries (initiative-C).

To make this commit smaller and easier to review, we're not yet adding
`EdgeSpec` to the `Initiative` type and will ignore the entries when
converting from `InitiativeFile` to `Initiative`.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/initiativesDirectory.js
+++ b/src/plugins/initiatives/initiativesDirectory.js
@@ -11,6 +11,7 @@ import {
   type ReferenceDetector,
   MappedReferenceDetector,
 } from "../../core/references";
+import {type EdgeSpecJson} from "./edgeSpec";
 import {
   type Initiative,
   type InitiativeRepository,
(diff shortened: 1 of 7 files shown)


src/plugins/initiatives/initiativesDirectory.js
```diff
@@ -148,18 +149,41 @@ export function _convertToInitiatives(
...
 }
 
+// TODO: this is a temporary function, which reverts an ?EdgeSpecJson to just
+// $ReadOnlyArray<URL>. It only exists to allow the `Initiative` type to add
+// support for EdgeSpec in a separate commit.
+export function _lossyURLFromEdgeSpecJson(
+  json?: EdgeSpecJson
+): $ReadOnlyArray<URL> {
+  return (json || {}).urls || [];
+}
```
Beanow, line 185: Note: this is removed in #1764.

PR #1764, opened 2020-04-23
Initiatives: add EdgeSpec support to Initiative type

Beanow:
Depends on #1763

Because the `Initiative` type now supports `EdgeSpec`, we're no longer
discarding entries when converting between `InitiativeFile` and
`Initiative`.

To make this commit smaller and easier to review, we're not yet adding
support to add `NodeEntry` to the graph though, instead
`createWeightedGraph` ignores the entries for now.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/initiatives/createGraph.js
+++ b/src/plugins/initiatives/createGraph.js
@@ -12,6 +12,7 @@ import {type WeightedGraph as WeightedGraphT} from "../../core/weightedGraph";
 import * as WeightedGraph from "../../core/weightedGraph";
 import type {NodeWeight} from "../../core/weights";
 import type {ReferenceDetector, URL} from "../../core/references";
+import type {EdgeSpec} from "./edgeSpec";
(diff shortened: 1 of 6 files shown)


src/plugins/initiatives/createGraph.js
```diff
@@ -90,9 +91,14 @@ export function createWeightedGraph(
...
     // Generic approach to adding edges when the reference detector has a hit.
     const edgeHandler = (
-      urls: $ReadOnlyArray<URL>,
+      edges: $ReadOnlyArray<URL> | EdgeSpec,
       createEdge: EdgeFactoryT
     ) => {
+      // TODO: this is a temporary implementation, which takes an EdgeSpec and
+      // just takes the $ReadOnlyArray<URL>. This is only done to allow support
+      // for graphing NodeEntries to be added in a separate commit.
+      const urls = Array.isArray(edges) ? edges : edges.urls;
```
Beanow, line 100: Note: this is removed in #1767

PR #1767, opened 2020-04-24
Initiatives: add EdgeSpec support to createGraph

Beanow:
Depends on #1764, #1763, #1758

As of this commit the plugin should fully support `EdgeSpec`. Meaning the
entries are included in the Graph and Cred computations.

As the champions field does support URLs but not NodeEntry from an
EdgeSpec, we're separating the URL handler from the EdgeSpec handler.

Test plan:
- `yarn test` should show nodes, edges and weights are correctly included.
- (optional) loading a project with Initiatives that have entries, and confirm it includes entries in the output Graph and Cred computation.

Note: this (like #1757) would be improved with a sharness test to automate the "load a project with initiatives that have entries" test.


src/plugins/initiatives/createGraph.js
```diff
@@ -89,28 +112,58 @@ export function createWeightedGraph(
       weights.nodeWeights.set(node.address, weight);
     }
 
-    // Generic approach to adding edges when the reference detector has a hit.
+    // Generic approach to adding edges and entries from EdgeSpec or URLs.
     const edgeHandler = (
       edges: $ReadOnlyArray<URL> | EdgeSpec,
-      createEdge: EdgeFactoryT
+      createEdge: EdgeFactoryT,
+      field: ?NodeEntryField
```
decentralion, line 119: Why is field optional?

Beanow: Champion URL reference detection uses this same handler, however we don't allow any NodeEntries for it. As NodeEntries describe a _contribution_ while we're expecting _users_.

Looking at this again though. This can be cleaned up. Let me factor it out to a new handler.

Beanow: Separated the handlers, so champions (the previous case where we took null here), just calls a `urlHandler`. And the others call the `edgeSpecHandler`, making this a required argument.

src/plugins/initiatives/createGraph.js
```diff
@@ -90,27 +113,58 @@ export function createWeightedGraph(
...
+      {urls, entries}: EdgeSpec,
+      createEdge: EdgeFactoryT,
+      field: NodeEntryField
+    ) => {
+      // Delegate handling the URLs.
+      urlHandler(urls, createEdge);
+
+      for (const entry of entries) {
+        // Add the NodeEntry contribution itself to the graph.
+        const node = nodeFromEntry(entry, initiative.id, field);
```
decentralion, line 138: nit: would be easier to understand if this were called `entryNode` rather than just `node`, since there are a few different nodes in play here

Beanow: Updated :+1:

PR #1770, opened 2020-04-25
flow: update `better-sqlite3` types for v7.0.0

wchargin:
Summary:
I first wrote these type definitions for v4.x.x. The library API has
changed since then. This patch updates the type definitions to match the
current API as of the v7.0.0 release:
<https://github.com/JoshuaWise/better-sqlite3/blob/v7.0.0/docs/api.md>

There are breaking changes, but none among functions that we use. On the
other hand, there are new features that will be useful to us.

Test Plan:
Running `yarn flow` still passes. There may be some errors among typing
of functions that we don’t actually use (particularly `aggregate`, which
is more complicated than the others). If so, we can cross those bridges
when we come to them.

wchargin-branch: flow-better-sqlite3-v7-api

### DIFF
--- a/flow-typed/npm/better-sqlite3_vx.x.x.js
+++ b/flow-typed/npm/better-sqlite3_vx.x.x.js
@@ -13,15 +13,24 @@ declare class bettersqlite3$Database {
     options?: bettersqlite3$Database$ConstructorOptions
   ): void;
   prepare(source: string): bettersqlite3$Statement;
-  exec(source: string): this;
-  transaction(sources: $ReadOnlyArray<string>): bettersqlite3$Transaction;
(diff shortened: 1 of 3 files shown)


wchargin:
We could, I guess? I’m confident that the core DML types are useful and
basically correct (or close enough that I haven’t run into issues with
any of the `Mirror` development). As mentioned, the type for `aggregate`
is definitely wrong, but perhaps still useful. There’s also the whole
bigint issue.

I’ll see if maybe I can send a PR to `flow-typed`, but it’s not super
high on my priority list, and I’m not interested in being on the hook to
maintain them diligently. (E.g., this PR to update the types for v5+ is
a year and a half late, which is fine for us because we didn’t need them
in the interim.)

PR #1774, opened 2020-04-29
Add types for the grain ledger

decentralion:
This is based on work in #1697, but streamlined to just have the event
and ledger types and no implementation. If we agree on these details, we
can then implement the logic in a followup.

Test plan: Human review. `yarn test` passes.

### DIFF
--- a/src/grain/ledger.js
+++ b/src/grain/ledger.js
@@ -1,0 +1,113 @@
+// @flow
+
+import {type NodeAddressT} from "../core/graph";
+import {type Grain} from "./grain";
+
+/**
+ * Tracks a transfer of grain between participants.
+ *
+ * A grainholder may transfer grain to themself (which is a no-op).
+ * A transfer that would result in the sender having a negative balance
+ * is illegal.
+ */
+export type GrainTransferredV1 = {|
+  +type: "GRAIN_TRANSFERRED",
+  +version: number,
+  +sender: NodeAddressT,
+  +recipient: NodeAddressT,
+  +amount: Grain,
+  +timestampMs: number,
+  +memo: string,
+|};
+
+// TODO: Import GrainAllocation once it merges.
+type GrainAllocation = {};
+
+/**
+ * Tracks a distrbution of newly-minted grain to contributors.
+ *
+ * Each Distribution contains zero or more allocations, which
+ * actually contain the mapping of new grain amounts to contributors.
+ */
+export type GrainDistributedV1 = {|
+  +type: "GRAIN_DISTRIBUTED",
+  +version: number,
+  +timestamMs: number,
+  +memo: string,
+  +allocations: $ReadOnlyArray<GrainAllocation>,
+|};
+
+/**
+ * Record that an "alias" identity is equivalent to a "canonical" identity.
+ *
+ * This event permanently records that one user NodeAddress (the alias) is
+ * equivalent to another NodeAddress (the canonical address). this is useful
+ * if, for example, a user has both a GitHub and a Discourse account. Once the
+ * identities are merged, they cannot be separated (as doing so risks creating
+ * negative balances). We may need to implement un-merging in the future; see
+ * discussion below:
+ * https://github.com/sourcecred/sourcecred/pull/1774#discussion_r419179404
+ *
+ * If a user has multiple aliases, then they should have multiple identity
+ * merge events.
+ *
+ * If an identity merge has the same address as alias and canonical, it is a
+ * no-op.
+ *
+ * It is allowed for the "canonical" address of one identity merge to later
+ * be the "alias" of another merge.
+ *
+ * After the identity merge, any events which reference an alias address will
+ * be treated as though they referenced the canonical address. (E.g., a
+ * transfer to an alias will be received by the canonical address.) This rule
+ * even applies to future identity merge events. As such, if we had an identity
+ * merge saying that "A" is an alias of "B", and a future merge saying that "B"
+ * is an alias of "A", this would be re-written to saying that "B" is an alias
+ * of "B", which is a no-op.
+ */
+export type IdentityMergedV1 = {|
+  +type: "IDENTITY_MERGED",
+  +version: number,
+  +timestampMs: number,
+  +alias: NodeAddressT,
+  +canonical: NodeAddressT,
+  +memo: string,
+|};
+
+export type LedgerEvent =
+  | GrainDistributedV1
+  | IdentityMergedV1
+  | GrainTransferredV1;
+
+/**
+ * A ledger tracks balances, earnings, and the event history of participants.
+ *
+ * We may refactor the `events` to be an async generator in the future,
+ * when we stop storing all the events in memory.
+ */
+export interface Ledger {
+  /**
+   * Stores the current grain balance of each address.
+   */
+  balances(): Map<NodeAddressT, Grain>;
+
+  /**
+   * Stores the lifetime earnings of each address.
+   *
+   * Necessary for computing which participants have been underpaid.
+   */
+  earnings(): Map<NodeAddressT, Grain>;
+
+  /**
+   * Retrieve the full history of all LedgerEvents.
+   */
+  events(): $ReadOnlyArray<LedgerEvent>;
+
+  /**
+   * For each address, list all known aliases.
+   *
+   * Addresses that are present in the history but have no aliases will be
+   * included, and will map to an empty array.
+   */
+  aliases(): Map<NodeAddressT, $ReadOnlyArray<NodeAddressT>>;
+}


src/grain/ledger.js
```diff
@@ -0,0 +1,111 @@
...
+|};
+
+/**
+ * Record that an "alias" identity is equivalent to a "canonical" identity.
+ *
+ * This event permanently records that one user NodeAddress (the alias) is
+ * equivalent to another NodeAddress (the canonical address). this is useful
+ * if, for example, a user has both a GitHub and a Discourse account. Once the
+ * identities are merged, they cannot be separated (as doing so risks creating
+ * negative balances).
```
hammadj, line 47: What about the situation where someone merges an identity that doesn't actually belong to them and then the original owner of the identity wants to claim their identity back?

src/grain/ledger.js
```diff
@@ -0,0 +1,111 @@
...
+export type LedgerEvent =
+  | GrainDistributedV1
+  | IdentityMergedV1
+  | GrainTransferredV1;
+
+/**
+ * A ledger tracks balances, earnings, and the event history of participants.
+ *
+ * We may refactor the `events` to be an async generator in the future,
+ * when we stop storing all the events in memory.
```
hammadj, line 84: I assume the events will need to be persisted to disk/JSON in some way still? Just like scores.json?

src/grain/ledger.js
```diff
@@ -0,0 +1,111 @@
...
+   */
+  events(): $ReadOnlyArray<LedgerEvent>;
+
+  /**
+   * For each address, list all known aliases.
+   *
+   * Addresses that are present in the history but have no aliases will be
+   * included, and will map to an empty array.
+   */
+  aliases(): Map<NodeAddressT, $ReadOnlyArray<NodeAddressT>>;
```
hammadj, line 110: How will this interplay with the identities field in project.json? Currently we are free to just edit that field as we need, but if the Ledger needs to know about when changes are made and prevent un-merging of identities etc, will we still be able to do it that way?

src/grain/ledger.js
decentralion: Woah, great question! I hadn't thought of that.

I suppose we could make an "identity unmerge" event that is explicit about how the grain balance gets split between the de-coupled identities. Also now that I think about it, I don't think we would actually get negative balance problems from splitting the identity. However, what would be tricky is retroactively splitting up their combined grain earnings, so we can figure out underpayment properly.

To see an example of when this matters, consider the following attack:
- I merge github/decentralion (which earns most of my "underlying" cred) and sourcecred/decentralion, with sourcecred/decentralion as canonical
- I earn a bunch of grain, paid to sourcecred/decentralion
- I unmerge, a naive algorithm doesn't move around my lifetime earnings to show that a lot of sourcecred/decentralion's earnings now retroactively look like they were earned by github/decentralion
- Now the balanced allocation strategy sees that github/decentralion has a ton of cred but no earnings, so I get a big allocation -- thus double paying me for my cred

So, when doing the unmerge, we would need some way to adjust the history for all the past earnings. We could do this by computing the historical cred scores at the time of the unmerge, looking at the ratio between the different identities cred in each period for which we had a distribution, and re-approportioning each distribution.

Which means the unmerge event would need to include both how to split the grain balance and how to split the earnings, basically. Seems complicated. Let's not implement it for now, but this is a great question to be asking, since we're likely to encounter this in the future.

src/grain/ledger.js
decentralion: Yeah, for now I'm just thinking JSON files.

src/grain/ledger.js
decentralion: I'm thinking once we put together the instance system, an "instance update" system will look like
- run thru the ledger to get the expected identity mapping
- compare to the project.json file
- if new identity contractions, write merge events to the ledger. if identity separations, error.
- compute cred for project, write graph and scores to disk.
- compute grain distribution based on that cred, write distribution events to ledger.

src/grain/ledger.js
hammadj: Yea I think its messy to deal with un-merging accounts, even for the future. The ideal way to handle it IMO is actually include "proofs of ownership" as a part of the identity definition. e.g. SourceCred can generate an ID and ask me to tweet it to prove that I own a certain Twitter account. It can ask me to create a gist on GitHub with that ID to prove I own a certain GitHub account. That prevents the ability for anyone to merge an identity that they haven't proved ownership of in the first place.

3Box and Keybase both do this to verify your various social accounts: https://docs.3box.io/faq/data-identity

The reason I even thought to ask this is because we are planning to implement this proof of identity system for MetaGame players so we can automate the whole process instead of manually having to edit the JSON file. MetaGame profiles are all going to be built on 3Box since its fully extensible and designed to be an open identity standard, would be awesome to be able to encode those proofs of identity as a part of SC as well.

Fun fact: The same team behind 3box is now building Ceramic!

decentralion:
@Beanow @hammadj seems this should be ready to merge.

src/grain/ledger.js
```diff
@@ -0,0 +1,113 @@
...
+  | IdentityMergedV1
+  | GrainTransferredV1;
+
+/**
+ * A ledger tracks balances, earnings, and the event history of participants.
+ *
+ * We may refactor the `events` to be an async generator in the future,
+ * when we stop storing all the events in memory.
+ */
+export interface Ledger {
```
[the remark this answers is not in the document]
Beanow: Additionally I'm slightly on the fence about whether this should be an interface or type.
Because these are all in-memory and projected beforehand. It could be just be a plain (deep frozen?) object that's the return type of a project function.

src/grain/ledger.js
Beanow: Because projections are transient, I think it's worth taking a look at what use-case we have for this first.
We can disregard data we don't need, and tweak the data structure to be specific for our use-case.

For example if we do `(old) => new` resolving most often, changing the map from `Map<new, old[]>` to `Map<old, new>` would give more efficient lookups. Maybe both directions are common and we want to have both maps.

My guess is, `Map<old, new>` will be used for the projection logic itself, if we want to support the post-merge event redirecting of old addresses to the new. Also think that future producers of events (commands) should prefer using the new address, so they would use this lookup too. Not seeing a huge need for the inverse lookup.

Beanow: > if new identity contractions, write merge events to the ledger. if identity separations, error.

I do think this is a good approach, but do want to point out, it's not the ledger that should care about this. The ledger with the current business rules (post-merge the old addresses are considered aliases of the new address), would support this scenario just fine.

Suppose: `github/Beanow` and `discourse/Beanow` merged into `sourcecred/Beanow`. The 2 merge events are added to the ledger, the old addresses now redirect to `sourcecred/Beanow`.

Should the project.json remove this mapping, I would receive Cred like this:
- `github/Beanow: 321`
- `discourse/Beanow: 123`
- `sourcecred/Beanow: undefined`

And the grain distribution strategies would also allocate on `github/Beanow` and `discourse/Beanow` accordingly. The mapping would then redirect these allocations to `sourcecred/Beanow`.

Nothing crashes here, and in terms of events this looks like a totally valid history!

There's 2 problems with this though.
For one the Cred scores would not be accurate here. Changing the graph topology makes a difference for Cred flow so _all the Cred scores_, not just mine, would be skewed. Skewing the Grain allocations along with it.

Second is, there's a semantics problem, if the "unlinking" in the project.json was intentional and supposed to mean that these are not to be considered the same identity any more, then the end result of it going to a single account is unexpected.

src/grain/ledger.js
Beanow: Unmerging I think will be messy indeed. Although fairly easy to support in the ledger.

It's not unlike a divorce situation. Having used a shared account for some time, it's infeasible to tell who should own what. So you'll have to make some imperfect choice, 50:50? Some other balance? :shrug: you'll have to figure that out _somehow_.

But as far as the ledger is concerned, it's not responsible for how to decide this. It just needs to record the result as an event.

For example `ACCOUNTS_SPLIT` (as the opposite of `ACCOUNTS_MERGED`). It would need to store details about which addresses are involved, and how the Grain is divided. Plus some business rules, like the invariant `sum(grain shares) == original grain` and to undo the the redirecting of aliases.

None of that is rocket science per say on the accounting side. It's the problem of how to divide the Grain that makes it difficult. And I think it's unlikely we'll have a one-size-fits-all solution to support doing this automatically. So crashing here and have an instance maintainer input these things is what I imagine the mid-term solution being.

decentralion: We need to split up both the Grain, and the historical Grain distribution. From perspective of the integrity of the system as a whole, we don't really care who gets the grain, provided that `sum(grain shares) == original grain` gets maintained. However, we do care how the historical grain distribution is attributed. Otherwise you could bind your high-cred account to a low-cred account, and then split accounts in such a way that your historical earnings get attributed to the low-cred account, and then you get to reap double distributions when the lifetime payout strategy sees that your high-cred account is underpaid.

The correct way to do this is to know the Cred scores *after* the split has been applied, and use that as a basis for computing the ledger update. But that is kind of tricky, because I'd like to have the Cred score computation pipeline use the Ledger as SoT for identities, but actually you'd need to compute Cred on the new (split) identities to get the data needed to properly split them.

src/grain/ledger.js
```diff
@@ -0,0 +1,113 @@
...
+ * After the identity merge, any events which reference an alias address will
+ * be treated as though they referenced the canonical address. (E.g., a
+ * transfer to an alias will be received by the canonical address.) This rule
+ * even applies to future identity merge events. As such, if we had an identity
+ * merge saying that "A" is an alias of "B", and a future merge saying that "B"
+ * is an alias of "A", this would be re-written to saying that "B" is an alias
+ * of "B", which is a no-op.
+ */
+export type IdentityMergedV1 = {|
+  +type: "IDENTITY_MERGED",
```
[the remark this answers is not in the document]
decentralion: I think coupling the user identities to the grain ledger makes sense, because we do really need these two to be in a consistent state. Otherwise you have a situation where:

T0:
Ledger: Foo and Bar are separate identities
Graph: Foo and Bar are separate identities
Result: Foo and Bar both earn Grain

T1:
Ledger: Foo and Bar are separate identities
Graph: Bar is an alias of Foo
Result: Foo now has really big Cred relative to past issuance. Foo gets "double-dipping" Grain, because we haven't properly netted out Bar's past allocation

Thus: if we allow the Cred Graph and the Ledger to get out of sync, we accidentally double-issue past Grain.

decentralion:
Superseded by all my other work on the Grain ledger.

PR #1780, opened 2020-05-06
Discord: add Mirror class

BrianLitwin:
Creates a class to handle the implementation
of fetching data from the Discord server and
saving it to the local mirror.

Test plan:

Unit tests added.


src/plugins/discord/mirror.js
```diff
@@ -0,0 +1,111 @@
...
+      after: Snowflake
+    ) => {
+      return await this._fetcher.members(this._guildId, after);
+    };
+
+    const processMember: (Model.GuildMember) => void = (
+      member: Model.GuildMember
+    ) => this._sqliteMirror.addMember(member);
+
+    await this._paginatedFetch(fetchMembers, processMember, "0");
```
BrianLitwin, line 63: I tried passing `this._sqliteMirror.addMember` directly, instead of writing `processMember`; I was getting an error in the `sqliteMirror` about `this` being `none`, where `this` should have referenced the `sqliteMirror`.

src/plugins/discord/mirror.test.js
```diff
@@ -0,0 +1,367 @@
...
+        testMessage("2"),
+        testMessage("3"),
+      ]);
+      expect(messages.mock.calls).toEqual([
+        [channelId, "0"],
+        [channelId, "2"],
+      ]);
+    });
+
+    it("fetches reactions", async () => {
```
BrianLitwin, line 217: I'm not super happy with the amount of state we need to build up to run this test. An option I considered was testing that the `_fetchReactions` method was called with the correct parameters and then testing `_fetchReactions` in isolation. However, that felt like testing the implementation versus the behavior. Note, I'm not crystal clear on that distinction/testing methodology.

src/plugins/discord/mirror.js
Beanow: Yeah, if you call the function outside of it's object context, `this` will be unset. There's a few ways around that using any of `Function.prototype.apply`, `call` or `bind`. Bind being easiest but also performance heavy, so avoid using in hot loops.

```js
myObj.func();
// this == myObj

let func = myObj.func;
func();
// this == undefined

func = myObj.func.bind(myObj);
func();
// this == myObj
```

Beanow:
I've created an example of how I imagine async generators can greatly simplify this.
https://github.com/sourcecred/sourcecred/compare/bean-mirror-example

The first commit: https://github.com/sourcecred/sourcecred/commit/f39c78541fbe10f5423c079610910cefca6ac638
Shows just how close your function was to being a generator :smile:
Adding a `*`, using `yield` and changing the return type.
```diff
-  async _paginatedFetch<T>(
+  async *_paginatedFetch<T>(
...
-  ): Promise<void> {
+  ): AsyncGenerator<T, void, void> {
...
-        await process(result);
+        yield result;
```

The second: https://github.com/sourcecred/sourcecred/commit/4a9be9b977995fa256a4e9e78aaef4c6a5bb5be4
Separate the functions to create these generators, from the code that consumes them.
So all SQLite stuff now happens in the update functions.

The third: https://github.com/sourcecred/sourcecred/commit/aa2a52bd24f6db8debdeed291459a78de297fe2f
Shows that once you've separated producer and consumer, handling the pagination can be it's own utility wrapper around Fetcher and Mirror becomes much simpler.

src/plugins/discord/mirror.js
```diff
@@ -0,0 +1,78 @@
...
+  ): Promise<$ReadOnlyArray<Model.Reaction>>;
+}
+
+export async function getPages<T>(
+  fetchPage: (endCursor: Snowflake) => Promise<ResultPage<T>>,
+  endCursor: Snowflake
+): Promise<$ReadOnlyArray<T>> {
+  let after = endCursor;
+  let hasNextPage = true;
+  let allResults = [];
```
BrianLitwin, line 62: This implementation accumulates results in this variable. The upside is it makes things very simple. The tradeoff may be that channels have tens of thousands of messages. It's not a lot of added complexity to pass in a callback that processes these results individually, instead of accumulating them all and returning them. However, opting for the simplest version in this pass.

BrianLitwin:
@decentralion this is the pr I mentioned in the dev meeting today

src/plugins/discord/mirror.js
decentralion: Cool. Yeah an async generator would conceptually fit well here, but fine with this implementation for now.

src/plugins/discord/mirror.js
```diff
@@ -0,0 +1,78 @@
...
+        )) {
+          sqliteMirror.addReaction(reaction);
+        }
+      }
+    }
+  }
+}
+
+// Note: most of this is about wrapping pagination.
+export interface StreamingFetcher {
```
decentralion, line 41: Let's call this a `DepaginatedFetcher`, since it no longer actually streams.

decentralion:
@BrianLitwin I went ahead and fixed my nits myself, since they were all quite trivial. Going to merge with the fixes. Thanks for the PR!

Issue #1773, opened 2020-04-26
Improve SourceCred data output format


decentralion:
Over on the forums, @s-ben [asks about how to analyze SourceCred data](https://discourse.sourcecred.io/t/analyzing-sourcecred-data/704/). At present, the data we export via the [scores CLI](https://github.com/sourcecred/sourcecred/pull/1223) only includes user data, which makes many interesting analyses impossible.

I think we should come up with a simple output representation that is more comprehensive and allows users to explore many common questions about the cred scores. For now, I propose that we don't actually encode the graph here or try to give users the ability to validate or recompute the scores, since all of that effort would need to be re-done after we switch to CredRank (#1686).

Here's a format that we could use, which would be "forwards compatible" to CredRank (i.e. it pre-processes the data so that it looks more like how the data will after CredRank, i.e. we only have timing info for contributors, not for the contributions). It would also provide explicit view into where [cred minting](https://roamresearch.com/#/app/SourceCred/page/bD3-yHgbX) is happening.

```js
type Contribution = {
  date: ISOTimestamp,
  address: string[],
  cred: number,
  mint: number,
};
type Contributor = {
  scores: {date: ISOTimestamp, cred: number}[],
  address: string[],
};
type Scores = {contributors: Contributor[], contributions: Contribution[]};

```

Here are some examples:
```js
const contribution = {
  date: "Sun Apr 26 2020 12:02:12 GMT-0700 (Pacific Daylight Time)",
  address: [
    "sourcecred", 
    "discourse", 
    "https://discourse.sourcecred.io", 
    "TOPIC", 
    "704"
  ],
  cred: 12.3,
  mint: 0,
}

const contributor = {
  address: ["sourcecred", "identity", "IDENTITY", "s-ben"],
  scores: [
    {
      date: "Sat Feb 01 2020 00:00:00 GMT-0800 (Pacific Standard Time)", 
      cred: 432.1
    }, 
    ...
  ]
}
```

I think we should consider including this in the Quality-of-Life initiative.

decentralion:
Here's a slightly tweaked version, to deduplicate storing the dates:

```js
type Contribution = {
  date: ISOTimestamp,
  address: string[],
  cred: number,
  mint: number,
};
type Contributor = {
  scores: number[], // indexed by intervalEndpoints
  address: string[],
};
type Scores = {
  contributors: Contributor[], 
  contributions: Contribution[], 
  intervalEndpoints: ISOTimestamp[],
};
```

[...]

PR #1781, opened 2020-05-06
Add output data formats

decentralion:
As explored in #1773, this commit adds some output data formats that we
can use to enable data analysis of cred instances, along with powering
new UIs. I've included three output formats in order of increasing
complexity, with the intention that we can ship the simpler output
formats quickly, and then enhance with more information to power better
analyses.

I've tried to ensure forward-compatibility with CredRank, so that we can
migrate to CredRank without needing to make major changes to this API.

I may want to include type information in the output format as well.

Test plan: Human review and consideration. `yarn test` passes.

Thanks to @s-ben and @Beanow for discussion, review, and inspiration.

### DIFF
--- a/src/analysis/output.js
+++ b/src/analysis/output.js
@@ -1,0 +1,106 @@
+// @flow
+
+/**
+ * This module defines a rich output format for cred scores, so that we can use
+ * it to drive UIs and data analysis.
+ */
+import type {Alias} from "../plugins/identity/alias";
+import type {PluginDeclaration} from "./pluginDeclaration";
+import type {TimestampMs} from "../util/timestamp";
+
+export type Index = number;
+export type CredFlow = {|+forwards: number, +backwards: number|};
+
+/**
+ * Describes an individual node in the contribution graph.
+ * Includes the information in the raw Graph node, along with scoring
+ * information, i.e. its total cred, and the amount of cred that it minted.
+ *
+ * While we are in TimelineCred, the "cred" field will have the total cred,
+ * summed across all time slices. Once we migrate to CredRank, it will just be
+ * the cred directly calculated by the CredRank algorithm.
+ */
+export type OutputNode = {|
+  +address: $ReadOnlyArray<string>,
+  +cred: number,
+  +minted: number,
+  +timestamp: TimestampMs | null,
+  // Description comes from the underlying Graph node, so it's determined by the
+  // plugin that added the node. May contain markdown.
+  +description: string,
+|};
+
+/**
+ * The first version of the output format will have the nodes, and the plugin
+ * declarations. Including the declarations makes it easy for consumers to do
+ * analyses like: how much cred was minted within each plugin? Within each
+ * contribution type? Etc.
+ */
+export type OutputV1 = {|
+  // Ordered by address
+  +orderedNodes: $ReadOnlyArray<OutputNode>,
+  +plugins: $ReadOnlyArray<PluginDeclaration>,
+|};
+
+/**
+ * Extra data for Contributors. Note that each Contributor corresponds
+ * to a specific node in the graph, which is retrievable via the NodeIndex.
+ *
+ * For contributors, we'll give the cred-over-time. We don't include this
+ * information for regular contribution nodes, both because it's not that
+ * interesting, and because once we migrate to CredRank, we won't have that
+ * level of resolution anymore.
+ *
+ * While we are in TimelineCred, the credOverTime will be the raw cred that the
+ * contributor recieved in each time slice. Once we switch to TimelineCred, the
+ * credOverTime will instead be the amount of cred that flowed from the
+ * contributor's epoch node to that contributor for each time period. (Cred at
+ * the contributor's epoch node may also flow to other destinations, e.g. to
+ * past/future epoch nodes through the webbing edges, or to a team or sponsor.)
+ *
+ *
+ * In either case, it will be an invariant that the sum of the credOverTime will
+ * correspond to the cred score on the raw OutputNode (up to floating point
+ * precision).
+ */
+export type Contributor = {|
+  +alias: Alias,
+  // Indexed into the orderedNodes
+  +nodeIndex: Index,
+  // Matches the interval endpoints in the full output format.
+  +credOverTime: $ReadOnlyArray<number>,
+|};
+
+/**
+ * The second version of the output format includes explicit contributor data,
+ * with cred-over-time and convenient aliases for looking up particular
+ * contributors.
+ */
+export type OutputV2 = {|
+  +orderedNodes: $ReadOnlyArray<OutputNode>,
+  +plugins: $ReadOnlyArray<PluginDeclaration>,
+  +contributors: $ReadOnlyArray<Contributor>,
+  +intervalEnd: $ReadOnlyArray<TimestampMs>,
+|};
+
+/**
+ * This edge format includes the standard Graph.Edge information, along with
+ * the cred flow across the edge. In TimelineCred, the cred flow will be summed
+ * across every time period; once we switch to CredRank, the summing will no
+ * longer be necessary.
+ */
+export type OutputEdge = {|
+  +address: $ReadOnlyArray<string>,
+  +credFlow: CredFlow,
+  +timestamp: TimestampMs,
+  +srcIndex: Index,
+  +dstIndex: Index,
+|};
+
+export type OutputV3 = {|
+  +orderedNodes: $ReadOnlyArray<OutputNode>,
+  +plugins: $ReadOnlyArray<PluginDeclaration>,
+  +contributors: $ReadOnlyArray<Contributor>,
+  +intervalEnd: $ReadOnlyArray<TimestampMs>,
+  +edges: $ReadOnlyArray<OutputEdge>,
+|};


src/analysis/output.js
```diff
@@ -0,0 +1,99 @@
...
+ * the amount of cred that it minted.
+ *
+ * While we are in TimelineCred, the "cred" field will have the total cred,
+ * summed across all time slices. Once we migrate to CredRank, it will just be
+ * the cred directly calculated by the CredRank algorithm.
+ */
+export type OutputNode = {|
+  +address: $ReadOnlyArray<string>,
+  +cred: number,
+  +mint: number,
```
Beanow, line 28: Some preference to call it `mints`.
Since this is presented as a current state snapshot, I think a present tense is fine.

src/analysis/output.js
```diff
@@ -0,0 +1,99 @@
...
+  +description: string,
+|};
+
+/**
+ * The first version of the output format will
+ * just have the nodes.
+ */
+export type OutputV1 = {|
+  // Ordered by address
+  +orderdNodes: $ReadOnlyArray<OutputNode>,
```
wchargin, line 41: s/orderd/ordered/ (throughout)

decentralion:
Note to WC: My hope is that when we land CredRank, we won't need to change these formats, we'll just include the epoch nodes in the list of contributions, and all the edges will be properly re-written so that edges that used to be incident to users will instead by incident to epoch nodes. Although, I'm not sure how we would include webbing edges, since they don't have an EdgeWeight per se.

src/analysis/output.js
```diff
@@ -0,0 +1,99 @@
...
+ * interesting, and because once we migrate to CredRank, we won't have that
+ * level of resolution anymore.
+ *
+ * While we are in TimelineCred, the credByTime will be the raw cred that the
+ * contributor recieved in each time slice. Once we switch to TimelineCred, the
+ * credByTime will instead be the amount of cred that flowed from the
+ * contributor's epoch node for each time period.
+ *
+ * In either case, it will be the an invariant that the sum of the credByTime
+ * will correspond to the cred score on the raw OutputNode.
```
wchargin, line 60: ”… (up to floating point imprecision).”

And strike *the* from “will be the an invariant”.

decentralion:
Maybe it would be better if in addition to including the raw EdgeWeights, we also have the normalized transitionProbabilities, in which case we could give webbing edges or sponsorship edges or boost edges a null EdgeWeight, but they would still have the transitionProbabilties? Although this would have to happen after CredRank, since right now we can't calculate transitionProbabilities, since they vary over time.

src/analysis/output.js
```diff
@@ -0,0 +1,99 @@
...
+ * This edge format includes the standard Graph.Edge information,
+ * along with the edge weight and (most interestingly), the
+ * cred flow across the edge. In TimelineCred, the cred flow
+ * will be summed across every time period; once we switch to CredRank,
+ * the summing will no longer be necessary.
+ */
+export type OutputEdge = {|
+  +address: $ReadOnlyArray<string>,
+  +edgeWeight: EdgeWeight,
+  +credFlow: EdgeWeight,
```
Beanow, line 92: This is interesting, we're sort of coercing absolute cred values into the `Weight` type, since they happen to both be numbers. For cleanliness would like a separate type. But otherwise the `{forward: x, backward: y}` structure works for me.

decentralion:
Also, I encode all the addresses as string-arrays so that consumers don't have to deal with NUL byte confusion. Though I also think we should ship lightweight consumer APIs in JS and Python that make it easier to use this data rather than needing to engage with the raw JSON. This would also make it easier to do common queries like finding all contributions matching one of the given types.

src/analysis/output.js
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * In either case, it will be the an invariant that the sum of the credByTime
+ * will correspond to the cred score on the raw OutputNode.
+ */
+export type Contributor = {|
+  +alias: Alias,
+  // Indexed into the orderedNodes
+  +nodeIndex: Index,
+  // Matches the interval endpoints in the full output format.
+  +credByTime: $ReadOnlyArray<number>,
```
Beanow, line 67: Nit: do think `credOverTime` is more common :]

src/analysis/output.js
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * For contributors, we'll give the cred-over-time. We don't include this
+ * information for regular contribution nodes, both because it's not that
+ * interesting, and because once we migrate to CredRank, we won't have that
+ * level of resolution anymore.
+ *
+ * While we are in TimelineCred, the credByTime will be the raw cred that the
+ * contributor recieved in each time slice. Once we switch to TimelineCred, the
+ * credByTime will instead be the amount of cred that flowed from the
+ * contributor's epoch node for each time period.
```
wchargin, line 57: Is it perhaps clearer to phrase this as “the amount of cred at the
contributor’s epoch node for each time period” rather than “…that flowed
from…” the epoch node? It’s technically equivalent for a stationary
distribution, but imho easier to think about instantaneously.

src/analysis/output.js
```diff
@@ -0,0 +1,104 @@
...
+ * This edge format includes the standard Graph.Edge information,
+ * along with the edge weight and (most interestingly), the
+ * cred flow across the edge. In TimelineCred, the cred flow
+ * will be summed across every time period; once we switch to CredRank,
+ * the summing will no longer be necessary.
+ */
+export type OutputEdge = {|
+  +address: $ReadOnlyArray<string>,
+  +edgeWeight: EdgeWeight,
+  +credFlow: EdgeWeight,
```
wchargin, line 92: > Maybe it would be better if in addition to including the raw
> EdgeWeights, we also have the normalized transitionProbabilities, in
> which case we could give webbing edges or sponsorship edges or boost
> edges a null EdgeWeight, but they would still have the
> transitionProbabilties?

My reading of this line (`credFlow: EdgeWeight`) was that this *was*
what was meant, and that the `EdgeWeight` type was being kind of coöpted
as a generic “bidirectional float”. If that’s not what was meant, please
clarify.

src/analysis/output.js
```diff
@@ -0,0 +1,104 @@
...
+ * This module defines a rich output format for cred scores, so that we can use
+ * it to drive UIs and data analysis.
+ */
+import type {Alias} from "../plugins/identity/alias";
+import type {EdgeWeight} from "../core/weights";
+import type {PluginDeclaration} from "./pluginDeclaration";
+
+// unenforced
+export type Integer = number;
+export type TimestampMs = Integer;
```
Beanow, line 13: We can use the TimestampMs type from the timestamp util.

Even if it would not be enforced if it came directly from `JSON.parse()`. As imo `JSON.parse()` should be treated as returning `mixed` rather than `any`.

src/analysis/output.js
```diff
@@ -0,0 +1,104 @@
...
+ * While we are in TimelineCred, the "cred" field will have the total cred,
+ * summed across all time slices. Once we migrate to CredRank, it will just be
+ * the cred directly calculated by the CredRank algorithm.
+ */
+export type OutputNode = {|
+  +address: $ReadOnlyArray<string>,
+  +cred: number,
+  +mint: number,
+  +timestamp: TimestampMs | null,
+  +description: string,
```
hammadj, line 30: How is the description generated / defined?

src/analysis/output.js
hammadj, line 67: Would be nice to get a breakdown of the cred to a contributor instead of just cred history. Knowing why/how someone was given a certain amount of cred is probably one of if not the most important data points to understand as a user of sourcecred.

Would that be possible in this current output structure? With outputV3 and output edges, could I just do something like this (rough code):

```ts
const getEdgesToNode = (output: OutputV3, nodeIndex: Index) => {
  const edgesToNode = output.edges.filter(e => e.dstIndex === nodeIndex);
  if (!edgesToNode.length) return [];
  
  return edgesToNode.map(edge => ({
    edge,
    sourceNode: output.orderedNodes[e.srcIndex],
    edgesToSourceNode: getCredFlowToNode(e.srcIndex), // recursively traverse graph
  }));
}

// find cred flow to a contributor
const edgesToContributor = getEdgesToNode(contributor.nodeIndex);
```

src/analysis/output.js
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * The first version of the output format will have the nodes, and the plugin
+ * declarations. Including the declarations makes it easy for consumers to do
+ * analyses like: how much cred was minted within each plugin? Within each
+ * contribution type? Etc.
+ */
+export type OutputV1 = {|
+  // Ordered by address
+  +orderdNodes: $ReadOnlyArray<OutputNode>,
```
hammadj, line 41: Why is it an array instead of a Map with the address as a key?

src/analysis/output.js
decentralion: Done.

src/analysis/output.js
decentralion: Done.

src/analysis/output.js
decentralion: No, it's not equivalent. The contributor's epoch node will flow to other nodes: e.g. thru the webbing edges, and (if sponsorship has occured) to the sponsor. Therefore, the contributor's share of their cred from that epoch node is the amount that flowed to the contributor from that epoch node, not the cred at the epoch node.

Updated the comment to make this clearer.

src/analysis/output.js
decentralion: Yeah, I co-opted the EdgeWeight as a sort of bidirectional float. My intention was for `EdgeWeight` to contain the original (unnormalized) edge weights, but I think that having it be the transition probability could be cleaner.

CredFlow shows how much cred flowed in each direction--re using the EdgeWeight type may be confusing. My intention at the moment is that the `edgeWeight` field will have the raw (unnormalized) edge weights, but I think we can take it out for now. In TimelineCred it's not tremendously helpful, and in CredRank we may want transition probabilties instead.

src/analysis/output.js
decentralion: Reasonable.

src/analysis/output.js
decentralion: OK, I declared an equivalent type called CredFlow locally.

src/analysis/output.js
decentralion: Changed.

src/analysis/output.js
decentralion: Done.

src/analysis/output.js
decentralion: The description comes from the underlying graph node, so it's defined by the plugin. Added a comment to clarify this.

src/analysis/output.js
decentralion: Correct, this kind of analysis becomes possible using the data from OutputV3. It won't be possible in OutputV2. The point of the explicit contributors field (other than the convenience of having aliases) is to provide time-resolution cred data for contributors when we don't provide it for all nodes in the graph. (for now, this is for space efficiency--as of CredRank, we actually will only compute time-resolved cred for contributors but not for nodes in general.)

src/analysis/output.js
decentralion: Several reasons:
- Including the nodes in an array means that every node as an index. Then we can normalize edges and contributors to reference nodes by index, which is much more efficient than referencing by address.
- This is explicitly intended as an output data format. Maps are not JSON-serializable.
- We're storing the addresses as arrays of strings because the raw address format has NUL bytes, and these are ugly when you serialize them in JSON.

decentralion:
Addressed review comments. @hammadj I don't think we should take another dep to database-ify this JSON... it would add a lot of complexity (new dep, new tech) for marginal gains. Esp. as a key goal here is supporting both Python and JS users, it'd be hard to align the dep for both cases.

Instead, I think we'll want to write a lightweight client over the data in both JS and Python which make it easy to do the domain-appropriate queries (find all entities by type, cred flow reconstructions, etc). No dep will understand the domain as well as we do, so we'll wind up with much cleaner APIs that way.

src/analysis/output.js
hammadj: For serialization problem, can just use a plain object with key/val being address/node

In terms of reference efficiency, key/val lookup on an object is O(1), and in JS the array actually is an Object with keys being 0, 1, 2, etc.

Since the ordering of the nodes in the array doesn't hold any semantic significance, I would argue that its actually less intuitive than key/val object. 

Additionally, these indices will be changing all the time as new nodes are added. Adding just a single node would cause the Index of every other node after it to change.

 In terms of pretty serialization, you could just hash the address string to come up with a "nodeId" instead of a "nodeIndex" as the key. That way if I see a certain nodeId in the serialized JSON I can easily look it up by searching for that NodeID. If it's just an array, there's no easy way to do so, would have to programatically look things up.

If I want to store references to certain NodeIds in other places, its important that the nodeIds be deterministic and based on the actual node address vs changing all the time. e.g. If I want to keep track of certain nodeIds in MetaGame web app database, it would only work if I had some level of guarantee that the nodeId would be consistent. Example use case would be storing each users NodeID in their profile and being able to easily look up their node, or just starting with an address and being able to easily map it to a node without having to know the index in advance. 

Additionally, the references to nodes wouldn't be changing all the time, it would be more "stable", and the git diffs wouldn't be as noisy.

hammadj:
> Addressed review comments. @hammadj I don't think we should take another dep to database-ify this JSON... it would add a lot of complexity (new dep, new tech) for marginal gains. Esp. as a key goal here is supporting both Python and JS users, it'd be hard to align the dep for both cases.
> 
> Instead, I think we'll want to write a lightweight client over the data in both JS and Python which make it easy to do the domain-appropriate queries (find all entities by type, cred flow reconstructions, etc). No dep will understand the domain as well as we do, so we'll wind up with much cleaner APIs that way.

Makes sense, yea that's exactly what I'm advocating for, an opinionated lightweight client over the data. Lowdb is just Lodash + Json, which is how I imagine we would be filtering and querying the data anyways.

src/analysis/output.js
decentralion: > In terms of reference efficiency, key/val lookup on an object is O(1), and in JS the array actually is an Object with keys being 0, 1, 2, etc.

The efficiency I'm worried about is space efficiency. Addresses are quite large, consider "sourcecred/sourcecred/github/PULL/sourcecred/sourcecred/1781" as an example. With this format, the number of addresses we need to write is `N + E`. With your format, we need to write `N + 3E + U`, where `N` is the number of nodes, `E` is the number of edges, and `U` is the number of users. That's why we do the same optimization in the graph serialization format, see: https://github.com/sourcecred/sourcecred/pull/374

Even with just the Output V1 format, the output data is 8.4M uncompressed for the sourcecred project. I believe that a large portion (greater than 50%, probably greater than 90%) of that data is address encoding. In practice the number of edges dominates the number of nodes, so it's not unreasonable to think that the index based format will have half the space requirement of a full address format.

> Since the ordering of the nodes in the array doesn't hold any semantic significance, I would argue that its actually less intuitive than key/val object.

This format is designed to be efficient, not intuitive. As I mentioned somewhere else, we'll make convenient wrapper clients in JS and Python that are intuitive to work with.

> Additionally, these indices will be changing all the time as new nodes are added. Adding just a single node would cause the Index of every other node after it to change.

Adding a new node to the output data is illegal. (To get any meaningful results, you'd need to add the node to the underlying graph, and then re-run cred analysis, which already requires data and context that isn't included in the output format.)

> If I want to store references to certain NodeIds in other places, its important that the nodeIds be deterministic and based on the actual node address vs changing all the time. e.g. If I want to keep track of certain nodeIds in MetaGame web app database, it would only work if I had some level of guarantee that the nodeId would be consistent. Example use case would be storing each users NodeID in their profile and being able to easily look up their node, or just starting with an address and being able to easily map it to a node without having to know the index in advance.

This is what node addresses are for. We don't need a new concept for this.

PR #1784, opened 2020-05-07
Enable address to alias conversion

decentralion:
This commit updates the alias module so that we may convert node
addresses into aliases. Naturally, the node address needs to be some
kind of user node address that is known to the aliasing scheme.

It's a big in-elegant that this creates a "hidden" integration point for
plugins, where plugins creating new user node types should add hardcoded
logic into the identity plugin's alias system. However, it is a
convenience and we currently use this system, so I'm just going to add
this functionality for now, and think about how the alias system should
work long term (or whether we should phase it out) for another
discussion.

This is needed for #1773.

I have `toAlias` return `null` when the address doesn't correspond to a
known aliasing scheme, rather than erroring. I think erroring would be
too harsh, given that it's quite possible that the user has loaded
third-party plugins that haven't registered aliasing schemes upstream
with us. In that case, the application should make a best effort attempt
to proceed without an alias (e.g. fallback to the full address), for
robustness.

Test plan: Unit tests included; `yarn test` passes.

### DIFF
--- a/src/plugins/identity/alias.js
+++ b/src/plugins/identity/alias.js
@@ -3,11 +3,15 @@
 import {type NodeAddressT} from "../../core/graph";
 import {githubOwnerPattern} from "../github/repoId";
 import {loginAddress as githubAddress} from "../github/nodes";
+import {userNodeType as githubUserType} from "../github/declaration";
 import {userAddress as discourseAddress} from "../discourse/address";
+import {userNodeType as discourseUserType} from "../discourse/declaration";
+import {identityType} from "./declaration";
 import {
   identityAddress,
   USERNAME_PATTERN as _VALID_IDENTITY_PATTERN,
 } from "./identity";
+import {NodeAddress} from "../../core/graph";
 
 /** An Alias is a string specification of an identity within another plugin.
  *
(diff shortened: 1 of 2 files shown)


Beanow:
> It's a big in-elegant that this creates a "hidden" integration point for
> plugins, where plugins creating new user node types should add hardcoded
> logic into the identity plugin's alias system. However, it is a
> convenience and we currently use this system,

Agreed, but that applies to the existing `resolveAlias` as well. For the future we could create a core interface for alias handling. Similar to what reference detectors do. For now we can punt on a refactor.

> This is needed for #1773.

I'm not directly seeing an "alias" field or mention of aliases there. Can you elaborate how we need this feature?

> I have toAlias return null when the address doesn't correspond to a
known aliasing scheme, rather than erroring.

This is a good approach for this abstraction level I think. It works well for reference detection as well which uses this pattern. For instance, here we're not considering a "miss" to be an error. Just a reason to check the next detector in the list.
https://github.com/sourcecred/sourcecred/blob/f183c5115914213a5da31e5822016f7e823f623e/src/core/references/cascadingReferenceDetector.js#L17-L22

src/plugins/identity/alias.js
```diff
@@ -64,3 +68,31 @@ export function resolveAlias(
...
+  let prefix = null;
+  if (NodeAddress.hasPrefix(n, githubUserType.prefix)) {
+    prefix = "github";
+  }
+  if (NodeAddress.hasPrefix(n, discourseUserType.prefix)) {
+    prefix = "discourse";
+  }
+  if (NodeAddress.hasPrefix(n, identityType.prefix)) {
+    prefix = "sourcecred";
+  }
```
Beanow, line 93: Nit: these should be mutually exclusive. Maybe use `else if`, early returns, or iife assignments?

Beanow: Actually, we could probably anticipate on the future "register your own resolvers" pattern, while solving this nit at the same time. What do you think about looping through a mapping?

```js
  const prefixes: Map<string, NodeAddressT> = new Map([
    ["github", githubUserType.prefix],
    ["discourse", discourseUserType.prefix],
    ["sourcecred", identityType.prefix],
  ]);

  for(const [prefix, nodePrefix] of prefixes.entries()) {
    if(NodeAddress.hasPrefix(n, nodePrefix)) {
      return `${prefix}/${terminator}`;
    }
  }

  return null;
```

decentralion:
> I'm not directly seeing an "alias" field or mention of aliases there. Can you elaborate how we need this feature?

See the actual output format as implemented here: https://github.com/sourcecred/sourcecred/pull/1781

```js
export type Contributor = {|
  +alias: Alias,
  // Indexed into the orderedNodes
  +nodeIndex: Index,
  // Matches the interval endpoints in the full output format.
  +credOverTime: $ReadOnlyArray<number>,
|};
```

Although I will actually change the alias field to `addressOrAlias: string | Alias` to accomodate the fact that some addresses may not have aliases, in which case we will use a stringified address `stringify(NodeAddress.toParts(address))`

src/plugins/identity/alias.js
decentralion: Yeah, that looks good to me. I'll do that.

PR #1786, opened 2020-05-07
Upgrade flow-bin

Beanow:
~~Continues from #1785~~
Closes #1711

Versions >= v0.121.0 require more specific suppression comments. Which caused the issues in #1711. See https://github.com/facebook/flow/releases/tag/v0.121.0

Test plan: `yarn flow`

### DIFF
--- a/src/plugins/github/nodes.test.js
+++ b/src/plugins/github/nodes.test.js
@@ -64,8 +64,8 @@ describe("plugins/github/nodes", () => {
 
   // Incorrect types should be caught statically, either due to being
   // totally invalid...
-  // $ExpectFlowError
   const _unused_badRepo: GN.RepoAddress = {
+    // $ExpectFlowError
     type: "REPOSITORY",
     owner: "foo",
     name: "bar",
--- a/src/util/map.test.js
+++ b/src/util/map.test.js
@@ -295,6 +295,7 @@ describe("util/map", () => {
       // $ExpectFlowError
       const _unused_badMap: Map<string, number> = MapUtil.merge([
         numberMap,
+        // $ExpectFlowError
         stringMap,
       ]);
     });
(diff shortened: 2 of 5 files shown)


src/plugins/github/nodes.test.js
```diff
@@ -64,8 +64,8 @@ describe("plugins/github/nodes", () => {
 
   // Incorrect types should be caught statically, either due to being
   // totally invalid...
-  // $ExpectFlowError
   const _unused_badRepo: GN.RepoAddress = {
+    // $ExpectFlowError
     type: "REPOSITORY",
```
Beanow, line 69: In this case we needed to move the error, as it happened on this specific field.

src/util/map.test.js
```diff
@@ -295,6 +295,7 @@ describe("util/map", () => {
       // $ExpectFlowError
       const _unused_badMap: Map<string, number> = MapUtil.merge([
         numberMap,
+        // $ExpectFlowError
         stringMap,
       ]);
```
Beanow, line 300: In this case we need two comments because this array is an invalid argument as all it's values types are expected to be the same. And the variable assignment is invalid as well, because it's type does not match the return type.

src/webutil/createRelativeHistory.test.js
```diff
@@ -46,10 +46,13 @@ describe("webutil/createRelativeHistory", () => {
...
-        expect(() => createHistory(undefined, "undefined/")).toThrow(
-          "basename: expected string, got: undefined"
-        );
+        expect(() =>
+          createHistory(
+            // $ExpectFlowError
+            undefined,
+            "undefined/"
+          )
+        ).toThrow("basename: expected string, got: undefined");
```
Beanow, line 55: We can't use only ExpectFlowError in this case anymore, because it depended on the suppression propagating to internal function calls too. The implicit function argument would be typed `undefined` and not match the expected `string`. Because that error is no longer suppressed, it should be explicitly typed.

PR #1788, opened 2020-05-08
Adopt a less restrictive TimestampMs type

decentralion:
This commit modifies the TimestampMs type so that it is no longer
opaque, which means that use of the `toNumber` function is no longer
enforced. This is because `toNumber` adds very little safety (I don't
believe it would have ever triggered, or caught any bugs, across the
history of the codebase), and needing to call a function every time we
make an example in test code is very tedious.

Thanks to this change, it becomes easy to roll out the TimestampMs type
across the codebase, which improve the legibility of our types. I found
replacement candidates by grepping for `timestampMs: number`, and
replaced all such usages.

Test plan: `yarn test` and `git grep "timestampMs: number"`. The only
hit is within the timestamp module itself.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -6,6 +6,7 @@ import deepFreeze from "deep-freeze";
 import {makeAddressModule, type AddressModule} from "./address";
 import {toCompat, fromCompat, type Compatible} from "../util/compat";
 import * as NullUtil from "../util/null";
+import {type TimestampMs} from "../util/timestamp";
 
 /**
  * This module contains the Graph, which is one of the most fundamental pieces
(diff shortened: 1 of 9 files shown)


decentralion:
Discussed offline: My claim is that enforcing usage of the `Timestamp.fromNumber` method fails cost benefit analysis because:
- the benefit is very low, true positive rate of these errors is close to zero
- the cost is moderately high, in particular updating test code to use this method consistently would be a pain

I am definitely not willing to do the work of updating all existing test code to use the TimestampMs type, because it feels like busywork. I posit no-one else wants to do that either, in which case we should just go with the less restrictive approach. The alternative, where we inconsistently do/don't use TimestampMs, seems like a worst-of-both-worlds because it adds more confusion and cognitive overhead.

I see your point that this maybe should be two PRs, although they are so small that I feel comfortable bundling them because they don't create much review overhead.

decentralion:
Decided to split this into two PRs per @Beanow's suggestion: #1789 and #1790

PR #1791, opened 2020-05-14
Update output format to include credOverTime

decentralion:
As requested by @s-ben, we map now include cred over time for all
contributions, not just contributors. Based on discussion with @Beanow,
we made it an optional field so that we can optionally filter to save
space instead.

I was initially concerned that we wouldn't be able to compute
credOverTime for non-user nodes in CredRank, which is why I left it out.
However, informed by discussions with @mZargham, I'm less concerned
because PageRank (and thus CredRank) is a linear operator on the seed
vector. So, if we want to compute the "cred over time" for individual
contributions in CredRank, we can do so by constructing time-specific
seed vectors (which flow only to activity minting cred in the specified
interval), and the sum of contributions time-scoped cred will be equal
to the non-time-scoped cred. It's good that we'll still have the epoch
nodes for users, as that will allow us to model sponsorship cred flow
dynamics.

cc @wchargin for CredRank considerations.

Test plan: Unit tests updated, `yarn test` passes.


src/analysis/output.js
```diff
@@ -33,6 +33,10 @@ export const COMPAT_INFO = {
 export type OutputNode = {|
   +address: $ReadOnlyArray<string>,
   +cred: number,
+  // Full cred over time (aligned with output interval boundaries).
+  // It's optional because it inflates the output size a lot -- we'll
+  // want to filter out low-cred nodes for large projects
+  +credOverTime: ?$ReadOnlyArray<number>,
```
Beanow, line 39: Nit: this isn't great at self-describing.
By that I mean, it will be tough to figure out from the output which filters were used. Some kind of filter details (a mode enum? set of whitelisted prefixes?) would be good to include. Since no filtering is actually implemented in this version, we could also make it required till we do support it.

decentralion: Ack. My intention is to let the filter be arbitrary logic, so there isn't a clean way to encode the result. I suppose we could standardize around a mode enum, but that adds a lot of complexity for the consumer to keep track of all the potential filter modes. 

As an example of a reasonably complex filter, we might have the composition of the following rules:
- keep every user node
- keep every node with greater than > 30 total cred
- keep all pull requests, issues, and topics from the last month

Would be a pain to track that via mode enums.

Beanow: Which makes sense but is a trade-off, because by necessity will require an out-of-band description of how this output was produced, should you want to have reproducibility of it. And in many cases you do want reproducibility, to support any claims you're deriving from it, or to expose it in a canonical instance.

PR #1792, opened 2020-05-15
Discord: minimal plugin declaration

Beanow:
Defines users, messages and the "sent message" edge.
This allows creating a minimal graph topology with some form of Cred flow.
Assuming we would be interested in minting Cred on messages.

Test plan: `test flow` and manual review

Issue #1682, opened 2020-02-29
Merge minted cred


decentralion:
We recently changed the Discourse plugin to only mint cred on likes, rather than on raw activity. (See [discussion](https://discourse.sourcecred.io/t/minting-discourse-cred-on-likes-not-posts/603).) The theory is simple: what makes a Discourse post valuable is not its existence per-se, but rather it is valuable if it contributed something meaningful in the eyes of other participants in the forum. Therefore, likes are a better proxy for value creation than raw activity.

This seems to have improved our Discourse quality-of-cred substantially, so I'd like to do the same for GitHub. One could imagine simply porting the same concept directly to GitHub (i.e. minting cred on reactions), but I don't think GH reactions work as a proxy for value creation; that's just not how developers use GH reactions, and many valuable PRs merge without any reactions at all.

However, to a first order approximation, the goal of collaborating on GitHub is to write code that merges into master, since such code is reviewed and then actually used by the project.

Therefore, I intend to modify the GitHub plugin in the following steps:
- For each pull request, get data from the API on which branch is its base ref
- If the pull merges, and the base ref is master, create a "merged-to-master" node
- Modify the default weights so that only merged-to-master nodes mint cred

decentralion:
Fixed by #1933 (simpler implementation).

PR #1793, opened 2020-05-15
Discord: minimal address functions

Beanow:
Depends on #1792 

Defines users, messages and the "sent message" addresses.
This allows creating a minimal graph topology with some form of Cred flow.
Assuming we would be interested in minting Cred on messages.

Test plan: `yarn flow` and manual review

### DIFF
--- a/src/plugins/discord/address.js
+++ b/src/plugins/discord/address.js
@@ -1,0 +1,56 @@
+// @flow
+
+import type {NodeAddressT, EdgeAddressT} from "../../core/graph";
+import {NodeAddress, EdgeAddress} from "../../core/graph";
+import {type Snowflake} from "./models";
+import {
+  userNodeType,
+  messageNodeType,
+  sentMessageEdgeType,
+} from "./declaration";
+
+/**
+ * These user IDs are considered globally unique by Discord.
+ * We can reference a user even across multiple guilds using this address.
+ * See https://discordapp.com/api/users/{userId}
+ *
+ * Importantly we don't include any membership details in the address.
+ * This way the graph topology stays identical when users leave / rejoin.
+ */
+export function userAddress(userId: Snowflake): NodeAddressT {
+  return NodeAddress.append(userNodeType.prefix, userId);
+}
+
+/**
+ * The URLs Discord creates to permalink these are:
+ * https://discordapp.com/channels/{guildId}/{channelId}/{messageId}
+ * Meanwhile the API retrieves these using:
+ * https://discordapp.com/api/channels/{channelId}/messages/{messageId}
+ *
+ * In other words the API finds just {channelId} sufficiently namespaced for unique messages.
+ * They're being conservative for the permalinks by namespacing as {guildId}/{channelId}.
+ * As they've clearly not felt the need to change the API, we'll use just {channelId} too.
+ */
+export function messageAddress(
+  channelId: Snowflake,
+  messageId: Snowflake
+): NodeAddressT {
+  return NodeAddress.append(messageNodeType.prefix, channelId, messageId);
+}
+
+/**
+ * Importantly we don't include any membership details in the address.
+ * This way the graph topology stays identical when users leave / rejoin.
+ */
+export function sentMessageAddress(
+  userId: Snowflake,
+  channelId: Snowflake,
+  messageId: Snowflake
+): EdgeAddressT {
+  return EdgeAddress.append(
+    sentMessageEdgeType.prefix,
+    userId,
+    channelId,
+    messageId
+  );
+}


src/plugins/discord/address.js
```diff
@@ -0,0 +1,56 @@
...
+ * These user IDs are considered globally unique by Discord.
+ * We can reference a user even across multiple guilds using this address.
+ * See https://discordapp.com/api/users/{userId}
+ *
+ * Importantly we don't include any membership details in the address.
+ * This way the graph topology stays identical when users leave / rejoin.
+ */
+export function userAddress(userId: Snowflake): NodeAddressT {
+  return NodeAddress.append(userNodeType.prefix, userId);
+}
```
decentralion, line 22: Assuming that a Discord alias looks like `discord/decentralion#1337`, it will be impossible to do address <-> alias mapping without an extra data store, which is sad.

Is it feasible to address users by the name and identifier? I know that these are mutable, which can be a pain, but the same is true for GitHub usernames (i have renamed my account several times). Since we treat the external data store as source of truth, any issues around user id renaming could be resolved by resetting the cache.

src/plugins/discord/address.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * The URLs Discord creates to permalink these are:
+ * https://discordapp.com/channels/{guildId}/{channelId}/{messageId}
+ * Meanwhile the API retrieves these using:
+ * https://discordapp.com/api/channels/{channelId}/messages/{messageId}
+ *
+ * In other words the API finds just {channelId} sufficiently namespaced for unique messages.
+ * They're being conservative for the permalinks by namespacing as {guildId}/{channelId}.
+ * As they've clearly not felt the need to change the API, we'll use just {channelId} too.
```
hammadj, line 32: One consideration for this might be the case where a SC instance has multiple Discord servers configured and wants to set different weights for each one, which would happen on the guildId level and therefore might warrant including it in the address structure

hammadj, line 32: One consideration would be SC instances that have multiple Discord servers in their configuration and want to set different weights for each one. For that to happen, we might need to include `guildId` in the Address structure.

src/plugins/discord/address.js
Beanow: In the prototype, for that reason the alias uses the user.id Snowflake.
`discord/143776454050709505`

For reference, Discord clients and bots are also expected to use Snowflakes for mentions and channel references if they want them to be connected in a message.
So clients do not send `Hi @Beanow, try posting that in #support`
but `Hi <@143776454050709505>, try posting that in <#454007860926611478>`.

This means that all translation between these is done with local maps and only the IDs are considered valid input on the server. We will have to rely on a cache or API to do this.

---

> Since we treat the external data store as source of truth, any issues around user id renaming could be resolved by resetting the cache.

I don't think this is true when combined with a Grain ledger. In terms of users an entire section of the graph might have a new node that receives it's Cred. So your previous identity will be overpaid and a brand new grain account will be underpaid and receiving that cred.

Having the same situation with GitHub might be more of a reason to change GitHubs node addresses than to do the same thing for Discord. :sweat_smile:

It's worth noting that GitHub also gives very explicit warnings that things are _likely to break_ when you rename things and requires you to confirm that you understand this. Plus most of the time you're talking about a more technical audience than discord users.
image

Worse yet, it seems that Discord offers the ability to control the discriminator (rather than have it randomized) as a premium feature. I don't know if they keep a blacklist of previously used pairs to avoid identity theft, but if they don't we could open ourselves up to trivial identity and Grain theft methods here.

src/plugins/discord/address.js
Beanow: Multiple servers can theoretically be supported even when we don't include the `guildId` here. This is because the `channelId` Snowflake is considered sufficiently unique to be in a global namespace.

But I think your point is, we can't express this in a single address prefix to include all messages in a particular server. So to do something like 2x on all messages in a particular server, we'd need to set this 2x on a more specific prefix, which is for each of it's channels.

A fair point. Though I think we shouldn't change our addresses to suit the weighting-by-prefix system. There's a couple of issues with the current weight configuration, so we'll likely want to have a more flexible weight configuration approach.

For example, if we want to set different weights in GitHub between open / closed / merged PRs (#1682) we'd be forced to include this (open/closed/merged) mutable piece of state into the node address to do this with prefix weights... which isn't the purpose of addresses, they're unique IDs.

Because we're also not supporting multiple servers yet, I think it's safe to punt on this :]

src/plugins/discord/address.js
decentralion: Thanks for the explanation @Beanow -- that makes a lot of sense.

decentralion:
Closing as stale.

PR #1796, opened 2020-05-15
Document graphToMarkovChain behavior on loops

decentralion:
This commit modifies the graphToMarkovChain test code so that we have
examples of how loops are processed. (Any loop will be treated like a
normal edge, and correspond to two connections: one with the loop's
forward weight, and one with the loop's backward weight.)

Test plan: `yarn test`

### DIFF
--- a/src/core/algorithm/graphToMarkovChain.test.js
+++ b/src/core/algorithm/graphToMarkovChain.test.js
@@ -122,6 +122,31 @@ describe("core/algorithm/graphToMarkovChain", () => {
         MapUtil.mapValues(map, (_, v) => sortBy(v, (x) => JSON.stringify(x)));
       expect(canonicalize(actual)).toEqual(canonicalize(expected));
     });
+    it("works on a chain with loops", () => {
+      const loop = edge("loop", n1, n1);
+      const e1 = edge("e1", n1, n2);
+
+      const g = new Graph().addNode(n1).addNode(n2).addEdge(loop).addEdge(e1);
+      const edgeWeight = () => ({forwards: 6.0, backwards: 3.0});
+      const actual = createConnections(g, edgeWeight, 1.0);
+      // Total out-weights (for normalization factors):
+      //   - for `n1`: 2 out, 1 in, 1 synthetic: 12 + 3 + 1 = 16
+      //   - for `n2`: 0 out, 1 in, 1 synthetic: 0 + 3 + 1 = 7
+      const expected = new Map()
+        .set(n1.address, [
+          {adjacency: {type: "SYNTHETIC_LOOP"}, weight: 1 / 16},
+          {adjacency: {type: "IN_EDGE", edge: loop}, weight: 6 / 16},
+          {adjacency: {type: "OUT_EDGE", edge: loop}, weight: 3 / 16},
+          {adjacency: {type: "OUT_EDGE", edge: e1}, weight: 3 / 4},
+        ])
+        .set(n2.address, [
+          {adjacency: {type: "SYNTHETIC_LOOP"}, weight: 1 / 4},
+          {adjacency: {type: "IN_EDGE", edge: e1}, weight: 6 / 16},
+        ]);
+      const canonicalize = (map) =>
+        MapUtil.mapValues(map, (_, v) => sortBy(v, (x) => JSON.stringify(x)));
+      expect(canonicalize(actual)).toEqual(canonicalize(expected));
+    });
   });
 
   describe("createOrderedSparseMarkovChain", () => {
@@ -197,6 +222,33 @@ describe("core/algorithm/graphToMarkovChain", () => {
           {
             neighbor: new Uint32Array([0, 1, 2]),
             weight: new Float64Array([0.5, 1.0, 1.0]),
+          },
+        ],
+      };
+      expect(normalize(osmc)).toEqual(normalize(expected));
+    });
+
+    it("works on a simple chain with a loop edge", () => {
+      // The loop edge will implicitly correspond to one forward edge with weight 2, and one backward edge with weight 1
+      const loop = edge("loop", n1, n1);
+      const e1 = edge("e1", n1, n2);
+
+      const g = new Graph().addNode(n1).addNode(n2).addEdge(e1).addEdge(loop);
+      const edgeWeight = () => ({forwards: 2, backwards: 1});
+      const osmc = createOrderedSparseMarkovChain(
+        createConnections(g, edgeWeight, 0.0)
+      );
+      const expected = {
+        nodeOrder: [n1.address, n2.address],
+        chain: [
+          {
+            neighbor: new Uint32Array([0, 1]),
+            // Reproduce floating point imprecision
+            weight: new Float64Array([2 / 5 + 1 / 5, 1]),
+          },
+          {
+            neighbor: new Uint32Array([0, 1]),
+            weight: new Float64Array([2 / 5, 0]),
           },
         ],
       };


src/core/algorithm/graphToMarkovChain.test.js
```diff
@@ -122,6 +122,31 @@ describe("core/algorithm/graphToMarkovChain", () => {
...
+        .set(n1.address, [
+          {adjacency: {type: "SYNTHETIC_LOOP"}, weight: 1 / 16},
+          {adjacency: {type: "IN_EDGE", edge: loop}, weight: 6 / 16},
+          {adjacency: {type: "OUT_EDGE", edge: loop}, weight: 3 / 16},
+          {adjacency: {type: "OUT_EDGE", edge: e1}, weight: 3 / 4},
+        ])
+        .set(n2.address, [
+          {adjacency: {type: "SYNTHETIC_LOOP"}, weight: 1 / 4},
+          {adjacency: {type: "IN_EDGE", edge: e1}, weight: 6 / 16},
+        ]);
```
Beanow, line 145: What was not obvious here, why the normalized weight groups are spread out across different map entries.
Highlighting as green and red pairs. 

```diff
      const expected = new Map()
        .set(n1.address, [
+         {adjacency: {type: "SYNTHETIC_LOOP"}, weight: 1 / 16},
+         {adjacency: {type: "IN_EDGE", edge: loop}, weight: 6 / 16},
+         {adjacency: {type: "OUT_EDGE", edge: loop}, weight: 3 / 16},
-         {adjacency: {type: "OUT_EDGE", edge: e1}, weight: 3 / 4},
        ])
        .set(n2.address, [
-         {adjacency: {type: "SYNTHETIC_LOOP"}, weight: 1 / 4},
+         {adjacency: {type: "IN_EDGE", edge: e1}, weight: 6 / 16},
        ]);
```

Reading through the implementation code (not the point of documentation :slightly_smiling_face:), finding this representation stores the weight as normalized on the other end of this edge.

So for a `n1 <-> n2` edge, in `n1`'s entry will be normalized to the `n2` total weights.

No idea why, but that appears to be intentional.

src/core/algorithm/graphToMarkovChain.test.js
```diff
@@ -122,6 +122,31 @@ describe("core/algorithm/graphToMarkovChain", () => {
...
+    it("works on a chain with loops", () => {
+      const loop = edge("loop", n1, n1);
+      const e1 = edge("e1", n1, n2);
+
+      const g = new Graph().addNode(n1).addNode(n2).addEdge(loop).addEdge(e1);
+      const edgeWeight = () => ({forwards: 6.0, backwards: 3.0});
+      const actual = createConnections(g, edgeWeight, 1.0);
+      // Total out-weights (for normalization factors):
+      //   - for `n1`: 2 out, 1 in, 1 synthetic: 12 + 3 + 1 = 16
+      //   - for `n2`: 0 out, 1 in, 1 synthetic: 0 + 3 + 1 = 7
```
Beanow, line 134: `0 + 3 + 1 = 7` :thinking:

src/core/algorithm/graphToMarkovChain.test.js
```diff
@@ -203,6 +228,33 @@ describe("core/algorithm/graphToMarkovChain", () => {
       expect(normalize(osmc)).toEqual(normalize(expected));
     });
 
+    it("works on a simple chain with a loop edge", () => {
+      // The loop edge will implicitly correspond to one forward edge with weight 2, and one backward edge with weight 1
```
Beanow, line 232: Why is this "implicit"? Doesn't this match the explicit weight resolver?
`const edgeWeight = () => ({forwards: 2, backwards: 1});`

src/core/algorithm/graphToMarkovChain.test.js
```diff
@@ -203,6 +228,33 @@ describe("core/algorithm/graphToMarkovChain", () => {
...
+        nodeOrder: [n1.address, n2.address],
+        chain: [
+          {
+            neighbor: new Uint32Array([0, 1]),
+            // Reproduce floating point imprecision
+            weight: new Float64Array([2 / 5 + 1 / 5, 1]),
+          },
+          {
+            neighbor: new Uint32Array([0, 1]),
+            weight: new Float64Array([2 / 5, 0]),
```
Beanow, line 251: These would be easier to read, if we had the same comment at the top about expected normalization.
```
      // Total out-weights (for normalization factors):
      //   - for `n1` ... = 5
      //   - for `n2` ... = 1?
```

src/core/algorithm/graphToMarkovChain.test.js
```diff
@@ -203,6 +228,33 @@ describe("core/algorithm/graphToMarkovChain", () => {
...
+      const edgeWeight = () => ({forwards: 2, backwards: 1});
+      const osmc = createOrderedSparseMarkovChain(
+        createConnections(g, edgeWeight, 0.0)
+      );
+      const expected = {
+        nodeOrder: [n1.address, n2.address],
+        chain: [
+          {
+            neighbor: new Uint32Array([0, 1]),
+            // Reproduce floating point imprecision
```
wchargin, line 246: We generally try to exactly representable numbers for tests like this,
so that you don’t have to be quite so careful here. Here, the
un-normalized out-flow from `n1` is `2 * f + b`, and the un-normalized
out-flow from `n2` is `b`. How about taking forward and backward weights
of `f = 3` and `b = 2` so that all the un-normalized out-weights are
powers of 2 and thus integer multiples of their reciprocals are exactly
representable? Thus:

```diff
diff --git a/src/core/algorithm/graphToMarkovChain.test.js b/src/core/algorithm/graphToMarkovChain.test.js
index f4f49d03..844e475b 100644
--- a/src/core/algorithm/graphToMarkovChain.test.js
+++ b/src/core/algorithm/graphToMarkovChain.test.js
@@ -234,7 +234,7 @@ describe("core/algorithm/graphToMarkovChain", () => {
       const e1 = edge("e1", n1, n2);
 
       const g = new Graph().addNode(n1).addNode(n2).addEdge(e1).addEdge(loop);
-      const edgeWeight = () => ({forwards: 2, backwards: 1});
+      const edgeWeight = () => ({forwards: 3, backwards: 2});
       const osmc = createOrderedSparseMarkovChain(
         createConnections(g, edgeWeight, 0.0)
       );
@@ -243,12 +243,11 @@ describe("core/algorithm/graphToMarkovChain", () => {
         chain: [
           {
             neighbor: new Uint32Array([0, 1]),
-            // Reproduce floating point imprecision
-            weight: new Float64Array([2 / 5 + 1 / 5, 1]),
+            weight: new Float64Array([5 / 8, 1]),
           },
           {
             neighbor: new Uint32Array([0, 1]),
-            weight: new Float64Array([2 / 5, 0]),
+            weight: new Float64Array([3 / 8, 0]),
           },
         ],
       };
```

src/core/algorithm/graphToMarkovChain.test.js
wchargin: +1

src/core/algorithm/graphToMarkovChain.test.js
wchargin: It’s certainly intentional. [The documentation][1] says, 

> This `weight` is a conditional probability: given that you're at the
> source of this connection's adjacency, what's the probability that you
> travel along this connection to the target?

That is, these are `Pr[X_{n+1} = dst | X_n = src]`, so the weights must
be normalized per source to form a proper probability distribution. When
you say “a `n1 <-> n2` edge”, remember that this really represents two
edges, `e1: n1 -> n2` and `e1': n2 -> n1`. So there are two adjacencies:
the adjacency for `e1` is stored in the map entry for `n2`, and is
normalized by the total out-weight of its source, `n1`; and the
adjacency for `e1'` is stored in the map entry for `n1`, and is
normalized by the total out-weight of its source, `n2`.

[1]: https://github.com/sourcecred/sourcecred/blob/ae181c2fda8719546d90acbaa0ab60d417473a2c/src/core/algorithm/graphToMarkovChain.js#L16-L18

PR #1797, opened 2020-05-15
Refactor timelinePagerank, exposing connections

decentralion:
This commit refactors internal helper methods in timelinePagerank so
that rather than piping around an OrderedSparseMarkovChain, we instead
provide the NodeToConnections from which that OSMC may be derived. This
is important because the NodeToConnections has the information necessary
to derive how score flowed across individual edges, and not just on the
adjacency topology of the graph. This will allow us to compute the
OutputEdge format with edge-specific cred flows as documented in #1773.

Test plan: `yarn test` passes. It's a simple refactor.

### DIFF
--- a/src/core/algorithm/timelinePagerank.js
+++ b/src/core/algorithm/timelinePagerank.js
@@ -20,12 +20,9 @@ import {type Distribution} from "./distribution";
 import {
   createOrderedSparseMarkovChain,
   createConnections,
+  type NodeToConnections,
 } from "./graphToMarkovChain";
-import {
-  findStationaryDistribution,
-  type PagerankParams,
-  type SparseMarkovChain,
-} from "./markovChain";
+import {findStationaryDistribution, type PagerankParams} from "./markovChain";
 
 /**
  * Represents raw PageRank distributions on a graph over time.
@@ -41,6 +38,8 @@ export type TimelineDistributions = $ReadOnlyArray<{|
   // Uses the canonical graph node order.
   +distribution: Distribution,
 |}>;
+
+export const SYNTHETIC_LOOP_WEIGHT = 1e-3;
 
 /**
  * Runs timeline PageRank on a graph.
@@ -127,7 +126,7 @@ export async function timelinePagerank(
     nodeEvaluator,
     intervalDecay
   );
-  const markovChainIterator = _timelineMarkovChain(
+  const nodeToConnectionsIterator = _timelineNodeToConnections(
     weightedGraph.graph,
     edgeCreationHistory,
     edgeEvaluator,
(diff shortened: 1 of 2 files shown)


decentralion:
Note to reviewers: I'm happy to take a "rubber stamp" approval here. I'm the only person with context on this code (wrote it during a period where I was the only person working on SC), and we're planning to replace this all in a few months with CredRank, so there's not a ton of value in ramping up on it.

Beanow:
> Note to reviewers: I'm happy to take a "rubber stamp" approval here. I'm the only person with context on this code (wrote it during a period where I was the only person working on SC), and we're planning to replace this all in a few months with CredRank, so there's not a ton of value in ramping up on it.

Same response as https://github.com/sourcecred/sourcecred/pull/1799#issuecomment-630268033

src/core/algorithm/timelinePagerank.js
```diff
@@ -166,12 +163,12 @@ export function* _timelineNodeWeights(
...
 }
 
-export function* _timelineMarkovChain(
+export function* _timelineNodeToConnections(
   graph: Graph,
   edgeCreationHistory: $ReadOnlyArray<$ReadOnlyArray<Edge>>,
   edgeEvaluator: EdgeWeightEvaluator,
   intervalDecay: number
-): Iterator<SparseMarkovChain> {
+): Iterator<NodeToConnections> {
```
Beanow, line 171: Nit: Suggest using `Generator<NodeToConnections, void, void>` as a return type.
As (Async)Generators implement both `Iterator` and `Iterable`, it allows for more flexible usage.

Function arguments should remain specific though. For example when using the argument in a for..of loop, it should be an `Iterable<T>`, even when passing in a Generator. This makes it easier to create a source of data. For example in a unit test, array literals like `[1, 2, 3]` can be used as they implement `Iterable<Number>` but aren't Generators.

src/core/algorithm/timelinePagerank.js
```diff
@@ -187,19 +184,7 @@ export function* _timelineMarkovChain(
...
-    // performance improvement. We will need to change the markov chain
-    // representation to do so (we should add a `totalOutWeight` array to the
-    // chain, so that we can efficiently update the total weight as we add new
-    // connections, rather than needing to re-normalize the whole chain for
-    // each interval).
-    const chain = createOrderedSparseMarkovChain(
-      createConnections(graph, currentEdgeWeight, 1e-3)
-    ).chain;
-    yield chain;
+    yield createConnections(graph, currentEdgeWeight, 1e-3);
```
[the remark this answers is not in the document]
decentralion: Sure, done.

src/core/algorithm/timelinePagerank.js
decentralion: I don't see any practical benefit to doing this--can you give an example of a plausible situation where we'd be glad for this change? In general, I think adding complexity for purely hypothetical benefits is an antipattern.

PR #1799, opened 2020-05-16
Factor out _intervalResult in timelinePagerank

decentralion:
As part of work for #1773, I want to add a lot more complexity to the
logic for computing individual time-slices of pagerank scores, so that
we can trace how much score flowed on each individual edge. This means
adding more complexity to the _computeTimelineDistribution function;
however, that function is an untested wrapper, so I'm hesitant to add
more complexity directly.

Instead, I'm first factoring out an _intervalResult method and a
corresponding type, which computes the scores for a given timeline
interval. I've also added sanity checking tests for this method. In a
followon commit, I'll add more logic for tracking edge-level score
flows.

Test plan: This is just a refactor, maintaining existing behavior and
adding tests. `yarn test --full` passes except for the issue underlying #1800. 
Since our sharness tests
include doing a full load (including timeline cred computation) on
realistic data from GitHub, this gives us confidence that there hasn't
been any change to cred semantics.

### DIFF
--- a/src/core/algorithm/timelinePagerank.js
+++ b/src/core/algorithm/timelinePagerank.js
@@ -24,10 +24,7 @@ import {
 } from "./graphToMarkovChain";
 import {findStationaryDistribution, type PagerankParams} from "./markovChain";
 
-/**
- * Represents raw PageRank distributions on a graph over time.
- */
-export type TimelineDistributions = $ReadOnlyArray<{|
+export type IntervalResult = {|
   // The interval for this slice
   +interval: Interval,
   // The total node weight within this interval (normalized to account for the
@@ -37,7 +34,11 @@ export type TimelineDistributions = $ReadOnlyArray<{|
   // The raw score distribution over nodes for this interval (i.e. sums to 1).
   // Uses the canonical graph node order.
   +distribution: Distribution,
-|}>;
+|};
+/**
+ * Represents raw PageRank distributions on a graph over time.
+ */
+export type TimelineDistributions = $ReadOnlyArray<IntervalResult>;
 
 export const SYNTHETIC_LOOP_WEIGHT = 1e-3;
 
@@ -206,28 +207,47 @@ export async function _computeTimelineDistribution(
     const nodeToConnections = NullUtil.get(
       nodeToConnectionsIterator.next().value
     );
-    const {chain} = createOrderedSparseMarkovChain(nodeToConnections);
-
-    const seed = weightedDistribution(nodeOrder, nodeWeights);
-    if (pi0 == null) {
-      pi0 = seed;
-    }
-    const params: PagerankParams = {chain, alpha, seed, pi0};
-    const distributionResult = await findStationaryDistribution(params, {
-      verbose: false,
-      convergenceThreshold: 1e-7,
-      maxIterations: 255,
-      yieldAfterMs: 30,
-    });
-    const intervalWeight = sum(nodeWeights.values());
-    results.push({
+    const result = await _intervalResult(
+      nodeWeights,
+      nodeToConnections,
+      nodeOrder,
       interval,
-      intervalWeight,
-      distribution: distributionResult.pi,
-    });
+      pi0,
+      alpha
+    );
+    results.push(result);
     // Use the latest convergce results as the starting point for the next run
     // of PageRank
-    pi0 = distributionResult.pi;
+    pi0 = result.distribution;
   }
   return results;
+}
+
+export async function _intervalResult(
+  nodeWeights: Map<NodeAddressT, number>,
+  nodeToConnections: NodeToConnections,
+  nodeOrder: $ReadOnlyArray<NodeAddressT>,
+  interval: Interval,
+  pi0: Distribution | null,
+  alpha: number
+): Promise<IntervalResult> {
+  const {chain} = createOrderedSparseMarkovChain(nodeToConnections);
+
+  const seed = weightedDistribution(nodeOrder, nodeWeights);
+  if (pi0 == null) {
+    pi0 = seed;
+  }
+  const params: PagerankParams = {chain, alpha, seed, pi0};
+  const distributionResult = await findStationaryDistribution(params, {
+    verbose: false,
+    convergenceThreshold: 1e-7,
+    maxIterations: 255,
+    yieldAfterMs: 30,
+  });
+  const intervalWeight = sum(nodeWeights.values());
+  return {
+    interval,
+    intervalWeight,
+    distribution: distributionResult.pi,
+  };
 }
--- a/src/core/algorithm/timelinePagerank.test.js
+++ b/src/core/algorithm/timelinePagerank.test.js
@@ -2,12 +2,13 @@
 
 import {sum} from "d3-array";
 import * as NullUtil from "../../util/null";
-import {node, edge} from "../graphTestUtil";
+import {node, edge, advancedGraph} from "../graphTestUtil";
 import {Graph, type EdgeAddressT, type Edge} from "../graph";
 import {
   _timelineNodeWeights,
   _timelineNodeToConnections,
   SYNTHETIC_LOOP_WEIGHT,
+  _intervalResult,
 } from "./timelinePagerank";
 import {
   createConnections,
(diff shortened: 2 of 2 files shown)


decentralion:
Note to reviewers: I'm happy to take a "rubber stamp" approval here. I'm the only person with context on this code (wrote it during a period where I was the only person working on SC), and we're planning to replace this all in a few months with CredRank, so there's not a ton of value in ramping up on it.

Beanow:
> Note to reviewers: I'm happy to take a "rubber stamp" approval here. I'm the only person with context on this code (wrote it during a period where I was the only person working on SC), and we're planning to replace this all in a few months with CredRank, so there's not a ton of value in ramping up on it.

Before actually reviewing, might I wholeheartedly reject this notion? :sweat_smile:
I would even strongly prefer to abandon this feature over going this route.

One of the reasons to do code reviews is _exactly_ to counteract this. Forcing there to be at least one other person on the planet who knows that the code does, and forcing the code to be explained at least once (which if you're going to do, might as well archive that as documentation).

I'd say this is a good opportunity to go from undocumented code without test coverage, to something better understood.

If this is too much friction, then we have a problem with our development workflow.
Which seems to be a trend: https://github.com/sourcecred/sourcecred/pull/1801#pullrequestreview-413626572

Besides, CredRank is out of scope for beta.1. Which is supposed to be a polished release we're comfortable supporting gets adopted in the wild. Adding on last minute "rubber stamped" core algorithm changes seems dangerous and contradictory.

decentralion:
I don't know if you understand why I want the rubber-stamp here @Beanow. 
This isn't work on CredRank. This is updates (in this case, minor refactors) to TimelineCred, so that we can compute edge flow information for #1773 (which allows us to debug the initiative plugin cred flows and thus unblock its usage, and enables a new UI as part of beta.1).

The TimelineCred code exists in a state of "review debt" because no one has ever reviewed it (because I was the only person working on SC at the time it was written). If we were planning to maintain it long term, then I'd say we should pay that review debt. However, we're actually planning to throw it away within ~one milestone of launching Beta.1, when this code gets replaced by CredRank. Therefore, I don't think there's a reason to pay down the cred debt.

We can be confident that these "last minute core algorithm changes" won't change the underlying cred scores, because we have snapshot tests for real cred outputs on example github. So there isn't a risk that we regress the underlying cred quality. However, these changes will be used for generating new inspection tools over cred, which will make it much easier to detect if there are underlying issues that we need to fix. 

That said, if you want to invest the time in reviewing these changes, I'm fine with this too. However, for reasons above, I don't think it's super high value.

decentralion:
Also:

> Before actually reviewing, might I wholeheartedly reject this notion? 😅
> I would even strongly prefer to abandon this feature over going this route.

I'd like to wholeheartedly reject **that** notion 😅

We have a code review practice to improve our effectiveness as software developers, not because code review is an end-in-itself. In this case (low-risk changes to legacy code that is soon-to-be-deprecated), the value of code review is relatively low. However, the value of this feature work is very high (unblocks actually merging initiatives, unblocks better UI, unblocks @s-ben in his data analysis workflows).

Abandoning a high-priority feature in order to ensure code review of deprecated legacy code is a priority inversion.

Beanow:
Sorry I don't mean to attack you personally. The rationale struck a nerve with me, and I used needlessly strong language. Also I didn't mean any value judgement for this PR, as I haven't reviewed. I'm only responding to the rubber stamping.

What I mean is that in general terms, rubber stamp approving code because something is high priority and urgent and in an unreviewable state, is a red flag for code quality and an example of "heroism".

The reason it's high priority and urgent, is because it's already predicted that this will be a critical component to delivering high value to users and likely to have many dependencies. Out of any code, this is the code that should be _most carefully_ reviewed. Instead of skipped through the very practices that are designed to make those (known critical) components appropriately solid.

Additionally I would claim that having code quality practices in place for everything _except_ the critical things is not going to have much benefit overall, and is a common cause of vicious cycles of heroism. Teams that have repeated overtime, putting out fires in the live environment, and high burnout rate, are likely to have this "fastlane" in their routine workflow. It's self-perpetuating because this pressure for heroism doesn't happen for low priority features, meaning the components who's code quality erode the most from it are the most critical ones.

As for why this is heroism: Under normal circumstances adding a feature takes longer, but in this case you're expected to ship faster than reasonable. Normally, devs are allowed mistakes, the reviewer is there to help catch those. But with urgency you're now expected to be writing bug-free code in a single attempt. Normally, review and approval means responsibility for shipped code is shared and at least 2 people could respond to issues should there be any. Now both of that falls onto one person.

decentralion:
> Sorry I don't mean to attack you personally. The rationale struck a nerve with me, and I used needlessly strong language. Also I didn't mean any value judgement for this PR, as I haven't reviewed. I'm only responding to the rubber stamping.

Thanks for that note. Having read your recent message, I think that at root we're miscommunicating via the term "rubber stamp"ing. You're reading rubber stamp as meaning: this code is so urgent and such a mess that there's no way to review it! Let's just ship it and not bother and hope nothing bad happens!

What I'm saying is: this code is perfectly reviewable (it's actually a super isolated refactor commit). However, I don't think code review is really worth it from a cost-benefit standpoint:

- It's a simple atomic refactor with test coverage, so the likelihood that it has any serious error is close to 0
- It's in a part of the codebase that no-one else has context on, so it's relatively more expensive to reivew
- It's in a part of the codebase which is deprecated, so ramping new people up on it is less useful
- Also, because it's due for deprecating, spotting potential improvements is also less valuable

Basically, I view "rubber stamping" less as an urgent/heroic mechanism, and more as a response to unusual situations where code review is not very valuable from a cost-benefit standpoint.

If it's contentious, I'm fine going through regular review. As I said, I don't think it should be especially difficult to review either this or the subsequent PR.

Beanow:
Well I think this was the straw that broke the camels back more than as grand of a problem in isolation that got me up in arms. But I'll try unpack my thoughts better.

We haven't used rubber stamping before, so I took it as one of the dictionary entries of: giving ones approval as a matter of routine. Which leaves the reason for this routine blank.

I inferred the reason for the routine from your other comments: only person with context, will replace soon, no value in ramping up. Later on: review debt, stressing the importance of what this is a dependency for.

I think that wasn't far off from how you're explaining, cost-benefit is not favorable for this review. So the motivation for giving routine approval would be: "not worth the time".

There is still a miscommunication I think, because I was referring to it as your request to rubber stamp, not the term itself.

---

Urgency is an assumption on my part. I feel like it's been an ongoing trend to dilute reviews under stress to move faster. To me, this hunch feels supported by why we would need to make a trade-off / cost-benefit analysis in the first place. Still an assumption though.

Whether true or not, I think diluting reviews would be counter productive so I've been prioritizing it, and as a lead by example attempt have put much of my time into being thorough with my reviews. While I have no problems putting that work in, what's been straining about that is feeling like I've been the designated bad news bringer, with more critical reviews and blocking. The request felt in some ways undermining of this work and I think was part of why that struck a nerve.

Another part is that this came across as a request. Stronger than a heads-up for reviewers. It felt to me like you were stating your desired outcome.

In discussing review culture, I came away from that thinking that the responsibility of making PRs easy to review lies with the author. Hence PRs should be small, well explained, tested, etc. If not, a reviewer is expected to push back and ask the author to try and improve that. Rationalizing a high review cost as the author is odd in that light, but no harm in making an argument in good faith. Ultimately though it's the reviewer who needs to make the cost-benefit call how they'll go about their review. And importantly they are accepting some accountability for what they're approving. Impeding on that would be offloading, so I'm feeling protective of this space for reviewers.

Notably though, the typical response would be to ask the author to take another stab if review costs are high. Turning a bind eye because it's on it's way out, in my opinion should be done as an exception with the consent of the reviewer, not routinely nor mainly on advice by the author (using rubber stamping here did suggest routine to me).

I'm seeing a danger of a routine approval due to "we're throwing it away soon anyway" cementing, and am much concerned about that. I believe it has already permeated to the point where it affects the quality of the software and value end-users get out of it and is slowing down development, through a cluttered desk effect of technical debt and unfinished business across the code, even ignored "de-prioritized" bugs. The way I see it, wanting a more polished beta.1 is a direct response to such symptoms. And here I'm concerned we're about to ignore one of the causes.

Actually while I didn't frame beta.1's goals, I will take some credit for getting the ball rolling on it, as that was in part me taking up the task of investigating issues with the development workflow and petitioning for team objectives and an updated roadmap as part of that.

So these discussions about defending accruing of various debts, do feel like pulling out the rug from what I'm trying to achieve with some of the other work I put forward on that. In fairness and retrospect I've probably dug my own grave on this. Taking an engineering approach of brooding on the problem and focusing on getting solutions implemented, rather than making rounds to build report and aligning in a manager style.

decentralion:
Thanks for the detailed reply @Beanow. I hear that you're feeling like the rug is being pulled under you re: review quality and review culture. And I'll admit that I (personally) like to be able to move fast when coding, and tend to be of a "hack it and ship it" mentality (I recognize this is something of a contradiction with the review culture that I do want us to promulgate). This has definitely caused frustration for other developers working with me in the past 😅 ; in SourceCred @wchargin was often the voice of caution that would balance me out. I guess that falls to you now. :)

I hear that you're feeling frustrated with being the "bearer of bad news" and feeling like I want to  route around your reviews. With the grain ledger, your reviews have slowed stuff down but also been really valuable -- we're laying foundation for code that we're going to have a deep dependency on for years to come, and you've had some really good insight into how we can make that system better designed. So going slower, but on better foundation, is of great benefit to the project.

That said, you're right that I don't really want review on these timeline cred changes. All of the timeline cred code is sloppy prototype code that I pushed out during a monthlong hackathon with no review. It's not the cleanest but I understand it and know how to change it without introducing regressions. And we *are* throwing it away soon. So as a practical matter, what's the upside of having in depth review and fixing nits and so forth? It just slows us down on beta.1 and improves the understanding and code quality of a system we're about to throw away.

On a more general sense, I think I'm feeling frustrated because it feels like our dev. velocity has slowed to a crawl. And maybe me pushing back against code review on this workstream is an attempt by me to get some momentum back. I think you and I should have a chat about this and make sure we understand root causes rather than just fighting over symptoms and bandaids though. Let's schedule a recurring dev meeting (could just be you and me to start, potentially open to everyone developing on SC) so we can get aligned.

src/core/algorithm/timelinePagerank.js
```diff
@@ -204,28 +205,47 @@ export async function _computeTimelineDistribution(
...
+  pi0: Distribution | null,
+  alpha: number
+): Promise<IntervalResult> {
+  const {chain} = createOrderedSparseMarkovChain(nodeToConnections);
+
+  const seed = weightedDistribution(nodeOrder, nodeWeights);
+  if (pi0 == null) {
+    pi0 = seed;
+  }
+  const params: PagerankParams = {chain, alpha, seed, pi0};
```
Beanow, line 238: Nit: prefer not to reassign function arguments. (wish this had an eslint rule, couldn't find one though)
Instead, can use something like:

```js
const params: PagerankParams = {
  chain,
  alpha,
  seed,
  pi0: pi0 || seed
};
```

src/core/algorithm/timelinePagerank.test.js
```diff
@@ -102,4 +103,67 @@ describe("src/core/algorithm/timelinePagerank", () => {
...
+        return result.distribution[idx];
+      }
+      const isoScore = getScore(nodes.isolated);
+      const srcScore = getScore(nodes.src);
+      const dstScore = getScore(nodes.dst);
+      expect(isoScore + srcScore + dstScore).toBeCloseTo(1);
+      // It has 2/3rd weight, and is isolated, so it's simple
+      expect(isoScore).toBeCloseTo(2 / 3);
+      // src has the weight, and dst doesnt, so it should have a higher score
+      expect(srcScore).toBeGreaterThan(dstScore);
```
Beanow, line 166: Because the graph structure is abstracted away between `graphTestUtil.js` and `example()`, validating these assertions takes a fair amount of context switching. Any thoughts how we can make this test more straightforward to read on it's own?

src/core/algorithm/timelinePagerank.test.js
```diff
@@ -102,4 +103,67 @@ describe("src/core/algorithm/timelinePagerank", () => {
...
+
+  describe("_intervalResult", () => {
+    async function example() {
+      const {graph1, nodes} = advancedGraph();
+      const g = graph1();
+      const nodeWeights = new Map()
+        .set(nodes.src.address, 1)
+        .set(nodes.isolated.address, 2);
+      const edgeFn = (_unused_edge) => ({forwards: 1, backwards: 0.5});
+      const nodeToConnections = createConnections(g, edgeFn, 1e-3);
```
Beanow, line 115: Nit: prefer `1e-3` magic number is named as a variable.

src/core/algorithm/timelinePagerank.test.js
decentralion: It'd be cleaner to create the graph locally rather than re-using it from graphTestUtil... but since this is all going to be replaced with CredRank, just having the tests as they are is fine IMO.

decentralion:
Merging on self review (TimelineCred is the "dirty" zone, though also #1773).

PR #1800, opened 2020-05-16
Update example-github snapshot

decentralion:
Without this change, `yarn test --full` fails.


decentralion:
Note to reviewers: I'm not sure what actually caused this change in the upstream data.

Beanow:
On the surface it would seem like GitHub changed the scheme to generate the ID?
What I'd like to know is, if this requires people to blow their cache to reflect ID changes or not.

PR #1801, opened 2020-05-17
Add params field to Project type

hammadj:
This will allow SourceCred instances to override the default alpha and interval decay parameters by adding an optional "params" field in their project.json file. If they don't include any of the fields, it will just fallback to use the default params.

Test Plan: Add params field in project.json file and see if the resulting SourceCred instance correctly picks up the custom alpha value.

CI Failing until #1800 is merged


decentralion:
I support merging this more-or-less as is. It adds technical debt, yes, but project.json is already a known storehouse of technical debt, which we're planning to fixup when we set up the instance system. Adding one more field to project.json doesn't meaningfully change the trajectory there. 

Also, it sounds like MetaGame feels they need this for the quality of their cred instance, and since they are a partner project, greenlighting this falls in the category of "prioritized responsiveness to help partners with their cred issues". Asking to block this until we design the instance system seems unfair, since we already have a number of high-priority projects underway, so we're not likely to make very fast progress on the instance system.

Beanow:
> but project.json is already a known storehouse of technical debt, which we're planning to fixup when we set up the instance system.

As I'm listed for implementing said instance system, that means the buck of this technical debt will be passed to me in a span of weeks.

> Adding one more field to project.json doesn't meaningfully change the trajectory there.

It also doesn't improve on the current situation. While creating the beginnings of a config system would.

> Asking to block this until we design the instance system seems unfair

I did not mean the instance system. I meant a configuration format. And by configuration format, we need a scaffolding similar to https://github.com/sourcecred/docs/ that will hold our broad-strokes categories while it's still in experimental stage.

By implementing an experimental feature (tweaking alpha) in an experimental general-purpose configuration format, which we know we're are going to need, there's 2 things we can improve on at once. I didn't suggest to block this on the instance system, I'm suggesting this wouldn't be accruing as much technical debt if it were in a first stab at a configuration system.

PR #1803, opened 2020-05-19
Discord: add latest message helper

BrianLitwin:
This adds a helper function to get the nth message
from last, for purposes of re-fetching recent messages
from the Discord server to catch edits and new emoji
reactions. We use this helper function in FetchDiscord
to refetch the previous n messages in a channel.

Test Plan: unit tests added.


src/plugins/discord/sqliteMirror.js
```diff
@@ -497,4 +497,17 @@ export class SqliteMirror {
...
       .map((e) => Model.refToEmoji(e.emoji));
   }
+
+  latestMessageId(channel: Model.Snowflake) {
+    return this._db
+      .prepare(
+        dedent`
+          SELECT id, max(timestamp_ms)
+          FROM messages
+          WHERE channel_id = :channel_id
```
Beanow, line 507: Sorting by `timestamp_ms` is technically imperfect. Because the real order is determined by sorting Snowflakes treated as `uint64`. The only time these are different is when the timestamp is identical for several entries, and the Snowflakes' additional data need to act as a tiebreaker.
https://discord.com/developers/docs/reference#snowflakes

More importantly, in the prototype what I ended up using was not a query for the last message but:
https://github.com/sourcecred/sourcecred/blob/d0dc11d2016dcb359606ee736a89e20235a7aaaf/src/plugins/discord/mirrorRepository.js#L208

The reason being, we'll want to "reload" some amount of messages we already cached. Because they could be edited (for example adding new mentions, changing the graph) and the emoji reactions on them may have changed since last mirror. The number of messages is a heuristic, trading off between faster mirroring and higher accuracy.

We don't care too much that the sorting is imperfect there, because the number `n` was arbitrary anyway. If we reload one message more or less, it wouldn't be a big issue. So I went with `timestamp_ms` sorting there.

---

Concluding, it would be nice-to-have correct sorting. But more importantly, want to check if `latestMessageId` is what's needed for our caching strategy or whether we'll be using `nthMessageFromTail` like queries instead?

src/plugins/discord/sqliteMirror.test.js
```diff
@@ -256,6 +258,25 @@ describe("plugins/discord/sqliteMirror", () => {
       });
     });
 
+    it("retrieves latest message's id", () => {
+      const channelId: Snowflake = "1";
+      const authorId: Snowflake = "2";
+      const mes1: Message = testMessage("1", channelId, authorId, 1);
+      const mes2: Message = testMessage("2", channelId, authorId, 1000);
+      const mes3: Message = testMessage("3", channelId, authorId, 5);
```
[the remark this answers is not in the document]
BrianLitwin: I see that that the real life data wouldn't look like the test data but I don't think it's contrived; it's verifying that the sorting mechanism is the timestamp and not the id or the insertion order. 

I agree we should handle a tie on timestamp; although, it wouldn't cause a problem to retrieve the second to last message for the purpose of the cache; we'd re-fetch one extra message.

src/plugins/discord/sqliteMirror.js
BrianLitwin: It feels like `nthMessageFromTail` will be more useful; I'm wondering if `n` is the appropriate parameter or if we should use something like time (say, all messages in the last five days) to re-fetch? 

E.g., 50 messages can take a month or a few minutes to accumulate in Discord. However, once messages are buried behind others, it's much less likely they'll be updated. 

What do you think?

src/plugins/discord/sqliteMirror.test.js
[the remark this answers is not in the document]
Beanow: Yeah good point. I did mean whether the input data shows a likely scenario. It does test with certainty that we're sorting by the field we're expecting it to sort by.

And definitely agree for caching purpose we can get away with imperfection in the sorting.
When we've settled on the implementation, let's include a note on this in comments. As it's an easy assumption to make that sorting will just work correctly and doesn't have this edge case.

src/plugins/discord/sqliteMirror.js
Beanow: Good idea, sounds like that would work better than Nth.

Say `n=30` I can definitely imagine a lot of unnecessary reloading of something like a locked down `#announcements` channel where messages trickle in every other week. It could mean reloading the whole channel. Versus possibly not enough reloading for high volumes of messages in something like `#support` or `#general` where 30 messages could be 5 minutes worth of conversation.

This burying of messages you're referring to is also the underlying assumption why Nth message would still gave acceptable accuracy. I think that's also true for a time offset :]

Depending on what acceptable accuracy means, you could tweak the parameter or discard the cache whenever you like. For example, using cache for updates of stats several times a day but reloading everything just before the weekly Grain distribution.

src/plugins/discord/sqliteMirror.test.js
```diff
@@ -256,6 +258,37 @@ describe("plugins/discord/sqliteMirror", () => {
...
+        return {sqliteMirror, channelId};
+      }
+      it("retrieves nth message from tail", () => {
+        const {sqliteMirror, channelId} = buildMessages();
+        expect(sqliteMirror.nthMessageIdFromTail(channelId, 3)).toBe("7");
+      });
+      it("handles n greater than total message count", () => {
+        const {sqliteMirror, channelId} = buildMessages();
+        expect(sqliteMirror.nthMessageIdFromTail(channelId, 20)).toBe(
+          undefined
```
BrianLitwin, line 287: I kind of think the caller is in a better position to handle the case where n > total message count, rather than `sqliteMirror`

src/plugins/discord/sqliteMirror.js
Beanow: Happy to accept either approach :]
Whether it holds up is we'll probably only find out by testing it in the wild.

src/plugins/discord/sqliteMirror.js
```diff
@@ -497,4 +497,20 @@ export class SqliteMirror {
       .all({channel_id: channel, message_id: message})
       .map((e) => Model.refToEmoji(e.emoji));
   }
+
+  nthMessageIdFromTail(channel: Model.Snowflake, n: number) {
```
Beanow, line 501: Should define a return type. `?Model.Snowflake`

src/plugins/discord/sqliteMirror.test.js
Beanow: I agree, for the caching strategy we don't need any extra information than the return type being nullable. If `n` > total messages, it simply means we should (re)load all messages in that channel :]

src/plugins/discord/sqliteMirror.test.js
```diff
@@ -256,6 +258,37 @@ describe("plugins/discord/sqliteMirror", () => {
...
+        const authorId: Snowflake = "2";
+
+        const db = new Database(":memory:");
+        const sqliteMirror = new SqliteMirror(db, "0");
+        sqliteMirror.addUser(testUser(authorId));
+        sqliteMirror.addChannel(testChannel(channelId));
+
+        const maxCount = 10;
+        var x = 1;
+        while (x <= maxCount) {
```
Beanow, line 273: This would be cleaner as a for-loop I think :]
```js
for(let x = 1; x <= 10; x++)
```

But when you do use mutable variables, prefer using `let` over `var`.

BrianLitwin: updated

src/plugins/discord/sqliteMirror.js
BrianLitwin: updated

src/plugins/discord/sqliteMirror.test.js
```diff
@@ -227,6 +227,43 @@ describe("plugins/discord/sqliteMirror", () => {
...
+          };
+          sqliteMirror.addMessage(message);
+        }
+        return {sqliteMirror, channelId};
+      }
+      it("retrieves nth message from tail", () => {
+        const {sqliteMirror, channelId} = buildMessages();
+        expect(sqliteMirror.nthMessageIdFromTail(channelId, 3)).toBe("7");
+      });
+      it("handles n greater than total message count", () => {
```
decentralion, line 259: nit: rephrase to `"returns undefined when n is greater than message count"`

src/plugins/discord/sqliteMirror.js
```diff
@@ -497,4 +497,20 @@ export class SqliteMirror {
       .all({channel_id: channel, message_id: message})
       .map((e) => Model.refToEmoji(e.emoji));
   }
+
+  nthMessageIdFromTail(channel: Model.Snowflake, n: number): ?Model.Snowflake {
```
decentralion, line 501: nit: might be nice to have a brief comment or docstring explaining that it returns undefined when `n` is greater than the number of messages -- kind of implied by the type signature but not 100% clear

BrianLitwin: Updated

src/plugins/discord/sqliteMirror.test.js
BrianLitwin: Updated

src/plugins/discord/sqliteMirror.test.js
```diff
@@ -222,11 +223,40 @@ describe("plugins/discord/sqliteMirror", () => {
...
 
+    describe("nthMessageIdFromTail", () => {
+      const channelId: Snowflake = "1";
+      const authorId: Snowflake = "2";
+
+      it("retrieves nth message from tail", () => {
+        const db = new Database(":memory:");
+        const sqliteMirror = new SqliteMirror(db, "0");
+        buildNMessages(5, sqliteMirror, channelId, authorId);
+        expect(sqliteMirror.messages(channelId).map((x) => x.id)).toEqual([
```
BrianLitwin, line 239: Added this sanity check bc it's cheap and makes it easier to reason about the test

src/plugins/discord/sqliteMirror.js
```diff
@@ -497,4 +497,21 @@ export class SqliteMirror {
...
+          SELECT id
+          FROM messages
+          WHERE channel_id = :channel_id
+          ORDER BY timestamp_ms DESC
+          LIMIT 1
+          OFFSET :offset
+        `
+      )
+      .pluck()
+      .get({channel_id: channel, offset: n - 1});
```
BrianLitwin, line 515: Might be off about this, but this indexes the n passed in from 1. I was confusing myself in the tests: I was passing in n = 4 to `nthMessageFromTail`,  with a message array `[1, 2 ,3, 4, 5]` , and getting back 1. This would return 2, which seems more intuitive. But happy to revert this as well.

decentralion:
Note: I accidentally merged #1890 into this PR, and then merged #1803 to master, when I should have done that in the reverse order.

All the code was approved and it all wound up in the right branch. But, oops!

PR #1804, opened 2020-05-19
Refactor TimelineCredScores data type

decentralion:
This commit refactors the TimelineCredScores data type so it is an
array-of-objects rather than an object-of-arrays. I want to add several
more fields (for forward cred flow, backwards cred flow, seed flow,
synthetic loop flow), and feel it will be a lot cleaner with an
array-of-objects.

This is a refactor of a local data type, and there's test coverage.
Likelihood of regression is very low.

Test plan: Updated tests; `yarn test` passes.

### DIFF
--- a/src/analysis/timeline/timelineCred.js
+++ b/src/analysis/timeline/timelineCred.js
@@ -209,12 +209,12 @@ export class TimelineCred {
     const addressToCred = new Map();
     for (let i = 0; i < nodeOrder.length; i++) {
       const addr = nodeOrder[i];
-      const addrCred = credScores.intervalCredScores.map((cred) => cred[i]);
+      const addrCred = credScores.map(({cred}) => cred[i]);
       addressToCred.set(addr, addrCred);
     }
     return new TimelineCred(
       weightedGraph,
-      credScores.intervals,
+      credScores.map((x) => x.interval),
       addressToCred,
       fullParams,
       plugins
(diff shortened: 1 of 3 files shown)


src/core/algorithm/distributionToCred.js
```diff
@@ -59,51 +51,48 @@ export function distributionToCred(
...
       intervalTotalScore === 0 ? 0 : intervalWeight / intervalTotalScore;
     const cred = distribution.map((x) => x * intervalNormalizer);
-    return cred;
+    return {interval, cred};
   });
-  return {intervalCredScores, intervals};
 }
 
-const COMPAT_INFO = {type: "sourcecred/timelineCredScores", version: "0.1.0"};
+const COMPAT_INFO = {type: "sourcecred/timelineCredScores", version: "0.2.0"};
```
Beanow, line 73: This serialization format seems to be unused other than it's unit tests.
Prefer a preceding PR to remove it, rather than updating it.

src/analysis/timeline/timelineCred.js
```diff
@@ -209,12 +209,12 @@ export class TimelineCred {
     const addressToCred = new Map();
     for (let i = 0; i < nodeOrder.length; i++) {
       const addr = nodeOrder[i];
-      const addrCred = credScores.intervalCredScores.map((cred) => cred[i]);
+      const addrCred = credScores.map(({cred}) => cred[i]);
       addressToCred.set(addr, addrCred);
     }
```
Beanow, line 214: Nit: this section is fairly opaque. Explicit types would probably help readability.

decentralion:
Merging on self-approval. Considering #1773 as the "dirty" scope here.

PR #1805, opened 2020-05-22
Make the analysis output type compatible

decentralion:
This changes the analysis output type (described in #1773) a Compatible
object by default. The rationale is that since the Output type is
intended to be an Output that is persisted to disk and shared across
system boundaries, it should always be Compatible at rest. The `extract`
method allows plucking the underlying RawOutput data from it.

This change will make some future work more convenient; in particular, I
want to automatically save the Output data as part of `sourcecred load`
rather than computing it on-the-fly with `sourcecred output`. This will
enable us to build frontends that explicitly depend on this rich output
data and may assume its existence.

Test plan: `yarn test` passes, and added a new unit test.

### DIFF
--- a/src/analysis/output.js
+++ b/src/analysis/output.js
@@ -11,11 +11,18 @@ import type {PluginDeclaration} from "./pluginDeclaration";
 import type {TimestampMs} from "../util/timestamp";
 import {TimelineCred} from "./timeline/timelineCred";
 import {nodeWeightEvaluator} from "../core/algorithm/weightEvaluator";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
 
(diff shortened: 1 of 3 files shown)


decentralion:
Fair questions all @Beanow. To be honest, it was because I needed a single function that I could plug into the LoadContext, and I didn't want to have to make another `computeFunction`-eqsue file which would just export a function that computed the output and then compatibalized it. (I posit that this is evidence that LoadContext is rotten, since it induced technical debt in upstream dependencies.)

However, I'm no longer planning to have the Output development path touch LoadContext at all, the automatic output generation will just happen as part of scv2. So I'll close this PR.

PR #1809, opened 2020-05-27
cli2: add skeleton

wchargin:
Summary:
This patch creates a new binary, `./bin/sc2`, which will be the home for
a rewrite of the CLI intended to implement an instance system. See:
<https://discourse.sourcecred.io/t/sourcecred-instance-system/244>

Paired with @decentralion.

Test Plan:
Run `yarn backend && node ./bin/sc2.js`, which should nicely fail with a
“not yet implemented” message.

wchargin-branch: cli2-skeleton

### DIFF
--- a/config/paths.js
+++ b/config/paths.js
@@ -30,6 +30,7 @@ module.exports = {
   // point within the build directory.
   backendEntryPoints: {
     sourcecred: resolveApp("src/cli/main.js"),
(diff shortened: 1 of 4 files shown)


decentralion:
Note: I pushed a fixup commit to fix the lint error in cli2/main.js.

wchargin:
Thanks! (You actually force-pushed, but I can manage.)

Issue #945, opened 2018-10-29
Define a SourceCred API


wchargin:
This is a proposal to tackle #638 by means of defining actual APIs for
configuration, for plugins, and for SourceCred itself. (This is a draft.
I’ve thought a fair amount about the concepts, but less about the
details.)

SourceCred needs to expose, first and foremost, a JavaScript API. The
`sourcecred.js` script provides a command-line entry point to this API.

This API has a top-level function:

```js
interface Sourcecred {
  load(config: LoadConfig): Promise<void>;
}

type LoadConfig = {|
  +version: "1",
  // ID for project to be loaded, like "ipfs/go-ipfs" or "ipfs-meta".
  // (The analogous concept in the current system is a repository ID,
  // which requires hacks for combining multiple repositories. This is a
  // vestige of a time when we only had single-repository support;
  // I think that the right thing to do is to decouple the project ID from
  // a specific repository.)
  +id: string,
  // Human-readable description. Optional. (We may not need this?)
  +description?: string,
  // SourceCred storage directory. Defaults to SOURCECRED_DIRECTORY.
  +directory?: string,
  // Plugins and their configurations.
  +plugins: $ReadOnlyArray<{|
    // Name of a plugin as known to SourceCred.
    +name: string,
    // Plugin-specific configuration.
    +options?: mixed,
  |}>,
|};
```

For instance:

```js
function sampleConfig(): LoadConfig {
  return {
    version: "1",
    id: "ipfs-top-2",
    description: "Meta-IPFS (top 2 repositories)",
    plugins: [
      {
        name: "sourcecred/git",
        options: {
          remoteUrls: [
            "https://github.com/ipfs/go-ipfs.git",
            "git@github.com:ipfs/secret-ipfs.git",
          ],
        },
      },
      {
        name: "sourcecred/github",
        options: {
          tokenEnvironmentVariable: "NO_PERMS_GITHUB_TOKEN",
          repositories: [
            {
              owner: "ipfs",
              name: "go-ipfs",
            },
            {
              owner: "ipfs",
              name: "secret-ipfs",
              tokenEnvironmentVariable: "PRIVATE_ACCESS_GITHUB_TOKEN",
            },
          ],
          cache: {
            ttlSeconds: 86400,
            connectionsLimit: 250, // tunable
          },
        },
      },
    ],
  };
}
```

Obviously, users aren’t expected to type all this out at the command
line. Instead, we can have configuration generators:

```js
[...]

PR #1810, opened 2020-05-27
cli2: add sketch of plugin loading

wchargin:
Summary:
This adds a `CliPlugin` interface and a basic implementation for the
GitHub plugin.

Paired with @decentralion.

Test Plan:
Create a new directory `/tmp/test-instance`, with:

```
// sourcecred.json
{"bundledPlugins": ["sourcecred/github"]}

// config/sourcecred/github/config.json
{"repositories": ["sourcecred/example-github"]}
```

Then, run

```
yarn backend &&
(cd /tmp/test-instance && node "$OLDPWD/bin/sc2.js" load)
```

and observe that the new instance has a cache directory containing a
GitHub database.

wchargin-branch: cli2-load


src/cli2/load.js
```diff
@@ -0,0 +1,24 @@
...
+    die(std, "usage: sourcecred load");
+  }
+  const baseDir = process.cwd();
+  const config = await loadInstanceConfig(baseDir);
+  for (const [name, plugin] of config.bundledPlugins) {
+    const dirContext = pluginDirectoryContext(baseDir, name);
+    plugin.load(dirContext);
+  }
+  return 0;
+};
```
Beanow, line 22: Want to suggest that there's some complecting going on here.

One part is the transport implementation we just-so-happen to be using (CLI) and the other is a use-case.

Since we're taking a greenfield approach at the CLI now, it's a distinction I'd love to see at some point. Doing so would allow exposing a first-class javascript API, similar to the likes of babel and webpack. While making the CLI transport implementation simple.

In this case the distinction would look like:
```js
const loadCommand: Command = async (args, std) => {
  // Positional arguments, transport detail.
  if (args.length !== 0) {
    die(std, "usage: sourcecred load");
  }

  // Environment data extracting, transport detail.
  const baseDir = process.cwd();

  // Call the use-case.
  loadUsecase(baseDir);

  // Exit codes, transport detail.
  return 0;
};

// Suitable to expose as Javascript API.
function loadUseCase(baseDir: string): void {
  const config = await loadInstanceConfig(baseDir);
  for (const [name, plugin] of config.bundledPlugins) {
    const dirContext = pluginDirectoryContext(baseDir, name);
    plugin.load(dirContext);
  }
}
```

src/plugins/github/cliPlugin.js
```diff
@@ -0,0 +1,108 @@
...
+): Promise<GithubConfig> {
+  const dirname = dirContext.configDirectory();
+  const path = pathJoin(dirname, "config.json");
+  const contents = await fs.readFile(path);
+  return Promise.resolve(parseConfig(JSON.parse(contents)));
+}
+
+// Shim to interface with `fetchGithubRepo`; TODO: refactor that to just
+// take a directory.
+class CacheProviderImpl implements CacheProvider {
```
Beanow, line 35: Actually, I'm a huge fan of using these interfaces. It decouples the rest of the code from the implementation details of using filesystems. Would recommend applying the same idea to the configuration and tokens too like `ConfigProvider` and `SecretsProvider`.

On the flipside, I would rather not see `fs.readFile` or `process.env` in a _use-case_.

To give a common example where this practice pays off. Many applications use ENV for secrets. However interesting when using Docker Swarm, ENV is not suitable as it will be passed around in plaintext. Instead it should use "Docker Secrets", which most application implement by accepting a `*_FILE` suffix to secrets and reading it from that file path.

Having a central `SecretsProvider` implementation, adding this in would be easy and something the core can implement once. Having `process.env` scattered across the plugins, this would mean every plugin maintainer has to figure this out.

src/cli2/common.js
```diff
@@ -0,0 +1,61 @@
...
+// Make a directory, if it doesn't exist.
+function mkdirx(path: string) {
+  try {
+    fs.mkdirSync(path);
+  } catch (e) {
+    if (e.code !== "EEXIST") {
+      throw e;
+    }
+  }
+}
```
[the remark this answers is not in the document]
wchargin: That does something else. We don’t want to make all the transitive
parent directories.

src/cli2/common.js
```diff
@@ -0,0 +1,61 @@
...
+  if (idParts.length !== 2) {
+    throw new Error(`Bad plugin name: ${pluginId}`);
+  }
+  const [pluginOwner, pluginName] = idParts;
+  const pathComponents = [...prefix, pluginOwner, pluginName];
+  let path = baseDir;
+  for (const pc of pathComponents) {
+    path = pathJoin(path, pc);
+    mkdirx(path);
+  }
```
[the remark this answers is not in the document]
wchargin: As above.

src/cli2/common.js
```diff
@@ -0,0 +1,61 @@
...
+
+import type {PluginDirectoryContext} from "./cliPlugin";
+import {parse as parseConfig, type InstanceConfig} from "./instanceConfig";
+
+export async function loadInstanceConfig(
+  baseDir: string
+): Promise<InstanceConfig> {
+  const projectFilePath = pathJoin(baseDir, "sourcecred.json");
+  const contents = await fs.readFile(projectFilePath);
+  return Promise.resolve(parseConfig(JSON.parse(contents)));
```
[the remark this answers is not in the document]
wchargin: Yes, but only because of a JavaScript [design flaw][1] that has really
aged like milk with the prevalence of TypeScript and Flow. I prefer to
write the code in the sensible, principled style, even though it’s not
strictly required by the language.

[1]: https://github.com/promises-aplus/promises-spec/issues/94#issuecomment-16176966

src/cli2/instanceConfig.js
```diff
@@ -0,0 +1,55 @@
...
+// dependencies.
+export type BundledPluginSpec = string;
+
+type JsonObject =
+  | string
+  | number
+  | boolean
+  | null
+  | JsonObject[]
+  | {[string]: JsonObject};
```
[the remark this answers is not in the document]
wchargin: We (@decentralion and I) chatted about this a bit. We chose to write the
correct-style parser here (perhaps because I was driving ;-) ) because
it really does give us much stronger guarantees than the `any`-downcast,
though we definitely aren’t big fans of the verbosity. I’m not super
inclined to reflect Flow types into parsers, but I’m up for abstracting
this with an attoparsec-like or optics-based approach if we think that
that would be helpful. But later, not now; let’s find out how we
actually use these first.

src/cli2/common.js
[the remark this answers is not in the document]
Beanow: Certainly won't argue with you that Promises have a bunch of design flaws. I've built several client projects with https://github.com/fluture-js/Fluture instead of Promises and there's a lot of great properties there that I'm sorely missing having gone back to Promises.

On the other hand, the language integration definitely yields productivity benefits over going against the unfortunate standard, and reduces learning curve.

In this case `return Promise.resolve()` is a pattern that makes use of one of those Promise design choices, to silently unpack Promises at every opportunity.

In other words:
```js
Promise.resolve(Promise.resolve(123))
> Promise { <state>: "fulfilled", <value>: 123 }
```
We get a `Promise<number>` not a `Promise<Promise<number>>`.
A surprising and undesirable property if you ask me.

My taste preference here is to avoid that surprise, so I generally encourage others not to `return Promise.resolve()`.

Anyhow, if you feel like there's a benefit to going the other way, curious to hear about it. But for this PR, I'm fine with either :]

src/cli2/load.js
Beanow: Based on the dev call today. If we do want to run a local daemon we can send commands over HTTP, I think this will help out.

IPFS and Docker are examples that invested in this structure, as they expose their commands through multiple transports (library, unix sockets and HTTP)

src/cli2/common.js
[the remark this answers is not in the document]
decentralion: Personally, I'm not a fan of this style. The linked comment does not explain why this is a design flaw, and I'm used to the semantics of just returning things from async functions and getting promises. I'm not going to ask for this to change, since it seems correct either way (see NHIWHDI), but in code that I write I'll continue to use the same idiom I've used in the past.

src/cli2/load.js
decentralion: I agree with Beanow that we'll want to expose a functional style for implementing this logic that isn't hard-wired into the CLI. I don't need that to happen in this commit--I think it will be easy to factor out in the future when we need it. Though it could also be nice to have. I approve either way.

src/cli2/common.js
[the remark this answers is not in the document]
teamdandelion: I prefer the non-redundant style, although haven't invested the effort to really understand why Willow avoids it. Since we don't have a lint rule to automatically enforce it, I say it's authors pick on which to use.

src/cli2/load.js
decentralion: I agree that we'll eventually want an api-ified version of this that is decoupled from the CLI. agnostic on when that happens, it should be easy to factor out later.

src/plugins/github/cliPlugin.js
decentralion: > On the flipside, I would rather not see fs.readFile or process.env in a use-case.

I think we're in agreement that stuff like `fs.readFile` and `process.env` should be kept at the outermost layer possible, and decoupled from implementing the logic.

IMO, losing the CacheProvider is not a big deal because we aren't actually supporting other filesystems, so as developers we aren't getting feedback on whether we're (for example) using CacheProvider consistently across the codebase. Without that feedback, there's no particular reason to believe that we *will* actually be using it consistently when we need that flexibility, or even that this is giving us the right flexibility compared to what we'll actually need when we need it. So we don't benefit much from it, and it does add extra indirection and cognitive overhead.

src/cli2/common.js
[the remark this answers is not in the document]
wchargin: That `resolve(resolve(123))` is equivalent to `resolve(123)` is
precisely the design flaw that I’m trying to avoid. If the standard
didn’t have that flaw, then you would have to return `resolve(123)`
here, so returning `123` is surfacing the flaw, not hiding it. When
I read `async function foo() { return 123; }`, which shouldn’t compile,
I have to remind myself that it works because of this flaw.

I’m happy to discuss in more detail, and I can provide a specific
example of code that this design flaw makes it impossible to write
cleanly (forcing ugly contortions in the form of boxing) as well as
point to some of the complexity that the flaw forces in real code. I’ll
keep it as is for now, and am fine with it being author’s pick, just as
so many other things are.

src/cli2/load.js
wchargin: Yes, I of course also agree that it’d be great to have a JS API that
backs the CLI and is also exposed. This is basically what I describe at
the end of #945. But I don’t think that defining such an API should
block developing this CLI.

src/cli2/cliPlugin.js
```diff
@@ -0,0 +1,17 @@
...
+
+import type {PluginDeclaration} from "../analysis/pluginDeclaration";
+import type {WeightedGraph} from "../core/weightedGraph";
+import type {ReferenceDetector} from "../core/references/referenceDetector";
+
+export interface CliPlugin {
+  declaration(): PluginDeclaration;
+  load(PluginDirectoryContext): Promise<void>;
+  graph(PluginDirectoryContext, ReferenceDetector): Promise<WeightedGraph>;
+  referenceDetector(PluginDirectoryContext): Promise<ReferenceDetector>;
```
decentralion, line 11: Let's provide a TaskReporter to each of these methods.

src/cli2/load.js
```diff
@@ -0,0 +1,24 @@
...
+
+const loadCommand: Command = async (args, std) => {
+  if (args.length !== 0) {
+    die(std, "usage: sourcecred load");
+  }
+  const baseDir = process.cwd();
+  const config = await loadInstanceConfig(baseDir);
+  for (const [name, plugin] of config.bundledPlugins) {
+    const dirContext = pluginDirectoryContext(baseDir, name);
+    plugin.load(dirContext);
```
decentralion, line 19: Let's have the task reporter report that load started for the plugin--that way we have consistent measurements for all the plugins, and the plugin can use the provided task reporter to give more detailed information (e.g. timing on sub-portions)

src/plugins/github/cliPlugin.js
wchargin: I’m not really familiar with the way in which you’re using the term
*use-case*. Perhaps you could clarify?

Currently, `fetchGithubRepo` takes a simple `token: GithubToken`. It’s
up to the caller to fetch that, directly from an environment variable or
from a file or from a cast5-decryption or whatever. It’s beyond the
scope of `fetchGithubRepo` to care where that token came from. I don’t
see how introducing a `secretProvider: () => GithubToken` (presumably?)
makes that any simpler.

PR #1811, opened 2020-05-27
sc2: add graph subcommand

wchargin:
Summary:
Paired with @decentralion.

Test Plan:
Follow the test plan for #1810, then additionally run

```
(cd /tmp/test-instance && node "$OLDPWD/bin/sc2.js" graph)
```

and note that the `output/graphs/...` directory has a graph JSON file.

wchargin-branch: cli2-graph


src/cli2/graph.js
```diff
@@ -0,0 +1,55 @@
...
+    ([k, _]) => k
+  )) {
+    const dirContext = pluginDirectoryContext(baseDir, name);
+    const rd = await plugin.referenceDetector(dirContext);
+    rds.push(rd);
+  }
+  return new CascadingReferenceDetector(rds);
+}
+
+export default loadCommand;
```
decentralion, line 55: should be graphCommand. (or just "command").

wchargin: Thanks! Fixed. I’ll keep `graphCommand` so that when we add back
`helpCommand` as before (assuming that this follows a similar pattern)
we don’t need to just rename it back.

PR #1814, opened 2020-05-30
cli2: add Discourse CLI plugin

decentralion:
This updates the v2 CLI so that it now supports the Discourse plugin.

Test plan:

Modify the test instance described in the previous commits so that the
root `sourcecred.json` file includes "sourcecred/discourse" in the list
of bundled plugins. Then, add a
`config/sourcecred/discourse/config.json` file with the following
contents:
`{"serverUrl": "https://sourcecred-test.discourse.group/"}`

Now, running `sc2 load` will load Discourse data, and `sc2 graph` writes
a Discourse graph in the output directory.

### DIFF
--- a/src/plugins/discourse/cliPlugin.js
+++ b/src/plugins/discourse/cliPlugin.js
@@ -1,0 +1,74 @@
+// @flow
+
+import Database from "better-sqlite3";
+import fs from "fs-extra";
+import {join as pathJoin} from "path";
+
+import type {CliPlugin, PluginDirectoryContext} from "../../cli2/cliPlugin";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {ReferenceDetector} from "../../core/references/referenceDetector";
+import type {WeightedGraph} from "../../core/weightedGraph";
+import {createGraph} from "./createGraph";
+import {declaration} from "./declaration";
+import {parseConfig, type DiscourseConfig} from "./config";
+import {weightsForDeclaration} from "../../analysis/pluginDeclaration";
+import {SqliteMirrorRepository} from "./mirrorRepository";
+import {Fetcher} from "./fetch";
+import {Mirror} from "./mirror";
+import {DiscourseReferenceDetector} from "./referenceDetector";
+import {type TaskReporter} from "../../util/taskReporter";
+
+async function loadConfig(
+  dirContext: PluginDirectoryContext
+): Promise<DiscourseConfig> {
+  const dirname = dirContext.configDirectory();
+  const path = pathJoin(dirname, "config.json");
+  const contents = await fs.readFile(path);
+  return Promise.resolve(parseConfig(JSON.parse(contents)));
+}
+
+async function repository(
+  ctx: PluginDirectoryContext,
+  serverUrl: string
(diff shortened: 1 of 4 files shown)


src/plugins/discourse/cliPlugin.js
```diff
@@ -0,0 +1,74 @@
...
+    const fetcher = new Fetcher({serverUrl});
+    const mirror = new Mirror(repo, fetcher, serverUrl, mirrorOptions);
+    await mirror.update(reporter);
+  }
+
+  async graph(
+    ctx: PluginDirectoryContext,
+    rd: ReferenceDetector
+  ): Promise<WeightedGraph> {
+    const _ = rd; // TODO(#1808): not yet used
```
wchargin, line 59: OK: I edited this issue to include the Discourse plugin, too.

src/plugins/discourse/config.js
```diff
@@ -0,0 +1,67 @@
...
+  if (typeof serverUrl !== "string") {
+    throw new Error("serverUrl not string: " + JSON.stringify(serverUrl));
+  }
+  const httpRE = new RegExp(/^https?:\/\//);
+  if (!httpRE.test(serverUrl)) {
+    throw new Error(
+      "expected server url to start with 'https://' or 'http://'"
+    );
+  }
+  if (raw.mirrorOptions !== undefined) {
```
wchargin, line 33: As noted, this code lets the `mirrorOptions` be omitted, but it doesn’t
let it be partial: if the object exists, so must all its keys. This is
sound, but it’s more strict than the `$Shape<MirrorOptions>` would
require. Okay with me to use that strict behavior for now, especially
because it’ll get cleaned up shortly.

src/plugins/discourse/config.js
```diff
@@ -0,0 +1,67 @@
...
+    if (!recheckTopicsInCategories.every((x) => typeof x === "number")) {
+      throw new Error(
+        "mirrorOptions.recheckTopicsInCategories must all be numbers, got " +
+          JSON.stringify(recheckTopicsInCategories)
+      );
+    }
+    recheckTopicsInCategories = recheckTopicsInCategories.map((x) => Number(x));
+    if (typeof recheckCategoryDefinitionsAfterMs !== "number") {
+      throw new Error(
+        "recheckCategoryDefinitionsAfterMs msut be number, got " +
```
wchargin, line 56: sp.: *msut* → *must*

src/plugins/discourse/config.test.js
```diff
@@ -0,0 +1,88 @@
...
+      };
+      expect(() => parseConfig(c1)).toThrowError();
+      const c2 = {
+        serverUrl: "https://server.io",
+        mirrorOptions: {
+          recheckCategoryDefinitionsAfterMs: 12,
+        },
+      };
+      expect(() => parseConfig(c2)).toThrowError();
+    });
```
wchargin, line 68: Duplicate of previous test; remove?

src/plugins/discourse/config.js
decentralion: Done.

src/plugins/discourse/config.test.js
decentralion: Done.

PR #1815, opened 2020-05-30
Implement bare-bones sc2 merge and score

decentralion:
This adds the missing `merge` and `score` commands to the sc2 CLI.
`merge` currently just merges the plugin graphs, and doesn't yet support
identity resolution or weight overrides.

`score` computes the new data output format (cf #1773) and writes it to
disk. It doesn't yet support using custom parameters.

Test plan:
Follow the test plans for the previous commits, then run `sc2 merge`.
It will create a combined graph at `output/graph.json`, which will
contain data for both Discourse and GitHub. Then run `sc2 score` and
the `output/output.json` file will contain scores for the combined
graph.

### DIFF
--- a/src/cli2/merge.js
+++ b/src/cli2/merge.js
@@ -1,0 +1,53 @@
+// @flow
+
+import fs from "fs-extra";
+import stringify from "json-stable-stringify";
+import {join as pathJoin} from "path";
+
+import {LoggingTaskReporter} from "../util/taskReporter";
+import type {Command} from "./command";
+import {makePluginDir, loadInstanceConfig} from "./common";
+import {
+  toJSON as weightedGraphToJSON,
+  fromJSON as weightedGraphFromJSON,
+  type WeightedGraph,
+  merge,
+} from "../core/weightedGraph";
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  return 1;
+}
+
+const mergeCommand: Command = async (args, std) => {
+  if (args.length !== 0) {
+    return die(std, "usage: sourcecred merge");
+  }
+  const taskReporter = new LoggingTaskReporter();
+  taskReporter.start("merge");
+  const baseDir = process.cwd();
+  const config = await loadInstanceConfig(baseDir);
+  const graphOutputPrefix = ["output", "graphs"];
+
+  async function loadGraph(pluginName): Promise<WeightedGraph> {
+    const outputDir = makePluginDir(baseDir, graphOutputPrefix, pluginName);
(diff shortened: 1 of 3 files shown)


src/cli2/merge.js
```diff
@@ -0,0 +1,53 @@
...
+} from "../core/weightedGraph";
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  return 1;
+}
+
+const mergeCommand: Command = async (args, std) => {
+  if (args.length !== 0) {
+    die(std, "usage: sourcecred merge");
```
wchargin, line 24: `return die`, no? lest we continue.

(The other commands miss this, too, but we can fix it here first.)

src/cli2/merge.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+  const taskReporter = new LoggingTaskReporter();
+  taskReporter.start("merge");
+  const baseDir = process.cwd();
+  const config = await loadInstanceConfig(baseDir);
+  const graphOutputPrefix = ["output", "graphs"];
+
+  async function loadGraph(pluginName): Promise<WeightedGraph> {
+    const outputDir = makePluginDir(baseDir, graphOutputPrefix, pluginName);
```
wchargin, line 33: Do we actually want to make this directory? It doesn’t seem semantically
right here—we’re expecting a file in that directory to exist, anyway.
Maybe add `getPluginDir` parallel to `makePluginDir` (in `common`) that
just returns the pathname?

src/cli2/score.js
```diff
@@ -0,0 +1,56 @@
...
+  const params = defaultParams();
+
+  const tc = await TimelineCred.compute({
+    weightedGraph: graph,
+    params,
+    plugins: declarations,
+  });
+  const output = fromTimelineCredAndPlugins(tc, declarations);
+  const outputJSON = stringify(toCompat(OUTPUT_COMPAT_INFO, output));
+  const outputPath = pathJoin(baseDir, "output", "output.json");
```
wchargin, line 50: Maybe `scores.json` rather than `output.json`, for clarity? It’s under
the `output/` directory already, after all, and it’s not the only
output.

src/cli2/merge.js
decentralion: Discussed offline, you'll make this in a separate change.

src/cli2/merge.js
decentralion: Done.

src/cli2/score.js
decentralion: Discussed offline; renamed to `cred.json`.

PR #1819, opened 2020-05-30
combo: allow field renaming

wchargin:
Summary:
This patch expands the API of `Combo.object` such that fields in the
input JSON may be renamed in the output JSON. This often occurs in
combination with `fmap`, as we convert a simple string field with a
user-facing name to a different structured representation. For example:

```javascript
C.object({
  repoIds: C.rename(
    "repositories",
    C.fmap(C.array(C.string), repoIdToString)
  ),
});
```

This is backward-compatible and invisible when not needed: the fields of
the argument to `C.object` may now be either parsers (as before) or
results of `C.rename`.

This patch also adds a check that the required and optional key sets
don’t overlap, which could technically have happened before but is more
important now that renames are possible.

Test Plan:

wchargin-branch: combo-rename-fields


src/util/combo.test.js
```diff
@@ -275,5 +275,69 @@ describe("src/util/combo", () => {
...
+      });
+      it("provides missing key errors using the user-facing name", () => {
+        const thunk = () => p.parseOrThrow({one: 1, cuatro: 4});
+        expect(thunk).toThrow('missing key: "dos"');
+      });
+      it("only accepts the user-facing keys", () => {
+        const thunk = () => p.parseOrThrow({one: 1, two: 2});
+        expect(thunk).toThrow('missing key: "dos"');
+      });
+      it("only accepts the user-facing keys", () => {
```
decentralion, line 326: Duplicated test.

src/util/combo.test.js
```diff
@@ -275,5 +275,69 @@ describe("src/util/combo", () => {
...
+        C.object({hmm: C.string}, {hmm: C.rename("hum", C.string)});
+      }).toThrow('duplicate key: "hmm"');
+    });
+    it("forbids renaming a rename at the type level", () => {
+      C.rename("old", C.string);
+      // $ExpectFlowError
+      (C.rename("old", C.string): C.Parser<string>);
+      // $ExpectFlowError
+      C.rename("hmm", C.rename("old", C.string));
+    });
```
decentralion, line 363: I don't understand what this test is doing.

src/util/combo.test.js
wchargin: Thanks—I think that I meant to make this a test for user-facing optional
keys, which I’ve now done.

src/util/combo.test.js
wchargin: Clarified as discussed.

PR #1822, opened 2020-05-30
distributionToCred propagates cred flow data

decentralion:
This commit modifies the distributionToCred module so that in addition
to normalizing the scores into cred, it also normalizes the score flows
(from #1802) into cred flows. This is another big part of #1773, so that
we can support the `OutputEdge` format and give meaningful info on cred
flows for each edge.

The change is quite simple, as we already compute a normalization
constant, we just need to apply it to a few more arrays.

Test plan: I added a new unit test that has values in all of the
different flows, so we can validate that they are normalized
consistently.


decentralion:
Merging on self-review.

PR #1825, opened 2020-05-31
combo: add `tuple` combinator

wchargin:
Summary:
This new combinator parses heterogeneous tuples: arrays of fixed length,
which may have different, fixed types at each index. For example:

```javascript
type CompactDistribution = [NodeAddressT, number][];
function compactDistributionParser(): C.Parser<CompactDistribution> {
  return C.array(
    C.tuple([
      C.fmap(C.array(C.string), NodeAddress.fromParts),
      C.number,
    ])
  );
}
```

Or:

```javascript
function compatParser<T>(underlying: C.Parser<T>): C.Parser<Compatible<T>> {
  return C.tuple([
    C.object({type: C.string, version: C.string}),
    underlying,
  ]);
}
```

Test Plan:

wchargin-branch: combo-tuple


wchargin:
Good point; done. (This type magic is fairly straightforward and
benign—it’s commonplace for things like `Promise.all`—but it’s
definitely good to have a test for it.)

PR #1829, opened 2020-05-31
combo: add `exactly` combinator for fixed values

wchargin:
Summary:
Write `C.exactly(["red", "green", "blue"])` to admit any of a fixed set
of primitive values. This can already be implemented in terms of `raw`
and `fmap`, but it’s more convenient to have a dedicated helper.

Test Plan:

wchargin-branch: combo-exactly


src/util/combo.js
```diff
@@ -114,6 +114,35 @@ export function pure<T>(t: T): Parser<T> {
...
+// provide two constants `ONE = {}` and `TWO = {}` (different objects),
+// and then expect that any value of type `T` would be identical to
+// either `ONE` or `TWO`. Using strict reference equality for array and
+// object types would be sound, but would not usually be what was
+// wanted, as it wouldn't match ~any actual output of `JSON.parse`.
+export function exactly<T: string | number | boolean | null>(
+  ts: $ReadOnlyArray<T>
+): Parser<T> {
+  const expected: string =
+    ts.length === 1 ? String(ts[0]) : `one of ${JSON.stringify(ts)}`;
```
decentralion, line 135: Why define expected out here, rather than inside the parser (i.e. construct it when needed for the failure)?

wchargin: I did it that way to reduce total work done: you can compile a parser
once and use it for many parses, and both clients and other combinators
(like `orElse`) may expect to recover from parse errors (i.e., they’re
not application-fatal). But it’s okay with me to defer instead; done.

PR #1830, opened 2020-06-01
Add the CredResult module

decentralion:
This module has a concise and clean type for storing all the output data
from running SourceCred, including:
- The graph
- The weights
- The cred scores and flows
- The interval timing info
- The plugins used
- The parameters used

We also have a method `compute` which computes one of these given the
weighted graph, parameters, and plugins. It's all quite clean and
simple. I feel good about thsi API.

Test plan:
The main `compute` function is only sanity checked via flow and unit
testing, which is appropriate since it's a pure-piping function with
little than can go wrong. I've also added JSON serialization; this is
tested with round trip testing.

### DIFF
--- a/src/analysis/credResult.js
+++ b/src/analysis/credResult.js
@@ -1,0 +1,87 @@
+// @flow
+
+import {
+  type WeightedGraph,
+  type WeightedGraphJSON,
+  toJSON as wgToJSON,
+  fromJSON as wgFromJSON,
+} from "../core/weightedGraph";
+import {type Compatible, toCompat, fromCompat} from "../util/compat";
+import {
+  type TimelineCredParameters,
+  type TimelineCredParametersJSON,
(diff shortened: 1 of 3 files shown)


src/analysis/credResult.js
```diff
@@ -0,0 +1,87 @@
...
+  +credData: CredData,
+  +params: TimelineCredParametersJSON,
+  +plugins: PluginDeclarationsJSON,
+|}>;
+
+export function toJSON(x: CredResult): CredResultJSON {
+  const {weightedGraph, credData, params, plugins} = x;
+  return toCompat(COMPAT_INFO, {
+    weightedGraph: wgToJSON(weightedGraph),
+    credData,
```
decentralion, line 72: Note: Currently the credData object is just a raw JSON object, so we don't need any special serialization here.

PR #1831, opened 2020-06-01
Compute CredData from TimelineCredScores

decentralion:
This commit adds the `analysis/credData` module, which processes raw
TimelineCredScores into a format which is better for serialization and
data analysis. In particular, this format explicitly stores the summary
(summed-across-time) data for nodes separately from the raw temporal
data, which will allow us to throw away raw data for uninteresting or
low-cred nodes, while keeping the summary.

Test plan: I've added some basic unit tests; run `yarn test`.

### DIFF
--- a/src/analysis/credData.js
+++ b/src/analysis/credData.js
@@ -1,0 +1,124 @@
+// @flow
+
+import type {TimestampMs} from "../util/timestamp";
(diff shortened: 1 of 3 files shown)


decentralion:
Merging on self review.

PR #1832, opened 2020-06-01
Add lossy cred compression strategy

decentralion:
This adds a new `compressByThreshold` method in the credResult module,
which compresses the CredResult by removing interval-level cred data for
flows of cred that are below a threshold.

Test plan: Unit tests included; `yarn test` passes.


decentralion:
Merging on self review.

PR #1833, opened 2020-06-02
cli2 writes CredResult as cred.json

decentralion:
This commit modifies the cli2 score command so that instead of writing
the output format described in #1773, it instead writes the new
CredResult format added in #1830. This is a much better engineered
format, and will be useful for building a new frontend as well as for
analysis by third parties (although we will need to provide client code
for parsing it).

Test plan: On a cli2 instance, run `sc2 score` and inspect the resulting
credResult.json file.


decentralion:
Note: It's a little redundant to store the `graph.json` alongside the `credResult.json` since the graph is also stored inside the credResult. However, it's convenient for now, both to give an intuition for how large the graphs are, and because some applications may want just the graph without the credResult.

So we'll do it for now, may improve on it later for efficiency.

PR #1844, opened 2020-06-08
credrank: add Markov process graph and basic CLI

wchargin:
Summary:
We introduce the *Markov process graph* data type: a weighted graph with
unidirectional edges whose out-weights sum to 1 for each node, and whose
mixing process semantics are given directly by the transition matrix.
Higher-order transformations like seed teleportation and epoch fibration
are compiled into the graph as new nodes and edges, which behave just
the same way as underlying graph nodes. For more background, see the
CredRank initiative:
<https://discourse.sourcecred.io/t/credrank-scalable-interpretable-flexible-attribution/654>

The main purpose of this patch is to start getting pieces of this work
into master to unblock downstream development. We include a basic CLI
entry point for manual testing, but no automated tests yet, and the
interfaces should be considered merely “not obviously wrong”.

Based on original work in collaboration with @decentralion.

Test Plan:
On a test instance with the GitHub plugin pointing at the SourceCred
core repository, running the `load`, `graph`, `credrank` sequence yields
an `output/credGraph.json`, and a spot-check of the top-scoring users—

```
jq '
    .scores as $scores
    | [
        $scores | keys | .[]
        | {raw: ., parts: split("\u0000")}
        | select(.parts[2:5] == ["github", "USERLIKE", "USER"])
        | {who: .parts | join("/"), score: $scores[.raw]}
    ]
    | sort_by(-.score)[:10]
    | map({(.who): .score})
    | add
'
```

—is roughly consistent with what we’d expect. The `credrank` step takes
about 1–2 seconds to build the Markov process graph and 500 ms to run
CredRank (on my laptop), compared to about 50 seconds to run timeline
cred analysis, so we’re looking at a speedup of multiple orders of
magnitude.

wchargin-branch: credrank-mpg-cli

### DIFF
--- a/src/core/credGraph.js
+++ b/src/core/credGraph.js
@@ -1,0 +1,99 @@
+// @flow
+
+import * as NullUtil from "../util/null";
+import * as MapUtil from "../util/map";
+import {type NodeAddressT, type EdgeAddressT} from "./graph";
+import {type NodeWeight} from "./weights";
+import {
+  MarkovProcessGraph,
+  type MarkovProcessGraphJSON,
+  type MarkovEdge,
+  type TransitionProbability,
+} from "./markovProcessGraph";
+
+export type Node = {|
+  +address: NodeAddressT,
+  +description: string,
+  +cred: number,
+  +weight: NodeWeight,
+|};
+
+export type Edge = {|
+  +address: EdgeAddressT,
+  +reversed: boolean,
+  +src: NodeAddressT,
+  +dst: NodeAddressT,
+  +transitionProbability: TransitionProbability,
+  +credFlow: number,
+|};
+
+export type CredGraphJSON = {|
+  +mpg: MarkovProcessGraphJSON,
+  +scores: {|+[NodeAddressT]: number|},
+|};
+
+export class CredGraph {
+  _mpg: MarkovProcessGraph;
+  _scores: Map<NodeAddressT, number>;
+
+  constructor(
+    markovProcessGraph: MarkovProcessGraph,
+    scores: Map<NodeAddressT, number>
+  ) {
+    this._mpg = markovProcessGraph;
+    this._scores = scores;
+  }
+
+  _cred(addr: NodeAddressT): number {
+    return NullUtil.get(this._scores.get(addr));
+  }
+
+  _credFlow(edge: MarkovEdge): number {
+    const srcCred = this._cred(edge.src);
+    return srcCred * edge.transitionProbability;
+  }
+
+  node(addr: NodeAddressT): ?Node {
+    const node = this._mpg._nodes.get(addr);
+    if (node == null) return undefined;
+    return {...node, cred: this._cred(addr)};
+  }
+
+  *nodes(): Iterator<Node> {
+    for (const node of this._mpg.nodes()) {
+      yield {...node, cred: this._cred(node.address)};
+    }
+  }
+
+  *edges(): Iterator<Edge> {
+    for (const edge of this._mpg.edges()) {
+      yield {...edge, credFlow: this._credFlow(edge)};
+    }
+  }
+
+  *scoringNodes(): Iterator<Node> {
+    for (const addr of this._mpg.scoringAddresses()) {
+      yield NullUtil.get(this.node(addr));
+    }
+  }
+
+  *inNeighbors(addr: NodeAddressT): Iterator<Edge> {
+    for (const edge of this._mpg.inNeighbors(addr)) {
+      yield {...edge, credFlow: this._credFlow(edge)};
+    }
+  }
+
+  toJSON(): CredGraphJSON {
+    return {
+      mpg: this._mpg.toJSON(),
+      scores: MapUtil.toObject(this._scores),
+    };
+  }
+
+  static fromJSON(j: CredGraphJSON): CredGraph {
+    return new CredGraph(
+      MarkovProcessGraph.fromJSON(j.mpg),
+      MapUtil.fromObject(j.scores)
+    );
+  }
+}
--- a/src/core/credGraph.js
+++ b/src/core/credGraph.js
@@ -3,7 +3,6 @@
 import * as NullUtil from "../util/null";
 import * as MapUtil from "../util/map";
 import {type NodeAddressT, type EdgeAddressT} from "./graph";
-import {type NodeWeight} from "./weights";
 import {
   MarkovProcessGraph,
   type MarkovProcessGraphJSON,
--- a/src/core/credGraph.js
+++ b/src/core/credGraph.js
@@ -15,7 +15,7 @@ export type Node = {|
   +address: NodeAddressT,
   +description: string,
   +cred: number,
-  +weight: NodeWeight,
+  +mint: number,
 |};
 
 export type Edge = {|
--- a/src/core/credGraph.js
+++ b/src/core/credGraph.js
@@ -48,7 +48,7 @@ export class CredGraph {
   }
 
   _credFlow(edge: MarkovEdge): number {
-    const srcCred = this._cred(edge.src);
+    const srcCred /* heh */ = this._cred(edge.src);
     return srcCred * edge.transitionProbability;
   }
 
--- a/src/core/credGraph.js
+++ b/src/core/credGraph.js
@@ -2,7 +2,7 @@
 
 import * as NullUtil from "../util/null";
 import * as MapUtil from "../util/map";
-import {type NodeAddressT, type EdgeAddressT} from "./graph";
+import type {NodeAddressT, EdgeAddressT} from "./graph";
 import {
   MarkovProcessGraph,
   type MarkovProcessGraphJSON,
--- a/src/core/markovProcessGraph.js
+++ b/src/core/markovProcessGraph.js
@@ -1,0 +1,579 @@
+// @flow
+
+/**
+ * Data structure representing a particular kind of Markov process, as
+ * kind of a middle ground between the semantic SourceCred graph (in the
+ * `core/graph` module) and a literal transition matrix. Unlike the core
+ * graph, edges in a Markov process graph are unidirectional, edge
+ * weights are raw transition probabilities (which must sum to 1) rather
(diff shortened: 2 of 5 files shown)


src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+ *     forces well-definedness and ergodicity of the Markov process (for
+ *     nonzero alpha, and assuming that there is at least one edge in
+ *     the underlying graph).
+ *
+ * The edges include:
+ *
+ *   - *contribution edges* between nodes in the underlying graph, which
+ *     are lifted to their corresponding contribution nodes or to epoch
+ *     nodes if either endpoint has been fibrated;
+ *   - *seed-in* / "redistribution" / "radiation" edges from nodes to
```
wchargin, line 43: > radiation

This name occurred to me while mulling. The implication is twofold[.][1]
First, the seed-in edges cause the normal cred flow to “lose energy” due
to a background radiation of warmth back to the root. But also seed-in
edges are required for the Markov process to be well defined, because
they define the action at “black hole” nodes—and black holes dissipate
via Hawking radiation.

I’d like *some* name better than “seed-in”, which I feel isn’t entirely
unambiguous (node *v*’s “seed-in” edge is actually an out-edge of *v*).
*Redistribution* is more direct, but sometimes an appropriate splash of
color doesn’t hurt, either. Thoughts welcome. :-)

[1]: https://en.wikipedia.org/wiki/%C4%AAh%C4%81m

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+        }
+      })()) {
+        if (edge.weight > 0) {
+          yield edge;
+        }
+      }
+    };
+
+    const srcNodes: Map<
+      NodeAddressT /* domain: (nodes with out-edges) U (epoch nodes) */,
```
decentralion, line 446: I originally interpreted this comment to mean "every node with out-edges and every epoch node will be included in this map". However:
- nodes with only zero-weight out edges are not included
- epoch nodes are only included if there was an out-edge from the corresponding scoring node with a timestamp in that epoch node's interval

suggest updating the comment for clarity

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+  }
+
+  *nodes(options?: {|+prefix: NodeAddressT|}): Iterator<MarkovNode> {
+    const prefix = options ? options.prefix : NodeAddress.empty;
+    for (const node of this._nodes.values()) {
+      if (NodeAddress.hasPrefix(node.address, prefix)) {
+        yield node;
+      }
+    }
+  }
```
decentralion, line 494: Lately I've been thinking we should just return arrays rather than iterators--i cant recall ever benefiting from using the iterate and it's a bit more cumbersome. 

no change requested.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+
+  *inNeighbors(nodeAddress: NodeAddressT): Iterator<MarkovEdge> {
+    for (const edge of this._edges.values()) {
+      if (edge.dst !== nodeAddress) {
+        continue;
+      }
+      yield edge;
+    }
+  }
```
decentralion, line 509: Placeholder comment since I haven't seen this used yet.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
+// @flow
+
+/**
+ * Data structure representing a particular kind of Markov process, as
+ * kind of a middle ground between the semantic SourceCred graph (in the
+ * `core/graph` module) and a literal transition matrix. Unlike the core
+ * graph, edges in a Markov process graph are unidirectional, edge
+ * weights are raw transition probabilities (which must sum to 1) rather
```
decentralion, line 8: The "must sum to 1" constraint doesn't seem to be enforced, e.g. for an epoch node with no out-edges. In which case the edge weights sum to `(1-transitionRemainder)`.  This is based on looking inside `static new` for any re-normalization in the case where there weren't any out edges, and not finding it.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+      inEdges.forEach((e, i) => {
+        // Note: We don't group-by src, so there may be multiple `j`
+        // such that `neighbor[j] === k` for a given `k` when there are
+        // parallel edges in the source graph. This should just work
+        // down the stack.
+        const result = nodeIndex.get(e.src);
+        if (result == null) {
+          throw new Error(e.src);
+        }
+        neighbor[i] = result;
```
decentralion, line 540: would prefer if `result` were renamed to `neighborIndex`

src/core/credGraph.js
```diff
@@ -0,0 +1,99 @@
...
+    for (const addr of this._mpg.scoringAddresses()) {
+      yield NullUtil.get(this.node(addr));
+    }
+  }
+
+  *inNeighbors(addr: NodeAddressT): Iterator<Edge> {
+    for (const edge of this._mpg.inNeighbors(addr)) {
+      yield {...edge, credFlow: this._credFlow(edge)};
+    }
+  }
```
decentralion, line 84: This method doesn't seem to be used?

wchargin:
Thanks for the review! Will revise and respond by end of week, hopefully
sooner.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+]);
+const EPOCH_PAYOUT = EdgeAddress.append(FIBRATION_EDGE, "EPOCH_PAYOUT");
+const EPOCH_WEBBING = EdgeAddress.append(FIBRATION_EDGE, "EPOCH_WEBBING");
+// Prefixes for seed edges.
+const SEED_IN = EdgeAddress.fromParts(["sourcecred", "core", "SEED_IN"]);
+const SEED_OUT = EdgeAddress.fromParts(["sourcecred", "core", "SEED_OUT"]);
+
+export type FibrationOptions = {|
+  // List of node prefixes for temporal fibration. A node with address
+  // `a` will be fibrated if `NodeAddress.hasPrefix(node, prefix)` for
```
decentralion, line 182: nit: `s/node/a` in the code example

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,579 @@
...
+  _scoringAddresses: Set<NodeAddressT>;
+
+  constructor(
+    nodes: Map<NodeAddressT, MarkovNode>,
+    edges: Map<MarkovEdgeAddressT, MarkovEdge>,
+    scoringAddresses: Set<NodeAddressT>
+  ) {
+    this._nodes = nodes;
+    this._edges = edges;
+    this._scoringAddresses = scoringAddresses;
```
decentralion, line 217: I really like that we take an explicit set of scoring addresses, rather than prefix matchers.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,605 @@
...
+
+    // Add graph nodes
+    const nwe = nodeWeightEvaluator(wg.weights);
+    for (const node of wg.graph.nodes()) {
+      const weight = nwe(node.address);
+      if (weight < 0) {
+        const name = NodeAddress.toString(node.address);
+        throw new Error(`Negative node weight for ${name}: ${weight}`);
+      }
+      addNode({
```
decentralion, line 299: maybe check that node is a scoring node implies that the weight is zero?

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,605 @@
...
+      }
+      const epochEndIndex = sortedIndex(timeBoundaries, edgeTimestampMs);
+      const epochStartIndex = epochEndIndex - 1;
+      const epochTimestampMs = timeBoundaries[epochStartIndex];
+      return epochNodeAddressToRaw({
+        type: "EPOCH_NODE",
+        owner: address,
+        epochStart: epochTimestampMs,
+      });
+    };
```
decentralion, line 407: Would be nice if we can get unit testing of this function in particular. Feels like if it were off-by-one our current test plan wouldn't catch it.

src/core/markovProcessGraph.js
decentralion: in retrospect, this seems a little pedantic, but still correct, so maybe worth following up on.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,605 @@
...
+      if (_edges.has(mae)) {
+        throw new Error("Edge conflict: " + mae);
+      }
+      const pr = edge.transitionProbability;
+      if (pr < 0 || pr > 1) {
+        const name = MarkovEdgeAddress.toString(mae);
+        throw new Error(`Invalid transition probability for ${name}: ${pr}`);
+      }
+      _edges.set(mae, edge);
+      _nodeOutMasses.set(edge.src, _nodeOutMasses.get(edge.src) || 0 + pr);
```
decentralion, line 281: Umm, I believe that addition is actually higher precedence than ||. so i think this code is wrong.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#:~:text=Operator%20precedence%20determines%20how%20operators,of%20operators%20with%20lower%20precedence.

src/core/markovProcessGraph.js
```diff
@@ -0,0 +1,605 @@
...
+      MapUtil.pushValue(inNeighbors, edge.dst, edge);
+    }
+
+    const chain = nodeOrder.map((addr) => {
+      const inEdges = NullUtil.orElse(inNeighbors.get(addr), []);
+      const inDegree = inEdges.length;
+      const neighbor = new Uint32Array(inDegree);
+      const weight = new Float64Array(inDegree);
+      inEdges.forEach((e, i) => {
+        // Note: We don't group-by src, so there may be multiple `j`
```
decentralion, line 558: It would be an efficiency improvement to collapse these together, right? Smaller/denser markov chain that is semantically equivalent. I dont think this happens much in practice so no change requested.

decentralion (approved):

PR #1849, opened 2020-06-09
Re-rasterize the png logos

decentralion:
In #1738, @wchargin updated the svg logo with a cleaner path structure.
This commit re-runs the rasterize script with the new svg, which
results in much cleaner pngs without aliasing.

Test plan: The new pngs look a lot cleaner.


wchargin:
Thanks!

PR #1853, opened 2020-06-12
Fork the frontend to build towards cli2 compat

decentralion:
The cli2 ("instance") system has a foundationally different assumption
about how the frontend works: rather than having a unified frontend that
abstracts over many separate SourceCred projects, we'll have a single
frontend entry per instance. This means we no longer need (for example)
to make project IDs available at build time.

Our frontend setup and server side rendering is pretty complex, so
rather than rebuild it from scratch, I'm going to fork it into an
independent copy and then change it to suit our needs.

To start here, I've duplicated the `src/homepage` directory into
`src/homepage2`, duplicated the webpack config to
`config/webpack.config.web2.js`, and duplicated the paths and
package.json scripts.

Test plan:
Run `yarn start2` and it will start an identical frontend, using the
duplicated directory. Run `yarn build2` and it will build a viable
frontend into the `build2` directory. `build2` is gitignored.


config/webpack.config.web2.js
```diff
@@ -0,0 +1,255 @@
+// @flow
+const express = require("express");
+/*::
+import type {
+  $Application as ExpressApp,
+  $Response as ExpressResponse,
+} from "express";
+*/
```
hammadj, line 8: nit: remove comment

config/webpack.config.web2.js
```diff
@@ -0,0 +1,255 @@
...
+      // We also include JSX as a common component filename extension to support
+      // some tools, although we do not recommend using it, see:
+      // https://github.com/facebookincubator/create-react-app/issues/290
+      // `web` extension prefixes have been added for better support
+      // for React Native Web.
+      extensions: [".web.js", ".mjs", ".js", ".json", ".web.jsx", ".jsx"],
+      alias: {
+        // Support React Native Web
+        // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
+        "react-native": "react-native-web",
```
hammadj, line 118: whoa where did rn come from. RN is my life lol

package.json
```diff
@@ -92,7 +92,9 @@
     "prettify": "prettier --write '**/*.js'",
     "check-pretty": "prettier --list-different '**/*.js'",
     "start": "NODE_ENV=development webpack-dev-server --config config/webpack.config.web.js",
+    "start2": "NODE_ENV=development webpack-dev-server --config config/webpack.config.web2.js",
```
hammadj, line 95: How long will we care about backwards compatibility? Do we want to be maintaining two versions?

src/homepage2/HomePage.js
```diff
@@ -0,0 +1,207 @@
+// @flow
+
+import React from "react";
+
+import type {Assets} from "../webutil/assets";
+import Link from "../webutil/Link";
+import {StyleSheet, css} from "aphrodite/no-important";
+
+export default class HomePage extends React.Component<{|+assets: Assets|}> {
```
hammadj, line 9: Do we really want to have all this stuff when there's the docs repo as the new homepage?

config/webpack.config.web2.js
decentralion: This is actually needed for flow checking -- this is a webpack config file, so it doesn't get run thru babel (since we integrate babel via webpack) so we needed to add the flow comments in comment-style syntax which is also valid javascript

config/webpack.config.web2.js
decentralion: I'm not sure, tbh. Note this is a copied file. :)

package.json
decentralion: We'll throw away the old version as soon as the new one can take over. The old stuff is all crap IMO.

PR #1854, opened 2020-06-12
Radically simplify homepage2

decentralion:
This drastically streamlines the new frontend entry directory.
All the logos, the nav bar, the pages, etc, are all gone.
Now there's just a landing page that reads "Under Construction".

Test plan: Run `yarn start2`, observe that there's just an empty landing
page that we'll rebuild from.


src/homepage2/routeData.js
```diff
@@ -26,37 +26,8 @@ type RouteDatum = {|
...
-      component: component,
-    },
-    title: "Inspection test for: " + name,
-    navTitle: null,
-  };
-}
-
 function makeRouteData(
-  projectIds /*: $ReadOnlyArray<string> */
+  _unused_projectIds /*: $ReadOnlyArray<string> */
```
hammadj, line 30: Why do we need to have this still?

decentralion: This gets cleaned in the next PR, my friend. :)

PR #1856, opened 2020-06-12
Configure frontend2 to load from instances

decentralion:
This modifies the frontend2 so that we can load real data from cli2
instances, including:
- the root `sourcecred.json` config
- files inside the `config/` directory
- files inside the `output/` directory

This works both for the dev server and the compiled output.

In the case of the dev server, it's now necessary to provide the path to
the cli2 instance you're developing against. You can pass this via the
`--instance` argument, as in `yarn start2 --instance=/path/`, or via
the `$SOURCECRED_DEV_INSTANCE` environment variable (recommended). If
neither is provided, or if that path doesn't look like a valid instance,
an error is thrown.

In the case of the built output, given a valid sc2 instance, you can set
it up via the following:

```
cp -r build2/favicon.png build2/index.html build2/static $INSTANCE
```

Then spin up a simple http server in the $INSTANCE.

You can look at console messages in the frontend to verify that the
instance is working (note it expects to see a discourse config file, not
a GitHub config file as was used in some previous examples).

Test plan:
Setup an example Discourse instance, and then turn on the server via
`yarn start2 --instance=PATH`,
`SOURCECRED_DEV_INSTANCE=path yarn start2`,
and by manually copying in the built outputs using the instructions
above.

In each case, you can load the homepage and check the console to see
that assets loaded successfully.

### DIFF
--- a/config/webpack.config.web2.js
+++ b/config/webpack.config.web2.js
@@ -6,8 +6,8 @@ import type {
   $Response as ExpressResponse,
 } from "express";
 */
-const os = require("os");
 const path = require("path");
+const fs = require("fs-extra");
 const webpack = require("webpack");
 const RemoveBuildDirectoryPlugin = require("./RemoveBuildDirectoryPlugin");
 const CopyPlugin = require("copy-webpack-plugin");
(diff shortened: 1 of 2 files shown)


config/webpack.config.web2.js
```diff
@@ -40,18 +40,55 @@ async function makeConfig(
...
+          // it's likely the developer has made a mistake and would find a quick failure helpful.
+          throw new Error(
+            `Provided instance directory ${developmentInstancePath} doesn't have a sourcecred.json file`
+          );
+        }
+
+        const configContents = fs.readFileSync(configPath);
+        const serveConfig = (_unused_req, res /*: ExpressResponse */) => {
+          res.status(200).send(configContents);
+        };
```
hammadj, line 71: @topocount Notice a lot of similar code to your PR (https://github.com/sourcecred/sourcecred/pull/1861) here, might be worth consolidating in the future

src/homepage2/HomePage.js
```diff
@@ -5,7 +5,23 @@ import React from "react";
...
 import {StyleSheet, css} from "aphrodite/no-important";
 
+async function loadAndReport(assets, path) {
+  const url = assets.resolve(path);
+  const response = await fetch(url);
+  if (!response.ok) {
+    console.error(path, response);
+  }
+  const json = await response.json();
+  console.log(path, json);
```
hammadj, line 15: Was this intentional, or left over from testing?

decentralion: This is intentional--see the test plan. We'll remove it once we actually have a frontend that displays data.

PR #1860, opened 2020-06-14
Create new initiative type for use with Editor

topocount:
This is the initial proposal for what a new initiative json file type
might look like.
This is very close to how the intiative editor service stores
intitiatives in `db.json`. No compatibility functions are submitted for
review as of yet, but compatibility might be achieved through use of
plugin-specific functions that derive urls from their respective node
addresses.

However, if a user-friendly interface such as the initiatives editor is
provided, compatibility is probably not necessary, since the
human-readability of the JSON database file can be less of a priority in
that case.

Test plan:
Beyond passing flow and linting tests, this addition is not yet handled
dynamically sounit tests aren't* included

### DIFF
--- a/src/plugins/initiatives/initiativeFile.js
+++ b/src/plugins/initiatives/initiativeFile.js
@@ -8,6 +8,7 @@ import {type InitiativeWeight, type InitiativeId, createId} from "./initiative";
 import {type InitiativesDirectory} from "./initiativesDirectory";
 import {type EdgeSpecJson} from "./edgeSpec";
 import {initiativeNodeType} from "./declaration";
+import {type NodeEntryJson} from "./nodeEntry";
 
 export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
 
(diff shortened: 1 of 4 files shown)


src/plugins/initiatives/initiativeFile.js
```diff
@@ -19,6 +20,28 @@ export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
...
+/**
+ * Stored inside a initiatives.json db file with other initiatives.
+ * presently the `id` prop is used as the key inside the db file
+ */
+export type InitiativeFileV100 = {|
+  id: string, // json file name right now. probably not needed if this is stored in side a json db file
+  // Compat info is usually stored within a header object, but in the db file parsing is more
+  // straightforward if it's folded inside the primary data object
+  type: string,
+  version: string,
```
decentralion, line 32: Can you explain why it's convenient to put it inside the primary data object? IMO it's a nice separation of concerns: once you know you have an InitiativeFileV100 you don't need a version or type info, because it's there in the type. So we include the type/version in the wrapper, but once you unwrap into a known type, you don't need that information in runtime any more.

src/plugins/initiatives/initiativeFile.js
```diff
@@ -19,6 +20,28 @@ export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
...
+  title: string,
+  timestampIso: TimestampISO,
```
decentralion, line 34: Let's use `TimestampMs` for consistency with all our other serialized timestamps, we were only using `TimestampISO` for human read-ability on the JSON files.

src/plugins/initiatives/initiativeFile.js
```diff
@@ -19,6 +20,28 @@ export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
...
+  weight: InitiativeWeight,
+  completed: boolean,
+  // should be user nodes: may need a a more specific type
+  champions: ?$ReadOnlyArray<NodeAddressT>,
+  dependencies: ?$ReadOnlyArray<NodeAddressT>,
+  // should be an activity node. Will be more difficult to specify this
+  references: ?$ReadOnlyArray<NodeAddressT>,
+  contributions: ?$ReadOnlyArray<NodeEntryJson>,
```
decentralion, line 42: Let's make them required (can just have an empty array if not set); it makes parsing the format a little simpler. We made them optional for ease of human-editability.

src/plugins/initiatives/initiativeFile.js
```diff
@@ -19,6 +20,28 @@ export const INITIATIVE_FILE_SUBTYPE = "INITIATIVE_FILE";
  */
 export type InitiativeFile = InitiativeFileV020;
 
+/**
+ * Stored inside a initiatives.json db file with other initiatives.
+ * presently the `id` prop is used as the key inside the db file
+ */
+export type InitiativeFileV100 = {|
```
decentralion, line 27: This type shouldn't really be called `InitiativeFile` because it no longer is describing a file, right? We're dropping the assumption that there's a 1:1 mapping between initiatives and files. So this could be called something like `InitiativeEntry` instead, and there'd be an `InitiativeStore` type which would correspond to the `db.json`.

src/plugins/initiatives/initiativeFile.js
topocount: I was using this in the frontend to determine how to parse files, originally. Will remove this.

topocount:
closing in favor of #1864

PR #1861, opened 2020-06-15
create admin panel service helper functions

topocount:
These functions are used by the admin panel node service to local files
and load them into memory.

`getFilePath` conducts a breadth-first-search in the supplied directory
for the file and returns the path to the directory containing the file.
This is used to set up static paths.

`loadFile` utilizes getFilePath and actually loads the file into memory.
This will be used by the conversion mondule to read distinct initiative
files and convert them before loading them into db.json

test plan
both functions are tested simultaneously in unit testing in order to
ensure they return a file matching the name, and error properly when
either the supplied directory or the supplied file name does not exist.

### DIFF
--- a/src/admin/localService/loadFile.js
+++ b/src/admin/localService/loadFile.js
@@ -1,0 +1,50 @@
+// @flow
+const fs = require("fs");
+const pathJoin = require("path").join;
+
+const readdir = fs.promises.readdir;
+
+/**
+ * load file into server memory
+ */
+async function loadFile(credDir: string, fileName: string): Promise<string> {
+  const filePath = await getFilePath(`${credDir}/`, fileName);
+  if (!filePath)
+    return Promise.reject(
+      new Error(`${fileName} not found. Please enter a full cred repo`)
+    );
+
+  return fs.promises.readFile(pathJoin(filePath, fileName), "utf8");
+}
+
+/**
+ * server utility function used to naively locate ancillary files needed in the frontend:
+ * weightedGraph.json
+ * project.json
+ * pluginDeclarations.json
+ */
+async function getFilePath(dir: string, targetFile: string): Promise<string> {
+  const files: Dirent[] = await ((readdir(dir, {
+    withFileTypes: true,
+  }): any): Promise<Dirent[]>); // Dirent is not currently in the builtin node flow-typed lib
+  if (files.map((f) => f.name).includes(targetFile)) return dir;
+  const search = files
+    .filter((f) => f.isDirectory())
+    .map(async (f) => {
+      return await getFilePath(pathJoin(dir, f.name), targetFile);
+    });
+  if (search.length > 0) {
+    const results: Array<string> = await Promise.all(search);
+    const result = results.find((e) => !!e) || "";
+    return result;
+  }
+  return "";
+}
+
+type Dirent = {|
+  name: string,
+  type: number,
+  isDirectory(): boolean,
+|};
+
+module.exports = {loadFile, getFilePath};
--- a/src/admin/localService/loadFile.test.js
+++ b/src/admin/localService/loadFile.test.js
@@ -1,0 +1,39 @@
+// @flow
+import {loadFile} from "./loadFile";
+import fs from "fs";
+
+describe("src/admin/localService/loadFile", () => {
+  beforeAll(async () => {
+    // Arrange
+    await fs.promises.mkdir("./test/dir", {recursive: true});
+    await fs.promises.writeFile("./test/dir/some-file.txt", "found me!");
+  });
+  it("finds and reads a file", async () => {
+    // Act
+    const filePath = await loadFile("test", "some-file.txt");
+    // Assert
+    expect(filePath).toEqual("found me!");
+  });
+  it("throws an error if no file is found in an existing directory", async () => {
+    // Act
+    return expect(loadFile("test", "missing-file.txt")).rejects.toThrow(
+      // Assert
+      "missing-file.txt not found. Please enter a full cred repo"
+    );
+  });
+  it("throws an error if the entered directory doesn't exist", async () => {
+    //arrange
+    expect.assertions(1);
+    // Act
+    return expect(
+      loadFile("missing-directory", "missing-file.txt")
+    ).rejects.toThrow(
+      "ENOENT: no such file or directory, scandir 'missing-directory/'"
+    );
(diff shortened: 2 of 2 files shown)


src/admin/localService/loadFile.js
```diff
@@ -0,0 +1,50 @@
...
+    .filter((f) => f.isDirectory())
+    .map(async (f) => {
+      return await getFilePath(pathJoin(dir, f.name), targetFile);
+    });
+  if (search.length > 0) {
+    const results: Array<string> = await Promise.all(search);
+    const result = results.find((e) => !!e) || "";
+    return result;
+  }
+  return "";
```
hammadj, line 41: Nit: return early, a bit easier to read. Also

```suggestion
  if (!search.length) return "";
  const results: Array<string> = await Promise.all(search);
  return results.find((e) => !!e) || "";
```

src/admin/localService/loadFile.test.js
```diff
@@ -0,0 +1,39 @@
...
+  beforeAll(async () => {
+    // Arrange
+    await fs.promises.mkdir("./test/dir", {recursive: true});
+    await fs.promises.writeFile("./test/dir/some-file.txt", "found me!");
+  });
+  it("finds and reads a file", async () => {
+    // Act
+    const filePath = await loadFile("test", "some-file.txt");
+    // Assert
+    expect(filePath).toEqual("found me!");
```
hammadj, line 15: Naming
```suggestion
    const contents = await loadFile("test", "some-file.txt");
    // Assert
    expect(contents).toEqual("found me!");
```

src/admin/localService/loadFile.test.js
```diff
@@ -0,0 +1,39 @@
...
+  it("throws an error if no file is found in an existing directory", async () => {
+    // Act
+    return expect(loadFile("test", "missing-file.txt")).rejects.toThrow(
+      // Assert
+      "missing-file.txt not found. Please enter a full cred repo"
+    );
+  });
+  it("throws an error if the entered directory doesn't exist", async () => {
+    //arrange
+    expect.assertions(1);
```
hammadj, line 26: what does this do exactly, I assume it ensures that a certain number of assertions are made? "arrange" isn't that helpful as a comment IMO, if anything it should just clarify what assertions is for

src/admin/localService/loadFile.test.js
```diff
@@ -0,0 +1,39 @@
...
+    return expect(
+      loadFile("missing-directory", "missing-file.txt")
+    ).rejects.toThrow(
+      "ENOENT: no such file or directory, scandir 'missing-directory/'"
+    );
+  });
+  afterAll(async () => {
+    await fs.promises.unlink("./test/dir/some-file.txt");
+    await fs.promises.rmdir("./test/dir");
+    return await fs.promises.rmdir("./test");
```
hammadj, line 37: Would be good to have these as constants at the top of the file, that way if someone changes the directory names in tests it will stay in sync here

src/admin/localService/loadFile.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type Dirent = {|
+  name: string,
+  type: number,
```
hammadj, line 46: Is this field used for anything? Noticed the `withFileTypes: true` option as well, but dont see anywhere that we check for it.

src/admin/localService/loadFile.js
```diff
@@ -0,0 +1,50 @@
...
+const readdir = fs.promises.readdir;
+
+/**
+ * load file into server memory
+ */
+async function loadFile(credDir: string, fileName: string): Promise<string> {
+  const filePath = await getFilePath(`${credDir}/`, fileName);
+  if (!filePath)
+    return Promise.reject(
+      new Error(`${fileName} not found. Please enter a full cred repo`)
```
hammadj, line 14: Might be good to have a check / error message for having the wrong cred directory vs a file not existing. Also, "Please enter a full cred repo" is a bit confusing, what does "full" mean in this context? Might be good to reword it.

src/admin/localService/loadFile.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+const fs = require("fs");
+const pathJoin = require("path").join;
+
+const readdir = fs.promises.readdir;
+
+/**
+ * load file into server memory
```
hammadj, line 8: A better description would be that it takes a directory and recursively traverses it until it finds a cetain filename, and then returns the contents of that file.

src/admin/localService/loadFile.test.js
topocount: > what does this do exactly, I assume it ensures that a certain number of assertions are made?

Yes, it ensures the test actually runs. Basically, if the assertion never executes due to some issue where the promise never resolves, we don't want a silent failure where jest just accepts the test as passing because nothing threw. I'm setting these for the two other tests for the sake of consistency.

> "arrange" isn't that helpful as a comment IMO, if anything it should just clarify what assertions is for

"arrange, act, assert (AAA)" is a simple organizing structure for setting up tests. I think setting the expected assertion quantity falls under the "arrange" step since it's testing instrumentation. I'm happy to remove these comments but I find them helpful.

src/admin/localService/loadFile.js
topocount: This would not be strictly necessary, except that this is the canonical shape of Node's `Dirent` class. The reason I pass in the `withFileTypes` flag is so that we have access to the `isDirectory method: https://github.com/topocount/sourcecred/blob/1e37a2714933e8c3ac0b5659250f2d07e85fa05c/src/admin/localService/loadFile.js#L35

Otherwise a simple string with the name of file or directory is returned. Since I'm not even technically declaring a class, I can remove it to avoid any confusion, but it's included in Node Dirents, so I figured I should keep it in there.

src/admin/localService/loadFile.js
topocount: I added some usage info to the server logic so users know what repo they're supposed to be using, and how to enter it. If a file doesn't exist in this case, it's pretty safe to assume that they're looking in the wrong directory from my observation, unless they've just never run any cred calculations before. not sure how we'd want to handle that caveat though, since I don't know what the instructions would be, or how to differentiate between a repo that has just never run calculations vs an incorrect repo.

topocount:
closing in favor of #1865

PR #1862, opened 2020-06-15
Admin panel server core

topocount:
This PR contains the core server logic and is meant as a follow-on to #1861
- [ ] merge https://github.com/sourcecred/sourcecred/pull/1861 first

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,11 +17,13 @@
     "d3-time-format": "^2.2.3",
     "deep-freeze": "^0.0.1",
     "express": "^4.17.1",
+    "flow-remove-types": "^2.127.0",
     "fs-extra": "^9.0.0",
     "globby": "^11.0.0",
     "history": "^3.3.0",
     "htmlparser2": "^4.1.0",
     "isomorphic-fetch": "^2.2.1",
+    "json-server": "^0.16.1",
     "json-stable-stringify": "^1.0.1",
     "lodash.clonedeep": "^4.5.0",
     "lodash.isequal": "^4.5.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -17,11 +17,13 @@
     "d3-time-format": "^2.2.3",
     "deep-freeze": "^0.0.1",
     "express": "^4.17.1",
+    "flow-remove-types": "^2.127.0",
```
hammadj, line 20: Where and how is this used?

src/admin/localService/loadFile.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+const fs = require("fs");
+const pathJoin = require("path").join;
+
+const readdir = fs.promises.readdir;
+
+/**
+ * load file into server memory
+ */
+async function loadFile(credDir: string, fileName: string): Promise<string> {
```
hammadj, line 10: This file is the same as the one in https://github.com/sourcecred/sourcecred/pull/1861/files, right? In that case, would be good to get that merged and rebase this on top. Once you have write access to the SC repo, next time this PR could just be merging into that other PR branch so we dont get duplicated diffs like this

package.json
```diff
@@ -139,6 +143,9 @@
       "mjs"
     ]
   },
+  "bin": {
+    "adminServe": "flow-node ./src/admin/localService/serve.js"
+  },
```
hammadj, line 148: curious, what's the reasoning for having this?

src/admin/localService/loadFile.js
decentralion: I sent @topocount an invite to join the sourcecred contributors team, which will give him write access

package.json
topocount: This hasn't been completely worked out yet, but the idea is  that, rather than running a development build of the admin panel, we could release sourceCred with a parcel-managed production build of the admin panel that is served directly from this node service

topocount:
> Should mention in the PR description that https://github.com/sourcecred/sourcecred/pull/1861/files should be merged first before this PR

Closing in favor of #1866 , which is directed towards it's immediate parent branch, simplifying the diff

PR #1864, opened 2020-06-16
New initiative format

topocount:
This Pull Request contains the implementation of the new Initiative types that will favor concise storage over human-readability. These initiative stores will be served into a web frontend, either via github or directly from disk, where users will be able to construct initiatives using actual graph and plugin data.

### DIFF
--- a/src/plugins/initiatives/initiativeEntry.js
+++ b/src/plugins/initiatives/initiativeEntry.js
@@ -2,11 +2,12 @@
 import {type TimestampMs} from "../../util/timestamp";
 import {type NodeAddressT} from "../../core/graph";
 import {type InitiativeWeight} from "./initiative";
-import {type NodeEntryJson} from "./nodeEntry";
+import {type NodeWeight} from "../../core/weights";
 
 /**
- * Stored inside a initiatives.json db file with other initiatives.
- * presently the `id` prop is used as the key inside the db file
+ * Stored inside a initiatives.json db file with other initiatives,
+ * modified and populated by a frontend.
+ * The `id` prop is used as the key inside the db file
  */
 export type InitiativeEntry = {|
   +id: string, // GUID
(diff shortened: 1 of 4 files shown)


src/plugins/initiatives/initiativeEntry.js
```diff
@@ -0,0 +1,31 @@
...
+export type InitiativeEntry = {|
+  +id: string, // GUID
+  +title: string,
+  +timestampMs: TimestampMs,
+  +weight: InitiativeWeight,
+  +completed: boolean,
+  // should be user nodes
+  +champions: $ReadOnlyArray<NodeAddressT>,
+  // other initiative nodes
+  +dependencies: $ReadOnlyArray<NodeAddressT>,
```
decentralion, line 20: (re comment): I think dependencies can be any nodes, not just initiatives nodes.

src/plugins/initiatives/initiativeEntry.js
```diff
@@ -0,0 +1,31 @@
...
+  // an activity node. Will be more difficult to specify this
```
decentralion, line 21: What do you have in mind when you say an activity node? why will it be more difficult to specify?

src/plugins/initiatives/initiativeEntry.js
```diff
@@ -0,0 +1,31 @@
...
+  +references: $ReadOnlyArray<NodeAddressT>,
+  +contributions: $ReadOnlyArray<NodeEntryJson>,
```
decentralion, line 23: The NodeEntryJson type uses url references, not node address references. Let's please add a new way of representing inline contributions, which does node address references.

decentralion: Also, let's have the new type be precise / exact (i.e. it will have every field) rather than making all the fields optional -- for the same reason as here, if we are not trying to ensure that the type is easy for manual human editing, then it is cleaner to just always have the fields.

src/plugins/initiatives/initiativeEntry.js
topocount: I just see references as non-user-like nodes, and I'm not sure there's a way to do this statically. Right now I"m filtering on user-types to create an array of "activity" nodes for this field.

decentralion:
I think this is obsoleted, to be replaced by the Creditor

PR #1866, opened 2020-06-17
create core admin server logic

topocount:
**NOTE** Please merge #1865 before this branch. This is directed towards that PR branch until it gets merged

This core server logic is pretty straightforward. It's more or less what
configuring and starting and express server looks like, only there's
some added sugar from the `json-server` package, including the rewriter
and the logging middlewares.

The configuration steps are as follows:
1. Prerequisite middlewares are installed, including the rewriter
2. static paths are configured for the needed files from the project
folder
3. If no db.json file is detected, one is intitialized
4. The json server router for that file is installed
5. The service is started.

Note that imports and compatibility shaping is descoped from this
commit. However, this service should work just fine with a fresh
initiatives plugin.

In the package.json file I installed a couple files in the "core"
dependencies as opposed to dev because I wanted this serive to be
accessible to individuals who might install sourcecred as a global
dependency and use it as a utility

Test Plan:
I did some manual testing to assure the middleware installation order
allowed everything to function as expected, but beyond that, it seems
testing might not be needed here, since this is mostly a configuration
of an express server

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,11 +17,13 @@
     "d3-time-format": "^2.2.3",
     "deep-freeze": "^0.0.1",
     "express": "^4.17.1",
+    "flow-remove-types": "^2.127.0",
     "fs-extra": "^9.0.0",
     "globby": "^11.0.0",
     "history": "^3.3.0",
     "htmlparser2": "^4.1.0",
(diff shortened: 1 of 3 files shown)


decentralion:
Closing in favor of #1869

PR #1867, opened 2020-06-17
add support for parsing compatible objects

decentralion:
This adds Combo parsing support to the compatible module. Now, rather
than writing `fromJSON` methods which implicitly take any, we can
instead write typesafe `parseJSON` methods which will parse compatible
headers, and then choose a version-appropriate parser.

Test plan: Added unit tests; `yarn test` passes.

### DIFF
--- a/src/util/compat.js
+++ b/src/util/compat.js
@@ -1,5 +1,6 @@
 // @flow
 
+import * as C from "./combo";
 export opaque type Compatible<T> = [CompatInfo, T];
 type CompatInfo = {|
   +type: string,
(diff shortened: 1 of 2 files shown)


src/util/compat.js
```diff
@@ -43,3 +44,34 @@ export function fromCompat<T>(
   }
   return result;
 }
+
+const headerParser = C.object({type: C.string, version: C.string});
+const wrappedParser = C.tuple([headerParser, C.raw]);
+
+export function compatibleParser<T>(
+  expectedType: string,
+  handlers: {[version: string]: C.Parser<T>}
```
wchargin, line 53: Should be read-only: `{+[version: string]: C.Parser<T>}`.

src/util/compat.test.js
```diff
@@ -72,6 +73,43 @@ describe("util/compat", () => {
...
+      );
+    });
+    it("throws on json with an invalid compatible header", () => {
+      expect(() =>
+        parseCompatible([{type: "foo", version: 3}, {}], "unused", {})
+      ).toThrow("unable to unwrap compatible");
+    });
+    it("throws on json with the wrong type", () => {
+      expect(() =>
+        parseCompatible([{type: "foo", version: "1.0.0"}, {}], "bar", {})
```
wchargin, line 92: Please use `toCompat` here instead to avoid breaking the opaque-type
abstraction boundary, which makes the test brittle. Likewise below.

src/util/compat.js
decentralion: Done.

src/util/compat.test.js
decentralion: Done.

src/util/compat.js
decentralion: Done.

PR #1868, opened 2020-06-17
cli2 common: add util to load json files

decentralion:
This adds two methods to cli2/common.js:
- loadJson, which loads a JSON file from disk and then parses it
- loadJsonWithDefault, which loads a JSON file from disk and parses it,
  or returns a default value if the file is not present.

Both methods are well documented and well tested.

Test plan: `yarn test`; see included unit tests which are thorough.

### DIFF
--- a/src/cli2/common.js
+++ b/src/cli2/common.js
@@ -5,6 +5,7 @@ import fs from "fs-extra";
 
 import type {PluginDirectoryContext} from "./cliPlugin";
 import {parse as parseConfig, type InstanceConfig} from "./instanceConfig";
+import * as C from "../util/combo";
 
 export async function loadInstanceConfig(
   baseDir: string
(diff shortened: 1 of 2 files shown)


src/cli2/common.js
```diff
@@ -59,3 +60,53 @@ export function pluginDirectoryContext(
     },
   };
 }
+
+/**
+ * Load and parse a JSON file from disk.
+ *
+ * If no file is present at that location, then an error is thrown.
```
wchargin, line 67: “If the file cannot be read, …”. (Also throws on EACCES, e.g.).

Likewise below.

src/cli2/common.js
```diff
@@ -59,3 +60,53 @@ export function pluginDirectoryContext(
...
+
+/**
+ * Load and parse a JSON file from disk, with a default to use if the file is
+ * not found.
+ *
+ * This is intended as a convenience for situations where the user may
+ * optionally provide configuration in a json file saved to disk.
+ *
+ * The default must be provided as a function that returns a default, to
+ * accomodation situations where the object may be mutable, or where constructing
```
wchargin, line 86: nit: accomodation → accommodate (part of speech, plus spelling)

src/cli2/common.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+describe("cli2/common", () => {
+  function tmpWithContents(contents: mixed) {
+    const name = tmp.tmpNameSync();
+    fs.writeFileSync(name, JSON.stringify(contents));
+    return name;
+  }
+  describe("loadJson / loadJsonWithDefault", () => {
+    const fooParser = C.object({foo: C.number});
+    const badPath = "not-a-real-path";
```
wchargin, line 16: This file *could* exist. Could you create a temp directory and put
`goodPath` and `badPath` in there so that the test is actually hermetic?

src/cli2/common.test.js
```diff
@@ -0,0 +1,57 @@
...
+      expect(await loadJsonWithDefault(f, fooParser, fooDefault)).toEqual(
+        fooInstance
+      );
+    });
+    it("loadJsonWithDefault loads default if file not present", async () => {
+      expect(await loadJsonWithDefault(badPath, fooParser, fooDefault)).toEqual(
+        fooDefault()
+      );
+    });
+    it("loadJsonWithDefault errors if parse fails", async () => {
```
wchargin, line 44: How about if the `JSON.parse` fails?

src/cli2/common.test.js
```diff
@@ -0,0 +1,57 @@
...
+      const f = tmpWithContents(barInstance);
+      const fail = async () =>
+        await loadJsonWithDefault(f, fooParser, fooDefault);
+      await expect(fail).rejects.toThrow("missing key");
+    });
+    it("loadJsonWithDefault errors if file loading fails for a non-ENOENT reason", async () => {
```
wchargin, line 50: Nice; thanks!

src/cli2/common.test.js
decentralion: Tested

src/cli2/common.test.js
decentralion: Did it for badPath

PR #1869, opened 2020-06-17
Add a scaffold Dashboard for @topocount

decentralion:
This PR adds a simple Dashboard component to the new frontend which
provides (in its state) the CredResult and the InstanceConfig.

The code is a bit raw and I'll want to incorporate proper typesafe
parsing later, but no need to do it right now.

Right now it's connected to our old router so we can find the instance
by running:

```
yarn start2 --instance=$PATH_TO_INSTANCE
```

And then navigate to `localhost:8080/dashboard`

I'm not particular that we keep this routing structure around long term,
but this should unblock you.

Test plan: Within a loaded cli2 instance, run `yarn start2`, navigate to
`/dashbaord`, and observe that the console shows the loaded state.


decentralion:
Superseded by #1870.

PR #1871, opened 2020-06-18
Add Address.fromRaw

decentralion:
This adds a fromRaw method to the Address module so that we may properly
parse serialized addresses. It has a type signature of `string =>
Address`, and throws an error if the string is not a valid address.

I basically copied the implementation out of the `assertValid` method. I
considered deduplicating them (i.e. having `assertValid` simply call
`fromRaw`), but `assertValid` had some extra logic around accepting a
prefix for the error message, and it felt simpler to simply copy+paste
rather than trying to wrap it.

Test plan:
I added unit tests; `yarn test` passes.

### DIFF
--- a/src/core/address.js
+++ b/src/core/address.js
@@ -5,7 +5,7 @@ import deepFreeze from "deep-freeze";
 
 import * as MapUtil from "../util/map";
 
-export interface AddressModule<Address> {
+export interface AddressModule<Address: string> {
   /**
    * Assert at runtime that the provided address is actually a valid
    * address of this kind, throwing an error if it is not. If `what` is
@@ -73,6 +73,19 @@ export interface AddressModule<Address> {
    * e.g., `toParts(["ban"])` is not a prefix of `toParts(["banana"])`.
    */
   hasPrefix(address: Address, prefix: Address): boolean;
+
+  /**
+   * Interpret the provided string as an Address.
+   *
+   * Addresses are natively stored as strings. This method verifies
+   * that the provided "raw" address is actually an Address, so that
+   * you can have a type-level assurance that a string is an Address.
+   *
+   * This is useful if e.g. you are loading serialized Addresses.
+   *
+   * Throws an error if the string is not a valid Address.
+   */
+  fromRaw(raw: string): Address;
 }
 
 export type Options = {|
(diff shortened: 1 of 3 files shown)


src/core/address.js
```diff
@@ -73,6 +73,13 @@ export interface AddressModule<Address> {
    * e.g., `toParts(["ban"])` is not a prefix of `toParts(["banana"])`.
    */
   hasPrefix(address: Address, prefix: Address): boolean;
+
+  /**
+   * Attempt to convert a raw string into an Address.
```
wchargin, line 78: Yep, okay. This is a little subtle, actually. Attempt to convert a raw
string into an address… how? In the same way that addresses are
naturally embedded as strings—but that embedding is not actually known
at the `AddressModule` level. Before now it didn’t really need to be: it
was important that (e.g.) `NodeAddressT` be a subtype of `string`, but
it wasn’t important that a generic `Address` be a subtype of `string`.
But now it is. So, could you apply this patch—

```diff
diff --git a/src/core/address.js b/src/core/address.js
index c80f7056..e97b71bc 100644
--- a/src/core/address.js
+++ b/src/core/address.js
@@ -5,7 +5,7 @@ import deepFreeze from "deep-freeze";
 
 import * as MapUtil from "../util/map";
 
-export interface AddressModule<Address> {
+export interface AddressModule<Address: string> {
   /**
    * Assert at runtime that the provided address is actually a valid
    * address of this kind, throwing an error if it is not. If `what` is
diff --git a/src/core/trie.js b/src/core/trie.js
index d8f9d7e9..ffe10e18 100644
--- a/src/core/trie.js
+++ b/src/core/trie.js
@@ -13,7 +13,7 @@ const EMPTY_ENTRY_SYMBOL = Symbol("EMPTY");
 
 type Entry<V> = {|+map: RecursiveMap<V>, value: V | typeof EMPTY_ENTRY_SYMBOL|};
 type RecursiveMap<V> = Map<string, Entry<V>>;
-class BaseTrie<K, V> {
+class BaseTrie<K: string, V> {
   addressModule: AddressModule<K>;
   entry: Entry<V>;
 
```

—and update the docs on `fromRaw` to clarify that this is just the
partial inverse of the canonical injection from `Address` to `string`?

decentralion: Done. Feel free to propose changes to the docstring.

PR #1872, opened 2020-06-18
combo: accept a key parser in `dict`

wchargin:
Summary:
An optional second parameter to `C.dict` can now be used to refine keys
to a subtype of string: e.g.,

    C.dict(C.number, C.fmap(C.string, NodeAddress.fromRaw))

is a `C.Parser<{[NodeAddressT]: number}>`. Before this change, `C.dict`
could only parse objects whose key type ranged literally over `string`.

Note that the argument order is `C.dict(valueParser, keyParser)`, not
the other way around. This is because `C.dict(valueParser)` is the more
general use case (you _need_ a value parser), and it’s simpler to have
the optional argument be at the end. (Analogously, the `forEach` methods
on `Array`s and `Map`s take a callback accepting `(value, key)`.)

Test Plan:
Unit tests included, retaining full coverage. Existing tests for the
experimental Discord plugin, which uses `C.dict`, also still pass.

wchargin-branch: combo-dict-key

### DIFF
--- a/src/util/combo.js
+++ b/src/util/combo.js
@@ -403,6 +403,16 @@ export function tuple<T: Iterable<Parser<mixed>>>(
   });
 }
 
+// Parser combinator for a dictionary whose keys are arbitrary strings.
+type PDictStringKey = <V>(Parser<V>) => Parser<{|[string]: V|}>;
+// Parser combinator for a dictionary whose keys are a specified subtype
+// of string.
+type PDictCustomKey = <V, K: string>(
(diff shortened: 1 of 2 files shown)


wchargin:
Yes, the extra map check and the extra indirection through `C.string`
are indeed not free. If we need to, we can split the implementations to
retain the old code for the unary case. I think that would be notably
harder to read (at least here all the nonobvious stuff happens at the
type level) so I’ll keep it as is for now.

PR #1873, opened 2020-06-18
Remove instructions from README

decentralion:
Our README.md has a hodge-podge of information about using SourceCred.
Much of it is about to go entirely out of date because we are killing
the old CLI and frontend.

As such, I'm cleaning out and greatly simplifying the README, so we
won't give our users inaccurate information. I didn't make an effort at
replacing the documentation, that will come afterwards once we're ready
to document the instance system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,32 +2,36 @@
 
 [![Build Status](https://circleci.com/gh/sourcecred/sourcecred.svg?style=svg)](https://circleci.com/gh/sourcecred/sourcecred)
 [![Discourse topics](https://img.shields.io/discourse/https/discourse.sourcecred.io/topics.svg)](https://discourse.sourcecred.io)
-[![Discord](https://img.shields.io/discord/453243919774253079.svg)](https://discord.gg/tsBTgc9)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,334 +2,34 @@
...
-  - Install [Yarn] (tested on v1.7.0).
-  - Create a [GitHub API token]. No special permissions are required.
-  - For macOS users: Ensure that your environment provides GNU
-    coreutils. [See this comment for details about what, how, and
-    why.][macos-gnu]
-
-[Node]: https://nodejs.org/en/
-[Yarn]: https://yarnpkg.com/lang/en/
-[GitHub API token]: https://github.com/settings/tokens
-[macos-gnu]: https://github.com/sourcecred/sourcecred/issues/698#issuecomment-417202213
```
hammadj, line 42: I feel like some of this stuff might still be relevant to new contributors and people exploring. Maybe we should move it to a temporary separate MD file and link to it from the main README for developers with a disclaimer that it might contain outdated info. Will likely also serve as a good reference when writing the new replacement docs as well.

decentralion: Good point. I kept around these dev instructions (they'll stay relevant) but left out the stuff for actually turning on the dev server, since that won't work anymore.

PR #1874, opened 2020-06-18
Kill the old homepage

decentralion:
This commit removes the old homepage entirely, to prepare for
removing react-router. This is needed to unblock work on the initiatives editor
(due to a dependency conflict between react-admin and our version of
react-router).

Because the homepage is gone, there's now no frontend included with
SourceCred. As such, we should merge #1873 alongside this one, so that
our README doesn't give any patently false information to our users.

Test plan: `yarn start` and `yarn build` are no
longer commands. (`yarn start2` and `yarn build2` will be renamed
later). `scripts/build_static_site.sh` is gone too. `yarn test --full` passes.


config/paths.js
```diff
@@ -12,11 +12,8 @@ module.exports = {
   root: appDirectory,
   dotenv: resolveApp(".env"),
   favicon: resolveApp("src/assets/logo/rasterized/logo_32.png"),
-  appBuild: resolveApp("build"),
   appBuild2: resolveApp("build2"),
```
hammadj, line 15: Nit, should we now rename appBuild2 to appBuild ?

decentralion: I'll do it in a followon.

decentralion: cf https://github.com/sourcecred/sourcecred/pull/1882

Issue #1807, opened 2020-05-27
Initiatives: handling users who aren't graphed 


Beanow:
The platform data plugins have mirroring strategies which don't include every user in the system. From their perspective this makes sense. Adding every Discord user and GitHub user that exists would be several millions of records bloating the mirror times, caches and graphs. For Discourse you might get away with this, but aren't currently adding every user to the graph.

The issue is, Initiatives might include URLs to accounts that do exist but weren't added to the graph by said plugin. For example because they never interacted with any of the repos or forums we're including for Cred scoring.

This causes a problem, because now we've already passed the mirroring stage (reference detection is done when creating the graph) and find out a potentially valid and existing account URL may be missing to complete the graph, but we haven't even mirrored it.

Currently reference detection also has a dependency on the mirror being populated, so it would be challenging to move it to the mirroring stage to surface these missing accounts. Even then it wouldn't make sense for the Initiatives plugin to implement mirroring of those accounts. So we may have to provide some form of lazy mirroring interface for other plugins too.

PR #1878, opened 2020-06-20
Hacky temporary fix for #1807

decentralion:
This modifies the cli1 pluginLoaders so that we construct a reference
detector that detects `@foo` references and treats them as pointers to
SourceCred identities. It violates the stated contract for reference
detectors, violates the intended contract for pluginLoaders, and is
untested. However, it will unblock MetaGame from actually using
SourceCred, and the areas where it adds technical debt (pluginLoaders
and to peoples serlialized v1 initiatives) are already deprecated and
being actively replaced. So I'm not worried about this tech debt.

Test plan: I went into an existing repo with initiatives and added an
`@identity` reference, and it did not print the "target not found" error
to console. @hammadj to test on MetaFam/TheSource before we merge this.

### DIFF
--- a/src/backend/pluginLoaders.js
+++ b/src/backend/pluginLoaders.js
@@ -18,6 +18,7 @@ import {
   type ReferenceDetector,
   CascadingReferenceDetector,
 } from "../core/references";
+import {identityAddress} from "../plugins/identity/identity";
 
 /**
  * A type combining all known plugin Loader interfaces.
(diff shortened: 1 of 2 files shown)


decentralion:
Note: If we decide to go with this, we'll need to cherry pick it onto a side branch to release, since master isn't yet usable.

decentralion:
Replaced by #2038.

PR #1879, opened 2020-06-21
flow: update libdefs for `express`, removing hacks

wchargin:
Summary:
This removes a SourceCred-specific hack because the upstream issue has
been fixed. Removing the hack is important because it uses a suppression
comment and thus hampers upgrading to latest Flow.

The libdef portion of this change was generated by removing the old file
and then running `flow-typed install express@4.17.1`.

Test Plan:
That `yarn flow` passes suffices.

wchargin-branch: libdefs-express


config/webpack.config.web.js
```diff
@@ -3,6 +3,7 @@ const express = require("express");
 /*::
 import type {
   $Application as ExpressApp,
+  $Request as ExpressRequest,
```
topocount, line 6: What is the significance of the `$` in this context? I've seen this used in a few different contexts, both involving and not involving JQuery. The use to denote JQuery objects is clear enough but I've found their use in reactive programming contexts and this context to be less clear. Is the answer simply to denote "mechanically generated code?"

wchargin: On our end, we don’t have a choice; despite the funny-looking name, this
is simply a normal type named `$Request` in the libdefs (which are
human-written, not machine-generated), so we have to import it by that
name. Presumably they chose to call it `$Request` rather than `Request`
because the latter is a window global in browser contexts, which could
be confusing. (IMHO it should instead be called `ExpressRequest` to make
this clear, or just `Request` and let people shadow.)

Yes, sometimes an `$` affix denotes a jQuery element or observable or
stream or something, as a kind of Hungarian notation; sometimes in Flow
libdefs it’s used as a path component delimiter as in `crypto$key` or
`express$Application`, as a kind of poor man’s namespacing, because
declared Flow types exist in a global namespace; and sometimes it’s used
in generated code because it’s a legal identifier character that can
often be excluded from input (e.g., our GraphQL Flow types). But none of
those is what’s happening here.

PR #1888, opened 2020-06-22
Setup React Admin frontend

hammadj:
This configures ReactAdmin to render as the frontend with a simple "fake data provider".
SSR issues need to be addressed before the UI will be visible

TestPlan: Ensure the frontend builds when running yarn start

PR #1889, opened 2020-06-24
implement initiative editor frontend

topocount:
This commit implements virtually the same frontend that's been demoed.

A couple styling considerations should be noted:
1. This frontend utilizes the out of the box materialUI dark theme
2. the `id` prop for Initiative entries and the `key` prop for
contributions are rendered as hidden. So for as I can tell right now,
	      this is necessary in order for the uuid value to be
	      generated for the key prop, but not for the id prop. This
	      is because initial values are not so easily populated for
	      the entire contribution because it's iterable. Some custom
	      work might be needed to implement this more elegantly and
	      prevent tampering with the DOM

The initiative shape will live in the App.js file until it is further
discussed as a part of Milestone 2. Then it will be moved to its own
file.

Test plan: Testing is manual. A mock state is provided and rendered.
All CRUD operations should be possible, and the state can be exported
as a json file, and saved to the file system.


src/ui/components/Initiatives.js
```diff
@@ -0,0 +1,260 @@
...
+      <Datagrid rowClick="edit">
+        <TextField label="Initiative Name" source="title" />
+        <BooleanField label="Completed" source="completed" />
+        <DateField label="Last Updated" source="timestampMs" />
+      </Datagrid>
+    </List>
+  );
+};
+
+export const InitiativeCreate = (scData: AppState) => (props: Object) => {
```
hammadj, line 93: Would be nice to split up the components into separate files so its a bit more digestible

src/ui/components/Initiatives.js
```diff
@@ -0,0 +1,260 @@
...
+    fakeLink.setAttribute("href", URL.createObjectURL(blob));
+    fakeLink.setAttribute("download", `initiatives.json`);
+    fakeLink.click();
+  }
+  fakeLink.remove();
+};
+
+const getPlainDescFromMd = ({description}) => removeMd(description);
+
+export const InitiativeList = (props: Object) => {
```
hammadj, line 81: Do we know what type this is supposed to be? I imagine there's probably some TS typings for the List component that we can just copy over?

src/ui/App.js
```diff
@@ -1,43 +1,135 @@
...
-      <Admin dataProvider={dataProvider} history={history}>
+      <Admin theme={theme} dataProvider={dataProvider} history={history}>
+        <Resource
+          name="initiatives"
+          list={InitiativeList}
+          create={InitiativeCreate(this.state)}
+          edit={InitiativeEdit(this.state)}
+        />
         <Resource name="people" list={ListGuesser} edit={EditGuesser} />
         <Resource name="comments" list={ListGuesser} edit={EditGuesser} />
```
hammadj, line 134: Should probably remove these Resources since they arent used

src/ui/App.js
```diff
@@ -1,43 +1,135 @@
...
-    {id: 0, title: "Hello, world!"},
-    {id: 1, title: "FooBar"},
-    {id: 2, title: "Bar foo"},
-  ],
-  comments: [
-    {id: 0, person_id: 0, author: "John Doe", body: "Sensational!"},
-    {id: 1, person_id: 0, author: "Jane Doe", body: "I agree"},
-  ],
-});
+function loadUsers(plugins, graph): Array<Object> {
```
hammadj, line 24: Should move the helper functions into a separate file from the components

src/ui/App.js
```diff
@@ -1,43 +1,135 @@
...
+const dataProvider = fakeDataProvider(fakeInitiatives, true);
 
 const history = createMemoryHistory();
 
-export default class App extends React.Component<{||}> {
+const theme = createMuiTheme({
+  palette: {
+    type: "dark",
+  },
+});
```
hammadj, line 76: Noticed that the text is hard to read in some of these fields, can we change the color to pink instead of blue easily?

image

src/ui/components/Initiatives.js
topocount: React Admin uses prop types for their components. I can copy them over and flow-ify them if you think it's worthwhile. 

We aren't really touching these at this point in the frontend lifecycle so it didn't seem worth it when I was typing these initially but it shouldn't be significant lift.

topocount:
> One thing I noticed, the timestampMs is just a date and not actually a timestamp, is that intentional? Also, initiatives have "id" and the inline contributions have "key", should they both just be "id"?
> 
> image

Good catch Hammad! the frontend was handling the ISO timestamps seemlessly and I didn't think to check again when I switched over to the epoch timestamps. Works now though

src/ui/App.js
```diff
@@ -1,45 +1,101 @@
...
+
+export type AppState = {|
+  graph: ?Graph,
+  plugins: ?(PluginDeclaration[]),
+  users: Node[],
+  activities: Node[],
+  loaded: boolean,
+|};
```
decentralion, line 53: Take a look at how I handled state in https://github.com/sourcecred/sourcecred/pull/1894: 

https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/ui/App.js#L8-L30
https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/ui/App.js#L33-L35

It's a much simpler way to track state IMO. We can factor that out into a util so that it's available for this frontend too.

decentralion: Also: rather than tracking the graph, plugins, etc, as separate state, you can just track the CredView. Then everything else is accessible:

https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/analysis/credView.js#L130-L132
https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/analysis/credView.js#L142-L144
https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/analysis/credView.js#L177-L185

Note that `CredNode` is a superset of `GraphNode`:
https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/analysis/credView.js#L37-L45

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,123 @@
...
+    timestampMs: Date.now(),
+    champions: [],
+    dependencies: [],
+    references: [],
+    contributions: [],
+    weight: {incomplete: 0, complete: 0},
+    completed: false,
+  };
+  const initiatives = scData.activities;
+  console.log("props: ", props);
```
decentralion, line 35: rm?

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,123 @@
...
+          label="Date"
+          source="timestampMs"
+        />
+        <NumberInput
+          label="Weight While Incomplete"
+          source="weight.incomplete"
+        />
+        <NumberInput label="Weight When Completed" source="weight.complete" />
+        <BooleanInput label="Completed" source="completed" />
+        {/**/}
```
decentralion, line 59: ?

src/ui/initiativeUtils.js
```diff
@@ -0,0 +1,116 @@
...
+    while (!nextUser.done) {
+      const user = {
+        ...nextUser.value,
+      };
+      users.push(user); // find a way to return users, possibly refactor
+      nextUser = userIterator.next();
+    }
+  });
+  return users;
+}
```
decentralion, line 89: just use `CredView.userNodes`.

src/ui/initiativeUtils.js
```diff
@@ -0,0 +1,116 @@
...
+
+export function loadActivities(
+  plugins: PluginDeclaration[],
+  graph: Graph
+): Array<Node> {
```
decentralion, line 94: This feels brittle to me. To start, can we leave this out and just have it show all the nodes in the graph, even users/likes/whatever? Then let's clean it up during the UI polish phase.

src/ui/initiativeUtils.js
decentralion: also, take a look at the implementation for an idea on how to implement basically the same function in a cleaner and more efficient fashion:

https://github.com/sourcecred/sourcecred/blob/219c4490b26f78d7eb7c1b716c11b37e00baae41/src/analysis/credView.js#L177-L185

topocount:
> For this to merge, it needs to not clobber the functional UI added by #1894 -- ie, running `yarn start` should bring you to the working cred explorer UI. While this is still under development, let's hide it behind a route, e.g. `localhost:8080/editor`.
@decentralion you expect this route to be configured in express for now, correct? I think configuring the routing to work with the current SSR config might be a bit involved if we want the editor to include the cred explorer anyway (as opposed to the cred explorer including a link to editor)

decentralion:
> @decentralion you expect this route to be configured in express for now, correct? I think configuring the routing to work with the current SSR config might be a bit involved if we want the editor to include the cred explorer anyway (as opposed to the cred explorer including a link to editor). I have no problem setting up the routing within SSR, just curious

> Edit On second configuring the route via SSR might make more sense since it'll be an SSR-rendered application anyhow. Curious to get your take

Let's integrate it in a way that works with SSR, so that it's a pattern we can actually use long-term, rather than just a dev hack.

src/ui/initiativeUtils.js
topocount: I've done as suggested, but the autocomplete dropdowns are now quite sluggish when using the makerdao cred repo. Looking forward to finding a more permanent solution

src/ui/App.js
topocount: thanks this was dead-easy to use

src/ui/mock/fakeInitiatives.js
```diff
@@ -0,0 +1,73 @@
...
+      completed: true,
+      title: "Dandelion Creates webpack Dev Service",
+    },
+    {
+      id: "95f29621-4761-42c9-815a-59e6c5217069",
+      timestampMs: 1592888400000,
+      champions: [],
+      dependencies: [],
+      references: [],
+      contributions: [
```
hammadj, line 39: Arent the dependencies, references and contributions supposed to be objects with the following shape:
```ts
export type EdgeSpec = {|
  +urls: $ReadOnlyArray<URL>,
  +entries: $ReadOnlyArray<NodeEntry>,
|};
```

Even if we arent using URLs, we should probably conform to the existing initiative format right? Otherwise they wont be working at all. This should probably be:
```suggestion
      dependencies: {},
      references: {},
      contributions: {
        entries: [
```

Or are we completely abandoning that old version and forcing upgrades to this format?

src/ui/initiativeUtils.js
```diff
@@ -0,0 +1,65 @@
...
+  // Defaults to an empty array.
+  +contributors: $ReadOnlyArray<NodeAddressT>,
+  // Timestamp of this node, but in ISO format as it's more human friendly.
+  +timestampMs: TimestampMs,
+  // Defaults to null.
+  +weight: NodeWeight,
+|};
+
+export const getPlainDescFromMd = ({description}: {description: string}) =>
+  removeMd((description: string));
```
hammadj, line 38: Don't think we need the type cast here
```suggestion
  removeMd(description);
```

src/ui/components/InitiativesEditor.js
```diff
@@ -0,0 +1,74 @@
...
+const history = createMemoryHistory();
+
+const theme = createMuiTheme({
+  palette: {
+    type: "dark",
+    primary: pink,
+  },
+});
+
+export default class App extends React.Component<Props, State> {
```
hammadj, line 28: naming nit:
```suggestion
export default class InitiativesEditor extends React.Component<Props, State> {
```

src/ui/components/InitiativesEditor.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+
+import React from "react";
+import {Admin, Resource, Loading} from "react-admin";
+import {
+  InitiativeList,
+  InitiativeCreate,
+  InitiativeEdit,
+} from "../components/Initiatives";
```
hammadj, line 9: Can shorten this import
```suggestion
} from "../Initiatives";
```

src/ui/components/InitiativesEditor.js
hammadj: Would be better to implement this as a functional component with hooks since class components are pretty much deprecated at this point:
```tsx
const InitiativesEditor = () => {
  const [loadResult, setLoadResult] = React.useState<LoadResult | null>(null);
  React.useEffect(() => {
    load().then(setLoadResult);
  }, []);

  if (!loadResult) {
    return (
      <Loading
        loadingPrimary="Fetching cred details..."
        loadingSecondary="Your patience is appreciated"
      />
    );
  }
  switch (loadResult.type) {
    case "FAILURE":
      return (
        <div>
          <h1>Load Failure</h1>
          <p>Check console for details.</p>
        </div>
      );
    case "SUCCESS":
      return (
        <Admin theme={theme} dataProvider={dataProvider} history={history}>
          <Resource
            name="initiatives"
            list={InitiativeList}
            create={InitiativeCreate(loadResult.credView)}
            edit={InitiativeEdit(loadResult.credView)}
          />
        </Admin>
      );
    default:
      throw new Error((loadResult.type: empty));
  }
};

export default InitiativesEditor;
```

src/ui/initiativeUtils.js
```diff
@@ -0,0 +1,65 @@
...
+export const dateFormatter = (t: TimestampMs): ?DateString => {
+  // v is a `Date` object
+  const v = new Date(t);
+  if (!(v instanceof Date) || isNaN(v)) return;
+  const pad = "00";
+  const yy = v.getFullYear().toString();
+  const mm = (v.getMonth() + 1).toString();
+  const dd = v.getDate().toString();
+  return `${yy}-${(pad + mm).slice(-2)}-${(pad + dd).slice(-2)}`;
+};
```
hammadj, line 65: Any reason we arent using the `src/util/timestamp` utils for formatting Dates? Also, react-admin date picker also has its own date parser/formatter, why arent we using the defaults from there?

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,118 @@
...
+} from "react-admin";
+import {
+  getPlainDescFromMd,
+  dateFormatter,
+  dateParser,
+} from "../../initiativeUtils";
+import {CredView} from "../../../analysis/credView";
+
+export const InitiativeCreate = (credView: CredView) => (props: Object) => {
+  const defaultInitiativeValues = {
```
hammadj, line 23: Should wrap this in React.useMemo so its not recreated on every render

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,118 @@
...
+    champions: [],
+    dependencies: [],
+    references: [],
+    contributions: [],
+    weight: {incomplete: 0, complete: 0},
+    completed: false,
+  };
+  return (
+    <Create title="Create New Initiative" {...props}>
+      <SimpleForm initialValues={defaultInitiativeValues}>
```
hammadj, line 35: This component is almost exactly the same as the InitiativeEdit component, except this has the defaultInitialValues. Should extract this inner SimpleForm component into a separate `InitiativeForm` component that takes an optional `initialValues` prop and then use that component for both InitiativeEdit and InitiativeCreate. Will make it easier to maintain that way

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,118 @@
...
+          optionValue="address"
+          optionText={getPlainDescFromMd}
+          label="Dependencies"
+          suggestionLimit={10}
+        />
+        <AutocompleteArrayInput
+          source="references"
+          allowDuplicates={false}
+          translateChoice={false}
+          choices={credView.nodes()}
```
hammadj, line 80: Should wrap this in React.useMemo so it doesn't recompute on every render:

```suggestion
          choices={React.useMemo(() => credView.nodes(), [credView])}
```

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,118 @@
...
+              parse={dateParser}
+              label="Date"
+              source="timestampMs"
+            />
+            <NumberInput label="Weight" source="weight" />
+            <AutocompleteArrayInput
+              source="contributors"
+              allowDuplicates={false}
+              translateChoice={false}
+              choices={credView.userNodes()}
```
hammadj, line 107: Should wrap this in React.useMemo so it doesn't recompute on every render:

```suggestion
          choices={React.useMemo(() => credView.userNodes(), [credView])}
```

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,118 @@
...
+  };
+  return (
+    <Create title="Create New Initiative" {...props}>
+      <SimpleForm initialValues={defaultInitiativeValues}>
+        <TextInput
+          source="id"
+          label="Initiative ID"
+          disabled
+          type="hidden"
+          style={{display: "none"}}
```
hammadj, line 41: Do we really want to hide it? Having it visible will allow us to debug and make sure its working correctly, can still have it disabled. If we do want to hide it, should just setup a css class that we can specify instead of having inline style objects

src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -0,0 +1,118 @@
...
+          label="References"
+          suggestionLimit={10}
+        />
+        <ArrayInput label="Contributions" source="contributions">
+          <SimpleFormIterator>
+            <TextInput
+              source="key"
+              label="Contribution Key"
+              disabled
+              initialValue={uuid()}
```
hammadj, line 92: Noticed that when I add multiple inline contributions it uses the same uuid for all of them. I realize this is a shortcoming of react-final-form-array / SimpleFormIterator, and that we likely wont be doing it like this with the proper new format, but will the graph still be functional without it? 

Issues for this on RA:
- https://github.com/marmelab/react-admin/issues/4805
- https://github.com/marmelab/react-admin/issues/4800

src/ui/components/Initiatives/InitiativeCreate.js
hammadj: addressed this in #1968

src/ui/components/Initiatives/InitiativeCreate.js
hammadj: addressed this in #1968

src/ui/components/Initiatives/InitiativeCreate.js
hammadj: addressed this in #1968

src/ui/components/Initiatives/InitiativeCreate.js
hammadj: addressed this in #1968

src/ui/components/Initiatives/InitiativeCreate.js
hammadj: addressed this in #1968

src/ui/components/InitiativesEditor.js
hammadj: addressed this in #1968

src/ui/components/InitiativesEditor.js
hammadj: addressed this in #1968

src/ui/mock/fakeInitiatives.js
topocount: @decentralion decided we should completely rewrite the underlying format to utilize node addresses.

src/ui/initiativeUtils.js
topocount: These are the react-admin-recommended methods for transposing dates to and from timestamps. The problem is that the date picker returns a stringified date that is not in ISO format.

decentralion:
Heya, I notice this PR isn't set to merge into master. What's the merge strategy here? Asking before I review in case it's going to be broken up into smaller pieces for review.

hammadj:
> Heya, I notice this PR isn't set to merge into master. What's the merge strategy here? Asking before I review in case it's going to be broken up into smaller pieces for review.

@decentralion This PR needs to be merged into master first: https://github.com/sourcecred/sourcecred/pull/1888

That one is doing the "react admin setup" and dependency updates etc to setup a solid "base" to build the initiative editor on. This PR is focused specifically on initiative editor implementation.

decentralion:
Looks like we have merge conflicts.

Would prefer to see this as (at least) 2 PRs:
- one which does all the integration so we can put an (empty) react-admin frontend in
- one which adds all the logic which is specific to initiatives

That said, I'm ok with merging this as-is on hammad's approval.

decentralion:
Closing as we've moved on from this plan.

PR #1890, opened 2020-06-25
Discord: Add refetch messages implementation

BrianLitwin:
This adds an option to refetch a given number of a channel's 
messages when fetching from the Discord server,
in order to capture updates.
    
Test Plan: unit tests added.

### DIFF
--- a/src/plugins/discord/mirror.js
+++ b/src/plugins/discord/mirror.js
@@ -10,9 +10,12 @@ import * as nullUtil from "../../util/null";
  * Mirrors data from the Discord API into a local sqlite db.
  */
 
+const REFETCH = 50;
+
 export async function fetchDiscord(
   sqliteMirror: SqliteMirror,
-  streams: DepaginatedFetcher
+  streams: DepaginatedFetcher,
+  refetchNMessages: number = REFETCH
 ) {
   for (const member of await streams.members()) {
     sqliteMirror.addMember(member);
@@ -20,8 +23,10 @@ export async function fetchDiscord(
 
   for (const channel of await streams.channels()) {
     sqliteMirror.addChannel(channel);
+    const endCursor: Snowflake =
+      sqliteMirror.nthMessageIdFromTail(channel.id, refetchNMessages) || "0";
 
-    for (const message of await streams.messages(channel.id)) {
+    for (const message of await streams.messages(channel.id, endCursor)) {
       sqliteMirror.addMessage(message);
 
       for (const emoji of message.reactionEmoji) {
--- a/src/plugins/discord/mirror.test.js
+++ b/src/plugins/discord/mirror.test.js
@@ -11,12 +11,14 @@ import {
   testMessage,
   customEmoji,
   testReaction,
+  buildNMessages,
 } from "./testUtils";
 
 describe("plugins/discord/mirror", () => {
   describe("updateMirror", () => {
     const activeChannelId = "12";
     const activeMessageid = "34";
+    const activeAuthorId = "56";
 
     function testMirror() {
       const db = new Database(":memory:");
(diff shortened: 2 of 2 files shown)


src/plugins/discord/mirror.js
```diff
@@ -10,18 +10,23 @@ import * as nullUtil from "../../util/null";
  * Mirrors data from the Discord API into a local sqlite db.
  */
 
+const REFETCH = 50;
```
decentralion, line 13: nit: DEFAULT_MESSAGES_TO_REFETCH

src/plugins/discord/mirror.test.js
```diff
@@ -33,11 +35,15 @@ describe("plugins/discord/mirror", () => {
...
         return [
-          testMessage(activeMessageid, channelId, authorId, reactionEmoji),
-          testMessage("2", channelId, authorId, []),
+          testMessage(
+            activeMessageid,
+            channelId,
+            activeAuthorId,
+            reactionEmoji
+          ),
+          testMessage("2", channelId, activeAuthorId, []),
```
decentralion, line 46: I recognize this hasn't changed since last time -- but it seems kind of weird that it returns message id 34 and then message id 2? Wouldn't the message ids be increasing in practice? Does this not matter?

src/plugins/discord/mirror.test.js
```diff
@@ -11,20 +11,22 @@ import {
...
   customEmoji,
   testReaction,
+  buildNMessages,
 } from "./testUtils";
 
 describe("plugins/discord/mirror", () => {
   describe("updateMirror", () => {
     const activeChannelId = "12";
     const activeMessageid = "34";
+    const activeAuthorId = "56";
```
decentralion, line 21: Why do we need an activeAuthorId now?

src/plugins/discord/mirror.js
```diff
@@ -10,18 +10,23 @@ import * as nullUtil from "../../util/null";
...
+
 export async function fetchDiscord(
   sqliteMirror: SqliteMirror,
-  streams: DepaginatedFetcher
+  streams: DepaginatedFetcher,
+  refetchNMessages: number = REFETCH
```
decentralion, line 18: It's not immediately obvious to me why we'd want to refetch messages, so it probably won't be obvious to future readers of this code either--could you please add a docstring or comment explaining why we have this parameter? is it based on the premise that recent messages may have had new reactions added?

if so, maybe instead of refetching a fixed # of messages, we should do it via timestamp? i.e. going back 50 messages might only go back a few minutes of history in a busy channel, maybe we should instead refetch all messages that are less than a day old? i recognize this could add more complexity.

decentralion: also, a nit: prefer `nMessagesToRefetch` vs `refetchNMessages`, usually when using `n` to indicate a numeric quantity, we put it at the front. having `refetchNMessages` makes me think that we are refetching a "NMessage", whatever that is 😝

BrianLitwin: My reasoning for fetching based on a fixed message number is that people will be more likely to update or react to posts that are visible, i.e. near the end of their feed, and that a fixed number would capture recent messages, but also older messages on less busy channels.

I'm envisioning a timestamp based approach as trying to capture the case where a user has 50+ new messages in a channel and they scroll through them and add reactions to some of them. I don't personally go through every unread message on a channel, so I'm projecting that on our users and suggesting it's a less probable case. But lmk if that's a flawed assumption. l don't think using a timestamp would add significant complexity, was actually doing that in an iteration with Beanow already.

decentralion: That makes sense, thanks for the explanation @BrianLitwin! Will merge.

PR #1899, opened 2020-06-26
Remove legacy explorer UIs

decentralion:
This commit removes (both) legacy explorer UIs, since they have been
replaced as of #1894.

Test plan: No code that is currently in use has been touched, `yarn test`
passes. Clean deletes are easy.


decentralion:
Goodbye, my sweet creations. You were tested too thoroughly for this world... from entropy you came and to entropy you shall return.

PR #1900, opened 2020-06-26
Remove the project type

decentralion:
This commit removes the `core/project.js` type. It was a sort of
"kitchen sink" config file where we just threw any random thing. This
was not a good approach to config, so now we let each plugin provide
it's own config format, and we no longer need (or want) the project
type.

I removed a test from the compatIO module, although we'll probably
remove that too ere long.

Test plan: `yarn test`.


src/core/project.js
```diff
@@ -1,166 +0,0 @@
...
-  |} | null,
-  +identities: $ReadOnlyArray<Identity>,
-|};
-
-const upgrades = {
-  "0.3.0": upgradeFrom030,
-  "0.3.1": upgradeFrom030,
-  "0.4.0": upgradeFrom040,
-  "0.5.0": upgradeFrom050,
-  "0.5.1": upgradeFrom051,
```
hammadj, line 165: Do we no longer want to have any sort of upgrade path from the legacy project.json to the newer format? At least something that will generate an error message letting the user know how they can migrate their config?

decentralion: The new setup is really different--IMO users are more likely to want to start fresh with a fork off the template repo. Trying to automate upgrades would be a lot of effort relative to our small number of users IMO.

PR #1905, opened 2020-06-28
cli: add `site` subcommand

wchargin:
Summary:
The `sourcecred site` command now sets up a static site for a cred
instance, which may be deployed to (e.g.) GitHub Pages or IPFS.

This implementation sets up a directory structure where the site is
contained in a `site/` subdirectory whose top-level entries are also
symbolically linked from the root of the instance. Thus:

```
my-instance/
    sourcecred.json
    config/
    output/
    site/
        index.html
        favicon.png
        static/
            static/js/...
    index.html -> site/index.html
    favicon.png -> site/favicon.png
    static -> site/static
```

The purpose of this is to enable the instance to be served directly as a
GitHub Pages site without the user having to navigate to a URL that
includes a `site/` path component.

It’s not a perfect solution. The top level of the instance is cluttered.
If we have any web pages other than the root directory, then they’ll
appear at top level, too. If we change the structure, we’ll need to
teach `sourcecred site` to clean up vestiges of old structures. We also
won’t be able to have any pages called `config` or `output` or `cache`
due to namespace collision. But we think that it’s at least a reasonable
stopgap, and doesn’t incur much conceptual overhead.

This isn’t tested or officially supported on Windows.

Paired with @decentralion.

Test Plan:
Run `yarn build` to build the frontend and backend, then `cd` into a
SourceCred instance and run `node "${OLDPWD}/bin/sourcecred.js site`.
Note that a `site/` directory is created, and when the static site is
served it works from either `site/` or the instance root. Note that this
all works when actually deployed to GitHub Pages as well.

wchargin-branch: cli-site


src/cli/site.js
```diff
@@ -0,0 +1,42 @@
...
+async function lnsf(src: string, dst: string): Promise<void> {
+  try {
+    await fs.symlink(src, dst);
+  } catch (e) {
+    if (e.code !== "EEXIST") {
+      throw e;
+    }
+  }
+  await fs.unlink(dst);
+  return fs.symlink(src, dst);
```
decentralion, line 39: nit: i would write `await fs.symlink(src, dst)` and then let it implicitly return.
semantically the same, looks more consistent (imo less likely to confuse someone with a weak grasp of promises)

wchargin: Sure, done.

PR #1915, opened 2020-07-03
Add a ledger/user type

decentralion:
This adds a user type to the ledger. Each user is tracked by a UUID, and
has a human-specified and renameable username. Each user also
corresponds to a node in the graph, and has a list of aliase addresses
that should be merged into it using `graph.contractNodes`.

It's important that we track this in the ledger because the ledger needs
to reconcile when aliases are added or removed.

Test plan: Unit tests added. Quite a simple module.


decentralion:
> I like the idea of using node addresses for more things instead of stuff like github/. Less middleware we need to write and maintain and makes it easier for plugins as well. Can always improve the UX by building a UI on top of the data instead of trying to make the data more human-readable.

Agreed. I'm going to add an aliasing system in a future PR, but it's going to have extremely isolated scope--aliases will only be used as a way to identify users in the config.json files, but won't be used elsewhere or in any internally generated data foramts.

src/ledger/user.js
```diff
@@ -0,0 +1,105 @@
...
+
+export type UserId = Uuid;
+export type User = {|
+  // UUID, assigned when the user is created.
+  +id: UserId,
+  +name: Username,
+  // Every other node in the graph that this user corresponds to.
+  // Does not include the user's "own" address, i.e. the result
+  // of calling (userAddress(user.id)).
+  +aliases: $ReadOnlyArray<NodeAddressT>,
```
Beanow, line 52: Late comment, but why not implement this as a Set?
It seems order isn't important, and duplicates don't make sense.

src/ledger/user.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Throws an error if the username is invalid.
+ */
+export function usernameFromString(username: string): Username {
+  const re = new RegExp(USERNAME_PATTERN);
+  const match = re.exec(username);
+  if (match == null) {
+    throw new Error(`invalid username: ${username}`);
+  }
+  return match[1];
```
wchargin, line 81: Should precompile this regex and just use

```javascript
  const match = username.match(_USERNAME_REGEXP);
  if (match == null) {
    throw new Error(`invalid username: ${username}`);
  }
  return match[1];
```

You only need the `exec` form when you care about multiple matches
(rather than one or more capturing groups), and it’s significantly more
expensive as well as more complicated, and this function will be called
many times when parsing a config.

src/ledger/user.js
```diff
@@ -0,0 +1,105 @@
...
+  type NodeAddressT,
+  NodeAddress,
+  type Node as GraphNode,
+} from "../core/graph";
+
+/**
+ * We validate usernames using GitHub-esque rules.
+ */
+export opaque type Username: string = string;
+const USERNAME_PATTERN = /^@?([A-Za-z0-9-_]+)$/;
```
wchargin, line 42: The `@` sign isn’t really part of the username, is it? It’s part of a
username *reference* in some free-form text contexts. But, for instance,
on the main page of this repo GitHub says “decentralion committed … 11
hours ago”, and that contributors include “decentralion” and “wchargin”
and “Beanow”, and that this pull request was opened by “decentralion”—no
`@`-signs anywhere.

I also think that this is confusing because `usernameFromString` accepts
non-canonical inputs and is not injective. As written, we could have a
config file of usernames like

```json
{
  "users": [
    {"username": "@decentralion", "id": "...", "aliases": []},
    {"username": "wchargin", "id": "...", "aliases": []}
  ]
}
```

which would be properly parsed even though the `username` forms are
inconsistent. This makes things like sorting a list of usernames harder.
It also means that logical identity on usernames is not the same as
physical identity, which is unfortunate.

Can we remove the `@` sign here and consistently just use the bare
username?

src/ledger/user.test.js
```diff
@@ -0,0 +1,53 @@
...
+    expect(node.timestampMs).toEqual(null);
+  });
+  describe("usernameFromString", () => {
+    it("fails on invalid usernames", () => {
+      const bad = ["With Space", "With.Period", "A/Slash", ""];
+      for (const b of bad) {
+        expect(() => usernameFromString(b)).toThrowError("invalid username");
+      }
+    });
+    it("succeeds on valid usernames", () => {
```
wchargin, line 46: `@`-stripping behavior not tested, but as above I recommend removing it,
anyway.

src/ledger/user.js
wchargin, line 42: We should not allow underscores in usernames; they’re not valid in DNS
labels, which precludes some useful behavior (same reason that GitHub
disallows underscores).

src/ledger/user.js
decentralion: Good question. I thought about this, but if using a set, we'd can no longer serialize users via `stringify(user)`, we'd have to have the `User` type and `UserSerialized` with converters between. Didn't seem like the Set added enough to justify the added complexity.

decentralion:
@wchargin, sorry for pre-empting your review -- I will add an update PR to fix the underscore and @ issues you surfaced.

src/ledger/user.js
```diff
@@ -0,0 +1,105 @@
...
+  +name: Username,
+  // Every other node in the graph that this user corresponds to.
+  // Does not include the user's "own" address, i.e. the result
+  // of calling (userAddress(user.id)).
+  +aliases: $ReadOnlyArray<NodeAddressT>,
+|};
+
+// It's not in the typical [owner, name] format because it isn't provided by a plugin.
+// Instead, it's a raw type owned by SourceCred project.
+export const USER_PREFIX = NodeAddress.fromParts(["sourcecred", "USER"]);
```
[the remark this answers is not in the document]
decentralion: Done (in followon)

src/ledger/user.js
decentralion: Done

src/ledger/user.js
decentralion: Done

decentralion: Done

src/ledger/user.js
Beanow: True, one way of this conversion can be handled by the parser, but `stringify` wouldn't unless it implemented a `toJSON` method. Fortunately those are easier to validate as we have the JsonObject we can set as return type now.

I'll leave it as a suggestion, but I suspect we'll end up with a serialization layer either way. Event types will probably include a fair bit of parsing logic. Serializing sets by comparison don't seem hard.

PR #1916, opened 2020-07-03
Ledger tracks user identities

decentralion:
This starts work on the Ledger module, which will accumulate instance
state for user identity updates, and Grain distribution.

A key requirement is that we durably store every state change, and are
able to replay it. This way we can have an auditable history of Grain
distributions, of when users' aliases changed, etc. (We need to keep the
user alias changes alongside the Grain distribution history, because
when users link or unlink aliases, we need to update the lifetime
distributions for each address accordingly.)

For now, we just handle user identity state, although we will enhance
the module with Grain distribution history in a future commit.

From an implementation perspective, the Ledger provides an imperative
API that is similar in feel to the Graph APIs. Under the hood, every
mutating API call constructs an "action", which is added to the action
log, and then processes that action. This way we can serialize the
history of actions and replay them later. I've taken care to ensure that
the ledger can't get into an inconsistent state, i.e. any action which
fails will fail before modifying the ledger state. This property is
tested.

Test plan: There are a lot of edge cases to handle, e.g. handling users'
"innate" addresses, ensuring that addresses and usernames are unlinked
when a user is removed, and so forth. Unit tests are quite
comprehensive.

### DIFF
--- a/src/ledger/ledger.js
+++ b/src/ledger/ledger.js
@@ -8,17 +8,24 @@
  * every action that's happened in the ledger, so that we can audit the ledger
  * state to ensure its integrity.
  */
-import deepEqual from "lodash.isequal";
-import {type UserId, type User, type Username, userAddress} from "./user";
+import {
+  type UserId,
+  type User,
+  type Username,
+  userAddress,
+  usernameFromString,
+} from "./user";
 import {type NodeAddressT, NodeAddress} from "../core/graph";
+import {type TimestampMs} from "../util/timestamp";
+import * as NullUtil from "../util/null";
+import {random as randomUuid} from "../util/uuid";
 
 /**
  * The ledger for storing identity changes and (eventually) Grain distributions.
  *
  * The following API methods change the ledger state:
- * - `addUser`
+ * - `createUser`
  * - `renameUser`
- * - `removeUser`
  * - `addAlias`
  * - `removeAlias`
  *
@@ -26,6 +33,11 @@ import {type NodeAddressT, NodeAddress} from "../core/graph";
  * the ledger's action log. The ledger state may be serialized by saving the
  * action log, and then reconstructed by replaying the action log. The
  * corresponding methods are `actionLog` and `Ledger.fromActionLog`.
+ *
+ * None of these methods are idempotent, since they all modify the Ledger state
+ * on success by adding a new action to the log. Therefore, they will all fail
+ * if they would not cause any change to the ledger's logical state, so as to
+ * prevent the ledger from permanently accumulating no-op clutter in the log.
  *
  * It's important that any API method that fails (e.g. trying to add a
  * conflicting user) fails without mutating the ledger state; this way we avoid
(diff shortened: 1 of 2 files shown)


decentralion:
> Before I dig into this, if we want to have identity changes accounted for in the graph, wondering if we also need to have weight changes accounted for. It would definitely be necessary if/when we want to allow for people to have a certain configuration for a certain time window and then apply different rules to a different time window. Would it make sense to use the ledger for that in the future?

No, I don't think the ledger needs to account for changing configuration over time. Reasoning is twofold:
1. The scope for the ledger is stuff where "if we don't have a consistent history for this, the system is broken / vulnerable to attacks or double spends". If we don't have a consistent history for when distributions or transfers occured, or for when we linked and unlinked accounts, ppl can attack the Grain distribution. So it needs to be in the ledger. However, retroactive changes in Cred are allowed and don't threaten the integrity of Grain distribution, so those don't need to be in the ledger.
2. We haven't come up with a general model for configuration, and it's likely to involve arbitrary plugin-specific configuration. Putting this in the ledger would create coupling between the ledger and plugins that would be extremely complicated and lead to lots of bugs, especially when plugins update and change their configuration formats. The data going into the ledger needs to be minimal so that we can confidently support the whole history.

src/ledger/ledger.js
```diff
@@ -0,0 +1,344 @@
...
+  _addUser(user: User) {
+    const existingUser = this._users.get(user.id);
+    if (existingUser != null) {
+      if (deepEqual(existingUser, user)) {
+        // No-op; This user already exists in the ledger.
+        return;
+      } else {
+        throw new Error(`addUser: conflicting user with id ${user.id}`);
+      }
+    }
```
Beanow, line 104: This seems fairly awkward to me and an example of why the 1:1 relation between API and events doesn't improve things. The UserId is it's unique reference, the other fields are state. So this is like a partial idempotence? Where a User in the same state as the existing one is silently ignored, but the same semantical user in a divergent state fails because this method wasn't supposed to check for differences while it could.

Abandoning the 1:1 mapping, this could actually be made much simpler and idempotent. In pseudo:

```js
trackUser(user: User): $ReadOnlyArray<Event> {
  // if user does not exist
    // return UserCreated event

  // const changes = [];

  // if username is different
    // add UsernameChanged to changes

  // if aliases are different
    // add AliasesChanged to changes

  // return changes
}
```

An existing user in the same state would be ignored (empty array of events). Changes would be reflected by change events. New users would have a create event.

src/ledger/ledger.js
```diff
@@ -0,0 +1,344 @@
...
+   * already taken, or if the user's innate address is already taken, or if any
+   * of the user's aliases are already taken.
+   */
+  addUser(user: User): Ledger {
+    return this._act({type: "ADD_USER", user: user});
+  }
+  _addUser(user: User) {
+    const existingUser = this._users.get(user.id);
+    if (existingUser != null) {
+      if (deepEqual(existingUser, user)) {
```
Beanow, line 98: Because the User type currently has an array of aliases, I think deep equality would give incorrect results.

```js
// Identical but returns false.
deepEqual(
  {id: "ABC123", name: "Beanow", aliases: ["github/Beanow", "discourse/Beanow"]},
  {id: "ABC123", name: "Beanow", aliases: ["discourse/Beanow", "github/Beanow"]},
);
```

src/ledger/ledger.js
```diff
@@ -0,0 +1,344 @@
...
+ *
+ * This is an opaque type; clients must not modify the log, since they
+ * could put it in an inconsistent state.
+ */
+export opaque type LedgerLog = $ReadOnlyArray<Action>;
+
+/**
+ * The Actions are used to store the history of Ledger changes.
+ */
+type Action = AddUser | RenameUser | RemoveUser | AddAlias | RemoveAlias;
```
Beanow, line 320: Oh, something I missed in previous review.
I would say adding timestamps and versions to every event is highly advisable.

Timestamps are not something you can easily reproduce and very commonly necessary.

Versioning is the only way to change the model after release (as you'll have to support it forever), so having a version type like `+version: 1 | 2 | 3` is cleaner than `+version?: 2 | 3` where undefined is an implicit v1.

Beanow:
Here's an example of what would also make sense to me in terms of events: https://github.com/sourcecred/sourcecred/commit/10b8e97c4f16f81861fd697faa0fce2c2a89d452

src/ledger/ledger.js
```diff
@@ -0,0 +1,344 @@
...
+|};
+type RenameUser = {|
+  +type: "RENAME_USER",
+  +userId: UserId,
+  +newName: Username,
+|};
+type RemoveUser = {|
+  +type: "REMOVE_USER",
+  +userId: UserId,
+|};
```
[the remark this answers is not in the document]
decentralion: I think of it as giving people autonomy--contributors should have the right to delete their account. Although these might not be the right semantics, since once the account is deleted, their aliases (if any) will go back to collecting Grain independently. So perhaps it should be something that marks the user deleted and prevents them from showing up in results or earning more Grain, but keeps lets them keep hold of their addresses.

src/ledger/ledger.js
```diff
@@ -0,0 +1,344 @@
...
+type AddAlias = {|
+  +type: "ADD_ALIAS",
+  +userId: UserId,
+  +alias: NodeAddressT,
+|};
+type RemoveAlias = {|
+  +type: "REMOVE_ALIAS",
+  +userId: UserId,
+  +alias: NodeAddressT,
+|};
```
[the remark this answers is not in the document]
decentralion: Good point.

src/ledger/ledger.js
decentralion: Good point.

src/ledger/ledger.js
decentralion: Good points. I discussed adding timestamps with @wchargin and we came out weakly on the side of not enforcing that ledger entries have timestamps, but I can't remember why. Seeing as I can't remember why; I'm actually pretty inclined to put them back in, I agree that I can imagine a lot of situations where we'll want them.

src/ledger/ledger.js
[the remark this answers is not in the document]
decentralion: I decided to remove `removeUser` for now. Semantics aren't clear enough, and we don't need it yet.

src/ledger/ledger.js
```diff
@@ -0,0 +1,344 @@
...
+
+/**
+ * The Actions are used to store the history of Ledger changes.
+ */
+type Action = AddUser | RenameUser | RemoveUser | AddAlias | RemoveAlias;
+
+type AddUser = {|
+  +type: "ADD_USER",
+  +user: User,
+|};
```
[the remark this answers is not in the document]
decentralion: I changed it to just record the creation of a user identity.

src/ledger/ledger.js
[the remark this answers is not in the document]
decentralion: Actually--having one-change-per-event is a nice feature. We're going to have to augment the RemoveAlias type with additional info on how to split up the historical allocation and the current Grain balance, so better to keep one-change-per-action.

src/ledger/ledger.js
decentralion: Yeah, it's awkward, I changed this so that the client doesn't create the Users.

src/ledger/ledger.js
decentralion: Added the timestamps and versions.

decentralion:
@Beanow I gave a lot of thought to your structural suggestions, but I've decided to keep the implementation more-or-less as is. You raise some interesting conceptual points, but the bottom line is that this approach is much easier to write, grok, and maintain than a more complex multi-system based approach.

It's true that there's a possibility of bugs where the class's state gets desync'd from what's in the log, but we can avoid that through being careful writing the code (update methods must not fail at any point after they've mutated state; this is now documented in comments), and in unit testing (every failure is tested to ensure ledger log wasn't mutated). This module has 100% coverage, so I'm pretty confident in the comprehensiveness of the testing.

That said, your review led to a number of improvements in the design, namely:
- remove `removeUser`
- replace `addUser` with simpler `createUser`
- add versions and timestamps

So, thanks for the review!

Beanow:
Respecting your choice, though I have a different opinion on the reasoning.

---

> we can avoid that through being careful writing the code

Famous last words of every C programmer that ever wrote security vulnerabilities.
We have Flow and eslint because compiler checked rules tend to hold up better than relying on humans to do hard things.

I think writing underlying abstractions once and having a typed API for it that will catch common mistakes at compile time, is more robust than needing to reason about multiple pieces of mutable state you need to manually sync for every new method.

---

For the worse-is-better reference, I think you're using easy and simple interchangeably.
If you define simple as: (objectively) single-fold; a lack of entanglement.
And easy as: (subjectively) familiar; low difficulty.

Reading the post with that lens:
- MIT style:
  * It is more important for the **interface** to be **simple** than the implementation.
  * Incorrectness is not allowed.
- worse-is-better style:
  * It is more important for the **implementation** to be **simple** than the interface.
  * It is slightly better to be simple than correct.

There's no mention of easy here. Which is nice because that means they're objective terms!
As I mentioned, I believe the current design has a lot of entanglement and is objectively complex.
And it focuses on the (MIT style) simplicity of the _interface_.
> I'm not a fan of the one-stop-shop architecture of this class, which implements a lot of things:
> - The event log
> - Current state model
> - State queries
> - Event replaying
> - Commands (called actions here)

Sure, my alternative still implements all of these things. However they're for the most part split into one solution per item on that list and later composed. Each individual component is simple or low complexity and has a more complex interface, but an interface Flow can enforce more rules of.

That doesn't mean my approach is easier. If you're unfamiliar with Commands, Projections and the like from the CQRS / Event Sourcing patterns, they're not going to seem like primitives to you in the way you would see `map` and `for..of` as easy building blocks (not needing to care about their implementation complexity).

PR #1918, opened 2020-07-03
Modify Discord to mint Cred for every emoji

decentralion:
This changes the Discord plugin so that, by default, it adds every
Reaction to the graph, so they will all mint Cred.

If custom reaction weights are added, they will act as multiplicative
modifiers on the Cred minting.

Test plan: I've tested this in production, by deploying SourceCred's own
Cred from this patch. The Discord plugin has no unit tests.

### DIFF
--- a/src/plugins/experimental-discord/createGraph.js
+++ b/src/plugins/experimental-discord/createGraph.js
@@ -204,18 +204,15 @@ export function createGraph(
   for (const channel of channels) {
     const messages = repo.messages(channel.id);
     for (const message of messages) {
-      const hasMentions = message.mentions.length > 0;
-      if (!hasMentions && message.reactionEmoji.length === 0) continue;
+      if (message.mentions.length === 0 && message.reactionEmoji.length === 0)
+        continue;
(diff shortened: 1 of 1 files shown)


src/plugins/experimental-discord/createGraph.js
```diff
@@ -226,15 +223,17 @@ export function createGraph(
           continue;
         }
 
-        hasWeightedEmoji = true;
         const node = reactionNode(reaction, message.timestampMs, guild);
-        wg.weights.nodeWeights.set(node.address, reactionWeight);
+        if (reactionWeight != null) {
```
hammadj, line 227: nit: strict equality check to be more explicit on the intent? (!==)

decentralion: Across the codebase we use `==` when checking null, because we don't care about the difference between null and undefined. In this case, using `!==` checking would actually induce a bug, since reactionWeight is undefined when it is unset.

Our eslint config blocks `==` checking everywhere except for on null/undefined.

PR #1920, opened 2020-07-05
Publish SourceCred with executable binary

decentralion:
This updates our build so that we include a shebang at the top of
`bin/sourcecred.js`, mark it executable, and make it the binary entry
point for the package.

As a result, dependents can install the `sourcecred` command via npm and
conveniently use it. Also, local invocation no longer needs to be
preceded with `node`, i.e. `./bin/sourcecred.js ARGS` is sufficient.

Test plan:

Verify that `yarn test --full` passes; and that
`./bin/sourcecred.js --version` prints the version.

Also, see sourcecred/example-instance#1, which uses a prerelease package
published from this commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "sourcecred",
-  "version": "0.7.0-beta-1",
+  "description": "a tool for communities to measure and reward value creation",
+  "homepage": "https://sourcecred.io",
+  "repository": "github:sourcecred/sourcecred",
+  "version": "0.7.0-beta-3",
   "private": false,
   "dependencies": {
     "aphrodite": "^2.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sourcecred",
-  "version": "0.7.0-beta-1",
+  "version": "0.7.0-beta-3",
```
hammadj, line 3: should also add in description, homepage, and repository so it shows up on the NPM page:
```suggestion
  "version": "0.7.0-beta-3",
  "description": "A tool for communities to measure and reward value creation",
  "homepage": "https://sourcecred.io",
  "repository": "github:sourcecred/sourcecred",
```

decentralion: Done, thanks. Did change the description, in the (near) future it won't just be the binary but also the API.

PR #1924, opened 2020-07-05
cli: add `./data` directory

decentralion:
Presently, instances have a `config` directory for storing
user-generated configs, and an `output` directory which has the data
generated by SourceCred. IMO, `output` should have the property that you
can always blow it away and regenerate it. However, we need some data
that is persistent but not user-maintained; a great example is the
ledger. Therefore, I am adding a data directory.

The first use of this directory is to store the plugins' graphs. We want
to keep each plugin's graph so that users can re-compute Cred scores
even if they are missing credentials to load any individual plugin;
since we have that plugin's graph, we can re-run the scoring pipeline,
and potentially even update the graphs for the other plugins.

In the longer term, I intend to refactor the CliPlugins so that the
`load` step writes data, and the `graph` step reads from config and data
but not from cache. That way, it will be possible to change plugins'
config in ways that modify the Graph (e.g. changing trust levels) and
then regenerate the Graph accordingly, without ever needing a load step.

Test plan: Run it on an example instance, check that the graphs are
written to the data dir, check that you can `rm -r output` and
immediately run `sourcecred score` without needing to regenerate graphs.


decentralion:
cc @wchargin

decentralion:
Update: I'm not sure if I want to go ahead with this.

Question: Should the ledger records be in `config` or in `data`?

Should we have an `output` directory? What does "output" imply exactly? Maybe we should just have `config`, `data`, and `cache`.

PR #1931, opened 2020-07-06
add JsonLog for serializing logs of JSON objects

decentralion:
This commit adds the JsonLog module, which has a JsonLog class for
managing an append-only log consisting entirely of JSON objects.

In addition to this basic functionality, it provides a toString() method
which creates a string in which each log item is on its own line. This
is a much better format for storing an append-only log inside of a Git
respository, as we will avoid churning the whole file's diff when adding
additional records.

Acknowledgements to @wchargin for suggesting the seralization format,
and to @Beanow for writing [another version] of this module.

[another version]: https://github.com/sourcecred/sourcecred/commit/42809547e5d08777b9f1cd1286687566ab669d86

Test plan:
Unit tests added; `yarn test` passes.


decentralion:
One thing I'm not sure about is whether this should have COMPAT-style headers as the first non-comment entry in the log.

I think it's not needed: since we are using Combo.Parser to parse the elements, it's impossible to not realize that you're reading the wrong type of log (so the "safety" from checking the stated type isn't needed), and the versioning should occur at the level of the log entries, not at the level of the log itself.

But open to being persuaded otherwise.

Beanow:
I think it's worth checking for https://github.com/ndjson/ndjson-spec compliance.
This would make 3rd party scripts a lot easier, as they'd have access to libraries in multiple languages.

src/util/jsonLog.js
```diff
@@ -0,0 +1,89 @@
...
+ * at runtime.
+ */
+export class JsonLog<T: C.JsonObject> {
+  _items: T[];
+
+  constructor() {
+    this._items = [];
+  }
+
+  append(items: $ReadOnlyArray<T>): JsonLog<T> {
```
Beanow, line 36: In my alt implementation I ended up switching this argument to `Iterable<T>`. They make commands much easier to write as you can conditionally yield events, rather than do conditional array building.

src/util/jsonLog.js
```diff
@@ -0,0 +1,89 @@
...
+ * `stringify` on the individual JSON objects. We support // comment syntax,
+ * and the toString and writeJsonLog methods allow you to provide prefix
+ * comments; we recommend putting prefix comments like
+ * ["AUTOGENERATED FILE", "do not manually edit"].
+ *
+ * The JsonLog is always parsed using a Combo.Parser, which ensures type safety
+ * at runtime.
+ */
+export class JsonLog<T: C.JsonObject> {
+  _items: T[];
```
Beanow, line 30: Think this should be either:
- `_items: $ReadOnlyArray<T>` so we'll replace the array reference rather than mutate it.
- `+_items: T[]` so we'll mutate it rather than replace the reference.

src/util/jsonLog.js
```diff
@@ -0,0 +1,89 @@
...
+ * Instead, we store each item on its own line in the output format, using
+ * `stringify` on the individual JSON objects. We support // comment syntax,
+ * and the toString and writeJsonLog methods allow you to provide prefix
+ * comments; we recommend putting prefix comments like
+ * ["AUTOGENERATED FILE", "do not manually edit"].
+ *
+ * The JsonLog is always parsed using a Combo.Parser, which ensures type safety
+ * at runtime.
+ */
+export class JsonLog<T: C.JsonObject> {
```
Beanow, line 29: `<T: C.JsonObject>` has the limitation that events can't have a custom `toJSON` such as for a `Set`, while they will require a custom parser per event type. Have a hunch we can group together serializing and parsing per event in a comprehensive way.

src/util/jsonLog.js
```diff
@@ -0,0 +1,89 @@
...
+  constructor() {
+    this._items = [];
+  }
+
+  append(items: $ReadOnlyArray<T>): JsonLog<T> {
+    this._items = [...this._items, ...items];
+    return this;
+  }
+
+  items(): $ReadOnlyArray<T> {
```
Beanow, line 41: Suggest using `values(): Iterator<T>` to be similar to the built-in classes and more memory friendly.
Map.prototype.values, Set.prototype.values, Array.prototype.values, etc.
We most likely don't need to copy the array as we do that on append instead.

src/util/jsonLog.js
```diff
@@ -0,0 +1,89 @@
...
+    // Make a copy, guarding against unsafe mutation, and avoiding potential
+    // confusion when the client adds more items to the log after requesting
+    // the items.
+    return this._items.slice();
+  }
+
+  toString(prefixComments: ?$ReadOnlyArray<string>): string {
+    const stringified = this._items.map((x) => stringify(x)).join("\n");
```
Beanow, line 49: This will surely cause memory problems when the log has a substantial size. Which I think would defeat the purpose of having ndjson. If we're not considering a memory bottleneck, then why not write the entire log as a single JSON array?

src/util/jsonLog.js
decentralion: I've talked with @wchargin about making a dual to Combo.Parser that would bundle serialization in too. Which I'd really like. He says it's not trivial but should be possible, but I didn't follow all the details.

If we need to support objects like sets, we could make a version of JsonLog where we don't assume T is a C.JsonObject, but in the `toString` method, you need to provide a serializer that maps `T => JsonObject`. Should be easy to add if/when we actually need it.

src/util/jsonLog.js
decentralion: I'm not worried about in-memory size, since the size of RAM on a users machine is much, much larger than the max file size limitation on GitHub (~gigabytes vs ~100MB), we will need to switch to a very different solution before memory becomes an issue. However, if we write the log as a JSON array, we'll get bad repository bloat on the Git side, because appending a single event would force the ledger to be an entirely new blob. However, if we use ndjson, then [Git packfiles](https://git-scm.com/book/en/v2/Git-Internals-Packfiles) will deduplicate most of the log for us, across commits.

src/util/jsonLog.js
decentralion: Changed.

src/util/jsonLog.js
decentralion: Good point, switched to readonly-mutable.

src/util/jsonLog.js
decentralion: Done

src/util/jsonLog.js
Beanow: I think it should be a concern moving forward though.

For dogfooding this may not be our first bottleneck, but the "git hosting" constraints can be avoided switching to Git LFS, IPFS or just keeping on regular filesystems. Today we're not able to run SC on larger projects as it will crash due to memory constraints. 16GB ram processes are the upper limit (regardless of actual ram + swap available) and I've hit it for tests with larger projects.

Streaming ndjson would be my recommendation, but the point remains. We can get the git optimizations with plain JSON. If we're writing our own serializer, might as well wrap it in `[\n` + `\n]\n` and write `,\n` at the end of each row to have plain old JSON on easy to compress rows.

decentralion: Regarding memory--the part of the system that hits the memory bottleneck (serializing huge graphs and cred results) has size `O(num weeks * num nodes)`. This log is being used for data that has size `O(num weeks * num users)`, which is at least 2 orders of magnitude smaller. So I'm not worried about memory pressure on this module right now; when/if we do find a bottleneck, we can improve the implementation without needing to change the output format.

(Also--this ledger data is wholly owned by SourceCred, if we need to upgrade it in the future we can still do that.)

I do like the idea of just writing plain JSON rather than ndjson. It's a neat trick where do still get the clean diffs, and we don't have the complexity of doing something that isn't regular JSON. So I changed it to your implementation.

Beanow: > (Also--this ledger data is wholly owned by SourceCred, if we need to upgrade it in the future we can still do that.)

I wrote a small essay on that before, but in my experience this isn't how upgrading works with event logs. https://github.com/sourcecred/sourcecred/pull/1697#discussion_r409020751
> For better or worse you will be supporting these event models until the end of time. I disagree with @decentralion here, with regards to "private data" and upgrade paths. These events are your source of truth and updating them after the fact is a breach of contract. Even re-encoding the same data with the same meaning into a new format should be frowned upon. If this were on-chain data, this would rightfully be made infeasible to do.

However switching from JSON arrays to NDJSON is something I think can be done, because they're both encodings of `Array<T>` that are easy to prove are equivalent. Plus backward compatibility would not be a big deal.

wchargin: For posterity/context, this change was reversed in #2075.

PR #1936, opened 2020-07-07
Allow inclusion of Grain in JSON objects

decentralion:
We store Grain as BigInts, and we would really like to be able to write
types that include Grain values, and treat them like JSON objects. That
would allow us to store events that include Grain values directly into
logs that are persisted to disk, and not need to worry with processing
each value for serialization and deserialization.

However, we have a problem, in that BigInts do not have a native
`toJSON` method. Therefore, this commit modifies the Grain module so
that on module load time, it overrides the BigInt prototype to include a
`toJSON` method which outputs the string representation, which is an
injection. Then, we provide a `grainParser` which will correctly parse
either raw BigInts, or stringified BigInts.

The net result is that you can stringify object containing Grain values
to your hearts content, and parse them using the provided parser, and
never need to think about the hacks that went into this! Brilliant!

The only downsides are:
1. We are modifying a global prototype (but the modification is so
reasonable that the MDN docs propose it), and
2. From the perspective of the parser, any logical Grain value has two
equivalent representations (string and BigInt)

Frankly, I think this is worth it and would happily sign my name onto
this hack.

Test plan: Unit tests included, check them out.

### DIFF
--- a/src/grain/grain.js
+++ b/src/grain/grain.js
@@ -1,5 +1,7 @@
 // @flow
 
+import * as C from "../util/combo";
+
 /**
  * This module contains the types for tracking Grain, which is the native
  * project-specific, cred-linked token created in SourceCred instances. In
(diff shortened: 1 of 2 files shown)


decentralion:
Discussed with @wchargin; he persuaded me that modifying the BigInt prototype is a Bad Thing, especially since we do expect others to depend on SourceCred as a library. Replaced with #1938.

PR #1937, opened 2020-07-07
WIP - Update Relative History Routing to V4 API

topocount:
The history V4 api has several changes:
- no `getCurrentLocation` method. The location state is accessed directly. (but only modified through hooks)
- private methods are no longer stored in the history object. Methods from V3 that aren't meant to be accessed publicly have been removed from the history context, and thus those methods have been removed here.
- the addtion of a `block` method, which prevents path changes within the react app when called

One of the trickier changes was exposing the delegate location state in a way that still conforms to the API. Canonical history implementations maintain their own state, rather than relying on nested history's state. Using a proxy handler, `relativeHistory` is able to handle top-level get requests without maintaining any of it's own state.

test plan: Unit tests have been updated to accomodate these changes and, ensure that the new endpoints are behaving as expected. Manual integration testing is also conducted using the [example instance](https://github.com/sourcecred/example-instance)

and executing the following shell commands:
```bash
# pick either this python server:
alias serve="python -m SimpleHttpServer 6060"
# or this javascript server:
yarn global add serve
alias serve="serve -l 6060" 
# then test each of these cases:
cd example-instance && serve  # then localhost:6060 should show a working frontend
cd .. && serve # then localhost:6060/example-instance should show a working frontend
```

## Remaining Tasks
- [x] add `action` prop
- [x] add `length` prop
- [x] add `location` prop
- [x] add `index` prop
- [x] add `entries` prop
- [x] add `block` method
- [x] polish comments to add new details
- [x] split changes into semantically atomic commits
- [x] write a helpful PR description
- [ ] conduct integration tests

### DIFF
--- a/flow-typed/npm/history_v4.x.x.js
+++ b/flow-typed/npm/history_v4.x.x.js
@@ -28,8 +28,8 @@ declare module "history/createBrowserHistory" {
     goBack(): void,
     goForward(): void,
     listen: Function,
-    block(message: string): typeof Unblock,
-    block((location: BrowserLocation, action: Action) => string): typeof Unblock,
+    block(message: ?string): typeof Unblock,
+    block((location: BrowserLocation, action: Action) => ?string): typeof Unblock,
   }
 
   declare type BrowserHistoryOpts = {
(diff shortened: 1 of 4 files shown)


topocount:
#1889 works with variably-sized paths thanks to https://github.com/sourcecred/sourcecred/pull/1889/commits/8d7c713aeeeed0581d612f244a791039ff1f5476. RelativeHistory is no longer needed.

PR #1942, opened 2020-07-08
Support $ReadOnlyMap

decentralion:
This modifies the Grain allocation module so that it uses
`$ReadOnlyMap`s. As a necessary and happy byproduct, we also update the
Map util module.


decentralion:
cc @wchargin -- this is nice.

PR #1946, opened 2020-07-08
ledger: add acounts and grain balances

decentralion:
This commit updates the Ledger so that we now support tracking and
updating users' Grain balances, and lifetime paid amounts.

We don't yet add any events which actually provide Grain, so it's
impossible to use the public API to get any non-empty balances in the
Ledger. Instead, we just add the underlying logic for tracking Grain
amounts. Most notably, we handle the edge case where a user is
un-linking an alias, and we need to ensure that we can record
that the alias has already been paid, so that attackers
can't use un-linking to double-dip against the BALANCED payout strategy.

Test plan: Comprehensive unit tests added; `yarn test` passes; coverage
still at 100%.

### DIFF
--- a/src/ledger/ledger.js
+++ b/src/ledger/ledger.js
@@ -19,6 +19,20 @@ import {type NodeAddressT, NodeAddress} from "../core/graph";
 import {type TimestampMs} from "../util/timestamp";
 import * as NullUtil from "../util/null";
 import {random as randomUuid} from "../util/uuid";
+import * as G from "./grain";
+
+/**
+ * An GrainAccount is an address that is present in the Ledger
+ * and has some associated Grain.
+ */
+type MutableGrainAccount = {|
+  +address: NodeAddressT,
+  +userId: UserId | null,
+  balance: G.Grain,
+  paid: G.Grain,
+|};
+
+export type GrainAccount = $ReadOnly<MutableGrainAccount>;
 
 /**
  * The ledger for storing identity changes and (eventually) Grain distributions.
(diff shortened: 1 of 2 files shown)


src/ledger/ledger.js
```diff
@@ -245,30 +287,62 @@ export class Ledger {
...
     };
+    const addr = userAddress(userId);
+    const aliasAccount = this._unsafeGetAccount(alias);
+    const userAccount = this._unsafeGetAccount(addr);
 
     // State mutations! Method must not fail past this comment.
     this._users.set(userId, updatedUser);
+    // Transfer alias's balance and paid to the canonical account
+    userAccount.paid = G.add(userAccount.paid, aliasAccount.paid);
+    userAccount.balance = G.add(userAccount.balance, aliasAccount.balance);
```
topocount, line 298: I have some concerns about not logging these transfers in the ledger. It'd definitely make returning maliciously captured funds through fraudulent alias accumulation more straightforward, since we can refer to a specific ledger entry, in addition to any ill-gotten grain through proportional cred accumulation

decentralion:
Thanks for the suggestion @topocount. You raise an excellent point, for auditability it's much better if we have an explicit balance transfer event rather than doing it implicitly. I've implemented your suggestion here: https://github.com/sourcecred/sourcecred/pull/1969 and wound up discovering a nasty bug in the process. So, double win.

Regarding proportional returns: We currently don't make any attempt to transfer Grain back from a user to an unlinked account. We can estimate how much SHOULD get transferred back, it's the exact same amount that we are setting as that alias's lifetime paid Grain. However, there's no guarantee that such a transfer back is actually possible; the user may have transferred the Grain away already. Since the ledger must not fail, it must not try to do that reversal transaction.

Note what it does do is move lifetime paid metadata to the alias. Otherwise, we imagine the following situation (imagine we have the policy of "Always pay users so they receive 100 Grain per Cred").

1. Attacker has raw account with address FOO. FOO has 1 Cred and has been paid 100 Grain
2. Attacker adds FOO to new user account User1. User1 has 1 Cred and has already been paid 100 Grain
2. Attacker disconnects FOO from user account User1. User1 has 0 Cred and has been paid 100 Grain. FOO now has 1 Cred but has been paid 0 (because the paid amount "stuck" to User1). FOO now gets paid 100 Grain
3. Attacker laughs their way to the Grain bank

Does that all make sense?

PR #1949, opened 2020-07-09
Discord: remove unused interface

BrianLitwin:
This Interface wasn't being used for the test
purpose it was originally intended for.

Test plan:

Yarn test passes

### DIFF
--- a/src/plugins/discord/fetcher.js
+++ b/src/plugins/discord/fetcher.js
@@ -9,7 +9,7 @@ type FetchEndpoint = (endpoint: string) => Promise<any>;
  * Provide the Guild ID to fetch against, and the 'limit'
  * parameter when fetching GuildMembers, Messages, and Reactions.
  */
-type FetchOptions = {|
+export type FetchOptions = {|
   membersLimit: number,
   messagesLimit: number,
   reactionsLimit: number,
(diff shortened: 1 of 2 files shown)


src/plugins/discord/fetcher.js
```diff
@@ -131,8 +108,7 @@ export class Fetcher implements DiscordFetcher {
       mentions: (x.mentions || []).map((user) => user.id),
     }));
     const hasNextPage = results.length === messagesLimit;
-    const endCursor =
-      response.length > 0 ? response[response.length - 1].id : null;
+    const endCursor = response.length > 0 ? response[0].id : null;
```
decentralion, line 111: I didn't notice this when I reviewed this PR, oops. @BrianLitwin can you explain what this change is?

src/plugins/discord/fetcher.js
```diff
@@ -148,7 +124,7 @@ export class Fetcher implements DiscordFetcher {
     const endpoint = `/channels/${channel}/messages/${message}/reactions/${emojiRef}?after=${after}&limit=${reactionsLimit}`;
     const response = await this._fetch(endpoint);
     const results = response.map((x) => ({
-      emoji: x.emoji,
+      emoji: emoji,
```
decentralion, line 127: @BrianLitwin can you explain this change?

src/plugins/discord/fetcher.js
BrianLitwin: The responses from the Discord Server (per our snapshots) come back in descending order (last message first, `5 4 3 2 1`).  I need to update the other methods that rely on endCursor for pagination; I made this change on messages because that was the method we used to test pagination with against the snapshots.

src/plugins/discord/fetcher.js
BrianLitwin: Yeah. We are asking for the users that responded with a particular emoji here, and the emoji isn't present in the response, but is passed in as a parameter to the Discord fetch. 

I need to follow up and update the unit tests in few places to reflect this. Thanks for the catch.

PR #1950, opened 2020-07-09
End-to-end implementation of Discord server fetch

BrianLitwin:
End-to-end implementation of Discord server fetch plan.

Test plan:

Uses data from our test Discord instance to populate
the responses from several mock requests (channels,
members, messages, reactions) and uses snapshots to
test that this data is saved correctly in local db mirror.

Open to replacing the snapshot tests with something more
granular, like property tests. Motivation for using snapshots
is that theses snapshots should not change frequently.

### DIFF
--- a/src/plugins/discord/fetch.js
+++ b/src/plugins/discord/fetch.js
@@ -1,0 +1,103 @@
+// @flow
+
+import fetch from "isomorphic-fetch";
+
+import {DepaginatedFetcher, getPages} from "./mirror";
+import {DiscordFetcher, type FetchOptions} from "./fetcher";
+import * as Model from "./models";
+import {type Snowflake} from "./models";
+import Database from "better-sqlite3";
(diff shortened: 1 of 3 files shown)


src/plugins/discord/fetch.js
```diff
@@ -0,0 +1,103 @@
...
+  return token;
+}
+
+export function buildDiscordFetch(fetch: typeof fetch) {
+  return async (endpoint: string) => {
+    const fetchOptions = {
+      method: "GET",
+      headers: {
+        Accept: "application/json",
+        Authorization: `Bot ${getTokenFromEnv()}`,
```
decentralion, line 65: Let's not bind in the token here as an impl detail--instead, let's have the caller pass the token in. In the future we may get tokens from different places than the environment (e.g. a config store) so we don't want to hard code this assumption.

PR #1951, opened 2020-07-09
Discord: fetch error-handling 

BrianLitwin:
Adds a method to handle network request errors.
    
Test plan: One unit test added to cover the general
case of a request returning a non-200 or 201 status.
There is room to be more granular if the need arises.

### DIFF
--- a/src/plugins/discord/fetch.js
+++ b/src/plugins/discord/fetch.js
@@ -56,19 +56,40 @@ function getTokenFromEnv(): string {
   return token;
 }
 
-export function buildDiscordFetch(fetch: typeof fetch) {
+function handleErrors(response: Response) {
+  // TODO: handle 50001
+  if (response.status === 429) {
+    // TODO: Retry strategy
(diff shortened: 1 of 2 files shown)


src/plugins/discord/fetch.js
```diff
@@ -56,19 +56,40 @@ function getTokenFromEnv(): string {
   return token;
 }
 
-export function buildDiscordFetch(fetch: typeof fetch) {
+function handleErrors(response: Response) {
+  // TODO: handle 50001
```
decentralion, line 60: What is 50001?

src/plugins/discord/fetch.js
```diff
@@ -56,19 +56,40 @@ function getTokenFromEnv(): string {
...
+  if (response.status === 429) {
+    // TODO: Retry strategy
```
decentralion, line 62: How do rate limits work on Discord?

PR #1959, opened 2020-07-10
Add stress-tested algo for splitting Grain budgets

decentralion:
In #1957, I encountered a serious bug in our Grain distribution logic
that could result in distributing more Grain than was budgetted. This
was kind of unsurprising; doing arithmetic that mixes floats and bigints
is nontrivial.

This commit adds a stress-tested method, `splitBudget`, which splits a
fixed Grain budget based on floating-point scores. It
maintains the invariant that the amount distributed is always exactly
equal to the budget, even though this may result in slightly unfair
splits sometimes. (The unfairness comes about at scales like 10^-15
Grain, so I'm not too worried about it.)

The method is stress-tested by increasingly crazily example splits, so
we can have some confidence that it will satisfy its contract in
production. It has some routines that act to hide errors (e.g. letting
the last address absorb any remaining budget). So that we can get
confidence that we're not hiding large discrepancies, I've left some
commented-out "error sentinels" in the code, which log the size of the
discrepancies. These may be uncommented to see the size of the errors.

Along the way, I improved the accuracy of the `multiplyFloat` method so
that in the special case of multiplying by `1`, it returns the input
Grain value exactly.

Test plan: Inspect included unit tests, perhaps uncommenting some of the
"error sentinels".


src/ledger/grain.js
```diff
@@ -185,4 +194,87 @@ export function toFloatRatio(numerator: Grain, denominator: Grain): number {
...
+    throw new Error("shares sum to 0");
+  }
+  if (!isFinite(remainingShares)) {
+    throw new Error(`shares sum not finite: ${remainingShares}`);
+  }
+  const pieces = shares.map((s) => {
+    if (s < 0) {
+      throw new Error("negative share: " + s);
+    }
+    if (s === 0 || remainingShares === 0) {
```
hammadj, line 223: Isn't the `remainingShares === 0` case already handled on line 213?

src/ledger/grain.js
```diff
@@ -185,4 +194,87 @@ export function toFloatRatio(numerator: Grain, denominator: Grain): number {
...
+/**
+ * Splits a budget of Grain into fractional shares.
+ *
+ * It guarantees that the sum of all the splits will be exactly equal
+ * to the budget. It does not guarantee fairness, i.e. some users may
+ * receive more or less Grain than others with equal shares.
+ */
+export function splitBudget(
+  budget: Grain,
+  shares: $ReadOnlyArray<number>
```
hammadj, line 206: Shares = cred balance? Have we used this wording anywhere else before? Why not call it `credBalances` or something like that?

src/ledger/grain.js
```diff
@@ -185,4 +194,87 @@ export function toFloatRatio(numerator: Grain, denominator: Grain): number {
...
+
+    if (piece === "0" && remainingBudget !== "0" && s > 0) {
+      piece = "1";
+    }
+    remainingBudget = sub(remainingBudget, piece);
+    remainingShares -= s;
+    return piece;
+  });
+  // istanbul ignore if
+  if (lt(remainingBudget, "0")) {
```
hammadj, line 261: Would be good to have a comment explaining why we are not testing for coverage here

src/ledger/grain.js
decentralion: This is a more abstract function for partitioning a bigInt into pieces proportional to floating point shares. The shares aren't necessarily Cred--if you look at the usage in the grain allocation module, in one case the shares represent Cred, in another they represent per-user fractions of the total underpayment.

src/ledger/grain.js
decentralion: No, in line 213 we haven't started mutating remainingShares, so at that point it's effectively the total shares.

src/ledger/grain.js
decentralion: It's an invariant checker--if we found a test case that triggered this case, then it would mean the function is broken, so we can't expect to hit it in testing.

src/ledger/grainAllocation.js
```diff
@@ -130,21 +130,16 @@ function computeImmediateReceipts(
...
+  const tuples: [NodeAddressT, number][] = mapToArray(
+    immediateCredMap,
+    (x) => x
+  );
+  const scores = tuples.map((x) => x[1]);
+  const pieces = G.splitBudget(budget, scores);
+  return tuples.map((x, i) => ({
+    address: x[0],
+    amount: pieces[i],
+  }));
```
hammadj, line 142: Nit to improve readability. Also, calling it just "tuples" is a bit vague, so even if you want to keep it would be good to have a more descriptive name (like you have for `underpaymentTuples` on line 223). 

Same with `pieces` > `grainPieces`
```suggestion
  const scores: number[] = mapToArray(
    immediateCredMap,
    (x) => x[1]
  );
  const grainPieces = G.splitBudget(budget, scores);
  return mapToArray(
    immediateCredMap,
    (x, i) => ({
      address: x[0],
      amount: grainPieces[i],
    })
  );
```

src/ledger/grainAllocation.js
```diff
@@ -225,25 +220,18 @@ function computeBalancedReceipts(
...
-  });
-  if (G.gt(totalPaid, budget)) {
-    throw new Error(
-      `invariant violation: paid ${totalPaid} greater than budget ${budget}`
-    );
-  }
-  return receipts;
+  const underpaymentTuples: [NodeAddressT, G.Grain][] = mapToArray(
+    userUnderpayment,
+    (x) => x
```
hammadj, line 225: Wondering if we can just have this as the default param for the mapToArray util. That way you can just do `mapToArray(userUnderpayment)`. Makes sense with the name of the function, its just turning it into an array.

src/ledger/grainAllocation.js
```diff
@@ -225,25 +220,18 @@ function computeBalancedReceipts(
...
+  );
+  const underpaymentFractions = underpaymentTuples.map(
+    (x) => Number(x[1]) / Number(totalUnderpayment)
+  );
+  const pieces = G.splitBudget(budget, underpaymentFractions);
```
hammadj, line 230: nit: descriptive name
```suggestion
  const grainPieces = G.splitBudget(budget, underpaymentFractions);
```

src/ledger/grain.js
hammadj: got it, didnt see the mutation below.

src/ledger/grain.js
```diff
@@ -185,4 +194,87 @@ export function toFloatRatio(numerator: Grain, denominator: Grain): number {
...
+      console.error(
+        `${delta} discrepancy due to capping ${s} / ${remainingShares} from ${
+          s / remainingShares
+        } to 1`
+      );
+    }
+    */
+
+    if (piece === "0" && remainingBudget !== "0" && s > 0) {
+      piece = "1";
```
hammadj, line 254: Not sure I understand the purpose of changing the piece to 1 if their fraction of shares was near 0. Don't think its to use up any excess budget since that's handled on line 264 at the end of this function. Doesn't seem like something we would want to be doing, would be good to add a comment explaining why we are

src/ledger/grain.js
```diff
@@ -185,4 +194,87 @@ export function toFloatRatio(numerator: Grain, denominator: Grain): number {
...
+export function splitBudget(
+  budget: Grain,
+  shares: $ReadOnlyArray<number>
+): $ReadOnlyArray<Grain> {
+  if (lt(budget, ZERO)) {
+    throw new Error("negative budget");
+  }
+  let remainingBudget = budget;
+  let remainingShares = shares.reduce((a, b) => a + b, 0);
+  if (remainingShares === 0) {
```
hammadj, line 213: Should also handle the case where remainingShares is negative. 
```suggestion
  if (remainingShares <= 0) {
```

src/ledger/grain.js
decentralion: Tweaked the implementation to make this clearer.

src/ledger/grain.js
decentralion: Added a comment.

src/ledger/grain.js
decentralion: I renamed it to `scores` which will hopefully be clearer.

src/ledger/grain.js
decentralion: Yeah, you're right to call this out. It was needed before I added the "use up any excess budget, logic, now that I've done that we don't need it anymore. I've updated the implementation to remove it, and updated the test cases.

src/ledger/grain.js
decentralion: Sure, done.

src/ledger/grainAllocation.js
decentralion: I'm fine taking nits if they are added thru the suggestion UI (just lower friction); could you please make the suggestion apply the whole rename?

decentralion:
Thanks for the review @hammadj, it led to removing some vestigial special-casing from the algorithm.

Gonna leave this one open for a few more days b.c. i'm curious for @wchargin's thoughts.

decentralion:
I updated the PR so that it has smaller scope, and only adds the new algorithm, without modifying the allocation module to use it. Hop on over to https://github.com/sourcecred/sourcecred/pull/1966 to see the followon.

PR #1964, opened 2020-07-11
Start the DiskInstance class

decentralion:
This class is going to abstract over interacting with SourceCred
instances that are on disk. This commit just starts the class with
extremely simple scope: it loads the appropriate plugins for a SC
instance, based on the contents of `sourcecred.json`, and it errors
if the `sourcecred.json` file can't be found, or if specified plugins
are unavailable. Since this needs to be async, it happens in a static
method, called `instantiate`.

I took care with this method to emit very clear errors when it fails,
communicating exactly what went wrong (baseDir wasn't a directory, or
there was no sourcecred.json file, or the sourcecred.json file was
invalid, etc). This way we will give the user helpful feedback when
there's something basic broken in their instance setup.

Test plan: See included tests; `yarn test` passes.

### DIFF
--- a/src/api/diskInstance.js
+++ b/src/api/diskInstance.js
@@ -1,0 +1,103 @@
+// @flow
+
+import {join} from "path";
+import fs from "fs-extra";
+import {type PluginId} from "./pluginId";
+import {type Plugin} from "./plugin";
+import {rawParser} from "./instanceConfig";
+import {type TaskReporter, SilentTaskReporter} from "../util/taskReporter";
+
+export class DiskInstance {
+  +_instanceDirectory: string;
+  +_plugins: $ReadOnlyMap<PluginId, Plugin>;
+  +_taskReporter: TaskReporter;
+
+  constructor(
+    instanceDirectory: string,
+    plugins: $ReadOnlyMap<PluginId, Plugin>,
+    // The TaskReporter that will receive task timing info from Plugin methods.
+    // Defaults to the SilentTaskReporter.
+    taskReporter: ?TaskReporter
+  ) {
+    this._instanceDirectory = instanceDirectory;
+    this._plugins = plugins;
+    this._taskReporter = taskReporter || new SilentTaskReporter();
+  }
+
+  /**
+   * Gets the filesystem directory containing this Instance.
+   */
+  instanceDirectory(): string {
+    return this._instanceDirectory;
+  }
+
+  /**
+   * Gets the Plugins that are active in this instance.
+   */
+  plugins(): $ReadOnlyArray<Plugin> {
+    return Array.from(this._plugins.values());
+  }
+
+  /**
+   * Instantiate a DiskInstance from an existing SourceCred directory.
+   *
+   * Arguments:
+   * - instanceDirectory: a valid SourceCred instance directory (i.e. has a
+   *   sourcecred.json file)
+   * - availablePlugins: runtime instantiations of all the bundled plugins
+   *   present in the instance
+   * - taskReporter: an optional TaskReporter that will receive logging info
+   *   from plugins as they progress in loads, etc.
+   *   Defaults to the SilentTaskReporter.
+   */
+  static async instantiate(
+    instanceDirectory: string,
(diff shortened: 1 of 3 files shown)


src/api/diskInstance.js
```diff
@@ -0,0 +1,103 @@
...
+  ) {
+    this._instanceDirectory = instanceDirectory;
+    this._plugins = plugins;
+    this._taskReporter = taskReporter || new SilentTaskReporter();
+  }
+
+  /**
+   * Gets the filesystem directory containing this Instance.
+   */
+  instanceDirectory(): string {
```
hammadj, line 30: For stuff like this I usually like to mark it as a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) so its clear that you aren't performing any action:
```suggestion
  get instanceDirectory(): string {
```

src/api/diskInstance.js
```diff
@@ -0,0 +1,103 @@
...
+          `cannot load instance: ${instanceDirectory} doesn't exist`
+        );
+      }
+      throw new Error(`cannot load instance: ${e}`);
+    }
+    if (!stat.isDirectory()) {
+      throw new Error(
+        `cannot load instance: ${instanceDirectory} is not a directory`
+      );
+    }
```
hammadj, line 72: I imagine we would be doing this a lot, would be nice to extract into a helper function
```js
try {
  checkDirectoryExists(instanceDirectory)
} catch (e) {
  throw new Error(`cannot load instance: ${e}`);
}
```

src/api/diskInstance.js
```diff
@@ -0,0 +1,103 @@
...
+    } catch (e) {
+      throw new Error(`error parsing sourcecred.json: ${e}`);
+    }
+    const {bundledPlugins} = config;
+    const pluginSet = new Set(bundledPlugins);
+    const plugins = new Map(availablePlugins.map((p) => [p.id, p]));
+    const result = new Map();
+    for (const id of pluginSet) {
+      const plugin = plugins.get(id);
+      if (plugin == null) {
```
hammadj, line 96: Curious, why not just `if (!plugin)` ?

src/api/diskInstance.test.js
```diff
@@ -0,0 +1,131 @@
...
+    });
+    it("errors if there is no config file", async () => {
+      const path = tmp.dirSync().name;
+      diskUtil.mkdirx(path);
+      const fail = DiskInstance.instantiate(path, []);
+      await expect(fail).rejects.toThrow(
+        `no sourcecred.json file present in ${path}`
+      );
+    });
+    it("errors if the config file can't isn't valid json", async () => {
```
hammadj, line 98: typo:
```suggestion
    it("errors if the config file isn't valid json", async () => {
```

src/api/diskInstance.test.js
```diff
@@ -0,0 +1,131 @@
...
+    },
+  };
+
+  function setupInstanceDirectory(config: RawInstanceConfig) {
+    const dir = tmp.dirSync().name;
+    fs.writeFileSync(join(dir, "sourcecred.json"), JSON.stringify(config));
+    return dir;
+  }
+
+  async function example() {
```
hammadj, line 45: naming: `setupFixture` would be clearer

src/api/diskInstance.test.js
```diff
@@ -0,0 +1,131 @@
...
+    });
+    it("errors if a plugin is not available", async () => {
+      const config = {bundledPlugins: [dummyPlugin.id]};
+      const dir = setupInstanceDirectory(config);
+      const fail = DiskInstance.instantiate(dir, []);
+      await expect(fail).rejects.toThrow(
+        "plugin sourcecred/dummy not available"
+      );
+    });
+    it("handles the case wth no plugins", async () => {
```
hammadj, line 118: typo
```suggestion
    it("handles the case with no plugins", async () => {
```

decentralion:
During a discussion today, we realized this isn't needed for beta milestone. Since my hands are full, I am going to close this PR for now; will re-open it after we've shipped beta.

PR #1966, opened 2020-07-12
Ensure Grain allocations match their budget

decentralion:
This commit updates the grainAllocation module so that distributions
always distribute their budgets precisely, or fail. It does this by
incorporating the new stress-tested `splitBudget` method introduced over
in #1959, and modifying the contract so that we fail rather than
returning an empty allocation in the case when totalCred sums to 0.

Test plan: `yarn test` passes, and we are now using a stress-tested core
algorithm for the distribution, rather than some ad-hoc but undertested
math.


src/ledger/grainAllocation.test.js
```diff
@@ -46,24 +46,18 @@ describe("src/ledger/grainAllocation", () => {
...
-          });
+        expect(actual).toEqual({
+          policy: zeroPolicy,
+          receipts: [],
         });
-
-        it("all the cred sums to 0", () => {
-          const actual = computeAllocation(
+      });
+      it("should error when the Cred sums to 0", () => {
```
hammadj, line 58: Should have a test case for negative cred as well

src/ledger/grainAllocation.js
```diff
@@ -117,34 +117,29 @@ function computeImmediateReceipts(
...
+  }
+
   const lastSlice = credHistory[credHistory.length - 1];
 
   const immediateCredMap = lastSlice.cred;
 
   const totalCred = sum(immediateCredMap.values());
   if (totalCred === 0) {
-    return [];
+    throw new Error("cannot allocate Grain: cred sums to 0");
```
hammadj, line 134: Should be <= 0 in case there's negative cred?

src/ledger/grainAllocation.test.js
decentralion: I considered adding it, but it's already enforced in the outer method:
https://github.com/sourcecred/sourcecred/pull/1966/files#diff-b3948ae03dddf0431e5873d71313bdaeR82-R92

And note the helpers getting tested in this test are only reachable via that wrapper method.

src/ledger/grainAllocation.js
decentralion: checking the sum wouldn't guarantee that there's no negative cred (e.g. `[1,-2,5]`). However, the `splitBudget` method will error if there's any negative cred. Maybe we should just remove this check?

I'm almost tempted to try to enforce this at the type level, and have a new math util module with:

```js
export opaque type Finite: number = number;
export opaque type FiniteNonNegative: number = number;
```

And associated parsers and converters so we can use them at runtime.
I'm tired of constantly checking params against `NaN, Infinity, -Infinity`, etc.

Curious what @wchargin thinks.

PR #1968, opened 2020-07-12
Initiative Editor Updates / Fixes

hammadj:
This PR addresses some of my feedback on #1889. I was already implementing the changes locally to test it out, so thought I might as well save @topocount's time and push the changes myself.

Main changes:
- Consolidate implementation of the Initiatives form into a single reused component, deduplicating code in the InitiativeCreate and InitiativeEdit components
- useMemo for things that we don't want/need to run on every render
- Refactor InitiativeEditor component from component class to functional + hooks
- Fixed bug where multiple inline contributions were all being created with the same ID / key

Test Plan: Ensure that the initiative editor frontend still runs and that multiple inline contributions are created with unique IDs

### DIFF
--- a/src/ui/components/Initiatives/InitiativeCreate.js
+++ b/src/ui/components/Initiatives/InitiativeCreate.js
@@ -1,23 +1,26 @@
 // @flow
 import React from "react";
-import { v4 as uuid } from "uuid";
-import { Create } from "react-admin";
-import { CredView } from "../../../analysis/credView";
-import { InitiativeForm } from "./InitiativeForm";
+import {v4 as uuid} from "uuid";
+import {Create} from "react-admin";
+import {CredView} from "../../../analysis/credView";
+import {InitiativeForm} from "./InitiativeForm";
 
 export const InitiativeCreate = (credView: CredView) => (props: Object) => {
-  const defaultInitiativeValues = React.useMemo(() => ({
-    id: uuid(),
-    title: '',
-    timestampMs: Date.now(),
-    champions: [],
-    dependencies: [],
-    references: [],
-    contributions: [],
-    weight: {incomplete: 0, complete: 0},
-    completed: false,
-  }), []);
-  
+  const defaultInitiativeValues = React.useMemo(
+    () => ({
+      id: uuid(),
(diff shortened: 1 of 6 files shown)


src/ui/components/Initiatives/InitiativeCreate.js
```diff
@@ -1,118 +1,32 @@
 // @flow
 import React from "react";
 import {v4 as uuid} from "uuid";
```
topocount, line 3: We should probably update this to use William's uuid implementation and shed this dependency: 
https://github.com/sourcecred/sourcecred/blob/master/src/util/uuid.js

src/ui/components/Initiatives/InlineKeyField.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import React from "react";
+import {v4 as uuid} from "uuid";
```
topocount, line 3: same here

src/ui/components/Initiatives/InlineKeyField.js
```diff
@@ -0,0 +1,36 @@
...
+    input,
+    meta: {touched, error},
+    isRequired,
+  } = useInput({
+    ...props,
+    initialValue: key,
+  });
+
+  return (
+    <TextField
```
topocount, line 26: Is this something we want to hide? I hid it because I figured users wouldn't want to see uuid fields, but I'm open to the idea of leaving it as shown, especially now that this bug was caught.

hammadj: Yea, lets leave it shown for now since its still going to be used by devs mostly, can hide it later once we are confident things are working properly. Will be handy for making sure things are wired up correctly if we can see the IDs in the UI

src/ui/components/Initiatives/InitiativeCreate.js
hammadj: done

src/ui/components/Initiatives/InlineKeyField.js
hammadj: done

PR #1973, opened 2020-07-12
Script to generate ledger from legacy project.json

hammadj:
Usage: `yarn createLedger <legacy-instance-dir>`

Loads the project.json file from the given instance directory, instantiates a Ledger, and fills it with
users and aliases based on the identities specified in the project.json. Prints the resulting eventLog to
stdout.

The users and their aliases are created in the same order as they are defined in the project.json. Note that the current credit instance has an invalid username for @s-ben ("s_ben"). Underscores are not allowed by the ledger, so it will need to be updated before this will actually work.

*TestPlan:* Run the createLedger command on the SourceCred cred instance and check if the generated eventLog is correct

Paired with: @decentralion 

Progress towards: #1976

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.8.7",
+    "@babel/node": "^7.8.7",
     "@babel/plugin-proposal-class-properties": "^7.8.3",
     "@babel/plugin-syntax-bigint": "^7.8.3",
     "@babel/preset-env": "^7.8.7",
@@ -109,6 +110,7 @@
     "build:backend": "NODE_ENV=development webpack --config config/webpack.config.backend.js",
     "api": "webpack --config config/webpack.config.api.js",
     "test": "node ./config/test.js",
+    "createLedger": "babel-node ./scripts/createLedgerFromLegacyProject.js",
     "unit": "BABEL_ENV=test NODE_ENV=test jest --env=jsdom",
     "sharness": "make -sC ./sharness prove PROVE_OPTS=-f TEST_OPTS='--chain-lint'",
     "sharness-full": "make -sC ./sharness prove PROVE_OPTS=-vf TEST_OPTS='-v --chain-lint --long'",
--- a/scripts/createLedgerFromLegacyProject.js
+++ b/scripts/createLedgerFromLegacyProject.js
@@ -1,0 +1,102 @@
+// @flow
+import fs from "fs-extra";
+import path from "path";
+
+import * as C from "../src/util/combo";
+import type {Identity} from "../src/plugins/identity/identity";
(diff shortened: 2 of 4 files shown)


scripts/createLedgerFromLegacyProject.js
```diff
@@ -0,0 +1,96 @@
...
+  try {
+    const contents = await fs.readFile(jsonPath);
+    return parseProject(JSON.parse(contents));
+  } catch (e) {
+    if (e.message.startsWith("ENOENT:")) {
+      throw `project from directory "${instanceDirectory}" not loaded`;
+    }
+    throw e;
+  }
+}
```
decentralion, line 67: Can just use loadJson:
https://github.com/sourcecred/sourcecred/blob/5b0faf284e53112581a21698b887b82767f07621/src/util/disk.js#L19-L31

src/util/uuid.js
```diff
@@ -32,7 +32,7 @@ function isClean(s: string): boolean {
 function randomUuidUnchecked(): string {
   const bytes = getRandomValues(new Uint8Array(16));
   const blob = [...bytes].map((n) => String.fromCharCode(n)).join("");
-  return btoa(blob).slice(0, -2); // drop "==" padding
+  return base64Encode(blob).slice(0, -2); // drop "==" padding
```
decentralion, line 35: this change should definitely be in its own PR, and reviewed by @wchargin.

package.json
```diff
@@ -49,6 +49,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.8.7",
+    "@babel/node": "^7.8.7",
```
decentralion, line 52: Why do we need babel/node? We don't need it for any of our other scripts.

Also: IMO this script does not need to be in package.json. Just putting it in `scripts/createLedger.js` and then wiring it into the backend build is fine.

decentralion: Followup: I guess the advantage to babel-node is that we don't need to manually wire it in as a backend entry point, unlike these?

https://github.com/sourcecred/sourcecred/blob/5b0faf284e53112581a21698b887b82767f07621/config/paths.js#L31-L43

I'm open to the idea that this is a better way to do things, but if so we should do it as a separate commit where we migrate all the scripts. For now can we just make this PR add a new backendEntryPoint?

decentralion:
Regarding the _ in username issue--I think we should follow GitHub's lead in disallowing underscores in usernames, it gives us more flexibility if the usernames are dns compliant -- for now let's just convert any underscores in usernames to hyphens. Users can change it later, and if it creates a username conflict the ledger will throw an error and we can deal with it then.

src/util/uuid.js
decentralion: I looked into it more, and it seems harmless.

PR #1989, opened 2020-07-14
Chore: combine frontends within React Admin

### LINKED ISSUE #1987: Integrate react-admin and wrap explorer within it

topocount:
At long last, The Explorer lives inside react admin. Additionally, deep links now work thanks to Hash routing. We could also easily utilize regular Browser routing instead, if that is preferred. 

A white background was hardcoded for the Explorer to ensure it's visible inside react admin.

test-plan: build and install this frontend in the example instance. Ensure the Explorer and intiitiatives editor are fully functional from within React Admin. Additionally, ensure the browser url updates as you click around to reflect the current page.

merge plan: redirect this to `master` after #1889 is merged

resolves #1987


topocount:
After some experimentation, I've found that the `webpack-dev-server` route is not a suitable environment for testing routing. This route provides a hot-reloading environment, but it's inside an iframe that provides a false base path that doesn't include the `webpack-dev-server` route and therefore react-router's attempts to update the path fail.
 
 This setting what determines whether or not the dev server is in iframe mode:
https://github.com/sourcecred/sourcecred/blob/edadc12b6d8c44953837e563f0d219e7cff14b74/config/webpack.config.web.js#L42-L43

setting `inline: true` causes a build error, and is probably the reason it's disabled.

All this is meant to convey that routing testing must be done either with a production build or with a successful dev build with inline set to true.

src/ui/components/Explorer.js
```diff
@@ -196,7 +196,7 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
     const sortedNodes = sortBy(nodes, (n) => -n.credSummary.cred);
     const total = sum(nodes.map((n) => n.credSummary.cred));
     return (
-      <div style={{width: "1000px", margin: "0 auto"}}>
+      <div style={{width: "1000px", margin: "0 auto", background: "white"}}>
```
hammadj, line 199: Can we have a responsive width? What happens if you set it to 100%?

src/ui/components/Menu.js
```diff
@@ -0,0 +1,47 @@
...
+import {MenuItemLink, getResources} from "react-admin";
+import {withRouter} from "react-router-dom";
+import TrendingUpIcon from "@material-ui/icons/TrendingUp";
+import DefaultIcon from "@material-ui/icons/ViewList";
+
+type menuProps = {|onMenuClick: Function, logout: Function|};
+
+const Menu = ({onMenuClick, logout}: menuProps) => {
+  const isXSmall = useMediaQuery((theme) => theme.breakpoints.down("xs"));
+  const open = useSelector((state) => state.admin.ui.sidebarOpen);
```
hammadj, line 15: What would it take for us to get this type safe? Seems like there's typedefs available: https://github.com/flow-typed/flow-typed/tree/master/definitions/npm/react-redux_v7.x.x

Not for this PR, just something that would be good to add in soon

decentralion:
It's a nice start!

From a functionality perspective, I want the following changes:
- The explorer should be the default view
- The explorer should have a lot more space (looks really cramped right now)
- The initiatives tab should only appear when the `sourcecred.json` instance config lists the initiatives plugin as being active

src/ui/components/Explorer.js
topocount: I'd argue that this is out of scope for this PR. This line change essentially ensures visual parity with the existing Explorer. Discussion about styling for the explorer is ongoing and this will be addressed within those changes.

src/ui/components/Menu.js
topocount: I added the react-redux libdefs but `react-router-dom` conflicts with an existing def in webutils. We could probably resolve manually in our files.

Additionally, we'll probably need to stub out material ui, if we think it's worthwhile to do so

src/ui/components/Explorer.js
topocount: 80% improves things but there are still some issues: https://github.com/sourcecred/sourcecred/pull/1989/commits/44932129468593dea9704a1e3a8aeb4a2d1671b0

PR #1996, opened 2020-07-15
Add interface for creating and Modifying ledger users

topocount:


src/ui/uiUtils.js
```diff
@@ -63,3 +72,26 @@ export const dateFormatter = (t: TimestampMs): ?DateString => {
...
+  if (window.navigator && window.navigator.msSaveOrOpenBlob) {
+    // accomodate IE11+ & Edge
+    window.navigator.msSaveOrOpenBlob(blob, `export.json`);
+  } else {
+    fakeLink.setAttribute("href", URL.createObjectURL(blob));
+    fakeLink.setAttribute("download", `initiatives.json`);
+    fakeLink.click();
+  }
+  fakeLink.remove();
+};
```
decentralion, line 97: uiUtils shouldn't reference initiatives

src/ui/uiUtils.js
```diff
@@ -34,6 +35,14 @@ type ContributionEntry = {|
   +weight: NodeWeight,
 |};
 
+type UserEntry = {|
+  +id: UserId,
+  +name: Username,
+  // stores aliases that have not yet been linked in the ledger module
```
decentralion, line 41: This feels fragile to me--I'd rather keep a ledger in memory, let the user mutate it (e.g. add an alias), and then use the updated ledger state to drive the UI. And then add a `save` button which serializes it back to disk. That way we don't need to explicitly store "make these mutations" as a separate state. What do you think?

topocount: The big driver for this was letting users modify linked aliases that haven't yet been saved to disk. Should we just make adding aliases a one-way operation regardless of whether or not they've been synced in the ledgerm

decentralion: The ledger is a mutable object. So just use the ledger, in memory, as the source of truth: i.e. when you call `ledger.createUser` or `ledger.addAlias`, that updates the data that is being rendered in the UI.

Then the save operation is just to save the updated ledger to disk.

This way we don't have to keep track of "state in ledger + changes we want to apply". We just keep track of the ledger state, and update it directly.

PR #1999, opened 2020-07-15
Create config parser for Initiatives Plugin

hammadj:
Setups up the parser to be used in the Initiatives plugin for the new CLI

TestPlan: Check to see if the parser has the required fields (i.e. remoteUrl)

### DIFF
--- a/src/plugins/initiatives/config.js
+++ b/src/plugins/initiatives/config.js
@@ -1,0 +1,12 @@
+// @flow
+
+import * as C from "../../util/combo";
+import {_validateUrl} from "./initiativesDirectory";
+
+export type InitiativesConfig = {|
+  +remoteUrl: string,
+|};
+
+export const parser: C.Parser<InitiativesConfig> = C.object({
+  remoteUrl: C.fmap(C.string, _validateUrl),
+});


src/plugins/initiatives/config.js
```diff
@@ -0,0 +1,16 @@
...
+import * as C from "../../util/combo";
+import {_validateUrl} from "./initiativesDirectory";
+
+export type InitiativesConfig = {|
+  +remoteUrl: string,
+|};
+
+export function parseConfig(raw: C.JsonObject): InitiativesConfig {
+  return configParser.parseOrThrow(raw);
+}
```
decentralion, line 12: Don't need this. Just use `loadJson` or `loadJsonWithDefault`, which takes the parser directly.

src/plugins/initiatives/config.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const configParser: C.Parser<InitiativesConfig> = C.object({
+  remoteUrl: C.fmap(C.string, _validateUrl),
+});
```
decentralion, line 16: I usually just export the parsers as `parser`, and then import with the needed name.

PR #2001, opened 2020-07-15
Create parser for Initiative Files

hammadj:
This parser will be used in the Initiatives Plugin for the new CLI. It supports both v0.1.0 and v0.2.0
Initiative files.

TestPlan: Sanity check that all the right fields are there and that upgrade strategy makes sense

### DIFF
--- a/src/plugins/initiatives/parseInitiative.js
+++ b/src/plugins/initiatives/parseInitiative.js
@@ -21,31 +21,29 @@ const CommonFields = {
   completed: C.boolean,
 };
 
-const Parse_020: C.Parser<InitiativeFileV020> = (() => {
-  const NodeEntryParser = C.object(
-    {
-      title: C.string,
-      timestampIso: TimestampParser,
-      contributors: C.array(URLParser),
-    },
-    {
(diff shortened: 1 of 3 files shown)


src/plugins/initiatives/parseInitiative.js
```diff
@@ -0,0 +1,65 @@
...
+    entries: C.array(NodeEntryParser),
+  });
+
+  return C.object(CommonFields, {
+    contributions: EdgeSpecParser,
+    dependencies: EdgeSpecParser,
+    references: EdgeSpecParser,
+    champions: C.array(URLParser),
+  });
+})();
```
decentralion, line 48: I generally prefer to assemble "bare" parsers (not inside an IIFE) and compose them outside of a function scope. Think its a bit simpler. Fine with this too, though (no change requested).

hammadj: yea, I just noticed a bunch of the other parsers were defined as IIFEs and copied that style since I thought that was preferred for some reason lol, was also confused.

PR #2006, opened 2020-07-17
refactor grain allocations

decentralion:
Re-writes the grain allocation module, around the new assumption that we
only allocate to ids.

Also: dramatically cleans the interface, and validates the arguments to
the underlying receipt allocator (saving them from needing to
redundantly check that there are identities, that the budget is
nonnegative, etc) and validates the output (to ensure that sum receipts
is always equal to the budget).

Also: pulls the distribution type out to a new file, and patches the
`Ledger.distributeGrain` method to consume the new style Distributions.

Test plan: Unit tests rewritten; run `yarn test`


decentralion:
cc @hammadj -- this is a re-write of the module we wrote together, you might find it interesting.

PR #2009, opened 2020-07-17
replace isomorphic-webcrypto

decentralion:
Unfortunately, isomorphic-webcrypto wasn't that isomorphic, and
attempting to depend on it in web code made it impossible to compile the
frontend.

This commit replaces isomorphic-webcrypto with get-random-values, which
has a pleasingly simple [implementation] and plays nicely with our build
stack. After effecting this dependency swap, it's now possible to import
the uuid module from frontend code.

However, there is an interesting caveat. Our server-side rendering setup
does not provide an environment in which crypto primitives are
available. Therefore, if any page that is server-side rendered attempts
to use this module (e.g. generate a UUID) _during initial render_, then
the server side rendering will fail. Thus, we'll need to be careful only
to generate UUIDs on non-initial renders (e.g. in response to user
action). I think this can work out in practice; and if it doesn't we can
cross that bridge when we come to it.

Test plan: Add an unused dependency on the uuid module to some frontend
code, and confirm that the build doesn't break.

[implementation]: https://github.com/KenanY/get-random-values/blob/31778f726fa97d1459f353db3823860e505518f9/index.js


wchargin:
I admit that I didn’t test it, but I _really_ thought it would be safe
to assume that a module with “isomorphic” and “web” as two-thirds of its
name would work isomorphically on the web. Oh, well; shows what I know…

PR #2015, opened 2020-07-21
implement ledger identity editor POC

topocount:
The ledger identity editor enables users to collect grain by allowing
them to claim aliases from various plugins and unite them under a
single identity.

This editor allows for the creation and modification of identity
usernames, along with the assignment of aliases.

Note: aliases cannot yet be unassigned as there is no graceful way to
accomplish this in the backend yet.

test plan:
 - run the dev web server and browse to the ledger admin view.
 - Try to create an identity, then assign a couple aliases to it.
 - Click on the newly created identity name in the identities list to
 reset the view and create a new identity
 - Create a second identity and assign a couple more aliases to it.
 - Note that assigned aliases are no longer visible in the dropdown
 - click between users and ensure the name and alias lists repopulate
 with the correct information

### DIFF
--- a/src/ui/components/AliasSelector.js
+++ b/src/ui/components/AliasSelector.js
@@ -1,0 +1,162 @@
+// @flow
+
+import React, {useState, useMemo} from "react";
+import {useCombobox, useMultipleSelection} from "downshift";
+import {Ledger} from "../../ledger/ledger";
+import {type Identity} from "../../ledger/identity";
+import {CredView} from "../../analysis/credView";
+import {type NodeAddressT} from "../../core/graph";
+import Markdown from "react-markdown";
+
+type Props = {|
+  +currentIdentity: Identity | null,
+  +ledger: Ledger,
+  +credView: CredView,
+  +setLedger: (Ledger) => void,
+  +setCurrentIdentity: (Identity) => void,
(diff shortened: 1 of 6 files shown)


topocount:
@decentralion I added the account-active checkbox :ballot_box_with_check: 

not sure what's going on with these tests atm. the failures seem unrelated my my changes here.

decentralion:
Yeah, the flake is unrelated--fixed in https://github.com/sourcecred/sourcecred/pull/2016

src/ui/components/AliasSelector.js
```diff
@@ -0,0 +1,162 @@
...
+    if (node) {
+      return node.description;
+    }
+    return "";
+  };
+
+  const getFilteredItems = (items) =>
+    items.filter(
+      (item) =>
+        !ledger._aliases.has(item) &&
```
decentralion, line 58: didn't catch this in initial review, but: if a property is `_named` like `ledger._aliases`, then it's a module-internal implementation detail, and it shouldn't be accessed externally. the right solution is either to to get the same data via existing apis (`ledger.accounts().map((a) => a.identity.aliases)` in this case) or to add a new API. Ill fix this up.

PR #2023, opened 2020-07-23
Add the `cli/grain` command for distributing Grain

decentralion:
This adds a new CLI command, `sourcecred grain`, which distributes
Grain. This happens based on a GrainConfig file, at `config/grain.json`.

The GrainConfig has a basic policy where you specify how much gets
distributed using the IMMEDIATE and BALANCED strategies. You also
specify how many weeks worth of distributions can be done at once.

This parameter is relevant because SC wants to create one distribution
per week, so that if (for example) you forgot to run the script for the
past 4 weeks, it will create 4 distributions for you. However, you might
not want this, or you might want to go back only a little while... e.g.
if you're running SC for the first time on an old project and don't want
to issue hundreds of distributions worth of Grain. So you can tune it
via the `maxSimultaneousDistributions` parameter, which defaults to
Infinity if unset.

Test plan: Tested via actual Grain distributions. This is the kind of
"wire everything together" code that's annoying to test and unlikely to
have subtle bugs, so the unit testing is spotty.

### DIFF
--- a/config/webpack.config.web.js
+++ b/config/webpack.config.web.js
@@ -89,6 +89,26 @@ async function makeConfig(
           "/config/",
           express.static(path.join(developmentInstancePath, "config"))
         );
+        app.use(
+          "/data/",
+          express.static(path.join(developmentInstancePath, "data"))
+        );
+
+        // configure the dev server to support writing the ledger to disk
+        app.use(express.text());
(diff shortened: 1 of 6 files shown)


decentralion:
Note: I accidentally merged this with a weird commit message, did a bad rebase. Oops.

PR #2027, opened 2020-07-23
Implement Initiatives plugin for new CLI

hammadj:
Sets up a cli2 plugin for the Initiatives plugin and registers it in the bundled plugins. This is for the current 0.2.0 initiative format, so it doesn't work with node addresses yet. Also still need to get the hack fix from #1878 in before this will work properly for MetaGame's instance.

Test Plan: Create an initiatives directory with some example initiatives in the example instance and check if they are successfully loaded into the graph

### DIFF
--- a/src/plugins/initiatives/plugin.js
+++ b/src/plugins/initiatives/plugin.js
@@ -1,0 +1,82 @@
+// @flow
+import type {Plugin, PluginDirectoryContext} from "../../api/plugin";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import {parser, type InitiativesConfig} from "./config";
+import {declaration} from "./declaration";
+import {join as pathJoin} from "path";
+import type {ReferenceDetector} from "../../core/references";
+import type {WeightedGraph} from "../../core/weightedGraph";
+import {weightsForDeclaration} from "../../analysis/pluginDeclaration";
+import {createWeightedGraph} from "./createGraph";
+import {
+  type PluginId,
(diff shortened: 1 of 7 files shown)


src/plugins/initiatives/plugin.js
```diff
@@ -0,0 +1,82 @@
...
+
+  // We dont need to load any data since all the initiative files are on disk
+  async load(): Promise<void> {}
+
+  async graph(
+    ctx: PluginDirectoryContext,
+    rd: ReferenceDetector
+  ): Promise<WeightedGraph> {
+    const {remoteUrl} = await loadConfig(ctx);
+    const localPath = getDirectoryFromEnv();
```
decentralion, line 54: Local path should be within the plugin config directory, so we should be using the pluginDirectoryContext.

E.g. `const localPath = path.join(ctx.configDirectory(), "initiative-files")`

src/plugins/initiatives/plugin.js
```diff
@@ -0,0 +1,82 @@
...
+
+function getDirectoryFromEnv(): string {
+  const rawDir = process.env[DIRECTORY_ENV_VAR_NAME] || "initiatives";
+  if (rawDir == null) {
+    throw new Error(
+      `No initiatives directory provided: set ${DIRECTORY_ENV_VAR_NAME}`
+    );
+  }
+  return rawDir;
+}
```
decentralion, line 37: Let's delete this function -- no reason to be getting stuff from ENV when we now have plugin directory contexts

PR #2034, opened 2020-07-25
Grain: Implement a parser for stringified floats

topocount:
This is most useful for processing form input values before passing them
into the ledger, since all form fields return strings

In this case, a "float string" is a string that returns a number value
when passed into `parseFloat`

The reason to circumvent any floating point values is to avoid losses in
precision. By modifying the string directly in a predictable pattern, we can
convert uer-generated floating point values to grain at full fidelity, and avoid
any fuzzy floating point arithmetic

The tradeoff here is around versatility. Values with more decimals than the
allowable precision will yield an error when passed in.

test-plan: thorough unit tests are included

### DIFF
--- a/src/ledger/grain.js
+++ b/src/ledger/grain.js
@@ -202,7 +202,7 @@ export function fromFloatString(
   if (typeof x !== "string") {
     throw new Error(`not a string: ${x}`);
   }
-  if (Number.isNaN(parseFloat(x))) {
+  if (!(isFinite(x) && x.trim())) {
     throw new Error(`input not a valid number: ${x}`);
   }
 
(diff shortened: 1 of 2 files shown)


src/ledger/grain.js
```diff
@@ -178,6 +178,45 @@ export function fromInteger(x: number): Grain {
...
+ * allowable precision will yield an error when passed in.
+ */
+export function fromFloatString(
+  x: string,
+  precision: number = DECIMAL_PRECISION
+): Grain {
+  if (typeof x !== "string") {
+    throw new Error(`not a string: ${x}`);
+  }
+  if (Number.isNaN(parseFloat(x))) {
```
decentralion, line 205: Use `isFinite` instead; then we'll also throw for Infinity and -Infinity

PR #2040, opened 2020-07-27
Add function to fetch discourse user data, snapshots, and tests

befitsandpiper:
Test plan: Added snapshot tests to cover new function


src/plugins/discourse/fetch.js
```diff
@@ -79,6 +79,11 @@ export type TopicWithPosts = {|
   +posts: $ReadOnlyArray<Post>,
 |};
 
+export type User = {|
+  +username: string,
+  +trust_level: number,
```
decentralion, line 84: should be `trustLevel` for consistency with the rest of the module.

src/plugins/discourse/fetch.test.js
```diff
@@ -16,6 +16,12 @@ describe("plugins/discourse/fetch", () => {
         await snapshotFetcher().likesByUser("dl-proto", 0)
       ).toMatchSnapshot();
     });
+    it("loads user data from snapshot dl-proto", async () => {
+      expect(await snapshotFetcher().getUserData("dl-proto")).toMatchSnapshot();
+    });
+    it("loads user data from snapshot system", async () => {
+      expect(await snapshotFetcher().getUserData("system")).toMatchSnapshot();
+    });
```
decentralion, line 24: nit: just one is fine, i don't think there's much benefit from testing a second user

PR #2042, opened 2020-07-27
Add trustLevel to likeAction, post, user tables

befitsandpiper:
Test Plan: test behavior for posts and likes


src/plugins/discourse/fetch.js
```diff
@@ -415,6 +423,7 @@ function parseLike(json: any): LikeAction {
     username: json.target_username,
     postId: json.post_id,
     timestampMs: Date.parse(json.created_at),
+    trustLevel: json.trustLevel,
```
decentralion, line 426: This doesn't seem right--there's no trust level on that object.

src/plugins/discourse/createGraph.test.js
```diff
@@ -44,13 +44,13 @@ describe("plugins/discourse/createGraph", () => {
     posts(): $ReadOnlyArray<Post> {
       return this._posts;
     }
-    users(): $ReadOnlyArray<string> {
-      const users = new Set();
+    users(): $ReadOnlyArray<User> {
+      const users = [];
```
decentralion, line 48: This array will now have two copies of `{username: "decentralion", trustLevel: 3}`, which is not what we want.

src/plugins/discourse/fetch.js
```diff
@@ -68,6 +68,8 @@ export type Post = {|
   // reply to the first post.
   +replyToPostIndex: number | null,
   +timestampMs: TimestampMs,
+  // user trustLevel
+  +trustLevel: number | null,
```
decentralion, line 72: Do we know that this can be null in the incoming data? Since the authorUsername is not nullable, I would guess the trustLevel isn't either.

src/plugins/discourse/fetch.js
```diff
@@ -287,11 +292,11 @@ export class Fetcher implements Discourse {
   }
 
   async likesByUser(
-    username: string,
+    targetUsername: string,
```
decentralion, line 295: why the rename?

src/plugins/discourse/mirror.js
```diff
@@ -198,7 +198,10 @@ export class Mirror {
       let offset = 0;
       let upToDate = false;
       while (!upToDate) {
-        const likeActions = await this._fetcher.likesByUser(user, offset);
+        const likeActions = await this._fetcher.likesByUser(
+          user.username,
+          offset
+        );
```
decentralion, line 204: nit: revert this file. (surprised, usually prettier is pretty good at not leaving spurious formatting changes)

src/plugins/discourse/createGraph.test.js
decentralion: We could fix this by having a set track the usernames, and only adding the user to the user array if their username is not yet in the set. However, I think that's too complicated: instead, let's just have this method explicitly return the right set of users that are included in the test case. That way we can also give the users some varying trust levels, which will be good for an example test case later once we actually use the trust levels.

PR #2046, opened 2020-07-27
don't render anything for empty aliases

topocount:
aliases connected to an identity get subsumed and no longer return when
naively queried from the graph after `$ sourcecred score` is executed.
This commit will not render anything in the ledger admin in these cases

test-plan: ensure no whitespace renders, including carriage returns,
  occur when an identity with linked aliases is rendered after
  credscores are recalculated.

### DIFF
--- a/src/ui/components/AliasSelector.js
+++ b/src/ui/components/AliasSelector.js
@@ -101,21 +101,18 @@ export function AliasSelector({
         <h2>Aliases:</h2>
       </label>
       <div>
-        {selectedItems.map(
-          (selectedItem, index) =>
-            getNodeDescription(selectedItem) && (
-              <span
-                key={`selected-item-${index}`}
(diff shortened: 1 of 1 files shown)


src/ui/components/AliasSelector.js
```diff
@@ -101,18 +101,21 @@ export function AliasSelector({
...
+                {...getSelectedItemProps({selectedItem, index})}
+              >
+                <Markdown
+                  renderers={{paragraph: "span"}}
+                  source={getNodeDescription(selectedItem)}
+                />
+                <br />
+              </span>
+            )
+        )}
```
decentralion, line 118: So.. basically it's an implicit filter that depends on the fact that react filters falsey-results, and presumably getNodeDescription is returning undefined for the aliases?

Too hard to reason about as a newcomer to the code imo. how about `selectedItems.filter(...).map`

PR #2055, opened 2020-07-28
migrate credAccounts and identities to analysis

topocount:
the identity and credAccount modules' functionality spans both frontend and core use.
Moving it to analysis will allow us to better evaluate how we want to
split this functionality between these two purposes.

test-plan: This change should not modify any tests for functionality.
Unit tests, es-linting, and flow-linting are still passing.
Frontend requires manual testing to ensure nothing has been
broken.


decentralion:
I don't like the idea of moving identity into analysis. I like to think of the ledger as a relatively self-contained module, and the identity is the foundational data type for the ledger, so if we move it to analysis it stops being self-contained. (true, it does dep on analysis for the node type concept, but in my mind that's the fault of the concepts in analysis leaking over, rather than the ledger fundamentally depending on analysis). 

Similarly, my intended suggestion was not that we move computeCredAccounts wholesale into analysis, but rather that we split it up, into the part needed for `computeDistribution` and `applyDistribution` (which stays in ledger) and the part needed for arbitrary UI / data analysis purposes (which goes to ledger). One way to start this would be to fork computeCredAccounts so that there's an analysis version with a different name, then we could winnow down the original computeCredAccounts and add more functionality to the new version

PR #2066, opened 2020-07-30
Creditor Nodes: initial implementation

topocount:
Creditor nodes are utilized to store user-created cred minting actions.
The cannot be deleted, but they are modifiable. In memory they are
stored in the CredNodeStore, which enforces their structure and
heirarchical relationships with other nodes.

Like Initiatives, Creditor nodes are saved to disk in the instance repo.
Unlike Initiatives, they are not meant to be user-modifiable in this
state, but are rather meant to be updated via a web frontend. Diffing
changes is facilitated through use of jsonLog, which outputs each node
on a separate line.

test-plan: unit tests are provided for all methods in the CredNodeStore

### DIFF
--- a/src/plugins/creditor/credNodeStore.js
+++ b/src/plugins/creditor/credNodeStore.js
@@ -1,0 +1,100 @@
+// @flow
+
+import {type CreditorNode, type NodeUuid, type NodeTagId} from "./creditorNode";
+import {JsonLog} from "../../util/jsonLog";
+
+export type NodeEntry = {|type: "NODE", entry: CreditorNode|};
+export type NodeDetails = {title: string, id: NodeUuid};
+type NodesFile = JsonLog<NodeEntry>;
+
+export function toNodeEntry(entry: CreditorNode): NodeEntry {
+  return {type: "NODE", entry};
+}
+
+export function fromNodeEntry({entry}: NodeEntry): [NodeUuid, CreditorNode] {
+  return [entry.id, entry];
+}
+
+export function getDetails({id, title}: CreditorNode): NodeDetails {
+  return {id, title};
+}
+
+/**
+ * CredNodeStore is the static, in-memory representation of creditor nodes
+ *
+ * This class is used to accept and validate frontend inputs, and export them in
+ * serialized format utilizing the jsonLog functionality when saving to a file.
+ * Nodes in this class are still modifiable, of course, and jsonLogging is utilized
+ * because it outputs a gitdiff-friendly format
+ *
+ * Nodes in the the cred node store are not deletable at this time. In lieu of deleting nodes,
+ * their weights (`mint` values) can simply be set to zero
+ */
+export default class CredNodeStore {
+  +_nodes: Map<NodeUuid, CreditorNode>;
+
+  constructor(fromDisk: NodesFile | void) {
+    if (!fromDisk) {
+      this._nodes = new Map();
+    } else {
+      const nodeKVArray = Array.from(fromDisk.values()).map(fromNodeEntry);
+      this._nodes = new Map(nodeKVArray);
+    }
+  }
+
+  serialize(): string {
+    const log = new JsonLog();
+    const sortedEntries = Array.from(this.values()).sort(
+      ({createdAt: a}, {createdAt: b}) => a - b
+    );
+    return log.append(sortedEntries.map(toNodeEntry)).toString();
+  }
+
+  set(node: CreditorNode): CredNodeStore {
+    const {
+      id,
+      tags,
+      title,
+      description,
+      graphTimestamp,
+      createdAt,
+      parent,
+    } = node;
+    if (!id) {
+      throw new Error("Creditor Node UUID required");
+    }
+    if (!Array.isArray(tags)) throw new Error("`tags` must be an array");
+    if (!(title || description))
+      throw new Error("`title` and `description` are required in Cred Nodes");
+    if (!(graphTimestamp || createdAt))
+      throw new Error(
+        "`createdAt` and `graphTimestamp` values must be non-zero"
+      );
+    if (parent && !this._nodes.has(parent))
+      throw new Error("cannot add node with nonexistent parent");
+    this._nodes.set(node.id, node);
+    return this;
+  }
+
+  get(id: NodeUuid): ?CreditorNode {
+    return this._nodes.get(id);
+  }
+
+  values(): Iterator<CreditorNode> {
+    return this._nodes.values();
+  }
+
+  referencingTag(id: NodeTagId): NodeDetails[] {
+    return Array.from(this._nodes.values())
+      .filter(({tags}) => tags.includes(id))
+      .map(getDetails);
+  }
+
+  getParent(child: NodeUuid): ?CreditorNode {
+    const childNode = this._nodes.get(child);
+    if (!childNode)
+      throw new Error(`cred node with Id ${child} does not exist`);
+    if (!childNode.parent) return null;
+    return this._nodes.get(childNode.parent);
+  }
+}
--- a/src/plugins/creditor/credNodeStore.test.js
+++ b/src/plugins/creditor/credNodeStore.test.js
@@ -1,0 +1,124 @@
+// @flow
+
+import {type CreditorNode} from "./creditorNode";
+import Store, {getDetails, toNodeEntry} from "./credNodeStore";
+import {random as randomUuid} from "../../util/uuid";
+
+describe("plugins/creditor/credNodeStore", () => {
+  const tagId1 = randomUuid();
+  const tagId2 = randomUuid();
+  const baseNode: CreditorNode = {
+    id: randomUuid(),
+    tags: [tagId1, tagId2],
+    title: "Implementing credNodes",
+    description: "creating credNodes for use in the creditor",
+    graphTimestamp: new Date("1/1/2000").getTime(),
+    createdAt: new Date("12/1/2000").getTime(),
+    mint: 0,
+    parent: null,
+  };
+
+  const childNode: CreditorNode = {
+    id: randomUuid(),
+    tags: [tagId2],
+    title: "Creating credNode type",
+    description: "specifying credNodes to test the store",
+    graphTimestamp: new Date("1/1/2000").getTime(),
+    createdAt: new Date("11/1/2000").getTime(),
+    mint: 20,
+    parent: baseNode.id,
+  };
+
+  describe("adding nodes to store", () => {
+    it("requires nodes to have expected members", () => {
+      const store = new Store();
+      const thunk = (badNode: Object) => () => store.set(badNode);
+
+      const badNodes = [
+        [{}, "Creditor Node UUID required"],
+        [{id: randomUuid()}, "`tags` must be an array"],
+        [
+          {id: randomUuid(), tags: []},
+          "`title` and `description` are required in Cred Nodes",
+        ],
+        [
+          {
+            id: randomUuid(),
+            tags: [],
+            title: "bad Node",
+            description: "badDescription",
+            createdAt: 0,
+            graphTimestamp: 0,
+          },
+          "`createdAt` and `graphTimestamp` values must be non-zero",
+        ],
+      ];
+      badNodes.forEach(([node, errorMsg]) =>
+        expect(thunk(node)).toThrowError(errorMsg)
+      );
+    });
+    it("can set and get valid nodes", () => {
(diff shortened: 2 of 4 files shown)


src/plugins/creditor/declaration.js
```diff
@@ -0,0 +1,49 @@
...
+import deepFreeze from "deep-freeze";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {NodeType, EdgeType} from "../../analysis/types";
+import {NodeAddress, EdgeAddress} from "../../core/graph";
+
+export const nodePrefix = NodeAddress.fromParts(["sourcecred", "creditor"]);
+export const edgePrefix = EdgeAddress.fromParts(["sourcecred", "creditor"]);
+
+export const CreditorEntryType: NodeType = deepFreeze({
+  name: "CredNode",
```
decentralion, line 12: IMO calling it a "CredNode" is more confusing either than "Nodes" or "Creditor Nodes". 
I vote either "Creditor Node" or "Manual Node"

src/plugins/creditor/declaration.js
```diff
@@ -0,0 +1,49 @@
...
+  pluralName: "CredNodes",
+  prefix: NodeAddress.append(nodePrefix, "credNode"),
+  defaultWeight: 0,
+  description:
+    "An creditor node, describing a scoped improvement to a project from proposal to completion. \
+    Creditor nodes can be nested  arbitrarily (but non-circularly) to represent complet projects and scopes",
```
decentralion, line 18: Suggest "A node manually created by the Creditor". "Scoped improvement..." imo confuses the matter, esp, when we use Creditor Nodes for stuff like meetings. Talking about how they can be nested is just gonna confuse people.

Note: This text shows up in the UI (in the weights pane atm)

src/plugins/creditor/declaration.js
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * A contributor (src) CONTRIBUTES TO (verb) an Creditor node (dst).
+ * Forward: a contributor towards the entry node has a small endorsement of that
+ * contribution.
+ * Backward: flows the value of the contribution to the contributors.
+ */
+export const CreditorEdgeType: EdgeType = deepFreeze({
+  forwardName: "contributes to entry",
+  backwardName: "entry is contributed to by",
```
decentralion, line 36:
```suggestion
  forwardName: "connects to",
  backwardName: "is connected to",
```

I suggest calling it "connects" rather than "contributes" because contribute actually implies a unidirectional edge (it's confusing if "is contributed to" has meaningful Cred flow) but the Creditor connections could be bidirectional.

src/plugins/creditor/declaration.js
```diff
@@ -0,0 +1,49 @@
...
+  It follows the core/graph.js documentation to use
+  a <subject> <verb> <object> format to figure out
+  the directionality.
+*/
+
+/**
+ * A contributor (src) CONTRIBUTES TO (verb) an Creditor node (dst).
+ * Forward: a contributor towards the entry node has a small endorsement of that
+ * contribution.
+ * Backward: flows the value of the contribution to the contributors.
```
decentralion, line 32: I think we should start with a really generic creditor connection for connecting arbitrary Creditor Nodes, and then later add the specific case where we're connecting to a contribution.

Otherwise most potential connections are inexpressible.

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+    if (!(title || description))
+      throw new Error("`title` and `description` are required in Cred Nodes");
+    if (!(graphTimestamp || createdAt))
+      throw new Error(
+        "`createdAt` and `graphTimestamp` values must be non-zero"
+      );
+    if (parent && !this._nodes.has(parent))
+      throw new Error("cannot add node with nonexistent parent");
+    this._nodes.set(node.id, node);
+    return this;
```
decentralion, line 76: We don't usually do runtime validation this extensively, because we assume that Flow will catch issues like the id being unset. However, if we do want to validate, it would be much more comprehensive to use the parser module:

```js
import * as C from "../../util/combo"

const creditorNodeParser: C.Parser<CreditorNode> = C.object({
  id: uuid.parser,
  tags: C.array(uuid.parser),
  title: C.string,
  description: C.string,
  graphTimestamp: C.number,
...
})
```

However, in keeping with the rest of the codebase: I'd say use a parser only when we're actually reading from arbitrary JSON objects, and for here we could just check stuff like validating that the parent exists.

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+    } = node;
+    if (!id) {
+      throw new Error("Creditor Node UUID required");
+    }
+    if (!Array.isArray(tags)) throw new Error("`tags` must be an array");
+    if (!(title || description))
+      throw new Error("`title` and `description` are required in Cred Nodes");
+    if (!(graphTimestamp || createdAt))
+      throw new Error(
+        "`createdAt` and `graphTimestamp` values must be non-zero"
```
decentralion, line 71: `0` is a valid timestamp. We shouldn't fail on valid input data.

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
+// @flow
+
+import {type CreditorNode, type NodeUuid, type NodeTagId} from "./creditorNode";
+import {JsonLog} from "../../util/jsonLog";
+
+export type NodeEntry = {|type: "NODE", entry: CreditorNode|};
+export type NodeDetails = {title: string, id: NodeUuid};
```
decentralion, line 7: Why do we need this NodeDetails type? Why not just return the CreditorNodes directly in methods that return NodeDetails?

decentralion, line 7:
```suggestion
export type NodeEntry = {|+type: "NODE", +entry: CreditorNode|};
export type NodeDetails = {|+title: string, +id: NodeUuid|};
```

Prefer making types exact and readonly unless we specifically want diff behavior. Wish this was the default tbh.

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+  +_nodes: Map<NodeUuid, CreditorNode>;
+
+  constructor(fromDisk: NodesFile | void) {
+    if (!fromDisk) {
+      this._nodes = new Map();
+    } else {
+      const nodeKVArray = Array.from(fromDisk.values()).map(fromNodeEntry);
+      this._nodes = new Map(nodeKVArray);
+    }
+  }
```
decentralion, line 43: Directly setting _nodes is a bad pattern. We will want to track more pieces of state (e.g. tag and parent relationships), and if we directly set _nodes, then we'll need to re-implement the state tracking here. That implies a class of bugs where we update the state inconsistently. It will be much safer to call _addNodes.

To that end, let's not overload the constructor where it can take a NodesFile. All this does is introduce the opportunity to code subtle bugs where the state is different depending on whether we loaded the nodes file directly in the constructor, or called the `setNode` API. Instead, let's have the constructor take no arguments, and have a static method which creates a CredNodeStore from the NodesFile. See Graph for example:

https://github.com/sourcecred/sourcecred/blob/8ade386780654e69a10b134d67d2b9ac66ac1c79/src/core/graph.js#L236-L251

https://github.com/sourcecred/sourcecred/blob/8ade386780654e69a10b134d67d2b9ac66ac1c79/src/core/graph.js#L836-L865

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+
+  referencingTag(id: NodeTagId): NodeDetails[] {
+    return Array.from(this._nodes.values())
+      .filter(({tags}) => tags.includes(id))
+      .map(getDetails);
+  }
+
+  getParent(child: NodeUuid): ?CreditorNode {
+    const childNode = this._nodes.get(child);
```
decentralion, line 94: We have a public API for this. As a heuristic, I recommend building functionality using the public APIs rather than direct property access when feasible; it makes the module easier to refactor and easier to reason about. (E.g. if we need to refactor `this._nodes` into a different type to make certain kinds of access more efficient, we wouldn't need to re-write this method.)

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+    if (!childNode)
+      throw new Error(`cred node with Id ${child} does not exist`);
+    if (!childNode.parent) return null;
```
decentralion, line 97: This method has a wonky/inconsistent contract:
- If the child doesn't exist, it throws an error
- If the child exists but doesn't have a parent, it returns null
- If the child exists, and has a parent, but the parent doesn't exist, it returns undefined

I think we should be a bit more consistent, e.g. always returning either undefined or a node and never erroring.

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+      );
+    if (parent && !this._nodes.has(parent))
+      throw new Error("cannot add node with nonexistent parent");
+    this._nodes.set(node.id, node);
+    return this;
+  }
+
+  get(id: NodeUuid): ?CreditorNode {
+    return this._nodes.get(id);
+  }
```
decentralion, line 81: Based on past experience, I think we may want a `getOrFail` method too:

```js
getOrFail(id: NodeUuid): CreditorNode {
  const result = this.get(id)
  if (result == null) { throw new Error(`No node with id: ${id}`);}
  return result;
}
```

This is a very common query in practice, so it just saves the client from needing to implement the failure logic.

src/plugins/creditor/credNodeStore.js
```diff
@@ -0,0 +1,100 @@
...
+
+  values(): Iterator<CreditorNode> {
+    return this._nodes.values();
+  }
+
+  referencingTag(id: NodeTagId): NodeDetails[] {
+    return Array.from(this._nodes.values())
+      .filter(({tags}) => tags.includes(id))
+      .map(getDetails);
+  }
```
decentralion, line 91: I find this name a little confusing. We could call it `nodesMatchingTag` instead?

src/plugins/creditor/credNodeStore.test.js
```diff
@@ -0,0 +1,124 @@
+// @flow
+
+import {type CreditorNode} from "./creditorNode";
+import Store, {getDetails, toNodeEntry} from "./credNodeStore";
+import {random as randomUuid} from "../../util/uuid";
+
+describe("plugins/creditor/credNodeStore", () => {
+  const tagId1 = randomUuid();
+  const tagId2 = randomUuid();
```
decentralion, line 9: Speaking from experience: Using random UUIDs in test code will just create headaches, e.g. whenever we want to do snapshot serialization. I recommend something like the following:
https://github.com/sourcecred/sourcecred/blob/8ade386780654e69a10b134d67d2b9ac66ac1c79/src/ledger/ledger.test.js#L27-L43

src/plugins/creditor/credNodeStore.test.js
```diff
@@ -0,0 +1,124 @@
...
+      expect(nextStore).toEqual(store);
+      const returnedBase = store.get(baseNode.id);
+      expect(returnedBase).toEqual(baseNode);
+    });
+    it("returns null for nonexistent nodes", () => {
+      const store = new Store();
+      const badId = randomUuid();
+      const result = store.get(badId);
+      expect(result).toBeEmpty;
+    });
```
decentralion, line 72: This test does not do what it says on the tin. The store actually returns `undefined`, but the test passes.

A correct implementation of the test (which would fail) would be:
`expect(store.get(badId)).toBe(null)`

It's really dumb that JS has null and undefined as distinct values, but it does. As a heuristic, when we're writing our code we almost always prefer using explicit null, except in the case where we are implementing `Map.get` esque-semantics, in which case we return `undefined` to be consistent with the language primitives.

src/plugins/creditor/credNodeStore.test.js
```diff
@@ -0,0 +1,124 @@
...
+  });
+  describe("node interactions", () => {
```
decentralion, line 74: we need tests verifying the semantics for replacing a node (same id, new props)
we also need to test how replacement interacts with other API methods. for example, if we have a node with tag Foo, and we replace that node so it now has tag Bar, check that it no longer shows up on a tag search for Foo but it does show up on a tag search for Bar

PR #2071, opened 2020-07-30
implement lint-staged

topocount:
lint-staged will help speed up the development process by failing on
commits if any of the repo's linters detect errors that can't be autofixed.

One consideration: flow has been added to lint-staged even though it's
not technically sensitive to git's stage. It will error if _any_ files
in the repo violate a flow rule. The flow check can be removed if we
find this to be irksome.

These checks can always be bypassed by running `git commit --no-verify`

test plan: there should be no unexpected errors from our linters

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -1,0 +1,6 @@
+// @flow
+module.exports = {
+  "{src,config}/**/*.js": "eslint --fix --max-warnings 0",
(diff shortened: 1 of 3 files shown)


topocount:
> We found it more convenient to configure our editors to automatically lint on every save

I have found this to be true myself. I'll close this for now.

PR #2077, opened 2020-08-02
jsonLog: let `append` act on values, not sequences

wchargin:
Summary:
On list-like collections, `append` generally suggests adding a single
element, as with Python’s `list.append`. As evidence that this is more
natural, note that both non-test uses of `JsonLog.append` passed a
singleton sequence. We retain `extend` in case it’s helpful; for long
sequences, it may be very slightly faster.

Test Plan:
Unit tests suffice.

wchargin-branch: jsonlog-append

### DIFF
--- a/src/ledger/ledger.js
+++ b/src/ledger/ledger.js
@@ -495,7 +495,7 @@ export class Ledger {
     }
     this._processAction(action);
     this._latestTimestamp = ledgerTimestamp;
-    this._ledgerEventLog.append([e]);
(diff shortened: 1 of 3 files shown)


wchargin:
For your consideration? I don’t feel super strongly about this, but
while writing #2075, I was surprised that `.append(v)` didn’t
type-check, and thought that this would be more natural.

wchargin:
(thanks for closing this out!)

PR #2079, opened 2020-08-05
Create AccountDropdown which uses material-ui

befitsandpiper:
update TransferGrain to use new dropdown

Test Plan: manual frontend testing

### DIFF
--- a/src/ui/components/AccountSelector.js
+++ b/src/ui/components/AccountSelector.js
@@ -37,7 +37,6 @@ export default function AccountDropdown({
       <Autocomplete
         onChange={onComboChange}
         fullWidth
-        id="combo-box-demo"
         options={items}
         getOptionLabel={(item) =>
           `${item.identity.name} (${format(item.balance, 2)})`
--- a/src/ui/components/AccountSelector.js
+++ b/src/ui/components/AccountSelector.js
@@ -1,102 +1,52 @@
 // @flow
 
-import React, {useState} from "react";
+import React from "react";
+import {createMuiTheme, ThemeProvider} from "@material-ui/core/styles";
+import {TextField} from "@material-ui/core";
+import {format} from "../../ledger/grain";
+import {Autocomplete} from "@material-ui/lab";
 import {Ledger, type Account} from "../../ledger/ledger";
-import {useCombobox} from "downshift";
 
 type DropdownProps = {|
   +ledger: Ledger,
-  +setCurrentIdentity: (Account) => void,
+  +setCurrentIdentity: (Account | null) => void,
+  +placeholder?: string,
 |};
 
+const theme = createMuiTheme({
+  overrides: {
+    // Name of the rule
+    text: {
+      // Some CSS
+      color: "black",
+    },
+  },
+});
+
 export default function AccountDropdown({
+  placeholder,
   setCurrentIdentity,
   ledger,
 }: DropdownProps) {
   const items = ledger.accounts().filter((a) => a.active);
-  const [inputItems, setInputItems] = useState<$ReadOnlyArray<Account>>(items);
-  const [inputValue, setInputValue] = useState<string>("");
-  const {
-    isOpen,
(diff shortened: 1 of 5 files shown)


sharness/__snapshots__/test-instance/data/ledger.json
```diff
@@ -5,4 +5,4 @@
 {"action":{"identityId":"mbzRr0D3A2wcD1E7p0H2rQ","type":"TOGGLE_ACTIVATION"},"ledgerTimestamp":1595960684984,"uuid":"lXJfGAKpX2C7eV7WQ13I8A","version":"1"},
```
decentralion, line 5: Irrelevant change, please revert

src/ui/components/TransferGrain.js
```diff
@@ -11,7 +11,7 @@ import {
 } from "../../ledger/grain";
 import {type Identity} from "../../ledger/identity";
 import {Ledger, type Account} from "../../ledger/ledger";
-import AccountDropdown from "./AccountSelector";
+import AccountDropdown from "./AccountDropdown";
```
decentralion, line 14: Looks like we added AccountDropdown without removing AccountSelector. Is there a reason we need both?

src/ui/components/TransferGrain.js
```diff
@@ -25,13 +25,13 @@ export const TransferGrain = ({ledger, setLedger}: Props) => {
   const [maxAmount, setMaxAmount] = useState<Grain>(fromInteger(0));
   const [memo, setMemo] = useState<string>("");
 
-  const setSender = (acct: Account) => {
-    setMaxAmount(acct.balance);
-    setSourceIdentity(acct.identity);
+  const setSender = (acct: Account | null) => {
```
decentralion, line 28: Why the change in signature?

src/ui/components/TransferGrain.js
befitsandpiper: It is used on other pages still. But I could make sure everything works with those pages too. Just figured that would lead to styling pages on other pages and this branch quickly getting out of scope.

src/ui/components/TransferGrain.js
befitsandpiper: I felt it was necessary to be able to set the account back to the null default if you cleared the combobox field. This disables the grain transfer button again and prevents unexpected behavior.

src/ui/components/TransferGrain.js
decentralion: Yeah, let's update in both places please. I think the PR will still be small & easy to review, and that way we guarantee we aren't adding debt by having two components that do nearly the same thing. It also "validates" the new component a bit by showing that we can re-use it.

befitsandpiper: Okay, I am using this in both places now. This did change how the SpecialGrainDistribution page looks a bit, but did not affect usability, and there will be styling updates there as well in the future. I didn't want to include css changes in this commit.

src/ui/components/TransferGrain.js
decentralion: Cool, makes sense.

befitsandpiper:
it is full width rn because I haven't applied new styling. This is just switching to material-ui. I have improved styling as well for another commit.
looks more like this:
Screenshot from 2020-08-04 19-50-30

src/ui/components/AccountSelector.js
```diff
@@ -1,102 +1,52 @@
...
-              {item.identity.name}
-            </li>
-          ))}
-      </ul>
-    </div>
+    <ThemeProvider theme={theme}>
+      <Autocomplete
+        onChange={onComboChange}
+        fullWidth
+        id="combo-box-demo"
```
decentralion, line 40: nit: the id looks vestigial

befitsandpiper: ah yes, oops

PR #2080, opened 2020-08-05
Ledger: support changing identity subtypes

decentralion:
We support a few different identity subtypes (USER, PROJECT,
ORGANIZATION, BOT, etc). Right now these subtypes aren't really used,
but we'll likely use them in UI filtering in the future.

We'll need to support changing these, much like we do identity names.
This commit adds a new CHANGE_IDENTITY_SUBTYPE action and tests support
for it.

Test plan: Unit tests updated; `yarn test` passes.

### DIFF
--- a/src/ledger/ledger.test.js
+++ b/src/ledger/ledger.test.js
@@ -273,6 +273,73 @@ describe("ledger/ledger", () => {
       });
     });
 
+    describe("changeIdentityType", () => {
+      it("works", () => {
+        const ledger = new Ledger();
+        setFakeDate(0);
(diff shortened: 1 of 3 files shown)


src/ledger/ledger.test.js
```diff
@@ -214,6 +214,73 @@ describe("ledger/ledger", () => {
       });
     });
 
+    describe("changeIdentitySubtype", () => {
+      it("works", () => {
```
hammadj, line 218: something more descriptive of the functionality would be better

```suggestion
      it("changes a USER identity type into a PROJECT identity type", () => {
```

PR #2082, opened 2020-08-06
Add a module for typesafe numerics

decentralion:
This commit adds the `util/numerics` module which exposes typesafe
numeric sub-types, such as Finite, FiniteNonnegative, Integer,
NonnegativeInteger, and Proportion. These types all have an associated
parsing method, and a parser.

Test plan: See included unit tests, a followon PR will refactor the
codebase to use these types.

### DIFF
--- a/src/util/numerics.js
+++ b/src/util/numerics.js
@@ -1,0 +1,70 @@
+// @flow
+
+import * as C from "./combo";
+
+/**
+ * A library for making type-level assertions about numeric properties;
+ * e.g. that a number is finite, or finite and non-negative, or an integer.
+ *
+ * We should avoid having these methods in numeric hot loops by doing "unsafe"
+ * operations in the hot loop, and then validating on exiting the loop.
+ *
+ * Nonetheless, we avoid unnecessary function calls for performance reasons (i.e.
+ * the validator functions avoid calling each other).
+ */
+
+// A number guaranteed to not be NaN, Infinity, or -Infinity
+export opaque type Finite: number = number;
+export function finite(n: number): Finite {
+  if (!isFinite(n)) {
+    throw new Error(`expected finite, got: ${n}`);
+  }
+  return n;
+}
+export const finiteParser: C.Parser<Finite> = C.fmap(C.number, finite);
+
+export opaque type FiniteNonnegative: number = number;
+export function finiteNonnegative(n: number): FiniteNonnegative {
+  if (n < 0 || !isFinite(n)) {
+    throw new Error(`expected finite nonnegative, got: ${n}`);
+  }
+  return n;
+}
+export const finiteNonnegativeParser: C.Parser<FiniteNonnegative> = C.fmap(
+  C.number,
+  finiteNonnegative
+);
+
+export opaque type Integer: number = number;
+export function integer(n: number): Integer {
+  if (!isFinite(n) || n !== Math.floor(n)) {
+    throw new Error(`expected integer, got: ${n}`);
+  }
+  return n;
+}
+export const integerParser: C.Parser<Integer> = C.fmap(C.number, integer);
+
+export opaque type NonnegativeInteger: number = number;
+export function nonnegativeInteger(n: number): NonnegativeInteger {
+  if (!isFinite(n) || n !== Math.floor(n) || n < 0) {
+    throw new Error(`expected nonnegative integer, got: ${n}`);
+  }
+  return n;
+}
+export const nonnegativeIntegerParser: C.Parser<NonnegativeInteger> = C.fmap(
+  C.number,
+  nonnegativeInteger
+);
+
+// A number guaranteed to be in the range [0, 1]
+export opaque type Proportion: number = number;
+export function proportion(n: number): Proportion {
+  if (!isFinite(n) || n < 0 || n > 1) {
+    throw new Error(`expected proportion in [0, 1], got: ${n}`);
+  }
+  return n;
+}
+export const proportionParser: C.Parser<Proportion> = C.fmap(
+  C.number,
+  proportion
+);
--- a/src/util/numerics.test.js
+++ b/src/util/numerics.test.js
@@ -1,0 +1,95 @@
+// @flow
+
+import * as N from "./numerics";
+
+describe("util/numerics", () => {
+  describe("finite", () => {
+    it("allows finite numbers", () => {
+      for (const f of [-1.2, 0, 999]) {
+        expect(N.finite(f)).toEqual(f);
+        expect(N.finiteParser.parseOrThrow(f)).toEqual(f);
+      }
+    });
+    it("disallows non-finite numbers", () => {
+      for (const f of [NaN, Infinity, -Infinity]) {
+        expect(() => N.finite(f)).toThrow("expected finite, got:");
+        expect(() => N.finiteParser.parseOrThrow(f)).toThrow(
+          "expected finite, got:"
+        );
+      }
+    });
+  });
+  describe("finiteNonnegative", () => {
+    it("allows finite nonnegative numbers", () => {
+      for (const f of [0, 0.12, 999]) {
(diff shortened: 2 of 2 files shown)


decentralion:
See #2083 for the followon which uses this across the codebase. Not requesting review on it because it's noisy and not super interesting.

src/util/numerics.js
```diff
@@ -0,0 +1,70 @@
...
+  return n;
+}
+export const finiteNonnegativeParser: C.Parser<FiniteNonnegative> = C.fmap(
+  C.number,
+  finiteNonnegative
+);
+
+export opaque type Integer: number = number;
+export function integer(n: number): Integer {
+  if (!isFinite(n) || n !== Math.floor(n)) {
```
wchargin, line 40: Can we use `Number.isSafeInteger` here? It’s hard to think of cases
where you want an argument to be an integer but don’t care whether it’s
a safe integer. For instance, naïve code might try to represent a 64-bit
uint as a JavaScript number, which would lose precision for 99.95% of
values; we should catch that. Or, from another perspective: any type
that does not have an injective “successor” operation consistent with
the “add 1” operation should not be called an integer type.

If you actually need such large integers, you should use bigints or
similar.

(also, fyi: `Number.isInteger`.)

src/util/numerics.js
```diff
@@ -0,0 +1,70 @@
...
+ * operations in the hot loop, and then validating on exiting the loop.
+ *
+ * Nonetheless, we avoid unnecessary function calls for performance reasons (i.e.
+ * the validator functions avoid calling each other).
+ */
+
+// A number guaranteed to not be NaN, Infinity, or -Infinity
+export opaque type Finite: number = number;
+export function finite(n: number): Finite {
+  if (!isFinite(n)) {
```
wchargin, line 19: In case it’s new to you:

```javascript
Number.isFinite("3"); // false (it's a string, of course)
isFinite("3"); // true...
```

Here we have `n: number`, but the function is exported, so it’s not
quite right if the Flow types are unsound—e.g., if an untyped client
grabs us from NPM. (Even if it were guaranteed safe, I’d recommend using
`Number.isFinite` anyway to avoid having to prove this each time I read
the line.)

Likewise below.

src/util/numerics.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+  return n;
+}
+export const nonnegativeIntegerParser: C.Parser<NonnegativeInteger> = C.fmap(
+  C.number,
+  nonnegativeInteger
+);
+
+// A number guaranteed to be in the range [0, 1]
+export opaque type Proportion: number = number;
```
wchargin, line 60: So this is an interesting one. :-)

First, I don’t think that _proportion_ is the right name. From context
in this module, I’d expect the name _proportion_ to be used in the
mathematical sense, but I can’t link that to its usage here. Constants
of proportionality can be greater than one, or negative. _Ratios_ can
also of course be greater than one.

Now, from context broader than this PR, I expect that you intend to use
this _proportion_ more like _share_ (“…of the pie”). In that case, it
seems that actually _portion_ (not _proportion_) fits quite nicely? (It
actually just seems more fitting to me overall; non-mathematically,
proportions are about how things relate to each other, like aspect
ratios or measurements, whereas portions are the things that get divvied
up.)

That said, I wonder whether you’ll end up wanting this type at all.
For validating portions, won’t you always want to verify that the
distribution sums to unity, or at least does not exceed unity? in which
case it suffices to take `FiniteNonnegative`s and assert on the sum.

wchargin:
By the way, in the past I’ve defined a `Probability` type for `[0, 1]`,
and found it awkward specifically dealing with whether the endpoints are
included or excluded in certain contexts. Basically, you want to be able
to represent both `Pr[E] = 0` and `Pr[E] = 1`, but when you “roll” from
a distribution you want to roll strictly in `[0, 1)`. So I ended up with
comments like “args: `x: Pr` should be strictly less than `1`”, which
was less than perfect.

wchargin:
One such usage:
<https://gist.github.com/wchargin/c1d9bf2eacf4d896dc62e772f9f7ee51#file-droptable-rs-L8-L9>

src/util/numerics.js
```diff
@@ -0,0 +1,70 @@
...
+  if (!isFinite(n) || n !== Math.floor(n)) {
+    throw new Error(`expected integer, got: ${n}`);
+  }
+  return n;
+}
+export const integerParser: C.Parser<Integer> = C.fmap(C.number, integer);
+
+export opaque type NonnegativeInteger: number = number;
+export function nonnegativeInteger(n: number): NonnegativeInteger {
+  if (!isFinite(n) || n !== Math.floor(n) || n < 0) {
```
hammadj, line 49: More readable to re-use the existing helpers:
```suggestion
  if (!integer(n) || n < 0) {
```

A nice side effect of this is more specific error messages. If its not an integer they get an error "expected integer", but if it's a negative integer they get an error "expected nonnegative integer".

src/util/numerics.js
hammadj: I like "portion", but if we want a more "official term" for this, could use "proper fraction" ? Still prefer "portion" though

src/util/numerics.test.js
```diff
@@ -0,0 +1,95 @@
+// @flow
+
+import * as N from "./numerics";
+
+describe("util/numerics", () => {
+  describe("finite", () => {
+    it("allows finite numbers", () => {
+      for (const f of [-1.2, 0, 999]) {
```
hammadj, line 8: Would be good to test out some irrational numbers too like Math.PI and repeating decimals like `5/7`.

src/util/numerics.test.js
```diff
@@ -0,0 +1,95 @@
...
+        expect(() => N.finiteNonnegativeParser.parseOrThrow(f)).toThrow(
+          "expected finite nonnegative, got:"
+        );
+      }
+    });
+  });
+
+  describe("integer", () => {
+    it("allows integers", () => {
+      for (const f of [-12, 0, 13]) {
```
hammadj, line 43: Could throw in `Number.MAX_SAFE_INTEGER` in there as well

src/util/numerics.test.js
```diff
@@ -0,0 +1,95 @@
...
+        expect(N.integer(f)).toEqual(f);
+        expect(N.integerParser.parseOrThrow(f)).toEqual(f);
+      }
+    });
+    it("disallows non-integers", () => {
+      for (const f of [NaN, Infinity, -Infinity, -1.23, 14.91]) {
```
hammadj, line 49: Could throw in `Number.MIN_VALUE` in there

src/util/numerics.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+  describe("proportion", () => {
+    it("allows finite numbers in [0, 1]", () => {
+      for (const f of [0, 0.3, 1]) {
+        expect(N.proportion(f)).toEqual(f);
+        expect(N.proportionParser.parseOrThrow(f)).toEqual(f);
+      }
+    });
+    it("disallows numbers outside [0, 1]", () => {
+      for (const f of [NaN, Infinity, -Infinity, -13, -1.23, 1.01]) {
```
hammadj, line 85: Could throw in `1 + Number.MIN_VALUE`

src/util/numerics.test.js
```diff
@@ -0,0 +1,95 @@
...
+        expect(() => N.nonnegativeIntegerParser.parseOrThrow(f)).toThrow(
+          "expected nonnegative integer, got:"
+        );
+      }
+    });
+  });
+
+  describe("proportion", () => {
+    it("allows finite numbers in [0, 1]", () => {
+      for (const f of [0, 0.3, 1]) {
```
hammadj, line 79: Could throw in `5/9`

decentralion:
Closing. I think I'll go with the lower-touch approach later where we just add the parsers.

PR #2088, opened 2020-08-06
First pass transferGrain styles

befitsandpiper:
Test Plan: manual FE testing


src/ui/components/TransferGrain.js
```diff
@@ -51,69 +106,90 @@ export const TransferGrain = ({ledger, setLedger}: Props) => {
...
+        </div>
+        <div className={classes.centerRow}>
+          <div
+            className={`${classes.dropdown} ${classes.centerRow} ${classes.third}`}
+          >
+            <AccountDropdown
+              ledger={ledger}
+              setCurrentIdentity={setSender}
+              placeholder="From..."
+            />
```
topocount, line 122: I like the UX of rendering the balance in this component, but after a transfer occurs, we're still left with the balance prior to the transfer. We'll need to utilize useEffect or ideally just utilize props directly to render out the balance so that the render isn't one step behind the actual state. I had to troubleshoot a similar issue where the component state is one render behind the latest prop value and this was my solution:
https://github.com/sourcecred/sourcecred/blob/74b4df4ab628968a5b88b81b1d18fc19c919dc03/src/ui/components/AliasSelector.js#L62-L64

This looks like a bug that snug in through #2079

befitsandpiper:
Lots of the issues you are seeing are because your screen is smaller than mine. Not a good excuse since I did overlook this though. I will update the designs and test on more screen sizes, put in some breakpoints and stuff. I can also just set it to be full width. I think the transferGrain area is only like 60% of the screen right now.

This is also not the final design, but based on Faruk's whimsical mockup. The components are roughly correct, but the css and colors will probably be tweaked quite a bit. Faruk is currently working a more final Figma design. I will update the page though. Is there a minimum screen resolution we want to be targeting here with this UI?

decentralion:
I'm on a 13'' MBP (2560 x 1600). I don't have a particular minimum screen resolution in mind, but I do think it should render cleanly on 13'' laptops. :)

befitsandpiper:
I changed the width of the transferGrain box from being a percent of screen to being a fixed pixel width. This should fix most of the issues you're seeing on most sized screens. Some other minor UI tweaks as well to fix the issues you mentioned:
(sorry my screenshot overlays a gray box on everything).

We'll be updating designs again soon, I hope this is good enough for now. Switching to a vertical flow would require a major overhaul of this work.

Screenshot from 2020-08-07 14-20-44

PR #2089, opened 2020-08-07
First pass transferGrain styles

befitsandpiper:
Revision of this closed PR: https://github.com/sourcecred/sourcecred/pull/2088

I changed the width of the transferGrain box from being a percent of screen to being a fixed pixel width. This should fix most of the style breaking issues seen before on smaller screen sizes. Some other minor UI tweaks as well to fix the issues mentioned [here](https://github.com/sourcecred/sourcecred/pull/2088).
(sorry my screenshot service overlays a gray box on everything).

Faruk will be updating designs again soon, I hope this is good enough for now. Switching to a vertical flow would require a major overhaul of this work.

Screenshot from 2020-08-07 14-20-44


src/ui/components/TransferGrain.js
```diff
@@ -18,19 +20,74 @@ export type Props = {|
...
+  },
+  memo: {
+    width: "960px",
+    height: "50px",
+    border: "1px solid #9EADBA",
+    paddingLeft: "20px",
+    color: "black",
+  },
+});
+
```
KuraFire, line 74: Is this the bulk of all the styling work on this UI (I saw small bits of CSS along the way, too), or is there more elsewhere?

befitsandpiper: This is the bulk, there are still a few one-off inline styles, but that's all. There is also the one theme declaration in the AccountSelector file. I am not very good with themes, but we will probably want to create custom theming for the material-ui components which we keep in some central file.

src/ui/components/AccountSelector.js
```diff
@@ -27,24 +26,46 @@ export default function AccountDropdown({
...
-        style={{margin: "20px"}}
-        renderInput={(params) => (
-          <TextField fullWidth {...params} label={placeholder} />
-        )}
+        getOptionLabel={(item) => {
+          const balStr = format(
+            balance !== fromInteger(0) && balance ? balance : item.balance,
+            2
+          );
+          return `${item.identity.name} (${balStr})`;
```
decentralion, line 65: Can we give an explicit type for `item`?

src/ui/components/AccountSelector.js
```diff
@@ -27,24 +26,46 @@ export default function AccountDropdown({
   placeholder,
   setCurrentIdentity,
   ledger,
+  balance,
 }: DropdownProps) {
   const items = ledger.accounts().filter((a) => a.active);
+  const [selectedAccount, setSelectedAccount] = useState<Account | null>(null);
```
decentralion, line 32: Looks like we can either set the balance explicitly as a prop, or we can get it implicitly from the account. Why both? I'd prefer to have a single source of truth for the balance (e.g. the account)

src/ui/components/TransferGrain.js
```diff
@@ -18,19 +20,74 @@ export type Props = {|
...
   const [destIdentity, setDestIdentity] = useState<Identity | null>(null);
+  const [destBalance, setDestBalance] = useState<Grain>(fromInteger(0));
   const [amount, setAmount] = useState<string>("0");
-  const [maxAmount, setMaxAmount] = useState<Grain>(fromInteger(0));
   const [memo, setMemo] = useState<string>("");
 
   const setSender = (acct: Account | null) => {
-    setMaxAmount(acct ? acct.balance : fromInteger(0));
+    setSourceBalance(acct ? acct.balance : fromInteger(0));
     setSourceIdentity(acct ? acct.identity : null);
```
decentralion, line 86: So we set the balance and the identity in separate calls -- why not just set the account directly, since it has both?

src/ui/components/AccountSelector.js
befitsandpiper: yes

src/ui/components/AccountSelector.js
befitsandpiper: Yes you are right. I came to this solution in a roundabout way trying to fix the balance updating issue topocount saw, but after adding this selectedAccount state line, the balance prop is redundant.

src/ui/components/TransferGrain.js
befitsandpiper: sure, done

src/ui/components/AccountSelector.js
```diff
@@ -29,22 +27,45 @@ export default function AccountDropdown({
   ledger,
 }: DropdownProps) {
   const items = ledger.accounts().filter((a) => a.active);
+  const [selectedAccount, setSelectedAccount] = useState<Account | null>(null);
 
-  const onComboChange = (e, inputObj) => setCurrentIdentity(inputObj);
+  const onComboChange = (e, inputObj) => {
+    setSelectedAccount(inputObj);
+    setCurrentIdentity(inputObj);
```
decentralion, line 34: similar question as before--why are we calling setCurrentIdentity and setSelectedAccount? the account includes the identity

src/ui/components/AccountSelector.js
```diff
@@ -29,22 +27,45 @@ export default function AccountDropdown({
...
-        }
-        style={{margin: "20px"}}
-        renderInput={(params) => (
-          <TextField fullWidth {...params} label={placeholder} />
-        )}
+        getOptionLabel={(item: Account) => {
+          const balStr = format(
+            selectedAccount && selectedAccount.balance !== fromInteger(0)
+              ? selectedAccount.balance
+              : item.balance,
```
decentralion, line 62: Can you help me understand what's the distinction between the selectedAccount and the item, and why we would want to render the balance for the selectedAccount rather than the item?

src/ui/components/TransferGrain.js
```diff
@@ -18,102 +18,192 @@ export type Props = {|
...
   const setSender = (acct: Account | null) => {
-    setMaxAmount(acct ? acct.balance : fromInteger(0));
-    setSourceIdentity(acct ? acct.identity : null);
+    setSourceAccount(acct);
   };
 
   const setReceiver = (acct: Account | null) => {
-    setDestIdentity(acct ? acct.identity : null);
+    setDestAccount(acct);
   };
```
decentralion, line 86: Now we should be able to remove these declarations entirely, and just set them as `const [sender, setSender] = useState<Account|null>(null)`. Since this assignment reduces to `const setSender = setSourceAccount`.

src/ui/components/TransferGrain.js
```diff
@@ -18,102 +18,192 @@ export type Props = {|
...
 
   const submitTransfer = (e) => {
     e.preventDefault();
-    if (sourceIdentity && destIdentity) {
+    if (
+      sourceAccount &&
+      sourceAccount.identity &&
```
decentralion, line 92: `sourceAccount` implies `sourceAccount.identity`. Per the type, identity on an account is non-nullable

src/ui/components/TransferGrain.js
```diff
@@ -18,102 +18,192 @@ export type Props = {|
...
+                  name="amount"
+                  min="0"
+                  placeholder="(amount)"
+                  value={amount !== "0" ? amount : ""}
+                  onChange={(e) => setAmount(e.currentTarget.value)}
+                />
+              </div>
+              <span>
+                {sourceAccount &&
+                  sourceAccount.identity &&
```
decentralion, line 142: as above: checking the identity on an account is redundant

src/ui/components/TransferGrain.js
```diff
@@ -18,102 +18,192 @@ export type Props = {|
...
+              </span>
+            </div>
+            <div className={classes.triangle} />
+          </div>
+          <div
+            className={`${classes.dropdown} ${classes.centerRow} ${classes.third}`}
+          >
+            <AccountDropdown
+              ledger={ledger}
+              setCurrentIdentity={setReceiver}
```
decentralion, line 153: finding it confusing that it looks like `setCurrentIdentity` actually takes an account, not an identity

src/ui/components/TransferGrain.js
```diff
@@ -18,102 +18,192 @@ export type Props = {|
...
+            onClick={() => {
+              fetch("data/ledger.json", {
+                headers: {
+                  Accept: "text/plain",
+                  "Content-Type": "text/plain",
+                },
+                method: "POST",
+                body: ledger.serialize(),
+              });
+            }}
```
decentralion, line 201: We should abstract over this pattern so that we create a `persistLedger` method once at the app top level and then pass the function down. No change requested in this PR, it's a separate cleanup, just noting that this is violating DRY.

src/ui/components/AccountSelector.js
befitsandpiper: This is primarily to have the balances on the menu items within the combobox prior to selecting an item. The way material-ui works the labels inside the dropdown menu are the same as the values displayed on select, so this was the most sensible way to do this I think. The item though does not update when the the balances change, so we need to use the selectedAccount for that.

src/ui/components/AccountSelector.js
befitsandpiper: This one is actually necessary as far as I can tell in order to update the balances in the child, since using the setCurrentIdentity only sets the identity at the parent level, and doesn't update the UI for the child level.

befitsandpiper: I will make a second pass today to see if there is a better way to do this. Because I agree if does seem strange.

src/ui/components/AccountSelector.js
befitsandpiper: So there were some weird sync issues I was seeing after transferring grain, and having account balances adjust. Removing balance logiv from the Autocomplete components fixed everything and cleaned up the code a lot, getting rid of these ugly code patterns you saw in the process.

src/ui/components/AccountSelector.js
befitsandpiper: Managed to get rid of this by removing balance displays from the Autocomplete component

befitsandpiper:
Slight change to the look, removing balances from inside Autocomplete components
Screenshot from 2020-08-08 18-20-19

src/ui/components/AccountSelector.js
```diff
@@ -1,47 +1,39 @@
...
-        onChange={onComboChange}
+        onChange={(e, inputObj) => setCurrentAccount(inputObj)}
         fullWidth
         options={items}
-        getOptionLabel={(item) =>
-          `${item.identity.name} (${format(item.balance, 2)})`
-        }
-        style={{margin: "20px"}}
+        getOptionLabel={(item: Account) => item.identity.name}
+        style={{margin: "20px", marginTop: "-16px"}}
```
hammadj, line 36: Why the negative top margin?

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
-          <input
-            type="number"
-            name="amount"
-            min="0"
-            step="any"
-            placeholder={`max: ${maxAmount}`}
-            required
-            value={amount}
-            onChange={(e) => setAmount(e.currentTarget.value)}
+    <div className={classes.root}>
```
hammadj, line 96: Can just use the Container component from MUI: https://material-ui.com/components/container/

If you need to set a maxWidth or something, you can do so from that components props. That way the look of this page can follow the theme instead of having a white BG

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
+          disableUnderline
+          type="text"
+          value={memo}
+          placeholder="Memo"
+          onChange={(e) => setMemo(e.currentTarget.value)}
+        />
+      </div>
+      <div className={classes.centerRow}>
+        <Button
+          className={`${classes.centerRow} ${classes.half}`}
```
hammadj, line 149: Why are the buttons centerRow?

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
+          style={{background: "#6558F5"}}
```
hammadj, line 150: Shouldn't hardcode colours like this. Instead, use the color prop so it can pull values from the theme: 
```suggestion
          color="primary"
```

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
-        <input
-          type="button"
-          value="save ledger to disk"
+          onClick={submitTransfer}
+        >
+          transfer grain
+        </Button>
+        <Button
+          className={`${classes.centerRow} ${classes.half}`}
+          style={{background: "#C9CED2"}}
```
hammadj, line 163: Default button is grey, shouldn't need to set this

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
+  third: {
+    width: "300px",
+    height: "50px",
+    margin: "30px 15px",
+  },
+  half: {
+    width: "450px",
+    height: "50px",
+    margin: "30px",
+  },
```
hammadj, line 62: Hardcoding sizes like this is pretty fragile. If you want something to be 50% width, you just just set the container to `display: flex` and the buttons inside to have `flex: 1` so that they automatically take up half the containers width each. That way, if a button is added or removed, you dont have to change anything, whereas with this you have to go around and update all these sizes manually.

In addition, instead of setting the height manually, you can use the `size` prop and set it to `size="large"`

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
+  memo: {
+    width: "960px",
+    height: "50px",
```
hammadj, line 65: Same with this, shouldn't hardcode sizes, just use %'s or flex box

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
+    border: "1px solid #9EADBA",
+    paddingLeft: "20px",
+    color: "black",
```
hammadj, line 68: Instead of setting the border like this, just use the `variant="outlined"` prop of the [`TextField`](https://material-ui.com/components/text-fields/#textfield) component. Should probably not use the Input component directly from material-ui

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
...
+  centerRow: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+  },
+  third: {
+    width: "300px",
+    height: "50px",
+    margin: "30px 15px",
+  },
```
hammadj, line 57: Similarly, shouldn't hardcode pixel values like this, use flex box instead

src/ui/components/TransferGrain.js
befitsandpiper: you're right, this is redundant.

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
   +setLedger: (Ledger) => void,
 |};
 
+const useStyles = makeStyles({
+  root: {
+    width: "1100px",
+    margin: "0 auto",
+    background: "white",
```
hammadj, line 24: Why is the background being set to white? Should just follow the theme so its consistent with the rest of the UI and its easy to update the look globally. Similarly, we shouldn't be setting all the text to black.

src/ui/components/AccountSelector.js
befitsandpiper: probably sufficient for this top margin to just be zero.

src/ui/components/TransferGrain.js
befitsandpiper: I wanted the background to be white though, I was basing this UI off of Faruk's whimsical mockup: https://whimsical.com/A6EEdJgLwX3K4VTroURP3b@2Ux7TurymMnWnyv6hKCm. 
If the BG isn't white everything else looks very strange and the colors all have to change.

src/ui/components/TransferGrain.js
```diff
@@ -18,90 +17,151 @@ export type Props = {|
   +setLedger: (Ledger) => void,
 |};
 
+const useStyles = makeStyles({
```
hammadj, line 20: If you need to access values from the theme, you can do this:
```suggestion
const useStyles = makeStyles((theme) => ({
  root: {
     background: theme.palette.background,
```

Just an example to illustrate, you probably dont even need to set the BG if you just use the provided components from MUI.

befitsandpiper:
@hammadj So was this more the look you were thinking?

Screenshot from 2020-08-09 22-26-59

src/ui/components/AccountSelector.js
```diff
@@ -1,47 +1,39 @@
...
 }: DropdownProps) {
   const items = ledger.accounts().filter((a) => a.active);
 
-  const onComboChange = (e, inputObj) => setCurrentIdentity(inputObj);
-
   return (
     <ThemeProvider theme={theme}>
       <Autocomplete
-        onChange={onComboChange}
+        onChange={(e, inputObj) => setCurrentAccount(inputObj)}
```
decentralion, line 32: I'm surprised that the unused variable `e` isn't producing an eslint error.

befitsandpiper: even when using the onComboChange as before there is an unused variable. Not sure how to get around this, as those are the default arguments passed, and I only need to use the 2nd argument.

befitsandpiper: It looks like there's a way to do this with array destructuring... (...[ , inputObj]), but it's kind of ugly and I've never really seen that used in this context

befitsandpiper: Okay, this is what I'm settling on I think:
onChange={(...args) => {
        const inputObj = args[1];
        setCurrentAccount(inputObj)
      }}

befitsandpiper: this is the most elegant way to do the thing, and it works, but Flow throws an error about the rest parameter being unexpected:
onChange={(...[ , inputObj]) => setCurrentAccount(inputObj)}

befitsandpiper:
Thank you everyone for the reviews on this!

I just pushed a MUI theme update, I'm going to go ahead and merge it now, since everything is functionally working and the last few reviews were not critical issues. If anything else comes up it can be changed on the next pass when me and @KuraFire set the MUI color palette, other theme stuff, finer tune margins, etc

src/ui/components/AccountSelector.js
decentralion: Use:

```js
onChange={(_, inputObj) => {
     setCurrentAccount(inputObj);
}}
```

Our eslint rules explicitly allow _ as an unused variable:
https://github.com/sourcecred/sourcecred/blob/9eec13c799b8b00e24c83279955191b5a6cc7623/.eslintrc.js#L30-L36

Issue #1941, opened 2020-07-08
Add a `tribute` plugin


decentralion:
I'd like to add a "tribute" plugin, which enables projects to flow a %-age of their cred upstream to other projects they want to thank.

The config will be quite simple:
```
type TributeConfig = {|
  +tributeProportion: number,
  +tributees: $ReadOnlyArray<Tributee>,
|}
type Tributee = {|
  +username: Username,
  +weight: number,
|}
```

The rule is: every week, `tributeProportion` fraction of the total Cred minted in that week is given to a tribute node, and that tribute node flows Cred to the tributees, in proportion to each Tributee's weight.

This plugin has an awkward/unusual signature, since its `createGraph` method actually depends on the merged graph from the other plugins, but we can give it hardcoded support by invoking it in the `score` method (when the merged graph is created), and that's not too bad a hack.

It may require some experimentation to get the `tributeProportion` reasonably tuned; the issue is that if we do this naively, with unidirectional edges from the tribute node to the tributees, the tributees are likely to get greatly inflated cred scores (relative to tributeProportion) because of the self-loops; they're a lot "sticker" to Cred than regular users, and the cred renormalization logic will re-normalize a lot of the cred sore away from regular users. However, I think we can mitigate this by making the tribute edges strongly bidirectional, so that most of the tribute cred "sticks" to the tribute node.

I'm marking this as a beta feature, because I want to be able to include sourcecred as a default tributee for the projects we onboard during beta.

decentralion:
I consulted @wchargin for feedback on how to implement this, since there was a
lot of design uncertainty. Takeaways from the discussion:

- Let's store this information directly in the ledger.
- Rather than creating new events for naming/renaming/aliasing dependencies,
  let's refactor User->Identity and then make (User, Dependency, Organization,
  Bot) all subtypes of Identity
- Rather than putting the extra Cred generation into the Graph, we should
  directly create raw Cred (not node minting) after we have finished
  Graph-based Cred calculation

Note for Cred archeologists: I think this was a considerable improvement to the
design compared to my initial ideas :)

Paired with: @wchargin

PR #2095, opened 2020-08-08
add dependenciesMintPolicy

decentralion:
From the module docstring:

This module adds a core type for specifying "Dependency Mint" policies.
The Dependency Minting system is how we intend to allow projects to
flow Cred to their dependencies; e.g. it is how a user of SourceCred may
flow Cred to SourceCred itself. We want to make this easily configured at
the whole-project level, so that e.g. we can make projects flow ~5% of their
Cred to SourceCred by default.

It's not obvious how to embed this information directly into the Contribution
Graph in a way that's elegant and computationally efficient. E.g. we could add
an edge from every individual contribution to each dependency, but that
greatly increases the size and max degree of the Graph, which is
undesirable. After some discussion and consideration of alternatives,
@decentralion and @wchargin decided it would be cleanest to handle this
outside of the normal Graph abstraction. Thus, the Dependency Mint Policy was
born. These policies specify that after normal, Graph-based Cred computation is
finished, some nodes will recieve "extra" Cred minting which is proportional
to the total pre-dependencies Cred.

The total amount of Cred that may be minted is unbounded; for example, if
the dependencies have a total weight of 0.2, then the total Cred will be
120% of the base Cred, but if the dependencies had a total weight of 1, then
the total Cred would be double the base Cred. This was a deliberate design
decision so that dependency minting would feel "non-rival", i.e. there is
not a fixed budget of dependency cred that must be split between the
dependencies. In many cases, it may be very reasonable for the total Cred
flowing to a project's dependencies to be larger than the total Cred flowing
directly to the project's contributors; consider that the total amount of
time/effort invested in building all the dependencies may be orders of
magnitude larger than investment in the project itself.

In the future, we may develop more elegant approaches that integrate
dependencies more naturally into the contribution graph itself, in which
case we could phase out this system. But for now, it seems like a
sufficiently useful and flexible abstraction to merit inclusion in core/.

Note that there's nothing in the implementation or data types that imply
that this is only used for dependencies; one could imagine using it for
other purposes, e.g. minting extra Cred for moderators of a forum, or for
maintainers of a project. However, we have a prior that all contributors to
a project should be paid via CredRank and the graph-based system, rather
than "out-of-band" fiat/configuration like the mint policies in this module.
Thus, we're naming it the dependencyMintPolicy to communicate that
intention.

Test plan: Unit tests included, run `yarn test`.

Progress on #1941.

### DIFF
--- a/src/core/dependenciesMintPolicy.js
+++ b/src/core/dependenciesMintPolicy.js
@@ -1,0 +1,133 @@
+// @flow
+
+/**
+ * This module adds a core type for specifying "Dependency Mint" policies.
+ * The Dependency Minting system is how we intend to allow projects to
+ * flow Cred to their dependencies; e.g. it is how a user of SourceCred may
+ * flow Cred to SourceCred itself. We want to make this easily configured at
+ * the whole-project level, so that e.g. we can make projects flow ~5% of their
+ * Cred to SourceCred by default.
+ *
+ * It's not obvious how to embed this information directly into the Contribution
+ * Graph in a way that's elegant and computationally efficient. E.g. we could add
+ * an edge from every individual contribution to each dependency, but that
+ * greatly increases the size and max degree of the Graph, which is
+ * undesirable. After some discussion and consideration of alternatives,
+ * @decentralion and @wchargin decided it would be cleanest to handle this
+ * outside of the normal Graph abstraction. Thus, the Dependency Mint Policy was
+ * born. These policies specify that after normal, Graph-based Cred computation is
+ * finished, some nodes will recieve "extra" Cred minting which is proportional
+ * to the total pre-dependencies Cred.
+ *
+ * The total amount of Cred that may be minted is unbounded; for example, if
+ * the dependencies have a total weight of 0.2, then the total Cred will be
+ * 120% of the base Cred, but if the dependencies had a total weight of 1, then
+ * the total Cred would be double the base Cred. This was a deliberate design
+ * decision so that dependency minting would feel "non-rival", i.e. there is
+ * not a fixed budget of dependency cred that must be split between the
+ * dependencies. In many cases, it may be very reasonable for the total Cred
+ * flowing to a project's dependencies to be larger than the total Cred flowing
+ * directly to the project's contributors; consider that the total amount of
+ * time/effort invested in building all the dependencies may be orders of
+ * magnitude larger than investment in the project itself.
+ *
+ * In the future, we may develop more elegant approaches that integrate
+ * dependencies more naturally into the contribution graph itself, in which
+ * case we could phase out this system. But for now, it seems like a
+ * sufficiently useful and flexible abstraction to merit inclusion in core/.
+ *
+ * Note that there's nothing in the implementation or data types that imply
+ * that this is only used for dependencies; one could imagine using it for
+ * other purposes, e.g. minting extra Cred for moderators of a forum, or for
+ * maintainers of a project. However, we have a prior that all contributors to
+ * a project should be paid via CredRank and the graph-based system, rather
+ * than "out-of-band" fiat/configuration like the mint policies in this module.
+ * Thus, we're naming it the dependencyMintPolicy to communicate that
+ * intention.
+ */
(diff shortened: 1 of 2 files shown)


src/core/dependenciesMintPolicy.js
```diff
@@ -0,0 +1,133 @@
...
+  +address: NodeAddressT,
+  // Information on how the minted Cred varies in time.
+  +periods: $ReadOnlyArray<DependencyMintPeriod>,
+|};
+
+export type DependencyMintPeriod = {|
+  // What proportion of the project's raw Cred should be minted to this
+  // dependency during this period. For example, if the weight is 0.05, and the
+  // project has a pre-dependencies total Cred of 1000 in a given interval,
+  // then this dependench will receive an additional 50 Cred.
```
befitsandpiper, line 62: nit: "dependency" typo

src/core/dependenciesMintPolicy.js
```diff
@@ -0,0 +1,133 @@
...
+    if (currentStartTimeMs > startTimeMs) {
+      throw new Error(
+        `mint periods out of order: ${currentStartTimeMs} > ${startTimeMs}`
+      );
+    }
+    currentStartTimeMs = startTimeMs;
+    if (weight < 0 || !Number.isFinite(weight)) {
+      throw new Error(`invalid mint weight: ${weight}`);
+    }
+  }
```
hammadj, line 118: Wondering if this is something we can make a parser for instead of manually validating like this.

src/core/dependenciesMintPolicy.js
```diff
@@ -0,0 +1,133 @@
...
+ * Thus, we're naming it the dependencyMintPolicy to communicate that
+ * intention.
+ */
+import {type NodeAddressT, NodeAddress} from "./graph";
+import {type TimestampMs} from "../util/timestamp";
+
+export type DependencyMintPolicy = {|
+  // The node address that will receieve the extra minted Cred
+  +address: NodeAddressT,
+  // Information on how the minted Cred varies in time.
```
hammadj, line 54: Shouldn't it be "information on how the weight varies in time"? since its not actually storing the # of cred minted

src/core/dependenciesMintPolicy.js
```diff
@@ -0,0 +1,133 @@
...
+|};
+
+/**
+ * The ProcessedMintPolicy is a DependencyMintPolicy which has
+ * been transformed so that it matches the abstractions available when
+ * we're doing raw cred computation: instead of an address, we track an index
+ * into the canonical node order, and rather than arbitrary client-provided
+ * periods, we compute the weight for each Interval.
+ */
+export type ProcessedMintPolicy = {|
```
hammadj, line 77: Should be named "ProcessedDependencyMintPolicy" IMO

src/core/dependenciesMintPolicy.test.js
```diff
@@ -0,0 +1,93 @@
...
+        0.5,
+      ]);
+    });
+    it("handles a case with a multiple periods", () => {
+      const period1 = {startTimeMs: 10, weight: 0.5};
+      const period2 = {startTimeMs: 20, weight: 0.1};
+      expect(
+        _alignPeriodsToIntervals([period1, period2], [0, 5, 10, 15, 20, 25])
+      ).toEqual([0, 0, 0.5, 0.5, 0.1, 0.1]);
+    });
```
hammadj, line 36: Should add a case where the period startTimeMs is in between a time interval, and possibly another where there's multiple periods between one time interval and the next

src/core/dependenciesMintPolicy.js
decentralion: Done.

src/core/dependenciesMintPolicy.js
decentralion: Fixed.

src/core/dependenciesMintPolicy.js
decentralion: Changed.

src/core/dependenciesMintPolicy.test.js
decentralion: Done.

Issue #2084, opened 2020-08-06
Ledger Admin: Alias selector autocomplete should match middle of string


decentralion:
The ledger admin alias selector supports autocomplete, but only finds matches that start with the string.

For example, suppose that a user has the following aliases:

- `github/@user`
- `discourse/@user`
- `discord/user#1337`

It would be great if we could type "user" into the search box and find all matches. However, currently typing "user" would match none; you'd need to write `"github/@user"` and so forth. Also, the search is noticably slow.

We should fix this.

befitsandpiper:
changing to searching the whole string is a 1 line change. However, I'm not sure how to test the speed here on the dev server. Any suggestions?

decentralion:
Download a larger cred instance (e.g. for makerdao) and then do `yarn start --instance PATH_TO_BIG_INSTANCE`
You can find a makerdao instance here: https://github.com/decentralion/makerdao-instance-demo/

PR #2100, opened 2020-08-11
Make match alias robust and performant

befitsandpiper:
This branch is for issue: https://github.com/sourcecred/sourcecred/issues/2084

Switch to  more performant material-ui over combobox
Use includes rather than startsWith for alias search.
Test plan: test performance on maker instance.
Also test matching via UI

It's not beautiful, but it's honest work:
Screenshot from 2020-08-10 23-52-03

I am going to branch off of this to do a full material-ui overhaul, and fix page styling. So don't worry, it won't look like this for long.

### DIFF
--- a/src/ui/components/AliasSelector.js
+++ b/src/ui/components/AliasSelector.js
@@ -1,13 +1,15 @@
 // @flow
-
 import React, {useState, useEffect} from "react";
-import {useCombobox} from "downshift";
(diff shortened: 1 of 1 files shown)


befitsandpiper:
Does this look good enough for now?
Screenshot from 2020-08-12 22-49-57

PR #2104, opened 2020-08-12
Clean up discourse createGraph test code

decentralion:
This makes a few improvements to the createGraph test code:
- The MockData class has been renamed to MockRepository, which is a
better name because it matches the class it mocks (a ReadRepository),
and disambiguates from the "GrpahData" concept
- The MockRepository class now creates consistent users and trust levels from
its input data, rather than arbitrarily giving every user turst level 3
- We no longer define multiple MockRepository classes, there's only one.
- Some existing tests have been refactored to use the new repository
more cleanly.

Test plan: `yarn test`

### DIFF
--- a/src/plugins/discourse/createGraph.test.js
+++ b/src/plugins/discourse/createGraph.test.js
@@ -6,12 +6,10 @@ import type {ReadRepository} from "./mirrorRepository";
 import type {Topic, Post, PostId, TopicId, LikeAction, User} from "./fetch";
 import {EdgeAddress, type Node, type Edge, Graph} from "../../core/graph";
 import {
-  createGraph,
   _createReferenceEdges,
   weightForTrustLevel,
   _createGraphData,
   _graphFromData,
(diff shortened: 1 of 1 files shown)


src/plugins/discourse/createGraph.test.js
```diff
@@ -52,20 +50,26 @@ describe("plugins/discourse/createGraph", () => {
...
       }
       for (const {authorUsername} of this.topics()) {
-        if (!userNames.has(authorUsername)) {
-          userNames.add(authorUsername);
-          users.push({username: authorUsername, trustLevel: 3});
-        }
+        usernames.add(authorUsername);
+      }
+      for (const {username} of this.likes()) {
+        usernames.add(username);
```
decentralion, line 64: @befitsandpiper this is a change I was suggesting on Discord--we now "realistically" choose the user trust level based on the underlying post data, rather than inventing it arbitrarily

befitsandpiper: I see, yeah that makes sense. Sorry about that.

Issue #2109, opened 2020-08-12
Improve the Identity system


decentralion:
We have room for several major improvements to the identity system:

1. We don't automatically generate Identities for user-like node addresses. This puts a significant burden on instance maintainers to manually create identities for every user, rather than just merging them. Therefore, we should change the system so that at `sourcecred score` time, every user-like node is automatically upgraded to an identity and canoncialized in the ledger. (Note: This would allow us to entirely remove the concept of "userlike node types", in favor of having all contributors be represented by Identities, which can lead to significant UI and PluginDeclaration simplification.) 
2. We encode the identity subtypes in the identity address, but we also need to allow subtype re-writing. This means the addresses are mutable, which is unacceptable. We need to remove the subtype from the identity node address.
3. The `Identity` data type does not clearly distinguish between the user-modifiable data and the system-generated data. We should split out the user modifiable piece, which would allow (among other things) cleaner Ledger events to re-write user data, as suggested by @hammadj [here](https://github.com/sourcecred/sourcecred/pull/2080#pullrequestreview-463906325)
4. The identity only has a "name" property, but we want this to both be a display name and a url-sanitized login string. This blocks a lot of display names that the user would want, e.g. names with spaces, underscores, or capitalization. We should distinguish the login name from the display name.

I plan to make these three improvements. Since this is a change to the data structure, and we already have Ledgers in production, I'll need to bump the Ledger event version and provide backcompat support (a good test of our ability to make these changes ;) ). We'll also need to tweak the Ledger Admin so that instead of adding aliases, we merge identities.

decentralion:
Ok, having started down the path described above, I've hit some friction:

- The identity type is pretty deeply embedded in the ledger. Re-writing the ledger to use a new identity type, while maintaining full backcompat for existing ledgers, is possible but shaping up to be a huge pain. I'd really rather avoid it.
- Having separate identity names, distinct from the logins, feels like it will actually be a source of ambiguity and not a particularly valuable feature. Initially I thought we might want to support markdown in them, but I now think that's a bad idea.
- Have I mentioned that porting the Ledger to a new identity abstraction is a pain?

So, I think I want to take a different approach:
- identities will continue to have the address embedded in the identity. For identities we've already made, we'll leave in the subtype scoping, but we won't include it in future identities because subtypes are allowed to change.
- we'll generate identities automatically for user-like nodes in the graph, but we'll do so using the current identity data type

decentralion:
I'm considering this complete.

PR #2117, opened 2020-08-13
Ledger: add support for merging identities

decentralion:
This commit adds a new MergeIdentities action to the Ledger, which
allows merging two identities together. One identity is the "base" and
the other is the "target"; the target is absorbed. The base absorbs the
target's balance, paid amount, all of its aliases, and its own innate
address.

This is part of #2109.

Test plan: Thorough unit tests added. Snapshots updated.


befitsandpiper:
I wonder if we should also create a function here that allows separating an identity back out into multiple identities. There are some open questions here around withdrawals though. 

Since addresses remain distinct post merging it is possible to recalculate cred for each address in a separation event, but it's unclear how remaining balances would be recalculated. This shouldn't really come up if there are never errors when merging. However, I'm a little wary of irreversible actions like this.

decentralion:
@befitsandpiper we'll definitely need the ability to split identities, e.g. when the identities have been merged by a scammer pulling off identity theft. as you noted, it's nontrivial to work out how to separate balances, and how to retroactively attribute the "paid" amount to the different identities, post-split.

i'm thinking we might put in the unmerge operation pre-beta-launch, or we might put it in after we've seen a case or two where identities need to be split and then use that experience to inform the unmerge semantics.

befitsandpiper:
Given a crisis situation, we can manually pick apart a faulty identity merge situation, right?

decentralion:
> Given a crisis situation, we can manually pick apart a faulty identity merge situation, right?

In a crisis situation, the offending identity can be de-activated, which will prevent it from getting any ill-gotten Grain until the situation gets sorted out. Adding a "demerge identities" operation would likely only take a few hours.

PR #2128, opened 2020-08-15
Add a system for identity proposals

decentralion:
This commit adds the `identityProposal` module, which specifies how a
plugin may propose a new identity to be added to the Ledger. The
proposal includes a name, the name of the plugin (for
disambiguation), the type of identity that should be created, and
the initial alias the identity should own.

If some existing identity has an alias with that address already, then
no action will be taken. However, if there is no identity matching that
alias, then a new identity will be created, based on the name (and, if
necessary, pluginName) of the proposal.

This will allow the Ledger to be auto-populated with new identities as
provided by plugins, which will save instance adminstrators from the
very tedious task of making an identity for each user in the instance.
This will also have other downstream benefits, like letting us do away
with the plugin node types in favor of assuming that all user accounts
correspond to identities (and, thus, identity addresses).

This is a big step towards #2109.

Test plan: Unit tests included, as per usual.

Note: I considered putting the identity proposal logic into the ledger
directly, but I think it's cleaner to write it as a consumer of the
ledger APIs. Primarily because by forcing it to go through API
boundaries, it's easier to reason about the implementation, and doesn't
add any bloat to the Ledger module, which is already quite complex.

### DIFF
--- a/src/ledger/identity/index.js
+++ b/src/ledger/identity/index.js
@@ -10,6 +10,7 @@ export {
 } from "./identity";
 
 export type {IdentityType} from "./identityType";
+export {parser as identityTypeParser} from "./identityType";
 
 export type {Name} from "./name";
 export {nameFromString, parser as nameParser} from "./name";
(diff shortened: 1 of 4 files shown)


decentralion:
Merging on self-review.

PR #2137, opened 2020-08-16
Support merging identities in the ledger admin

decentralion:
As of #2134, we create identities for every user. Thus, we no longer
want to add aliases to identities, but instead we merge them together
(this functionality from #2117). This commit updates the frontend so we
can merge identities.

Thanks to @hammadj for major help letting this land in a bug free
fashion. (Via #2140.)

Test plan: Run `yarn start`, and merge some identities together.


src/ui/components/IdentityMerger.js
```diff
@@ -0,0 +1,80 @@
...
+          }
+        }}
+        onChange={(_, selectedItem, reason) => {
+          if (reason === "select-option") {
+            setLedger(
+              ledger.mergeIdentities({
+                base: selectedIdentityId,
+                target: selectedItem.id,
+              })
+            );
```
decentralion, line 63: @hammadj I'm not sure why, but this `setLedger` isn't causing the parent component to re-render with the new ledger, even though in the parent component whenever we call `setLedger` it works as expected.

src/ui/components/IdentityMerger.js
```diff
@@ -0,0 +1,75 @@
...
+      <Autocomplete
+        onInputChange={(_, value, reason) => {
+          if (reason === "input") {
+            setSearch(value);
+            setInputValue(value);
+          }
+        }}
+        onChange={(_, selectedItem, reason) => {
+          if (reason === "select-option") {
+            console.log("setLedger in IdentityMerger");
```
hammadj, line 52: stray console.log

PR #2138, opened 2020-08-16
frontend: wrap ledger state to avoid state bugs

decentralion:
Currently, the React component tree tracks a Ledger directly as state.
This leads to a class of confusing bugs due to the fact that ledgers
themselves track state, and a ledger update results in a referentially
equivalent object in the React state.

This commit wraps the Ledger in an object which updates every time the
ledger is changed, so that we'll no longer have referentially equal
state before and after a state update, so we will get React re-renderes
appropriately.

Test plan: Integrate with a downstream commit and confirm the state bugs
go away.

### DIFF
--- a/src/ui/components/AdminApp.js
+++ b/src/ui/components/AdminApp.js
@@ -92,13 +92,32 @@ const AdminApp = () => {
 };
 
 /**
+ * The ledger is a piece of mutable state.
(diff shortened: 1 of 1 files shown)


decentralion:
Superseded by #2140.

PR #2144, opened 2020-08-17
Setup Github Action to auto-publish to NPM

hammadj:
This sets up a Github Action that triggers every time a new release is created in GH. It runs tests, builds, updates the
package.json version to the one defined by the tag in the Release, then publishes to NPM.

This makes the process of creating a release super simple and only one step: Publish a release with the desired version
in Github.

Note: For this action to work, an [NPM_TOKEN](https://docs.npmjs.com/creating-and-viewing-authentication-tokens) will need to be added to the repo secrets. 

**Test Plan:** Create a new Release in the Github UI with a tag that is one version higher than the current version in package.json. Ensure this workflow is triggered and that the new version is released to NPM and the package.json version is updated as well on master.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,6 +24,8 @@ jobs:
 
       - name: Test
         run: yarn test --full --ci
+        env:
+          SOURCECRED_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       # Uses the version defined by the tag created in the Release
       # and pushes a commit that updates the version in package.json
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,38 @@
+name: Publish Release
+on:
+  # Publish when a new release is created on GitHub
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: 'https://registry.npmjs.org'
+
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: nodeModules-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install packages
+        run: yarn --frozen-lockfile
+
+      - name: Test
+        run: yarn test --ci
+
+      # Uses the version defined by the tag created in the Release
+      # and pushes a commit that updates the version in package.json
+      - name: Publish to NPM
+        run: |
+          git config --global user.name $GITHUB_ACTOR
+          git config --global user.email ${GITHUB_ACTOR}@users.noreply.github.com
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: nodeModules-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install packages
+        run: yarn --frozen-lockfile
+
+      - name: Test
+        run: yarn test --ci
```
decentralion, line 26: Can we make it test-full?

package.json
```diff
@@ -119,7 +119,8 @@
     "sharness-full": "make -sC ./sharness prove PROVE_OPTS=-vf TEST_OPTS='-v --chain-lint --long'",
     "coverage": "yarn run unit --coverage",
     "flow": "flow",
-    "lint": "eslint src config --max-warnings 0"
+    "lint": "eslint src config --max-warnings 0",
+    "prepublishOnly": "yarn build"
```
decentralion, line 123: Is this a yarn feature? That it runs this before publish?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      # Uses the version defined by the tag created in the Release
+      # and pushes a commit that updates the version in package.json
+      - name: Publish to NPM
+        run: |
+          git config --global user.name $GITHUB_ACTOR
+          git config --global user.email ${GITHUB_ACTOR}@users.noreply.github.com
+          yarn publish --new-version ${GITHUB_REF#"refs/tags/"} --no-git-tag-version
+          git push
```
decentralion, line 35: Is it pushing to master? Will we need to whitelist the github actor so it has permissions to do this?

package.json
hammadj: Yep!

hammadj:
> How can we test this? Maybe run it on a fork?

@decentralion Here's a passing run on a fork: https://github.com/hammadj/sourcecred/runs/1043059049

It was triggered when I created this release from Github's UI: https://github.com/hammadj/sourcecred/releases/tag/0.7.0-publish-6

Published version on NPM: https://www.npmjs.com/package/sourcecred-publish-test

However, with the branch protection rule, this action fails to push updates directly into master. Three things we can do here:
- Instead of using the built-in GITHUB_TOKEN, you can generate a personal access token and add that to the secrets as GITHUB_TOKEN_ADMIN. Using that inside the push action will succeed IF the "Include administrators" option in the branch protection rule is unchecked.
- If we want to keep the "Include administrators", then we will have to add some steps in the action to temporarily enable/disable that option when its pushing to master. Will still need the GITHUB_TOKEN_ADMIN in secrets.
- The action can just make a PR instead of merging to master.

Context: https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101

decentralion (approved):

PR #2145, opened 2020-08-17
Implement custom currencies in UI

### LINKED ISSUE #2112: Allow customizing what "Grain" is called
Not all projects call their currency Grain -- other examples in production are SEEDS and DAI. We should allow projects to customize this name.

I think we should, in `config/grain.json`, allow specifying a name (which defaults to "Grain") and a suffix (which defaults to "g"). Then the frontend should use these names appropriately.

We should also likely change the url structure to be name agnostic -- e.g. change the "Grain Accounts" url to use "accounts" as the slug instead of "grain".

topocount:
Custom currencies can now be configured in an instance's
config/grain.json file. The fields `currencyName` and currencySuffix`
will be parsed by the frontend to display these values in place of the
default "Grain" and "g", respectively.
    
test plan: run `yarn start` and ensure that the "Grain" name and "g" are
not visible in the Menu bar, in the transfer arrow when selecting an
amount, or in the Accounts view. In each of these locations, they should
be replace with Dai. Other currencies can be tested by modifying the
sharness example instance's grain/config file

resolves #2112

### DIFF
--- a/src/api/currencyConfig.js
+++ b/src/api/currencyConfig.js
@@ -2,6 +2,7 @@
 
 import {DEFAULT_SUFFIX} from "../ledger/grain";
 import * as C from "../util/combo";
+import * as NullUtil from "../util/null";
 
 /**
  * Shape of concurrencyDetails.json on disk
@@ -29,8 +30,8 @@ export const DEFAULT_NAME = "Grain";
  */
 function upgrade(c: SerializedCurrencyDetails): CurrencyDetails {
   return {
-    name: c.currencyName || DEFAULT_NAME,
-    suffix: c.currencySuffix || DEFAULT_SUFFIX,
+    name: NullUtil.orElse(c.currencyName, DEFAULT_NAME),
+    suffix: NullUtil.orElse(c.currencySuffix, DEFAULT_SUFFIX),
   };
 }
 
--- a/src/api/currencyConfig.js
+++ b/src/api/currencyConfig.js
@@ -1,0 +1,40 @@
+// @flow
+
+import {DEFAULT_SUFFIX} from "../ledger/grain";
+import * as C from "../util/combo";
+
+/**
+ * Shape of concurrencyDetails.json on disk
+ */
+type SerializedCurrencyDetails = {|
+  +currencyName?: string,
+  +currencySuffix?: string,
+|};
+
+/**
+ * Shape of currencyDetails json in memory after parsing
+ */
+export type CurrencyDetails = {|
+  +name: string,
+  +suffix: string,
+|};
+
+export const DEFAULT_NAME = "Grain";
+
+/**
+ * Utilized by combo.fmap to enforce default currency values
+ * when parsing. This engenders a "canonical default" since there
+ * will be no need to add default fallbacks when handling currency
+ * detail values after parsing the serialized file.
+ */
+function upgrade(c: SerializedCurrencyDetails): CurrencyDetails {
+  return {
+    name: c.currencyName || DEFAULT_NAME,
+    suffix: c.currencySuffix || DEFAULT_SUFFIX,
+  };
+}
+
+export const parser: C.Parser<CurrencyDetails> = C.fmap(
+  C.object({}, {currencyName: C.string, currencySuffix: C.string}),
+  upgrade
+);
--- a/src/ui/components/AccountOverview.js
+++ b/src/ui/components/AccountOverview.js
@@ -9,7 +9,7 @@ import TableHead from "@material-ui/core/TableHead";
 import TableRow from "@material-ui/core/TableRow";
 import Paper from "@material-ui/core/Paper";
 import {type Account} from "../../ledger/ledger";
-import {type CurrencyDetails} from "../load";
+import {type CurrencyDetails} from "../../api/currencyConfig";
 import * as G from "../../ledger/grain";
 import {useLedger} from "../utils/LedgerContext";
 
--- a/src/ui/components/AccountOverview.js
+++ b/src/ui/components/AccountOverview.js
@@ -13,7 +13,7 @@ import {type CurrencyDetails} from "../../api/currencyConfig";
 import * as G from "../../ledger/grain";
 import {useLedger} from "../utils/LedgerContext";
 
-type OverviewProps = {|currency: CurrencyDetails|};
+type OverviewProps = {|+currency: CurrencyDetails|};
 
 export const AccountOverview = ({
   currency: {suffix: currencySuffix},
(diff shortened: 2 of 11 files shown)


src/ui/load.js
```diff
@@ -37,7 +42,18 @@ export async function load(): Promise<LoadResult> {
     const rawLedger = await responses[2].text();
     const ledger = Ledger.parse(rawLedger);
     const {hasBackend} = await responses[3].json();
-    return {type: "SUCCESS", credView, bundledPlugins, ledger, hasBackend};
+    const {
+      currencySuffix: suffix = DEFAULT_SUFFIX,
+      currencyName: name = "Grain",
+    } = grainConfigParser.parseOrThrow(await responses[4].json());
```
decentralion, line 48: I'd prefer to have the defaults live at the level of the config file, e.g. we could have a type `SerializedGrainConfig` which has the fields as optional, and then a `GrainConfig` has the fields as required, and we'd have some sort of `upgrade(g: SerializedGrainConfig): GrainConfig` function which applies the defaults. Then the parser would compose a `SerilaizedGrainConfig` parse step along with using `fmap` to apply the upgrade. That way, we can ensure that the same defaults are used consistently across the system, v.s. a case where we might call the default `"Grain"` here and `"grain"` elsewhere and have slight inconsistencies. 

Also, now that I'm thinking about it more, we may want to have a config file which is only focused on the currency display stuff, e.g. call it `config/currencyDetails.json`. Looking at `config/grain.json`, to have it both give currency details and payout strategy feels like a messy mixing of concerns... the payout strategy and the rendering info are actually quite distinct. What do you think? Would merge it either way, but I think it may be a better design to separate them.

src/ui/components/AccountOverview.js
```diff
@@ -36,19 +41,21 @@ export const GrainAccountOverview = () => {
...
+const AccountRow = (account: Account, suffix: string) => (
   <TableRow key={account.identity.id}>
     <TableCell component="th" scope="row">
       {account.identity.name}
     </TableCell>
     <TableCell align="right">{account.active ? "✅" : "🛑"}</TableCell>
-    <TableCell align="right">{G.format(account.balance)}</TableCell>
-    <TableCell align="right">{G.format(account.paid)}</TableCell>
+    <TableCell align="right">{G.format(account.balance, 0, suffix)}</TableCell>
+    <TableCell align="right">{G.format(account.paid, 0, suffix)}</TableCell>
```
decentralion, line 59: While we're here--mind shifting to 2 decimals precision?

src/ui/load.js
topocount: My instinct was to create a separate `currencyDetails.json` file (I need to start putting these thoughts in the PR descriptions).  Combining data used for completely different applications indeed seems inconsistent with the design philosophy of the project, so I agree we should fix this here. I'm excited to dig deeper into Combo functionality as well!

topocount:
@decentralion re-requesting a review on this to ensure I've implemented this new type and parser as you were expecting

src/api/currencyConfig.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Utilized by combo.fmap to enforce default currency values
+ * when parsing. This engenders a "canonical default" since there
+ * will be no need to add default fallbacks when handling currency
+ * detail values after parsing the serialized file.
+ */
+function upgrade(c: SerializedCurrencyDetails): CurrencyDetails {
+  return {
+    name: c.currencyName || DEFAULT_NAME,
+    suffix: c.currencySuffix || DEFAULT_SUFFIX,
```
decentralion, line 33: nit: I generally prefer `c.currencyName == null ? DEFAULT_NAME : c.currencyName` or (alternatively) `NullUtil.orElse(c.currencyName, DEFAULT_NAME)`. 

That way we treat all input strings consistently, rather than having an implicit special case where if the user specifies the empty string we refuse to respect the user's choice and treat it as unset instead. It comes from the general principle in our codebase of trying to treat the empty element of a type consistently wherever possible.

(change not required).

src/ui/components/AccountOverview.js
```diff
@@ -9,10 +9,15 @@ import TableHead from "@material-ui/core/TableHead";
 import TableRow from "@material-ui/core/TableRow";
 import Paper from "@material-ui/core/Paper";
 import {type Account} from "../../ledger/ledger";
+import {type CurrencyDetails} from "../../api/currencyConfig";
 import * as G from "../../ledger/grain";
 import {useLedger} from "../utils/LedgerContext";
 
-export const GrainAccountOverview = () => {
+type OverviewProps = {|currency: CurrencyDetails|};
```
decentralion, line 16:
```suggestion
type OverviewProps = {|+currency: CurrencyDetails|};
```
Should be readonly, right?

src/api/currencyConfig.js
```diff
@@ -0,0 +1,40 @@
...
+  };
+}
+
+export const parser: C.Parser<CurrencyDetails> = C.fmap(
+  C.object({}, {currencyName: C.string, currencySuffix: C.string}),
+  upgrade
+);
```
decentralion, line 40: Yup, this is what I had in mind!

src/ui/components/Menu.js
```diff
@@ -5,23 +5,26 @@ import {MenuItemLink} from "react-admin";
...
       <MenuItemLink
         to="/explorer"
-        primaryText="explorer"
+        primaryText="Explorer"
         leftIcon={<TrendingUpIcon />}
         onClick={onMenuClick}
         sidebarIsOpen={open}
       />
       <MenuItemLink
         to="/grain"
```
decentralion, line 26:
```suggestion
        to="/accounts"
```

What do you think?

src/ui/components/Transfer.js
```diff
@@ -44,7 +45,11 @@ const useStyles = makeStyles((theme) => ({
   pageHeader: {color: theme.palette.text.primary},
 }));
 
-export const TransferGrain = () => {
+type TransferProps = {|currency: CurrencyDetails|};
```
decentralion, line 48:
```suggestion
type TransferProps = {|+currency: CurrencyDetails|};
```

src/ui/load.js
```diff
@@ -11,19 +15,25 @@ export type LoadSuccess = {|
...
   +hasBackend: Boolean,
+  +currency: CurrencyDetails,
 |};
 export type LoadFailure = {|+type: "FAILURE", +error: any|};
 
 export async function load(): Promise<LoadResult> {
+  // TODO (@topocount) refactor to better
+  // utilize functional programming best practices.
+  // Optional loads require some better organization
+  // than ternaries. There's also a lot of repeated code here
```
decentralion, line 26: Agreed :) Would be a good fit for a small refactor commit in its own PR.

PR #2147, opened 2020-08-18
Add dependency minting to cred compute pipeline

decentralion:
This commit injects the concept of dependency minted Cred (from #2095)
to the TimelineCred computation pipeline, in `credData.js`. The file is
lightly tested and I added another test. I also added some comments
along the way, since I needed the clarification.

Note: This code path is deprecated, and will get replaced by
TimelineCred in the coming months. As such, I'm ok with somewhat lower
test comprehensiveness.

Test plan: See attached unit test. Logic is also pretty simple.

Part of #1941.


src/analysis/credData.js
```diff
@@ -99,13 +111,28 @@ export function computeCredData(scores: TimelineCredScores): CredData {
...
       nodeOverTime[n].syntheticLoopFlow[i] = syntheticLoopFlow[n];
+      // Pre-fill with 0 to ensure a value for every node
+      nodeOverTime[n].dependencyMintedCred[i] = 0;
+    }
+    for (const {nodeIndex, intervalWeights} of mintPolicies) {
+      const weight = intervalWeights[i];
+      const mintedCred = weight * intervalTotalCred;
+      // The following is needed to avoid flow errors.
+      // I have no idea why. It's already typed as number.
+      const idx = (nodeIndex: number);
```
decentralion, line 131: @wchargin any guesses?

PR #2150, opened 2020-08-20
Update Explorer Table to material-ui

youngkidwarrior:
Just a basic PR to get the material-ui kicked off.

Using material-ui buttons to start.

### DIFF
--- a/src/ui/components/Explorer.js
+++ b/src/ui/components/Explorer.js
@@ -1,7 +1,10 @@
 // @flow
 
 import React, {type Node as ReactNode} from "react";
-import {Button} from "@material-ui/core";
+import {Button, IconButton} from "@material-ui/core";
+import AddIcon from "@material-ui/icons/Add";
+import RemoveIcon from "@material-ui/icons/Remove";
+import {styled} from "@material-ui/core/styles";
 import deepEqual from "lodash.isequal";
 import {StyleSheet, css} from "aphrodite/no-important";
 import Markdown from "react-markdown";
@@ -142,6 +145,7 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
     const analyzeButton = (
       <Button
         variant="contained"
+        color="primary"
         disabled={this.state.recalculating || paramsUpToDate}
         onClick={() => this.analyzeCred()}
       >
@@ -156,6 +160,7 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
           <span style={{flexGrow: 1}} />
           <Button
             variant="contained"
+            color="primary"
             onClick={() => {
               this.setState(({showWeightConfig}) => ({
                 showWeightConfig: !showWeightConfig,
@@ -343,8 +348,10 @@ class TableRow extends React.Component<TableRowProps, TableRowState> {
       <React.Fragment>
         <tr style={{backgroundImage}} className={css(styles.hoverHighlight)}>
           <td>
-            <Button
-              variant="contained"
+            <IconButton
+              aria-label="expand"
+              color="primary"
+              size="medium"
               style={{
                 marginRight: 5,
                 marginLeft: 15 * indent + 5,
@@ -355,8 +362,8 @@ class TableRow extends React.Component<TableRowProps, TableRowState> {
                 }));
               }}
             >
-              {expanded ? "\u2212" : "+"}
-            </Button>
+              {expanded ? <RemoveIcon /> : <AddIcon />}
+            </IconButton>
             <Markdown renderers={{paragraph: "span"}} source={description} />
           </td>
           <td style={{textAlign: "right"}}>{format(".1d")(cred)}</td>
--- a/src/ui/components/Explorer.js
+++ b/src/ui/components/Explorer.js
@@ -131,7 +131,11 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
               })
             }
           >
-            <ListItemText primary={"Filter"} />
+            <ListItemText
+              primary={
+                this.state.name ? `Filter: ${this.state.name}` : "Filter"
+              }
+            />
             {this.state.anchorEl ? (
               <KeyboardArrowUpIcon />
             ) : (
@@ -148,8 +152,8 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
           open={Boolean(this.state.anchorEl)}
           onClose={this.handleMenuClose}
           getContentAnchorEl={null}
-          anchorOrigin={{vertical: "bottom", horizontal: "center"}}
-          transformOrigin={{vertical: "top", horizontal: "center"}}
+          anchorOrigin={{vertical: "bottom", horizontal: "left"}}
+          transformOrigin={{vertical: "top", horizontal: "left"}}
         >
           <MenuItem
             key={"All users"}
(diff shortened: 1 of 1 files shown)


topocount:
screenshot of the changes: 
image

youngkidwarrior:
yeah sure thing. I debated making them look all fancy. Give me a few

decentralion:
@youngkidwarrior no need to worry about the test_full failures, it's just because you PR'd from a fork so you don't have all the needed credentials. I sent you an invite to join the sourcecred org, once you accept that you can PR from branches inside sourcecred/sourcecred rather than from your fork.

youngkidwarrior:
new spanky buttons 
image

decentralion:
I'm concerned that they look a lot less like buttons now, and so it's a lot less intuitive to a user that they are meant to be clicked on. cc @KuraFire

youngkidwarrior:
I think material-ui when fully Integrated will add some lines in-between or possibly a card like element with a box shadow. Might make things easier. 

e.g.



I could also just make the buttons filled :)

KuraFire:
@youngkidwarrior Ooh, I like it, but I see @decentralion's concern. Could you add a preview of what those buttons would look like in the filled button style?

youngkidwarrior:
image

KuraFire:
@youngkidwarrior I definitely like the non-filled cleaner look better, but as discussed on Discord let's see it with the expand/collapse arrow icons instead of +/-, which are a little too old-school and not super intuitive here.

```
➡️  Participant (collapsed)
➡️  Participant (collapsed)
⬇️  Expanded Participant 
```

But with pretty Material UI icons :)

src/ui/components/Explorer.js
```diff
@@ -36,6 +55,9 @@ export type ExplorerState = {|
   showWeightConfig: boolean,
   view: CredView,
   recalculating: boolean,
+  anchorEl: any,
+  selectedIndex: number,
```
decentralion, line 59: What are these for?

decentralion:
Also, could you update the PR description?

src/ui/components/Explorer.js
youngkidwarrior: those are to handle the dropdown menu selection. I can rename and properly type anchorEl

src/ui/components/Explorer.js
```diff
@@ -36,6 +55,9 @@ export type ExplorerState = {|
   showWeightConfig: boolean,
   view: CredView,
   recalculating: boolean,
+  anchorEl: any,
```
topocount, line 58: Let's use HTMLElement here
```suggestion
  anchorEl: HTMLElement | null,
```

src/ui/components/Explorer.js
topocount, line 59: Based on my testing, `selectedIndex` can be removed, since we aren't highlighting the selection in the menu itself.

src/ui/components/Explorer.js
```diff
@@ -49,44 +71,108 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
...
+            ) : (
+              <KeyboardArrowDownIcon />
+            )}
+          </ListItem>
+          <Divider style={{backgroundColor: "#F20057", height: "2px"}} />
+        </List>
+
+        <Menu
+          id="lock-menu"
+          anchorEl={this.state.anchorEl}
```
topocount, line 153: Let's move the dropdown to below the the actual toggle:
```suggestion
          anchorEl={this.state.anchorEl}
          getContentAnchorEl={null}
          anchorOrigin={{vertical: "bottom"}}
          transformOrigin={{vertical: "top"}}
```

src/ui/components/Explorer.js
```diff
@@ -49,44 +71,108 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
...
+        <List component="div" aria-label="Device settings">
+          <ListItem
+            button
+            aria-haspopup="true"
+            aria-controls="filter-menu"
+            aria-label="filters"
+            onClick={(event) =>
+              this.setState({
+                anchorEl: event.currentTarget,
+              })
```
topocount, line 136: Let's change the name back to "Filter" here
```suggestion
              this.setState({
                name: "Filter",
                anchorEl: event.currentTarget,
              })
```

topocount (commented):
Hey @youngkidwarrior This is looking quite good. I have a couple requests around making the menu more "dropdown-like" and some cleanup involving errors in the console, which may or may not be caused by your changes, but either way, this is a good opportunity to fix the "duplicate key" errors.

Definitely let me know if you have any followup questions!

src/ui/components/Explorer.js
youngkidwarrior: We could have `Filter: ${currentTarget}` to retain information

youngkidwarrior: Actually came up with something better @topocount. Check it

src/ui/components/Explorer.js
```diff
@@ -160,12 +147,21 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
           keepMounted
           open={Boolean(this.state.anchorEl)}
           onClose={this.handleMenuClose}
+          getContentAnchorEl={null}
+          anchorOrigin={{vertical: "bottom", horizontal: "center"}}
+          transformOrigin={{vertical: "top", horizontal: "center"}}
```
topocount, line 152:
```suggestion
          anchorOrigin={{vertical: "bottom", horizontal: "left"}}
          transformOrigin={{vertical: "top", horizontal: "left"}}
```

src/ui/components/Explorer.js
topocount: How about we do "both, and"? I like your idea of having the filter name in the description (`Filter: ${currentTarget`), and I also like the idea of putting the filter in the table header.

src/ui/components/Explorer.js
```diff
@@ -316,7 +312,7 @@ export class Explorer extends React.Component<ExplorerProps, ExplorerState> {
           <TableHead>
             <TableRow>
               <TableCell align="left" style={{color: "black"}}>
-                Node
+                {name ? name : "All users"}
```
topocount, line 315: Nit: Would it be possible to prefix the filter group (like the plugin name here, so if you select "Pull requests" from the dropdown you'll see "Github: Pull Requests"

topocount (approved):
Almost there! The alignment of the dropdown should be "left" and the other two filter wording comments are nits. I'll approve this and let you decide on the nits.

src/ui/components/Explorer.js
youngkidwarrior: This is possible but probably a bigger pain then what its worth. (Adding a bunch of state logic and splitting strings)

decentralion (commented):
So I'm noticing a few issues with this PR:

1. When recalculating Cred after changing the weights, there is no visual indicator that reveals when the computation is done. On a big instance (e.g. SourceCred's own Cred) it can take upward of a minute, so we need some way for the user to know when the recompute is done and they can start exploring the updated Cred scores. In the current version this is handled really crudely https://github.com/sourcecred/sourcecred/blob/ab8a13d8355b65993a2bfa59fa207c6021d04bc3/src/ui/components/Explorer.js#L208 and it seems to have vanished in this PR; maybe we could have a button show up next to "Computing Cred" that shows that recalculation is in progress?

2. We have a console error appearing when expanding one of the rows:
image

I know that @befitsandpiper is blocked on this PR merging, so I'm OK with merging this as-is, so long as either @youngkidwarrior or @befitsandpiper will take on fixing these regressions before the next release.

PR #2161, opened 2020-08-24
add cache-clearing step in load cli command

topocount:
This is a relatively naive implementation of a cache clearing retry for
the load command. It works as follows:
1. check all plugins to be loaded for an existing cache db and save the
state to a Map
2. For each plugin if the load fails for any reason and a cache exists,
  the cache is deleted and the load is retried. Otherwise the load fails
  as before.
3. If the single retry fails, the load will fail.

Some alternative implementations were considered but decided against for
the following reasons:
- the CLI context has no knowledge of the db file name composition
semantics, so naively checking the cache dir was implemented in favor of
looking for a specific file
- It's difficult to catalog all reasons the cache may be
invalidated across plugins, so this implementation does not parse the
error type in order to determine whether a retry is the proper next step

test plan:
1. run `path/to/sourcred.js load` to set up an initital cache
2. perform an action that would cause a sqlite db error or a
graphql error using the master cli load command
3. run `path/to/sourcred.js load` again to verify the error does not
occur. Instead, the cache db should be deleted and recreated

### DIFF
--- a/src/cli/load.js
+++ b/src/cli/load.js
@@ -4,39 +4,30 @@ import * as NullUtil from "../util/null";
 import type {Command} from "./command";
 import {loadInstanceConfig, pluginDirectoryContext} from "./common";
 import {LoggingTaskReporter} from "../util/taskReporter";
-import * as pluginId from "../api/pluginId";
-import {Plugin, PluginDirectoryContext} from "../api/plugin";
+import {type PluginId, parser as pluginIdParser} from "../api/pluginId";
 import {isDirEmpty} from "../util/disk";
 import fs from "fs-extra";
+import chalk from "chalk";
 
 function die(std, message) {
   std.err("fatal: " + message);
   return 1;
 }
 
-type PluginLoadContext = {|
-  +plugin: Plugin,
-  +dirContext: PluginDirectoryContext,
-  +taskReporter: LoggingTaskReporter,
-  +task: string,
-|};
-
-function loadPlugin(ctx: PluginLoadContext) {
-  const {plugin, dirContext, taskReporter, task} = ctx;
-  return plugin
-    .load(dirContext, taskReporter)
-    .then(() => taskReporter.finish(task));
+function warn(std, task: string, message: string) {
+  const label = chalk.bgYellow.bold.white(" WARN ");
+  std.out(`${label} ${task}: ${message}`);
 }
 
 const loadCommand: Command = async (args, std) => {
-  let pluginsToLoad = [];
+  let pluginsToLoad: PluginId[] = [];
   const baseDir = process.cwd();
   const config = await loadInstanceConfig(baseDir);
   if (args.length === 0) {
-    pluginsToLoad = config.bundledPlugins.keys();
+    pluginsToLoad = Array.from(config.bundledPlugins.keys());
   } else {
     for (const arg of args) {
-      const id = pluginId.fromString(arg);
+      const id = pluginIdParser.parseOrThrow(arg);
       if (config.bundledPlugins.has(id)) {
         pluginsToLoad.push(id);
       } else {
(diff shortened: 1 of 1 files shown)


src/cli/load.js
```diff
@@ -34,14 +51,28 @@ const loadCommand: Command = async (args, std) => {
...
+      taskReporter,
+      task,
+    };
+    const promise = loadPlugin(pluginCtx)
+      .catch((e) => {
+        if (!cacheEmpty.get(name)) {
+          // clear the cache and try again
+          fs.emptyDirSync(dirContext.cacheDirectory());
+          return loadPlugin(pluginCtx);
+        }
```
decentralion, line 73: I think it'd be appropriate to log the error here, and report that we're re-trying

src/cli/load.js
```diff
@@ -34,14 +51,28 @@ const loadCommand: Command = async (args, std) => {
...
-      .load(dirContext, taskReporter)
-      .then(() => taskReporter.finish(task))
+    cacheEmpty.set(name, isDirEmpty(dirContext.cacheDirectory()));
+    const pluginCtx: PluginLoadContext = {
+      plugin,
+      dirContext,
+      taskReporter,
+      task,
+    };
+    const promise = loadPlugin(pluginCtx)
```
decentralion, line 67: might be nice to rename `promise` to `loadWithPossibleRetry` or such

src/cli/load.js
wchargin: +1; if this happens habitually for non-transient reasons, we’d want to
know about it rather than just silently retrying, which would waste
time/compute and hide bugs.

src/cli/load.js
```diff
@@ -5,12 +5,29 @@ import type {Command} from "./command";
...
+  +taskReporter: LoggingTaskReporter,
+  +task: string,
+|};
+
+function loadPlugin(ctx: PluginLoadContext) {
+  const {plugin, dirContext, taskReporter, task} = ctx;
+  return plugin
+    .load(dirContext, taskReporter)
+    .then(() => taskReporter.finish(task));
+}
```
wchargin, line 29: This can be more concisely (and imho more clearly) expressed as a
closure, capturing the environment rather than reifying it:

```javascript
    const loadPlugin = () =>
      plugin
        .load(dirContext, taskReporter)
        .then(() => taskReporter.finish(task));
    const promise = loadPlugin()
      .catch((e) => {
        if (!cacheEmpty.get(name)) {
          // clear the cache and try again
          fs.emptyDirSync(dirContext.cacheDirectory());
          return loadPlugin();
        }
        throw e;
      })
      .catch((e) => {
        console.error(e);
        failedPlugins.push(name);
      });
```

Net diff is −16 lines locally, plus you drop an import.

src/cli/load.js
```diff
@@ -5,12 +5,29 @@ import type {Command} from "./command";
 import {loadInstanceConfig, pluginDirectoryContext} from "./common";
 import {LoggingTaskReporter} from "../util/taskReporter";
 import * as pluginId from "../api/pluginId";
+import {Plugin, PluginDirectoryContext} from "../api/plugin";
```
wchargin, line 8: By the way: when you only use imported symbols in type positions rather
than runtime positions—i.e., when the imports are unused after removing
Flow type annotations—please use `import type` rather than `import`.
This prevents adding an unnecessary runtime dependency edge.

wchargin:
In the future, please request re-review after making changes that aren’t
just responses to the review comments. The diff from the last approval
on this PR to the code that was actually merged is substantial, and
I would have liked to request some changes.

This is the flip-side to “nit and approve”—we can only comfortably
pre-approve if we don’t have to worry about the approval being taken
advantage of.

Also, in the future, please make sure to check the commit message before
merging. Usually, it’s most appropriate to just paste the PR description
into the commit message—that way, the commit message itself is part of
the review. What actually got merged here has a lot of duplication as
well as transient information that doesn’t need to be in the permanent
log (like “Address review comments from…”):
<https://github.com/sourcecred/sourcecred/commit/59000eb64ef10c18686e143e9851c71cbadb40eb>

(I wish that GitHub would automate this and not suggest a trash commit
message by default, and I’ve requested it, but for now we’ll have to
make do with what we have.)

wchargin:
Okay. Just getting back to this. Let me elaborate.

First, there are a couple simple errors. The `console.log` statements
for debugging need to go. The `warn`s for retrying and reporting errors
probably should have clearer messages/presentation. And the prefix test
via `new RegExp("^" + name)` is only barely safe (for non-local
reasons), and should be simplified to `taskId.startsWith(name)`.

But these are all simple local changes. If they were the only issues,
I’d just fix them. More important is that this PR represents a shift in
design. Prior to this change, the `TaskReporter` interface was only used
as a simple… well, task reporter. It formatted messages and kept track
of a stopwatch. But it was never load-bearing. As of this PR, the actual
logic depends on the `taskReporter.activeTasks` map; there is no
precedent for this in the codebase outside of test code.

Per my [comment on the other PR][1], I don’t think that this is a good
pattern, and was actually planning to remove it from the existing test
code. Conceptually, it’s a mismatch, which you can see from the fact
that only `LoggingTaskReporter` supports this behavior, not the
`TaskReporter` interface itself. At an implementation level, anything
that’s load-bearing like this needs to be more careful about its
bookkeeping. Representing tasks only by string names is fine when all
you do is print them to the console, but relying on the `${name}:`
prefix format to determine which tasks should be cancelled is too much.

[1]: https://github.com/sourcecred/sourcecred/pull/2164#discussion_r478009866

If my review had been re-requested here, I would have held this
discussion before approving (note that I never did approve). Of course,
the approval status of this PR doesn’t matter anymore, but please
consider my requests still active. @topocount, can you take care of the
small issues and propose a path forward for the design change? Happy to
work with you or advise, as you desire.

(Generally what I’d suggest in the case of a premature merge is to just
roll back and then re-open a PR for review, but in this case it’s been a
few weeks so that’s a less attractive option. The latency there is on
me, so my apologies for that.)

topocount:
@wchargin I apologize for hastily merging this.

I think a more robust solution would be a task tree. As per your comments about the map, we would move this stateful construct for tasks outside of the task reporter and create something of a ManagedTasks construct, with methods for recirsively pruning the task tree at the point of failure. I could understand if this is a bit heavy-weight of a solution for the problem though.

Earlier in this PR I tried to instantiate new task reporters in subcontexts. This cleaned up the logic involving filtering and killing tasks but I got the sense that it was considered an inelegant approach. I'd like to better understand why it's not a preferred approach, short of refactoring the inter task map out of the LoggingTaskReporter.

PR #2163, opened 2020-08-25
Update special distribution to use material-ui

befitsandpiper:
Test Plan: manual FE testing, pass existing tests
Screenshot from 2020-08-25 18-17-33

Success state after transferring:
Screenshot from 2020-08-25 18-18-08


befitsandpiper:
Sorry, I'm seeing a bug I didn't see before, false alarm.

PR #2164, opened 2020-08-25
Add Scoped Task Reporter

topocount:
ScopedTaskReporter is a higher-order task reporter for generating
opaque scopes meant to be passed into child contexts. In this case, a
scope is a log prefix indicating the parent context in which the
current task is running. This allows for reliable filtering and
searching on existing tasks by prefix. Any instance of a TaskReporter
may be used with the scopedTaskReporter.

test plan: unit tests included

Paired with @wchargin

Co-authored-by: William Chargin <wchargin@gmail.com>

### DIFF
--- a/src/util/taskReporter.js
+++ b/src/util/taskReporter.js
@@ -116,12 +116,18 @@ export class SilentTaskReporter implements TaskReporter {
  * in which the current task is running.
  *
  * This allows for reliable filtering and searching on existing tasks
- * by prefix, so long as no child contexts attempt to utilize the same prefix
- * and then filter on it. Care should be taken to ensure that the same prefix does
- * not exist in different scopes, so far as error handling is concerned, or a
- * filter may incorrectly catch and finish an sill-running scope while error-handling
+ * by prefix. Care should be taken to ensure that the same prefix does
+ * not exist in peer task contexts, so far as error handling is concerned,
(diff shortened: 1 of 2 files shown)


wchargin:
I see that this has been updated, but I’m not sure if you’re ready for
re-review yet. When you are, could you please re-request my review by
pushing the button in the “Reviewers” panel?

![Screenshot of “Reviewers” panel with a “refresh” icon-button next to
`@wchargin`’s name][ss]

[ss]: https://user-images.githubusercontent.com/4317806/91379036-9b22fb00-e7d6-11ea-9118-70a57c6c4863.png

decentralion:
> Since my initial review comment led to a full redesign here, it’s good
form to indicate that in the PR description and commit message. To
indicate this, you should append
> 
> > Co-authored-by: William Chargin <wchargin@gmail.com>

Surprised that you're proposing co-authored by but not paired-with? Is there a reason to do this in a way that SC isn't aware of?

wchargin:
I think that we should support `Co-authored-by`, but you’re right that
that’s a separate issue (#2167), so I’m fine with just using an inline
`Paired with @wchargin.` for now.

topocount:
@wchargin I've updated the PR description to include both the online co-author and paired trailer. Should I only use the co-author tag?

PR #2165, opened 2020-08-26
Update special distribution to use material-ui

befitsandpiper:
Test Plan: manual FE testing, pass existing tests
Screenshot from 2020-08-25 18-17-33

Success state after distributing:
Screenshot from 2020-08-25 22-01-36

### DIFF
--- a/src/ui/components/SpecialDistribution.js
+++ b/src/ui/components/SpecialDistribution.js
@@ -1,38 +1,78 @@
 // @flow
-
 import React, {useState} from "react";
-import {div, fromInteger, fromFloatString, ZERO} from "../../ledger/grain";
-import {type IdentityId} from "../../ledger/identity";
+import {Button, Container, TextField} from "@material-ui/core";
+import {Alert} from "@material-ui/lab";
+import {makeStyles} from "@material-ui/core/styles";
+import {div, fromFloatString, lt, ZERO} from "../../ledger/grain";
 import {type Account} from "../../ledger/ledger";
+import {type CurrencyDetails} from "../../api/currencyConfig";
 import AccountDropdown from "./AccountSelector";
 import {computeAllocation} from "../../ledger/grainAllocation";
 import * as uuid from "../../util/uuid";
 import type {TimestampMs} from "../../util/timestamp";
 import {useLedger} from "../utils/LedgerContext";
 
-export const SpecialDistribution = () => {
+const useStyles = makeStyles((theme) => ({
+  root: {
+    width: "80%",
+    minWidth: "700px",
+    margin: "0 auto",
+    padding: "0 5em 5em",
+  },
+  arrowBody: {
+    color: theme.palette.text.primary,
+    flex: 1,
+    background: theme.palette.background.paper,
+    padding: "5px 20px",
+    display: "flex",
+    alignItems: "center",
+  },
+  triangle: {
+    width: 0,
+    height: 0,
+    background: theme.palette.background,
+    borderTop: "30px solid transparent",
+    borderBottom: "30px solid transparent",
+    borderLeft: `30px solid ${theme.palette.background.paper}`,
+  },
+  dropdownWrapper: {
+    border: `1px solid ${theme.palette.text.primary}`,
+    borderRadius: "50px",
+  },
+  centerRow: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+  },
+  element: {flex: 1, margin: "20px"},
+  arrowInput: {display: "inline-block"},
+  headerText: {color: theme.palette.text.primary},
+}));
+
+type SpecialDistributionProps = {|+currency: CurrencyDetails|};
+
+export const SpecialDistribution = ({
+  currency: {name: currencyName},
+}: SpecialDistributionProps) => {
   const {ledger, updateLedger} = useLedger();
 
+  const classes = useStyles();
   const [credTimestamp, setCredTimestamp] = useState<TimestampMs>(+Date.now());
-  const [recipient, setRecipient] = useState<IdentityId | null>(null);
-  const [amount, setAmount] = useState<string>("0");
+  const [recipient, setRecipient] = useState<Account | null>(null);
+  const [amount, setAmount] = useState<string>("");
   const [memo, setMemo] = useState<string>("");
+  const [successMessage, setSuccessMessage] = useState<string>("");
 
-  const setRecipientFromAccount = (acct: Account | null) => {
-    setRecipient(acct ? acct.identity.id : null);
-  };
-
-  const submitDistribution = (e) => {
-    e.preventDefault();
-    if (recipient != null && amount !== ZERO) {
+  const submitDistribution = () => {
+    if (recipient && amount && amount !== ZERO) {
       const policy = {
         policyType: "SPECIAL",
         budget: fromFloatString(amount),
         memo,
-        recipient,
+        recipientId: recipient.identity.id,
       };
       const allocation = computeAllocation(policy, [
-        {cred: [1], paid: ZERO, id: recipient},
+        {cred: [1], paid: ZERO, id: recipient.identity.id},
       ]);
       const distribution = {
         id: uuid.random(),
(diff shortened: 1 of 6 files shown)


src/ui/components/SpecialDistribution.js
```diff
@@ -41,88 +76,118 @@ export const SpecialDistribution = () => {
...
+          className={classes.element}
+          type="number"
+          value={credTimestamp}
+          placeholder="Timestamp"
+          onChange={(e) => {
+            const newTimestamp = +e.currentTarget.value;
+            if (!Number.isFinite(newTimestamp)) {
+              throw new Error(`invalid timestamp: ${e.currentTarget.value}`);
+            }
+            setCredTimestamp(newTimestamp);
```
topocount, line 162: A couple thoughts about this component:

1. Do we need it? I suggest getting rid of it above, assuming this component is to be used to create a user-generated timestamp. Is a user-modifiable timestamp a desired feature here?
2. Assuming the answer to 1 is yes, let's allow users to configure this using a date picker of some sort. I personally struggle to eyeball unix epoch timestamps, and I imagine most other people to do.

src/ui/components/SpecialDistribution.js
```diff
@@ -41,88 +76,118 @@ export const SpecialDistribution = () => {
...
       const nextLedger = ledger.distributeGrain(distribution);
       updateLedger(nextLedger);
-      setAmount(fromInteger(0));
-      setRecipient(null);
+      setSuccessMessage(
+        `You distributed ${amount} grain to ${recipient.identity.name}!`
+      );
+      setAmount("");
+      setMemo("");
+      setCredTimestamp(+Date.now());
```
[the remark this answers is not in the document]
befitsandpiper: this is to reset the timestamp to a new value after a distribution is made.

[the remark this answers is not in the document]
befitsandpiper: much as we default the timestamp initially. But we shouldn't repeat timestamps.

src/ui/components/SpecialDistribution.js
befitsandpiper: I'm all for removing it, and just setting the timestamp in the submit function, but it's up to @decentralion

src/ui/components/SpecialDistribution.js
```diff
@@ -1,38 +1,73 @@
...
 import AccountDropdown from "./AccountSelector";
 import {computeAllocation} from "../../ledger/grainAllocation";
 import * as uuid from "../../util/uuid";
 import type {TimestampMs} from "../../util/timestamp";
 import {useLedger} from "../utils/LedgerContext";
 
+const useStyles = makeStyles((theme) => ({
+  root: {
+    width: "80%",
+    minWidth: "1100px",
```
hammadj, line 17: Seems pretty high for a minWidth. IMO should only set the minWidth of specific components if they need to have a minWidth set. Root level minwidth just makes it hard to support smaller screens.

src/ui/components/SpecialDistribution.js
```diff
@@ -1,38 +1,73 @@
...
+  triangle: {
+    width: 0,
+    height: 0,
+    background: theme.palette.background,
+    borderTop: "30px solid transparent",
+    borderBottom: "30px solid transparent",
+    borderLeft: `30px solid ${theme.palette.background.paper}`,
+  },
+  dropdownWrapper: {
+    border: "1px solid #9EADBA",
```
hammadj, line 38: Should use color values from the theme instead of hardcoding

src/ui/components/SpecialDistribution.js
```diff
@@ -1,38 +1,73 @@
...
-    e.preventDefault();
-    if (recipient != null && amount !== ZERO) {
+  const submitDistribution = () => {
+    if (recipient && amount && amount !== ZERO) {
       const policy = {
         policyType: "SPECIAL",
         budget: fromFloatString(amount),
         memo,
-        recipient,
+        recipient: recipient.identity.id,
```
hammadj, line 67: The name of this field should be `recipientId` instead of `recipient` since its not a recipient, its just the id

src/ui/components/SpecialDistribution.js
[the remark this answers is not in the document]
hammadj: IMO we should use the timestamp when the user submits

src/ui/components/SpecialDistribution.js
hammadj: Users might want to do special distributions in the past. e.g. For MetaGame, we would want to distribute SEEDs in the ledger based on when we did our distribution on chain a few months ago.

src/ui/components/SpecialDistribution.js
```diff
@@ -41,88 +76,118 @@ export const SpecialDistribution = () => {
...
+          distribute special grain
+        </Button>
+        <Button
+          size="large"
+          color="primary"
+          variant="contained"
+          className={classes.element}
+          onClick={postLedger}
+        >
+          save ledger to disk
```
hammadj, line 188: Should Capitalize This

src/ui/components/SpecialDistribution.js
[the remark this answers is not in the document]
topocount: My bad! I got th `ledgerTimestamp` and `credTimestamp` values mixed up when inspecting the ledger.json output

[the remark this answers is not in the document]
topocount: I recommend putting all these state resets behind a `reset()` function but that's a nit

src/ui/components/SpecialDistribution.js
befitsandpiper: According to conversation in discord, it seems this should stay

src/ui/components/SpecialDistribution.js
befitsandpiper: set to 700px instead. Any smaller than this and it really starts to look terrible without a minimum width.

src/ui/components/SpecialDistribution.js
```diff
@@ -1,38 +1,79 @@
...
+  const [amount, setAmount] = useState<string>("");
   const [memo, setMemo] = useState<string>("");
+  const [successMessage, setSuccessMessage] = useState<string>("");
 
-  const setRecipientFromAccount = (acct: Account | null) => {
-    setRecipient(acct ? acct.identity.id : null);
+  const resetState = () => {
+    setAmount("");
+    setMemo("");
+    setCredTimestamp(+Date.now());
```
decentralion, line 64: It's a feature, not a bug, that the timestamp and memo don't reset. Oftentimes when I"m using this tool I want to make a few distributions with the same TS and memo, e.g. "legacy Grain distribution for July" going out to a few accounts.

src/ui/components/SpecialDistribution.js
```diff
@@ -1,38 +1,79 @@
...
+}));
+
 export const SpecialDistribution = () => {
   const {ledger, updateLedger} = useLedger();
 
+  const classes = useStyles();
   const [credTimestamp, setCredTimestamp] = useState<TimestampMs>(+Date.now());
-  const [recipient, setRecipient] = useState<IdentityId | null>(null);
-  const [amount, setAmount] = useState<string>("0");
+  const [recipient, setRecipient] = useState<Account | null>(null);
```
decentralion, line 56: Why would we switch this to have an account as state? Note that this implies the account state in the component can desynchronize from the account in the ledger. It's nice to use the ledger as the source of truth for all account data.

src/ui/components/SpecialDistribution.js
befitsandpiper: ah, okay, that's fair I will revert.

src/ui/components/SpecialDistribution.js
befitsandpiper: The main reason I did this is we do the same thing in the transfer grain UI, and I wanted access to the name in the alert. Should I just have a separate state variable for the recipientName? I had two variables before in the transfer grain UI and you had me just set the account in state there. Is there a reason that we should treat these two UIs differently in that regard?

src/ui/components/SpecialDistribution.js
```diff
@@ -41,88 +76,116 @@ export const SpecialDistribution = () => {
       };
       const nextLedger = ledger.distributeGrain(distribution);
       updateLedger(nextLedger);
-      setAmount(fromInteger(0));
-      setRecipient(null);
+      setAmount("");
+      setRecipient(nextLedger.account(recipient.identity.id));
+      setSuccessMessage(
+        `You distributed ${amount} grain to ${recipient.identity.name}!`
```
decentralion, line 82: nit: should use the name from currency details, cf: https://github.com/sourcecred/sourcecred/pull/2145

KuraFire:
I know I missed the boat on this one, sorry! 😅 That said, also, I don't have a local Cred Explorer running so it's hard for me to review stuff on design & interactions. Static & cropped images may not represent the full screen, and in this case for example I don't see an input field for the name of the identity you are transferring Grain _from_. Does that mean it now only supports transferring from your own account? Is this a new view that is separate from the admin ledger view?

(In the future I'll aim to keep a closer watch on my GitHub notifications, but feel free to ping me on Discord as reminders if it's not too troublesome and you really want my eyes on it ❤️ 😊

PR #2214, opened 2020-09-08
Account Overview: make table fixed height

decentralion:
This refactors the Account Overview page so that we now set a max height
(40 ems) on the table, and the user can scroll within the table to find
more entries. This is cleaner than simply having the table spill over
infinitely.

We might want to choose a more principled size than 40 ems (e.g. a % of
view) but I couldn't get that to work and I figured this minimal change
is still an improvement. I'd also like for us to set up scrollbar styles
that are more in line with our dark theme.

I set the `stickyHeader` property so that the header will always be visible at the top.

Test plan: Load with an instance which has many accounts (i.e. not the
test instance) and observe the nice scrolling behavior. Remember that
you can run `yarn start --instance PATH_TO_INSTANCE`.

### DIFF
--- a/src/ui/components/AccountOverview.js
+++ b/src/ui/components/AccountOverview.js
@@ -12,13 +12,23 @@ import {type Account} from "../../ledger/ledger";
 import {type CurrencyDetails} from "../../api/currencyConfig";
 import * as G from "../../ledger/grain";
 import {useLedger} from "../utils/LedgerContext";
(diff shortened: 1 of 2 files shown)


decentralion:
Here's how it looks post-change, using the Maker instance as an example. Note the scroll bar, and the fact that I've scrolled midway in but the header is still visible. 
image

befitsandpiper (approved):
LGTM! Seems like a good first step to improve usability

PR #2215, opened 2020-09-08
disable esModules in css-loader

topocount:
css-loader v4 has a breaking change that enables esModules for imports
by default. setting the config to false returns behavior to what we were
using in v3

This impacts our current build because the `require` here: https://github.com/sourcecred/sourcecred/blob/0dfeca47da0c2a566647309bb5e3720584de9f3a/src/ui/server.js#L42

is imported as a module without this added config to set `esModule: false`

test plan: `yarn && yarn start` and inspect the index.html file to ensure the
index.css body styling is correctly loaded under the <script> tag

### DIFF
--- a/config/webpack.config.web.js
+++ b/config/webpack.config.web.js
@@ -218,6 +218,9 @@ async function makeConfig(
             {
               test: /\.css$/,
               loader: "css-loader", // TODO(@wchargin): add csso-loader
+              options: {
+                esModule: false,
+              },
             },
             {
               test: /\.svg$/,


decentralion (approved):
Confirming this fixes the errant scroll bar in the UI.
Suggest removing the `yarn.lock` changes, it's clean on master. (I think we're getting unclean yarn.lock in master when we merge multiple dependabot PRs quickly--after one merges it takes a few mins to rebase the others, I think.)

Issue #2204, opened 2020-09-08
Last fixes for the MUI explorer frontend 


topocount:
https://github.com/sourcecred/sourcecred/pull/2150#pullrequestreview-483716130

topocount:
Also worth noting: the filter dropdown text is still white. and this is the only view with a white background. This is probably worth addressing here as well

PR #2225, opened 2020-09-09
Explorer tweaks and polish

befitsandpiper:
fix issues from #2150 & use dark theme #2204 

Also switches to using hooks and ES6+ syntax where the changes are simple.

Test Plan: existing test suite, manual testing

Main visual difference vs previous Explorer UI is the switch to using a dark theme:

Screenshot from 2020-09-09 13-16-40

**Note: There are things not included here that I would normally go over with a finer tooth comb, like changing the Explorer file itself to use hooks, and probably a number of code simplications within this set of components. However we are about to do a major overhaul of the Explorer anyway, so spending the time on that now seems wasteful**


decentralion (commented):
It's hard to review this because it refactors the components (resulting in no usable diffs) while also changing them. Can you please split this up into two PRs, where the first refactors without making any style or functional changes, and the second updates the styles?

decentralion:
Hey @befitsandpiper, I'd still like to see this split up into smaller PRs. E.g. styling in one PR, refactoring in another PR, and fixing #2150 issues in a third. You can write them as a sequence of dependent branches (i.e. A merges into master, B merges into A, and C merges into B) and GitHub makes it pretty easy to merge them.

I recognize it may seem a bit pedantic, but I think holding the line on atomic commits will make our work together as a team much smoother. As an added benefit, with the current Cred semantics, it will be 3x more Cred for you. :) 

Also, just in case we weren't synced on this: I don't want the new explorer to get built as an overwrite of the existing UI. Rather, I'd like to have a new component (e.g. explorerv2, or overview) which is accessible but hidden (not in sidebar), and then once it's ready we will drop it in as a replacement for this one. So work on this PR shouldn't block developing the new explorer.

decentralion (commented):
Hey @befitsandpiper, I'd still like to see this split up into smaller PRs. E.g. styling in one PR, refactoring in another PR, and fixing #2150 issues in a third. You can write them as a sequence of dependent branches (i.e. A merges into master, B merges into A, and C merges into B) and GitHub makes it pretty easy to merge them.

I recognize it may seem a bit pedantic, but I think holding the line on atomic commits will make our work together as a team much smoother. As an added benefit, with the current Cred semantics, it will be 3x more Cred for you. :)

Also, just in case we weren't synced on this: I don't want the new explorer to get built as an overwrite of the existing UI. Rather, I'd like to have a new component (e.g. explorerv2, or overview) which is accessible but hidden (not in sidebar), and then once it's ready we will drop it in as a replacement for this one. So work on this PR shouldn't block developing the new explorer.

befitsandpiper:
Haha, I didn't want to be greedy, and I was trying to be quick ;)
Ah, I think I misunderstood there then, I thought this was being entirely replaced with the new UIs.
Okay, I will slow down, split this up more and work on it in conjunction with the new explorer mockup then.

PR #2228, opened 2020-09-09
dependencies config: allow dependency periods

decentralion:
This adds support for the dependencies config so that you can specify
how the weight changes over time (e.g. starting minting only at a
particular time) rather than only setting a consistent policy across all
time.

Test plan: Unit tests updated to accomodate the new case.

### DIFF
--- a/src/api/dependenciesConfig.js
+++ b/src/api/dependenciesConfig.js
@@ -44,6 +44,11 @@
  */
 import * as C from "../util/combo";
 import {Ledger} from "../ledger/ledger";
+import {
+  type TimestampISO,
+  timestampISOParser,
+  fromISO,
+} from "../util/timestamp";
 import {type IdentityId, type Name, nameParser} from "../ledger/identity";
 import {parser as uuidParser} from "../util/uuid";
 import {type DependencyMintPolicy} from "../core/dependenciesMintPolicy";
@@ -76,9 +81,10 @@ export type DependencyConfig = {|
   // from the start of the project up until the first specified weightPeriod.
   // If this is set and there are no explicit weightPeriods, this weight will
   // prevail for the whole timeline of the project.
-  // Currently, weight periods aren't yet implemented, so whatever startWeight is
-  // set will apply across all time.
   +startWeight: number,
+  // The time periods for which we're minting Cred. Each period has a start time and a
+  // mint weight; it is assumed to end as soon as the next period begins.
+  +periods?: $ReadOnlyArray<MintPeriod>,
   // Whether the dependency should be "active" for Grain collection by default.
   // If this is true, then when the corresponding account is auto-created in the
   // ledger, it will also be automatically activated.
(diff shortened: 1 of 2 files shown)


src/api/dependenciesConfig.js
```diff
@@ -76,9 +81,8 @@ export type DependencyConfig = {|
   // from the start of the project up until the first specified weightPeriod.
   // If this is set and there are no explicit weightPeriods, this weight will
   // prevail for the whole timeline of the project.
-  // Currently, weight periods aren't yet implemented, so whatever startWeight is
-  // set will apply across all time.
   +startWeight: number,
+  +periods?: $ReadOnlyArray<MintPeriod>,
```
topocount, line 85: What do you think about naming this "additionalPeriods"? It wasn't clear to me that the initial period shouldn't be set here as well

topocount (approved):
I just have one naming semantics nit. Otherwise it looks good!

src/api/dependenciesConfig.js
decentralion: We could also remove the `startWeight` setting, and just have users set periods if they want to do any minting. It's a simpler API. What do you think?

hammadj: +1, should simplify to use only periods

src/api/dependenciesConfig.js
```diff
@@ -148,8 +163,13 @@ export function toDependencyPolicy(
...
-  const period = {weight: config.startWeight, startTimeMs: -Infinity};
-  const policy = {address, periods: [period]};
-  return policy;
+  const startPeriod = {weight: config.startWeight, startTimeMs: -Infinity};
+  const processedPeriods = rawPeriods.map(({startTime, weight}) => ({
+    weight,
+    startTimeMs: fromISO(startTime),
+  }));
+  const periods = [startPeriod].concat(processedPeriods);
+  return {address, periods};
```
hammadj, line 174: Removing start period will simplify this greatly as well

src/util/timestamp.js
```diff
@@ -66,3 +68,18 @@ export function validate(timestampMs: number): TimestampMs {
...
+}
+
+export const timestampMsParser: C.Parser<TimestampMs> = C.fmap(
+  C.number,
+  validateTimestampMs
+);
+export const timestampISOParser: C.Parser<TimestampISO> = C.fmap(
+  C.string,
+  validateTimestampISO
+);
```
hammadj, line 85: Did something similar in initiatives parser:
```
const TimestampParser = C.fmap(C.string, (t) => toISO(fromISO(t)));
```

Not sure if there's any practical benefit in forcing it to go through our `toISO` function, but thought I'd share.

hammadj (approved):
Simplifying to only use periods would be cleaner, but looks good otherwise!

befitsandpiper (approved):

decentralion:
Cool, thanks for the feedback yall. I'm going to merge this as is, and remove the startWeight in a separate PR.

src/util/timestamp.js
decentralion: That's a neat trick, could have made the implementation for https://github.com/sourcecred/sourcecred/pull/2226 a bit more concise. WOuldn't have needed to factor out the validator function.

PR #2230, opened 2020-09-09
Refactor Explorer.js into many components

befitsandpiper:
Test Plan: existing tests, check that UI still works

The refactors here should make future Explorer updates easier, and improves readability of the code.

No UI changes in this branch.

**Note: The Refactors in this branch made it more obvious that there is a circular dependency between the *FlowsRow* and *NodeRow* Components. While this is supported in ES6 and the UI functions correctly, it is a major anti-pattern. It may not be possible to avoid this due to the cyclical nature of the Cred Graph, but it is something we should remain aware of**


decentralion (approved):
Approving, didn't test it manually but trusting that CI passing is sufficient for a simple refactor.

PR #2238, opened 2020-09-10
Fix inability for credbot to push to master in Release workflow

hammadj:
Updated the workflow to fetch all history for commits which is required for the refs to properly be pushed into
 master according to the docs of the push action (https://github.com/ad-m/github-push-action). Also disables
 persist-credentials which prevents the auto-generated GITHUB_TOKEN for actions from being used.

Verified successful run on a fork with same repo security settings (https://github.com/hammadj/sourcecred/runs/1098839653?check_suite_focus=true)

Test Plan: Create a release and ensure the action runs successfully and updates master branch with new version.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,7 +8,10 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@master
+        with:
+          persist-credentials: false # allows usage of custom access token with admin permissions
+          fetch-depth: 0 # fetches all commits / tags. Otherwise, it fails to push refs to dest branch
       - uses: actions/setup-node@v1
         with:
           node-version: 12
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -20,14 +23,14 @@ jobs:
       - name: Test
         run: yarn test --full --ci
         env:
-          SOURCECRED_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SOURCECRED_GITHUB_TOKEN: ${{ secrets.SOURCECRED_GITHUB_TOKEN }}
```
decentralion, line 26: Why this change? If we want the Credbot token for its permissions, we could use `CREDBOT_TOKEN`. But since this is for `yarn test`, the default GITHUB_TOKEN should be fine, no?

decentralion (commented):
After this change, do we need the CREDBOT admin token to be present at all?

.github/workflows/publish.yml
hammadj: Adding `persist-credentials: false` disables the default GITHUB_TOKEN that goes into Actions, so changed it to use this one instead.

hammadj:
> After this change, do we need the CREDBOT admin token to be present at all?

Yea we still need it there

.github/workflows/publish.yml
decentralion: Got it.

decentralion (approved):
Will cut a release today most likely, so we'll get to test this in prod soon. :)

PR #2241, opened 2020-09-11
Fix issue in calculating Grain distributions

### LINKED ISSUE #2239: Grain distributions sometimes do not go out on schedule
We've had a number of automated Grain distributions fail to distribute any Grain. After reviewing the code, I'm fairly confident the culprit is the following logic:
https://github.com/sourcecred/sourcecred/blob/c5688c45b310635c1af0287c847d1e49a8d50626/src/ledger/applyDistributions.js#L70-L74

As you can see, we ignore the latest Cred interval, on the assumption that the latest interval is incomplete (i.e. for the current week). However, the intervals are populated based on the timestamps of real contributions. Therefore, if we run the grain distribution shortly after the interval start (like we do in GitHub actions), then it's possible that the latest interval will actually be complete.

To fix this, we should change the logic so it takes the current timestamp into account, rather than blindly assuming that the latest interval is not yet complete.

decentralion:
This commit fixes #2239. See the bug for context. Honestly, the API for
the Grain command is really wonky and should probably be changed more,
but for now I'll settle for just fixing the bug.

Test plan: There's a unit test added for the case where the latest
interval is complete.

### DIFF
--- a/src/cli/grain.js
+++ b/src/cli/grain.js
@@ -42,7 +42,8 @@ const grainCommand: Command = async (args, std) => {
   const distributions = applyDistributions(
     distributionPolicy,
     credView,
-    ledger
+    ledger,
+    +Date.now()
   );
 
   let totalDistributed = G.ZERO;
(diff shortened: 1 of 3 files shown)


befitsandpiper (approved):
LGTM

topocount (approved):

PR #2242, opened 2020-09-11
dependencies config: specify all deps via periods

decentralion:
This removes the `startWeight` setting for the dependencies config, in
favor of making `periods` mandatory. This is a simpler API, as discussed
[here](https://github.com/sourcecred/sourcecred/pull/2228#discussion_r486663635).

It also discourages having "infinite-time" dependencies, in favor of
choosing when the dependency truly began--which is semantically better.

One interesting thing to consider: how can we have the default period
for dependency on SourceCred consistently start around the time when
someone forks the template instance? I considered setting up a GitHub
action for this, but then it would keep getting updated once people fork
the template repo. Food for thought.

Test plan: `yarn test` passes. Note this is a breaking change, but AFAIK
the only user right now is the template instance; I'll fix it when I
upgrade the instance to this version.


befitsandpiper (approved):
Feels like this is a really nice simplification conceptually, moving away from infinite time periods.

topocount (approved):
it might be good to include testing for when no periods are entered, even if that validation is not explicitly handled outside the dependencies module itself.

decentralion:
I added a test case with an empty periods array. We also have testing downstream when the periods get used:
https://github.com/sourcecred/sourcecred/blob/99f27742e65a1d81ab46af9b7840196cdc2a661e/src/core/dependenciesMintPolicy.test.js#L12-L17

PR #2247, opened 2020-09-11
Prettify markdown files

decentralion:
I've updated `yarn prettify` and `yarn check-pretty` so that we now
prettify markdown files.

Test plan: Pretty trivial change. Running `yarn prettify` resulted in
the markdown changes included in this commit. Running `yarn
check-pretty` passes, as does `yarn test`. Note that `yarn check-pretty`
has been modified to also check markdown.


wchargin (approved):
I forgot about the Factorio plugin. :-)

PR #2248, opened 2020-09-11
Prettify yml files

decentralion:
This is basically #2247, but for yml. Similar approach and test plan.

Test plan: CI passes.


topocount (approved):
thanks for enabling this

PR #2249, opened 2020-09-11
Dark theme and associated  styling changes

befitsandpiper:
Switch to dark theme, fix some needing styling organization as well.

Test Plan: tests pass, manual FE test

Screenshot from 2020-09-11 19-42-33


decentralion:
Issues I see:

* Hyperlinks are nigh unreadable:
image

* Cred over time chart is very hard to read
image

decentralion (commented):
Issues I see:

* Hyperlinks are nigh unreadable:
image

* Cred over time chart is hard to read (too little contrast)
image

befitsandpiper:
Switched to cyan over dark blue:
Screenshot from 2020-09-12 23-22-20

src/ui/components/Explorer/Explorer.js
```diff
@@ -17,6 +17,7 @@ import {
 } from "@material-ui/core";
 import KeyboardArrowDownIcon from "@material-ui/icons/KeyboardArrowDown";
 import KeyboardArrowUpIcon from "@material-ui/icons/KeyboardArrowUp";
+import {StyleSheet, css} from "aphrodite/no-important";
```
youngkidwarrior, line 20: material ui has its own styled-component solution. Might want to use those instead to be consistent. 
https://material-ui.com/styles/basics/#styled-components-api

src/ui/components/Explorer/Explorer.js
```diff
@@ -31,6 +32,31 @@ import {type TimelineCredParameters} from "../../../analysis/timeline/params";
 
 import NodeRow from "./NodeRow";
 
+const styles = StyleSheet.create({
+  combobox: {margin: "0px 32px 16px"},
+  menuHeader: {fontWeight: "bold"},
+  divider: {backgroundColor: "#F20057", height: "2px"},
```
youngkidwarrior, line 38: Nit: I see you changed the divider color to gray. I rather liked the pink underline. 😊️ It'd be really cool if it changed colors when hovered over.

image

src/ui/components/Explorer/Explorer.js
befitsandpiper: This doesn't work with the explorer, bc the material-ui makeStyles only works with functional components. There is a HOC solution as well, but I had trouble getting it working with flow. I will change this on another branch where I am switching everything to hooks and functional components.

src/ui/components/Explorer/Explorer.js
befitsandpiper: I don't think I changed this... but ultimately the color scheme will be decided by Faruk and the design team.

src/ui/components/Explorer/Explorer.js
youngkidwarrior: The styled components API should work with components I believe.

decentralion (commented):
I think it now shows the same link color for visited and unvisited links. Can we keep two separate colors  for them? (For consistency with the rest of the web.)

Also: It would be nice to restyle the sliders to look good on the dark background:
image

befitsandpiper:
Sliders are handled in this PR: https://github.com/sourcecred/sourcecred/pull/2253

befitsandpiper:
added pink color for visited state.

src/ui/components/Explorer/Explorer.js
befitsandpiper: https://stackoverflow.com/questions/56432167/how-to-style-components-using-makestyles-and-still-have-lifecycle-methods-in-mat

If you do not use the HOC method, our ususal way of using makeStyles doesn't work inside of class components. This is bc makeStyles uses react hooks, and react forbids hooks within class components.

There's a way to use HOC instead of the normal makeStyles classes pattern, but I was having trouble getting this to work with flow. I could spend more time fiddling with this, but I will be converting this component to a functional component soon anyway, so I figure I will just use our standard pattern then.

befitsandpiper:
I didn't think about using the material-ui Link component till just now for some reason, but I tested it and they also don't have a different visited state for some reason...

src/ui/components/Explorer/Explorer.js
youngkidwarrior: This is using styled. Should work with components.

```
Import { styled } from '@material-ui/core/styles'; 
import Button from '@material-ui/core/Button'; 
const MyButton = styled(Button)({ background: 'linear-gradient(45deg, #FE6B8B 30%, #FF8E53 90%)', border: 0, borderRadius: 3, boxShadow: '0 3px 5px 2px rgba(255, 105, 135, .3)', color: 'white', height: 48, padding: '0 30px', });
```

befitsandpiper:
Added Pink for visited state, and changed links to use the MUI Link Component over anchor tags.

decentralion (approved):

PR #2250, opened 2020-09-12
Add temporary style fixes for color and padding

youngkidwarrior:
Temporary color and padding fixes for styles until UI gets updated to using dark theme.


wchargin (approved):
This is okay with me for a temporary fix so that we can unbreak the live
site. But I don’t think that it actually addresses the root cause. The
`react-admin` framework is setting `color: #fff` on one of its root
elements. That’s obviously not what we want, so we should figure out why
it’s doing that and how to stop it from doing so rather than peppering
around `color: black`s here and in other PRs.

But okay with me to merge this for now.

youngkidwarrior:
@wchargin yeah for further clarification, I believe it's our material-ui theme that sets that value. We have a dark theme on the page, but use use a white component for explorer

PR #2253, opened 2020-09-13
Mui explorer regressions and spacing

befitsandpiper:
In master currently, the weights explorer page has spacing issues. This is because we partially converted to use MUI grid components, but still had a number of raw html elements that don't work well with the MUI stylings. Converting these raw html elements in the Explorer and WeightConfig files to MUI components allows us to fix these spacing issues. It also improves our consistency since we are gradually switching to using MUI everywhere in our UI.

I have also removed some extra indentation in the CredRow file and changed alignment in the TableCells within the Explorer file. This improves the alignment and look of the Table on the Explorer page.

The Collapse element has been removed from CredRow. This element served no functional purpose and was causing an error in the console because it was a non-table element within the table portion of the DOM tree.

Test Plan: unit tests are passing, check to make sure UI still works as expected.

Screenshot from 2020-09-15 16-27-57

Screenshot from 2020-09-15 16-22-00


src/ui/weights/WeightSlider.js
```diff
@@ -55,25 +56,30 @@ export type Props = {|
...
+          <Grid item xs={5}>
+            <Slider
+              value={weightToSlider(this.props.weight)}
+              min={MIN_SLIDER}
+              max={MAX_SLIDER}
+              step={1}
+              valueLabelDisplay="off"
+              onChange={(_, val) => {
+                const weight: Weight = sliderToWeight(val);
+                console.log({weight});
```
decentralion, line 73: looks like debug code

decentralion (commented):
Can you rewrite the PR description to be clearer? for example, what regressions is this fixing? what collapse is being removed?

it seems like the major change in this PR is switching to use MUI Grid where appropriate, and that's not even mentioned in the description. can you provide some context on why we are using grid and how you decided where to use it?

the slider change is logically separate and should be another PR.

src/ui/weights/WeightSlider.js
befitsandpiper: oops, yep, that's debug code. Maybe we should have a lint rule that errors on lines with console logs.

decentralion: That's a reasonable suggestion. There are some cases where we want to do it deliberately, but we could just disable the lint rule for those lines.

decentralion (approved):
Thanks for the clarified description. I did a true blind A-B test (opened up both before and after this PR and explored them both, without being sure which was which). And I can confirm this one is definitely better 👍🏽. It also doesn't have the console errors, which is great.

Issue #2257, opened 2020-09-14
Issues with SC release process


decentralion:
Issues with the current release flow:

- We tag and then update package.json, which means if you check out the tag, it has the wrong version
- We don't update [version.js](https://github.com/sourcecred/sourcecred/blob/master/src/core/version.js), so `sourcecred --version` will lie.
- We turned off 2fa on npm package deploy, which gives us worse security around the npm package. (We still have 2fa for getting credentials.)

decentralion:
I think we should just have a script, `release.sh`, which updates all the relevant files, creates a git tag, publishes to npm. Then after running that, the tag already exists, and the maintainer can write the release notes on GitHub. It's a two step process, but has little room for error (e.g. since the Git tag is created in release.sh, we don't have to worry about desync between the GitHub release and the npm release, even if someone merges during the process).

Also: the script should nag you to QA the release, maybe linking to a QA process checklist.

PR #2258, opened 2020-09-14
Simplify CLI version command to use version from package.json

hammadj:
This will make it easier to keep the package version and the CLI version in sync by directly using the version in package.json instead of maintaining a separate versioning system.

I also went pretty aggro in removing some of the extra formatting and git-state handling logic in the file since I didn't see it being used anywhere and @decentralion mentioned that it wasn't as necessary now that we have the NPM package as the main entrypoint for users. @wchargin if you think we should still keep that stuff around let me know!

Related: #2257

Test Plan: Run `sourcecred --version` and ensure it returns "v0.7.0-beta-16", or whatever is currently
in the package.json


src/cli/sourcecred.test.js
```diff
@@ -28,7 +28,7 @@ describe("cli/sourcecred", () => {
   it("responds to '--version'", async () => {
     expect(await run(sourcecred, ["--version"])).toEqual({
       exitCode: 0,
-      stdout: [expect.stringMatching(/^sourcecred v\d+\.\d+\.\d+$/)],
+      stdout: [expect.stringMatching(/^sourcecred v\d+\.\d+\.[a-z\d-]+$/)],
```
hammadj, line 31: Added this to allow for versions such as `0.7.0-beta-16`

decentralion (commented):
I think as written, it would be impossible to depend on the version in the frontend. However, showing version info in the frontend would be really nice. I've made a PR to do this: https://github.com/sourcecred/sourcecred/pull/2259. Can you try re-basing this on top of that PR, and see if it still works?

hammadj:
> I think as written, it would be impossible to depend on the version in the frontend. However, showing version info in the frontend would be really nice. I've made a PR to do this: #2259. Can you try re-basing this on top of that PR, and see if it still works?

Yea, I was gonna implement that in a separate PR by injecting the version into the frontend build. This way might work too, but it would result in the package.json showing up in client side code, which is not ideal

hammadj:
@decentralion Verified it works: 
image

decentralion (approved):
Lets give @wchargin a chance to look before merge

src/core/version.js
```diff
@@ -1,85 +1,5 @@
...
-  +gitState: GitState,
-  +environment: Environment,
-|};
-export type GitState = {|
-  +commitHash: string,
-  +commitTimestamp: string, // YYYYmmdd-HHMM, in commit-local time
-  +dirty: boolean, // does the worktree have unstaged/uncommitted changes?
-|};
-export type Environment = "development" | "production" | "test";
+const packageJSON = require("../../package.json");
```
wchargin, line 3: I’m curious—as you mentioned, this might pull `package.json` into the
frontend, which isn’t ideal (though not end of world); have you checked
whether it actually does, and if so whether destructuring here lets
Webpack tree-shake (`const {version: packageVersion} = require(…)`)?
I’m not familiar enough with Webpack to guess whether this will help.

wchargin, line 3: nit: `packageJson`, please; we use consistent camel-case except where
there’s a strong API reason not to (e.g., `toJSON` integration point).
Cf. `JsonLog`.

wchargin (approved):
I’m okay with this. I do think that it can be really helpful for the
frontend to report the exact Git version at which it was built, and we
used to display that, but we don’t anymore. I’m happy to add it back in
later with an easier mechanism: e.g., setting it in an environment
variable at build time.

PR #2259, opened 2020-09-14
Log SourceCred version on load

decentralion:
This commit prints the SourceCred version into the console on frontend
load. This will be useful to help provide tech support for users, since
we'll be able to find what version of SourceCred they'll be using.

Test plan: Run `yarn start`, and check the console.


hammadj (commented):
Verified this works correctly with the new versioning setup in #2258

hammadj (approved):

PR #2262, opened 2020-09-15
Refactor Explorer.js file before splitting into components

befitsandpiper:
An involved refactor of the Explorer.js file, and removing references to "this" from the render functions.
This may look a little ugly right now, but will allow a follow up PR to pull render functions out into separate components. Before RenderConfigureRow and RenderFilterSelect were tightly coupled to the Explorer class notation, whereas now they receive all of their dependencies as props and can easily be pulled into their own files and be refactored further from here.

Once these components are pulled out, the Explorer.js file will also be able to be converted into a functional component. There are probably at least two more PRs broadly related to refactoring this file: one which will pull these render functions into separate files, and one which will refactor all 3 files into functional components and reorganize state to minimize the number of props being passed down.


decentralion (commented):
Can I see what the code will look like after this sequence of refactors is complete, and the code has been split into smaller functional components?

One thing I'm wondering is: how much do we get out of this sequence of refactors? Do we imagine re-using the configuration row component elsewhere in the codebase? My sense is we might want to just replace this component with new, better designed and more narrowly scoped views.

befitsandpiper:
Maybe you're right. I guess it's not that useful necessarily unless we plan continuing to use these sub components or a modified version of them. If that is not the plan, then I should stop this process. But if that is the plan, this makes the code easier to read and to maintain.

befitsandpiper:
But, yeah, to answer your question: this file would be much shorter, and likely have less state included in it. The sub components would instead have some of that state that is directly related to their functions and also be pulled out into separate files. Probably at least half of the work has already been done here. 

One of these planned PRs would be just pulling out the functions into separate files and no other changes to avoid unusable diffs. The 2nd PR would be more involved but would mostly just be a reorganization of state and refactoring the main file into hooks.

PR #2264, opened 2020-09-15
Mui explore weights sliders

befitsandpiper:
Change the WeightSlider file to use all MUI Components instead of raw html. The MUI Slider has a better color scheme for the dark theme of our weights page. These changes broke the Slider test file, so I updated several of the tests. 

I did remove the onChange input tests, because changing the internal input element does not trigger the onChange event of the wrapping MUI Slider. I received confirmation from @decentralion that this is acceptable.

Screenshot from 2020-09-15 16-42-27


decentralion (approved):
I'm approving the test removal b.c. candidly I think the FE tests have demonstrated little benefit relative to the cost of writing and maintaining them. 

I did perk up seeing the test that verifies that the weight can be a non-power-of-two removed, so I manually tested that and verified it still works as intended.

Also tested the frontend and verified the sliders all still work (and look better).

PR #2266, opened 2020-09-17
Implement TaskManager

topocount:
TaskManager organizes and maintains a hierarchy of active tasks

It utilizes TaskReporters internally for logging, and allows them to
maintain their own task state for internal use. The primary concern of
the manager is creating and enforcing task scopes via a tree structure.
When a plugin task needs to be terminated in the load CLI command, it is important
that all child tasks also are terminated in case a restart of the task is
necessary. A well-behaved plugin should terminate all tasks it spawns on its own,
but in the case of a cache error or other unexpected error, this class makes it
possible to quickly find and terminate all child tasks.

paired with: @decentralion 
paired with: @wchargin 

test plan: unit tests are provided

### DIFF
--- a/src/util/taskManager.js
+++ b/src/util/taskManager.js
@@ -7,8 +7,10 @@ import {
   SilentTaskReporter,
 } from "./taskReporter";
 
+type TaskNodeId = string;
+type RootedTaskId = TaskNodeId[];
 type TaskNode = {|
-  id: TaskId,
+  reporter: TaskReporter,
   children: Map<TaskId, TaskNode>,
 |};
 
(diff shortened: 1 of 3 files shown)


decentralion (changes requested):
I find this API confusing. If I'm understanding correctly, the TaskManager is essentially intended to be a drop-in replacement for a TaskReporter, with the added feature that it allows you to create and close scopes. But this isn't clear from the signature (should `TaskManager` implement `TaskReporter`?), and the API leaks a lot of implementation details. For example the `TaskNode` type seems like an implementation detail that clients should not be aware of, but you can't construct a `TaskManager` without providing one.

Based on reading the module docstring, it's not clear that the `TaskManager` should be a util class at all. Util is intended for general-purpose tools, but the documentation for TaskManager suggests that it's narrowly intended for solving problems that come up with plugin loading. Should this just be an unexposed helper inside the load api? I suspect that this actually does solve a more general-purpose function, but atm this doesn't have the clarity or generality that I'd expect from a util.

Here's a suggestion: start by writing and getting feedback on an interface (not implementation). That should help clarify exactly what problem this is solving, and what API is needed.

I suspect that rather than writing a new class, we should just change the semantics of the `TaskReporter` to include a natural concept of scoping. What if instead of having `type TaskId = string`, we make a `TaskAddress` type, re-using the address module. Then, when calling `finish(address)`, that automatically finishes every task beneath it in the address structure.

We wouldn't even have to write any new tree logic: if we are using the [address module](https://github.com/sourcecred/sourcecred/blob/master/src/core/address.js) to create the TaskAddresses, that means we can also set up a TaskAddressTrie using the existing [trie module](https://github.com/sourcecred/sourcecred/blob/master/src/core/trie.js).

wchargin (commented):
@topocount and I discussed this some today. Some notes follow.

It looks like the interface here still relies to some degree on the
colon-hierarchy of string task names. I do see now (which I didn’t
understand correctly in our conversation) that clients don’t have to
provide the full ancestor-scoped task name, just the local part—which is
good. But it would be nice to eliminate that altogether from the control
flow, using the colons only for display.

The usage pattern for scopes also seems strange to me:

```javascript
const root: TaskManager = new TaskManager();
root.start("load-github");
const githubManager = root.scope("load-github");
// ... tasks with `githubManager`
root.finish("load-github")
```

It sounds like the `root.start` call is mandatory and the `root.scope`
call must be for the same task name. Could it perhaps make more sense to
just always return a scoped task manager from `start`, which clients may
discard if they don’t care about it? Thus:

```javascript
const root: TaskManager = new TaskManager();
const githubManager: TaskManager = root.start("load-github");
githubManager.start("restore-cache");
githubManager.finish("restore-cache");
root.finish("load-github");
```

Then there’s one fewer concept and the API surface area is reduced.

I note that `TaskManager` is implicitly being used as a `TaskReporter`
when you call `plugin.load(dirContext, scopedManager)` in the `load` CLI
command. I think that you should feel free to modify `TaskReporter` as
needed, both at the type level (e.g., `TaskReporter.start` may want to
return `void`, not `this`) and at the conceptual level.

So, as a client, I think that I might like an interface something like:

```javascript
interface TaskManager {
  // `start` and `finish` take task "base names": i.e., just the last
  // component of the colon-separated name. `start` returns a scoped
  // task manager in which subtasks can be defined. `finish` must be
  // called with a task previously started and not yet finished.
  start(task: string): TaskManager;
  finish(task: string);
}
```

But I also agree that I want to see clarified how this overlaps with
`TaskReporter`. There are a few things that we care about: logging that
tasks have been started/finished, measuring runtime, cleaning up
subtrees. Who handles what?

(FWIW, I think it’s at least worth considering the division that a
`TaskReporter`-like interface should _only_ handle the logging. The
three implementations there of CLI, in-memory, and silent task reporters
make sense to me.)

To @decentralion’s comment about using a trie: I agree that what we want
here is conceptually tree-like and/or trie-like, but I don’t see why the
`core/trie.js` module is what we want. That kind of trie is nice if you
want to be able to create/manipulate tasks keyed by an arbitrary path,
but I think we will only ever want to manipulate tasks that are the
direct children of a given task reporter node. That is, it should _not_
be valid to write

```javascript
const root = new TaskManager();
const load = root.start("load");
const github = load.start("github");
github.start("restore-cache");
root.finish("load: github: restore-cache");
// ^ invalid: the only child of `root` is "load", not "load: ..."
```

implicitly based on the colon-hierarchy. Since that’s invalid, the
peculiarities of an address-keyed `Trie` (like, the “get all nodes on
path to leaf” query) don’t seem particularly compelling, and it’s nice
to avoid using addresses when possible.

topocount:
@wchargin and @decentralion thanks for the thorough reveiws! I've refactored the TaskManager to no longer utilize colons for any kind of splitting at all. I've simplified the functionality (since the multilevel tree traveral isn't necessary) to eliminate a lot of recursive complexity and nested typing.

I've also followed William's suggested API interface with scoped TaskManagers returned from the `start` method. Happy to add a "paired with" line and/or "co-authored" trailer if this sticks since you helped me to pare down the API.

wchargin:
FYI: a “request changes” review means “this PR shouldn’t be merged until
*I* specifically re-approve”. Dismissing a stale review is an admin
action that should only be used when someone’s on vacation for a week
and their review is blocking your otherwise-approved merge. If you
believe that you’ve addressed the comments, the right response is to ask
the reviewer to please take another look.

(Dismissing a review also retroactively changes that review from
“request changes” to “commented”, which is another reason to avoid using
it.)

decentralion:
@wchargin I like that @topocount dismissed the review, so it will show back up in my "reviews requested" queue. Otherwise I might lose it because my GH notifications are too noisy.

topocount:
@wchargin my intention was to ensure @decentralion saw that this needed attention again. I'm pretty sure I can re-request a review without dismissing a changes-requested review so I'll avoid dismissing reviews from now on.

wchargin:
Yes, the correct thing to do is to request re-review.

src/util/taskManager.js
```diff
@@ -0,0 +1,92 @@
...
+  type TaskId,
+  type TaskReporter,
+  ScopedTaskReporter,
+  SilentTaskReporter,
+} from "./taskReporter";
+
+export type TaskNode = {|
+  reporter: TaskReporter,
+  children: Map<TaskId, TaskNode>,
+|};
```
decentralion, line 13: Why is this type exported?

If it's just because the private methods of TaskManager export it, maybe we can name it `_TaskNode` or such to communicate that it's an implementation detail?

src/util/taskManager.test.js
```diff
@@ -0,0 +1,116 @@
...
+    fooScope.finish("test");
+    expect(testCase.reporter.entries()).toEqual([
+      {"taskId": "foo", "type": "START"},
+      {"taskId": "foo: test", "type": "START"},
+      {"taskId": "foo: test", "type": "FINISH"},
+    ]);
+  });
+
+  describe("Scoped TaskManager", () => {
+    it("cannot create scope on nonexistent task", () => {});
```
decentralion, line 71: empty test

src/util/taskManager.test.js
```diff
@@ -0,0 +1,116 @@
...
+      this.taskManager = new TaskManager(this.reporter);
+    }
+    start(task: string): TaskManager {
+      return this.taskManager.start(task);
+    }
+    finish(task: string) {
+      this.taskManager.finish(task);
+      return this;
+    }
+  }
```
decentralion, line 22: personally, i would probably have just written something like

```js
function example() {
  const reporter = new SilentTaskReporter();
  const manager = new TaskManager(reporter);
  return {manager, reporter};
```

Your way is fine and maybe better, no change requested.

src/util/taskManager.test.js
```diff
@@ -0,0 +1,116 @@
...
+      {"taskId": "foo", "type": "START"},
+    ]);
+    fooScope.start("test");
+    fooScope.finish("test");
+    expect(testCase.reporter.entries()).toEqual([
+      {"taskId": "foo", "type": "START"},
+      {"taskId": "foo: test", "type": "START"},
+      {"taskId": "foo: test", "type": "FINISH"},
+    ]);
+  });
```
decentralion, line 68: just to be 100% sure the recursion is working, it would be nice to have a double-nested task and check that finishing root finishes the nested task.

decentralion (commented):
Given that a `TaskManager` is basically a wrapper around a `TaskNode`, have you considered that instead of having the task node at all, you could directly put the the reporter as a property of the `TaskManager`, and have also give the property `children: Map<TaskId, TaskManager>`.

I believe this would simplify the implementation materially because the `_findTask`, `_createTask`, and `_finishTask` would no longer need a second argument to specify which task node is in scope. 

Then:

`new TaskManager(reporter)` gives a new empty task manager
`manager.start(id)` gives a new manager with the new id
`manager.finish()` recursively calls `child.finish()` for all children, and then finish it's root task

WDYT?

src/util/taskManager.test.js
topocount: @decentralion that is tested here I believe:
https://github.com/sourcecred/sourcecred/blob/71c1634861758b49fea97bcfc4fa4542978aaf4a/src/util/taskManager.test.js#L40-L54

src/util/taskManager.js
```diff
@@ -0,0 +1,78 @@
...
+    parent._reporter.finish(idToKill);
+  }
+
+  _finishChildren(task: TaskManager) {
+    const children = Array.from(task._children.keys());
+    for (const id of children) {
+      this._finishTask(id, task);
+    }
+  }
+}
```
decentralion, line 78: Why does `_finishTask` need the `parent` argument? Why does `_finishChildren` need the `task` argument?

it seems like the following would work:

```js
_finishTask(idToKill: TaskId) {
  const taskToKill = this._children.get(idToKill);
  if (!taskToKill) // etc
  taskToKill._finishChildren();
}

_finishChildren() {
  for (const id of this._children.keys()) {
    this.finishTask(id);
  }
}
```

decentralion (approved):
LGTM

PR #2273, opened 2020-09-19
Add no-load flag to go CLI command

topocount:
Now the load step can be skipped if desired, without using lower-level
commands.

test plan: `path/to/local/build/of/sourcecred.js go` to ensure no regressions,  
`... go --no-load`, and
`path/to/local/build/of/sourcecred.js go --bad-flag` to ensure extraneous
arguments cause the command to throw

### DIFF
--- a/src/cli/go.js
+++ b/src/cli/go.js
@@ -11,22 +11,24 @@ function die(std, message) {
   return 1;
 }
 
-const noLoadArg = "--no-load";
+const NO_LOAD_ARG = "--no-load";
 
 const goCommand: Command = async (args, std) => {
+  let noLoad = false;
+  if (args.length === 1 && args[0] === NO_LOAD_ARG) {
+    noLoad = true;
+  } else if (args.length !== 0) {
+    return die(std, "usage: sourcecred go [--no-load]");
+  }
+
   const commandSequence = [
     {name: "load", command: load},
     {name: "graph", command: graph},
     {name: "score", command: score},
   ];
 
-  if (args.length === 1 && args[0] === noLoadArg) {
-    commandSequence.shift();
-  } else if (args.length !== 0) {
-    return die(std, "usage: sourcecred go [--no-load]");
-  }
-
   for (const {name, command} of commandSequence) {
+    if (name === "load" && noLoad) continue;
     const ret = await command([], std);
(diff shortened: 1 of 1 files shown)


src/cli/go.js
```diff
@@ -11,15 +11,21 @@ function die(std, message) {
   return 1;
 }
 
+const noLoadArg = "--no-load";
```
decentralion, line 14: we generally CAPS_CASE args like this

src/cli/go.js
```diff
@@ -11,15 +11,21 @@ function die(std, message) {
...
-    return die(std, "usage: sourcecred go");
-  }
   const commandSequence = [
     {name: "load", command: load},
     {name: "graph", command: graph},
     {name: "score", command: score},
   ];
+
+  if (args.length === 1 && args[0] === noLoadArg) {
+    commandSequence.shift();
```
decentralion, line 24: IMO this is a little bit cute and would generally prefer something more explicit, but this command is so trivial that I think its fine.

decentralion (approved):

src/cli/go.js
```diff
@@ -11,22 +11,24 @@ function die(std, message) {
...
   ];
 
-  if (args.length === 1 && args[0] === noLoadArg) {
-    commandSequence.shift();
-  } else if (args.length !== 0) {
-    return die(std, "usage: sourcecred go [--no-load]");
-  }
-
   for (const {name, command} of commandSequence) {
+    if (name === "load" && noLoad) continue;
```
wchargin, line 31: I like this better than the previous version, because it’s more explicit
about the branching. IMHO even clearer would be something like:

```javascript
const commandSequence = [];
if (!noLoad) {
  commandSequence.push({name: "load", command: load});
}
commandSequence.push({name: "graph", command: graph});
commandSequence.push({name: "score", command: score});
```

because `commandSequence` actually corresponds to the sequence of
commands that will be executed.

(Also, it would probably be clearer to call the variable `shouldLoad`
rather than `noLoad`, since `if (shouldLoad)` is slightly easier to
understand than `if (!noLoad)`.)

wchargin (commented):
Post-hoc approval, with suggestion for improvement.

PR #2278, opened 2020-09-21
Utilize taskManager

topocount:
test plan: on an example instance:
1. sourcecred load
2 sqlite3 ./cache/sourcecred/github/*.db "UPDATE meta SET config = 'bad'"
3. sourcecred load
4. ensure nothing breaks and the reload is successful


topocount:
this is technically blocked on #2266

decentralion:
@topocount is this PR obsolete?

topocount:
this isn't deprecated, I just need to do a bit more refactoring to feel good about merging it

src/plugins/experimental-discord/mirror.js
```diff
@@ -23,21 +23,21 @@ export class Mirror {
     this.guild = guild;
   }
 
-  async update(reporter: TaskReporter) {
+  async update(manager: TaskManager) {
```
decentralion, line 26: nit: `reporter` was a fairly self-descriptive variable name (it reports things), `manager` less so because it could be managing anything.

maybe we should call it `reporterManager`? or maybe the `TaskManager` class should itself be called `ScopedTaskReporter`?

src/cli/load.js
```diff
@@ -51,50 +52,30 @@ const loadCommand: Command = async (args, std) => {
...
-      warn(
-        std,
-        task,
-        `Error updating cache. clearing cache and restarting.
-        This is the error from the plugin:\n
-        ${error}`
-      );
-      taskReporter.start(task);
+      manager.finish(task);
+      warn(std, task, `clearing cache: ${error}`);
```
decentralion, line 71: imo including "Error [while] updating cache. clearing cache and restarting" was much more descriptive than the new message.

decentralion (commented):
the PR description doesn't orient reviewers on what TaskReporters are, or TaskManagers, or what's changing. however, you've requested review from a number of folks who haven't touched that part of the codebase before. i think it would be quite helpful to re-write the description to provide more context to reviewers to understand why the change is justified. also, the test plan of "do a thing and verify nothing bad happens" is underspecified--what behavior am I expecting to see, and to not see? 

since this change is basically about console output, i'd like to see, as part of the PR description, a before/after of what the console output is from the test plan before and after this change.

hammadj (commented):
Not really clear on what this PR is trying to accomplish / improve, don't have a ton of opinions / expertise on this area of the codebase either, will defer to others. Reading through it there's no issues I have, seems like a fairly innocent refactor

PR #2279, opened 2020-09-22
credrank: create epoch accumulator nodes

wchargin:
Summary:
Previously, user-epoch nodes (indexed by a user and an epoch) sent
payout flow directly to user (scoring) nodes. As of this change,
user-epoch nodes now send payout flow to new _epoch accumulator_ nodes,
which are indexed by an epoch. Thus, the structure looks like:

```
contribution1 -> userepoch[@torvalds, "week 2020-01-01"]
    with probability according to the base edge weight
userepoch[@torvalds, "week 2020-01-01"] -> epoch["week 2020-01-01"]
    with probability \beta
epoch["week 2020-01-01"] -> seed
    with probability 1
```

Since all base edges incident to scoring nodes are rewritten to point to
user-epoch nodes, scoring nodes are now isolated and are thus removed
from the graph entirely. A user’s cred can be determined by summing the
transition probabilities across the payout edges. (This held previously
as well, though previously this value was also equal to the cred at the
scoring node, which no longer exists.)

Test Plan:
Run `sourcecred credrank` to generate `output/credGraph.json`, then use
the following `jq(1)` command to generate the top-10 cred distribution:

```
jq '
    . as $root | .scores as $scores | .mpg as $mpg
    | [
        $mpg[1].edges[]
        | select(.address | contains("EPOCH_PAYOUT"))
        | {"edge": ., "who": (.src | split("\u0000")[5:] | join("\u0000"))}
    ]
    | group_by(.who)
    | map({
        who: .[0].who,
        edges: [
            .[].edge
            | {src, pr: .transitionProbability}
            | {flow: (.pr * $scores[.src])}
        ],
    })
    | map({
        who: (.who | split("\u0000") | join("/")),
        cred: ([.edges[].flow] | add),
    })
    | sort_by(-.cred)
    | .[:10]
    | map({(.who): .cred})
    | add
'
```

Note that the numbers are _approximately_ the same before and after this
change, though there definitely is some difference (usually on the order
of about 0.1–1%). I’m not sure exactly where this is coming from.

wchargin-branch: credrank-epoch-accumulators

### DIFF
--- a/src/core/markovProcessGraph.js
+++ b/src/core/markovProcessGraph.js
@@ -138,13 +138,15 @@ const SEED_DESCRIPTION = "\u{1f331}"; // U+1F331 SEEDLING
 // Node address prefix for epoch nodes.
 const USER_EPOCH_PREFIX = NodeAddress.append(CORE_NODE_PREFIX, "USER_EPOCH");
 
-export type EpochNodeAddress = {|
+export type UserEpochNodeAddress = {|
   +type: "USER_EPOCH",
   +owner: NodeAddressT,
   +epochStart: TimestampMs,
 |};
 
-export function epochNodeAddressToRaw(addr: EpochNodeAddress): NodeAddressT {
+export function userEpochNodeAddressToRaw(
+  addr: UserEpochNodeAddress
+): NodeAddressT {
   return NodeAddress.append(
     USER_EPOCH_PREFIX,
     String(addr.epochStart),
@@ -152,7 +154,9 @@ export function epochNodeAddressToRaw(addr: EpochNodeAddress): NodeAddressT {
   );
 }
 
-export function epochNodeAddressFromRaw(addr: NodeAddressT): EpochNodeAddress {
+export function userEpochNodeAddressFromRaw(
+  addr: NodeAddressT
+): UserEpochNodeAddress {
   if (!NodeAddress.hasPrefix(addr, USER_EPOCH_PREFIX)) {
     throw new Error("Not an epoch node address: " + NodeAddress.toString(addr));
   }
@@ -372,7 +376,7 @@ export class MarkovProcessGraph {
         mint: 0,
       });
       for (const scoringAddress of _scoringAddresses) {
-        const thisEpoch = epochNodeAddressToRaw({
+        const thisEpoch = userEpochNodeAddressToRaw({
           type: "USER_EPOCH",
           owner: scoringAddress,
           epochStart: boundary,
@@ -394,7 +398,7 @@ export class MarkovProcessGraph {
           transitionProbability: fibration.beta,
         });
         if (lastBoundary != null) {
-          const lastEpoch = epochNodeAddressToRaw({
+          const lastEpoch = userEpochNodeAddressToRaw({
             type: "USER_EPOCH",
             owner: scoringAddress,
             epochStart: lastBoundary,
@@ -464,7 +468,7 @@ export class MarkovProcessGraph {
       const epochEndIndex = sortedIndex(timeBoundaries, edgeTimestampMs);
       const epochStartIndex = epochEndIndex - 1;
       const epochTimestampMs = timeBoundaries[epochStartIndex];
-      return epochNodeAddressToRaw({
+      return userEpochNodeAddressToRaw({
         type: "USER_EPOCH",
         owner: address,
         epochStart: epochTimestampMs,
--- a/src/core/markovProcessGraph.js
+++ b/src/core/markovProcessGraph.js
@@ -27,9 +27,11 @@
  * The Markov process graphs in this module have three kinds of nodes:
  *
  *   - *base nodes*, which are in 1-to-1 correspondence with the nodes
- *     in the underlying core graph (including users);
- *   - *epoch nodes*, which are created for each time period for each
- *     scoring node; and
+ *     in the underlying core graph that are not scoring nodes;
+ *   - *user-epoch nodes*, which are created for each time period for
+ *     each scoring node; and
+ *   - *epoch accumulators*, which are created once for each epoch to
+ *     aggregate over the epoch nodes,
  *   - the *seed node*, which reifies the teleportation vector and
  *     forces well-definedness and ergodicity of the Markov process (for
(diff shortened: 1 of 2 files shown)


src/core/markovProcessGraph.js
```diff
@@ -133,37 +136,69 @@ const SEED_ADDRESS = NodeAddress.append(CORE_NODE_PREFIX, "SEED");
...
   const owner = NodeAddress.fromParts(parts.slice(epochPrefixLength + 1));
   return {
-    type: "EPOCH_NODE",
+    type: "USER_EPOCH",
     owner,
     epochStart,
   };
 }
 
+// TODO(@wchargin): Expose more cleanly.
```
wchargin, line 170: I’ve TODOed this because I expect subsequent changes to churn here again.

src/core/markovProcessGraph.js
```diff
@@ -133,37 +136,69 @@ const SEED_ADDRESS = NodeAddress.append(CORE_NODE_PREFIX, "SEED");
 const SEED_DESCRIPTION = "\u{1f331}"; // U+1F331 SEEDLING
 
 // Node address prefix for epoch nodes.
-const EPOCH_PREFIX = NodeAddress.append(CORE_NODE_PREFIX, "EPOCH");
+const USER_EPOCH_PREFIX = NodeAddress.append(CORE_NODE_PREFIX, "USER_EPOCH");
 
 export type EpochNodeAddress = {|
```
decentralion, line 141: `UserEpochNodeAddress` for consistency?

src/core/markovProcessGraph.js
```diff
@@ -133,37 +136,69 @@ const SEED_ADDRESS = NodeAddress.append(CORE_NODE_PREFIX, "SEED");
...
-  +type: "EPOCH_NODE",
+  +type: "USER_EPOCH",
   +owner: NodeAddressT,
   +epochStart: TimestampMs,
 |};
 
 export function epochNodeAddressToRaw(addr: EpochNodeAddress): NodeAddressT {
```
decentralion, line 147: `userEpochNodeAddressToRaw` for consistency?

src/core/markovProcessGraph.js
```diff
@@ -172,7 +207,14 @@ const FIBRATION_EDGE = EdgeAddress.fromParts([
 ]);
 const EPOCH_PAYOUT = EdgeAddress.append(FIBRATION_EDGE, "EPOCH_PAYOUT");
 const EPOCH_WEBBING = EdgeAddress.append(FIBRATION_EDGE, "EPOCH_WEBBING");
-const EPOCH_RADIATION = EdgeAddress.append(FIBRATION_EDGE, "EPOCH_RADIATION");
+const USER_EPOCH_RADIATION = EdgeAddress.append(
+  FIBRATION_EDGE,
+  "USER_EPOCH_RADIATION"
+);
```
decentralion, line 213: Let's remove?

wchargin: If we remove this, then the out-edges from a user epoch node are:

  - a payout edge with mass β
  - webbing edges with mass γ₁ + γ₂
  - base edges (to/from other contributions) with remaining mass

But in the case where there are no base edges for a user-epoch (which
will be quite common), this means that 1 − β − (γ₁ + γ₂) of the mass is
unaccounted for. Where should it go—to the seed? Anywhere that you send
it seems to create strange incentives. So we’d like to hold off on this.

src/core/markovProcessGraph.js
wchargin: Sure: done.

src/core/markovProcessGraph.js
wchargin: Sure: done.

decentralion (approved):

wchargin:
Oh, I should mention: for the test plan to work, you’ll need a copy of
`jq(1)` with a fix for this issue:
<https://github.com/stedolan/jq/issues/1732>

As far as I can tell, no such versions have been officially released
yet, so you’ll need to build from source. It’s easy, though; the
instructions Just Work for me.

wchargin:
Okay, indeed this _does_ seem like a large difference from live (which
is what I had thought earlier; I confused myself into thinking that it
was okay). With identity contractions, the top-10 set is:

```json
{
  "decentralion": 14102.991539172672,
  "s-ben": 4425.241703927829,
  "kurafire": 4205.921601942201,
  "lbstrobbe": 4108.230399526834,
  "mzargham": 3805.5115568014694,
  "beanow": 2561.935728690981,
  "wchargin": 1708.3907410921875,
  "bex": 1237.7415470456422,
  "hammad": 845.8524205905647,
  "burrrata": 803.4439951707769
}
```

but this is substantially different in both ratios (e.g., @decentralion
having over 3× next-highest) and ordering compared to the live explorer.

Before this patch, the top-20 (since no contractions) set was:

```
{
  "github/decentralion": 5110.167218462159,
  "discord/decentralion#1337": 5066.912642226158,
  "discourse/decentralion": 3800.7751362414374,
  "github/wchargin": 3153.45178989316,
  "discord/LBS#6541": 2849.5026559624857,
  "discord/s_ben#7872": 2423.183373772889,
  "discourse/LB": 1770.2109304375724,
  "discourse/s_ben": 1769.8258395605437,
  "discord/Beanow#5887": 1685.3385667598184,
  "discord/hammadj#6801": 1407.8391183846709,
  "github/hammadj": 1346.0545919176654,
  "github/Beanow": 1241.4698309035766,
  "discourse/burrrata": 1068.5817980147087,
  "discord/mZ#3472": 943.040521344988,
  "discord/Faruk#9564": 773.7940569542584,
  "discourse/Beanow": 753.1215564458081,
  "discord/Bex#7189": 690.8774085956492,
  "discord/burrrata#0232": 669.3924188507306,
  "discord/wchargin#8658": 462.15849737070437,
  "discourse/wchargin": 457.2764189978515
}
```

…which at a glance looks roughly consistent with the contracted top-10
above. So perhaps this patch is correct but the results are just very
different compared to our live instance. We should look into that—last
time we checked, they seemed to be comparable.

PR #2283, opened 2020-09-23
Feature 2172/ledger admin QOL

### LINKED ISSUE #2172: Improve the "Ledger Admin" page
As discussed on a call with @befitsandpiper:

- Currently UI is hard to use and navigate, confusing that the update pane is at the bottom after all identities
- Suggest putting identities into a fixed-size scrollable pane so update pane is always visible
- Then add a search to filter down identities

youngkidwarrior:
closes #2172 
static height for identity list
 
  - I made the decision for height to be static so the identity manager doesn't jump up and down

Autocomplete input updates identity list

Testing N/A

### DIFF
--- a/src/ui/components/LedgerAdmin.js
+++ b/src/ui/components/LedgerAdmin.js
@@ -16,6 +16,7 @@ import {
   TextField,
 } from "@material-ui/core";
 import {useLedger} from "../utils/LedgerContext";
+import {useAutocomplete} from "@material-ui/lab";
 
 const useStyles = makeStyles((theme) => {
   const marginNum = 20;
@@ -35,7 +36,7 @@ const useStyles = makeStyles((theme) => {
     },
     centerRow: {
       display: "flex",
-      justifyContent: "center",
+      justifyContent: "space-around",
       alignItems: "center",
     },
     element: {flex: 1, margin: `${marginNum}px`},
@@ -47,6 +48,9 @@ const useStyles = makeStyles((theme) => {
     checkboxElement: {flexGrow: 1, flexBasis: 0, margin: `${marginNum}px`},
     promptStringHeader: {margin: `${marginNum}px`, marginBottom: 0},
     IdentitiesHeader: {margin: `${marginNum}px`},
+    input: {
+      margin: `${marginNum}px`,
+    },
   };
 });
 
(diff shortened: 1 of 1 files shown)


src/ui/components/LedgerAdmin.js
```diff
@@ -98,35 +114,58 @@ export const LedgerAdmin = () => {
...
-
   return (
     <Container className={classes.root}>
       <span className={classes.centerRow}>
-        <h1 className={classes.IdentitiesHeader}>Identities</h1>{" "}
-        {ledger.accounts().length > 0 && <h3> (click one to update it)</h3>}
+        <h1 className={classes.IdentitiesHeader}>Identities</h1>
+        <TextField
+          className={classes.input}
+          label="Search"
```
topocount, line 159: If we leave this where it is I'm inclined to call this a filter box since it doesn't really return anything. it just filters the list. 
```suggestion
          label="Filter"
```

topocount (commented):
I think it might make sense to dynamically collapse the container when an identity. It'd probably be best to utilize a combo box here instead, and to just copy and reintegrate the identity merging combo box below. We can also utilize that box's faster filter function. The filter box is quite slow with larger repos.
Also, It looks like this container doesn't really "contain" instances with a larger number of contributors at this point:

image

youngkidwarrior:
@topocount Per our conversation, here is a rough layout

Note, the update ledger tab needs to be changed to reflect the new layout, e.g. "Select a user to update"
Screen Shot 2020-09-26 at 1 52 42 PM

befitsandpiper:
I don't know if this is out of scope for the changes we are trying to make here, but can the active checkboxes be moved into the actual table? This will allow much more rapid activation/deactivation of many accounts at once, and also better visibility on who is active by just scrolling through the table.

youngkidwarrior:
@befitsandpiper great idea. I think a seperate PR though. Just to keep this one about the search bar and the layout around it

topocount (commented):
I've started a discussion here to get some consensus on the proposed design: https://discordapp.com/channels/453243919774253079/454007880111095808/760153863457341450

decentralion:
This has been blocked for more than a month, so I'm going to close it to keep our open PRs a bit cleaner. Suggest syncing up with @KuraFire on Discord to coordinate more changes to this UI.

PR #2291, opened 2020-09-26
Set default period if it doesn't exist in dependenciesConfig

hammadj:
This fixes the issue of having to keep the "startTime" value up to date in the template
instance by setting it automatically to one week from the time the dependencyId was first
created in the ledger (i.e. when the user first added the dependency). The default weight
is set to 0.05, or 5% of the total cred.

Test Plan: Create a dependencyConfig that has an empty array for periods, run sourcecred
commands to generate the instance, and ensure the dependencyConfig.json is updated with a
period startTime of one week from the current time.

### DIFF
--- a/src/api/dependenciesConfig.js
+++ b/src/api/dependenciesConfig.js
@@ -88,6 +88,12 @@ export type DependencyConfig = {|
   // in new instances.)
   // Defaults to false if unset.
   +autoActivateOnIdentityCreation?: boolean,
+
+  // If this is set, then a default time period will be injected in the
+  // dependency config with the weight set to this value (e.g. 0.05 = 5% of total cred).
+  // (Mostly included so we can have SourceCred receiving cred by default)
+  // Does not inject a starting period if unset.
+  +autoInjectStartingPeriodWeight?: number,
 |};
 
 export type MintPeriod = {|
@@ -95,12 +101,16 @@ export type MintPeriod = {|
   +weight: number,
 |};
 
-const ONE_WEEK_IN_MS = 7 * 24 * 60 * 60 * 1000;
-
 export const mintPeriodParser: C.Parser<MintPeriod> = C.object({
   startTime: timestampISOParser,
   weight: C.number,
 });
+
+function checkWeightValid(x: number): number {
+  if (x > 1 || x < 0)
+    throw new Error(`must be a number between 0 and 1, got ${x}`);
+  return x;
+}
 
 export const dependencyConfigParser: C.Parser<DependencyConfig> = C.object(
   {
@@ -110,6 +120,7 @@ export const dependencyConfigParser: C.Parser<DependencyConfig> = C.object(
   {
     id: uuidParser,
     autoActivateOnIdentityCreation: C.boolean,
+    autoInjectStartingPeriodWeight: C.fmap(C.number, checkWeightValid),
   }
 );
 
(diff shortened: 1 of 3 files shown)


src/api/dependenciesConfig.js
```diff
@@ -133,6 +136,16 @@ export function ensureIdentityExists(
       if (dep.autoActivateOnIdentityCreation) {
         ledger.activate(id);
       }
+      if (!dep.periods.length) {
+        // Set default start period to one week from adding the dependency
+        return {
+          ...dep,
+          id,
+          periods: [
+            {"startTime": toISO(Date.now() + ONE_WEEK_IN_MS), "weight": 0.05},
```
decentralion, line 145: maybe just use Date.now? more robust to the change where we make the period 1 day instead of 1 week.
(neutral, could go either way)

decentralion (changes requested):
I like the idea, but not the interface. Having a period with 5% weight get auto-injected whenever there are no periods is way too magical. Nothing in the interface would cause someone to expect this to happen. Also, what if they want to auto-inject, but with a weight of 3% or 10%?

Instead, I propose making a separate optional key in the config, something like `autoInjectStartingPeriodWeight`. If that key is present, and the periods are currently empty, then it will auto-inject a starting period with the specified weight. way less magical

src/api/dependenciesConfig.js
```diff
@@ -87,6 +88,12 @@ export type DependencyConfig = {|
   // in new instances.)
   // Defaults to false if unset.
   +autoActivateOnIdentityCreation?: boolean,
+
+  // If this is set, then a default time period will be injected in the
+  // dependency config with the weight set to this value (e.g. 0.05 = 5% of total cred).
+  // (Mostly included so we can have SourceCred receiving cred by default)
```
decentralion, line 94: This isn't precisely correct. a weight of 0.05 mints an additional 5% of base contributor cred, which means that it will be 0.05 / 1.05 of total cred, or about 4.8%

src/api/dependenciesConfig.js
```diff
@@ -99,6 +106,12 @@ export const mintPeriodParser: C.Parser<MintPeriod> = C.object({
   weight: C.number,
 });
 
+function checkWeightValid(x: number): number {
+  if (x > 1 || x < 0)
+    throw new Error(`must be a number between 0 and 1, got ${x}`);
```
decentralion, line 111: actually, having a weight greater than 1 is technically fine. if there's a single dep with weight 1, it gets half the total cred. if the weight were 2, it would get 2/3rds of the total cred.

i suggest just checking that the weight is a nonnegative number

src/api/dependenciesConfig.js
```diff
@@ -133,6 +147,20 @@ export function ensureIdentityExists(
       if (dep.autoActivateOnIdentityCreation) {
         ledger.activate(id);
       }
+      const weight = dep.autoInjectStartingPeriodWeight;
+      if (weight != null && !dep.periods.length) {
+        // Set default start period to one week from adding the dependency
```
decentralion, line 152: comment out of date. (i prefer this implementation which does as-of Date.now()).

src/api/dependenciesConfig.test.js
```diff
@@ -46,15 +48,57 @@ describe("api/dependenciesConfig", () => {
...
+        name: n("foo"),
+        periods: [],
+        autoInjectStartingPeriodWeight,
+      };
+      const {periods} = ensureIdentityExists(config, ledger);
+      expect(periods).toHaveLength(1);
+      expect(periods[0].weight).toBe(autoInjectStartingPeriodWeight);
+      expect(fromISO(periods[0].startTime)).toBe(fakeNow);
+    });
+    it("does not inject default period if it already exists", () => {
```
decentralion, line 73: nit: `s/if it already exists/if any period already exists/`. (this test doesn't actually check the case where "it" (the default period) already exists, which is good, it tests the more general case where some other period exists.)

decentralion (approved):
approved modulo nits

PR #2299, opened 2020-09-28
Allow Grain table height to conform to the size of the window

topocount:
Using "vh" units allows element to adjust its size relative to the space
it's provided by the viewport.

test plan: 1. yarn start --instance <path to instance with many
participants>
2. browse to localhost:8080/#/accounts and resize the window a few times
to ensure a second outer scroll bar doesn't appear

### DIFF
--- a/src/ui/components/AccountOverview.js
+++ b/src/ui/components/AccountOverview.js
@@ -19,7 +19,7 @@ type OverviewProps = {|+currency: CurrencyDetails|};
 const useStyles = makeStyles(() => {
   return {
     container: {
-      maxHeight: "40em",
+      maxHeight: "80vh",
     },
   };
 });


topocount:
@decentralion I wasn't able to replicate the issue you showed. I've tried in chromium and Firefox.

decentralion (commented):
If I shrink my window down, I get a scrollbar-in-scrollbar effect. Any idea why?

decentralion (approved):
I'm approving b.c. I think the new issue (scroll-within-scroll when height is small) is less annoying than the existing issue. But I'm curious why we're seeing scroll-within-scroll.

topocount:
> If I shrink my window down, I get a scrollbar-in-scrollbar effect. Any idea why?

@decentralion It's because the react admin Layout component that contains the main content (in this case the table containing grain balances) has fixed-pixel padding: 
image

I looked into changing this when investigating the scrollbar regression with the border (fixed by #2215) and we'd have to start decomposing the the Admin component, which is probably worth pursuing when we want to customize our frontend beyond what react admin can offer us. 

In the meantime I've reduced the maxHeight to `80vh` to further reduce the likelihood of the scrollbar appearing during intended use.

PR #2301, opened 2020-09-29
Disable github publish workflow for now

topocount:
test plan: using the publish UI on github should not trigger the npm
publish workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,8 +1,9 @@
 name: Publish Release
 on:
   # Publish when a new release is created on GitHub
-  release:
-    types: [published]
+  # disable for now in favor of authenticated CLI workflow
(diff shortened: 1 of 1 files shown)


decentralion (approved):

wchargin (commented):
I don’t have context here. Deferring.

wchargin (approved):
Upon further estimation, I think that this is a good idea.

wchargin:
This appears to break the action: it’s failing at every commit with

> **Error**
>
> No event triggers defined in `on`

Sample failure:
<https://github.com/sourcecred/sourcecred/actions/runs/284710170>

Please fix?

(imho, this patch _should_ be fine; this is another example of a system
that fails to treat zero as a number, leading to unexpected breakage.)

PR #2304, opened 2020-09-30
flow: fix and enable “types first” mode

wchargin:
Summary:
“Types First” is a new architecture for Flow that is now on by default
and will be mandatory by January 2021:
<https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb>

The change basically amounts to “module boundaries must be explicitly
typed rather than inferred”, which is generally a good thing for
readability anyway.

Currently, we’ve opted out of types-first mode. This patch fixes all the
errors (150-ish of them) and opts us in.

This is written with the help of a codemod (~400 lines generated by
`yarn flow codemod annotate-exports --write --repeat src/`) but with
many manual changes on top (~350 lines), amounting to:

  - migrating the `config/` subtree manually, using comment syntax
  - fixing the inferred type for `Schema.unfaithful`
  - fixing an incorrectly absolute import path
  - defining a typed interface for the `api` module
  - replacing `$IMPORTED_TYPE$` placeholders with direct resolutions
  - retargeting and opacifying `"React"`
  - rewriting imports from `"filename.js"` (rather than `"filename"`)
  - enabling types-first mode in `.flowconfig`

Test Plan:
That `yarn flow` passes is the main thing; that `yarn test` passes
primarily indicates that the codemod didn’t break any import paths. (Or,
rather, that I fixed all the import paths that it broke.)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -10,6 +10,5 @@ flow-typed
 
 [options]
 include_warnings=true
-types_first=false
 
 [strict]
--- a/config/env.js
+++ b/config/env.js
@@ -57,9 +57,29 @@ process.env.NODE_PATH = (process.env.NODE_PATH || "")
   .map((folder) => path.resolve(appDirectory, folder))
   .join(path.delimiter);
 
+/*::
+type Env = {|
+  // Environment like `process.env`: e.g., `raw["NODE_ENV"] === "development"`.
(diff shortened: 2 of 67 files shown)


topocount (approved):
Thanks for powering through this upgrade!

PR #2307, opened 2020-10-01
Make CredGraph smaller on disk

decentralion:
This modifies the MarkovProcessGraph and CredGraph JSON serialization
format so that edges reference src and dst by index into the node order,
rather than by full address. The same transformation is applied for node
scores.

This reduces space consumption by a factor of 2.45x, reducing
credGraph.json for sourcecred/cred from 277M to 113M. However, this is
still above GitHub's file size limit, and considerably larger than the
status quo from TimelineCred (34M). Clearly, move improvements of
similar magnitude are needed.

Ideas for further improvement:
- For "organic" edges, rather than storing two wholly separate MarkovEdges
on disk, we can store a single edge with two transition probabilties,
which may be "unpacked" into two MarkovEdges
- Come up with a more efficient scheme for enumerating / unpacking the
"inorganic" nodes and edges. E.g. there's really no need to store every
participant epoch node address on disk, provided that we have the list
of epochs boundaries and of participant epochs. Likewise with radiation
edges; their presence may be inferred.

Test plan: The CredRank command output indicates that scores are
unchanged.

### DIFF
--- a/src/core/credGraph.js
+++ b/src/core/credGraph.js
@@ -1,7 +1,6 @@
 // @flow
 
 import * as NullUtil from "../util/null";
-import * as MapUtil from "../util/map";
 import {type NodeAddressT, type EdgeAddressT} from "./graph";
 import {
   MarkovProcessGraph,
@@ -10,7 +9,9 @@ import {
   type TransitionProbability,
   markovEdgeAddress,
   payoutAddressForEpoch,
+  COMPAT_INFO as MPG_COMPAT_INFO,
 } from "./markovProcessGraph";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
 
 export type Node = {|
   +address: NodeAddressT,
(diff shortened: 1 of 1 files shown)


src/core/credGraph.js
```diff
@@ -35,10 +36,13 @@ export type Participant = {|
   +credPerEpoch: $ReadOnlyArray<number>,
 |};
 
-export type CredGraphJSON = {|
+export type CredGraphJSON = Compatible<{|
   +mpg: MarkovProcessGraphJSON,
-  +scores: {|+[NodeAddressT]: number|},
-|};
+  // scores for each node in node-sorted-order
```
wchargin, line 41: Please specify that the ordering is as in the Markov process graph.

src/core/markovProcessGraph.js
```diff
@@ -692,9 +705,25 @@ export class MarkovProcessGraph {
...
   toJSON(): MarkovProcessGraphJSON {
+    const nodes = Array.from(this._nodes.values());
+    const sortedNodes = sortBy(nodes, (n) => n.address);
+    const nodeIndex: Map<
+      NodeAddressT,
+      number /* index into nodeOrder */
+    > = new Map();
+    sortedNodes.forEach((n, i) => {
+      nodeIndex.set(n.address, i);
+    });
```
wchargin, line 716: Cool. We may at some point want this to also just be the internal
representation (i.e., instance variables). But we can do that whenever.

wchargin (approved):

PR #2312, opened 2020-10-03
GitHub: Retry when hitting API rate limit

### LINKED ISSUE #2201: GitHub plugin load should not die due to rate limit
When we're loading the GitHub plugin, it currently dies when the API rate limit is exceeded. When the user is working with a local cache, this is OK because they can wait an hour and re-try. However, when this is happening on a GitHub action, it would be far better if the job would simply wait an hour and then continue loading, so that it would eventually succeed (and save all the data in cache).

I believe we should add the wait/and/retry logic here:
https://github.com/sourcecred/sourcecred/blob/cd92a24ba1456163040124868a5d9944dcf9d4b8/src/plugins/github/fetchGithubRepo.js#L263-L268

decentralion:
This modifies the GitHub fetchGithubRepo file so that when the GitHub
API rate limit is exceeded, we retry. We also modify the retry
parameters to ensure that our last retry will be after just after one
hour, which definitely gives time for us to hit the next API rate limit
reset.

We are using the same retry strategy for every error. This means that on
RATE_LIMIT_EXCEEDED, we'll retry after 1s, 2.8s, 22s, etc. These early
retries are really unlikely to work, since the API rate limit is reset
on an hourly cadence. However, the retry library doesn't allow modifying
the retry strategy in flight, so let's take this expedient hack. I don't
think the 7 extra failed API calls will bother GitHub overmuch.

Test plan: The retry logic is untested, which is reasonable as it would
be finicky to test, and doesn't get modified very much. As a test plan,
I'll try loading a large instance which is known to fail due to rate
limit exceeded, and see if it succeeds instead.

Fixes #2201

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -134,7 +134,7 @@ const GITHUB_GRAPHQL_SERVER = "https://api.github.com/graphql";
 type GithubResponseError =
   | {|+type: "FETCH_ERROR", retry: false, error: Error|}
   | {|+type: "GRAPHQL_ERROR", retry: false, error: mixed|}
-  | {|+type: "RATE_LIMIT_EXCEEDED", retry: false, error: mixed|}
+  | {|+type: "RATE_LIMIT_EXCEEDED", retry: true, error: mixed|}
   | {|+type: "GITHUB_INTERNAL_EXECUTION_ERROR", retry: true, error: mixed|}
   | {|+type: "BAD_CREDENTIALS", retry: false, error: mixed|}
   | {|+type: "NO_DATA", retry: true, error: mixed|};
@@ -165,7 +165,7 @@ function tryGithubFetch(fetch, fetchOptions): Promise<any> {
             return Promise.reject(
               ({
                 type: "RATE_LIMIT_EXCEEDED",
-                retry: false,
+                retry: true,
                 error: x,
               }: GithubResponseError)
             );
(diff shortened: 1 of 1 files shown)


befitsandpiper (approved):
Looks good

src/plugins/github/fetchGithubRepo.js
```diff
@@ -206,7 +206,24 @@ function tryGithubFetch(fetch, fetchOptions): Promise<any> {
...
+    // the limit has reset.
+    // In principle, we shouldn't bother having a bunch of immediate retrys
+    // when we hit rate limit exceeded, but it won't cause any problems, and the
+    // retry library does not offer any way to vary the retry period based
+    // on the specific error received.
+    //
+    // Verification that we will issue a retry after an hour:
+    // 2.8 ** 8 = 3778
+    // Minimum timeout is one second (1000 ms)
+    // 3778 seconds = 1.05 hours
```
wchargin, line 221: This shows that the delay _between_ the 7th and 8th retries is
3778 seconds. Don’t you actually want to show that the _total_ time
spent waiting between the 0th and 8th retries is just over an hour?

That is, you want to pick initial timeout `a`, ratio `r`, and retry
count `n` such that

```latex
\frac{a (r^0 - r^{n+1})}{1 - r} = 3600 + \varepsilon
```

(because the finite geometric series is just the difference of two
infinite geometric series with different initial terms).

So, with `a = 1.0` and `n = 8`, you want `(1 - r^9) / (1 - r)` to equal
(say) `3600 * 1.05`, which is satisfied for `r \approx 2.638`.

Confirm math:

```
$ ghc -e 'let { r = 2.638; a = 1.0; n = 8 } in sum [a * r^k | k <- [0..n]] / 3600'
1.049026649634996
```

Confirm API:

```
[✓] 21:38:25(1601699905) [5e0894f64d]
wchargin@wchargin-t440s ~/git/sourcecred
$ cat test.js
const retry = require("retry");
const op = retry.operation({factor: 2, retries: 3, minTimeout: 1000});
const first = Date.now();
let last = first;
op.attempt(() => {
  const now = Date.now();
  console.log(`delta: ${now - last} ms; total: ${now - first} ms`);
  last = now;
  op.retry(true);
});

[✓] 21:38:26(1601699906) [5e0894f64d]
wchargin@wchargin-t440s ~/git/sourcecred
$ node test.js
delta: 0 ms; total: 0 ms
delta: 1004 ms; total: 1004 ms
delta: 2002 ms; total: 3006 ms
delta: 4000 ms; total: 7006 ms
```

wchargin (commented):
Not sure the math is right? See inline.

src/plugins/github/fetchGithubRepo.js
wchargin: FWIW, the error as written causes the requests to be sent at 0.58 hours past and then again at 1.63 hours past, so I do think that this matters, since you burn 38 minutes:

```
$ ghc -e 'let { r = 2.8; a = 1.0 / 3600; n = 8 } in mapM_ print $ scanl1 (+) [a * r^k | k <- [0..n]]'
2.777777777777778e-4
1.0555555555555555e-3
3.233333333333333e-3
9.33111111111111e-3
2.640488888888888e-2
7.421146666666664e-2
0.20806988444444435
0.5828734542222219
1.632323449599999
```

PR #2314, opened 2020-10-03
helper fn for prepping graph, ledger, and deps

decentralion:
This adds a cli/common helper method for loading the graph, ledger, and
dependency policies for Cred computation. These three are intertwined in
that the dependency policies may modify the ledger, and the ledger then
modifies the graph, so it makes sense to have a single helper that
orchestrates this. I also added helpers for loading/saving the ledger.
This commit refactors the existing CLI score command to use the helper;
it will make a future change to CredRank easier as well.

Test plan: run `sourcecred score` in an existing instance. Scores should
not change. Run `sourcecred score` in an instance with unset ids in the
dependency file, and verify that they are set and serialized to both the
ledger and the dependencies file correctly. (nb. the template instance
is fine for this purpose.)

### DIFF
--- a/src/cli/common.js
+++ b/src/cli/common.js
@@ -3,20 +3,29 @@
 import {join as pathJoin} from "path";
 import fs from "fs-extra";
 import {loadJson, mkdirx} from "../util/disk";
+import deepEqual from "lodash.isequal";
+import stringify from "json-stable-stringify";
 
+import {type DependencyMintPolicy} from "../core/dependenciesMintPolicy";
 import type {PluginDirectoryContext} from "../api/plugin";
 import {
   parser as configParser,
   type InstanceConfig,
 } from "../api/instanceConfig";
-
+import {Ledger} from "../ledger/ledger";
+import {contractions as identityContractions} from "../ledger/identity";
+import {
+  parser as dependenciesParser,
+  ensureIdentityExists,
+  toDependencyPolicy,
+} from "../api/dependenciesConfig";
 import {
   type WeightedGraph,
   merge,
   overrideWeights,
   fromJSON as weightedGraphFromJSON,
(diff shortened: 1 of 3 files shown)


src/cli/common.js
```diff
@@ -60,6 +69,68 @@ export function pluginDirectoryContext(
   };
 }
 
+/**
+ * This method pipelines the data loading needed to run Cred analysis.
+ *
+ * It's a bit of a kitchen sink, but I think it's still valuable to have as its
+ * own method. In particular, there are some non-obvious relationships between the constituent components:
+ * it's important that the identities be computed after loading the dependencies, because
+ * loading the dependencies may create new identities. Thus, I prefer to expose this API which sequences
```
wchargin, line 78: Great point; this is a good thing to encode here.

src/cli/common.js
```diff
@@ -60,6 +69,68 @@ export function pluginDirectoryContext(
...
+export async function prepareCredData(
+  baseDir: string,
+  config: InstanceConfig
+): Promise<{|
+  graph: WeightedGraph,
+  ledger: Ledger,
+  dependencies: $ReadOnlyArray<DependencyMintPolicy>,
+|}> {
+  const weightedGraph = await loadWeightedGraph(baseDir, config);
+  const ledger = await loadLedger(baseDir);
```
wchargin, line 90: Maybe

```javascript
const [weightedGraph, ledger] = await Promise.all([
  loadWeightedGraph(baseDir, config),
  loadLedger(baseDir),
]);
```

? Since we’re steeped in async-land, we might as well try to get some of
the benefits.

wchargin (approved):

src/cli/common.js
decentralion: Done.

PR #2282, opened 2020-09-22
DRAFT: credrank: add dependency cred

wchargin:
Summary:
Conceptually paired with @decentralion: the gadget was mostly their idea
(in a joint conversation), though I wrote the code.

Test Plan:
Yes, please do!

wchargin-branch: credrank-dependency-cred

PR #2316, opened 2020-10-03
DRAFT2: credrank: add dependency cred

decentralion:
This is a functional version of #2282. It still needs testing.

One small test: run `credrank` inside the sc cred instance. Currently,
mints about 7x more dependency cred for SourceCred than is intended. We
should demonstrate that the mint amount is stable around our intention,
and independent of the choice of beta.

Paired with @wchargin, see #2282


decentralion:
This work was all obsoleted by the virtualization push, I now have a distinct branch which adds dependency Cred. (Haven't PR'd it yet, waiting to fix math issues.)

PR #2317, opened 2020-10-04
util: add `retry` module

wchargin:
Summary:
The GitHub plugin uses the NPM `retry` module to retry individual
queries. But that module’s API is too restrictive for us, on two counts:

 1. There’s no way to say “wait until this specific time, after which my
    API quota will have refreshed”. Or, from another perspective, every
    operation has just a single retry policy.
 2. The jitter range is hardcoded to `Uniform[1.0, 2.0]`; I’d prefer
    that it be something more like `1.2` to reduce variance.

This patch introduces a simple `retry` module that fits our needs more
precisely. It distinguishes a “retry with backoff” due to transient
error from a “wait until this time” due to quota exhaustion. Other
benefits of this module include:

  - The API is simpler. Clients don’t have to maintain a reference to a
    stateful operation object and manually request retries. Instead, a
    client simply provides an async function.
  - Since “retries” and “waits” are distinguished, we can test that
    logic in the retry module itself.
  - The module has Flow types and is written in a type-friendly style.

If useful, the `Result<T>` type could be enhanced with a record of all
the retry events: “waited 2.1 seconds due to transient error, then
waited 10 minutes for budget refresh, then waited 3.9 seconds due to
transient error”. This would be easy to add, but we don’t need it yet.

Test Plan:
Unit tests included, with full coverage.

wchargin-branch: util-retry

### DIFF
--- a/src/util/retry.js
+++ b/src/util/retry.js
@@ -1,0 +1,142 @@
+// @flow
+
+// The outcome of a single attempt of a retryable operation.
+export type AttemptOutcome<+T, +E> =
+  // The operation has terminated successfully.
+  | {|+type: "DONE", +value: T|}
+  // The operation has terminated with a fatal error.
+  | {|+type: "FATAL", +err: E|}
+  // The operation has encountered a transient error and should retry
+  // after a delay, if retries are still available.
+  | {|+type: "RETRY", +err: E|}
+  // The operation has exceeded its rate-limit budget and should wait
+  // until the given timestamp for that budget to be refreshed, then
+  // retry.
+  | {|+type: "WAIT", +until: Date, +err: E|};
+
+// The final result of executing an operation after retrying zero or more
+// times.
+export type Result<+T, +E> =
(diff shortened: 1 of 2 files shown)


src/util/retry.js
```diff
@@ -0,0 +1,144 @@
...
+        if (waits >= fullPolicy.maxWaits) {
+          return {type: "FAILED", err: outcome.err};
+        }
+        waits++;
+        const now = io.now();
+        const delayMs = outcome.until - now;
+        if (delayMs < 0) {
+          throw new Error(
+            "wait-until time in the past: " +
+              `@${String(+outcome.until)} < @${String(+now)}`
```
decentralion, line 112: nit: @ implies seconds, this is ms

decentralion (approved):

src/util/retry.js
wchargin: Indeed it does. Fixed.

PR #2318, opened 2020-10-04
markovProcessGraph: take intervals & refactor `new`

decentralion:
This modifies the MPG so that it takes the time intervals as an explicit
argument, rather than implicitly choosing intervals based on week-long
period. This will make the MPG easier to test. I also refactored the
`new` method so that it takes an object of arguments, and combined the
`seedOptions` and `fibrationOptions` into `parameters`.

This is mostly just a refactor, but we're using a slightly different
implementation for choosing weeklong intervals from a graph, such that
we now include node timestamps when picking intervals, not just edge
timestamps. This shouldn't matter in practice.

Test plan: Cred scores are unchanged.

### DIFF
--- a/src/core/markovProcessGraph.js
+++ b/src/core/markovProcessGraph.js
@@ -55,8 +55,6 @@ import {type Uuid} from "../util/uuid";
  * spectral analysis via the `toMarkovChain` method.
  */
 
-import {max, min} from "d3-array";
-import {weekIntervals} from "./interval";
 import sortedIndex from "lodash.sortedindex";
 import sortBy from "../util/sortBy";
 import {makeAddressModule, type AddressModule} from "./address";
@@ -76,6 +74,7 @@ import * as NullUtil from "../util/null";
 import * as MapUtil from "../util/map";
 import type {TimestampMs} from "../util/timestamp";
 import {type SparseMarkovChain} from "./algorithm/markovChain";
+import {type IntervalSequence} from "./interval";
 
 export type TransitionProbability = number;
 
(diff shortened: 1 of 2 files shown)


src/core/markovProcessGraph.js
```diff
@@ -260,7 +259,16 @@ const CONTRIBUTION_RADIATION = EdgeAddress.fromParts([
...
 
-export type FibrationOptions = {|
+export type Arguments = {|
+  +weightedGraph: WeightedGraphT,
+  +participants: $ReadOnlyArray<Participant>,
+  +intervals: $ReadOnlyArray<Interval>,
+  +parameters: Parameters,
+|};
+
+export type Parameters = {|
```
wchargin, line 269: Heh. It _is_ a little weird to have one parameter called `Arguments` and
another parameter called `Parameters` (especially since most people seem
to use “argument” and “parameter” interchangeably, anyway). I’m fine
with this for now, and might come in with a better name if I can think
of one.

(Are these hyperparameters?)

src/core/markovProcessGraph.js
```diff
@@ -347,15 +347,11 @@ export class MarkovProcessGraph {
...
-        wg.graph.edges({showDangling: false})
-      ).map((x) => x.timestampMs);
-      const start = min(edgeTimestamps);
-      const end = max(edgeTimestamps);
-      const boundaries = weekIntervals(start, end).map((x) => x.startTimeMs);
-      return [-Infinity, ...boundaries, Infinity];
-    })();
+    const timeBoundaries = [
+      -Infinity,
+      ...intervals.map((x) => x.startTimeMs),
```
wchargin, line 352: We depend on `intervals` being sorted; could you please check that it
is? (Probably want `timeBoundaries` to be strictly-sorted, in fact.)

wchargin (approved):
Approved modulo assertion rqeuest

src/core/markovProcessGraph.js
decentralion: Done via #2323

PR #2319, opened 2020-10-04
github: use `util/retry` to fetch repositories

wchargin:
Summary:
We switch `fetchGithubRepo.js` from the NPM `retry` module to our own
`util/retry`. This commit is a pure refactor; in particular, we keep the
default policy settings from NPM `retry` (10 retries, no jitter). In a
follow-up commit, we can use the new “wait until” feature.

Test Plan:
Point a test instance at `sourcecred/example-github`, for loading speed.
Note that running `sourcecred load` works, loading in ~4 seconds. Then,
patch `resolveRepositoryGraphqlId` to make the query invalid: say,
change `"owner"` to `"ownerz"`. For convenience, change the policy from
10 retries to 3. Re-build, and note that `load` fails promptly (500 ms)
with a nice error. Then, patch the types so that `GRAPHQL_ERROR` is
considered retryable, and update the value layer to fix the resulting
type error. Re-build, and note that `load` fails after much longer
(somewhere around 9 seconds), indicating that it’s backing off and
retrying before failing.

wchargin-branch: github-fetchrepo-util-retry

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -6,7 +6,6 @@
 
 import Database from "better-sqlite3";
 import fetch from "isomorphic-fetch";
-import retry from "retry";
 
 import {type RepoId, repoIdToString} from "./repoId";
 import {Mirror} from "../../graphql/mirror";
@@ -18,6 +17,7 @@ import schema from "./schema";
 import {type GithubToken} from "./token";
 import {cacheIdForRepoId} from "./cacheId";
 import {type CacheProvider} from "../../backend/cache";
+import retry from "../../util/retry";
 
 type FetchRepoOptions = {|
   +token: GithubToken,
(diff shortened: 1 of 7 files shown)


src/plugins/github/fetchGithubRepo.js
```diff
@@ -6,7 +6,6 @@
 
 import Database from "better-sqlite3";
 import fetch from "isomorphic-fetch";
-import retry from "retry";
```
decentralion, line 9: This is the only usage of "retry" -- remove the dependency?

decentralion (approved):

src/plugins/github/fetchGithubRepo.js
wchargin: Huh, so it is. I just assumed that Discourse and Discord would also be
using it. Removed; thanks.

PR #2320, opened 2020-10-04
github: reduce retry cap, add jitter

wchargin:
Summary:
The default for the NPM `retry` module was to retry requests 10 times.
That seems a bit exorbitant for transient errors, and makes debugging
annoying, since `2 ** 10` seconds is just over 17 minutes. We reduce the
count to 5 (arbitrarily).

We also add jitter to avoid large simultaneous load on GitHub servers.

Test Plan:
None; the functionality is tested in `util/retry.test.js`, and this is
just a policy change.

wchargin-branch: github-retry-jitter


decentralion (approved):
Really farming that cred minting ;)

wchargin (commented):
I split this PR from its neighbors because I wanted #2319 to be a pure
refactor and wanted #2321 to be a pure bugfix. I think you’ll agree that
this is on-brand. :-) Would you prefer that I squash #2321 into this?

(I did actually consider splitting the retry cap and the jitter into
separate PRs, since they’re logically unrelated, but decided that it
would be too noisy.)

decentralion:
I think it would make sense to squash it with #2321, but I don't object to this either. I was teasing you about the cred farming. In the long run, this "attack" will cease to garner undue rewards, as we switch away from activity-based minting entirely. In the short term, it's pretty clear that you are not an attacker, and an extra PR has a negligible impact on your overall Cred :)

wchargin:
Cool. If you don’t object, I’ll keep the bugfix separate, yes. Thanks!

PR #2323, opened 2020-10-04
add IntervalSequence

decentralion:
An IntervalSequence is an opaque read-only array of intervals, with
runtime validation that the intervals are non-overlapping and also have
no gaps between them, and that every interval has non-negative duration.
This way we can use intervals and not worry that the interval may be
out-of-order or have gaps, etc.

Test plan: Unit tests added to verify that the interval sequences have
the expected properties. Also, `git grep -F '$ReadOnlyArray<Interval>'`
only has hits within the interval module itself.

### DIFF
--- a/src/core/interval.js
+++ b/src/core/interval.js
@@ -19,6 +19,43 @@ export type Interval = {|
 |};
 
 /**
+ * An interval sequence is an array of intervals with the following guarantees:
+ * - Every interval has positive time span (i.e. the end time is greater
+ *   than the start time).
+ * - Every interval except for the first starts at the same time that the
+ *   previous interval ended.
+ * - No interval may have a NaN start or end time. (Infinity is OK.)
+ */
+export opaque type IntervalSequence: $ReadOnlyArray<Interval> = $ReadOnlyArray<Interval>;
+
+export function intervalSequence(
+  intervals: $ReadOnlyArray<Interval>
+): IntervalSequence {
+  let lastEndTime = null;
+  for (const {startTimeMs, endTimeMs} of intervals) {
+    if (isNaN(startTimeMs) || isNaN(endTimeMs)) {
+      throw new Error(`NaN in interval [${startTimeMs}, ${endTimeMs}]`);
+    }
+    if (lastEndTime != null && startTimeMs !== lastEndTime) {
+      throw new Error(
+        `last interval ended at ${lastEndTime} but this interval starts at ${startTimeMs}`
+      );
+    }
+    if (endTimeMs <= startTimeMs) {
+      throw new Error(
+        `interval must have positive length, but got [${startTimeMs}, ${endTimeMs}]`
+      );
+    }
+    lastEndTime = endTimeMs;
+  }
+  // Reconstruct the objects so mutating the input won't break the IntervalSequence
+  return intervals.map(({startTimeMs, endTimeMs}) => ({
+    startTimeMs,
+    endTimeMs,
+  }));
+}
+
+/**
  * Represents a slice of a time-partitioned graph
  * Includes the interval, as well as all of the nodes and edges whose timestamps
  * are within the interval.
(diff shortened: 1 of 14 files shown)


src/core/interval.js
```diff
@@ -18,6 +18,34 @@ export type Interval = {|
...
+      );
+    }
+    if (endTimeMs < startTimeMs) {
+      throw new Error(
+        `negative length interval: ${endTimeMs} < ${startTimeMs}`
+      );
+    }
+    lastEndTime = endTimeMs;
+  }
+  return intervals;
```
wchargin, line 46: Unsound; this type-checks but fails:

```javascript
    it("resists attack", () => {
      const i0 = {startTimeMs: 0, endTimeMs: 100};
      const i1 = {startTimeMs: 100, endTimeMs: 200};
      const is = intervalSequence([i0, i1]);
      i1.startTimeMs = 300;
      intervalSequence((is: IntervalSequence)); // fails
    });
```

Deep-copy?

src/core/interval.test.js
```diff
@@ -3,9 +3,70 @@
...
+      const i1 = {startTimeMs: 100, endTimeMs: 99};
+      const thunk = () => intervalSequence([i1, i0]);
+      expect(thunk).toThrowError("negative length interval: 99 < 100");
+    });
+    it("accepts a sequence with -Infinity and +Infinity", () => {
+      const i0 = {startTimeMs: -Infinity, endTimeMs: 100};
+      const i1 = {startTimeMs: 100, endTimeMs: +Infinity};
+      expect(intervalSequence([i0, i1])).toEqual([i0, i1]);
+    });
+    it("rejects a sequence with NaNs", () => {
```
wchargin, line 62: NaNs can still slip through. This fails:

```javascript
    it("rejects a sequence with just one NaN interval", () => {
      const i0 = {startTimeMs: NaN, endTimeMs: NaN};
      const thunk = () => intervalSequence([i0]);
      expect(thunk).toThrowError("NaN");
    });
```

src/core/interval.js
```diff
@@ -18,6 +18,34 @@ export type Interval = {|
   +endTimeMs: TimestampMs,
 |};
 
+/**
+ * An interval sequence is an array of intervals with the following guarantees:
+ * - Every interval has non-negative time spam (i.e. the end time is greater
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/interval.js
```diff
@@ -18,6 +18,34 @@ export type Interval = {|
   +endTimeMs: TimestampMs,
 |};
 
+/**
+ * An interval sequence is an array of intervals with the following guarantees:
```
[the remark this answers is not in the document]
decentralion: Done.

src/core/interval.js
decentralion: Fixed.

src/core/interval.test.js
decentralion: Fixed.

src/core/interval.js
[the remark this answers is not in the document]
decentralion: Yeah, I think being more restrictive here makes sense.

src/core/interval.test.js
```diff
@@ -3,9 +3,87 @@
...
+      const i1 = {startTimeMs: 100, endTimeMs: 200};
+      const thunk = () => intervalSequence([i0, i1]);
+      expect(thunk).toThrowError("NaN");
+    });
+    it("resists mutation of the intervals", () => {
+      const i0 = {startTimeMs: 0, endTimeMs: 100};
+      const i1 = {startTimeMs: 100, endTimeMs: 200};
+      const is = intervalSequence([i0, i1]);
+      i1.startTimeMs = 300;
+      intervalSequence(is);
```
wchargin, line 83: Mind explicitly checking that `is[0]` still has valid values, rather
than just checking that it still validates? This should be sufficient,
but it’s less obvious to me, and it’s nice for tests to be obvious.

wchargin (approved):
Thanks!

src/core/interval.test.js
decentralion: Done.

wchargin:
> the expected properties. Also, `git grep -F "$ReadOnlyArray<Interval>"`

I’ve updated this to `git grep -F '$ReadOnlyArray<Interval>'`; the
double-quoted version doesn’t do what you want in POSIX shells.

Also, it’s pleasing to see all the existing uses gaining this
validation, with no errors!

PR #2325, opened 2020-10-04
migrate ledger and identities to core

decentralion:
The `ledger` and `identity` modules are both now de-facto core concepts,
in that they are extensively depended upon by the plugins, and part of
the long-term solid APIs we're building out. Thus, it makes sense to
migrate them to core.

This adds a few dependency edges from core into analysis, but they will
be short-lived. Specifically, as soon as we remove TimelineCred and
update the ledger grain distribution logic to reference CredRank scores,
the core->analysis dep edges will be gone.

Test plan: `yarn flow` suffices for a file rename


wchargin (approved):

wchargin:
enjoy your cred minting ;-)

PR #2326, opened 2020-10-04
github: decouple error types from dispositions

wchargin:
Summary:
Previously, our GitHub fetch errors included a `retry` flag in the error
itself (with a value mandated by the variant). We now care about more
than just a boolean “retry” flag, since the correct response to an error
may be to wait until a certain time. Instead of just changing the field
on the error values, we introduce a new `errorDisposition` to specify
how to respond to a given error. This opens the door to more substantial
logic, like querying for the time at which quotas refresh, or retrying
only some kinds of `FETCH_ERROR`s (`ETIMEDOUT`).

wchargin-branch: github-error-decouple-retry

### DIFF
--- a/src/plugins/github/fetchGithubRepo.js
+++ b/src/plugins/github/fetchGithubRepo.js
@@ -17,7 +17,7 @@ import schema from "./schema";
 import {type GithubToken} from "./token";
 import {cacheIdForRepoId} from "./cacheId";
 import {type CacheProvider} from "../../backend/cache";
-import retry from "../../util/retry";
+import retry, {type AttemptOutcome} from "../../util/retry";
 
 type FetchRepoOptions = {|
   +token: GithubToken,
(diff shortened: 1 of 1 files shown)


src/plugins/github/fetchGithubRepo.js
```diff
@@ -163,47 +162,56 @@ function tryGithubFetch(fetch, fetchOptions): Promise<any> {
...
       }),
     (e) =>
-      Promise.reject(
-        ({type: "FETCH_ERROR", retry: false, error: e}: GithubResponseError)
-      )
+      Promise.reject(({type: "FETCH_ERROR", error: e}: GithubResponseError))
   );
 }
 
+function errorDisposition(
```
hammadj, line 192: Now that the error handling logic is separated from the fetch logic, would be good to add tests for this

hammadj (approved):
Looks good, might be worth writing tests now for errorDisposition

src/plugins/github/fetchGithubRepo.js
```diff
@@ -163,47 +162,56 @@ function tryGithubFetch(fetch, fetchOptions): Promise<any> {
...
+    return {type: "RETRY", err: e};
+  }
+  if (e.type === "RATE_LIMIT_EXCEEDED") {
+    // Wait in 15-minute increments. TODO(@wchargin): Ask GitHub
+    // when our token resets (`{ rateLimit { resetAt } }`) and wait
+    // until just then.
+    const delayMs = 15 * 60 * 1000;
+    return {type: "WAIT", until: new Date(Date.now() + delayMs), err: e};
+  }
+  throw new Error((e.type: empty));
```
hammadj, line 212: This function is called in a catch block, so throwing here would cause an uncaught error. The previous implementation just returned a `{type: "FATAL", err: e}` instead, seems like that's what we would want here as well?

wchargin: This line is unreachable, which you can tell because it has access to a
value at the empty type. We only hit this case if the types are unsound.

This `throw new Error((...: empty))` pattern is very common throughout
our codebase; try `git grep `: empty` for examples.

src/plugins/github/fetchGithubRepo.js
wchargin: Hmm—unit tests here don’t seem useful to me. The only nontrivial case is
`RATE_LIMIT_EXCEEDED`, so we could test that. But the tests would be
immediately removed in #2327, because this function becomes impure. Note
that in #2327 I _do_ add unit tests for the new logic.

What tests would you like to see, to fix which failure mode?

> Naming nit: `getErrorDisposition` instead of just `errorDisposition`

No; superfluous `get`s provide no information, and we avoid them. See
`Graph.node`, `MarkovProcessGraph.participants`, `CredView.weights`,
`CacheProvider.database`, `Ledger.account`, ….

decentralion (approved):

PR #2334, opened 2020-10-05
Improve Ledger Admin UX

hammadj:
The editor being at the bottom of the Identity list makes it difficult to edit identities since users have to scroll through the list, click an identity, and then scroll all the way down to edit it, all without any visual indicator that a new editor menu is open.

This cleans up the UX by making the list of identities a scrollable box within the page and moving the editor fields / button to the top of the page so that they stay in the same location and user isn't required to scroll through potentially thousands of list items. It also improves some of the labels and buttons to be more descriptive / cleaner.

image

Also renamed the sidebar button from "Ledger Admin" to "Identities" since that's the only thing this page really does. Grain transfer page also modifies the ledger, which means "ledger admin" tasks happen across multiple pages, and therefore no single page should be called "Ledger Admin".

Before:

After:

### DIFF
--- a/src/ui/components/AliasView.js
+++ b/src/ui/components/AliasView.js
@@ -4,7 +4,6 @@ import React, {type Node as ReactNode} from "react";
 import {type IdentityId} from "../../core/identity";
 import Markdown from "react-markdown";
 
-import {makeStyles} from "@material-ui/core/styles";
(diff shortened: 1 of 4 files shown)


topocount (approved):
Thanks! This is much needed

PR #2335, opened 2020-10-05
Implement MVP of Ledger History Viewer

hammadj:
Super basic start for having the ledger event history visible on the frontend. Currently
shows the event name, date / time, and info about the identities for "CREATE_IDENTITY" and
"TOGGLE_ACTIVATION" events. For advanced users, to get the identityId for a specific user/event,
you can hover over the name of the identity and it will appear in a tooltip.

Will expand on the info for other events in future PRs, wanted to get the basics in first

Test Plan:
Ensure the ledger UI is accessible from the "Ledger History" button in the sidebar,
and that it displays a chronological list of events from the ledger with correct dates.

image

### DIFF
--- a/src/ui/components/AdminApp.js
+++ b/src/ui/components/AdminApp.js
@@ -19,6 +19,7 @@ import {withRouter} from "react-router-dom";
 import AppBar from "./AppBar";
 import createMenu from "./Menu";
 import {LedgerProvider} from "../utils/LedgerContext";
+import {LedgerViewer} from "./LedgerViewer";
 
 const dataProvider = fakeDataProvider({}, true);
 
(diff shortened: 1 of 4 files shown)


decentralion (approved):
Looks like a reasonable start. A few thoughts:
- It's better to show reverse-chronological order, in general recent activity is much more interesting, and more relevant for review purposes
- It's pretty slow b.c. it loads all the events in one view--can we implement some kind of pagination to avoid this?

also, naturally, we'll want features like filtering event types, diving into distribution info, etc :)
one thing I'm noticing here is the "CREATE_IDENTITY" then "ADD_ALIAS" pattern is kind of noisy--we should probably make a single ledger event that adds many identities with starting aliases--just to clean things up.
Gonna approve so as to give you discretion on how to sequence those improvements.

PR #2336, opened 2020-10-05
Cypress Demo

topocount:
Cypress is a versatile testing framework that can be used for unit and
end-to-end frontend tests. It utilizes mocha and chai libraries for
writing tests.

The flow libdefs for all of these libraries are quite lacking, however
and I'm strongly considering using a .flow file that is localized to the
cypress folder to resolve this, and to mitigate collisions
with Jest.

Cypress has not been hooked up to the larger testing sequence we have,
	as no useful test have been written yet.

test plan: ensure all existing automated test pass


topocount:
I'm opening up this Pull Request for review for two reasons:   
1. We still need to figure out how we want to use cypress:
- snapshot testing: probably
- frontend unit tests: maybe, but we've heard some justified concerns regarding maintainability. I personally believe that Cypress' api is simple enough and their documentation is so good that there isn't a significant learning curve here. 

2. Configuration details: I've run into some issues setting up the flow libdefs to work with cypress' testing libraries (see [here](https://github.com/sourcecred/sourcecred/pull/2336/commits/2c3bfa56034c0abaef9823f86d0c4bc02dddcad2#diff-dff35056b37059d00cf7d738022c414eR4-R13) for details) and there doesn't seem to be an ideal path forward.

Issue #2339, opened 2020-10-05
Reduce ledger spam from auto-generating identities


decentralion:
Currently, the identity proposal system spams the ledger log with entries like:

```
CREATE_IDENTITY
ADD_ALIAS
CREATE_IDENTITY
ADD_ALIAS
CREATE_IDENTITY
ADD_ALIAS
```

Because every new identity proposal comes with an alias, but we aren't allowed to create identities with aliases.

We can clean this up by:
- first, allowing creating identities with aliases
- second, adding some kind of "BULK_CREATE_IDENTITIES" action that abstracts over creating many identities

This way, running the identity proposal system would create a single entry in the ledger, which would be a lot cleaner. Would generally make reviewing ledger changes much more sane and approachable.

hammadj:
What implications does this have on existing ledgers? Are we able to update the history, or will older ledgers be forever doomed to original design decisions? Is there / should there be a ledger migration process? I fear that as we make more changes, some of them might not be backwards compatible and we would end up having to support a bunch of legacy stuff in front end ledger viewer as well.

decentralion:
We'll support full backwards compatibility. Old ledgers will have ugly history, but that doesn't really matter. I want to make the review flow better, ppl are mostly looking at the latest activity not the old stuff.

We could write a ledger migration process; we might need it eventually. But I think maintaining it will be a headache, we should think a lot about the process before we set it up. Fortunately, I don't think we need it for this change. :)

wchargin:
Saw this earlier, want to leave a brief comment:

I encourage you to care much more about the format of the data stored in
the ledger than the workflows currently built atop it. As you mention,
we want to support full backward compatibility. The ledger structure is
the core, and a core should be composed of simple pieces that compose
rather than a partially filled matrix of checkboxes.

My inclination is that adding identities in bulk may be a reasonable
operation, but I’m more hesitant about bundling together identity
creation with alias linking.

By tenuous analogy, `mkdir -p` is easy to understand, but there’s no
standard operation `mkleaf foo/bar/baz` that makes directories `foo` and
`foo/bar/` and then creates an empty file at `foo/bar/baz`.

If the problem is review, solve that problem better. Create a tiny
script that takes a ledger diff and summaries the changes. Given (pseudo):

```json
{type: "CREATE_IDENTITY", id: "xxx", name: "alice"}
{type: "ADD_ALIAS", id: "xxx", target: "github/alice"}
[...]

PR #2345, opened 2020-10-06
allow creating new identity with starting aliases

teamdandelion:
Part of work on #2339. Decided to factor this out as a nice clean atomic
starting commit. I added logic around verifying that aliases are not
duplicated, so that the updated `ledger.createIdentity` could have
consistent behavior with `ledger.addAlias` (erroring if there's an
attempt to give the same alias to the same identity multiple times).

Test plan: `yarn test`, note unit tests were added.

### DIFF
--- a/src/core/identity/identity.js
+++ b/src/core/identity/identity.js
@@ -43,19 +43,29 @@ export const IDENTITY_PREFIX: NodeAddressT = NodeAddress.fromParts([
   "IDENTITY",
 ]);
 
-export function newIdentity(subtype: IdentityType, name: string): Identity {
+export function newIdentity(
+  subtype: IdentityType,
+  name: string,
+  aliases?: $ReadOnlyArray<Alias>
+): Identity {
   const id = randomUuid();
   try {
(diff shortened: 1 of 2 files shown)


src/core/identity/identity.js
```diff
@@ -43,19 +43,29 @@ export const IDENTITY_PREFIX: NodeAddressT = NodeAddress.fromParts([
   "IDENTITY",
 ]);
 
-export function newIdentity(subtype: IdentityType, name: string): Identity {
+export function newIdentity(
+  subtype: IdentityType,
+  name: string,
+  aliases?: $ReadOnlyArray<Alias>
```
hammadj, line 49: Instead of having two different variables (aliases + actualAliases), can just set default value here:
```suggestion
  aliases: $ReadOnlyArray<Alias> = []
```

hammadj (approved):
small nit

wchargin (commented):
Non-blocking review, but please read my comment here:
<https://github.com/sourcecred/sourcecred/issues/2339#issuecomment-704500827>

decentralion:

PR #2346, opened 2020-10-06
ledger: support creating identities with aliases

decentralion:
This is some more work on #2339. The Ledger `createIdentity` method now
supports creating identities with initial aliases. Happily, the
underlying action type support full identities, so no changes were made
to the data layer, and full backwards-compatibility is maintained.

I've updated the unit tests to reflect the new behavior.

Test plan: `yarn test`, and careful inspection of the code.

### DIFF
--- a/src/core/ledger/ledger.js
+++ b/src/core/ledger/ledger.js
@@ -169,8 +169,12 @@ export class Ledger {
    *
    * Will fail if the name is not valid, or already taken.
    */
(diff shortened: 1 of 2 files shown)


hammadj:
Removed the ready for review label based on @wchargin's comments on https://github.com/sourcecred/sourcecred/pull/2346, seems like this will be reworked a little

decentralion:
Yep, thanks @hammadj.

PR #2347, opened 2020-10-06
Reduce autogenerated ledger spam by 2x

decentralion:
This commit makes somes solid progress on #2339. Now that we support
creating identities with aliases, we can just do that inside the
identity proposal system. This reduces the number of event entries by
factor 2 when autogenerating identities.

Test plan: Existing behavior is well-covered by unit tests, added one
more checking that we output only a single event.


decentralion:

PR #2356, opened 2020-10-08
Implement viewer for Grain Distributions

hammadj:
Adds Components to view Grain Distribution events in the Ledger viewer with a breakdown of the
different policies used and the grain amounts received by the participants.

Test Plan: Open the LedgerViewer page, click on the grain distribution amount, and ensure a list of grain receipts is shown

### DIFF
--- a/src/ui/components/AdminApp.js
+++ b/src/ui/components/AdminApp.js
@@ -67,7 +67,7 @@ const customRoutes = (
       <AccountOverview currency={currency} />
     </Route>,
     <Route key="ledger" exact path="/ledger">
-      <LedgerViewer />
+      <LedgerViewer currency={currency} />
     </Route>,
   ];
   const backendRoutes = [
--- a/src/ui/components/LedgerViewer.js
+++ b/src/ui/components/LedgerViewer.js
@@ -11,6 +11,7 @@ import Table from "@material-ui/core/Table";
 import Typography from "@material-ui/core/Typography";
 import Box from "@material-ui/core/Box";
 import TableBody from "@material-ui/core/TableBody";
+import VisibilityIcon from "@material-ui/icons/Visibility";
 import Dialog from "@material-ui/core/Dialog";
 import * as G from "../../core/ledger/grain";
 import DialogContent from "@material-ui/core/DialogContent";
@@ -199,8 +200,11 @@ const LedgerEventRow = React.memo(
                   0,
                   currencySuffix
                 )}`}
+                clickable
                 onClick={() => handleClickOpen(a)}
+                onDelete={() => handleClickOpen(a)}
                 size="small"
+                deleteIcon={<VisibilityIcon />}
               />
             </Box>
           ));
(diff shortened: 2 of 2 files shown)


topocount (approved):
This is a fine addition to the current rendering logic, but I think we should step back and rethink how we're rendering this list of ledger events in the context of performance. Rendering this view with any production-sized instance blocks the whole frontend for 10 seconds or more. 

I'm thinking we should move towards react-virtualized, potentially, but don't have strong opinions yet

hammadj:
> This is a fine addition to the current rendering logic, but I think we should step back and rethink how we're rendering this list of ledger events in the context of performance. Rendering this view with any production-sized instance blocks the whole frontend for 10 seconds or more.
> 
> I'm thinking we should move towards react-virtualized, potentially, but don't have strong opinions yet

Yea, going to paginate it in another PR

decentralion (commented):
This is how it looks in an instance with a larger number of recipients per distribution:

image

Personally, I don't love the pop-up modal, in part b.c. it feels it'll be hard to extend with other features, e.g. a way to filter in the recipient list. But I also agree that having this (as-implemented) is a big improvement over having nothing. 

(Not approving, not blocking, just providing feedback.)

topocount:
@hammadj how do you feel about an "accordion"-type view? I know the component exists for mui tables

hammadj:
> @hammadj how do you feel about an "accordion"-type view? I know the component exists for mui tables

I tried that first, but it had performance issues since every row needed to have the accordion and handle open / close state which was tricky. The modal moves things outside the list hierarchy.

Might be able to optimize it better once its paginated.

> Personally, I don't love the pop-up modal, in part b.c. it feels it'll be hard to extend with other features, e.g. a way to filter in the recipient list.
@decentralion Having a filter button in the modal wouldn't be an issue. I think this actually makes it easier since we can have context specific UI show up depending on what the event is instead of having to cram everything inside the main table view.

decentralion:
@KuraFire do you want to offer any thoughts on this from a design perspective?

decentralion:
NB: I def think this is an improvement over not having it, so happy for it to merge, even w/o review from @KuraFire -- ex-post-facto reviews always welcome and if Faruk has thoughts on where the design should go, we can fold it into the next iteration.

KuraFire:
@decentralion @hammadj I'm excited to see this! Responding here post-merge for in-context purposes:

1. I'm not a fan of heavy-use accordion experiences; modals are usable to a point, depending on how frequently one encounters them in a common usage scenario. They also kind of limit UI features, and perhaps most relevant for us, modals prevent alternation, i.e. clicking on one source to see the contents, and then clicking on another to compare the contents of that one.
2. What about a right pane that works much like GitHub's pane on the Projects board, showing contextual information about selected issues but in our case, giving you a pane with the table of grain distributions as well as an optional filter field?

E.g.:
github right pane sourcecred ledger

PR #2365, opened 2020-10-12
MPG: Expose node and edge order

decentralion:
This commit updates the MarkovProcessGraph so that it exposes a
canonical node and edge order. On its face, this is a nice change
because it cleans up the API in CredGraph (we no longer poke inside the
JSON representation in order to get a canonical node order). However,
this has an added benefit. It allows us to re-write the `nodes` and
`edges` accessors so that they just iterate over the order,
filtering if appropriate, and then call the `node` or `edge` accessor
for each element in the order.

This is nice because it makes it cleaner to apply future refactors which
"virtualize the gadgets", i.e. stop representing each gadget node and
edge directly in maps (which is wasteful of memory/disk). Instead, we
can just make sure that the virtualized gadgets are included in the
order, and that the node/edge accessors can retrieve virtualized gadget
nodes and edges.

Test plan: Some unit tests added, `yarn test` passes, and I also re-ran
CredRank on the real SC instance and verified the scores haven't
changed.

Thanks to @wchargin for significant improvements to this commit.

### DIFF
--- a/src/core/markovProcessGraph.js
+++ b/src/core/markovProcessGraph.js
@@ -56,7 +56,6 @@ import {type Uuid} from "../util/uuid";
  */
 
 import sortedIndex from "lodash.sortedindex";
-import sortBy from "../util/sortBy";
 import {makeAddressModule, type AddressModule} from "./address";
 import {
   type NodeAddressT,
@@ -308,7 +307,7 @@ export type IndexedMarkovEdge = {|
   +transitionProbability: TransitionProbability,
 |};
 export type MarkovProcessGraphJSON = Compatible<{|
-  +sortedNodes: $ReadOnlyArray<MarkovNode>,
+  +nodes: $ReadOnlyArray<MarkovNode>,
   +indexedEdges: $ReadOnlyArray<IndexedMarkovEdge>,
   +participants: $ReadOnlyArray<Participant>,
   // The -Infinity and +Infinity epoch boundaries must be stripped before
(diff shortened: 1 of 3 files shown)


src/core/markovProcessGraph.js
```diff
@@ -644,33 +644,64 @@ export class MarkovProcessGraph {
...
   }
 
+  /**
+   * Returns a canonical ordering of the nodes in the graph.
+   *
+   * No assumptions should be made about the node order, other than
+   * that it is stable for any given MarkovProcessGraph.
+   */
+  nodeOrder(): $ReadOnlyArray<NodeAddressT> {
+    return Array.from(this._nodes.keys()).sort();
```
wchargin, line 654: Can we make `this._nodes`’s insertion order stable and just `yield*
this._nodes.keys()`, returning an iterator with constant time and
constant per-element advancement rather than linearithmic upfront time?

src/core/markovProcessGraph.js
```diff
@@ -644,33 +644,64 @@ export class MarkovProcessGraph {
...
+   * The nodes are always iterated over in the node order.
+   */
   *nodes(options?: {|+prefix: NodeAddressT|}): Iterator<MarkovNode> {
     const prefix = options ? options.prefix : NodeAddress.empty;
-    for (const node of this._nodes.values()) {
-      if (NodeAddress.hasPrefix(node.address, prefix)) {
-        yield node;
+    for (const address of this.nodeOrder()) {
+      if (NodeAddress.hasPrefix(address, prefix)) {
+        yield NullUtil.get(this.node(address));
```
wchargin, line 672: Follow-up to above: just iterate over `this._nodes` here and pull out
`[address, value]` rather than forcing another `this.node` lookup and
`NullUtil.get` (the lookup is more important, since it hits memory).

src/core/markovProcessGraph.js
```diff
@@ -679,21 +710,24 @@ export class MarkovProcessGraph {
   }
 
   toMarkovChain(): OrderedSparseMarkovChain {
-    const nodeOrder = Array.from(this._nodes.keys()).sort();
+    // Array-ify the iterators so we can iterate over them multiple times
+    // without needing to re-generate them.
```
wchargin, line 714: No need: just iterate over `this._nodes`/`this._edges` as much as you
need. You’re inside the abstraction boundary.

wchargin (approved):

PR #2366, opened 2020-10-12
Reorganize CredRank-related modules

decentralion:
This commit moves the core CredRank modules into a folder called
core/credrank. The motivation is that I want to start splitting apart
MarkovProcessGraph into smaller modules, and doing so without moving the
files would result in a disorganized core directory.


wchargin (approved):
illustrious

PR #2368, opened 2020-10-12
credrank: factor out node gadgets

decentralion:
This adds a new module, nodeGadgets.js, which provides a consistent
interface for handling node gadgets. This will be quite useful when we
start "virtualizing" the gadget nodes. Also, this reduces the bloat of
the markovProcessGraph module, and allows a separation of concerns
between the core logic of markovProcessGraph, and the incidental logic
of having representation strategies for gadget addresses and nodes.

I also did a few slight tweaks, e.g. renaming the "USER_EPOCH" to 
"PARTICIPANT_EPOCH" and changing the descriptions for those nodes.

Test plan: tests pass, also credrank results are stable.


src/core/credrank/nodeGadgets.js
```diff
@@ -0,0 +1,104 @@
...
+  fromRaw(NodeAddressT): T;
+  // Generate a full node for this gadget from the structured address.
+  node: (T) => MarkovNode;
+}
+
+export const seedGadget: NodeGadget<void> = (() => {
+  const description: string = "\u{1f331}"; // U+1F331 SEEDLING
+  const prefix = NodeAddress.append(CORE_NODE_PREFIX, "SEED");
+  const toRaw = () => prefix;
+  const fromRaw = () => {};
```
wchargin, line 32: Per spec, should error if the (unbound) argument is not `prefix`?

src/core/credrank/nodeGadgets.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+  function fromRaw(addr) {
+    if (!NodeAddress.hasPrefix(addr, this.prefix)) {
+      throw new Error(
+        "Not an epoch node address: " + NodeAddress.toString(addr)
+      );
+    }
+    const prefixLength = NodeAddress.toParts(prefix).length;
+    const parts = NodeAddress.toParts(addr);
+    const epochStart = +parts[prefixLength];
```
wchargin, line 53: To be precise here, you would check that `String(epochStart)` matches
`parts[prefixLength]` and fail otherwise. As written, I can give an
address whose last part is `2.00000`, which will parse just the same as
if it were `2`, even though that address did not come from this gadget.

That is, you want to draw a bijection between `T` and a subobject of
`NodeAddressT`.

Similarly, fail if `parts.length` is too long.

src/core/credrank/nodeGadgets.js
```diff
@@ -0,0 +1,104 @@
...
+  function toRaw(addr) {
+    return NodeAddress.append(this.prefix, String(addr.epochStart));
+  }
+  function fromRaw(addr) {
+    if (!NodeAddress.hasPrefix(addr, this.prefix)) {
+      throw new Error(
+        "Not an epoch node address: " + NodeAddress.toString(addr)
+      );
+    }
+    const prefixLength = NodeAddress.toParts(prefix).length;
```
wchargin, line 51: Can we define this outside `fromRaw`, right after `prefix`? I think it’s
just as clear, and no need to recompute it every time.

src/core/credrank/nodeGadgets.js
```diff
@@ -0,0 +1,104 @@
...
+    return NodeAddress.append(
+      prefix,
+      String(addr.epochStart),
+      ...NodeAddress.toParts(addr.owner)
+    );
+  }
+  function fromRaw(addr: NodeAddressT): ParticipantEpochAddress {
+    if (!NodeAddress.hasPrefix(addr, prefix)) {
+      throw new Error(
+        "Not an epoch node address: " + NodeAddress.toString(addr)
```
wchargin, line 84: Make error message not same as `EpochAccumulatorAddress.fromRaw`’s?

src/core/credrank/nodeGadgets.js
```diff
@@ -0,0 +1,104 @@
...
+  return Object.freeze({prefix, toRaw, fromRaw, node});
+})();
+
+export type EpochAccumulatorAddress = {|
+  +epochStart: TimestampMs,
+|};
+export const accumulatorGadget: NodeGadget<EpochAccumulatorAddress> = (() => {
+  const prefix = NodeAddress.append(CORE_NODE_PREFIX, "EPOCH_ACCUMULATOR");
+  function toRaw(addr) {
+    return NodeAddress.append(this.prefix, String(addr.epochStart));
```
wchargin, line 43: Here and below: let’s just use the names from the closure rather than
accessing via `this`. (So, `this.prefix` → `prefix`, `this.toRaw` →
`toRaw`, etc.) Reduces potential for binding errors due to bizarrities
of `this` semantics.

wchargin (approved):
Review is mostly for the `NodeGadget` structure; I skimmed the fallout
changes and they seem reasonable.

src/core/credrank/nodeGadgets.js
decentralion: Done

PR #2379, opened 2020-10-15
Create Ethereum plugin

topocount:
This plugin will enable authentication with Ethereum addresses as aliases. There is still a lot to implement,  but this lays out the logic for creating Identity proposals from the ethereumAddresses.json file.

I originally wanted to use a compatible header for this store, which would be represented as a JsonLog, but ran into issues preserving a readable/git diff-able format when serializing the file. Discussions about how to implement versioning with JsonLog entries are ongoing and will be for a future PR.

The next thing I'm keen to implement is the class representation of this information in memory, to be used by the frontend to add and read entries.

test plan: see individual commits for test plans.


src/plugins/ethereum/ethAddresses.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
+import type {EthAddress} from "./ethAddressFile";
+import {nodePrefix} from "./declaration";
+
+//type KeyEntries = {|[address: EthAddress]: boolean|};
```
decentralion, line 7: Vestigial?

src/plugins/ethereum/ethAddresses.js
```diff
@@ -0,0 +1,13 @@
...
+
+export function nodeAddressForEthAddress(address: EthAddress): NodeAddressT {
+  return NodeAddress.append(nodePrefix, address);
+}
+
+//function createKeyEntries(): KeyEntries {}
```
decentralion, line 13: vestigial?

src/plugins/ethereum/ethAddressFile.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+
+import {isAddress, toChecksumAddress} from "web3-utils";
+
+import {type NodeAddressT, NodeAddress} from "../../core/graph";
+import * as C from "../../util/combo";
+import {compatibleParser} from "../../util/compat";
+import {nodePrefix} from "./declaration";
+
+//import type {Name} from "../../core/identity";
```
decentralion, line 10: vestigial?

decentralion (commented):
Feels like there's four separate atomic changes here:
1. add deps
2. add eth address logic
3. skeleton of ethereum plugin
4. integrate into the test instance and update snapshots

would prefer that we PR individually atomic commits, it tends to result in higher quality reviews

PR #2384, opened 2020-10-15
Virtualize the seed node

decentralion:
This starts the work of virtualizing nodes in the MarkovProcessGraph,
beginning with the seed node. This has a negligible impact on the output
graph size (we only remove one node), but this lets us focus instead on
the new mechanics for supporting virtualization.

Test plan: Tests pass, credrank results are stable, and using the
following `jq` expression before and after the change reveals that the
number of sortedNodes has decreased by exactly one:

```
jq <../cred/output/credGraph.json '.[1].mpg | .[1].sortedNodes | length'
```

### DIFF
--- a/src/core/credrank/markovProcessGraph.js
+++ b/src/core/credrank/markovProcessGraph.js
@@ -168,7 +168,9 @@ export class MarkovProcessGraph {
     this._epochBoundaries = deepFreeze(epochBoundaries);
     this._participants = deepFreeze(participants);
     // Precompute the index maps
-    this._nodeIndex = new Map([...nodes.keys()].map((a, i) => [a, i]));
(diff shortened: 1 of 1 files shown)


src/core/credrank/markovProcessGraph.js
```diff
@@ -398,8 +396,17 @@ export class MarkovProcessGraph {
       }
     }
 
+    function* realAndVirtualNodes(): Iterator<MarkovNode> {
+      for (const node of _nodes.values()) {
+        yield node;
+      }
```
wchargin, line 402: `yield* _nodes.values()` (etc. below)

wchargin (approved):
This structure looks good to me.

PR #2388, opened 2020-10-15
Virtualize the fibration edges

decentralion:
This commit marks the first batch of edge virtualization. I decided to
virtualize all of the fibration edges (payout, forward webbing, and
backward webbing) in one step since they are all so similar.

Now this is space saving! We're down to 57% of original.

| Step | Virtualize | # Nodes | # Edges | Disk Space | % of full |
| ---- | ---------- | ------- | ------- | ---------- | -------- |
| 0 | None | 96144 | 416365 | 116M | 100% |
| 1 | Seed | 96143 | 416365 | 116M | 100% |
| 2 | Accumulators | 96001 | 416365 | 116M | 100% |
| 3 | Epochs | 30823 | 416365 | 102M | 88% |
| 4 | Fibrations | 30823 | 220833 | 66M | 57% |

Warning: When running this variant, I sense a slight disturbance in the
Cred. We should track it down and understand what's causing it.

### DIFF
--- a/src/core/credrank/markovProcessGraph.js
+++ b/src/core/credrank/markovProcessGraph.js
@@ -178,7 +178,12 @@ export class MarkovProcessGraph {
         ...virtualizedNodeAddresses(epochBoundaries, participants),
       ].map((a, i) => [a, i])
     );
-    this._edgeIndex = new Map([...edges.keys()].map((a, i) => [a, i]));
+    this._edgeIndex = new Map(
+      [
+        ...edges.keys(),
+        ...virtualizedMarkovEdgeAddresses(epochBoundaries, participants),
+      ].map((a, i) => [a, i])
+    );
   }
 
   static new(args: Arguments): MarkovProcessGraph {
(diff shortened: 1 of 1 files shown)


src/core/credrank/markovProcessGraph.js
```diff
@@ -270,22 +274,22 @@ export class MarkovProcessGraph {
...
             forwardWebbingGadget.markovEdge(webbingAddress, gammaForward)
           );
-          addEdge(
+          recordTransitionProbability(
             backwardWebbingGadget.markovEdge(webbingAddress, gammaBackward)
           );
         }
-        lastBoundary = boundary;
       }
+      lastBoundary = boundary;
```
decentralion, line 292: Note to reviewer (and to self): I believe this `lastBoundary` assignment was in the wrong loop, and this was causing the wrong fibration edges to get written. I believe that is why the Cred shifted slightly in this commit. For cleanliness and sanity checking, let's pull this change out into a preceding commit before merging.

src/core/credrank/markovProcessGraph.test.js
```diff
@@ -380,6 +393,12 @@ describe("core/credrank/markovProcessGraph", () => {
     });
   });
 
+  describe("toMarkovChain", () => {
+    it("does not violate its own invariant checker", () => {
+      markovProcessGraph().toMarkovChain();
+    });
+  });
```
decentralion, line 400: Added this because this failed for a bad implementation, and having it is helpful for debugging.

src/core/credrank/markovProcessGraph.test.js
```diff
@@ -29,18 +29,30 @@ describe("core/credrank/markovProcessGraph", () => {
...
+    address: na("participant2"),
+    timestampMs: null,
+  };
+  const id2 = uuid.fromString("YVZhbGlkVXVpZE20TGFzdA");
+  const participant2 = {
+    description: participantNode2.description,
+    address: participantNode2.address,
+    id: id2,
+  };
+
```
decentralion, line 55: Added a second participant because during implementation I found a bug that only repro'd with multiple participants (had to do with the whether an assignment after all participants or in between participants--see above). I believe adding this makes the test suite generally more effective, it also means we have a participant with no contributions, which is realistic.

src/core/credrank/markovProcessGraph.js
decentralion: Done, see #2402.

PR #2389, opened 2020-10-15
Add pagination to LedgerViewer

hammadj:
The LedgerViewer is now much more performant with a large number of events. The pagination
controls at the bottom allow users to go back and forth between pages of ledger events.
They will also be able to select between 25, 50, or 100 rows per page (default 25).
Tested with 2600+ ledger events with snappy performance.
UI for mobile also cleaned up to be much more usable.

Test Plan:
- On the LedgerViewer page, click the right arrow to go go the next page of the ledger events and ensure
they are in the correct order.
- Change the number of Rows to 50 and 100 and ensure the ledger events are shown in the correct order.

### DIFF
--- a/src/ui/components/LedgerViewer.js
+++ b/src/ui/components/LedgerViewer.js
@@ -35,7 +35,6 @@ import * as G from "../../core/ledger/grain";
 const useStyles = makeStyles((theme) => {
   return {
     container: {
-      maxWidth: "60em",
       width: "100%",
     },
     table: {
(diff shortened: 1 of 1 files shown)


topocount (commented):
This is looking nice! My only observation is that there seems to be a lot of empty space in landscape mode. Since most other views only work well in landscape mode, I think it makes sense to at least allow the Details column more width:
image

befitsandpiper (approved):

topocount (approved):
thanks!

PR #2390, opened 2020-10-15
Add roleWeights to discord plugin

befitsandpiper:
add to config, set user reaction weights based on highest role weight they have.
Also create a default weight property for users without specified roles.

Test Plan: all tests pass, ran on 1hive instance and sanity checked cred distribution

### DIFF
--- a/src/plugins/experimental-discord/config.js
+++ b/src/plugins/experimental-discord/config.js
@@ -2,7 +2,7 @@
 
 import * as Combo from "../../util/combo";
 import * as Model from "./models";
-import {type EmojiWeightMap, type RoleWeightMap} from "./createGraph";
+import {type EmojiWeightMap, type RoleWeightConfig} from "./createGraph";
 
 export type {BotToken as DiscordToken} from "./models";
 
@@ -30,7 +30,7 @@ export type DiscordConfig = {|
   // }
   // Note that roles have a snowflake identifier.
   // default is used to set weights for members who don't have a specified role
-  +roleWeights: RoleWeightMap,
+  +roleWeightConfig: RoleWeightConfig,
 |};
 
 export const parser: Combo.Parser<DiscordConfig> = (() => {
@@ -38,6 +38,9 @@ export const parser: Combo.Parser<DiscordConfig> = (() => {
   return C.object({
     guildId: C.string,
     reactionWeights: C.dict(C.number),
-    roleWeights: C.dict(C.number),
+    roleWeightConfig: C.object({
+      defaultWeight: C.number,
+      roleWeights: C.dict(C.number),
+    }),
   });
 })();
--- a/src/plugins/experimental-discord/config.js
+++ b/src/plugins/experimental-discord/config.js
@@ -24,12 +24,15 @@ export type DiscordConfig = {|
   +reactionWeights: EmojiWeightMap,
   // An object mapping a role to a weight, as in:
   // {
-  //   "default": 0,
-  //   "core:626763367893303303": 2,
-  //   "contributor:456763457893303303": 1,
+  //   "defaultWeight": 0,
+  //   "roleWeights": {
+  //     "759191073943191613": 0.5,
+  //     "762085832181153872": 1,
+  //     "698296035889381403": 1
+  //   }
   // }
-  // Note that roles have a snowflake identifier.
-  // default is used to set weights for members who don't have a specified role
+  // Note that roles use a snowflake id only.
+  // defaultWeight is used to set weights for members who don't have a specified role
   +roleWeightConfig: RoleWeightConfig,
 |};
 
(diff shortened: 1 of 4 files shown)


src/plugins/experimental-discord/config.js
```diff
@@ -22,12 +22,22 @@ export type DiscordConfig = {|
...
   +reactionWeights: EmojiWeightMap,
+  // An object mapping a role to a weight, as in:
+  // {
+  //   "default": 0,
+  //   "core:626763367893303303": 2,
+  //   "contributor:456763457893303303": 1,
+  // }
+  // Note that roles have a snowflake identifier.
+  // default is used to set weights for members who don't have a specified role
+  +roleWeights: RoleWeightMap,
```
decentralion, line 33: Suggest having the default be an explicit property, rather than an entry in the map.

```js
type RoleWeightConfig = {|
  +defaultWeight: number,
  +roleWeights: $ReadOnlyMap<DiscordSnowflake, number>
|}
```

This is a lot clearer and easier to reason about than having the string key "default" play a special role.

decentralion (commented):
I'd be comfortable merging this without explicit tests b.c. it matches where the rest of the experimental discord plugin is at.

src/plugins/experimental-discord/config.js
```diff
@@ -22,12 +22,25 @@ export type DiscordConfig = {|
   // You can get a custom emoji ID by right clicking the custom emoji and
   // copying it's URL, with the ID being the image file name.
   +reactionWeights: EmojiWeightMap,
+  // An object mapping a role to a weight, as in:
+  // {
+  //   "default": 0,
+  //   "core:626763367893303303": 2,
+  //   "contributor:456763457893303303": 1,
+  // }
```
decentralion, line 30: Comment out of date

src/plugins/experimental-discord/createGraph.js
```diff
@@ -212,18 +219,28 @@ export function createGraph(
       const reactions = repo.reactions(channel.id, message.id);
       for (const reaction of reactions) {
         const emojiRef = Model.emojiToRef(reaction.emoji);
-        const reactionWeight = emojiWeights[emojiRef];
+        let reactionWeight = emojiWeights[emojiRef];
+        if (reactionWeight === null || reactionWeight === undefined) {
```
decentralion, line 223: `reactionWeight === null || reactionWeight === undefined` is equivalent to `reactionWeight == null`. It's the one good time to use `==` rather than `===` in js.

src/plugins/experimental-discord/createGraph.js
```diff
@@ -212,18 +219,28 @@ export function createGraph(
...
-        const node = reactionNode(reaction, message.timestampMs, guild);
-        if (reactionWeight != null) {
-          wg.weights.nodeWeights.set(node.address, reactionWeight);
+        // get the weight of the highest weight role the reacting user has
+        let roleWeight = roleWeightConfig.defaultWeight;
+        const roleWeights = roleWeightConfig.roleWeights;
+        for (const roleRef of reactingMember.roles) {
+          if (roleWeights[roleRef] > roleWeight) {
+            roleWeight = roleWeights[roleRef];
+          }
```
decentralion, line 239: `roleWeights[roleRef]` may be undefined, so we're doing `undefined > number` comparisons.  this works but it's sketchy. let's do something like:

```js
const matchingWeight = roleWeights[roleRef];
if (matchingWeight != null && matchingWeight > roleWeight) {
  roleWeight = matchingWeight;
}
```

src/plugins/experimental-discord/createGraph.js
```diff
@@ -212,18 +219,28 @@ export function createGraph(
       const reactions = repo.reactions(channel.id, message.id);
       for (const reaction of reactions) {
         const emojiRef = Model.emojiToRef(reaction.emoji);
-        const reactionWeight = emojiWeights[emojiRef];
+        let reactionWeight = emojiWeights[emojiRef];
```
decentralion, line 222: cleaner: `const reactionWeight = nullUtil.orElse(emojiWeights[emojiRef], 1)`

src/plugins/experimental-discord/createGraph.js
befitsandpiper: Do we really want to use the soft equality like this? I'll change it though.

decentralion: Yeah, we use soft equality this way throughout the codebase. `git grep "== null" | wc -l` has 160 hits

decentralion (approved):
This looks good from an implementation standpoint. However, one usability issue I'm concerned about: as setup, any existing discord plugin will need to specify the roleWeightConfig or else get a parse error. How about having it defualt to `{defaultWeight: 1, roleWeights: {}}`?

crisog:
> This looks good from an implementation standpoint. However, one usability issue I'm concerned about: as setup, any existing discord plugin will need to specify the roleWeightConfig or else get a parse error. How about having it default to `{defaultWeight: 1, roleWeights: {}}`?

This can probably be done in a better way, but what comes to my mind right now is:

Add a new optional parameter to support default fields in combo.js.

`export const object: PObject = (function object(
  requiredFields,
  optionalFields?,
  defaultFields?
)`

topocount (approved):
👍

PR #2404, opened 2020-10-17
credrank: virtualize radiation edges

decentralion:
This commit applies the virtualization strategy to radiation edges. This
is more complicated because we have a radiation edge for every (real and
virtual) node in the graph, with the exception of the seed node, and we
cannot easily derive any edge's transition probability without knowing a
lot about the structure of the graph.

Therefore:
- the virtualized edge iterator needs to know the MPG node order, so as
to generate every radiation edge
- we store the transition probability for every single radiation edge
(can try optimizing this more later)

Even with these limitations, we still get some solid space saving:

| Step | Virtualize | # Nodes | # Edges | Disk Space | % of full |
| ---- | ---------- | ------- | ------- | ---------- | -------- |
| 0 | None | 96144 | 416365 | 116M | 100% |
| 1 | Seed | 96143 | 416365 | 116M | 100% |
| 2 | Accumulators | 96001 | 416365 | 116M | 100% |
| 3 | Epochs | 30823 | 416365 | 102M | 88% |
| 4 | Fibrations | 30823 | 220833 | 66M | 57% |
| 5 | Radiation | 30823 | 124690 | 45M | 39% |

Test plan: Unit tests pass, CredRank scores are stable.

### DIFF
--- a/src/core/credrank/markovProcessGraph.js
+++ b/src/core/credrank/markovProcessGraph.js
@@ -146,6 +146,7 @@ export type MarkovProcessGraphJSON = Compatible<{|
   // JSON serialization.
   +finiteEpochBoundaries: $ReadOnlyArray<number>,
   +parameters: Parameters,
+  +radiationTransitionProbabilities: $ReadOnlyArray<number>,
 |}>;
 
 export class MarkovProcessGraph {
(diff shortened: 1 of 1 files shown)


wchargin (approved):
Clean.

src/core/credrank/markovProcessGraph.js
```diff
@@ -728,5 +776,11 @@ function virtualizedMarkovEdge(
       parameters.gammaBackward
     );
   }
+  if (MarkovEdgeAddress.hasPrefix(address, radiationGadget.prefix)) {
+    const nodeAddress = radiationGadget.fromRaw(address);
+    const index = NullUtil.get(nodeIndex.get(nodeAddress));
+    const probability = radiationTransitionProbabilities[index];
+    return radiationGadget.markovEdge(nodeAddress, probability);
+  }
```
decentralion, line 784: Note to self: radiation edges are most common, so we get a slight perf optimization if we check for them first.

PR #2410, opened 2020-10-19
Discord Asymptotic Reaction Weights

befitsandpiper:
This PR adds a flag to the discord config that switches discord reaction weights from being linear on the number of reacts to having an exponential decay making the total cred one member can mint on a single message roughly approach 2x the highest weight emoji, asymptotically.

The asymptoticReactionWeights function first separates reactions on a message by author. After this the reactions for a member are sorted from highest to lowest reaction weight. The function then calls a modified form of the old reaction weighting loop in 
the sharedReactionWeights function. Here there is a conditional to apply a weight dropoff, reducing each successive emoji weight by powers of 2. In the case where asymptotic weights are not applied, the sharedReactionWeights function is called directly, which has the old default behavior.

The motivation behind this feature is to reduce the effectiveness of discord members trying to cheat the cred system through light sybilling and/or long term small group collusion spamming huge numbers of emojis on each other's messages.

Test Plan: Tests pass locally, no errors when running on 1hive instance, sanity check cred against 1hive instance

### DIFF
--- a/src/plugins/experimental-discord/createGraph.js
+++ b/src/plugins/experimental-discord/createGraph.js
@@ -187,7 +187,12 @@ function mentionsEdge(message: Model.Message, member: Model.GuildMember): Edge {
   };
 }
 
-function sharedReactionWeights({
+/* This function creates edges and nodes for discord reactions.
+
+By default weights for all reactions are linear. Setting the 
+useAsymptoticReactionWeights flag makes the weights of multiple reactions
+given by the same member dropoff exponentially by powers of 2 */
+function createReactionEdgesAndNodes({
   guild,
   emojiWeights,
   roleWeightConfig,
@@ -195,16 +200,15 @@ function sharedReactionWeights({
   memberMap,
   message,
   reactions,
+  hasEdges,
   useAsymptoticReactionWeights,
-}) {
-  let hasEdges = false;
-
-  for (const [ind, reaction] of reactions.entries()) {
+}): boolean {
+  for (const [index, reaction] of reactions.entries()) {
     const emojiRef = Model.emojiToRef(reaction.emoji);
     let reactionWeight = NullUtil.orElse(emojiWeights[emojiRef], 1);
     if (useAsymptoticReactionWeights) {
       // drop off reactionWeight by powers of 2
-      reactionWeight *= 0.5 ** ind;
+      reactionWeight *= 0.5 ** index;
     }
 
     const reactingMember = memberMap.get(reaction.authorId);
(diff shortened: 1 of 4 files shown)


src/plugins/experimental-discord/createGraph.js
```diff
@@ -187,6 +187,94 @@ function mentionsEdge(message: Model.Message, member: Model.GuildMember): Edge {
   };
 }
 
+function sharedReactionWeights({
```
topocount, line 190: I'd name this something more clear like `createReactionEdgesAndNodes`

src/plugins/experimental-discord/createGraph.js
```diff
@@ -187,6 +187,94 @@ function mentionsEdge(message: Model.Message, member: Model.GuildMember): Edge {
...
+}) {
+  // sort reactions from highest to lowest weight emojis
+  const sortedReactions = [...reactions].sort((a, b) => {
+    const bEmojiWeight = emojiWeights[Model.emojiToRef(b.emoji)] || 1;
+    const aEmojiWeight = emojiWeights[Model.emojiToRef(a.emoji)] || 1;
+    return bEmojiWeight - aEmojiWeight;
+  });
+
+  // separate reactions into groups per member (retains sorted order)
+  const mapMemberIdToReactions = new Map();
```
topocount, line 257: To me, this is named more like function rather than a data structure. not a huge deal, just an observation.

src/plugins/experimental-discord/plugin.js
```diff
@@ -66,12 +66,18 @@ export class DiscordPlugin implements Plugin {
...
+      useAsymptoticReactionWeights,
+      reactionWeights,
+      roleWeightConfig,
+    } = await loadConfig(ctx);
     const repo = await repository(ctx, guildId);
     const declarationWeights = weightsForDeclaration(declaration);
     const defaultRoleWeightConfig = {defaultWeight: 1, roleWeights: {}};
     return await createGraph(
       guildId,
+      useAsymptoticReactionWeights || false,
```
topocount, line 80: Let's use NullUtil here too

src/plugins/experimental-discord/createGraph.js
```diff
@@ -187,6 +187,94 @@ function mentionsEdge(message: Model.Message, member: Model.GuildMember): Edge {
...
+  sortedReactions.forEach((reaction) => {
+    const authorData = mapMemberIdToReactions.get(reaction.authorId) || [];
+    authorData.push(reaction);
+    mapMemberIdToReactions.set(reaction.authorId, authorData);
+  });
+
+  for (const memberReactions of mapMemberIdToReactions.values()) {
+    return sharedReactionWeights({
```
topocount, line 265: This is going to return on the first iteration, right? Your probably don't want this since it'll remove reactions when more than one person reacts

topocount (commented):
These are some big, complext functions with a lot going  on. Let's add some docstrings above each declaration so others looking at this code don't need to parse them just to get a high-level understanding of whats happening.

src/plugins/experimental-discord/createGraph.js
befitsandpiper: Ah, yes... this should not be a return...

src/plugins/experimental-discord/createGraph.js
```diff
@@ -236,43 +240,49 @@ function sharedReactionWeights({
...
-      message,
+  for (const memberReactions of memberIdToReactionsMap.values()) {
+    hasEdges = createReactionEdgesAndNodes({
+      ...args,
+      hasEdges,
       reactions: memberReactions,
-      useAsymptoticReactionWeights,
     });
   }
+  return hasEdges;
```
topocount: Rather than continually setting `hasEdges` as true as each iteration runs, let's revert to initializing it inside `createReactionEdgesAndNodes`, and in  `prepareAsymptoticReactionWeights` use `hasEdges` or better yet something more accurately named like `atleastOneNodeHasedges` as a flag that is only flipped `true` without passing it through both of these functions.

topocount (approved):
thanks! excited to see this help mitigate gaming

decentralion:
Please hold off on merging this until I get a chance to review.

decentralion (changes requested):
So, I have two issues with this PR that makes me prefer it doesn't merge as-is:

1. the logic is getting quite complicated, and it's untested. at this point, I don't think we can add more stuff at this level of complexity without unit tests or it will just turn into a ball of mysterious spaghetti. https://github.com/sourcecred/sourcecred/pull/2434 sets up a system where we can start to meaningfully test reaction weight logic

2. i think it'd be better to have something like a per-week limit on how much reaction weight a user can generate, rather than a per-message limit (as implemented here). if the problem is an attacker is able to generate too much Cred by spamming reactions, having a limit per-message would just encourage them to spam a small number of reactions on a large number of messages. putting in a throttle on the total amount of cred minting the attacker can do per week would cover more cases.

befitsandpiper:
Lol, can I get some cred for trying? Alright, I'll close the PR

PR #2414, opened 2020-10-21
feature: add channel weights to discord plugin

crisog:
This PR builds on #2390. It adds the functionality to set weights to channels, along with a default property for the ones that are not weighted in the config.

I ran this plugin on the current 1Hive instance, sanity checked with @befitsandpiper and all tests are passing.


crisog:
Closed this because tests were not passing, but I was not aware that it would fail because of credentials.

PR #2418, opened 2020-10-21
Eth Plugin: Create the EthAddress type and helpers

topocount:
EthAddresses types represent ethereum addresses. They are stored in the ethereum plugin in a registry file.

This is one of a series of PRs that implements the snapshot-tested skeleton of the Ethereum address plugin:
- #2418 (this one)
- #2419: Scaffolding
- #2420: integrate plugin into instance snapshot tests

The immediate purpose of this plugin is to act as a backend store for provably owned ethereum addresses. 

**User story this work supports:** In self-service admin, users will be able to prove ownership of these addresses by submitting a message signed with the address' private key. The admin service will attempt to verify the signature, and if successful, the address will be added to the store and the identity in the identity asserted in the message will be created if it doesn't exist and possibly linked if prior ownership of the existing Identity has been established. I'm working in conjunction with @hammadj to implement this story.

For now, these addresses are intended to just be an identity alias for participants
primarily for the purpose of authenticating in a self-service frontend.

Under the hood, eth addresses are serialized into a jsonArray in the plugin config.
Serialized compat files don't preserve the readability of jsonLogs, so
there is little reason to utilize it now. Some compat serialization
logic is needed to preserve readability. This will be important to
implement before wider release in order to better accomodate readable
diffs in a UI and within Git.

test plan: unit tests are provided for the functions utilized by
parsers.

### DIFF
--- a/src/plugins/ethereum/ethAddress.js
+++ b/src/plugins/ethereum/ethAddress.js
@@ -1,0 +1,39 @@
+// @flow
+
+import {isAddress, toChecksumAddress} from "web3-utils";
+
+import * as C from "../../util/combo";
+import {compatibleParser} from "../../util/compat";
+
+export opaque type EthAddress: string = string;
+
+export function parseAddress(s: string): EthAddress {
+  if (!isAddress(s)) {
+    throw new Error(`not a valid ethereum address: ${s}`);
+  }
+  return toChecksumAddress(s);
+}
+
+// Utilized for a more readable and recognizable address. This is
+// needed to allow the address to fit within the Identity name requirements
+export function truncateEthAddress(address: EthAddress): string {
+  const prefix = address.slice(0, 6);
+  const suffix = address.slice(-4);
+  return `${prefix}...${suffix}`;
+}
+
+export const ethAddressParser: C.Parser<EthAddress> = C.fmap(
+  C.string,
+  parseAddress
+);
+
+export const COMPAT_INFO = {
+  type: "sourcecred/ethAddressFile",
+  version: "0.0.1",
+};
+
+const addressEntriesParser = C.array(ethAddressParser);
+
+export const parser: C.Parser<
+  Array<EthAddress>
+> = compatibleParser(COMPAT_INFO.type, {"0.0.1": addressEntriesParser});
--- a/src/plugins/ethereum/ethAddress.js
+++ b/src/plugins/ethereum/ethAddress.js
@@ -28,7 +28,7 @@ export const ethAddressParser: C.Parser<EthAddress> = C.fmap(
 );
 
 export const COMPAT_INFO = {
-  type: "sourcecred/ethAddressFile",
+  type: "sourcecred/ethAddress",
   version: "0.0.1",
 };
 
--- a/src/plugins/ethereum/ethAddress.js
+++ b/src/plugins/ethereum/ethAddress.js
@@ -7,6 +7,27 @@ import {compatibleParser} from "../../util/compat";
 
 export opaque type EthAddress: string = string;
 
+/**
+ * parseAddress will accept any 20-byte hexadecimal ethereum address encoded as
+ * a string, optionally prefixed with `0x`.
+ *
+ * Per EIP-55 (https://eips.ethereum.org/EIPS/eip-55), parseAddress throws if
+ * the provided string is mixed-case but not checksum-encoded. All valid
+ * addresses in lower- and upper-case format will not throw.
+ *
+ * For consistency, all valid addresses are converted and returned in
+ * mixed-case form with the `0x` prefix included
+ *
+ * valid formats:
+ * "2Ccc7cD913677553766873483ed9eEDdB77A0Bb0"
+ * "0x2Ccc7cD913677553766873483ed9eEDdB77A0Bb0"
+ * "0X2CCC7CD913677553766873483ED9EEDDB77A0BB0"
+ * "0x2ccc7cd913677553766873483ed9eeddb77a0bb0"
+ *
+ * invalid formats:
+ * "0x2ccc7cD913677553766873483ed9eEDdB77A0Bb0"
+ * "2ccc7cD913677553766873483ed9eEDdB77A0Bb0"
+ */
 export function parseAddress(s: string): EthAddress {
   if (!isAddress(s)) {
     throw new Error(`not a valid ethereum address: ${s}`);
--- a/src/plugins/ethereum/ethAddress.test.js
+++ b/src/plugins/ethereum/ethAddress.test.js
@@ -17,6 +17,10 @@ describe("plugins/ethereum/ethAddress", () => {
         "0x2Ccc7cD913677553766873483ed9eEDdB77A0Bb0",
         "0x2Ccc7cD913677553766873483ed9eEDdB77A0Bb0".toUpperCase(),
         "0x2Ccc7cD913677553766873483ed9eEDdB77A0Bb0".toLowerCase(),
+        // upper-case with lower-cased `0x`
+        "0x2CCC7CD913677553766873483ED9EEDDB77A0BB0",
+        // lower-case with upper-cased `0x`
+        "0X2ccc7cd913677553766873483ed9eeddb77a0bb0",
       ].forEach((a: string) => {
         expect(parseAddress(a)).toBe(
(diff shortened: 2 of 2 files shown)


src/plugins/ethereum/ethAddress.js
```diff
@@ -0,0 +1,39 @@
...
+  return `${prefix}...${suffix}`;
+}
+
+export const ethAddressParser: C.Parser<EthAddress> = C.fmap(
+  C.string,
+  parseAddress
+);
+
+export const COMPAT_INFO = {
+  type: "sourcecred/ethAddressFile",
```
decentralion, line 31: our compat types are generally data types, not file names. e.g. "sourcecred/graph" not "sourcecred/graphFile"

src/plugins/ethereum/ethAddress.js
```diff
@@ -0,0 +1,39 @@
...
+  version: "0.0.1",
+};
+
+const addressEntriesParser = C.array(ethAddressParser);
+
+export const parser: C.Parser<
+  Array<EthAddress>
+> = compatibleParser(COMPAT_INFO.type, {"0.0.1": addressEntriesParser});
```
decentralion, line 39: are we using JSON log, or a file with an array of addresses? Confused based on the PR description.

src/plugins/ethereum/ethAddress.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+  describe("compatible JsonLog", () => {
+    const address1 = "0x2Ccc7cD913677553766873483ed9eEDdB77A0Bb0";
+    const address2 = "0xb4124cEB3451635DAcedd11767f004d8a28c6eE7";
+    const address3 = "0x8401Eb5ff34cc943f096A32EF3d5113FEbE8D4Eb";
+    const addressArray = [address1, address2, address3];
+    it("can create an exported JsonLog instance", () => {
+      const compatibleEthLog = toCompat(COMPAT_INFO, addressArray);
+      const result = parser.parseOrThrow(compatibleEthLog).values();
+      expect(Array.from(result)).toEqual(addressArray);
```
decentralion, line 62: It mentions JsonLog but never uses it?

decentralion (commented):
> Serialized compat files don't preserve the readability of jsonLogs, so
there is little use in utilizing it now. Some compat serialization
logic is needed to preserve readability. This will be important to
implement before wider release in order to better accomodate readable
diffs in a UI and within Git.

I don't understand that comment. The implementation and inline comments and tests all seem contradictory (e.g. the test that says it uses json log, but never uses json log). What's going on?

topocount:
> > Serialized compat files don't preserve the readability of jsonLogs, so
> > there is little use in utilizing it now. Some compat serialization
> > logic is needed to preserve readability. This will be important to
> > implement before wider release in order to better accomodate readable
> > diffs in a UI and within Git.
> 
> I don't understand that comment. The implementation and inline comments and tests all seem contradictory (e.g. the test that says it uses json log, but never uses json log). What's going on?

@decentralion thanks for catching this. I've thought about it and decided that using JsonLog doesn't make sense in this case. JsonLog entries are each meant to be immutable, so having an inline `version` prop to determine how to parse each entry makes sense. However, classifying serialized ethAddress entries as immutable doesn't seem necessary to me. As the feature set of the Ethereum plugin grows, all entries should be able to be updated to contain additional data. The reason I was using JsonLog was for the ease of diffing the file. I want this "diff"-ability to be accessible in compat for use-cases where entries are not immutable, so for now I'm just using a simple array of strings within compat.

I'm open to discussing this further, but I think changing the shape of the the ethereum plugin's user data does not necessarily need to be auditable (at least right now), which I see as the best argument for using jsonLog. I can possibly be convinced otherwise, though.

src/plugins/ethereum/ethAddress.js
```diff
@@ -0,0 +1,39 @@
...
+export const COMPAT_INFO = {
+  type: "sourcecred/ethAddress",
+  version: "0.0.1",
+};
+
+const addressEntriesParser = C.array(ethAddressParser);
+
+export const parser: C.Parser<
+  Array<EthAddress>
+> = compatibleParser(COMPAT_INFO.type, {"0.0.1": addressEntriesParser});
```
hammadj, line 39: What exactly is the purpose of this parser? Why are we storing ETH addresses in an array like this separate from the ledger? Is it supposed to be a whitelist or something?

Curious about the value this data structure provides

src/plugins/ethereum/ethAddress.js
wchargin: File with array of addresses: see #2420. I like this format because it
keeps the `Compatible` structure without paying a 50-byte penalty on
every 40-byte line, and also keeps the format more easily extensible to
other kinds of configuration, which seem plausibly desirable.

Do clarify PR description

src/plugins/ethereum/ethAddress.js
wchargin: On behalf of @topocount (correct me if mistaken): this plugin loads data
entirely from its config. Contrast:

  - The GitHub/Discourse/Discord plugins read their config to determine
    a server/repo to hit, and load data by querying it.
  - The Ethereum plugin reads its config to determine addresses, and
    loads data by simply emitting that set of addresses plus node and
    edge structure derived from it.

The PR description implies this in the first two paragraphs, I think. In
general I think this kind of thing is nice to have a brief design doc
for: say, a Discourse post describing the structure and purpose of this
plugin.

src/plugins/ethereum/ethAddress.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+import {isAddress, toChecksumAddress} from "web3-utils";
+
+import * as C from "../../util/combo";
+import {compatibleParser} from "../../util/compat";
+
+export opaque type EthAddress: string = string;
+
+export function parseAddress(s: string): EthAddress {
```
wchargin, line 10: This function has a contract. Could you document it? Specifically:

  - The input string should be a 40-byte hex digest of an Ethereum
    address, with optional `0x` prefix. (Is that `0x` case-sensitive?)
  - It may be all-lowercase or all-uppercase.
  - It may also be mixed-case, in which case it MUST match the checksum.

Maybe this is obvious to you, but Eth noobs like me have to decipher it
from the tests. It’s a testament to the tests that I can do this! But it
should also be a testament to the docs that I don’t have to. ;-)

src/plugins/ethereum/ethAddress.js
```diff
@@ -0,0 +1,39 @@
...
+  if (!isAddress(s)) {
+    throw new Error(`not a valid ethereum address: ${s}`);
+  }
+  return toChecksumAddress(s);
+}
+
+// Utilized for a more readable and recognizable address. This is
+// needed to allow the address to fit within the Identity name requirements
+export function truncateEthAddress(address: EthAddress): string {
+  const prefix = address.slice(0, 6);
```
wchargin, line 20: Curious: is this the standard abbreviation form? e.g., for PGP keys, you
generally just take the last 8, 16, or 32 characters of the fingerprint
depending on how long you want the name to be.

If it’s standard/recognizable, no problem.

wchargin (approved):
LGTM; please update PR description to describe current state (and, as
always, make sure that the merged commit message is the PR description).

src/plugins/ethereum/ethAddress.js
topocount: @hammadj check out the added detail in the PR description

src/plugins/ethereum/ethAddress.js
topocount: Good question. This is an unofficial convention for rendering addresses when screen real estate is at a premium (see [metamask](https://metamask.io/) and [1hive/aragon](https://apiary.1hive.org/profile?dao=0x12b013c23f1918eac4b5342117f0b202e048566c) UIs for examples). This pull request set doesn't necessarily require this function if we just bump up the maximum name length by a few characters, which I'm actually in favor of if @decentralion gives their blessing because that way the theoretical namespace for ethereum addresses increases signficantly in size.

Even if we do bump up the name character limit, I think this should be included as a convenience function, since it needs to be included in this file due to the opaque nature of the EthAddress type.

decentralion: I approved the bump-the-size PR.

decentralion (approved):

PR #2429, opened 2020-10-25
Implement PayableAddressStore for Ledger Accounts

topocount:
This is the first in a series of PRs that will implement single-click reward distributions for maintainers.

Note that this is functionally distinct from the ethereum plugin. That plugin is meant for users to authenticate via an EOA (externally owned account). The `PayableAddressStore` is tightly coupled to the ledger and allows users to configure the accounts they want grain paid out to.

Future work includes:
-  creating the config for the distribution and/or buyback currencies
- writing the dapp code that is capable of distributing grain payout and buybacks onchain
- writing the EVM event parsing code that verifies the distribution or buyback, and then writes the
appropriate activity to the Ledger.

PayableAddressStore maps addresses of EVM-based currencies to a participant's
address capable of accepting the currency. This structure exists to
accomodate safe migration for grain/payout token changes. Users must verify
themselves that the address they are supplying is capable of receiving their
share of a grain distribution.

An address may be deleted by passing in an empty string for the
`address` parameter. This is useful in case the underlying private key
is compromised or the exchange hosting a custodial account is hacked.

Merging Identities has no effect on the base Identity's configured payable addresses. The target
identity's address are dropped and all payouts after this point will target the base account's configured
address.

The CurrencyId objects are currently serialized before use in the Map.
This is due to the fact that Objects used as keys must be directly
accessible in order to retrieve the mapped value. A new object with
matching members is not equal to the object used as the key because each
has a distinct memory reference. By stringifying the object before
entering it as the key, deep equality can be observed and the
composition of the object become the keys, rather than the object's
memory reference.

test plan: Unit test are provided


src/core/ledger/ledger.js
```diff
@@ -575,6 +602,52 @@ export class Ledger {
...
+   * contract) and effectively lose that reward.
+   *
+   * An address may be deleted by passing in an empty string for the
+   * `address` parameter. This is useful in case the underlying private key
+   * is compromised or the exchange hosting a custodial account is hacked.
+   */
+  setPayoutAddress(
+    id: IdentityId,
+    currency: CurrencyAddress,
+    address: EthAddress | ""
```
decentralion, line 619: why pass in `""` as the delete param instead of `null`?

src/core/ledger/ledger.js
```diff
@@ -575,6 +602,52 @@ export class Ledger {
...
+      );
+    }
+    if (address === "") {
+      account.addresses.delete(currency);
+      return;
+    }
+    const addressResult = ethAddressParser.parse(address);
+    if (!addressResult.ok) {
+      throw new Error(`setPayoutAddress: invalid payout address: ${address}`);
+    }
```
decentralion, line 647: having a method throw an error after a mutation is scary in this module, since it may leave the ledger in an inconsistent state. please follow the convention from the other methods, and have a clear demarcation between the argument-checking logic and error handling (above the demarcation line) and any mutation (below the demarcation line).

(in this case, since the mutation that is above the error checking returns immediately, it's safe, but it's not obviously safe, and it's easy to imagine someone in the future making an innocent-looking change that actually makes the code unsafe.)

src/core/ledger/ledger.test.js
```diff
@@ -784,6 +789,52 @@ describe("core/ledger/ledger", () => {
...
+        "0xffffffffffffffffffffffffffffffffffffffff"
+      );
+      const ledger = new Ledger();
+      const id = ledger.createIdentity("USER", "Bob");
+      ledger.setPayoutAddress(id, ethAddress, fullAddress);
+      const account = ledger.accountByName("Bob");
+      if (!account) {
+        expect(account).not.toBeNull();
+        return;
+      }
```
decentralion, line 807: we don't re-use mutable state between test cases. it invites "spooky error at a distance" behavior where one test could pass while breaking a subsequent test. please refactor this so that each test has its own independent ledger instance, see other tests for examples.

decentralion (commented):
this solution seems to have good support for tracking multiple ERC-20s, which is nice. but it's super coupled to the EVM. what if we wanted to use it to represent FIL or BTC?

i'm wondering if we could have something as simple as having currencyAddress be an object that has a "chain identifier" (e.g. BTC, ETH, XDAI) and a "currency identifier". When the chain identifier points to "ETH" or "XDAI", then the currency identifier points to the right contract. If the chain identifier is BTC, then the currency identifier can be superfluous.

src/core/ledger/ledger.js
topocount: type-consistency but null makes sense

decentralion:
Removing myself from the review -- I think @hammadj is a better reviewer since he's more invested in this feature path

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  return id;
+}
+
+export const evmChainParser: C.Parser<EvmChainId> = C.fmap(
+  C.string,
+  parseEvmChainId
+);
+
+declare type EvmId = {
```
hammadj, line 32: Maybe `evmToken` would be a better name?

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+declare type ProtocolSymbol = string;
+
+/**
+ * Chains like Bitcoin and Filecoin do not have "production" sidechains so
+ * we represent them as a string, as specified in the ProtocolSymbol type
+ */
+declare type ProtocolId = {
+  chainId: ProtocolSymbol,
+  // currencyAddress is unused here
+  currencyAddress?: any,
```
hammadj, line 63: If its unused, why have it declared?

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+  {
+    chainId: C.exactly(["BTC", "FIL"]),
+  },
+  {
+    currencyAddress: C.pure<void>(),
+  }
+);
+
+export type ChainId = ProtocolSymbol | EvmChainId;
+export type CurrencyId = ProtocolId | EvmId;
```
hammadj, line 76: ProtocolSymbol being assigned to chainId is a bit confusing. Feel like it might be cleaner to have a single type for all tokens / currencies, something like:

```ts
declare type Currency = {
  symbol: string,
  chainId?: EvmChainId,
  tokenAddress?: EvmAddress
}
```

Where `chainId` and `tokenAddress` would be undefined for non-EVM currencies.

src/core/ledger/ledger.js
```diff
@@ -575,6 +602,58 @@ export class Ledger {
...
 
+  /**
+   * setPayoutAddress allows participants to set a payable address to collect
+   * grain. These addresses are keyed on a specific currency, which ensures that
+   * users don't erroneously receive a grain distribution to an address that
+   * cannot handle it (such as a custodial wallet, or rigidly-designed
+   * contract) and effectively lose that reward.
+   *
+   * An address may be deleted by passing in an empty string for the
+   * `address` parameter. This is useful in case the underlying private key
```
hammadj, line 613: you mention an `address` param but there isn't one, should be `payoutAddress`. Also, instead of an empty string should it be passing null since that's a valid type for that param?

src/core/ledger/ledger.js
```diff
@@ -575,6 +602,58 @@ export class Ledger {
...
+    payoutAddress: PayoutAddress | null,
+    chainId: ChainId,
+    currencyAddress: EthAddress | void
+  ): Ledger {
+    this._createAndProcessEvent({
+      type: "SET_PAYOUT_ADDRESS",
+      id,
+      payoutAddress,
+      currencyId: {chainId, currencyAddress},
+    });
```
hammadj, line 627: Would be cleaner to take in the `Currency` object type I described above instead, something like this:
```ts
  setPayoutAddress(
     id: IdentityId,
     currency: Currency,
     payoutAddress: PayoutAddress | null,
  ): Ledger {
    this._createAndProcessEvent({
      type: "SET_PAYOUT_ADDRESS",
      id,
      currency,
      payoutAddress,
    });
```

Also, this structure implies that different people would be able to get payouts in different tokens on different chains. Is there any use case where you think that might be relevant? Feels redundant to have the currency object in every ledger entry for setting a payout address.

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+
+// @topocount TODO: enable protocolIdParser once we support its address types
+export const currencyIdParser: C.Parser<CurrencyId> = evmIdParser;
```
hammadj, line 79: Likewise instead of `currencyIdParser` , it can just be `currencyParser`

src/core/ledger/ledger.js
```diff
@@ -43,9 +48,30 @@ type MutableAccount = {|
   // may not receive or transfer Grain. Accounts start inactive, and must
   // be explicitly activated.
   active: boolean,
+  // key-value store of addresses the account receives grain to
+  addresses: PayableAddressStore,
```
hammadj, line 52: Could be confused for nodeAddresses or something else. `payoutAddresses` would be a better name

src/core/ledger/ledger.test.js
```diff
@@ -784,6 +790,67 @@ describe("core/ledger/ledger", () => {
...
+        const ledger = ledgerWithIdentities();
+        ledger.setPayoutAddress(id1, fullAddress, protocolChainId);
+        const account = ledger.account(id1);
+        expect(account.addresses.get(protocolId.toString())).toBe(fullAddress);
+      });
+      xit("doesn't work with a non-EVM protocol Id", () => {
+        const ledger = ledgerWithIdentities();
+        const thunk = () => ledger.setPayoutAddress(id1, fullAddress, "BCH");
+        expect(thunk).toThrow("Invalid chainId or currencyAddress:");
+      });
```
hammadj, line 852: These two tests are skipped, any reason why or just forgot to un-skip?

src/core/ledger/ledger.test.js
```diff
@@ -784,6 +790,67 @@ describe("core/ledger/ledger", () => {
         expect(ledger.accountByName("target")).toEqual(null);
       });
     });
+
+    describe("setPayoutAddress", () => {
```
hammadj, line 794: Should also add a tests to ensure payoutAddresses are preserved after merging identities:
- identityA with address merged into identityB without address
- identityA without address merged into identityB with address
- identityA with address merged into identityB with address

hammadj (commented):
few naming / test things, overall looks good! this will help a ton

src/core/ledger/currency.js
topocount: @hammadj I considered this monolithic type but I thought is was more confusing for static typing. By splitting between these two subtypes, it's clearer in flow what type we're expecting and what sort of details it should possess.

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * CurrencyAddress is a subset of all available EthAddresses.
+   *
+   * A currency address is the address of the token contract for an ERC20 token,
+   * or the 20 byte-length equivalent of 0x0, which is the conventional address
+   * used to represent ETH on the ethereum mainnet, or the native currency on an
+   * EVM-based sidechain see here for more details on these semantics:
+   * https://ethereum.org/en/developers/docs/intro-to-ethereum/#eth
+   */
+  currencyAddress: EthAddress,
```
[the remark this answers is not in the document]
topocount: EthAddress is "more" correct since the base currency isn't technically a token

src/core/ledger/currency.js
topocount: Same explanation as above.

src/core/ledger/currency.js
topocount: This was removed but I never rebased this code

src/core/ledger/ledger.js
topocount: @hammadj I considered this but I feel like this is leaking implementation details. Consumers of this API to only need to pass in `ChainId`s and and and an optional ethereum address to generate a `CurrencyId` internally.

src/core/ledger/currency.js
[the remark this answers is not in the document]
hammadj: The vast majority of SC users would not be distributing Ether, they would be distributing some sort of ERC-20 token. Also, I was referring to the `currencyAddress` field and not `EthAddress` type.

src/core/ledger/currency.js
hammadj: Not sure I follow. `EvmId` implies that this type is an ID of some sort when in reality it's an object that contains information about some token and the chain it's running on. The EvmId name might be more appropriate if this type was a string in the form of `"evm:chainId:tokenAddress"` and coming out of a parser, but since its an object calling it an ID has a smell to it

src/core/ledger/currency.js
hammadj: Instead of doing the Union at the ID level, could just do something like this:

```ts

type EvmToken = {
  symbol: string,
  chainId: EvmChainId,
  tokenAddress: EvmAddress
}

type ProtocolToken = {
  symbol: string,
}

type Currency = EvmToken | ProtocolToken
```

This would also allow us to add additional fields scoped to different protocol tokens (e.g. if Polkadot has something like ERC20's on their network then we could have a tokenAddress field on ProtocolToken as well but enforce it to follow the semantics of its protocol instead of allowing a EvmChainId to be paired with a ProtocolId type)

src/core/ledger/ledger.js
hammadj: Wouldn't the `symbol` be a valuable thing to have in there as well, especially for rendering user-readable ledger history and other UI elements?

src/core/ledger/currency.js
```diff
@@ -82,3 +82,19 @@ export const currencyIdParser: C.Parser<CurrencyId> = C.orElse([
   evmIdParser,
   protocolIdParser,
 ]);
+
+export function buildCurrencyId(
+  chainId: string,
+  currencyAddress?: string
+): CurrencyId {
```
hammadj, line 89: Calling this CurrencyId makes me thing it behaves like a string since almost all instances of "IDs" that I can think of have been strings

src/core/ledger/ledger.js
```diff
@@ -65,9 +75,30 @@ type MutableAccount = {|
...
+  payoutAddresses: PayableAddressStore,
 |};
 export type Account = $ReadOnly<MutableAccount>;
 
+/**
+ * The CurrencyId key must be stringified to ensure the data is retrievable.
+ * Keying on the raw CurrencyId object means keying on the object reference,
+ * rather than the contents of the object.
+ */
+type CurrencyIdString = string;
```
hammadj, line 88: It would be a lot more intuitive id `CurrencyId` was the string that's generated from a `Currency` object

hammadj (commented):
Still think the naming needs some adjustment but happy to reconsider if other devs disagree. I know this PRs been open a while so don't want to be too nitty, but I think naming is super important to get right / consistent. If you feel strongly about your naming and no one else feels strongly about my suggestions, then I'm happy to yield to your discretion :)

src/core/ledger/ledger.js
```diff
@@ -881,6 +978,21 @@ export type TransferGrain = {|
   +amount: G.Grain,
   +memo: string | null,
 |};
+
+type SetPayoutAddress = {|
+  +type: "SET_PAYOUT_ADDRESS",
+  +id: IdentityId,
+  +currencyId: Currency,
```
hammadj, line 985:
```suggestion
  +currency: Currency,
```

src/core/ledger/ledger.js
```diff
@@ -881,6 +978,21 @@ export type TransferGrain = {|
...
+  +payoutAddress: EthAddress | null,
+|};
+
+const setPayoutAddressParser: C.Parser<SetPayoutAddress> = C.object({
+  type: C.exactly(["SET_PAYOUT_ADDRESS"]),
+  id: uuid.parser,
+  currencyId: currencyParser,
```
hammadj, line 992: same here, should be `currency`

src/core/ledger/currency.js
```diff
@@ -53,32 +53,48 @@ const protocolSymbolParser: C.Parser<ProtocolSymbol> = C.exactly([
...
 export const evmChainParser: C.Parser<EvmChainId> = C.fmap(
   C.string,
   parseEvmChainId
 );
 
-export const evmIdParser: C.Parser<EvmId> = C.object({
+export const evmParser: C.Parser<Evm> = C.object({
   type: C.exactly(["EVM"]),
   chainId: evmChainParser,
   currencyAddress: ethAddressParser,
```
hammadj, line 72: Much nicer :) Laaaaast nit would be to name this field `tokenAddress` or `tokenContract` instead of `currencyAddress` since that's the standard name used for this field in most web3 project APIs / docs etc.

e.g. from OpenSea JS: 
image

From Balancer:
image

From Uniswap:
image

hammadj (approved):
Approving with one last nit, feel free to merge after addressing

src/core/ledger/currency.js
topocount: @hammadj I'd be happy to use `token` and `tokenAddress` but we have a special case for native ETH itself. (or xDAI on the xDai sidechain) and I want to be clear that we're accounting for those. We also have other special cases like SAI which don't conform to the ERC-20 standard and therefore begs the question: Is SAI actually a token if it's not ERC-20 conformant?

PR #2431, opened 2020-10-26
Update max name length to 42 chars

topocount:
This increase accomodates full-fidelity Ethereum addresses.


decentralion (approved):
Maybe put in a full ethereum address as a test case

PR #2434, opened 2020-10-27
discord: refactor and test reaction weight logic

decentralion:
The logic for computing Discord reaction weights was starting to get
complicated, and it's all untested, and we want to add more logic still.
This feels like a good opportunity to refactor all the logic before
trying to add any more complexity.

This commit:
- pulls all the logic for computing reaction weights into its own
`reactionWeights.js` module
- tests that logic

I also made a slight tweak to the config format, but I only changed the
new fields that were added since the last release. 1hive will need to
change their config.json, but anyone on the 0.7 branch will not
experience breaking changes. So as to make a cleaner internal interface,
I split the Discord config into two layers: one which is the
json-serialized layer, and one which is the internal in-memory layer,
with an "upgrade" step in-between. Prior to the 0.8.0 release, I'd like
to just refactor the public config file to be a cleaner API, at which
point we can probably remove the upgrade step.

Test plan: All the new unit tests pass, and I tested for
cred-score-stability on an existing instance using the Discord plugin.

### DIFF
--- a/src/plugins/experimental-discord/config.js
+++ b/src/plugins/experimental-discord/config.js
@@ -1,16 +1,26 @@
 // @flow
 
-import * as Combo from "../../util/combo";
+import * as NullUtil from "../../util/null";
+import * as C from "../../util/combo";
 import * as Model from "./models";
 import {
   type EmojiWeightMap,
   type RoleWeightConfig,
   type ChannelWeightConfig,
-} from "./createGraph";
+  type WeightConfig,
+} from "./reactionWeights";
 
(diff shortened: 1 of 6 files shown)


topocount (approved):
Looks good to me. @befitsandpiper and @crisog might want to verify it works with their instance before merging.

befitsandpiper (approved):
Everything seems to work on my end

crisog (approved):
Ran a local test with the 1hive instance, it's looking good. 👍

PR #2436, opened 2020-10-28
testing: Set Up Cypress

topocount:
Cypress is a versatile testing framework that can be used for unit and
end-to-end frontend tests. It utilizes mocha and chai libraries for
writing tests.

The flow libdefs for all of these libraries are quite lacking, however
and I'm strongly considering using a .flow file that is localized to the
cypress folder to resolve this, and additionally, mitigate collisions
with Jest.

Cypress has not been hooked up to the larger testing sequence we have,
	as no useful tests have been written yet.
for an example test see #2437 

test plan: ensure all existing automated test pass

PR #2437, opened 2020-10-28
Cypress: Set up frontend image snapshot testing

topocount:
These changes install a dom/image snapshot testing plugin for Cypress
and initialize snapshots and snapshot tests for the frontend.

No CI integration is configured here. This just sets up the tests within
cypress itself, so environmental configs are out of scope at the moment.

**nb** I realize that this might be under-documented, and I'm very open to suggestions
on how to best share details behind some of these configs.

test plan: `yarn start > /dev/null & yarn cypress run snapshot --headless --browser chrome`


BrianBelhumeur (approved):
Snapshots can be a very useful tool when there are no good alternatives, but I encourage sparse use of them due to their inherent brittleness (e.g. a single minor CSS change can break every snapshot test).

Having a baseline "is the admin UI even loading" is a good start, and we may eventually replace the snapshots with more robust Cypress tests.

cypress/integration/snapshot.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    // this step ensures the frontend has loaded
+    // before taking a snapshot
+    cy.get('[href="#/explorer"]');
+    // $FlowIgnore[prop-missing]
+    cy.document().toMatchImageSnapshot();
+  });
+}
+
+context("Snapshots", () => {
+  it("renders the Explorer view", () => {
```
decentralion, line 33: I notice these are going to map onto file names directly (e.g. `/integration/__image_snapshots__/Snapshots renders the Explorer view #0.png`). I usually see capitalization and spaces in the name like this as a code smell. Is this the cypress default? Is it configurable?

cypress/integration/snapshot.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+function getSnapShotOnLoad(route: string) {
+  cy.visit(route).then(() => {
+    // this step ensures the frontend has loaded
+    // before taking a snapshot
+    cy.get('[href="#/explorer"]');
+    // $FlowIgnore[prop-missing]
+    cy.document().toMatchImageSnapshot();
+  });
+}
```
decentralion, line 30: From a performance standpoint: we load the page from scratch every time (including waiting for loading the cred data). Is there a way we can load the site once and then navigate around it?

decentralion (commented):
so... what's up with all the lint errors?

cypress/integration/snapshot.spec.js
topocount: yes we can definitely do that. The test won't self-document quite as nicely but i think it's worth it. I was actually anticipating this suggestion

topocount:
> so... what's up with all the lint errors?

The solution to fixing them was deleting and rebuilding yarn.lock so I'm not really sure.

topocount:
Will reopen this and #2436 to get them merged later

PR #2439, opened 2020-10-30
api: fix Observable compatibility

### LINKED ISSUE #2351: SourceCred require broken in observable

wchargin:
Summary:
This patch replaces our `getRandomValues` polyfill with a fork that
works in the browser, in Node, in Jest, and in Observable. It’s slightly
hacky because we can’t figure out how to get Webpack’s externs to play
nicely, but the hackiness is entirely contained in this file.

Paired with @topocount.

Fixes #2351.

Test Plan:
Run `yarn build:api`. Test in Node:

```
$ node
> (new (require("./dist/api").default.ledger.ledger.Ledger)()).createIdentity("USER", "z")
'GtJslqp9kl7J8Dr869xGKw'
```

Test in Observable:

 1. Visit <https://observablehq.com/new> and make any change to the
    notebook contents.
 2. Click the “Notebook actions” horizontal ellipsis button to the left
    of the “Publish” button in the top of the screen.
 3. Select “File attachments”.
 4. Click “Upload”, and select `dist/api.js` under your repo.
 5. This will upload `api.js`, or `api@N.js` for some value of `N` if
    you’ve done this before. If you see `api@N.js`, adjust accordingly
    in the following steps.
 6. In an Observable cell, run:

    ```javascript
    {
      const sc = (await require(await FileAttachment("api.js").url()));
      return new sc.default.ledger.ledger.Ledger().createIdentity("USER", "z");
    }
    ```

 7. Observe that it spits out a UUID without crashing.

wchargin-branch: api-observable-build

### DIFF
--- a/src/util/getRandomValues.js
+++ b/src/util/getRandomValues.js
@@ -29,14 +29,15 @@ function getRandomValues(buf: Uint8Array): Uint8Array {
     // Late-import `crypto` to avoid `require` in Observable notebooks,
     // and avoid using a literal `require(...)` to prevent Webpack from
     // rewriting this. (Getting Webpack externals to work properly is .)
+    //
+    /* eslint-disable camelcase */
+    /* eslint-disable no-undef */
     const realRequire =
       // $FlowExpectedError[cannot-resolve-name]
       typeof __non_webpack_require__ !== "undefined"
         ? __non_webpack_require__
         : require; /* needed for Jest */
     const nodeCrypto = realRequire("crypto");
-    if (!nodeCrypto.randomBytes) {
-    }
     if (buf.length > 65536) {
       const e = new Error();
       (e: any).code = 22;
--- a/src/util/getRandomValues.js
+++ b/src/util/getRandomValues.js
@@ -1,0 +1,54 @@
+// @flow
+
+/*
+ * API shim for `window.crypto.getRandomValues` in the browser, for
+ * compatibility with browser, Node, Jest, and Observable.
+ *
+ * Forked from NPM `get-random-bytes` by Kenan Yildirim, which is
+ * released under the MIT License.
+ */
+
+/**
+ * Fill the given buffer with cryptographically secure random bytes. The
+ * buffer length must not exceed 65536.
+ */
+function getRandomValues(buf: Uint8Array): Uint8Array {
+  if (typeof window !== "undefined") {
+    if (window.crypto && window.crypto.getRandomValues) {
+      return window.crypto.getRandomValues(buf);
+    }
+    if (
+      typeof window.msCrypto === "object" &&
+      typeof window.msCrypto.getRandomValues === "function"
+    ) {
+      return window.msCrypto.getRandomValues(buf);
+    }
+  }
+
+  if (typeof require !== "undefined") {
+    // Late-import `crypto` to avoid `require` in Observable notebooks,
+    // and avoid using a literal `require(...)` to prevent Webpack from
+    // rewriting this. (Getting Webpack externals to work properly is .)
+    const realRequire =
+      // $FlowExpectedError[cannot-resolve-name]
+      typeof __non_webpack_require__ !== "undefined"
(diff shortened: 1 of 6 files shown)


src/util/getRandomValues.js
```diff
@@ -0,0 +1,55 @@
...
+      typeof window.msCrypto.getRandomValues === "function"
+    ) {
+      return window.msCrypto.getRandomValues(buf);
+    }
+  }
+
+  if (typeof require !== "undefined") {
+    // Late-import `crypto` to avoid `require` in Observable notebooks,
+    // and avoid using a literal `require(...)` to prevent Webpack from
+    // rewriting this. (Getting Webpack externals to work properly is .)
```
topocount, line 31: seems like you had a thought to finish here

src/util/getRandomValues.js
```diff
@@ -0,0 +1,55 @@
...
+    /* eslint-disable no-undef */
+    const realRequire =
+      // $FlowExpectedError[cannot-resolve-name]
+      typeof __non_webpack_require__ !== "undefined"
+        ? __non_webpack_require__
+        : require; /* needed for Jest */
+    const nodeCrypto = realRequire("crypto");
+    if (buf.length > 65536) {
+      const e = new Error();
+      (e: any).code = 22;
```
topocount, line 43: what is the significance of this? the `code` property isn't documented anywhere that I've seen

src/util/getRandomValues.js
```diff
@@ -0,0 +1,55 @@
...
+  if (typeof require !== "undefined") {
+    // Late-import `crypto` to avoid `require` in Observable notebooks,
+    // and avoid using a literal `require(...)` to prevent Webpack from
+    // rewriting this. (Getting Webpack externals to work properly is .)
+    //
+    /* eslint-disable camelcase */
+    /* eslint-disable no-undef */
+    const realRequire =
+      // $FlowExpectedError[cannot-resolve-name]
+      typeof __non_webpack_require__ !== "undefined"
```
topocount, line 37: ah this is what we were missing!

src/util/getRandomValues.js
```diff
@@ -0,0 +1,55 @@
...
+      throw e;
+    }
+    const bytes = nodeCrypto.randomBytes(buf.length);
+    buf.set(bytes);
+    return buf;
+  }
+  throw new Error("No secure random number generator available.");
+}
+
+module.exports = getRandomValues;
```
topocount, line 55: Out of curiosity, why not use an ES5 export on the function?

topocount (approved):
So glad to see a solution for this that allows us to use our api build everywhere, strange as it is.

src/util/getRandomValues.js
wchargin: Ah, yes; thanks. Let’s go with *daunting*.

src/util/getRandomValues.js
wchargin: Yeah; with or without `externals`, Webpack still munges the `require`
logic with its own machinery. I poked at it for a while and opted to
just avoid it. You can work around it without `__non_webpack_require__`
by doing something like `globals["require"](...)`, but that doesn’t work
in Jest because reasons… this block of code is more finicky than it
might seem.

src/util/getRandomValues.js
wchargin: Yoinked from `!npm get-random-values`, but the purpose is to emulate the
browser error. Node’s `crypto.randomBytes` lets you ask for any number
of random bytes, but in the browser it throws an error with this code
and name if you ask for more than 65536.

src/util/getRandomValues.js
wchargin: An ES6 export does appear to work. I wasn’t sure whether it would, since
it sometimes causes the module-resolution-related things to act
differently, and we’re in finicky module-resolution land. But since it
does work, agreed that it’s nicer. Done.

PR #2447, opened 2020-11-01
grainAllocation: implement Recent policy

elihanover:
This commit implements a Recent policy that rewards contributions based on exponentially weigted recent cred.  You can also interpret this as progressively discounted previous cred.

__Next__
Once this policy is sanity checked, I can include this policy the grain configuration process.  The discount factor is set within `recentReceipts`; however, it makes most sense to include this as a parameter a user can set in their instance's grain configuration file (`config/grain.json`).

Thanks to @decentralion for this idea.

__Test Plan__
Unit tests are provided here, which cover the same base cases as other policies, and also checks the distribution on a simple hand computed distribution.  Given that my mental and programmatic logic could be running the same mistake, a sanity check would help from another party.

### DIFF
--- a/src/core/ledger/grainAllocation.js
+++ b/src/core/ledger/grainAllocation.js
@@ -38,6 +38,15 @@ export type Balanced = "BALANCED";
 export type Immediate = "IMMEDIATE";
 
 /**
+ * The Recent policy distributes cred using a time discount factor, weighing
+ * recent contributions higher.
+ *
+ * Note that this is a generalization of the Immediate policy, where Immediate
+ * is the same as Recent with a discount factor 1 (100%).
+ */
+export type Recent = "RECENT";
+
+/**
  * The Special policy is a power-maintainer tool for directly paying Grain
  * to a target identity. I'm including it because we will use it to create
  * "initialization" payouts to contributors with prior Grain balances in our old
@@ -48,7 +57,11 @@ export type Immediate = "IMMEDIATE";
  */
 export type Special = "SPECIAL";
 
-export type AllocationPolicy = BalancedPolicy | ImmediatePolicy | SpecialPolicy;
+export type AllocationPolicy =
+  | BalancedPolicy
+  | ImmediatePolicy
+  | RecentPolicy
+  | SpecialPolicy;
 
 export type BalancedPolicy = {|
   +policyType: Balanced,
@@ -58,6 +71,12 @@ export type BalancedPolicy = {|
 export type ImmediatePolicy = {|
   +policyType: Immediate,
   +budget: G.Grain,
+|};
+
+export type RecentPolicy = {|
+  +policyType: Recent,
+  +budget: G.Grain,
+  +discount: Discount,
 |};
 
 export type SpecialPolicy = {|
@@ -177,6 +196,8 @@ function receipts(
   switch (policy.policyType) {
     case "IMMEDIATE":
       return immediateReceipts(policy.budget, identities);
+    case "RECENT":
+      return recentReceipts(policy.budget, identities, policy.discount);
     case "BALANCED":
       return balancedReceipts(policy.budget, identities);
     case "SPECIAL":
@@ -189,7 +210,7 @@ function receipts(
 
 /**
  * Split a grain budget in proportion to the cred scores in
- * the most recent time interval
+ * the most recent time interval.
  */
 function immediateReceipts(
   budget: G.Grain,
@@ -199,6 +220,23 @@ function immediateReceipts(
     budget,
     identities.map((i) => i.mostRecentCred)
   );
+  return identities.map(({id}, i) => ({id, amount: amounts[i]}));
+}
+
+/**
+ * Split a grain budget based on exponentially weighted recent
+ * cred.
+ */
+function recentReceipts(
+  budget: G.Grain,
+  identities: ProcessedIdentities,
+  discount: Discount
+): $ReadOnlyArray<GrainReceipt> {
+  const computeDecayedCred = (i) =>
+    i.cred.reduce((acc, cred) => acc * (1 - discount) + cred, 0);
+  const decayedCredPerIdentity = identities.map(computeDecayedCred);
+  const amounts = G.splitBudget(budget, decayedCredPerIdentity);
+
   return identities.map(({id}, i) => ({id, amount: amounts[i]}));
 }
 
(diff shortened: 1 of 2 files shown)


src/core/ledger/grainAllocation.js
```diff
@@ -51,7 +60,7 @@ export type Special = "SPECIAL";
 export type AllocationPolicy = RegularPolicy | SpecialPolicy;
 
 export type RegularPolicy = {|
-  +policyType: Immediate | Balanced,
+  +policyType: Immediate | Recent | Balanced,
   +budget: G.Grain,
 |};
```
decentralion, line 65: A `RegularPolicy` is characterized by the fact that it only has a budget. However, the Recent policy needs a discount factor, so it actually can't be a regular policy. Suggest making:

```js

export type RecentPolicy = {|
  +policyType: "RECENT",
  +budget: G.Grain,
  +discountFactor: number,
|}
```

However, if we do this, then we'll have a strange `AllocationPolicy` line:

```js
export type AllocationPolicy = RegularPolicy | SpecialPolicy | RecentPolicy;
```

I recommend having a separate commit (and separate PR) which renames the policies and splits `RegularPolicy` into just `ImmediatePolicy` and `BalancedPolicy`. Could do that one first, which would follow our [recommended](https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md) workflow.

src/core/ledger/grainAllocation.js
```diff
@@ -197,6 +208,29 @@ function immediateReceipts(
...
+  const amounts = G.splitBudget(
+    budget,
+    identities.map((i) =>
+      i.cred.reduce(
+        (acc, cred, index) =>
+          acc + cred * (1 - discount) ** (identities.length - index),
+        0
+      )
+    )
+  );
```
decentralion, line 229: This seems correct, but I think the following implementation is slightly more straightforward:

```js
i.cred.reduce((acc, cred) => acc * (1-discount) + cred, 0)
```

i.e. we don't need to explicitly find the right power to raise each cred component to, we can just decay it as we scan.

I also recommend factoring this slightly different, as in:

```js
const computeDecayedCred = (i) => i.cred.reduce((acc, cred) => acc * (1-discount) + cred, 0)
const decayedCredPerIdentity = identities.map(computeDecayedCred)
const amounts = G.splitBudget(budget, decayedCredPerIdentity)
```

decentralion (commented):
Thanks for this PR, and nice work @elihanover!

Let's have the discount rate be one of the parameters in the allocation policy. This leaves a much better "audit trail" for inspecting the ledger later, and removes a lot of potential confusion if the project changes the discount rate after having already issued some distributions. I added some suggestions on how to do so.

src/core/ledger/grainAllocation.js
elihanover: Ooh pretty, yes I agree this is a lot better.

src/core/ledger/grainAllocation.js
elihanover: What if `AllocationPolicy` was a type with some `policyType`, `budget`, and `receipts` function?  This way, whenever we add or rename policies, we don't have to redefine the `AllocationPolicy`.  This feels like the essence of an `AllocationPolicy`.

```js
// each policy has a receipts function of the same type, so let's define that type
export type ReceiptGenerator = (
   policy: AllocationPolicy,
   identities: ProcessedIdentities
) => $ReadOnlyArray<GrainReceipt>;

export type AllocationPolicy = {
   +policyType: string,
   +budget: G.Grain,
   +receipts: ReceiptGenerator
}
```

It also lets us remove the `receipts` function and instead do call the `AllocationPolicy`'s own `receipts` function:
```js
export function computeAllocation(
  policy: AllocationPolicy,
  identities: $ReadOnlyArray<AllocationIdentity>
): Allocation {
  const validatedPolicy = _validatePolicy(policy);
  const processedIdentities = _processIdentities(identities);
  return _validateAllocationBudget({
    policy,
    receipts: validatedPolicy.receipts(processedIdentities), // <-- this 
    id: randomUuid(),
  });
}

// REMOVE
function receipts(
  policy: AllocationPolicy,
  identities: ProcessedIdentities
  switch (policy.policyType) {
    case "IMMEDIATE":
      return immediateReceipts(policy.budget, identities);
    case "BALANCED":
      return balancedReceipts(policy.budget, identities);
    case "SPECIAL":
      return specialReceipts(policy, identities);
    // istanbul ignore next: unreachable per Flow
    default:
      throw new Error(`Unknown policyType: ${(policy.policyType: empty)}`);
  }
}
```

decentralion: AllocationPolicy needs to be serializable:

Allocation has the policy:
https://github.com/sourcecred/sourcecred/blob/db854f6925e15601428073641da3484b8d1b3241/src/core/ledger/grainAllocation.js#L70-L73

Allocation is included in a distribution:
https://github.com/sourcecred/sourcecred/blob/db854f6925e15601428073641da3484b8d1b3241/src/core/ledger/distribution.js#L8-L12

And a distribution is included in the ledger:
https://github.com/sourcecred/sourcecred/blob/db854f6925e15601428073641da3484b8d1b3241/src/core/ledger/ledger.js#L776-L779

Since a function isn't serializable, this wouldn't work. Nice suggestion tho.

elihanover: Just to make sure I understand:
```js
export type RegularPolicy = BalancedPolicy | ImmediatePolicy;

export type BalancedPolicy = {|
  +policyType: Balanced,
  +budget: G.Grain,
|}

  +policyType: Immediate,
  +budget: G.Grain,
|}
```
I'm a fan of this style where each policy has it's own `______Policy` object and parser, since we can compose them into groups like `RegularPolicy`.  If we're on the same page, I'll open a separate PR.

src/core/ledger/grainAllocation.js
```diff
@@ -48,13 +57,26 @@ export type Immediate = "IMMEDIATE";
  */
 export type Special = "SPECIAL";
 
-export type AllocationPolicy = RegularPolicy | SpecialPolicy;
+export type AllocationPolicy = RegularPolicy | RecentPolicy | SpecialPolicy;
+
+export type RegularPolicy = BalancedPolicy | ImmediatePolicy;
```
decentralion, line 62: In that case, we don't have a need for the `RegularPolicy` type anymore, right? Let's just have `AllocationPolicy = Immediate | Balanced | Recent | Special`

src/core/ledger/grainAllocation.js
```diff
@@ -288,3 +342,12 @@ export const allocationParser: P.Parser<Allocation> = P.object({
   id: uuidParser,
   receipts: P.array(grainReceiptParser),
 });
+
+export opaque type Discount = number;
```
decentralion, line 346: by making this opaque, it means that other modules won't be allowed to use it like a number (e.g. calling `.toFixed()`), which is a little too restrictive. What we'd want is an opaque subtype of number, i.e. `export opaque type Discount: number = number`

decentralion (approved):

decentralion:
Needs rebase on #2453

elihanover:
My rebasing has devolved into a hot mess.  Would appreciate if someone could me get this sorted out.

decentralion (approved):

PR #2166, opened 2020-08-26
Sourcecred Gitcoin Plugin 

psudoanon:
[Architecture Specification can be found here](https://gist.github.com/psudoanon/3847235ad13835145f14d76c76b47d6f)
[Instructions for running with Gitcoin can be found here](https://gist.github.com/psudoanon/24ff5f3e93f887fa8c5ca3fb0f68209c)

Let me know what you guys think :)

PR #2452, opened 2020-11-03
Sourcecred Gitcoin Plugin 2

chibie:
sequel to #2166 
paired with: @adamallidina
this PR requires `GITCOIN_USER_WHITELIST` environmental variable if a config json file doesn't exist.

### DIFF
--- a/src/plugins/gitcoin/declaration.js
+++ b/src/plugins/gitcoin/declaration.js
@@ -1,0 +1,97 @@
+// @flow
+
+import deepFreeze from "deep-freeze";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {NodeType, EdgeType} from "../../analysis/types";
+import {NodeAddress, EdgeAddress} from "../../core/graph";
+
+export const nodePrefix = NodeAddress.fromParts(["sourcecred", "gitcoin"]);
+export const edgePrefix = EdgeAddress.fromParts(["sourcecred", "gitcoin"]);
+
+export const userNodeType: NodeType = deepFreeze({
+  name: "User",
+  pluralName: "Users",
+  prefix: NodeAddress.append(nodePrefix, "user"),
+  defaultWeight: 0,
+  description: "A user account on GitCoin",
+});
+
+export const postNodeType: NodeType = deepFreeze({
+  name: "Post",
+  pluralName: "Posts",
+  prefix: NodeAddress.append(nodePrefix, "post"),
+  defaultWeight: 8,
+  description: "A townsquare post on GitCoin",
+});
+
+export const commentNodeType: NodeType = deepFreeze({
+  name: "Comment",
+  pluralName: "Comments",
+  prefix: NodeAddress.append(nodePrefix, "comment"),
+  defaultWeight: 4,
+  description: "A comment by some user, directed at some post",
+});
+
+export const createsPostEdgeType: EdgeType = deepFreeze({
+  forwardName: "creates post",
+  backwardName: "post is created by",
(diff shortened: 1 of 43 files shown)


src/plugins/gitcoin/declaration.js
```diff
@@ -0,0 +1,97 @@
+// @flow
+
+import deepFreeze from "deep-freeze";
+import type {PluginDeclaration} from "../../analysis/pluginDeclaration";
+import type {NodeType, EdgeType} from "../../analysis/types";
+import {NodeAddress, EdgeAddress} from "../../core/graph";
```
topocount, line 6:
```suggestion
import {
  NodeAddress,
  EdgeAddress,
  type NodeAddressT,
  type EdgeAddressT,
} from "../../core/graph";
```

src/plugins/gitcoin/declaration.js
```diff
@@ -0,0 +1,97 @@
...
+
+export const nodePrefix = NodeAddress.fromParts(["sourcecred", "gitcoin"]);
+export const edgePrefix = EdgeAddress.fromParts(["sourcecred", "gitcoin"]);
```
topocount, line 9:
```suggestion
export const nodePrefix: NodeAddressT = NodeAddress.fromParts([
  "sourcecred",
  "gitcoin",
]);
export const edgePrefix: EdgeAddressT = EdgeAddress.fromParts([
  "sourcecred",
  "gitcoin",
]);
```

topocount (commented):
This is looking great! After I made the changes I suggested below, this branch passes the full test suite. A couple more steps are needed to get this approved and merged but you're extremely close:

1. Rebase on sourcecred's master branch, since we bumped our flow dependency a while ago and it's bit more strict than what's getting tested here

2. merge or re-implement my suggested changes below to fix the lastest flow error (this is due to a new convention in flow called [types-first](https://flow.org/en/docs/lang/types-first/) if you're curious)

topocount (approved):
Thanks!

wchargin:
This breaks the build for me due to the new dep on `pg-native`, which
requires platform-specific installs. That complicates the install
process for all SourceCred users.

I have tried to determine how and why `pg-native` is used, but the
config format is not documented or (that I can tell) tested. Could you
please explain to me why we need a Postgres client? Is this implementing
a local mirror repository using Postgres as a RDBMS, or talking to a
GitCoin-hosted Postgres server to read data, or something else?

wchargin:
@topocount also tells me that he had to install Postgres to keep
building as of this change, so it’s not just a broken setup on my end.
This is not good: SourceCred has not historically required any system
package management for our end users.

So I think the questions in my earlier comment still stand; @chibie,
please let me know at your earliest convenience. Thanks!

topocount:
@wchargin @chibie update on this. I installed postgres while troubleshooting my installation issues, but I just verified that I don't (and shouldn't need) postgres itself to successfully `yarn install`

wchargin:
(Per chat, @topocount does still need `libpq-dev` installed.)

topocount:
@chibie we've decided to revert this for now because of the libpq dependency. This would be a big ask of all sourcecred users to install libpq-dev, but this can be resolved by utilizing a postgres client that does not require libpq bindings and ideally is pure javascript, such as [node-postgres](https://www.npmjs.com/package/pg) as an example.

chibie:
@wchargin @topocount 

yes, this makes sense! i'll make the adjustments to use [node-postgres](https://www.npmjs.com/package/pg) and raise a new PR after some further house cleaning

PR #2456, opened 2020-11-04
Added compareGraphs method

blueridger:
This adds a compareGraphs method with tests. This method will ultimately be used to create a `-d` option for the `sourcecred graph` command that will report the difference with the existing written graph.

### Automated Tests:
`yarn test` passes

### Manual Tests:
None

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -1212,3 +1212,70 @@ export function edgeToParts(
   const timestampMs = edge.timestampMs;
   return {addressParts, srcParts, dstParts, timestampMs};
 }
+
+export type GraphComparison = {|
+  +graphsAreEqual: boolean,
+  +uniqueNodesInFirst: $ReadOnlyArray<Node>,
+  +uniqueNodesInSecond: $ReadOnlyArray<Node>,
+  +nodeTuplesWithDifferences: $ReadOnlyArray<[Node, Node]>,
+  +uniqueEdgesInFirst: $ReadOnlyArray<Edge>,
+  +uniqueEdgesInSecond: $ReadOnlyArray<Edge>,
+  +edgeTuplesWithDifferences: $ReadOnlyArray<[Edge, Edge]>,
+|};
+
+export function compareGraphs(
+  firstGraph: Graph,
+  secondGraph: Graph
+): GraphComparison {
+  const uniqueNodesInFirst = [];
+  const uniqueNodesInSecond = [];
+  const nodeTuplesWithDifferences = [];
+  const uniqueEdgesInFirst = [];
+  const uniqueEdgesInSecond = [];
+  const edgeTuplesWithDifferences = [];
+  const graphsAreEqual = firstGraph.equals(secondGraph);
+
+  if (!graphsAreEqual) {
+    for (const firstNode of firstGraph.nodes()) {
+      const secondNode = secondGraph.node(firstNode.address);
+      if (secondNode) {
+        if (!deepEqual(firstNode, secondNode))
+          nodeTuplesWithDifferences.push([firstNode, secondNode]);
+      } else {
+        uniqueNodesInFirst.push(firstNode);
+      }
+    }
+    for (const secondNode of secondGraph.nodes()) {
+      const firstNode = firstGraph.node(secondNode.address);
+      if (!firstNode) {
+        uniqueNodesInSecond.push(secondNode);
+      }
+    }
+
+    for (const firstEdge of firstGraph.edges({showDangling: true})) {
+      const secondEdge = secondGraph.edge(firstEdge.address);
+      if (secondEdge) {
+        if (!deepEqual(firstEdge, secondEdge))
+          edgeTuplesWithDifferences.push([firstEdge, secondEdge]);
+      } else {
+        uniqueEdgesInFirst.push(firstEdge);
+      }
+    }
+    for (const secondEdge of secondGraph.edges({showDangling: true})) {
+      const firstEdge = firstGraph.edge(secondEdge.address);
+      if (!firstEdge) {
+        uniqueEdgesInSecond.push(secondEdge);
+      }
+    }
+  }
+
+  return {
+    graphsAreEqual,
+    uniqueNodesInFirst,
+    uniqueNodesInSecond,
+    nodeTuplesWithDifferences,
+    uniqueEdgesInFirst,
+    uniqueEdgesInSecond,
+    edgeTuplesWithDifferences,
+  };
+}
--- a/src/core/graph.test.js
+++ b/src/core/graph.test.js
@@ -18,6 +18,7 @@ import {
   edgeToString,
   edgeToStrings,
   edgeToParts,
+  compareGraphs,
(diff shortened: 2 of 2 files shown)


src/core/graph.js
```diff
@@ -1212,3 +1212,73 @@ export function edgeToParts(
...
+  secondGraph: Graph
+): {|
+  +isEqual: boolean,
+  +uniqueNodesInFirst: Node[],
+  +uniqueNodesInSecond: Node[],
+  +nodeTuplesWithDifferences: [Node, Node][],
+  +uniqueEdgesInFirst: Edge[],
+  +uniqueEdgesInSecond: Edge[],
+  +edgeTuplesWithDifferences: [Edge, Edge][],
+|} {
```
decentralion, line 1227: I recommend pulling this type out as an exported `GraphComparison`, it's helpful for complex types like this so we can explicitly say (e.g.) that a `printComparisonResult` function takes a `GraphComparison` as an argument.

src/core/graph.js
```diff
@@ -1212,3 +1212,73 @@ export function edgeToParts(
   const timestampMs = edge.timestampMs;
   return {addressParts, srcParts, dstParts, timestampMs};
 }
+
+export function compareGraphs(
+  firstGraph: Graph,
+  secondGraph: Graph
+): {|
+  +isEqual: boolean,
```
decentralion, line 1220: nit: `isEqual` feels a little odd as a name, since the graph comparison is an object describing two graphs. so `areEqual` might be the right conjugation? or the even more explicit `graphsAreEqual`?

also, did you know that there's a (more efficient) `graph.equals(otherGraph)` method already defined? it could make sense for callers to check graph equality using that method, and then compute the (more expensive) full comparison result only when the graphs are not equal, in which case this field might not be as necessary. that said, having this field does seem like a nice convenience, so not requesting it be removed.

src/core/graph.js
```diff
@@ -1212,3 +1212,73 @@ export function edgeToParts(
...
+  +uniqueNodesInFirst: Node[],
+  +uniqueNodesInSecond: Node[],
+  +nodeTuplesWithDifferences: [Node, Node][],
+  +uniqueEdgesInFirst: Edge[],
+  +uniqueEdgesInSecond: Edge[],
+  +edgeTuplesWithDifferences: [Edge, Edge][],
```
decentralion, line 1226: Recommend changing all the arrays to read-only arrays, e.g. `s/Node[]/$ReadOnlyArray<Node>/`.

there's no reason why a caller would want the ability to mutate the graph comparison result, since doing so would just make the comparison inaccurate.

src/core/graph.test.js
```diff
@@ -1663,4 +1664,87 @@ describe("core/graph", () => {
       expect(edgeToParts(edge)).toEqual(expected);
     });
   });
+
+  describe("compareGraphs", () => {
+    describe("equivalent graphs", () => {
+      const graph1 = new Graph().addNode(src).addNode(dst).addEdge(simpleEdge);
+      const graph2 = new Graph().addNode(src).addNode(dst).addEdge(simpleEdge);
```
decentralion, line 1671: These two graphs are very simple (e.g. they do not include dangling edges) and also have the exact same creation history. It would be better to stress test, e.g.:

graphs that have dangling edges
two graphs which are the same but have very different history (e.g. nodes that were added and removed)

Happily, we have helper methods for doing so:
https://github.com/sourcecred/sourcecred/blob/master/src/core/graphTestUtil.js

src/core/graph.test.js
```diff
@@ -1663,4 +1664,87 @@ describe("core/graph", () => {
...
+
+      it("returns empty arrays", () => {
```
decentralion, line 1673: recommend renaming test case to "equivalent graphs are equal with no differences" or such

src/core/graph.test.js
```diff
@@ -1663,4 +1664,87 @@ describe("core/graph", () => {
       expect(edgeToParts(edge)).toEqual(expected);
     });
   });
+
+  describe("compareGraphs", () => {
```
decentralion, line 1668: it'd be good to have a test verifying that two graphs which are the same except for a single dangling edge still produces diffs occasionally. (i.e. if someone were to go and swap `showDangling: false` the test should fail)

decentralion (commented):
Looks good overall, small changes requested

decentralion (approved):
Looks good; approved. My recommendation would be to implement and get reviews the full feature (i.e. through to having a flag on the `graph` CLI command which prints the diff) before merging this PR. In my experience, writing the intended consumer for an API often reveals opportunities to improve the original API, and writing such improvements is smoother if the PR implementing the original API hasn't merged yet.

Mechanically, you'd do this by pushing subsequent PRs which merge back into the `compare-graphs` branch, getting those reviewed and approved, and then merging the stack from bottom-to-top (merge this PR, rebase the next one, merge it, etc). The process is slightly more work mechanically, but I find that when I follow this process, I tend to write better code.

Also, as a matter of review etiquette, I'd appreciate if you respond to comments when you mark it resolved (something as concise as "done", or "hear you, not bothering since it's a tiny nit"). It makes it feel less like I'm talking into a void as a reviewer, and makes the code easier to re-review since I know how you addressed (or didn't deliberately didn't address) the comments.

src/core/graph.js
blueridger: done

src/core/graph.js
blueridger: done and done

src/core/graph.js
blueridger: done

src/core/graph.test.js
blueridger: done

src/core/graph.test.js
blueridger: done

src/core/graph.test.js
blueridger: done

src/core/graph.js
```diff
@@ -1212,3 +1212,70 @@ export function edgeToParts(
   const timestampMs = edge.timestampMs;
   return {addressParts, srcParts, dstParts, timestampMs};
 }
+
+export type GraphComparison = {|
+  +graphsAreEqual: boolean,
+  +uniqueNodesInFirst: $ReadOnlyArray<Node>,
+  +uniqueNodesInSecond: $ReadOnlyArray<Node>,
+  +nodeTuplesWithDifferences: $ReadOnlyArray<[Node, Node]>,
```
decentralion, line 1220: Riffing off [this comment](https://github.com/sourcecred/sourcecred/pull/2458/files#r519002370):

What if we had a type `NodeDiff = {|+first: Node | null, +second: Node | null|}`
And then instead of having `uniqueNodesInFirst`, `uniqueNodesInSecond`, `nodeTuplesWithDifferences`, we could just have a single array of `NodeDiffs`?

blueridger: accepted, I can bake this into the other PR.

PR #2458, opened 2020-11-05
Add compareWeights function

blueridger:
This adds a comparison function for the Weights object. I will prioritize using the `compareWeightedGraphs` function I made instead of this function for the `sourcecred graph -d` diff flag, but I still think this can add value and I may want to play with adding granular options to the flag later.

## Automated testing
`yarn test`

## Manual testing
None

### DIFF
--- a/src/core/weights.js
+++ b/src/core/weights.js
@@ -1,5 +1,6 @@
 // @flow
 
+import deepEqual from "lodash.isequal";
 import * as MapUtil from "../util/map";
 import * as C from "../util/combo";
 import {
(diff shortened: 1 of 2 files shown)


decentralion (commented):
Currently merging into the wrong branch--makes it difficult to review.

src/core/weights.js
```diff
@@ -165,3 +166,72 @@ export function toJSON(weights: Weights): WeightsJSON {
 export function fromJSON(json: WeightsJSON): Weights {
   return parser.parseOrThrow(json);
 }
+
+export type WeightsComparison = {|
+  +weightsAreEqual: boolean,
+  +uniqueNodesInFirst: $ReadOnlyArray<[NodeAddressT, NodeWeight]>,
```
decentralion, line 172: The name `uniqueNodesInFirst` is confusing, since we don't have any `Node`s in this data structure at all

How about the simpler form like this:

`export type NodeWeightDiff = {|+first: NodeWeight | null, +second: NodeWeight | null|}`
then insetad of having uniqueNodesInFirst, uniqueNodesInSecond, nodeTuplesWithDifferences, there can be a single field of diffs

blueridger: accepted

src/core/weights.js
```diff
@@ -165,3 +166,90 @@ export function toJSON(weights: Weights): WeightsJSON {
...
+            second: secondWeight,
+          });
+      } else {
+        nodeWeightDiffs.push({
+          address: address,
+          first: firstWeight,
+          second: null,
+        });
+      }
+    }
```
decentralion, line 213: how about:

```js

const addresses = new Set([...firstWeights.nodeWeights.keys(), ...secondWeights.nodeWeights.keys()])
for (const a of addresses) {
  const first = firstWeights.nodeWeights.get(a)
  const second = secondWeights.nodeWeights.get(a)
  if (first !== second) {
    nodeWeightDiffs.push({
      address: a,
      first: first === undefined ? null : first,
     second: second === undefined ? null : second
  })
```

This way you only need one loop for both sides of the diff.

(and likewise for edges)

Note if we want to avoid the `first === undefined ? null : first` bit, we could just change the typedef so instead of being `NodeWeight | null` it's just `?NodeWeight`. Often I like specifying that we'll have `null` not `undefined`, but in this case it's kind of natural since `Map.get(` returns undefined if there's no entry for that key.

src/core/weights.js
```diff
@@ -165,3 +166,90 @@ export function toJSON(weights: Weights): WeightsJSON {
...
+  +edgeWeightDiffs: $ReadOnlyArray<EdgeWeightDiff>,
+|};
+
+export function compareWeights(
+  firstWeights: Weights,
+  secondWeights: Weights
+): WeightsComparison {
+  const nodeWeightDiffs = [];
+  const edgeWeightDiffs = [];
+  const weightsAreEqual = deepEqual(firstWeights, secondWeights);
```
decentralion, line 194: I'm not sure this is needed, feels like an unecessary performance optimization. Also, off the top of my head I'm not 100% sure that `new Map().set("a", 1).set("b", 2)` is deep equal to `new Map().set("b", 2).set("a", 1)`. Since afaik maps do iterate in insertion order, there is an observable diff between the two. But probably deep equal would return true?

src/core/weights.js
blueridger: sure, that's nicer.

src/core/weights.js
blueridger: yeah from my tests, the deep equals is order-agnostic. it does seem like an opportunity for unexpected behavior to creep in, though, and I only added it on your request, so I'm happy to take it out.

decentralion: Yeah, in the case of graph comparison, there was more overhead in iterating over all the nodes v.s. using the `graph.equals` method which is a lot more efficient. Here the overhead is a lot smaller. But I think in either case, my suggestion was a premature optimization that added complexity in a case that's very unlikely to be the perf-critical path.

src/core/weights.js
```diff
@@ -165,3 +166,68 @@ export function toJSON(weights: Weights): WeightsJSON {
...
+        first: first,
+        second: second,
+      });
+    }
+  }
+  const weightsAreEqual =
+    nodeWeightDiffs.length === 0 && edgeWeightDiffs.length === 0;
+
+  return {
+    weightsAreEqual,
```
decentralion, line 229: If i were writing this API, I would not include `weightsAreEqual` b.c. the `nodeWeightDiffs || edgeWeightDiffs` check is trivial. However, I don't object to it being included. (Hence approval)

decentralion (approved):

PR #2461, opened 2020-11-06
Compare weighted graphs

blueridger:
# Notes
This adds a comparator for the WeightedGraph type. This will be used for the `sourcecred graph -d` diff flag instead of the other two comparators I was working on. I intend to keep the other two, though, and maybe experiment with adding more granular options to the flag.

# Automated Tests
`yarn test`

# Manual Tests
Tried it in WIP cli

### DIFF
--- a/src/core/graphTestUtil.js
+++ b/src/core/graphTestUtil.js
@@ -10,6 +10,12 @@ import {
   type NodeAddressT,
   type EdgeAddressT,
 } from "./graph";
+import {
+  type NodeWithWeight,
+  type EdgeWithWeight,
+  type WeightedGraph,
+} from "./weightedGraph";
+import {empty as emptyWeights} from "./weights";
 import type {TimestampMs} from "../util/timestamp";
 
 /**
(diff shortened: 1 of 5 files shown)


src/core/graphTestUtil.js
```diff
@@ -262,3 +268,24 @@ export function advancedGraph(): {|
   };
   return {nodes, edges, graph1, graph2};
 }
+
+export function testWeightedGraph(
+  nodesWithWeights: NodeWithWeight[],
+  edgesWithWeights: EdgeWithWeight[]
+): WeightedGraph {
```
decentralion, line 275: Note: This API makes it impossible to construct graphs that have prefix-level weights that aren't bound to any node in particular.

src/core/weightedGraph.js
```diff
@@ -76,3 +77,108 @@ export function overrideWeights(
...
+
+export type WeightedGraphComparison = {|
+  +weightedGraphsAreEqual: boolean,
+  +uniqueNodesInFirst: $ReadOnlyArray<NodeWithWeight>,
+  +uniqueNodesInSecond: $ReadOnlyArray<NodeWithWeight>,
+  +nodeTuplesWithDifferences: $ReadOnlyArray<[NodeWithWeight, NodeWithWeight]>,
+  +uniqueEdgesInFirst: $ReadOnlyArray<EdgeWithWeight>,
+  +uniqueEdgesInSecond: $ReadOnlyArray<EdgeWithWeight>,
+  +edgeTuplesWithDifferences: $ReadOnlyArray<[EdgeWithWeight, EdgeWithWeight]>,
+|};
```
decentralion, line 99: This comparison is incapable of representing the case where the WeightedGraphs differ in prefix-set weights (i.e. weights that do not correspond to a specific node in the graph)

src/core/weightedGraph.js
```diff
@@ -76,3 +77,108 @@ export function overrideWeights(
   });
   return {graph: wg.graph, weights};
 }
+
+export type NodeWithWeight = {|
+  +node: Node,
+  +weight: ?Weights.NodeWeight,
```
decentralion, line 83: There's two concepts of weight for a node or edge in the graph. The "own weight" which is the weight directly assigned to this object (which may be undefined), and the "evaluated weight" which is the implied weight from the structure of the weighted graph (which is always defined, and defaults to 1 in the case of nodes). Should take care to define which one we're talking about here.

decentralion (commented):
I don't like this API, in part because I really don't like the weighted graph APIs. We should maybe talk about this a bit more--I think we could do something good here if we refactor the WeightedGraph class, but if we want to take the expedient path, maybe we should just diff the weights and diff the graph like you were originally planning.

Issue #2460, opened 2020-11-06
`yarn graph` gives opaque error when `ledger.json` is empty


BrianBelhumeur:
Currently, `yarn graph` requires an empty array in `ledger.json` when computing a new graph. Without this array, the following error occurs:
```
➜  template-instance git:(master) ✗ yarn graph
yarn run v1.22.5
$ dotenv sourcecred graph
  GO   graph
Error: line 1 is not valid JSON: SyntaxError: Unexpected end of JSON input
    at eval (webpack:///./src/util/jsonLog.js?:23:551)
    at Array.forEach (<anonymous>)
    at Function.fromString (webpack:///./src/util/jsonLog.js?:23:463)
    at Function.parse (webpack:///./src/core/ledger/ledger.js?:173:97)
    at loadLedger (webpack:///./src/cli/common.js?:40:618)
    at async graphCommand (webpack:///./src/cli/graph.js?:21:1158)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Since the error offers little in the way of debugability, I suggest one of two solutions when `yarn graph` is run and `ledger.json` is empty:
1. Default the contents to an empty array and continue
or
2. Throw an error that points to the need for an empty array

topocount:
I've actually replicated this as a bug in a completely fresh template instance. Running `yarn go` without configuring any plugins completely wipes out the contents of the ledger.json file when `yarn graph` runs.

BrianBelhumeur:
Ah ha, thats how I triggered it! I saw on my git diffs that the `[]` had been removed from `ledger.json` but not sure how. I did indeed run `yarn load` before configuring plugins. This seems like an additional place we could add some better error messaging (e.g. “warning: no plugins configured!”) as well as fixing the bug that removes the empty array from `ledger.json`

topocount:
> This seems like an additional place we could add some better error messaging (e.g. “warning: no plugins configured!”) as well as fixing the bug that removes the empty array from `ledger.json`

haha I'm just seeing this @BrianBelhumeur. That's exactly what I did in #2485 and #2486.

PR #2465, opened 2020-11-06
Support `sourcecred go` out of the box

topocount:
This PR allows `sourcecred go` to succeed out of the box in the template instance. Each small bugfix and refactor needed to make this happen is contained in its own semantically atomic commit. 

test plan:
1. yarn build
2. run `scdev go` in the template-instance repo

### DIFF
--- a/src/cli/common.js
+++ b/src/cli/common.js
@@ -34,12 +34,6 @@ export async function loadInstanceConfig(
 ): Promise<InstanceConfig> {
   const projectFilePath = pathJoin(baseDir, "sourcecred.json");
   const config = await loadJson(projectFilePath, configParser);
-  if (config.bundledPlugins.size === 0) {
-    throw new Error(
-      "No plugins configured; Please set up at least one plugin: " +
-        "https://github.com/sourcecred/template-instance#supported-plugins"
-    );
-  }
   return config;
 }
 
--- a/src/cli/common.js
+++ b/src/cli/common.js
@@ -29,9 +29,18 @@ import {loadFileWithDefault, loadJsonWithDefault} from "../util/disk";
 
 import * as Weights from "../core/weights";
 
-export function loadInstanceConfig(baseDir: string): Promise<InstanceConfig> {
+export async function loadInstanceConfig(
+  baseDir: string
+): Promise<InstanceConfig> {
   const projectFilePath = pathJoin(baseDir, "sourcecred.json");
-  return loadJson(projectFilePath, configParser);
+  const config = await loadJson(projectFilePath, configParser);
+  if (config.bundledPlugins.size === 0) {
+    throw new Error(
+      "No plugins configured; Please set up at least one plugin: " +
+        "https://github.com/sourcecred/template-instance#supported-plugins"
+    );
+  }
+  return config;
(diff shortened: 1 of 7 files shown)


decentralion (commented):
What are the unintended side effects we're trying to prevent? What are the root causes of those side effects? Are there other code paths that trigger the same side effects, which aren't fixed by this change?

While I don't object to this change, it feels a little hacky and ad hoc to me. If we're going to be validating the bundledPlugins, we should probably also verify that each putative bundled plugin is actually a valid plugin.

topocount:
@decentralion this is basically a an effort to prevent #2460. Initially I set out to make changes everything runs without error with no plugins (because why shouldn't that be okay?), but The sourcecred dependency-minting config and the default `[]` in the empty ledger file were two special cases that would need to be addressed in these changes. 

I've found that naive users can fall into a hopeless loop trying to test out the example instance, where the first `go` call deletes empty array from ledger.json each time `graph` runs, and then once they realize they need to replace that to run the `score` command, they'll get another error because the sourcecred dependency node throws this error on `score`: `Error: no Account for identity: 2zWwkGHnBGXmEcYneeWAzw`

I figured that, rather than trying to accommodate these and potentially more special cases, we could prevent users from doing something that would just lead them down a path of opaque errors, and could also direct them towards setting up the instance properly. 

I'm happy to add some validation for plugins at this point in the code, but I'm aware that it already occurs later as well. This change is mostly meant to minimize the amount of friction users encounter while setting up an instance,

decentralion:
> Initially I set out to make changes so everything runs without error with no plugins (because why shouldn't that be okay?)

My @wchargin simulator strongly favors making this OK.

> but The sourcecred dependency-minting config and the default [] in the empty ledger file were two special cases that would need to be addressed in these changes.

Yeah, those sound like root issues that should be fixed, rather than papering over them (which this PR does, imo).

> the first go call deletes empty array from ledger.json each time graph runs

Why does it do that? Can we fix that?

topocount:
> > Initially I set out to make changes so everything runs without error with no plugins (because why shouldn't that be okay?)
> 
> My @wchargin simulator strongly favors making this OK.
> 
> > but The sourcecred dependency-minting config and the default [] in the empty ledger file were two special cases that would need to be addressed in these changes.
> 
> Yeah, those sound like root issues that should be fixed, rather than papering over them (which this PR does, imo).
> 
@decentralion While I am in favor making the "base case" (clean template instance) run cleanly, I don't think it's particularly helpful for users for that to happen, unless we log a message after everything completes with similar instructions to what I pushed. I think it's important that we provide users toying around with the instance with some direction after everything runs, since it decreases needed support labor and helps them to get started on their own.

decentralion (commented):
Cool, these changes look good to me now. I recommend splitting them into 4 smaller PRs which each have a well-defined scope. This will be slightly more work since each PR will need its own test plan, which is a Good Thing :). (For the PRs with unit tests, those likely suffice for the test plan).

This will also get you more Cred, which IMO is quite well deserved.

topocount:
Closing in favor of #2485, #2486, #2487, 2488

PR #2466, opened 2020-11-06
add description and timestamp to nodeToString

blueridger:
As a part of the `graph -d` diff flag project, I am updating the nodeToStirng method to include all attributes.

## Testing
yarn test
used it in my WIP cli branch and worked as expected.

### DIFF
--- a/src/core/graph.js
+++ b/src/core/graph.js
@@ -1162,7 +1162,9 @@ export class Graph {
  */
 export function nodeToString(node: Node): string {
   const address = NodeAddress.toString(node.address);
-  return `{address: ${address}}`;
+  const timestamp =
+    node.timestampMs == null ? "null" : String(node.timestampMs);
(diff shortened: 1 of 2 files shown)


src/core/graph.js
```diff
@@ -1162,7 +1162,8 @@ export class Graph {
  */
 export function nodeToString(node: Node): string {
   const address = NodeAddress.toString(node.address);
-  return `{address: ${address}}`;
+  const timestamp = node.timestampMs || "null";
```
decentralion, line 1165: In the case where the timestampMs is 0, this would improperly report null. `0` is a valid timestamp.

Recommend using `NullUtil.orElse`, or `node.timestampMs == null ? "null" : String(node.timestampMs)`

blueridger: accepted

decentralion (approved):

PR #2468, opened 2020-11-07
Discord: Refactor createGraph

blueridger:
# Decentralion
Paired with: @decentralion 
This refactors the createGraph function in the Discord plugin to separate
data processing from graph creation. Basically, it splits it into a
function called `_findGraphMessages` which yields all of the information
needed to produce the graph, and `_createGraphFromMessages` generates
the actual graph using this data.

Doing it this way allows us to simplify the control flow. As an example:
we have a rule that we don't emit message nodes for messages with no
reactions or mentions. Previously, this was tracked via a `hasEdges`
boolean which would track whether we had added any reaction or mention
edges. Now, the `findGraphMessages` function explicitly only yields
messages that have mentions or reactions.

This refactor makes it easier to write unit tests (and allows us to
prioritize testing the areas that will have more logic and control flow
cases, i.e. the `_findGraphMessages` function). However, I haven't
actually added such tests, mostly because producing a mock
`SqliteMirrorRepository` for testing purposes will be a pain, and I'm
short on bandwidth. I'm comfortable merging this sans-testing because
I think it's a meaningful improvement/cleanup of the current logic
(which is also untested).

# Thena
I fixed some bugs and made a change that preserves previous behavior by allowing author to be null. As a follow up, I will start designing improved behavior, further refactors, and unit testing options.
## Automated testing
yarn test 
this code is untested, though

## Manual testing
I used my WIP `graph -d` command to ensure that both the weights and the graph remain unchanged. I tested against the SC test discord server by switching between these two commands and ensuring that no changes were reported:
```
yarn clean && scdev load && scdev graph -d

yarn clean && yarn load && scdev graph -d
```

### DIFF
--- a/src/plugins/experimental-discord/createGraph.js
+++ b/src/plugins/experimental-discord/createGraph.js
@@ -188,9 +188,85 @@ function mentionsEdge(message: Model.Message, member: Model.GuildMember): Edge {
   };
 }
 
+export type GraphReaction = {|
+  +reaction: Model.Reaction,
+  +reactingMember: Model.GuildMember,
+|};
+
+/**
+ * All of the information necessary to add a message to
+ * the graph, along with its reactions and its mentions.
+ */
+export type GraphMessage = {|
+  +message: Model.Message,
+  +author: Model.GuildMember | null,
+  +reactions: $ReadOnlyArray<GraphReaction>,
+  +mentions: $ReadOnlyArray<Model.GuildMember>,
+  // Included because we want the channel name in the node description.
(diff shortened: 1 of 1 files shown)


topocount (approved):
I really like these changes, @blueridger! This is significantly more readable now. I look forward to tests getting added later.

decentralion (approved):
For those interested (e.g. @topocount) this is kind of a Discord port of the following PR: https://github.com/sourcecred/sourcecred/pull/2049

PR #2469, opened 2020-11-07
rename `dependencyMintPolicy` to `bonusMinting`

decentralion:
This commit renames the dependencyMintPolicy module to bonusMinting, and
changes typenames and downstream usage accordingly. This is motivated,
most proximately, by my need to re-write this system to output a bonus
minting subgraph as part of the CredRank integration. (We want to have
dependency minting occur directly within the Graph, rather than adding
extra logic to the whole score computation pipeline.)

While writing the graph module, I found that the naming became a big
mess. The problem, I think, is that "DependencyMint" is a confusing
name: it describes the intended use case (paying dependencies) but not
the mechanism at hand (minting a bonus %age of total Cred that flows to
a particular recipient). I think renaming the mechanism to better
reflect its behavior will make the core logic more readable.

I've left the config API un-renamed, i.e. the config file is still
called dependnecies.json. That makes sense, since the intended use case
is still for minting Cred for dependencies.

Test plan: All existing tests pass. It's just a bunch of renames so that
suffices.

### DIFF
--- a/src/analysis/credData.js
+++ b/src/analysis/credData.js
@@ -1,10 +1,7 @@
 // @flow
 
 import type {TimelineCredScores} from "../core/algorithm/distributionToCred";
-import {
-  type DependencyMintPolicy,
-  processMintPolicy,
-} from "../core/dependenciesMintPolicy";
+import {type BonusPolicy, processBonusPolicy} from "../core/bonusMinting";
 import {type NodeAddressT, NodeAddress} from "../core/graph";
 import {IDENTITY_PREFIX} from "../core/identity";
 import {type IntervalSequence, intervalSequence} from "../core/interval";
@@ -74,7 +71,7 @@ export type EdgeCredOverTime = {|
 export function computeCredData(
   scores: TimelineCredScores,
   nodeOrder: $ReadOnlyArray<NodeAddressT>,
-  dependencyPolicies: $ReadOnlyArray<DependencyMintPolicy>
+  bonusPolicies: $ReadOnlyArray<BonusPolicy>
 ): CredData {
   const numIntervals = scores.length;
   if (numIntervals === 0) {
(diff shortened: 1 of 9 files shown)


src/analysis/credData.js
```diff
@@ -74,7 +71,7 @@ export type EdgeCredOverTime = {|
 export function computeCredData(
   scores: TimelineCredScores,
   nodeOrder: $ReadOnlyArray<NodeAddressT>,
-  dependencyPolicies: $ReadOnlyArray<DependencyMintPolicy>
+  dependencyPolicies: $ReadOnlyArray<BonusPolicy>
```
hammadj, line 74: Should rename the param to `bonusPolicies` as well

src/analysis/credData.js
```diff
@@ -88,7 +85,7 @@ export function computeCredData(
   }
   const intervals = intervalSequence(scores.map((d) => d.interval));
   const processedDependencyPolicies = dependencyPolicies.map((p) =>
```
hammadj, line 87: should be `processedBonusPolicies`

src/analysis/credData.test.js
```diff
@@ -97,7 +97,7 @@ describe("src/analysis/credData", () => {
         syntheticLoopFlow: new Float64Array([0.1, 0]),
       },
     ];
-    const mintPolicies: DependencyMintPolicy[] = [
+    const mintPolicies: BonusPolicy[] = [
```
hammadj, line 100:
```suggestion
    const bonusPolicies: BonusPolicy[] = [
```

hammadj (approved):
just a few renames missing

topocount (approved):

src/analysis/credData.js
decentralion: Done

src/analysis/credData.js
decentralion: Done

src/analysis/credData.test.js
decentralion: Done

PR #2479, opened 2020-11-07
grainConfig: implement Recent configuration

elihanover:
This adds configuration logic for the Recent policy to read from an instance's `config/grain.json` file.  Here, we add an optional `discount` parameter to the `GrainConfig` type, which checked for if there is a non-zero Recent policy.

This is followup to #2447

__Test Plan__
I've added some unit tests covering the `GrainConfig` parser.  I also added tests to cover errors related to missing a missing `discount`.

__Next__
I'd like to remove the hardcoding of policy names from this module.  Ideally, the `grainConfig` would read any valid policy that exist in `core/ledger/grainAllocation`.

### DIFF
--- a/src/api/grainConfig.js
+++ b/src/api/grainConfig.js
@@ -4,10 +4,13 @@ import {type DistributionPolicy} from "../core/ledger/applyDistributions";
 import * as G from "../core/ledger/grain";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
+import {toDiscount} from "../core/ledger/grainAllocation";
 
 export type GrainConfig = {|
   +immediatePerWeek: number,
   +balancedPerWeek: number,
+  +recentPerWeek: number,
+  +recentWeeklyDecayRate?: number,
   +maxSimultaneousDistributions?: number,
 |};
 
@@ -15,8 +18,10 @@ export const parser: C.Parser<GrainConfig> = C.object(
   {
     immediatePerWeek: C.number,
     balancedPerWeek: C.number,
+    recentPerWeek: C.number,
   },
   {
+    recentWeeklyDecayRate: C.number,
     maxSimultaneousDistributions: C.number,
   }
 );
@@ -25,6 +30,11 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
   if (!isNonnegativeInteger(x.immediatePerWeek)) {
     throw new Error(
       `immediate budget must be nonnegative integer, got ${x.immediatePerWeek}`
+    );
+  }
+  if (!isNonnegativeInteger(x.recentPerWeek)) {
+    throw new Error(
+      `recent budget must be nonnegative integer, got ${x.recentPerWeek}`
     );
   }
   if (!isNonnegativeInteger(x.balancedPerWeek)) {
@@ -37,6 +47,17 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
     allocationPolicies.push({
       budget: G.fromInteger(x.immediatePerWeek),
       policyType: "IMMEDIATE",
+    });
+  }
+  if (x.recentPerWeek > 0) {
+    const {recentWeeklyDecayRate} = x;
+    if (recentWeeklyDecayRate == null) {
+      throw new Error(`no recentWeeklyDecayRate specified for recent policy`);
+    }
+    allocationPolicies.push({
+      budget: G.fromInteger(x.recentPerWeek),
+      policyType: "RECENT",
+      discount: toDiscount(recentWeeklyDecayRate),
     });
   }
   if (x.balancedPerWeek > 0) {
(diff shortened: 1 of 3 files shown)


src/api/grainConfig.js
```diff
@@ -4,19 +4,24 @@ import {type DistributionPolicy} from "../core/ledger/applyDistributions";
 import * as G from "../core/ledger/grain";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
+import {toDiscount} from "../core/ledger/grainAllocation";
 
 export type GrainConfig = {|
   +immediatePerWeek: number,
   +balancedPerWeek: number,
+  +recentPerWeek: number,
+  +discount?: number,
```
decentralion, line 13: for clarity, let's rename to `recentPerWeekTimeDecayRate` or something. otherwise it's not at all clear what `discount` does, or that it's connected to the `recentPerWeek`

src/api/grainConfig.js
```diff
@@ -39,6 +49,17 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
       policyType: "IMMEDIATE",
     });
   }
+  if (x.recentPerWeek > 0) {
+    const discount = NullUtil.orElse(x.discount, undefined);
+    if (discount === undefined) {
+      throw new Error(`no discount specified for recent policy`);
+    }
```
decentralion, line 56: No need for NullUtil.orElse in this case, `if (discount == null)` is fine. Note that `undefined == null` and `null == null` are both true. It's the one acceptable use for the `==` operator.

src/api/grainConfig.js
```diff
@@ -39,6 +49,17 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
...
+    allocationPolicies.push({
+      budget: G.fromInteger(x.immediatePerWeek),
```
decentralion, line 58: there's a subtle and serious bug on this line. maybe consider adding tests to catch this issue. the only reason i don't insist on it is that i agree that we should refactor this config entirely.

decentralion (commented):
I agree, this `grainConfig` API kind of sucks. It'd be nice to see one that just allows any valid policies, not this weird extra API that wraps around the policies.

src/api/grainConfig.js
elihanover: Thanks for pointing this out.  Added another test to cover this and check values directly before/after.  In general this module is lacking on tests, so I'll try to populate it with more as I go along.

src/api/grainConfig.js
elihanover: Agreed. I implemented name change, but stuck with `discount` in the actual `RecentPolicy` type.  I think this confusion is indicative that there's probably a clearer way to do this.  I like the idea of adding policy "tickets".
```js
{
  policies: [
    "BALANCED": 0
    "IMMEDIATE": 0,
    "RECENT": {
       perWeek: 10,
       discount: 0.5
    }
  ],
]
```
or
```js
{
  policies: [
    {
       policyType: "BALANCED",
       perWeek: 10
    },
    {
       policyType: "IMMEDIATE",
       perWeek: 10
    },
    {
       policyType: "RECENT",
       perWeek: 10,
       discount: 0.5
    },
  ],
}
```
I also am strongly biased towards not having to include each policy as mandatory in the config.  This does not scale well.

src/api/grainConfig.js
```diff
@@ -4,19 +4,24 @@ import {type DistributionPolicy} from "../core/ledger/applyDistributions";
 import * as G from "../core/ledger/grain";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
+import {toDiscount} from "../core/ledger/grainAllocation";
 
 export type GrainConfig = {|
   +immediatePerWeek: number,
   +balancedPerWeek: number,
+  +recentPerWeek: number,
+  +recentWeeklyDecayRate?: number,
```
decentralion, line 13: Add a comment explaining what this parameter will do (e.g. will choosing 0.1 result in high or low decay? If I want "total decay" i.e. the immediate policy, would that correspond to 0 or to 1)?

src/api/grainConfig.test.js
```diff
@@ -0,0 +1,121 @@
+// @flow
+import {parser, toDistributionPolicy, type GrainConfig} from "./grainConfig";
```
decentralion, line 2: nit: we usually have a blankline after the flow pragma before the first import

decentralion (approved):
Thanks! Approved with small nits.

BTW, you should have write perms in this repo now, since I invited you to the team. Feel free to make branches on the main repo rather than PR'ing from forks.

src/api/grainConfig.js
elihanover: I added more detail in `src/core/ledger/grainAllocation.js` where the Recent policy is typed.  I think it would also be worthwhile to add details to the template instance, but I feel like the confusion around naming here is indicative of a larger problem of readability.  I plan on changing the grain config process to be clearer semantically in regards to typing.

src/api/grainConfig.test.js
elihanover: One nit a day makes the codebase go hooray

Issue #2450, opened 2020-11-03
InstanceManager: an organ of the LedgerService stack


hammadj:
## Abstract
On our journey towards an [identity system for sourcecred](https://github.com/sourcecred/sourcecred/issues/2371), and the Creditor, one of the first things we need to implement will be a way to programatically update the ledger in GitHub.

InstanceManager will manage the diffing and syncing of the ledger event log between GitHub and the in-memory ledger. It will ensure the correctness of the ledger by replaying local changes on top of the remote ledger in GitHub right before serializing and writing it back to GH. This helps keep things more [ACIDic](https://database.guide/what-is-acid-in-databases/).

## Prologue

There's a lot of problems we need to solve before we get to our ideal end to end integrated setup. In the spirit of atomic changes, this spec is detailing how we can handle persisting the ledger to GitHub programatically with good-enough safety to enable easier instance management by abstracting away git for the end user. This spec is not trying to solve for highly concurrent / frequent read/writes to the ledger, just iterating on the current single player instance management while having safety guarantees if there is any inconsistent state.

Not in scope:
- how we enable real time interaction with the ledger
- how we OAuth users or handle permissions
- how we cache data, or where this code is run
- how we construct commits or branches or PRs (would just be merging direct to master)

As we start building out the rest of the stack (account linking, self service admin, ETH withdrawals), the InstanceManager class will be composed by all those services who can use it to talk to GitHub. e.g. we could build a more realtime cache / db layer in between that uses this class to persist data back to GitHub while allowing for frequent concurrent read/writes at the application level.

## Benefits
- Ensures correctness of the ledger state, prevents any possibility of double spending or inconsistent data getting merged via git
- Better user experience: don't have to remember to pull changes from git, minimized probability of merge conflicts
- Gives more confidence for us to merge directly to master = less operational overhead
- Can be used on front end initially so we have something working ASAP, and then reused in our backend service once we build that out

## Spec

Here's some rough code:

```ts
class InstanceManager {
  ledger: Ledger
  _authToken: string
  _githubUrl: string
  
  // instantiates ledger from the ledger.json in GitHub
  constructor(githubUrl: string, authToken: string) {
    this._githubUrl = githubUrl;
    this._authToken = authToken;
  }
  
  // Queries and return latest ledger.json from GH
  _getRemoteLedgerLog(): Promise<LedgerLog>
  
  getUnprocessedEvents(remoteLog: LedgerLog): Promise<LedgerEvents[]> {
[...]

PR #2480, opened 2020-11-08
Implement LedgerManager class

hammadj:
Related issue: #2450

The LedgerManager will handle the diffing and syncing of Ledger event logs between an in-memory
ledger and a "remote ledger" through a generic LedgerStorage interface, allowing it to be persisted
to a variety of storage backends (e.g. GitHub, filesystem, database, etc).

Test Plan:
- Unit tests pass
- Interface makes sense for initial use case of GitHub storage

### DIFF
--- a/src/api/ledgerManager.js
+++ b/src/api/ledgerManager.js
@@ -87,7 +87,7 @@ export class LedgerManager {
         localChanges,
       };
     } catch (e) {
-      // Reset local ledger to remote state
+      // Set local ledger to the remote state + the local actions that were successfully applied
       this._ledger = remoteLedger;
       return {
         error: `Unable to apply local changes: ${e.message}, resetting to remote ledger`,
--- a/src/api/ledgerManager.js
+++ b/src/api/ledgerManager.js
@@ -1,8 +1,7 @@
 // @flow
-import {Ledger} from "../core/ledger/ledger";
-import type {LedgerLog} from "../core/ledger/ledger";
-import {diffLedger} from "../core/ledger/diffLedger";
-import type {LedgerDiff} from "../core/ledger/diffLedger";
+
+import {Ledger, type LedgerLog} from "../core/ledger/ledger";
+import {diffLedger, type LedgerDiff} from "../core/ledger/diffLedger";
 
 /**
  * Generic adaptor for persisting a Ledger to some storage backend
@@ -87,8 +86,8 @@ export class LedgerManager {
         localChanges,
       };
     } catch (e) {
-      // Set local ledger to the remote state + the local actions that were successfully applied
-      this._ledger = remoteLedger;
+      // Reset local ledger to the remote state
+      this._ledger = await this._storage.read();
       return {
         error: `Unable to apply local changes: ${e.message}, resetting to remote ledger`,
         remoteChanges,
--- a/src/api/ledgerManager.js
+++ b/src/api/ledgerManager.js
@@ -1,0 +1,99 @@
+// @flow
+import {Ledger} from "../core/ledger/ledger";
+import type {LedgerLog} from "../core/ledger/ledger";
+import {diffLedger} from "../core/ledger/diffLedger";
+import type {LedgerDiff} from "../core/ledger/diffLedger";
+
+/**
+ * Generic adaptor for persisting a Ledger to some storage backend
+ * (e.g. GitHub, local filesystem, a database, etc)
+ */
+export interface LedgerStorage {
+  read(): Promise<Ledger>;
+  write(ledger: Ledger): Promise<void>;
+}
+
+type LedgerManagerConfig = {|
+  +storage: LedgerStorage,
+  +initLogs?: LedgerLog,
+|};
+
+type ReloadResult = {|
+  +error: string | null,
+  +localChanges: LedgerDiff,
+  +remoteChanges: LedgerDiff,
+|};
+
+export class LedgerManager {
+  _ledger: Ledger;
+  +_storage: LedgerStorage;
+
+  constructor(config: LedgerManagerConfig) {
+    this._storage = config.storage;
+    this._ledger = config.initLogs
+      ? Ledger.fromEventLog(config.initLogs)
+      : new Ledger();
+  }
+
+  get ledger(): Ledger {
+    return this._ledger;
+  }
+
+  /**
+   * Returns a list of LedgerEvents that have not been persisted to storage yet
+   */
+  _getLocalChanges(remoteLedger: Ledger): LedgerDiff {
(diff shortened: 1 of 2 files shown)


src/api/ledgerManager.js
```diff
@@ -0,0 +1,112 @@
+// @flow
+import {Ledger} from "../core/ledger/ledger";
+import type {LedgerLog} from "../core/ledger/ledger";
+import diffBy from "lodash.differenceby";
+
+/**
+ * Generic adaptor for persisting a JSON log to some storage backend
+ * (e.g. GitHub, local filesystem, a database, etc)
```
decentralion, line 8: docstring says it's a generic adapter for persisting a JSON log, but the interface actually only describes ledgers.

src/api/ledgerManager.js
```diff
@@ -0,0 +1,112 @@
...
+ */
+export interface LedgerStorage {
+  read(): Promise<Ledger>;
+  write(ledger: Ledger): Promise<void>;
+}
+
+type LedgerManagerConfig = {|
+  +storage: LedgerStorage,
+  +initLogs?: LedgerLog,
```
decentralion, line 17: why is this argument optional, and why is it the log and not just a Ledger? smells funny to me

decentralion (commented):
a few thoughts:

1. We add logic for computing and applying ledger diffs. I'd like to see this factored out as clean APIs in the ledger module and robustly tested. In this commit, the functions aren't explicitly tested (e.g. no explicit testing for diffLedgerLog). Pulling out ledger diff support into its own commit would follow the "make the change easy, then make the change" development philosophy.

2. I think we should distinguish between LedgerLogs and LedgerDiffs. In my mind, we should call something a LedgerLog only if it actually corresponds to a self-consistent ledger. The results of a diff generally won't be a LedgerLog since you couldn't instantiate a Ledger from them. So we could make `LedgerLog` an opaque type, so you can't construct them except when they came directly from a valid Ledger.

src/api/ledgerManager.js
hammadj: The reason it's optional is because the Ledger manager can just load its state from the remote storage. The main reason having it here was more-so for testing purposes, but there could be more practical reasons to have it here (e.g. instantiating ledger instance from a users local ledger in case its not the same as the ledger persisted in GH)

The reason it takes the LedgerLog instead of the Ledger is because we want the LedgerManager to encapsulate its own instance of the ledger. We need to ensure that the user always goes through the LM when making changes instead of using their own external instance of a ledger since the reference to that instance would get thrown away by the LedgerManager when it does a re-load and there's new remote changes to sync. You can see in this line here that the internal ledger is reassigned to a new instance: https://github.com/sourcecred/sourcecred/pull/2480/files#diff-b1a73d3817f14a44e23c07abab8e24108919774aecf52e6cb8fa220668b7aaa9R96

Right now we still allow the user to make any method calls on the internal ledger instance, but in the future we could add specific API's that restrict ledger interaction to a subset of these, or add additional checks / verifications before forwarding events/actions to the underlying ledger.

src/api/ledgerManager.test.js
```diff
@@ -0,0 +1,280 @@
...
+      expect(res.localChanges).toEqual(expectedLocalChanges);
+
+      expect(manager.ledger.eventLog()).toEqual([
+        ...baseEventLog,
+        ...expectedRemoteChanges,
+        ...expectedLocalChanges,
+      ]);
+    });
+
+    it("should prevent replaying conflicting changes on remote ledger and return error", async () => {
```
blueridger, line 224: Due to other comment, would you add a test case where the first replay succeeds and the second replay fails?

blueridger (changes requested):

topocount (approved):

src/api/ledgerManager.js
```diff
@@ -0,0 +1,99 @@
...
+
+      this._ledger = remoteLedger;
+      return {
+        error: null,
+        remoteChanges,
+        localChanges,
+      };
+    } catch (e) {
+      // Reset local ledger to remote state
+      this._ledger = remoteLedger;
```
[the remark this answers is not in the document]
hammadj: Its intentional we keep the local changes that we can IMO, will update comment

blueridger (approved):

src/api/ledgerManager.js
```diff
@@ -0,0 +1,99 @@
+// @flow
```
decentralion, line 1: (nit, no change required): we usually have a newline after the flow pragma

src/api/ledgerManager.js
```diff
@@ -0,0 +1,99 @@
...
+        remoteChanges,
+        localChanges,
+      };
+    } catch (e) {
+      // Set local ledger to the remote state + the local actions that were successfully applied
+      this._ledger = remoteLedger;
+      return {
+        error: `Unable to apply local changes: ${e.message}, resetting to remote ledger`,
+        remoteChanges,
+        localChanges,
```
decentralion, line 95: Seems a little inconsistent; the error message says it's resetting to the remote ledger, but in actuality its resetting to remote ledger + a subset of local changes. Also, looking at the type, it's not obvious that some local changes applied and some didn't (potentially).

decentralion (commented):
I'm a little wary of the logic where when we do a reload and it creates conflicts, we apply some but not all of the local changes. Feels quite non-transactional and likely to create additional cognitive overhead for the end user, since now they can't follow the policy of "replay all my local changes", but instead need to think "ok, which of my local changes failed? I should replay only from that point forward...."

Is there a reason we don't just have the local changes all succeed or all fail transactionally?

hammadj:
> I'm a little wary of the logic where when we do a reload and it creates conflicts, we apply some but not all of the local changes. Feels quite non-transactional and likely to create additional cognitive overhead for the end user, since now they can't follow the policy of "replay all my local changes", but instead need to think "ok, which of my local changes failed? I should replay only from that point forward...."
> 
> Is there a reason we don't just have the local changes all succeed or all fail transactionally?

@decentralion Throwing away all the changes might be annoying to the user since they would have to repeat all their work again even for the actions that did succeed. The UI implementation will be the most important in terms of communicating effectively what changes the user is making. Applying these changes is still only local when reloading the ledger, so we can still do a diff with the remote ledger to show the user exactly what they are persisting to the remote ledger before we save it to GitHub.

However, if we feel that its better to just fail the whole thing and require users to repeat all the steps, happy to change it.

decentralion (approved):
Approving. I like the idea of starting with the more conservative behavior (don't partially apply the changes), open to re-visiting it once we have a functioning end-to-end example if it makes for a better UX.

PR #2481, opened 2020-11-08
LedgerManager persist logic

hammadj:
The persist method writes the local ledger changes back to the ledger storage while ensuring that there are no conflicts or inconsistencies in the ledger history as well as minimizing the potential for remote changes to get overwritten due to race conditions.

This is not super airtight or ready to be used my many people frequently updating the ledger at the same time, but it should cover the majority of sync conflicts that might arise.

Test Plan: Unit tests + test with github storage backend once that's implemented

### DIFF
--- a/src/api/ledgerManager.js
+++ b/src/api/ledgerManager.js
@@ -53,6 +53,47 @@ export class LedgerManager {
     return diffLedger(remoteLedger, this._ledger);
   }
 
+  /**
+   * Persists the local (in-memory) Ledger to the ledger storage. Reloads the
+   * remote ledger from storage right before persisting it to minimize the
+   * possibility of overwriting remote changes that were not synced to the local
+   * ledger and ensure consistency of the ledger events (e.g. no double spends).
+   *
+   * A race condition is present in this function: if client A runs reloadLedger
+   * and then client B writes to the remote ledger before client A finishes writing,
+   * then the changes to the ledger that client B made would be overwritten by the
+   * changes from client A. The correctness and consistency of the ledger will still
+   * be maintained, its just that client B might experience data loss of whatever
+   * events they were trying to sync. To detect if this has occurred, we reload
+   * the ledger again after writing the data to ensure the local changes were
+   * not overwritten. If they were, we can show an error message to client B with
+   * a list of changes that failed to sync.
+   */
+  async persist(): Promise<ReloadResult> {
+    // START RACE CONDITION
+    const preWriteReloadRes = await this.reloadLedger();
+    if (preWriteReloadRes.error) {
+      return preWriteReloadRes;
+    }
+    await this._storage.write(this._ledger);
+    // END RACE CONDITION
+
+    // Reload ledger again to ensure all the changes were persisted into storage
+    const postWriteReloadRes = await this.reloadLedger();
+    if (postWriteReloadRes.error) {
+      return postWriteReloadRes;
+    }
+
+    if (postWriteReloadRes.localChanges.length !== 0) {
+      return {
+        ...postWriteReloadRes,
+        error: "Some local changes have not been persisted",
+      };
+    }
+
+    return postWriteReloadRes;
+  }
+
   /** Reloads the persisted Ledger from storage and replays any local changes
    *  on top of any new remote changes, if they exist.
    *
--- a/src/api/ledgerManager.test.js
+++ b/src/api/ledgerManager.test.js
@@ -28,7 +28,10 @@ describe("api/ledgerManager", () => {
 
   const mockStorage = {
     read: jest.fn(() => Promise.resolve(new Ledger())),
-    write: jest.fn(() => Promise.resolve()),
+    write: jest.fn((ledger: Ledger) => {
+      setRemoteLedger(ledger);
+      return Promise.resolve();
+    }),
   };
 
   const setRemoteLedger = (remoteLedger: Ledger) => {
@@ -59,7 +62,8 @@ describe("api/ledgerManager", () => {
     id: uuid.random(),
   };
 
-  // fixture for testing sync conflicts
+  // fixture for testing sync conflicts. Remote Ledger contains an event that
+  // the local ledger in the manager does not
   const syncEventsFixture = () => {
     const baseLedger = ledgerWithIdentities();
     baseLedger.activate(id1);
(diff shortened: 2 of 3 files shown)


src/api/ledgerManager.js
```diff
@@ -67,6 +67,33 @@ export class LedgerManager {
...
+   */
+  async persist(): Promise<ReloadResult> {
+    let reloadRes = await this.reloadLedger();
+    if (reloadRes.error) {
+      return reloadRes;
+    }
+
+    // START RACE CONDITION
+    await this._storage.write(this._ledger);
+    // END RACE CONDITION
```
decentralion, line 84: It says the race condition starts and ends here, but doesn't it actually start on line 77 when we reload the ledger?

Suppose you have client A using this code. If client B writes any time between when the ledger was reloaded and when it gets written, it potentially triggers conflicts, yeah?

decentralion: Maybe write more detailed comments explaining what kind of race condition specifically exists on line 83?

src/api/ledgerManager.test.js
```diff
@@ -0,0 +1,382 @@
...
+
+  function ledgerWithIdentities() {
+    resetFakeUuid();
+    const ledger = new Ledger();
+    setNextUuid(id1);
+    ledger.createIdentity("USER", "steven");
+    setNextUuid(id2);
+    ledger.createIdentity("ORGANIZATION", "crystal-gems");
+    return ledger;
+  }
```
decentralion, line 40: I think I'm looking at diffs including the parent PR, maybe needs rebase?

src/api/ledgerManager.js
hammadj: Yea, you're right, that's how I intended to communicate it. Race condition starts after the ledger is reloaded, so I put the "START RACE CONDITION" comment after the reloadLedger command, but putting it before would be easier to understand. Will update!

hammadj: Added this comment in the code, reposting here for context:
```
   * A race condition is present in this function: if client A runs reloadLedger
   * and then client B writes to the remote ledger before client A finishes writing,
   * then the changes to the ledger that client B made would be overwritten by the
   * changes from client A. The correctness and consistency of the ledger will still
   * be maintained, its just that client B might experience data loss of whatever
   * events they were trying to sync. To detect if this has occurred, we reload
   * the ledger again after writing the data to ensure the local changes were
   * not overwritten. If they were, we can show an error message to client B with
   * a list of changes that failed to sync.
```

blueridger (approved):

src/api/ledgerManager.js
```diff
@@ -54,6 +54,42 @@ export class LedgerManager {
...
+
+    // Reload ledger again to ensure all the changes were persisted into storage
+    reloadRes = await this.reloadLedger();
+    if (reloadRes.localChanges.length !== 0) {
+      return {
+        ...reloadRes,
+        error: "Some local changes have not been persisted",
+      };
+    }
```
decentralion, line 89: if `reloadRes` has an error on this second load, then we are silently discarding the error message. I think we should check for error, explicitly return with the error if there was one, and only then try to see whether we had a problem with localChanges.

src/api/ledgerManager.js
```diff
@@ -54,6 +54,42 @@ export class LedgerManager {
...
+   * changes from client A. The correctness and consistency of the ledger will still
+   * be maintained, its just that client B might experience data loss of whatever
+   * events they were trying to sync. To detect if this has occurred, we reload
+   * the ledger again after writing the data to ensure the local changes were
+   * not overwritten. If they were, we can show an error message to client B with
+   * a list of changes that failed to sync.
+   */
+  async persist(): Promise<ReloadResult> {
+    // START RACE CONDITION
+    let reloadRes = await this.reloadLedger();
```
decentralion, line 75: recommend not re-assigning and just giving them separate names. e.g. `preWriteReload` and `postWriteReload`

src/api/ledgerManager.test.js
```diff
@@ -380,4 +397,89 @@ describe("api/ledgerManager", () => {
...
+
+      const res = await manager.persist();
+
+      expect(mockStorage.read).toBeCalledTimes(1);
+      expect(mockStorage.write).not.toBeCalled();
+      expect(res.error && res.error.indexOf("insufficient balance") > 0).toBe(
+        true
+      );
+      expect(res.remoteChanges).toHaveLength(1);
+      expect(res.localChanges).toHaveLength(1);
```
decentralion, line 426: Let's test the final state of the manager's ledger (after the persist)

src/api/ledgerManager.test.js
```diff
@@ -380,4 +397,89 @@ describe("api/ledgerManager", () => {
...
+        memo: "local transfer",
+      });
+
+      const res = await manager.persist();
+
+      expect(mockStorage.read).toBeCalledTimes(2);
+      expect(mockStorage.write).toBeCalledTimes(1);
+      expect(res.error).toBe(null);
+      expect(res.remoteChanges).toEqual([]);
+      expect(res.localChanges).toEqual([]);
```
decentralion, line 445: Let's test the final state of the manager's ledger (after the persist)

src/api/ledgerManager.test.js
```diff
@@ -380,4 +397,89 @@ describe("api/ledgerManager", () => {
...
+          action: {
+            type: "TRANSFER_GRAIN",
+            amount: "4",
+            memo: "local transfer",
+            from: id1,
+            to: id2,
+          },
+        },
+      ]);
+    });
```
decentralion, line 483: Let's test the final state of the manager's ledger (after the persist)

yarn.lock
```diff
@@ -7253,6 +7253,11 @@ nan@^2.12.1:
   resolved "https://registry.yarnpkg.com/nan/-/nan-2.14.1.tgz#d7be34dfa3105b91494c3147089315eff8874b01"
   integrity sha512-isWHgVjnFjh2x2yuJ/tj3JbwoHu3UC2dX5G/88Cm24yB6YopVgxvBObDY7n5xW6ExmFhJpSEQqFPvq9zaXc8Jw==
 
+nan@^2.14.0:
+  version "2.14.2"
+  resolved "https://registry.yarnpkg.com/nan/-/nan-2.14.2.tgz#f5376400695168f4cc694ac9393d0c9585eeea19"
+  integrity sha512-M2ufzIiINKCuDfBSAUr1vWQ+vuVcA9kqx8JJUsbQi6yf1uGRyb7HfpdfUr5qLXf3B/t8dPvcjhKMmlfnP47EzQ==
+
```
decentralion, line 7260: nb. i think this will disappear when rebased

decentralion (approved):

PR #2483, opened 2020-11-08
grainAllocation: inject optional mapping over cred scores

elihanover:
This commit allows injecting an optional `CredMapping :: CredScores => CredScores` into the receipt functions for each policy type (besides special by design).  In practice, we will use this for calculating grain distributions based on quadratically weighted cred; however, it lets us add any `CredScores` garnish more generally.  This makes policy building easier as we can tweak policies and their parameters on the fly without changing the underlying types.  This should also speed up any future cryptoeconomics research.

The obvious option is to create a new policy for each, but modifying policies does not scale will in this case.  Instead, injecting a more general mapping (which defaults to `id`) allows us to use the core logic of the core `Balanced`, `Immediate`, and `Recent` policies.

__Test Plan__
Here, I've added unit tests that test this by injecting a function to quadratically weight cred.

### DIFF
--- a/src/core/ledger/grain.js
+++ b/src/core/ledger/grain.js
@@ -60,6 +60,9 @@ export function mul(a: Grain, b: Grain): Grain {
 export function div(a: Grain, b: Grain): Grain {
   return bigInt(a).divide(bigInt(b)).toString();
 }
+export function pow(a: Grain, b: Grain): Grain {
+  return bigInt(a).pow(bigInt(b)).toString();
+}
 export function lt(a: Grain, b: Grain): boolean {
   return bigInt(a).lt(bigInt(b));
 }
--- a/src/core/ledger/grainAllocation.js
+++ b/src/core/ledger/grainAllocation.js
@@ -66,17 +66,20 @@ export type AllocationPolicy =
 export type BalancedPolicy = {|
   +policyType: Balanced,
   +budget: G.Grain,
+  +credMapping?: CredMapping,
 |};
 
 export type ImmediatePolicy = {|
   +policyType: Immediate,
   +budget: G.Grain,
+  +credMapping?: CredMapping,
 |};
 
 export type RecentPolicy = {|
   +policyType: Recent,
   +budget: G.Grain,
   +discount: Discount,
+  +credMapping?: CredMapping,
 |};
 
 export type SpecialPolicy = {|
@@ -85,6 +88,9 @@ export type SpecialPolicy = {|
   +memo: string,
   +recipient: IdentityId,
 |};
+
+export type CredScores = $ReadOnlyArray<number>;
+export type CredMapping = (CredScores) => CredScores;
 
 export type GrainReceipt = {|
   +id: IdentityId,
@@ -98,7 +104,7 @@ export type Allocation = {|
 |};
 
 export type AllocationIdentity = {|
-  +cred: $ReadOnlyArray<number>,
+  +cred: CredScores,
   +paid: G.Grain,
   +id: IdentityId,
 |};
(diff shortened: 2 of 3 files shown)


src/core/ledger/grainAllocation.js
```diff
@@ -66,17 +66,20 @@ export type AllocationPolicy =
 export type BalancedPolicy = {|
   +policyType: Balanced,
   +budget: G.Grain,
+  +mapping?: Function,
```
hammadj, line 69: Should add type information for the inputs / outputs of this function for typesafety

src/core/ledger/grainAllocation.js
```diff
@@ -66,17 +66,20 @@ export type AllocationPolicy =
...
 |};
 
 export type ImmediatePolicy = {|
   +policyType: Immediate,
   +budget: G.Grain,
+  +mapping?: Function,
```
hammadj, line 75: The name "mapping" isn't 100% clear, leads me to believe that this is a mapping for grain amounts, but seems like its actually a `credMapping`

src/core/ledger/grainAllocation.js
```diff
@@ -214,12 +222,13 @@ function receipts(
...
+  identities: ProcessedIdentities,
+  mapping?: Function = (id) => id
 ): $ReadOnlyArray<GrainReceipt> {
-  const amounts = G.splitBudget(
-    budget,
-    identities.map((i) => i.mostRecentCred)
-  );
+  const credPerIdentity = identities.map((i) => i.mostRecentCred);
+  // if (!mapping) mapping = (id) => id
+  const modifiedCred = credPerIdentity.map(mapping);
```
hammadj, line 230: I think the mapping function should accept the entire array of cred scores instead of each individual cred score. That way it can have context for what the entire range of values are and allow for more advanced mapping strategies (e.g. bell curve)

src/core/ledger/grain.js
```diff
@@ -60,6 +60,9 @@ export function mul(a: Grain, b: Grain): Grain {
 export function div(a: Grain, b: Grain): Grain {
   return bigInt(a).divide(bigInt(b)).toString();
 }
+export function pow(a: Grain, b: Grain): Grain {
+  return bigInt(a).pow(bigInt(b)).toString();
+}
```
decentralion, line 65: This doesn't seem right. Why would we be raising something to a `Grain` power? As an example, let `one` be one grain. Then `Grain.power(one, one)` is `(10^18) ^ (10^18)`. Also this forbids doing non-integer powers which frankly are the powers I"m most interested in (e.g. sqrt(x) is pow(x, 0.5)).

Also: if adding a new mathematical primitive to the grain module, please make that into its own PR.

src/core/ledger/grainAllocation.js
```diff
@@ -66,17 +66,20 @@ export type AllocationPolicy =
...
   +policyType: Immediate,
   +budget: G.Grain,
+  +credMapping?: CredMapping,
 |};
 
 export type RecentPolicy = {|
   +policyType: Recent,
   +budget: G.Grain,
   +discount: Discount,
+  +credMapping?: CredMapping,
```
decentralion, line 82: This is a no-go. The Policies need to be serializable (they are included in the ledger event), but functions are not serializable.

decentralion (changes requested):
Requesting changes because we can't have arbitrary functions in the Policy types, because those types all get serialized inside the ledger, and functions can't be serialized. Need to take a different approach.

src/core/ledger/grainAllocation.js
elihanover: Ok, you had mentioned this previously but I had forgotten.  I'll close this PR.

PR #2488, opened 2020-11-10
Refactor computeCredAccounts to function in the empty case

topocount:
The public computeCredAccounts command would fail if no intervals had
been computed yet, because userNodes were always assumed to contain some
cred values. This refactoring allows for computeCredAccounts to succeed
without any credIntervals needing to be created.

This is needed to allow the template instance to successfully run `yarn
go` out of the box (even if it doesn't do anything substantial without
plugins)

test plan: The outer function is untested, but the inner function has
been updated to demonstrate support for both the empty case and the
"default case", meaning the case in the template instance

### DIFF
--- a/src/core/ledger/credAccounts.js
+++ b/src/core/ledger/credAccounts.js
@@ -47,15 +47,19 @@ export function computeCredAccounts(
 ): CredAccountData {
   const grainAccounts = ledger.accounts();
   const userlikeInfo = new Map();
+  const intervals = credView.intervals();
+  const noIntervals = intervals.length === 0;
   for (const {address, credOverTime, description} of credView.userNodes()) {
-    if (credOverTime == null) {
+    if (noIntervals) {
+      userlikeInfo.set(address, {cred: [], description});
+    } else if (credOverTime === null) {
       throw new Error(
         `userlike ${NodeAddress.toString(address)} does not have detailed cred`
       );
+    } else {
+      userlikeInfo.set(address, {cred: credOverTime.cred, description});
     }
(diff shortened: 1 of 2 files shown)


decentralion (approved):

src/core/ledger/credAccounts.js
```diff
@@ -47,15 +47,19 @@ export function computeCredAccounts(
...
   for (const {address, credOverTime, description} of credView.userNodes()) {
-    if (credOverTime == null) {
+    if (noIntervals) {
+      userlikeInfo.set(address, {cred: [], description});
+    } else if (credOverTime == null) {
       throw new Error(
         `userlike ${NodeAddress.toString(address)} does not have detailed cred`
       );
+    } else {
+      userlikeInfo.set(address, {cred: credOverTime.cred || null, description});
```
blueridger, line 60: null case is untested, and for good reason. The downstream typing doesn't accept null, and the upstream credOverTime.cred attribute must be defined according to typing (which is why Flow doesn't fail this, because it knows  `|| null` will always be short circuited).

All that to say, you should be able to remove ` || null` and you probably oughta to avoid confusion.

blueridger (changes requested):

src/core/ledger/credAccounts.js
topocount: Good catch @blueridger ! This was dangling from an earlier iteration and I forgot to remove it.

blueridger: That was my guess :)

blueridger (approved):

hammadj (approved):

PR #2496, opened 2020-11-11
Add PR template

blueridger:
# Notes
https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository
# Manual Testing
Pasted it into this description and hit preview. Saw the headers as expected.
After merge, I will test that it is automatically loaded into a new PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,17 @@
+<!-- Please read our contributor guide before submitting your PR: -->
+<!-- https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md -->
+
+# Description
+
+<!-- Possible prompts:
+  Why is the change important?
+  Did you consider and reject alternate formulations of the same idea?
+  Are there relevant issues or discussions elsewhere?
+-->
+
+# Test Plan
+
+<!-- Possible prompts:
+  How well covered by automated tests is this change?
+  How did you manually test this change? How can a reviewer replicate the tests?
+-->


wchargin (changes requested):
We have had this in the past, but removed it (#618). Was there recent
discussion about adding it again? Please explain reasoning. Thanks!

Briefly, my preference for not having a template: it creates a huge,
40-line block in the PR description, after the commit message, that you
have to manually delete every time. It does this unconditionally, even
if your commit message already explains what’s asked for in the
template.

blueridger:
@wchargin We discussed it during the Dev meeting this week, and we aligned on adding it. Reasoning: DL and William's personal workflow is to add the info to the commit message, so not having it made sense when DL and William were doing most of the dev work. However, the org is transitioning to more decentralized development, so having a template makes sense to encourage best practices across a larger and looser set of developers. DL agreed to take the minor hit on their own workflow in order to improve PR quality organizationally.

wchargin (approved):
> We discussed it during the Dev meeting this week, and we aligned on adding it.

Okay, cool; thanks!

One question about the template itself: what is the “revision history”
section for? I’m not used to seeing people track this (and I don’t track
it myself).

blueridger:
> One question about the template itself: what is the “revision history”
> section for?

We do this at my old job. It's to provide a checkpoint to mitigate people making a change and then quietly cutting corners on re-testing. It also helps maintain context and let reviewers know what the diff of the diffs is. Though, our code review infrastructure actually had the idea of discrete revisions embedded into the ui, so it was more intuitive. I wouldn't mind leaving it out, since it isn't intuitive here.

decentralion (commented):
At present, this template is inconsistent with our [contributing guidelines](https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md). Overall, I quite like our contributing guidelines, so I request that this doc be changed to:

1. link to the contributing guidelines at the top
2. match the expectations set in the contributing guidelines (e.g. a single test plan, no revision history, etc)
3. tone match the contributing guidelines (i.e. less terse, more friendly and inviting)

If you think we should be making changes to the contributing workflow, let's either make those changes to the contributing guidelines first (and then merge the template), or merge a template that matches the guidelines and then update the two in tandem.

blueridger:
Updated to mirror contributing guidelines.

decentralion (approved):

topocount (approved):

wchargin (commented):
Thanks! I like this revised template a lot more.

PR #2502, opened 2020-11-11
Create Generic CurrencyId type

topocount:
A generic CurrencyId will provide support for non-evm protocols once we
support their address formats. For EVM-based blockchains, the numeric
chainId is stored as part of the key, along with the address of the
currency paid out as grain.

Prerequisite for #2429 

test plan: unit tests have been included for parsing logic

### DIFF
--- a/src/core/ledger/currency.js
+++ b/src/core/ledger/currency.js
@@ -1,0 +1,84 @@
+// @flow
+
+import {
+  ethAddressParser,
+  type EthAddress,
+} from "../../plugins/ethereum/ethAddress";
+import * as C from "../../util/combo";
+
+/**
+ * EvmChainId is represented in the form of a stringified integer for all
+ * EVM-based chains, including mainnet (1), and xDai (100).
+ * The reason for this is that ethereum's client configuration utilizes
+ * a number to represent chainId, and this way we can just transpose that
+ * chainId here as a component of the currency Id, since the web3 client will
+ * return a stringified integer when the chainId is requested.
+ */
+export opaque type EvmChainId: string = string;
+
+export function parseEvmChainId(id: string): EvmChainId {
+  const result = parseInt(id, 10).toString();
+  if (Number.isNaN(result) || result !== id) {
+    throw new Error(`Invalid EVM chainId value: ${id}`);
+  }
+  return id;
+}
+
+export type EvmId = {|
+  +type: "EVM",
+  +chainId: EvmChainId,
+  /**
+   * CurrencyAddress is a subset of all available EthAddresses.
+   *
+   * A currency address is the address of the token contract for an ERC20 token,
+   * or the 20 byte-length equivalent of 0x0, which is the conventional address
+   * used to represent ETH on the ethereum mainnet, or the native currency on an
+   * EVM-based sidechain see here for more details on these semantics:
+   * https://ethereum.org/en/developers/docs/intro-to-ethereum/#eth
+   */
+  +currencyAddress: EthAddress,
+|};
+
+/**
+ * Example protocol symbols: "BTC" for bitcoin and "FIL" for Filecoin
+ */
+export opaque type ProtocolSymbol: string = string;
+
+const protocolSymbolParser: C.Parser<ProtocolSymbol> = C.exactly([
+  "BTC",
+  "FIL",
+]);
+
+/**
+ * Chains like Bitcoin and Filecoin do not have "production" sidechains so
+ * we represent them as a string, as specified in the ProtocolSymbol type
+ */
+export type ProtocolId = {|
+  +type: "PROTOCOL",
+  +chainId: ProtocolSymbol,
+|};
+
+export type ChainId = EvmChainId | ProtocolSymbol;
+export type CurrencyId = EvmId | ProtocolId;
+
+export const evmChainParser: C.Parser<EvmChainId> = C.fmap(
+  C.string,
+  parseEvmChainId
+);
+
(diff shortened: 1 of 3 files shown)


src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+export type EvmChainId = string;
+
+export function parseEvmChainId(id: string): EvmChainId {
+  const result = parseInt(id, 10).toString();
+  if (Number.isNaN(result) || result !== id) {
+    throw new Error(`Invalid EVM chainId value: ${id}`);
+  }
+  return id;
+}
```
decentralion, line 25: if you're gonna add a parser, let's make the `EvmChainId` an opaque subtype of string, so as to get real type safety.

`export opaque type EvmChainId: string = string`

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+  parseEvmChainId
+);
+
+declare type EvmId = {
+  chainId: EvmChainId,
+  /**
+   * CurrencyAddress is a subset of all available EthAddresses.
+   *
+   * A currency address is the address of the token contract for an ERC20 token,
+   * or the 20 byte-length equivalent of 0x0, which is the conventional address
```
decentralion, line 38: Can we export the 20-byte equivalent of 0x as a constant?

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+
+export const evmIdParser: C.Parser<EvmId> = C.object({
+  chainId: evmChainParser,
+  currencyAddress: ethAddressParser,
+});
+
+/**
+ * Example protocol symbols: "BTC" for bitcoin and "FIL" for Filecoin
+ */
+declare type ProtocolSymbol = string;
```
decentralion, line 54: hmm why do we use `declare type`? I'm not used to seeing `declare`. are we not exporting it?

also, since we have a parser --let's make it opaque like above?

src/core/ledger/currency.js
```diff
@@ -0,0 +1,79 @@
...
+  {
+    chainId: C.exactly(["BTC", "FIL"]),
+  },
+  {
+    currencyAddress: C.pure<void>(),
+  }
+);
+
+export type ChainId = ProtocolSymbol | EvmChainId;
+export type CurrencyId = ProtocolId | EvmId;
```
decentralion, line 76: At runtime, how can a consumer tell if they have a `ProtocolId` or a `EvmId`?

topocount: I've added a C.orElse parser to cycle through the two, even though it's not strictly necessary yet.

src/core/ledger/currency.js
```diff
@@ -0,0 +1,84 @@
...
+
+import {
+  ethAddressParser,
+  type EthAddress,
+} from "../../plugins/ethereum/ethAddress";
+import * as C from "../../util/combo";
+
+/**
+ * EvmChainId is represented in number form for all EVM-based chains,
+ * including mainnet (1), and xDai (100)
```
decentralion, line 11: Docstring suggests it's represented in number form, but is actually a string? By "number form" do you mean "as a stringified number"?

src/core/ledger/currency.js
decentralion: Marked resolved but not done? Is there a reason this is a bad idea? (n.b. when you resolve a comment please leave a comment explaining what the resolution is)

EDIT: i see this was added in another file. request to explain what a comment's resolution is stands. (e.g. "done, added in other file")

src/core/ledger/currency.js
```diff
@@ -0,0 +1,84 @@
...
+
+export const protocolIdParser: C.Parser<ProtocolId> = C.object({
+  type: C.exactly(["PROTOCOL"]),
+  chainId: protocolSymbolParser,
+});
+
+// @topocount TODO: enable protocolIdParser once we support its address types
+export const currencyIdParser: C.Parser<CurrencyId> = C.orElse([
+  evmIdParser,
+  //protocolIdParser,
```
decentralion, line 83: seeing as this has a return type of `CurrencyId`, all clients will need to handle the protocolId case for typechecking, even if we never actually parse them. so we may as well parse them here, there's no harm in doing so.

src/ui/server.js
```diff
@@ -36,6 +36,7 @@ export default function render(
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width,initial-scale=1" />
       <link rel="shortcut icon" href="${assets.resolve("/favicon.png")}" />
+      <link rel="manifest" href="${assets.resolve("/manifest.json")}" />
```
decentralion, line 39: Why is this a part of this PR?

src/core/ledger/currency.js
topocount: Thanks, will always add resolution notes going forward.

src/ui/server.js
topocount: I don't know. I have no memory of adding this line, or what it does.

src/core/ledger/currency.js
topocount: Fair enough. added

src/core/ledger/currency.js
topocount: I went back and forth on how to describe this, and decided (wrongly) to let the code demonstrate what was meant. Thanks for the suggestion.

decentralion (approved):
Seems reasonable. May need some changes down the line (i don't know the space well enough to be confident), but what I *am* confident in is our ability to refactor as needed later.

Issue #2491, opened 2020-11-10
Discourse plugin failing with 520 error


s-ben:
When re-calculating scores for the [Maker instance](https://github.com/sourcecred/makerdao-cred), the Discord plugin keeps failing with a 520 error.

```
Ss-MacBook-Pro:makerdao-cred sben$ yarn; yarn load; yarn graph; yarn score; yarn site
yarn install v1.22.4
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.32s.
yarn run v1.22.4
$ sourcecred load
  GO   load
  GO   loading sourcecred/discourse
  GO   sourcecred/discourse: discourse
  GO   sourcecred/discourse: discourse/topics
 DONE  sourcecred/discourse: discourse: 22s
 WARN  sourcecred/discourse: discourse: Parent task restarting. Retrying
 DONE  sourcecred/discourse: discourse/topics: 22s
 WARN  sourcecred/discourse: discourse/topics: Parent task restarting. Retrying
 DONE  loading sourcecred/discourse: 22s
 WARN  loading sourcecred/discourse: Error updating cache. clearing cache and restarting.
        This is the error from the plugin:

        Error: not OK status 520 on https://forum.makerdao.com/latest.json?order=activity&ascending=false&page=20
  GO   loading sourcecred/discourse
  GO   sourcecred/discourse: discourse
  GO   sourcecred/discourse: discourse/topics
 FAIL  sourcecred/discourse: Error: not OK status 520 on https://forum.makerdao.com/t/359.json
 DONE  load: 2m 27s
fatal: load failed for plugins: sourcecred/discourse
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
yarn run v1.22.4
```

Have tried re-running probably a dozen times now. Always with the same result. The URL that fails is different every time. Usually it's a topic, but sometimes a page of topics as shown in the above example. Every URL that I've pasted into the browser returns a live, working page (or JSON in the case of the URL above). Have tried deleting the repo and cloning a fresh one, clearing cache with `yarn clean`, still get the same error.

topocount:
fixed in #2503

PR #2503, opened 2020-11-11
discourse: retry on 520 errors

decentralion:
We've started seeing ephemeral 520 errors from Discourse (see #2491).
520 is a catch-all for server issues, and the errors are not
reproducible, so I think the issue is just flaky-ness on Discourse's
end.

As such, I added retry logic. Whenever we hit a 520 error, we retry 3
times before failing. In testing, this makes it possible to successfully
load large Discourse instances, e.g. Maker.

Test plan:

**Manual**: Loading the MakerDAO instance now succeeds.
**Automated**: Added unit tests verifying that we try on 520 errors.

### DIFF
--- a/src/plugins/discourse/fetch.test.js
+++ b/src/plugins/discourse/fetch.test.js
@@ -40,6 +40,37 @@ describe("plugins/discourse/fetch", () => {
     };
     const fetcherWithStatus = (status: number) =>
       new Fetcher(options, fakeFetch(status), 0);
+
+    it("retries on 520", async () => {
+      const mockFetch = jest.fn();
+      mockFetch
+        .mockReturnValueOnce(
+          Promise.resolve(new Response("", {status: 520, url: "/something"}))
+        )
+        .mockReturnValueOnce(
+          Promise.resolve(new Response("", {status: 520, url: "/something"}))
+        )
+        .mockReturnValue(
+          Promise.resolve(new Response("", {status: 200, url: "/something"}))
+        );
+
+      const fetcher = new Fetcher(options, mockFetch, 0);
+      const response = await fetcher._fetch("/something");
+
+      expect(response.status).toEqual(200);
+      expect(mockFetch.mock.calls.length).toEqual(3);
+    });
+
+    it("errors after repeated 520 errors", async () => {
+      const mockFetch = jest.fn();
+      mockFetch.mockReturnValue(
+        Promise.resolve(new Response("", {status: 520, url: "/something"}))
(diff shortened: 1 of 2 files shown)


src/plugins/discourse/fetch.test.js
```diff
@@ -40,6 +40,30 @@ describe("plugins/discourse/fetch", () => {
...
       new Fetcher(options, fakeFetch(status), 0);
+
+    it("retries on 520", async () => {
+      let nTries = 0;
+      const fetch = (url) => {
+        const status = nTries < 2 ? 520 : 200;
+        nTries++;
+        const response = new Response("", {status, url});
+        return Promise.resolve(response);
+      };
```
blueridger, line 51: This is a little more complicated than what I would call best practice for a unit test. Generally unit tests should not have any logic (Quis custodiet ipsos custodes? -- who's going to test the test?).  I'd recommend using [Jest's built in capabilities](https://jestjs.io/docs/en/mock-functions), since it's already a dependency.
```
let mockFetch = jest.fn();
// various useful functions you could use for this case:
mockFetch
  .mockReturnValueOnce(Promise.resolve(new Response("", {status: 520, "/something"})))
  .mockImplementationOnce(url => Promise.resolve(new Response("", {status: 520, url})))
  .mockReturnValue(Promise.resolve(new Response("", {status: 200, "/something"})))

const fetcher = new Fetcher(options, mockFetch, 0);
const response = await fetcher._fetch("/something");

expect(response.status).toEqual(200);
expect(mockFetch.mock.calls.length).toEqual(3);
```

decentralion: Done, thanks for the suggestion.

topocount (approved):
Thanks Dandelion!

src/plugins/discourse/fetch.test.js
```diff
@@ -40,6 +40,37 @@ describe("plugins/discourse/fetch", () => {
...
+        )
+        .mockReturnValue(
+          Promise.resolve(new Response("", {status: 200, url: "/something"}))
+        );
+
+      const fetcher = new Fetcher(options, mockFetch, 0);
+      const response = await fetcher._fetch("/something");
+
+      expect(response.status).toEqual(200);
+      expect(mockFetch.mock.calls.length).toEqual(3);
```
blueridger, line 61: nit: I'd say this assertion is even more valuable in the failure case.

blueridger (approved):
Sweet!

PR #2504, opened 2020-11-12
Implement function for diffing event histories of two Ledgers

hammadj:
This diffLedger function will be useful for syncing two different Ledgers and improving
the front end UX for showing what changes a user has made to their local Ledger. It will
be used in the LedgerManager class to help avoid merge conflicts and breaking ledger entries
for self-service admin.

Test Plan: Wrote unit tests to verify functionality

### DIFF
--- a/src/core/ledger/diffLedger.js
+++ b/src/core/ledger/diffLedger.js
@@ -1,0 +1,28 @@
+// @flow
+
+import {Ledger} from "./ledger";
+import type {LedgerEvent} from "./ledger";
+import diffBy from "lodash.differenceby";
+
+export type LedgerDiff = $ReadOnlyArray<LedgerEvent>;
+
+/**
+ * Returns an array of ledger events that exist in ledger "a" but not in "b".
+ * An event is considered equal to another if it has the same uuid.
+ *
+ * This will not return any events from "b" that don't exist in "a", so the
+ * order of the params matters.
+ *
+ * Example 1:
+ *  - Ledger A: [1, 2, 3]
+ *  - Ledger B: [1, 3, 4, 5]
+ *  - Returns: [2]
+ *
+ * Example 2:
+ *  - Ledger A: [1, 3, 4, 5]
+ *  - Ledger B: [1, 2, 3]
+ *  - Returns: [4, 5]
+ */
+export function diffLedger(a: Ledger, b: Ledger): LedgerDiff {
+  return diffBy(a.eventLog(), b.eventLog(), "uuid");
+}
--- a/src/core/ledger/diffLedger.test.js
+++ b/src/core/ledger/diffLedger.test.js
@@ -1,0 +1,99 @@
+// @flow
+
+import {diffLedger} from "./diffLedger";
+import * as uuid from "../../util/uuid";
+import {Ledger} from "./ledger"; // for spy purposes
+
+describe("core/ledger/diffLedger", () => {
+  const randomMock = jest.spyOn(uuid, "random");
+
+  let nextFakeUuidIndex = 0;
+  function resetFakeUuid() {
+    nextFakeUuidIndex = 0;
+  }
+  function nextFakeUuid(): uuid.Uuid {
+    const uuidString = String(nextFakeUuidIndex).padStart(21, "0") + "A";
+    nextFakeUuidIndex++;
+    return uuid.fromString(uuidString);
+  }
+
+  randomMock.mockImplementation(nextFakeUuid);
+  function setNextUuid(x: uuid.Uuid) {
+    randomMock.mockImplementationOnce(() => x);
+  }
+
+  const id1 = uuid.fromString("YVZhbGlkVXVpZEF0TGFzdA");
+  const id2 = uuid.fromString("URgLrCxgvjHxtGJ9PgmckQ");
+
+  function ledgerWithIdentities() {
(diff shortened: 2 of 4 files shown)


src/core/ledger/diffLedger.js
```diff
@@ -0,0 +1,28 @@
...
+
+import {Ledger} from "./ledger";
+import type {LedgerEvent} from "./ledger";
+import diffBy from "lodash.differenceby";
+
+export opaque type LedgerDiff = $ReadOnlyArray<LedgerEvent>;
+
+/**
+ * Returns an array of ledger events that exist in ledger "a" but not in "b".
+ * An event is considered equal to another if it has the same uuid.
```
decentralion, line 11: What is the events have the same uuid but are distinct? It in principle, it shouldn't happen, but it definitely COULD happen, and it could happen maliciously (e.g. attacker pushes a commit to the ledger which modifies a real transfer to transfer to the attacker instead of the intended recipient).

src/core/ledger/diffLedger.test.js
```diff
@@ -0,0 +1,99 @@
...
+  function nextFakeUuid(): uuid.Uuid {
+    const uuidString = String(nextFakeUuidIndex).padStart(21, "0") + "A";
+    nextFakeUuidIndex++;
+    return uuid.fromString(uuidString);
+  }
+
+  randomMock.mockImplementation(nextFakeUuid);
+  function setNextUuid(x: uuid.Uuid) {
+    randomMock.mockImplementationOnce(() => x);
+  }
```
decentralion, line 23: I believe this is copy-pasta'd from ledger. Would be great to have a separate PR which just creates a `fakeUuid` module (or `uuidTestUtil`) which exports this functionality.

src/core/ledger/diffLedger.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Example 1:
+ *  - Ledger A: [1, 2, 3]
+ *  - Ledger B: [1, 3, 4, 5]
+ *  - Returns: [2]
+ *
+ * Example 2:
+ *  - Ledger A: [1, 3, 4, 5]
+ *  - Ledger B: [1, 2, 3]
+ *  - Returns: [4, 5]
```
decentralion, line 24: I appreciate the examples

src/core/ledger/diffLedger.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import {Ledger} from "./ledger";
+import type {LedgerEvent} from "./ledger";
+import diffBy from "lodash.differenceby";
+
+export opaque type LedgerDiff = $ReadOnlyArray<LedgerEvent>;
```
decentralion, line 7: Is there a reason we make the API asymmetrical? What if we did something like:

```js
export type LedgerDiff = {|
  +onlyInFirstLedger: $ReadOnlyArray<LedgerEvent>,
  +onlyInSecondLedger: $ReadOnlyArray<LedgerEvent>,
  // Events that had the same UUID but were not equal
  +uuidConflicts: $ReadOnlyArray<[LedgerEvent, LedgerEvent]>,
|}

File: src/core/ledger/diffLedger.js (hunk 1 above)
Comment (hammadj) [reply src/core/ledger/diffLedger.js:11]: If someone has permissions to write events directly to the ledger in GitHub, there's no way for us to actually prevent them from exploiting it. They don't have to modify an existing transfer, they can just do new transfer events to their own grain account.

Those type of protections can be put in place once we have a LedgerService backend running that's taking authenticated user actions (limited to a subset of all possible ledger actions) and modifying the ledger accordingly. Users won't be writing to the ledger directly or passing in the entire ledger history, they will only be allowed to pass in discrete actions.

If we really want to do this properly the uuid should be a "transaction hash" of the ledger action payload and a nonce. Then we just add a consensus protocol on top for a decentralized network of LedgerService nodes that's verifying the validity of the transaction. Then we do a big ICO for CredChain.

Or we just implement this using an existing blockchain when we want to do it properly. [Matic](https://matic.network) is a great option for high scalability / cheap tx L2 that works today and is being used in prod for many big projects. Most recently I found out that https://polymarket.com is using Matic and has strong mainstream adoption too, so UX is definitely there. Don't think we need to take this step yet though, but at a certain point once we are trying for proper hard guarantees and safety, it makes sense to cross the bridge.

File: src/core/ledger/diffLedger.js (hunk 4 above)
Comment (hammadj) [reply src/core/ledger/diffLedger.js]: Yea that might be handy. Don't think we need the uuidConflicts as per my comment above, which would allow us to return a tuple for some smooth ergonomics:

```ts
export type LedgerDiff = [$ReadOnlyArray<LedgerEvent>, $ReadOnlyArray<LedgerEvent>];

// In ledgerManager:
const [localChanges, remoteChanges] = diffLedger(this._ledger, remoteLedger);
```

src/core/ledger/diffLedger.test.js
hammadj: Done in this PR: https://github.com/sourcecred/sourcecred/pull/2505

src/core/ledger/diffLedger.js
decentralion: Makes sense, thanks.

decentralion (approved):

PR #2505, opened 2020-11-12
Consolidate uuid mocking in tests to createUuidMock function

hammadj:
Extracted the commonly used uuid mocks to a single helper function that can be used across all
the tests that need to mock out uuid generation

Test Plan: Ensure unit tests pass


decentralion (approved):
I would have made this commit the first commit (i.e. going directly to master) and then rebased the other work to use it. Falls into the pattern of "make the change easy, then make the change", and simplifies the workstream since this one is super easy to approve and merge, no risk of needing to rebase it again if the ledger-manager changes.

That said, both merge strategies are OK. Approved.

topocount (approved):
It might make sense to consolidate test utils to one location in the repo at some point down the road but I think this is right where it needs to be for now.

PR #2506, opened 2020-11-12
Release version 0.7.3

topocount:
test plan: Ran some canary testing on the template and prod sourcecred instances locally.


befitsandpiper (approved):
blessed

Issue #2462, opened 2020-11-06
Improve Identity Activation Workflow


topocount:
Currently, activating and deactivating a user involves clicking on the user in the Identities view and then toggling the activation checkbox. Most regular administration of existing users seems to revolve around managing the activation of muiltiple users on a regular basis, whether they're onboarding or losing good standing within the group.

This workflow can be greatly improved by moving the activation checkbox from the Identity details to the list of Identities, so every identity entry has a checkbox next to it, that reflects their current activation status.

As is currently the case, toggling the checkbox should immediately write an entry to the ledger via the `activate` Ledger api:
https://github.com/sourcecred/sourcecred/blob/7cee24d8d5fefa80c0d3b885d3741d301ae4fe45/src/core/ledger/ledger.js#L385-L391

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038626529529) by [Unito](https://www.unito.io/learn-more)

PR #2508, opened 2020-11-12
LedgerAdmin: improve activation workflow

crisog:
This closes PR #2462.

paired with @blueridger 

The GIF looks kinda slow but shows the functionality.

ActivationWorkflow

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038547504865) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/ui/components/IdentityListItems.js
+++ b/src/ui/components/IdentityListItems.js
@@ -1,24 +1,34 @@
 // @flow
 import React, {type Node as ReactNode} from "react";
-import {Divider, ListItem} from "@material-ui/core";
-import {type Identity} from "../../core/identity";
+import {Divider, ListItem, ListItemText, Checkbox} from "@material-ui/core";
+import {type Identity, type IdentityId} from "../../core/identity";
+import {type Account} from "../../core/ledger/ledger";
 
 type IdentityListItemsProps = {
-  +identities: $ReadOnlyArray<Identity>,
-  +onClick: Function,
+  +accounts: $ReadOnlyArray<Account>,
+  +onClick: (Identity) => void,
+  +onCheckbox: (IdentityId) => void,
 };
 
 export const IdentityListItems = ({
-  identities,
+  accounts,
   onClick,
+  onCheckbox,
 }: IdentityListItemsProps): ReactNode => {
-  const lastIndex = identities.length - 1;
+  const lastIndex = accounts.length - 1;
 
   if (lastIndex > -1) {
-    return identities.map((identity, index) => (
-      <React.Fragment key={identity.id}>
-        <ListItem button onClick={() => onClick(identity)}>
-          {identity.name}
+    return accounts.map((account, index) => (
+      <React.Fragment key={account.identity.id}>
+        <ListItem button onClick={() => onClick(account.identity)}>
+          <ListItemText primary={account.identity.name} />
+          <Checkbox
+            onClick={(e) => e.stopPropagation()}
+            onChange={() => onCheckbox(account.identity.id)}
+            checked={account.active}
+            name="active"
+            color="primary"
+          />
         </ListItem>
         {index < lastIndex && <Divider />}
       </React.Fragment>
--- a/src/ui/components/LedgerAdmin.js
+++ b/src/ui/components/LedgerAdmin.js
@@ -10,6 +10,8 @@ import {
   FormControlLabel,
   List,
   TextField,
+  ListSubheader,
+  Divider,
 } from "@material-ui/core";
 import {useLedger} from "../utils/LedgerContext";
 import {useTableState} from "../../webutil/tableState";
@@ -61,6 +63,11 @@ const useStyles = makeStyles((theme) => {
     addEditPrompt: {
       margin: 0,
     },
+    listSubheader: {
+      display: "grid",
+      gridTemplateColumns: "auto auto",
+      justifyContent: "space-between",
+    },
   };
 });
 
@@ -71,10 +78,7 @@ export const LedgerAdmin = (): ReactNode => {
   const [nextIdentityName, setIdentityName] = useState<string>("");
   const [selectedId, setSelectedId] = useState<IdentityId | null>(null);
   const [promptString, setPromptString] = useState<string>("Add Identity:");
-  const [checkboxSelected, setCheckBoxSelected] = useState<boolean>(false);
-  const accounts = useMemo(() => ledger.accounts().map((a) => a.identity), [
-    ledger._latestTimestamp,
-  ]);
+  const accounts = useMemo(() => Array.from(ledger.accounts()), [ledger]);
   const accountsTableState = useTableState({data: accounts});
 
   const changeIdentityName = (event: SyntheticInputEvent<HTMLInputElement>) =>
@@ -94,10 +98,8 @@ export const LedgerAdmin = (): ReactNode => {
     let nextLedger;
     if (ledger.account(id).active) {
       nextLedger = ledger.deactivate(id);
-      setCheckBoxSelected(false);
     } else {
       nextLedger = ledger.activate(id);
-      setCheckBoxSelected(true);
     }
     updateLedger(nextLedger);
   };
@@ -105,14 +107,12 @@ export const LedgerAdmin = (): ReactNode => {
   const resetIdentity = () => {
     setIdentityName("");
     setSelectedId(null);
-    setCheckBoxSelected(false);
     setPromptString("Add Identity: ");
   };
 
   const setActiveIdentity = (identity: Identity) => {
     setIdentityName(identity.name);
     setSelectedId(identity.id);
-    setCheckBoxSelected(ledger.account(identity.id).active);
     setPromptString("Update Identity: ");
   };
 
@@ -125,8 +125,8 @@ export const LedgerAdmin = (): ReactNode => {
       .join("+.*");
     const regex = new RegExp(filterString);
 
-    accountsTableState.createOrUpdateFilterFn("filterIdentities", (identity) =>
-      regex.test(identity.name.toLowerCase())
+    accountsTableState.createOrUpdateFilterFn("filterIdentities", (account) =>
+      regex.test(account.identity.name.toLowerCase())
     );
   };
 
@@ -152,8 +152,8 @@ export const LedgerAdmin = (): ReactNode => {
             className={classes.checkboxElement}
             control={
               <Checkbox
-                checked={checkboxSelected}
                 onChange={() => toggleIdentityActivation(selectedId)}
+                checked={ledger.account(selectedId)?.active}
                 name="active"
                 color="primary"
               />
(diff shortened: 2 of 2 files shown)


src/ui/components/LedgerAdmin.js
```diff
@@ -74,38 +87,60 @@ export const LedgerAdmin = (): ReactNode => {
...
+
+    ledger
+      .accounts()
+      .map((a) => a.identity)
+      .map((identity) => {
+        identities = {
+          ...identities,
+          [`${identity.id}`]: ledger.account(identity.id).active,
+        };
+      });
```
BrianBelhumeur, line 119: Hey @crisog, I'd like to suggest a simplified version of this code as follows:
```suggestion
    const identities = {};

    ledger.accounts().map(({identity, active}) => {
      identities[`${identity.id}`] = active;
    });
```

It eliminates the second `map()`, prevents re-spreading the `identities` object each time, and eliminates the need to access `ledger.account(id)` to get each active status. I've pulled your branch and tested this locally, and it seems to work fine though I'm new to the project and I'd suggest testing for yourself in case I missed something.

src/ui/components/LedgerAdmin.js
```diff
@@ -74,38 +87,60 @@ export const LedgerAdmin = (): ReactNode => {
...
+  useEffect(() => loadEntitiesCheckList(), [ledger]);
+
   const renderIdentities = () => {
     const renderIdentity = (i: Identity, notLastElement: boolean) => (
       <>
         <ListItem button onClick={() => setActiveIdentity(i)} key={i.id}>
           {i.name}
+          <Checkbox
+            onChange={handleSingleCheck}
+            checked={NullUtil.orElse(isChecked[i.id], false)}
```
BrianBelhumeur, line 133: There may be other reasons that `NullUtil.orElse()` is used commonly in the codebase (I'm still learning a lot about how things are done here) but you should also be able to use:
```suggestion
            checked={Boolean(isChecked[i.id])}
```
Same goes for line 179 below. If so, you can scratch the import on line 21.

crisog:
Just realized this is still pending, if any of you has a chance to review this let me know @decentralion @topocount @BrianBelhumeur @befitsandpiper.

src/ui/components/LedgerAdmin.js
```diff
@@ -137,12 +166,12 @@ export const LedgerAdmin = (): ReactNode => {
         />
         {selectedId && (
           <FormControlLabel
-            fullWidth
+            fullwidth="true"
```
befitsandpiper, line 169: not sure why you added the string "true" on line 159 and 169. This technically works since the "true" string is not falsey, but the previous way of just setting the key as a Component property is more correct, since it assigns a boolean. Otherwise this branch looks good.

befitsandpiper: @crisog

crisog: Yea I changed the property to `fullwidth` because of a React warning, but after checking MaterialUI it seems that the correct way to do it was just `fullWidth`. About the string, yea that should not be there in any of the cases.

src/ui/components/LedgerAdmin.js
```diff
@@ -74,38 +87,54 @@ export const LedgerAdmin = (): ReactNode => {
...
+    const identities = {};
+
+    ledger.accounts().map(({identity, active}) => {
+      identities[`${identity.id}`] = active;
+    });
+
+    setIsChecked(identities);
+  };
+
+  useEffect(() => loadEntitiesCheckList(), [ledger]);
```
hammadj, line 118: No need to define the function outside of useEffect

```suggestion
  useEffect(() => {
    const identities = {};

    ledger.accounts().map(({identity, active}) => {
      identities[`${identity.id}`] = active;
    });

    setIsChecked(identities);
  }, [ledger]);
```

src/ui/components/LedgerAdmin.js
```diff
@@ -127,7 +156,7 @@ export const LedgerAdmin = (): ReactNode => {
       <h3>{promptString}</h3>
       <div className={classes.centerRow}>
         <TextField
-          fullWidth
+          fullwidth="true"
```
hammadj, line 159: Doesn't need the `="true"`

src/ui/components/LedgerAdmin.js
```diff
@@ -55,7 +58,17 @@ export const LedgerAdmin = (): ReactNode => {
...
+
+  const handleSingleCheck = (event: SyntheticInputEvent<HTMLInputElement>) => {
+    const id = event.currentTarget.id;
+    const checked = event.currentTarget.checked;
+
+    const uuid = uuidFromString(id);
+    toggleIdentityActivation(uuid);
+
+    setIsChecked({...isChecked, [id]: checked});
+  };
```
hammadj, line 71: should wrap in `useCallback`

hammadj (changes requested):

src/ui/components/LedgerAdmin.js
```diff
@@ -137,12 +166,12 @@ export const LedgerAdmin = (): ReactNode => {
...
           <FormControlLabel
-            fullWidth
+            fullwidth="true"
             className={classes.checkboxElement}
             control={
               <Checkbox
-                checked={checkboxSelected}
-                onChange={() => toggleIdentityActivation(selectedId)}
+                checked={NullUtil.orElse(isChecked[selectedId], false)}
+                disabled={true}
```
hammadj, line 174: this would disable the checkbox permanently, seems like a typo or left over debugging stuff?

src/ui/components/LedgerAdmin.js
```diff
@@ -178,7 +207,7 @@ export const LedgerAdmin = (): ReactNode => {
         </>
       )}
       <div className={classes.centerRow}>
-        <List fullWidth className={classes.identityList}>
+        <List fullwidth="true" className={classes.identityList}>
```
hammadj, line 210: this change should be reverted as well

crisog:
I ran `yarn sharness-full` on my local, and it seems to be passing, though it doesn't in the CI server.

KuraFire:
For the table (list) of identities, I made a wireframe of the UI structure for improving activation workflow:

Wireframes - Admin workflow - IDENTITIES

Please implement in a similar way, meaning: the activation toggle checkboxes are right-aligned (and all aligned vertically across), and there is a column header labeled `Active` for them, and one above the participant names labeled `Participant`.

**IMPORTANT NOTE:**
The including of aliases in the different way represented in the wireframe should be considered a second-stage implementation in a new ticket, separate from this one.

Once implemented, please provide an image or GIF of the UI for design review. Thanks! :)

crisog:
> Please implement in a similar way, meaning: the activation toggle checkboxes are right-aligned (and all aligned vertically across), and there is a column header labeled `Active` for them, and one above the participant names labeled `Participant`.
> 

This is how it looks like now, the behavior is the same as the initial gif.

[Image from Gyazo](https://gyazo.com/c79c134b42f60a0abe6215cc3d175e3c)

KuraFire (approved):
Looking great, @crisog !

src/ui/components/LedgerAdmin.js
```diff
@@ -74,38 +101,52 @@ export const LedgerAdmin = (): ReactNode => {
...
 
+  useEffect(() => {
+    const identities = {};
+
+    ledger.accounts().map(({identity, active}) => {
+      identities[`${identity.id}`] = active;
+    });
+
+    setIsChecked(identities);
+  }, [ledger]);
```
BrianBelhumeur, line 130: I recommend converting this to a `useMemo` so it's synchronous, as well as passing `ledger.accounts()` as the arg list.

BrianBelhumeur (approved):

blueridger:
Rebased, fixed some nasty merge conflicts, and made some minor tweaks. @hammadj could you give this another pass?

topocount (approved):

src/ui/components/IdentityListItems.js
```diff
@@ -1,24 +1,34 @@
...
-import {Divider, ListItem} from "@material-ui/core";
-import {type Identity} from "../../core/identity";
+import {Divider, ListItem, ListItemText, Checkbox} from "@material-ui/core";
+import {type Identity, type IdentityId} from "../../core/identity";
 
 type IdentityListItemsProps = {
   +identities: $ReadOnlyArray<Identity>,
   +onClick: Function,
+  +isCheckedHash: Map<IdentityId, boolean>,
+  +onCheckbox: Function,
```
hammadj, line 10: The type signature of just `Function` means that anyone can pass it any params and it won't complain, which isn't great for type safety. Its usage below indicates it should be a function that accepts an identityId as its first param and a change event as its second param.

src/ui/components/LedgerAdmin.js
```diff
@@ -71,12 +85,28 @@ export const LedgerAdmin = (): ReactNode => {
...
+
+  const handleSingleCheck = useCallback(
+    (id: IdentityId, event: SyntheticInputEvent<HTMLInputElement>) => {
+      const checked = event.currentTarget.checked;
+      toggleIdentityActivation(id);
+
+      isCheckedHash.set(id, checked);
+    }
+  );
```
hammadj, line 108: This setup is kinda funky and bug-prone since there's two sources of truth for whether an account is active or not (the ledger itself + the `isCheckedHash` map).

Additionally, the way that isCheckedHash is using `useMemo` will cause the same logic to run many times unnecessarily. For example:
1. User loads page and `isCheckedHash` map is created and memoized based on the current ledger
2. User clicks a check box to activate an identity which sets the corresponding entry in `isCheckedHash` and also updates the ledger when it calls `toggleIdentityActivation`
3. Since the ledger has changed, the useMemo call for `isCheckedHash` runs again which iterates over all the ledger accounts and sets the same values that are already set, which will cause UI lag and choppiness as well as potential desynchronization of what's in the ledger and what the UI is showing.

The solution to this is just to not have an `isCheckedHash` value at all and instead use the ledger itself as the single source of truth for whether an account is active or not. Instead of doing `isCheckedHash.get(identityId)` for determining if a row is checked or not, you can just do `ledger.account(identityId)?.active` and in `handleSingleCheck` you just call the `toggleIdentityActivation` function and get rid of the `isCheckedHash.set(id, checked)` call. Keeps it much more simple and declarative instead of trying to copy state back and forth between ledger and `isCheckedHash`

hammadj (commented):
much cleaner, but still think we should improve the state management logic to prevent bugs / rendering bottlenecks that would hurt the UX.

blueridger:
Made the requested changes. Also took the liberty of changing the UX so that when checking inline checkboxes, the toggled identity is not selected as active for the widget at the top of the page. This feels important because before, when checking inline checkboxes, the table would jump around under your mouse as the top widget changed sizes due to the active identity changing.

hammadj (approved):

PR #2513, opened 2020-11-14

elihanover:
This commit allows users to specify an array of policies, which allows them two new advantages:
__1. Do not need to specify each policy.__
__2. Can add multiple version of the same policy (in the case that you want to use different parameters).__

Currently, users specify a budget for every valid policy in their instance's `config/grain.json` file, even if they are not using that grain allocation policy.
```js
{
  "immediatePerWeek": 0,
  "balancedPerWeek": 10,
  "recentPerWeek": 0,
}
```

This also forces us to handle each policy explicitly downstream, which could be a lot cleaner if we could leverage the `AllocationPolicy` type we already have.  This approach uses typing to check whether the user has specified a valid policy implicitly.

```js
{
  policies: [
    {
      policyType: "BALANCED",
      budget: "100"
    },
    {
      policyType: "RECENT",
      budget: "50",
      discount: 0.1
    },
    {
      policyType: "RECENT",
      budget: "100",
      discount: 0.5
    }
}
```

__Test Plan__
Unit tests are provided here for both the `parser`, and well as `toDistributionPolicy`, which serves to check the config.  Note, we can reduce this function entirely with stricter parsers and types.  For example, we have to check for negative grain values.  This calls for a new "`BudgetedGrain`" type, which I will follow up with.

### DIFF
--- a/src/api/grainConfig.js
+++ b/src/api/grainConfig.js
@@ -4,75 +4,42 @@ import {type DistributionPolicy} from "../core/ledger/applyDistributions";
 import * as G from "../core/ledger/grain";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
-import {toDiscount} from "../core/ledger/grainAllocation";
+import {
+  type AllocationPolicy,
+  allocationPolicyParser,
+} from "../core/ledger/grainAllocation";
 
 export type GrainConfig = {|
-  +immediatePerWeek: number,
-  +balancedPerWeek: number,
-  +recentPerWeek: number,
-  +recentWeeklyDecayRate?: number,
+  +allocationPolicies: $ReadOnlyArray<AllocationPolicy>,
   +maxSimultaneousDistributions?: number,
 |};
 
 export const parser: C.Parser<GrainConfig> = C.object(
   {
-    immediatePerWeek: C.number,
-    balancedPerWeek: C.number,
-    recentPerWeek: C.number,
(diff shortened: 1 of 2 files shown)


topocount:
@elihanover looks like you might have a merge conflict to tackle here, just a heads up

topocount (commented):
@elihanover After thinking about this further, I think it makes sense to allow these configs to not exist (for back-compat reasons) as they did before this change, and allow them to parse to 0 if absent from the file. An array of policies is much more difficult for users to manipulate by hand, and unless we write a serializer that will auto-update existing grain config files, I don't believe this is sustainable since this change doesn't need to be breaking.

decentralion:
Agree with @topocount, I'd rather not make a breaking change here unless there's a strong motivation for doing so.

topocount:
@elihanover you might like to know that an `optional` param exists in the Object parser we have, for fields that are parsed if they exist, but won't cause the parser to throw if missing: https://github.com/sourcecred/sourcecred/blob/55c803ee60e1668d3d3262f0df096d6c30092c29/src/util/combo.js#L314-L321

I used this pattern when setting up something similar:
https://github.com/sourcecred/sourcecred/blob/55c803ee60e1668d3d3262f0df096d6c30092c29/src/api/currencyConfig.js#L25-L41

where the config is optionally parsed and then processed by NullUtil if absent to generate the shape I want to work with. There are other patterns as well if you want to utilize the `pure` parser for example, but I like this because It's kind of verbose and obvious how the values are processed.

PR #2520, opened 2020-11-17
Add ArrayView util for sorting, filtering, and paginating Arrays

blueridger:
# Notes
This is a utility Array wrapper that I am creating for the Explorer Home react view, but I expect this will continue to be useful for any page that has a filtered+paginated+sorted table. The purpose is to wrap the state management needed to apply those 3 transformations, in order to simplify both the UI code and the data construction code.

The pagination and filtering here is solely for UI presentation. It does not offer memory savings since the entire array is still stored in memory (and additional state data is stored on top).

# Automated Tests
`yarn test`
added thorough unit tests

# Manual Testing
None

### DIFF
--- a/src/util/sortBy.js
+++ b/src/util/sortBy.js
@@ -1,6 +1,6 @@
 //@flow
 
-type PluckFn<T> = (T) => any;
+export type PluckFn<T> = (T) => any;
 
 const identity = (x) => x;
 
--- a/src/webutil/arrayState.js
+++ b/src/webutil/arrayState.js
@@ -1,0 +1,104 @@
+// @flow
+
+import {type PluckFn} from "../util/sortBy";
+import sortBy from "../util/sortBy";
+import cloneDeep from "lodash.clonedeep";
+
+export const SortOrders = {
+  ASCENDING: "ASCENDING",
+  DESCENDING: "DESCENDING",
+};
+type SortOrderTypes = $Keys<typeof SortOrders>;
+
+/**
+  This is a utility for managing the state of sorted, filtered, and/or paginated
+  table data on the front end.
+ */
+export class ArrayState<T> {
+  _array: Array<T>;
+  _filteredArray: Array<T>;
+  _rowsPerPage: number;
+  _sortArgs: ?[PluckFn<T>, SortOrderTypes];
+  _filterFn: ?(T) => boolean;
+
+  _pages(): number {
+    return Math.ceil(
+      this._filteredArray.length /
+        Math.min(this._rowsPerPage, this._filteredArray.length)
+    );
+  }
+
+  constructor(data: $ReadOnlyArray<T>) {
+    this.setData(data);
+    this._rowsPerPage = +Infinity;
+  }
+
+  /**
+    data: a new baseline data set. Existing transformations will be applied.
+    return: the new number of pages
+   */
+  setData(data: $ReadOnlyArray<T>): number {
+    this._array = cloneDeep(data);
+    return this._filterFn ? this.filter(this._filterFn) : this.filter();
+  }
+
+  /**
+    pluckArg: a function of an object that returns the element by which to sort
+    order: "ASCENDING" or "DESCENDING"
+   */
+  sortBy(pluckArg: PluckFn<T>, order: SortOrderTypes) {
+    this._filteredArray = sortBy(this._filteredArray, pluckArg);
+    if (order === SortOrders.DESCENDING) {
+      this._filteredArray.reverse();
+    }
+    this._sortArgs = [pluckArg, order];
+  }
+
+  /**
+    Removes any existing filters and applies the new filter if provided.
+
+    filterFn: a function that returns true for objects that should be included, omit to remove filter
+    return: the new number of pages
+   */
+  filter(filterFn?: (T) => boolean): number {
+    this._filterFn = filterFn;
+    this._filteredArray = filterFn ? this._array.filter(filterFn) : this._array;
+    if (this._sortArgs) this.sortBy(...this._sortArgs);
+    return this._pages();
+  }
+
+  /**
+    rowsPerPage: how many rows should be displayed per page, +Infinity to display all rows on one page
+    return: the new number of pages
+   */
+  setRowsPerPage(rowsPerPage: number): number {
+    if (rowsPerPage !== parseInt(rowsPerPage) && rowsPerPage !== +Infinity)
+      throw "rowsPerPage must be an integer";
+    if (rowsPerPage < 1) throw "rowsPerPage must be at least 1";
+    this._rowsPerPage = rowsPerPage;
+    return this._pages();
+  }
+
+  /**
+    Get the total number of pages.
+   */
+  getPageCount(): number {
+    return this._pages();
+  }
+
+  /**
+    Get the contents of the specified page.
+   */
+  getPage(pageNumber: number): $ReadOnlyArray<T> {
+    if (pageNumber !== parseInt(pageNumber))
+      throw "pageNumber must be an integer";
+    if (pageNumber < 1 || pageNumber > this._pages())
+      throw "Invalid page number. Use an integer in the range [1, pageCount].";
+    const startIndex =
+      Math.min(this._rowsPerPage, this._filteredArray.length) *
+      (pageNumber - 1);
+    const endIndex =
+      Math.min(this._rowsPerPage, this._filteredArray.length) + startIndex;
+    return this._filteredArray.slice(startIndex, endIndex);
+  }
+}
--- a/src/webutil/arrayState.test.js
+++ b/src/webutil/arrayState.test.js
@@ -1,0 +1,231 @@
+// @flow
+
+import {ArrayState, SortOrders} from "./arrayState";
+
+describe("src/webutil/arrayState", () => {
+  describe("when the type is an object", () => {
+    let arrayState;
+    beforeEach(() => {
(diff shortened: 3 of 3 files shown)


src/util/arrayView.js
```diff
@@ -0,0 +1,74 @@
...
+  /**
+    pluckArg: a function of an object that returns the element by which to sort
+    order: "ascending" or "descending"
+   */
+  sortBy(pluckArg: PluckFn<T>, order: "ascending" | "descending") {
+    this._array = sortBy(this._array, pluckArg);
+    this._filteredArray = sortBy(this._filteredArray, pluckArg);
+    if (order === "descending") {
+      this._array.reverse();
+      this._filteredArray.reverse();
```
decentralion, line 28: Hmm, sort is `O(n * log n)`, and this function is likely to be in the hot path for re-rendering the frontend (i.e. user changes the sort order in the UI). This function does two sorts, which is unnecessary. In particular, we never need to sort the full array since we don't actually return directly from it, we just need to sort the filtered array. It's easy to imagine cases where the filtered array is dramatically smaller than the full array (e.g. filtered down to just contributions in October 2020, or filtered down to just GitHub pull requests), so this could be a real change in performance.

How about we cache the sortBy arguments, so that way when the filter changes, we can compute a new filteredArray and then re sort it?

decentralion (commented):
I'd like to see an example of migrating existing code to use this interface before using it. I wonder about how cleanly it will integrate... as an example, the rowsPerPage is stored inside this class, but is not accessible. However, the UI using this class will want to know the rowsPerPage setting so that it can render it and update appropriately. In that case, will the containing UI also have its own representation of the rowsPerPage state? Are there potential bugs where these two pieces of state desynchronize? Could we avoid those issues by having this class expose rowsPerPage? etc

src/util/sortBy.js
```diff
@@ -1,6 +1,6 @@
 //@flow
 
-type PluckFn<T> = (T) => any;
+export type PluckFn<T> = (T) => any;
```
decentralion, line 3: Probably shouldn't be a part of this PR, but I wonder if we should change the type signature to `(T) => number | string`. I can't really think of a case where trying to compare on something other than numbers or strings is a good idea.

src/util/arrayView.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+
+import {type PluckFn} from "./sortBy";
+import sortBy from "./sortBy";
+
+export class ArrayView<T> {
+  _array: Array<T>;
+  _filteredArray: Array<T>;
+  _rowsPerPage: number;
+  _pages: number;
```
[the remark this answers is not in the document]
blueridger: sure

src/util/arrayView.js
blueridger: Good idea, I'll add a cache.

src/util/arrayView.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+    rowsPerPage: how many rows should be displayed per page, +Infinity to display all rows on one page
+    return: the new number of pages
+   */
+  setRowsPerPage(rowsPerPage: number): number {
+    rowsPerPage = Math.min(rowsPerPage, this._filteredArray.length);
+    rowsPerPage = Math.max(rowsPerPage, 1);
+    rowsPerPage = parseInt(rowsPerPage.toFixed(0));
```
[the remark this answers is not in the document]
blueridger: This line is to remove decimals since JS doesn't have an int type, because lol I forgot about Math.round
decimals would cause not good things to happen later on if we stored them, so this is just for robustness.

src/util/arrayView.js
```diff
@@ -0,0 +1,74 @@
...
+import sortBy from "./sortBy";
+
+export class ArrayView<T> {
+  _array: Array<T>;
+  _filteredArray: Array<T>;
+  _rowsPerPage: number;
+  _pages: number;
+
+  constructor(data: $ReadOnlyArray<T>) {
+    this._array = JSON.parse(JSON.stringify(data)); //Deep Copy
```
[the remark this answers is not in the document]
blueridger: accepted

src/util/arrayView.js
```diff
@@ -0,0 +1,74 @@
...
+  getPageCount(): number {
+    return this._pages;
+  }
+
+  /**
+    Get the contents of the specified page.
+   */
+  getPage(pageNumber: number): $ReadOnlyArray<T> {
+    pageNumber = parseInt(pageNumber.toFixed(0));
+    if (pageNumber < 1 || pageNumber > this._pages) return [];
```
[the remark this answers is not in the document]
blueridger: sure, will change to an error.

src/util/sortBy.js
blueridger: I was wondering that too

src/util/arrayView.js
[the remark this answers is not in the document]
decentralion: `parseInt(1.2)` already returns `1` tho

src/util/arrayView.js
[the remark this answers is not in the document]
decentralion: Are you keeping the 1-based indexing? if so, why?

src/util/arrayView.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+
+import {type PluckFn} from "./sortBy";
+import sortBy from "./sortBy";
+
+export class ArrayView<T> {
```
decentralion, line 6: Another thing: Let's have a module docstring explaining what the intention is for this module. The context that it's specifically intended for displaying data in frontends would be helpful.

Since it's intended for frontend use only, we should probably put it in `src/webutil` instead of `src/util`

blueridger:
> I'd like to see an example of migrating existing code to use this interface before using it. I wonder about how cleanly it will integrate... as an example, the rowsPerPage is stored inside this class, but is not accessible. However, the UI using this class will want to know the rowsPerPage setting so that it can render it and update appropriately. In that case, will the containing UI also have its own representation of the rowsPerPage state? Are there potential bugs where these two pieces of state desynchronize? Could we avoid those issues by having this class expose rowsPerPage? etc

So d3 won't need to track rows per page for the table building, though it will track it separately for the selection mechanism. The data flow on this attribute is generally unidirectional from the front end to the mdule. The only reason the front end might need it exposed by this module is for the initial state, which practically speaking is just "All pages" rather that the actual number, so I'm skeptical how useful exposing it will be. The cool thing about this module, is that it is easily tweakable and upgradable if we change our minds during implementation.

I will look around for somewhere we could integrate it into existing code.

src/util/arrayView.js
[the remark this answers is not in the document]
blueridger: today I learned

src/util/arrayView.js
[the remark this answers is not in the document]
blueridger: Because the page number is ultimately a UI concept, which will be 1 indexed on the page selector front end element, it seems intuitive to me for `page` to be 1-indexed whereas `index` would be 0-indexed.

src/util/arrayView.js
blueridger: sure, will move it and add a doc comment.

blueridger:
Addressed feedback, and added a setData method to be able to change the data set, in the case that external filtering happens like in the ExplorerHomeDataBuilder.

decentralion:
> So d3 won't need to track rows per page for the table building, though it will track it separately for the selection mechanism.

I'm thinking about React moreso than d3.

An example of an existing UI is a table with a rowsPerPage selection is the ledger history viewer: https://cred.sourcecred.io/#/ledger

An example of a UI that would really benefit from being made into a table with sorting, searching, and pagination is the Grain Account viewer: https://cred.sourcecred.io/#/accounts

Either could be a good test case for demonstrating the suitability of this util.

src/webutil/arrayState.js
```diff
@@ -0,0 +1,104 @@
...
+  /**
+    Removes any existing filters and applies the new filter if provided.
+
+    filterFn: a function that returns true for objects that should be included, omit to remove filter
+    return: the new number of pages
+   */
+  filter(filterFn?: (T) => boolean): number {
+    this._filterFn = filterFn;
+    this._filteredArray = filterFn ? this._array.filter(filterFn) : this._array;
+    if (this._sortArgs) this.sortBy(...this._sortArgs);
```
BrianBelhumeur, line 66: Have you run prettier on this file?

src/webutil/arrayState.test.js
```diff
@@ -0,0 +1,231 @@
...
+      });
+    });
+
+    describe(".setRowsPerPage", () => {
+      it("should work with one record per page", () => {
+        expect(arrayState.setRowsPerPage(1)).toEqual(5);
+        expect(arrayState.getPage(2)).toEqual([{num: 1, str: "1"}]);
+      });
+
+      it("should create no less that 1 page", () => {
```
BrianBelhumeur, line 47:
```suggestion
      it("should create no less than 1 page", () => {
```

BrianBelhumeur (approved):

src/webutil/arrayState.js
decentralion: CI wouldn't be passing if the code weren't prettified

blueridger:
So I dived into React, and I'm seeing how this doesn't integrate super well. Going to do some more learning on React, and maybe come back to something like this with that knowledge.

Issue #2244, opened 2020-09-11
Measure "active participants"


decentralion:
For the new Explorer dashboard (https://github.com/sourcecred/sourcecred/issues/2227), we are planning to introduce an "Active participants" metric. However, we don't currently have this data.

What we have is a Cred Graph with:
- nodes, usually representing contributions, with timestamps
- edges, which also have timestamps
- cred scores on nodes, and cred flows on edges

@KuraFire has proposed a metric where someone is active in a time period if they have made a contribution. Currently, we don't have the ability to surface this, because we don't know which edges represent authorship.

Therefore, for now, I propose we define someone as active within an interval iff they are incident to any edge which was created in that interval.

Once we have added the graph tag system (#2243), we can add a core `[sourcecred, core, authorship]` tag which represents authorship in a cross-plugin way. Then, we can actually compute the active participants based on authorship.

KuraFire:
Just to be clear, I proposed Active Participant as “someone who is active in a time period where they have made a contribution _that earned Cred_.”

I know that does not necessarily solve the authorship aspect, but I do believe that matches the edge… case… umm. The edges existing on the node graph. (yes?)

decentralion:
Thanks for the clarification, @KuraFire. In our dev meeting today, @hammadj and i decided that you should be considered active in an interval if your Cred in that interval was over some configurable threshold. It's more sibyll resistant and flexible.

KuraFire:
That's a great solution for this, nice and flexible. Product Design seal of approval ;) 

> It's more sibyll resistant …

Question about this: wouldn't having multiple identities and making each of them make a different contribution ultimately be self-defeating regardless of this threshold, since each identity's internal reputation stays lower than if you had just the one?

decentralion:
> > It's more sibyll resistant …
> 
> Question about this: wouldn't having multiple identities and making each of them make a different contribution ultimately be self-defeating regardless of this threshold, since each identity's internal reputation stays lower than if you had just the one?

It depends on the specific incentives being gamed.

I think that using Cred is trivially more Sybil resistent than just counting "number of people who made at least one contribution", because obviously it's trivial to spam new accounts that each make one post or such, and game that metric.

So let's consider the much higher-effort attack of making multiple accounts, and having each account make real contributions. This attack would probably succeed, in that the attacker would successfully control multiple "active" accounts with nontrivial Cred. Would it be worth the effort? That depends on the attacker's goals and incentives.

[...]

PR #2521, opened 2020-11-17
ExplorerHomeView Interface

blueridger:
# Notes
I decided to work backwards from the immediate problem, which is to populate the ExplorerHome page with the data it needs. This is the skeleton of a class whose sole purpose is to do that. It will consume smarter/more generic APIs (which I plan being a slightly improved CredGraph API and a new GrainData API), which will be prerequisite to the implementation of this skeleton. It will leave transformations other than filtering by interval to downstream logic (https://github.com/sourcecred/sourcecred/pull/2520).

# Testing
`yarn test`
not tested, just a skeleton for early review.

### DIFF
--- a/src/analysis/explorerHomeView.js
+++ b/src/analysis/explorerHomeView.js
@@ -1,0 +1,82 @@
+// @flow
+import {CredGraph} from "../core/credrank/credGraph";
+import {type TimestampMs} from "../util/timestamp";
+import {type Interval} from "../core/interval";
+import * as Grain from "../core/ledger/grain";
+
+export type DataOverTime = $ReadOnlyArray<number>;
+
+export type UserRow = {|
+  +name: string,
+  +credEarnedInTimeRange: number,
+  +grainHarvestedInTimeRange: Grain.Grain,
+  +grainBalance: Grain.Grain,
+  +allTimeCred: DataOverTime,
+  +category: UserCategory,
+|};
+
+export type UserCategory = "participant" | "bot" | "project" | "organization";
+
+export type ExplorerHomeView = {|
+  +credEarned: number,
+  +grainHarvested: Grain.Grain,
+  +participantCount: number, //Question: what is this supposed to capture? Is it supposed to change depending on the time range selected?
+  +grainPerCred: Grain.Grain,
+  +userRows: $ReadOnlyArray<UserRow>,
+  +intervalsIncluded: $ReadOnlyArray<Interval>,
+|};
+
+/**
+  Responsible for transforming data from the CredGraph and LedgerLog(?) into the information needed by the Explorer Home web page.
+ */
+export class ExplorerHomeViewBuilder {
+  constructor(credGraph: CredGraph) {
+    credGraph;
+    //Other params, depending on implementation. Probably a new GrainData API.
+    //Saves params or transforms them, depending on implementation strategy
+  }
+
+  /**
+    Returns all available intervals.
+    Useful for determining appropriate time ranges to offer for selection.
+   */
+  getAllIntervals(): $ReadOnlyArray<Interval> {
+    return [];
+  }
+
+  /**
+    Returns the timelines aggregated over all available intervals.
+   */
+  getAggregateTimelines(): {|
+    +credOverTime: DataOverTime,
+    +grainOverTime: DataOverTime,
+    +totalParticipantsOverTime: DataOverTime,
+  |} {
+    return {
+      credOverTime: [],
+      grainOverTime: [],
+      totalParticipantsOverTime: [],
+    };
+  }
+
+  /**
+    1. Narrows time range to only include intervals that start and end in the range [startTime-inclusive, endTime-inclusive]
+    2. Grain distributions are included if their timestamp is in the narrowed range (firstIntervalStartTime-exclusive, lastIntervalEndTime-inclusive]
+   */
+  getExplorerHomeView(
+    startTime?: TimestampMs,
+    endTime?: TimestampMs
+  ): ExplorerHomeView {
+    //to pass build
+    startTime;
+    endTime;
+    return {
+      credEarned: 0,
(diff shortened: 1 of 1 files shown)


src/analysis/explorerHomeView.js
```diff
@@ -0,0 +1,82 @@
+// @flow
+import {CredGraph} from "../core/credrank/credGraph";
+import {type TimestampMs} from "../util/timestamp";
+import {type Interval} from "../core/interval";
+import * as Grain from "../core/ledger/grain";
```
decentralion, line 5: nit / no change required: If I had written this, I might do something like:

```js
import * as Grain from "../core/ledger/grain";
import type {Grain as GrainT} from "../core/ledger/grain";
```

src/analysis/explorerHomeView.js
```diff
@@ -0,0 +1,82 @@
...
+  +allTimeCred: DataOverTime,
+  +category: UserCategory,
+|};
+
+export type UserCategory = "participant" | "bot" | "project" | "organization";
+
+export type ExplorerHomeView = {|
+  +credEarned: number,
+  +grainHarvested: Grain.Grain,
+  +participantCount: number, //Question: what is this supposed to capture? Is it supposed to change depending on the time range selected?
```
decentralion, line 23: We have this idea of defining "active participants" per https://github.com/sourcecred/sourcecred/issues/2244

It will change over time because whether you are considered "active" depends on your level of activity in the chosen period

src/analysis/explorerHomeView.js
```diff
@@ -0,0 +1,82 @@
...
+export type UserRow = {|
+  +name: string,
+  +credEarnedInTimeRange: number,
+  +grainHarvestedInTimeRange: Grain.Grain,
+  +grainBalance: Grain.Grain,
+  +allTimeCred: DataOverTime,
+  +category: UserCategory,
+|};
+
+export type UserCategory = "participant" | "bot" | "project" | "organization";
```
decentralion, line 18: Use the `IdentityType` type:

`import type {IdentityType} from "../core/identity"`

src/analysis/explorerHomeView.js
```diff
@@ -0,0 +1,82 @@
...
+
+export type ExplorerHomeView = {|
```
decentralion, line 20: Kind of a nit, but can we rename this to something like `ExplorerHomeData`? I think naming it `...View` invites some potential confusion with the Model, View, Controller design pattern (where a View is responsible for rendering in a UI). So when I saw the name `View` I was expecting it to be a class of some kind, not a data type.

src/analysis/explorerHomeView.js
```diff
@@ -0,0 +1,82 @@
...
+    //Saves params or transforms them, depending on implementation strategy
+  }
+
+  /**
+    Returns all available intervals.
+    Useful for determining appropriate time ranges to offer for selection.
+   */
+  getAllIntervals(): $ReadOnlyArray<Interval> {
+    return [];
+  }
```
decentralion, line 45: this would actually return an `IntervalSequence` type. it's basically a ROA of intervals with some extra guarantees. 

https://github.com/sourcecred/sourcecred/blob/55c803ee60e1668d3d3262f0df096d6c30092c29/src/core/interval.js#L21-L29

src/analysis/explorerHomeView.js
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+    Returns the timelines aggregated over all available intervals.
+   */
+  getAggregateTimelines(): {|
+    +credOverTime: DataOverTime,
+    +grainOverTime: DataOverTime,
```
decentralion, line 52: Grain is not a number, so this type signature is unsatisfiable

decentralion (commented):
This review has my thoughts and impressions on this API. I'm going to take some time to think about how I would implement it and chime in with more thoughts after.

src/analysis/explorerHomeView.js
blueridger: Ah yes, I like that.

src/analysis/explorerHomeView.js
blueridger: right, but for the first pass we said totalParticipants, right? does that just mean a static value that doesn't change with time selection?

src/analysis/explorerHomeView.js
blueridger: there we go. accepted.

src/analysis/explorerHomeView.js
blueridger: yep

src/analysis/explorerHomeView.js
blueridger: it will need to be converted to a number to build line graphs. would you rather that happen downstream?

src/analysis/explorerHomeView.js
blueridger: cool cool

src/analysis/explorerHomeView.js
decentralion: Yeah, I think we should keep data accurate/fully faithful as long as possible. When we get to the line graphs, we can do numeric conversion locally as needed.

decentralion:
I took a stab at implementing the interface over in #2524. It's pretty similar to your work, which I think is a good sign. Main differences:

- Different naming, specifically the names build "up" from the underlying domain (a wrapper around CredGraph and Ledger exposing enhanced participant data) rather than "down" from the intended use case (rows in an explorer table)
- I used the `Identity` type which includes id, name, and identity type
- The "builder" class implicitly wraps a view-across-all-time for convenience.
- Elaborated on the APIs for retrieving data (sorting, filtering, retrieving just one participant, etc)

blueridger:
Closing in favor of https://github.com/sourcecred/sourcecred/pull/2524

PR #2522, opened 2020-11-17
grainConfig: optional budgets

elihanover:
This commit adds support for omitted budgets in a SourceCred instance's `config/grain.json` file.

Now the parser takes all fields optionally, and maps them as an Object over the `zeroMissingBudgets` function, which zeroes missing budget parameters (`balancedPerWeek`, `immediatePerWeek`, or `recentPerWeek`).

For example, we omit the `balancedPerWeek` and `immediatePerWeek` fields here:
```js
{
  recentPerWeek: 10,
  recentWeeklyDecayRate: 0.5
}
```

__Test Plan__
Unit tests are provided here for the `zeroMissingBudgets` helper, checking that it corrently handles omitted fields and passes along the other values untouched.  I've also modified the parser tests to reflect the optional nature of these few parameters.

### DIFF
--- a/src/api/grainConfig.js
+++ b/src/api/grainConfig.js
@@ -8,22 +8,37 @@ import {toDiscount} from "../core/ledger/policies/recent";
 
 export type GrainConfig = {|
   +immediatePerWeek: number,
-  +balancedPerWeek: number,
   +recentPerWeek: number,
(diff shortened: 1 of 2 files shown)


decentralion:
Merged #2523.

PR #2523, opened 2020-11-17
grainConfig: optional budgets (alternative to #2522)

elihanover:
This commit adds support for omitted budgets in a SourceCred instance's `config/grain.json` file.

Here, the `GrainConfig` type takes the policy budgets optionally, and passes the responsibilty to handle null values to `toDistributionPolicy`.
```js
export type GrainConfig = {|
  +immediatePerWeek?: number,
  +balancedPerWeek?: number,
  +recentPerWeek?: number,
  +recentWeeklyDecayRate?: number,
  +maxSimultaneousDistributions?: number,
|};
```
```js
export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
  const immediatePerWeek = NullUtil.orElse(x.immediatePerWeek, 0);
  const recentPerWeek = NullUtil.orElse(x.recentPerWeek, 0);
  const balancedPerWeek = NullUtil.orElse(x.balancedPerWeek, 0);
  // ...
```
This is an alternative implementation to sourcecred#2522.  The advantage here is that we consolidate all the configuration checks on the data in `toDistributionPolicy`, as opposed to splitting its responsibilities with the parser.

__Test Plan__
I've modified the parser's unit tests to allow missing budget fields.  In addition, I've modified unit tests for `toDistributionPolicy` to test that missing parameters are handled properly.

### DIFF
--- a/src/api/grainConfig.js
+++ b/src/api/grainConfig.js
@@ -7,62 +7,69 @@ import * as NullUtil from "../util/null";
 import {toDiscount} from "../core/ledger/policies/recent";
 
 export type GrainConfig = {|
-  +immediatePerWeek: number,
-  +balancedPerWeek: number,
-  +recentPerWeek: number,
+  +immediatePerWeek?: number,
+  +balancedPerWeek?: number,
+  +recentPerWeek?: number,
   +recentWeeklyDecayRate?: number,
   +maxSimultaneousDistributions?: number,
 |};
 
 export const parser: C.Parser<GrainConfig> = C.object(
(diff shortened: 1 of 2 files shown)


decentralion (approved):
Looks good, thanks!

@elihanover, I recommend you start creating branches on the sc/sc repo, so that the `test_full` job won't fail. (It only passes when not run from a fork, for security reasons.)

topocount (approved):
Thanks @elihanover! Looks good to me

PR #2524, opened 2020-11-18
Sketch a CredLedger interface

decentralion:
This is a sketch of a data interface for accessing aggregated data from
a CredGraph and a Ledger, i.e. wrapping Cred scores and Grain earnings
together.

This is based on @blueridger's work in #2521.

Main differences:
- Different naming, specifically the names build "up" from the underlying domain (a wrapper around CredGraph and Ledger exposing enhanced participant data) rather than "down" from the intended use case (rows in an explorer table)
- I used the `Identity` type which includes id, name, and identity type
- The "builder" class implicitly wraps a view-across-all-time for convenience.
- Elaborated on the APIs for retrieving data (sorting, filtering, retrieving just one participant, etc)

Test plan: It's just an interface; flow and lint pass.

### DIFF
--- a/src/core/credLedger.js
+++ b/src/core/credLedger.js
@@ -1,0 +1,128 @@
+// @flow
+
+/**
+ * This module exposes a class that accesses participant data, aggregating
+ * between a CredGraph and a Ledger.
+ *
+ * It is useful for cases where you want to view a participant's Cred and Grain
+ * data simultaneously, for example for creating summary dashboards.
+ */
+import {CredGraph} from "./credrank/credGraph";
+import {type Identity, type IdentityType, type IdentityId} from "./identity";
+import {type Grain} from "./ledger/grain";
+import {Ledger} from "./ledger/ledger";
+import {type TimestampMs} from "../util/timestamp";
+import {type IntervalSequence} from "./interval";
+
+/**
+ * Cred and Grain data for a given participant.
+ *
+ * Implicitly has an associated time scope, which will
+ * be the time scope of the TimeScopedCredLedger that generated
+ * this.
+ */
+export type ParticipantData = {|
+  +identity: Identity,
+  // Total Cred earned during the time scope.
+  +cred: number,
+  // Cred earned in each interval within the time scope.
+  +credOverTime: $ReadOnlyArray<number>,
+  // Total Grain earned during the time scope.
+  +grainEarned: Grain,
+  // Grain earned in each interval within the time scope.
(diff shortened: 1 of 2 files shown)


src/core/credLedger.js
```diff
@@ -0,0 +1,128 @@
...
+  ) {
+    this._credGraph = credGraph;
+    this._ledger = ledger;
+    this._startTimeMs = startTimeMs;
+    this._endTimeMs = endTimeMs;
+  }
+
+  participants(options: ParticipantsOptions): $ReadOnlyArray<ParticipantData> {
+    const _ = options;
+    throw new Error("not yet implemented");
```
blueridger, line 117: stub with throws to avoid having to construct the return type. putting that in my back pocket!

src/core/credLedger.js
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Aggregates data across a CredGraph and Ledger.
+ *
+ * By default, it includes data across all time present in the instance.
+ * Callers can call `withTimeScope` to get a `TimeScopedCredLedger` which
+ * returns data that only includes a continuous subset of cred and grain data
+ * across time.
+ */
+export class CredLedger {
```
blueridger, line 59: I'm a big fan of verbose and self-explaining naming over term coining naming. I might name it something like CredAndGrainSummary/History/Ledger.

src/core/credLedger.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  participants(options: ParticipantsOptions): $ReadOnlyArray<ParticipantData> {
+    return this._defaultTimeScope.participants(options);
+  }
+}
+
+export class TimeScopedCredLedger {
+  _credGraph: CredGraph;
+  _ledger: Ledger;
```
blueridger, line 99: I imagine an intermediary data type being saved instead during implementation. Maybe even simply a `ParticipantData[]`

blueridger (approved):
This is suuuper similar to something I was sketching, but there were a few dots I was having trouble connecting and I gave up on it in favor of https://github.com/sourcecred/sourcecred/pull/2521. Kicking myself for not seeking help with my first idea instead, because this was my actual dream. Lesson learned.

src/core/credLedger.js
decentralion: Yeah, although we might not want the totals to be present for the cached data.

src/core/credLedger.js
decentralion: Changed it to `CredGrainView`, can make further tweaks...

PR #2530, opened 2020-11-19
Bugfix(frontend): remove unstyled initial render

topocount:
We're now utilizing MaterialUI's Server-side styling library to collect
and pre-render the needed css for  the initial loading screen.

test plan:
1. yarn start

2. navigate to http://localhost:8080 and disable javascript. Reload the
page and ensure it's styled

3. Reenable javascript, reload the page and ensure everything loads
fully as expected

### DIFF
--- a/src/ui/server.js
+++ b/src/ui/server.js
@@ -4,6 +4,7 @@ import {StyleSheetServer} from "aphrodite/no-important";
 import React from "react";
 import ReactDOMServer from "react-dom/server";
 import {StaticRouter} from "react-router";
+import {ServerStyleSheets} from "@material-ui/core/styles";
 
 import dedent from "../util/dedent";
 import {Assets, rootFromPath} from "../webutil/assets";
@@ -22,13 +23,20 @@ export default function render(
 
   function renderStandardRoute() {
     const bundlePath = locals.assets["main"];
+
+    // Reference for MUI stylesheet injections into SSR HTML:
+    // https://material-ui.com/guides/server-rendering/#handling-the-request
+    const sheets = new ServerStyleSheets();
     const {html, css} = StyleSheetServer.renderStatic(() =>
       ReactDOMServer.renderToString(
-        <StaticRouter location={path} context={context}>
-          <App />
-        </StaticRouter>
+        sheets.collect(
+          <StaticRouter location={path} context={context}>
+            <App />
+          </StaticRouter>
+        )
       )
     );
+
     const page = dedent`\
       <!DOCTYPE html>
       <html>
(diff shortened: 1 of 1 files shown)


src/ui/server.js
```diff
@@ -22,25 +22,30 @@ export default function render(
...
       <head>
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width,initial-scale=1" />
       <link rel="shortcut icon" href="${assets.resolve("/favicon.png")}" />
-      <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
-      <link href="https://fonts.googleapis.com/css?family=Roboto+Condensed" rel="stylesheet">
       <title>SourceCred</title>
       <style>${require("./index.css")}</style>
-      <style data-aphrodite>${css.content}</style>
+      <style id="jss-server-side">${css}</style>
```
wchargin, line 48: We still use Aphrodite, so let’s keep both? I expect that it should be
possible to nest the callbacks.

wchargin (commented):
Lovely! This is what I was hoping the patch would look like.

src/ui/server.js
```diff
@@ -22,25 +23,31 @@ export default function render(
...
+
     const page = dedent`\
       <!DOCTYPE html>
       <html>
       <head>
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width,initial-scale=1" />
       <link rel="shortcut icon" href="${assets.resolve("/favicon.png")}" />
-      <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
-      <link href="https://fonts.googleapis.com/css?family=Roboto+Condensed" rel="stylesheet">
```
wchargin, line 40: Is this related? I am happy to drop these, but should we move that to a
different PR?

yarn.lock
```diff
@@ -7261,6 +7261,11 @@ nan@^2.12.1:
   resolved "https://registry.yarnpkg.com/nan/-/nan-2.14.1.tgz#d7be34dfa3105b91494c3147089315eff8874b01"
   integrity sha512-isWHgVjnFjh2x2yuJ/tj3JbwoHu3UC2dX5G/88Cm24yB6YopVgxvBObDY7n5xW6ExmFhJpSEQqFPvq9zaXc8Jw==
 
+nan@^2.14.0:
```
wchargin, line 7264: Why are we changing the lockfile when `package.json` is unchanged?

wchargin (commented):
I can’t test this because of #2531, but the patch looks good to me
modulo inlines.

src/ui/server.js
topocount: Yup! having the nested callbacks works fine

src/ui/server.js
topocount: These lines have been un-removed from the pull request

yarn.lock
topocount: This is no longer necessary after the reversion

decentralion (approved):

wchargin (commented):
post-hoc LGTM; thanks!

PR #2532, opened 2020-11-20
Add TableState webutil for paginated+filtered+sorted arrays

blueridger:
# Notes
This is the product of my side quest to learn React. It is a state manager for tables built in ReactNodes. It encapsulates Pagination, Sort, and Filtering.

# Design choices
I chose to use a function/type architecture instead of class architecture because it seemed to better align the with React "use..." architecture, and because it results in much more readable syntax.

In opposition to a discussion DL and I was having on a predecessor PR (https://github.com/sourcecred/sourcecred/pull/2520), I chose to sort the original unfiltered table. This is to enable filtering without sorting, making filtering faster, which is important because sorting is a low-frequency mouse-click operation whereas filtering wants to be a high-frequency key-press operation.

# Automated Testing
`yarn test` however this code is **not** tested by unit tests. I find this acceptable because this code is in the front end and manual UI testing is sufficiently indicative of code health. There is also no precedent that I see of unit testing React code.

# Manual Testing
See test plan in https://github.com/sourcecred/sourcecred/pull/2616

If you want to try filtering for yourself, add in the test filter inputs using this code:
```
<input type="text" onChange={(event) => {
        const value = event.target.value;
        ts.createOrUpdateFilterFn("eventName", (ledgerEvent) => ledgerEvent.action.type.replace("_", " ").toLowerCase().includes(value.toLowerCase()))}}/>
<input type="text" onChange={(event) => {
        const value = event.target.value;
        ts.createOrUpdateFilterFn("eventName2", (ledgerEvent) => ledgerEvent.action.type.replace("_", " ").toLowerCase().includes(value.toLowerCase()))
}}/>
```

### DIFF
--- a/src/webutil/tableState.js
+++ b/src/webutil/tableState.js
@@ -1,0 +1,273 @@
+// @flow
+
+import {type PluckFn} from "../util/sortBy";
+import sortBy from "../util/sortBy";
+import {useMemo, useState, useCallback} from "react";
+import deepFreeze from "deep-freeze";
+
+export const SortOrders: {|
+  +ASC: "asc",
+  +DESC: "desc",
+|} = deepFreeze({
+  ASC: "asc",
+  DESC: "desc",
+});
+type SortOrderTypes = "asc" | "desc";
+
+export const DEFAULT_SORT = SortOrders.ASC;
+
+/**
+  A state manager for a React data table that provides encapsuled hooks for
+  Sort, Pagination, and Filtering. Useful for any table that needs at least
+  one of those features, and particularly useful for managing combinations.
+
+  EXAMPLES:
+  Import:
+      import {
+        useTableState,
+        SortOrders,
+        DEFAULT_SORT
+      } from "../../webutil/tableState";
+  Preparation:
+      const DATE_SORT = deepFreeze({
+        name: Symbol("Date"),
+        fn: (e: LedgerEvent) => e.ledgerTimestamp
+      });
+      const PAGINATION_OPTIONS = deepFreeze([25, 50, 100]);
+      ...
+      const ts = useTableState(eventLog, {
+        initialRowsPerPage: PAGINATION_OPTIONS[0],
+        initialSort: {
+          sortName: DATE_SORT.name,
+          sortOrder: SortOrders.DESC,
+          sortFn: DATE_SORT.fn,
+        }
+      });
+  Pagination:
+      <TablePagination
+        rowsPerPageOptions={PAGINATION_OPTIONS}
+        component="div"
+        count={ts.length}
+        labelRowsPerPage="Rows"
+        rowsPerPage={ts.rowsPerPage}
+        page={ts.pageIndex}
+        onChangePage={(event, newPage) => ts.setPageIndex(newPage)}
+        onChangeRowsPerPage={(event) => ts.setRowsPerPage(event.target.value)}
+      />
+  Filtering:
+      <input type="text" onChange={(event) => {
+        // Do not include an Event object from the DOM in the filter function
+        // as Event objects do not persist.
+        const value = event.target.value;
+        ts.createOrUpdateFilterFn(
+          "filterEventName",
+          (ledgerEvent) => ledgerEvent.action.type.replace("_", " ")
+            .toLowerCase()
+            .includes(value.toLowerCase())
+        )    
+      }}/>
+  Sorting:
+      <TableSortLabel
+          active={ts.sortName === DATE_SORT.name}
+          direction={
+            ts.sortName === DATE_SORT.name ? ts.sortOrder : DEFAULT_SORT
+          }
+          onClick={() => ts.setSortFn(DATE_SORT.name, DATE_SORT.fn)}
+      >Date</TableSortLabel>
+ */
+export type TableState<T> = {|
+  /**
+    A stateful representation of the current page data.
+   */
+  +currentPage: $ReadOnlyArray<T>,
+  /**
+    A stateful representation of the length of the filtered unpaginated data.
+   */
+  +length: number,
+  /**
+    A stateful representation of the current page index.
+   */
+  +pageIndex: number,
+  /**
+    A stateful representation of the number of rows per page.
+   */
+  +rowsPerPage: number,
+  /**
+    A stateful representation of the current sort name (the last sort name
+    passed to setSortFn).
+   */
+  +sortName: symbol,
+  /**
+    A stateful representation of the current order,
+    represented by "asc" or "desc".
+   */
+  +sortOrder: SortOrderTypes,
+  /**
+    A function that changes the currentPage to represent the indicated index.
+   */
+  +setPageIndex: (number) => void,
+  /**
+    A function that sets the rows per page and updates the pageIndex
+    to an equivalent place.
+   */
+  +setRowsPerPage: (number) => void,
+  /**
+    A function that adds a filter to the map of filters or overwrites an
+    existing filter that matches the key. Every filter stored will be applied
+    together as an AND operation (results will only include elements
+    that pass all filters).
+    Sets pageIndex to 0.
+
+    string: A key that represents the filter in the map.
+    function?: A function that returns false if an element should be excluded
+      from the result. Null to remove the filter at the key.
+   */
+  +createOrUpdateFilterFn: (string, null | ((T) => boolean)) => void,
+  /**
+    A function that sorts the data. Current pageIndex is maintained.
+
+    string: A key representing the sort field. It is is used to toggle sort
+      order and increase efficiency.
+    function?: A function that returns the property of the element that
+      should be used in the sort comparator.
+   */
+  +setSortFn: (symbol, PluckFn<T>) => void,
+|};
+
+type FilterFns<T> = {[string]: null | ((T) => boolean)};
+
+/**
+  A builder function for initializing a new TableState for use in a ReactNode.
+  It encapsulates Sort, Pagination, and Filtering.
+ */
+export function useTableState<T>(
+  // The base array. It should ideally be memoized before being passed in.
+  data: Array<T>,
+  initialOptions?: {
+    // The initial rows per page. Defaults to unpaginated.
+    initialRowsPerPage?: number,
+    // Defaults to unsorted (maintains order of the original data).
+    initialSort?: {
+      // The starting order of the initial sort, represented by "asc" or "desc".
+      sortOrder: SortOrderTypes,
+      // A key representing the sort field. It is is used to toggle sort order
+      // and increase efficiency.
+      sortName: symbol,
+      // A function that returns the property of the element that should be used
+      // in the sort comparator.
+      sortFn: PluckFn<T>,
+    },
+  }
+): TableState<T> {
+  const optionDefaults = {
+    initialRowsPerPage: Infinity,
+    initialSort: {},
+    ...initialOptions,
+  };
+  const sortDefaults = {
+    sortOrder: SortOrders.ASC,
+    sortName: Symbol(),
+    sortFn: null,
+    ...optionDefaults.initialSort,
+  };
+  const [sortName, setSortName] = useState(sortDefaults.sortName);
+  const [sortOrder, setSortOrder] = useState<SortOrderTypes>(
+    sortDefaults.sortOrder
+  );
+  const [sortFn, setSortFn] = useState(() => sortDefaults.sortFn);
+  const [pageIndex, setPageIndex] = useState(0);
+  const [rowsPerPageUnsafe, setRowsPerPageInternal] = useState(
+    optionDefaults.initialRowsPerPage
+  );
+  const rowsPerPage = useMemo(() => {
+    if (rowsPerPageUnsafe !== Infinity) return rowsPerPageUnsafe;
+    else return data.length;
+  }, [rowsPerPageUnsafe, data]);
+  const [filterFns, setFilterFns] = useState<FilterFns<T>>({});
+
+  const sortedArray = useMemo(() => {
+    const array = data.slice();
+    if (sortFn) {
+      sortBy(array, sortFn);
+      if (sortOrder === SortOrders.DESC) array.reverse();
+    }
+    return array;
+  }, [data, sortFn, sortOrder]);
+
+  const filteredArray = useMemo(() => {
+    const reducer = (array, key) => {
+      const fn = filterFns[key];
+      if (fn) return array.filter(fn);
+      return array;
+    };
+    return Object.keys(filterFns).reduce(reducer, sortedArray);
+  }, [sortedArray, filterFns]);
+
+  const setRowsPerPage = useCallback(
+    (newRowsPerPage: number) => {
+      // Keep the user from losing their place
+      // when changing # of results per page
+      const intRowsPerPage = parseInt(newRowsPerPage, 10);
+      const currentFirstVisibleRow = pageIndex * rowsPerPage;
+      const newPage = Math.floor(currentFirstVisibleRow / intRowsPerPage);
+
+      setRowsPerPageInternal(intRowsPerPage);
+      setPageIndex(newPage);
+    },
+    [pageIndex, rowsPerPage, data]
+  );
+
+  const setSortFnCallback = useCallback(
+    (newSortName: symbol, newSortFn: PluckFn<T>) => {
+      if (newSortName === sortName) {
+        setSortOrder((currentSortOrder) =>
+          currentSortOrder === SortOrders.DESC
+            ? SortOrders.ASC
+            : SortOrders.DESC
+        );
+      } else {
+        setSortOrder(DEFAULT_SORT);
+        setSortFn(() => newSortFn);
+      }
+      setSortName(newSortName);
+    },
+    [sortName]
+  );
+
+  const createOrUpdateFilterFn = useCallback(
+    (key: string, newFilterFn: null | ((T) => boolean)) => {
+      setPageIndex(0);
+      setFilterFns((fns): FilterFns<T> => {
+        return {
+          ...fns,
+          [key]: newFilterFn,
+        };
+      });
+    },
+    []
+  );
+
+  const currentPage = useMemo(
+    () =>
+      filteredArray.slice(
+        pageIndex * rowsPerPage,
+        pageIndex * rowsPerPage + rowsPerPage
+      ),
+    [filteredArray, pageIndex, rowsPerPage]
+  );
+
+  const length = filteredArray.length;
+
+  return {
+    pageIndex,
+    setPageIndex,
+    rowsPerPage,
+    sortName,
+    sortOrder,
+    setRowsPerPage,
+    createOrUpdateFilterFn,
+    setSortFn: setSortFnCallback,
+    currentPage,
+    length,
+  };
+}
--- a/src/util/sortBy.js
+++ b/src/util/sortBy.js
@@ -1,6 +1,6 @@
 //@flow
 
-type PluckFn<T> = (T) => any;
+export type PluckFn<T> = (T) => any;
 
 const identity = (x) => x;
 


yarn.lock
```diff
@@ -7261,6 +7261,11 @@ nan@^2.12.1:
   resolved "https://registry.yarnpkg.com/nan/-/nan-2.14.1.tgz#d7be34dfa3105b91494c3147089315eff8874b01"
   integrity sha512-isWHgVjnFjh2x2yuJ/tj3JbwoHu3UC2dX5G/88Cm24yB6YopVgxvBObDY7n5xW6ExmFhJpSEQqFPvq9zaXc8Jw==
 
+nan@^2.14.0:
```
blueridger, line 7264: Not sure why this got added. halp.

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+  const sortDefaults = defaults.initialSort || {};
+  const [originalArray, setOriginalArray] = useState([]);
+  const [filteredArray, setFilteredArray] = useState([]);
+  const [sortName, setSortName] = useState(sortDefaults.sortName || "");
+  const [sortOrder, setSortOrder] = useState(
+    sortDefaults.sortOrder || SortOrders.ASC
+  );
+  const [pageIndex, setPageIndex] = useState(0);
+  const [rowsPerPage, setRowsPerPageInternal] = useState(
+    defaults.initialRowsPerPage || 20
```
hammadj, line 99: Would be good to extract the `20` as a constant `DEFAULT_ROWS_PER_PAGE` before the function definition so its not so much a "magic number".

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+    Sets pageIndex to 0.
+   */
+  +setData: ($ReadOnlyArray<T>) => void,
+|};
+
+/**
+  A builder function for initializing a new TableState for use in a ReactNode.
+  It encapsulates Sort, Pagination, and Filtering.
+ */
+export function useTableState<T>(defaults?: {
```
hammadj, line 78: Maybe a name like `defaultOptions` or `defaultState` would be more descriptive

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+  const currentPage = useMemo(
+    () =>
+      filteredArray.slice(
+        pageIndex * rowsPerPage,
+        pageIndex * rowsPerPage + rowsPerPage
+      ),
+    [filteredArray, pageIndex, rowsPerPage]
+  );
+
+  const length = useMemo(() => filteredArray.length, [filteredArray]);
```
hammadj, line 170: No need for useMemo here since `.length` isn't computationally intensive, and since its a plain number it will be strict equal between renders as well (e.g. you might need useMemo if you are returning an object that you want to be referentially equal between renders as a performance optimization to prevent re-renders of pure components, but its not needed here and would actually slow it down).

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+    _filterFns.set(key, newFilterFn);
+    _filter(originalArray.slice());
+  };
+
+  const setData = (data: $ReadOnlyArray<T>) => {
+    setPageIndex(0);
+    const _array = cloneDeep(data);
+    setOriginalArray(_array);
+    _sortAndFilter(_array.slice(), sortOrder);
+  };
```
hammadj, line 159: Would wrap this in a `useCallback` to prevent creating a new function on every render, which might help prevent re-renders (which is valuable for large list type components)

hammadj (commented):
useTableState looks amazing, deserves to be its own open source package haha, super impressive work considering you're learning React! You're lucky you're learning it with Hooks, its a much better dev experience than the old class components where you had to "unlearn" the old way to get used to the new way with Hooks.

Made a bunch of suggestions / nits, overall looking good though.

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+  const [sortName, setSortName] = useState(sortDefaults.sortName || "");
+  const [sortOrder, setSortOrder] = useState(
+    sortDefaults.sortOrder || SortOrders.ASC
+  );
+  const [pageIndex, setPageIndex] = useState(0);
+  const [rowsPerPage, setRowsPerPageInternal] = useState(
+    defaults.initialRowsPerPage || 20
+  );
+  let _sortFn: PluckFn<T> | null = sortDefaults.sortFn || null;
+  const _filterFns: Map<string, null | ((T) => boolean)> = new Map();
```
hammadj, line 102: This Map is outside of React's state management, so its likely to cause bugs / issues with re-rendering components when the filter functions change. Its best to keep everything within Reacts state management. Additionally, Map isn't a good fit because when you call `.set()`, the data inside changes but the Map itself remains referentially equal to its previous state, which means React doesn't detect any changes and therefore doesn't re-render the UI.

A better way to do this would be to also put this in useState:
```ts
type FilterFnState = { [key: string]: null | ((T) => boolean) };

// ...
const [filterFns, setFilterFns] = useState<FilterFnState>({});

const createOrUpdateFilterFn = (
     key: string,
     newFilterFn: $Values<FilterFnState>
   ) => {
     setPageIndex(0);
     // if you pass in a function instead of a value, you can get the current filter state as a param and return the new state
     setFilterFns(oldFilters => ({
       ...oldFilters,
       [key]: newFilterFn,
     }));
     // no need to do .slice since .filter returns a new array instead of mutating the input array
     _filter(originalArray);
   };
```

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+      if (_sortOrder === SortOrders.DESC) _array.reverse();
+    }
+    setOriginalArray(_array);
+    _filter(_array);
+  };
+
+  const _filter = (_array: Array<T>) => {
+    _filterFns.forEach((fn) => {
+      if (fn) _array = _array.filter(fn);
+    });
```
hammadj, line 116: Would be better to use .reduce here. Also shouldn't mutate the params, better to initialize a new variable:
```suggestion
    let filteredArray = _array;
    _filterFns.forEach((fn) => {
      if (fn) filteredArray = filteredArray.filter(fn);
    });
```

yarn.lock
hammadj: I got that too, seems like someone didn't commit an updated yarn.lock file from some dependency update. Should be fine

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+export function useTableState<T>(defaults?: {
+  initialRowsPerPage?: number,
+  initialSort?: {
+    // The starting order of the initial sort, represented by "asc" or "desc".
+    sortOrder: SortOrderTypes,
+    // A key representing the sort field. It is is used to toggle sort order and increase efficiency.
+    sortName: string,
+    // A function that returns the property of the element that should be used in the sort comparator.
+    sortFn: PluckFn<T>,
+  },
```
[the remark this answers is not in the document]
blueridger: The idea is that `initialSort` can be optional, but if it's included, all of the subfields must be included.

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+import {type PluckFn} from "../util/sortBy";
+import sortBy from "../util/sortBy";
+import cloneDeep from "lodash.clonedeep";
+import {useMemo, useState} from "react";
+
+export const SortOrders = {
+  ASC: "asc",
+  DESC: "desc",
+};
+type SortOrderTypes = "asc" | "desc";
```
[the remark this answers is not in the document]
blueridger: I did find that in my research. The reason I did it this way was because the TableSortLabel can use the lowercase strings as input, which makes this line cleaner:
```
direction={
                    ts.sortName === DATE_SORT[0] ? ts.sortOrder : SortOrders.ASC
                  }
```

I didn't check if it would accept uppercase though. I'll play with it to see.

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+    // A key representing the sort field. It is is used to toggle sort order and increase efficiency.
+    sortName: string,
+    // A function that returns the property of the element that should be used in the sort comparator.
+    sortFn: PluckFn<T>,
+  },
+}): TableState<T> {
+  defaults = defaults || {};
+  const sortDefaults = defaults.initialSort || {};
+  const [originalArray, setOriginalArray] = useState([]);
+  const [filteredArray, setFilteredArray] = useState([]);
```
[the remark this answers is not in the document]
blueridger: accepted

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+  initialSort?: {
+    // The starting order of the initial sort, represented by "asc" or "desc".
+    sortOrder: SortOrderTypes,
+    // A key representing the sort field. It is is used to toggle sort order and increase efficiency.
+    sortName: string,
+    // A function that returns the property of the element that should be used in the sort comparator.
+    sortFn: PluckFn<T>,
+  },
+}): TableState<T> {
+  defaults = defaults || {};
```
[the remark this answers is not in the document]
blueridger: accepted

src/webutil/tableState.js
```diff
@@ -0,0 +1,185 @@
...
+  const sortDefaults = defaults.initialSort || {};
```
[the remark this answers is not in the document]
blueridger: accepted

src/webutil/tableState.js
blueridger: accepted

src/webutil/tableState.js
blueridger: yeah... I'm thinking `initialOptions`

src/webutil/tableState.js
blueridger: I don't fully understand the nuance here because I still don't really understand React under the hood, but I'll take this suggestion and seek more understanding later.

src/webutil/tableState.js
blueridger: I'll look into it.

src/webutil/tableState.js
blueridger: Again, don't really get the nuance, but I'll play with this.

src/webutil/tableState.js
blueridger: accepted

yarn.lock
topocount: @blueridger This is from an reverted merge. Rebasing should eliminate it

src/ui/components/LedgerViewer.js
```diff
@@ -59,8 +61,7 @@ export const LedgerViewer = ({
   const {ledger} = useLedger();
   const classes = useStyles();
   const [allocation, setAllocation] = useState<Allocation | null>(null);
-  const [page, setPage] = useState(0);
-  const [rowsPerPage, setRowsPerPage] = React.useState(25);
+  const DATE_SORT = ["Date", (e) => e.ledgerTimestamp];
```
BrianBelhumeur, line 64: nit: I advocate for named vs positional attributes. E.g.:
```suggestion
  const DATE_SORT = {
      name: "Date", 
      fn: (e) => e.ledgerTimestamp,
  };
```
then you can access them below like:
```
    sortName: DATE_SORT.name,
    sortOrder: SortOrders.DESC,
    sortFn: DATE_SORT.fn,
```

The reason being is it takes less cognitive load to remember positions (though presumably typing takes care of accidentally using the wrong one?) and it's more a descriptive when reading back the code.

BrianBelhumeur (approved):
@hammadj covered a lot of great points, so I only have a nit to add. Great work @blueridger! You're catching on to React very quickly. 👍

topocount (commented):
@blueridger This is awesome! I recommend adding some more to the docstring about use cases and maybe a simple example sketch.

src/webutil/tableState.js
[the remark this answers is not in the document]
blueridger: Yeah as I guessed, the TableSortLabel wants it lowercase. So we can make the constant names lowercase to add this sugar or we can leave it. I'll probably leave it.

src/webutil/tableState.js
blueridger: actually, I think I will just make this a required field instead, since there is no real case where you can get away with not providing it.

src/webutil/tableState.js
blueridger: accepted

src/webutil/tableState.js
blueridger: accepted

src/ui/components/LedgerViewer.js
blueridger: accepted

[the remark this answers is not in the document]
blueridger: accepted

src/webutil/tableState.js
blueridger: accepted

src/ui/components/LedgerViewer.js
```diff
@@ -59,8 +65,11 @@ export const LedgerViewer = ({
   const {ledger} = useLedger();
   const classes = useStyles();
   const [allocation, setAllocation] = useState<Allocation | null>(null);
-  const [page, setPage] = useState(0);
-  const [rowsPerPage, setRowsPerPage] = React.useState(25);
+  const [DATE_SORT] = useState({
+    name: "Date",
+    fn: (e: LedgerEvent) => e.ledgerTimestamp,
+  });
+  const [PAGINATION_OPTIONS] = useState([25, 50, 100]);
```
hammadj, line 72: These are both using useState, but the state is never being changed here (no second `setPaginationOptions` or `setDateSort` options assigned in the return value). If the state is never updated, you can just define the value outside of the LedgerViewer component as a constant.

src/ui/components/LedgerViewer.js
```diff
@@ -99,21 +102,29 @@ export const LedgerViewer = ({
...
-              <TableCell align="right">Date</TableCell>
+              <TableCell align="right">
+                <TableSortLabel
+                  active={ts.sortName === DATE_SORT.name}
+                  direction={
+                    ts.sortName === DATE_SORT.name ? ts.sortOrder : DEFAULT_SORT
+                  }
+                  onClick={() => ts.setSortFn(DATE_SORT.name, DATE_SORT.fn)}
+                >
+                  Date
```
hammadj, line 113:
```suggestion
                  {DATE_SORT.name}
```

src/webutil/tableState.js
```diff
@@ -0,0 +1,236 @@
...
+  const [sortName, setSortName] = useState(sortDefaults.sortName);
+  const [sortOrder, setSortOrder] = useState<SortOrderTypes>(
+    sortDefaults.sortOrder
+  );
+  const [sortFn, setSortFn] = useState(() => sortDefaults.sortFn);
+  const [pageIndex, setPageIndex] = useState(0);
+  const [rowsPerPage, setRowsPerPageInternal] = useState(
+    optionDefaults.initialRowsPerPage
+  );
+  type FilterFns = {[string]: null | ((T) => boolean)};
```
hammadj, line 155: Usually types are defined outside of function bodies, should move it out for consistency sake and readability.

src/webutil/tableState.js
```diff
@@ -0,0 +1,236 @@
...
+    return array;
+  }, [data, sortFn, sortOrder]);
+
+  const filteredArray = useMemo(() => {
+    const reducer = (array, key) => {
+      const fn = filterFns[key];
+      if (fn) return array.filter(fn);
+      return array;
+    };
+    return Object.keys(filterFns).reduce(reducer, sortedArray);
```
hammadj, line 173: +1 on filtering the sorted array instead of sorting the filtered array, this means that people will be able to change filters without needing to recalculate the sort (opposite isn't true however, updating sort function will trigger filter to run again, but filtering should be cheaper than sorting so not a biggie at all)

src/webutil/tableState.js
```diff
@@ -0,0 +1,236 @@
...
+
+  const setRowsPerPage = useCallback((newRowsPerPage: number) => {
+    // Keep the user from losing their place when changing # of results per page
+    const intRowsPerPage = parseInt(newRowsPerPage, 10);
+    const currentFirstVisibleRow = pageIndex * rowsPerPage;
+    const newPage = Math.floor(currentFirstVisibleRow / intRowsPerPage);
+
+    setRowsPerPageInternal(intRowsPerPage);
+    setPageIndex(newPage);
+  });
```
hammadj, line 184: missing second param here (surprised it didnt get caught by linter, should def fix that)
```suggestion
  }, [pageIndex, rowsPerPage]);
```

src/webutil/tableState.js
```diff
@@ -0,0 +1,236 @@
...
+
+  const setSortFnCallback = useCallback(
+    (newSortName: string, newSortFn: PluckFn<T>) => {
+      if (newSortName === sortName) {
+        setSortOrder(
+          sortOrder === SortOrders.DESC ? SortOrders.ASC : SortOrders.DESC
+        );
```
hammadj, line 191: Also missing array of dependencies in this useCallback. Note that there's a way to not have to include them in here by passing a function into setSortOrder instead, the param will be the current value:

```suggestion
        setSortOrder(currentSortOrder =>  
          currentSortOrder === SortOrders.DESC ? SortOrders.ASC : SortOrders.DESC
        );
```

src/webutil/tableState.js
```diff
@@ -0,0 +1,236 @@
...
+      const setterFn: (FilterFns) => FilterFns = (fns) => {
+        return {
+          ...fns,
+          [key]: newFilterFn,
+        };
+      };
+      setPageIndex(0);
+      setFilterFns(setterFn);
+    }
+  );
```
hammadj, line 211: Although this one doesn't have any dependencies on values outside the callback, you still need to pass an empty array as the second param to useCallback in order to indicate that

src/webutil/tableState.js
```diff
@@ -0,0 +1,236 @@
...
+
+  const createOrUpdateFilterFn = useCallback(
+    (key: string, newFilterFn: null | ((T) => boolean)) => {
+      const setterFn: (FilterFns) => FilterFns = (fns) => {
+        return {
+          ...fns,
+          [key]: newFilterFn,
+        };
+      };
```
hammadj, line 207: If you specify this inline to `setFilterFns` you shouldn't have to add those type annotations (would be inferred)

src/ui/components/LedgerViewer.js
blueridger: aha, accepted

src/ui/components/LedgerViewer.js
blueridger: accepted

src/webutil/tableState.js
blueridger: accepted

src/webutil/tableState.js
blueridger: accepted

src/webutil/tableState.js
blueridger: accepted

src/webutil/tableState.js
blueridger: sure

src/webutil/tableState.js
blueridger: Flow throws a fit about not knowing which case to pick due to ambiguity yadda yadda. I did find an improvement that allows me to specify the function inline with only 1 type annotation, though.

hammadj (approved):
Great work :)

src/ui/components/LedgerViewer.js
```diff
@@ -51,6 +57,13 @@ const useStyles = makeStyles((theme) => {
   };
 });
 
+const DATE_SORT = {
+  name: "Date",
+  fn: (e: LedgerEvent) => e.ledgerTimestamp,
+};
+
+const PAGINATION_OPTIONS = [25, 50, 100];
```
decentralion, line 65: nit: i suggest `deepFreeze`ing so that the defaults can't get modified.

src/ui/components/LedgerViewer.js
```diff
@@ -51,6 +57,13 @@ const useStyles = makeStyles((theme) => {
   };
 });
 
+const DATE_SORT = {
+  name: "Date",
+  fn: (e: LedgerEvent) => e.ledgerTimestamp,
+};
```
decentralion, line 63: nit: i suggest `deepFreeze`ing so that the defaults can't get modified.

src/webutil/tableState.js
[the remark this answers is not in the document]
decentralion: For another take on how we could set up options like this, take a look at the Direction const in core/graph:

https://github.com/sourcecred/sourcecred/blob/651db6de017cc74dfc85f16960144d0849fab105/src/core/graph.js#L151-L160

It sounds like in this case we can't take that approach due to the lowercase keys being expected by TableSortLabel. However, I do recommend using deepFreeze -- allowing something to modify these values at runtime would never be a good thing.

src/webutil/tableState.js
```diff
@@ -0,0 +1,244 @@
...
+  ASC: "asc",
+  DESC: "desc",
+};
+type SortOrderTypes = "asc" | "desc";
+
+export const DEFAULT_SORT = SortOrders.ASC;
+
+/**
+  A state manager for a React data table that provides encapsuled hooks for Sort, Pagination, and Filtering.
+  Useful for any table that needs at least one of those features, and particularly useful for managing combinations.
```
decentralion, line 17: Nit; wrap to 80 char

src/webutil/tableState.js
```diff
@@ -0,0 +1,244 @@
...
+|};
+
+type FilterFns<T> = {[string]: null | ((T) => boolean)};
+
+/**
+  A builder function for initializing a new TableState for use in a ReactNode.
+  It encapsulates Sort, Pagination, and Filtering.
+ */
+export function useTableState<T>(
+  // The base array. It should ideally be memoized before being passed in.
```
decentralion, line 122: What do you mean by memoizing the array?

src/webutil/tableState.js
```diff
@@ -0,0 +1,244 @@
...
+      sortOrder: SortOrderTypes,
+      // A key representing the sort field. It is is used to toggle sort order and increase efficiency.
+      sortName: string,
+      // A function that returns the property of the element that should be used in the sort comparator.
+      sortFn: PluckFn<T>,
+    },
+  }
+): TableState<T> {
+  const optionDefaults = {
+    initialRowsPerPage: data.length,
```
decentralion, line 139: It's kind of a code smell to me that the default is dependent on the data. I imagine it could result in issues where e.g. the data is set to have 5 rows, and then the data is updated to have 6 rows, and now the table is only showing the first 5 rows, resulting in user confusion.

That said: do we actually want the default to be "show all data"? Since this is UI code, we can imagine cases where' that's quite bad, e.g. it tries to render 20,000 rows and hangs. Maybe go with a safer default like 25?

src/webutil/tableState.js
```diff
@@ -0,0 +1,244 @@
...
+  const [rowsPerPage, setRowsPerPageInternal] = useState(
+    optionDefaults.initialRowsPerPage
+  );
+  const [filterFns, setFilterFns] = useState<FilterFns<T>>({});
+
+  const sortedArray = useMemo(() => {
+    const array = data.slice();
+    if (sortFn) {
+      sortBy(array, sortFn);
+      if (sortOrder === SortOrders.DESC) array.reverse();
```
decentralion, line 164: if we want a perf improvement here, we could just sort with a reversed comparator.
looking at the implementation, we could do this either by:
1. specifying that PluckFns have type `(T) => number` and then doing `sortBy(array, (x) => -sortFn(x))`
or
2. creating a new `sortByReversed` function

either way, we'd save the call to `array.reverse()`. 

probably not needed for now though.

decentralion (commented):
Seems like a well designed class. The bit that feels the strangest to me is the way that the sort and filter objects have string names which are load bearing. So we could run into issues: collisions, or cases where the producer renames the string key and not all consumers realize, etc.

Have you considered using [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) instead of strings here? Each sort or filter object can have a Symbol. You can still give them names for convenience, but they're guaranteed to be unique, so there will never be collisions. And since they need to be passed by reference, it would be impossible to get a desync where the producer tries to change the key and the consumer doesn't get the memo. Flow would catch it.

Also, it's nice to have some real usage available to review alongside seeing the design. However, I'd prefer for that to be in a separate PR -- they are semantically distinct commits and also merit different review and test plans. The first PR (adding TableState) merits design discussion and detailed code review, and the second (migrating LedgerViewer) would mostly just need regression testing and visual inspection. Having these as two separate review steps helps ensure things don't fall through the cracks.

Finally, it'd be nice to also see an example of the filtering in action, e.g. by adding an event type filter to the ledger viewer. I'd encourage writing it before merging this PR, as another check on the design, but wouldn't block merging this either.

src/ui/components/LedgerViewer.js
blueridger: accepted

src/ui/components/LedgerViewer.js
blueridger: accepted

blueridger:
> Have you considered using [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) instead of strings here?

Love it! Didn't know about this.

blueridger:
Pulling out the LedgerViewer changes into a separate PR.

blueridger:
> it'd be nice to also see an example of the filtering in action

Did you see that in the Manual Testing section of this PR, I've already included a functional code snippet of an event type filter, and have included it in the linked video demo? I didn't want to commit it since I think the UX would need hammered out a little more. Does that fulfill your request sufficiently?

src/webutil/tableState.js
blueridger: That it should be referentially equivalent on a re-render, wrapped in a react memo/state. Is there a better way to say this?

src/webutil/tableState.js
blueridger: wrapping all to 80, and adding a vertical ruler to my IDE for the future.

src/webutil/tableState.js
blueridger: This issue is that it is the responsibility of the UI view to create pagination controls, and the TableState cannot coerce the UI view to use pagination. I prefer the unidirectional semantics of TableState offering Sort, Filter, and Pagination and the UI view cherry-picking which it needs.

src/webutil/tableState.js
blueridger: Yeah I had that on my mind, but this way it supports both number and string sorts, which will be really nice from a UX perspective.

src/webutil/tableState.js
```diff
@@ -0,0 +1,273 @@
...
+  );
+
+  const createOrUpdateFilterFn = useCallback(
+    (key: string, newFilterFn: null | ((T) => boolean)) => {
+      setPageIndex(0);
+      setFilterFns((fns): FilterFns<T> => {
+        return {
+          ...fns,
+          [key]: newFilterFn,
+        };
```
blueridger, line 244: unfortunately, I was not able to get symbols to work for the filterFns. I had to use `// $FlowFixMe` since symbols are not supported as keys by flow, and even then, it didn't work. I'd like to continue using strings as a good-enough solution, with hopes to maybe figure out why symbols don't work in the future.

PR #2533, opened 2020-11-20
Revert "Sourcecred Gitcoin Plugin 2 (#2452)"

### LINKED ISSUE #2351: SourceCred require broken in observable

### LINKED ISSUE #2531: Build broken without system-level Postgres dep
When I run `yarn` on #2452 or later, I get a fatal error:

```
error /home/wchargin/git/sourcecred/node_modules/libpq: Command failed.
Exit code: 1
Command: node-gyp rebuild
Arguments: 
Directory: /home/wchargin/git/sourcecred/node_modules/libpq
Output:
...
You need to install postgresql-server-dev-X.Y for building a server-side extension or libpq-dev for building a client-side application.
gyp: Call to '/usr/bin/pg_config --libdir' returned exit status 1 while in binding.gyp. while trying to load binding.gyp
```

I’m assuming that this affects all end users since it affects both
@topocount and me (i.e., it’s not just a bad setup on one person’s
machine) and the `pg-native` dep is in our `dependencies`, not
`dev-dependencies` or the like.

Unless we have a very good reason, this isn’t okay. We can’t expect all
end users of SourceCred to have Postgres dev tools installed.

This is a development-blocking issue. I can’t test pull requests for
which my review has been requested because I can’t build our codebase.

Why do we have a Postgres dep? Is it for a local mirror, for connecting
to a remote GitCoin-hosted Postgres server, or something else?

wchargin:
Summary:
This reverts commit 0387b7d70f79a80233ea71f4261ec6ceb345e072.

That commit introduced a dependency on the `pg-native` NPM module, but
that module requires non-standard system packages to be installed on end
user machines. As a result, you cannot `npm install sourcecred` without
first running `sudo apt-get install libpq-dev`, which is far from a
usual Node workflow and is too high a barrier of entry for users.

Fixes #2531.

Test Plan:
On my machine, which does not have `libpq-dev` installed, running `yarn`
fails before this commit and passes after it, and tests still pass.

wchargin-branch: revert-gitcoin-plugin


topocount (approved):
Thanks for handling this William

PR #2534, opened 2020-11-20
Ledger: Add grain distribution history to accounts

blueridger:
# Notes
This adds an allocationHistory property to the Accounts structure provided by the Ledger. The new AllocationReceipt type offers three attributes:
- `allocationId` provides a way to hook into the new Allocation and Distribution APIs (https://github.com/sourcecred/sourcecred/pull/2550) in order to query for related activity on the same allocation or distribution, as well as to query the Allocation policy type.
- `grainReceipt` provides the grain amount as well as the original IdentityId that it was distributed to, which may not be the same as the account's id due to identity merges.
- `timestampMs` provides easy access to the distribution timestamp. Although this would already be accessible via a 3 step operation (`allocationId>distributionId>distribution>credTimestamp`), I am including it for ease of access since I anticipate it being a high-traffic property.

For the Merge Identities action, I chose to also merge the allocation histories under the same account. This implies that features built on this data will likely report all distributions as retroactively having gone to an identity which exists in the present, even if the distribution had originally gone to an identity which no longer exists because it has since been merged. However, since the original identity is preserved in the grainReceipt property and via the Allocations API, non-retroactive reporting could also be built if desired.

Through this change, features such as the new ExplorerHome page will be able to query information about historical grain payments for an account. This will provide us transparent reporting around who much people are getting paid over time.

# Automated Testing
`yarn test`
The updates included to existing tests should provide satisfactory coverage. `yarn coverage` reports ledger.js still has 100% coverage.

I updated the snapshot using the update_snapshots script so that CI passes.

# Manual Testing
I only did regression testing:
- I ran `yarn start` and ensured that grain account reporting had not changed on the front end.
- I ran `yarn grain` and `scdev grain` on the same ledger and ensured that the distributions were the same.
- I merged two identities with grain and ensured it succeeded and was reflected in the Grain Accounts page.


decentralion (commented):
Several comments:

1. This PR is quite hard to review, in part because the description is not very descriptive. A good PR description will give the reviewer an orientation on the problem being solved, the solution being proposed, and alternatives considered. Take a look at https://github.com/sourcecred/sourcecred/pull/2317 for an example. In this case, the description boils down to "I added a field to the ledger". How are we tracking it? Why does it have a `policyType` subfield? What are the semantics on account merge? The reviewer shouldn't need to guess at this by reading the code.

2. Storing the `timestampMs` and `policyType` on the distribution feels like a half-hearted solution to me, since it gives some of the info I might want to investigate, but not all of it (e.g. what was the budget for that allocation?). Instead of picking and choosing a few fields to include, we can leverage the fact that each `Distribution` and `GrainAllocation` has a unique id. Thus, we can just include a pointer to the relevant Distribution or Allocation. In principle, we could just have the account include an array of `AllocationIds` and then we could look up all info by inspecting the relevant allocation. But it probably does make sense to include the timestamp and amount explicitly for API convenience. But I think as far as looking up the policy type, we may as well just have the allocation id.

(Note: to realize this, we need to add new accessors to the ledger, for retrieving an allocation or distribution by id, and for looking up the containing distribution ID given a specific allocation ID. This is a good sign, since it means we'd be building an API out of flexible primitives rather than hardcoding for specific use cases.)

3. The `scripts/update_snapshots.sh` file has everything you need for updating snapshots. To run that script, you need our test discord API key... hit up me or @topocount for it. Although you don't actually need the key for the specific updates you want to do, since you don't need to re-run the `sourcecred load` command, just `sourcecred score` if I'm not mistaken. So you could just manually do that update (see the `update_snapshots.sh` script to see how).

blueridger:
@decentralion re: #2. Since the policy type was just a little extra sugar that we don't actually need right now, would you be alright with me just removing that field, and then we can add the allocation and distribution ids and lookups later when we need them?

blueridger:
I'll add the new allocation and distribution APIs in a separate PR and come back to this one.

src/core/ledger/ledger.js
```diff
@@ -33,7 +33,20 @@ import {
...
-import {type Allocation, type AllocationId} from "./grainAllocation";
+import {
+  type Allocation,
+  type AllocationId,
+  type GrainReceipt,
+} from "./grainAllocation";
+
+/**
+ * Timestamped record of a grain payment made to an Identity from a specific Allocation.
+ */
```
decentralion, line 44: Nit: wrap to 80 char

src/core/ledger/ledger.js
```diff
@@ -44,6 +57,8 @@ type MutableAccount = {|
   balance: G.Grain,
   // The amount of Grain this account has received in past Distributions
   paid: G.Grain,
+  // A history of Grain allocations to the account in chronological order
```
decentralion, line 60: Maybe worth mentioning here the merge logic, where an account will have allocations that it received directly or indirectly via merged accounts

src/core/ledger/ledger.js
```diff
@@ -530,8 +549,8 @@ export class Ledger {
     for (const allocation of distribution.allocations) {
       this._allocations.set(allocation.id, allocation);
       this._allocationsToDistributions.set(allocation.id, distribution.id);
-      for (const {id, amount} of allocation.receipts) {
-        this._allocateGrain(id, amount);
+      for (const receipt of allocation.receipts) {
+        this._allocateGrain(receipt, allocation.id, distribution.credTimestamp);
```
decentralion, line 553: The distribution's Cred timestamp is the timestamp of the Cred data that was used to distribute the Grain. However, it could be quite far removed from when the Grain was actually distributed. E.g. if we want to turn on SourceCred for a project and retroactively distribute grain for every week in the project's history, we could have a distribution that we make today, but with a cred timestamp from a year ago.

However, having the UI show the Grain as being received a year ago would be confusing-- in this example the Grain is distributed based on last year's cred, but it is distributed today. So I think we should be using the ledger timestamp for the ledger event containing this distribution instead.

Implementing this change will be kind of annoying because the data structure is that the `LedgerEvent` has a timestamp, and the `LedgerEvent` has an `Action`, and this code gets called through a `_processAction` handler that doesn't have the full ledger event. So I think that in order to implement this correctly, we'll need a pre-factor that makes it feasible for action handlers to get access to the full ledger event.

src/core/ledger/ledger.js
```diff
@@ -744,10 +763,15 @@ export class Ledger {
...
   // Helper method for recording that Grain was allocated to a identity.
   // Increases the identity's paid amount and balance in sync.
-  _allocateGrain(recipient: IdentityId, amount: G.Grain) {
-    const account = this._mutableAccount(recipient);
-    account.paid = G.add(amount, account.paid);
-    account.balance = G.add(amount, account.balance);
+  _allocateGrain(
+    grainReceipt: GrainReceipt,
+    allocationId: AllocationId,
+    timestampMs: TimestampMs
```
decentralion, line 769: nit / for readability, a weak preference that we refactor this to take an object with named arguments rather than positional args.

src/core/ledger/ledger.test.js
```diff
@@ -432,13 +437,31 @@ describe("core/ledger/ledger", () => {
...
+            grainReceipt: {id: id2, amount: g("10")},
+            timestampMs: 1,
+            allocationId: allocationId3,
+          },
+          {
+            grainReceipt: {id: id1, amount: g("1")},
+            timestampMs: 2,
+            allocationId: allocationId2,
+          },
+        ]);
```
decentralion, line 464: the test is titled `"gives the target's grain balance and paid to the base account"` and now its' actually testing extra properties about the allocation history and the sort order. this feels like potential confusion, e.g. what if someone re-writing this test doesn't realize that the timestamps were carefully chosen to require merge sorting and refactors in a way that stops testing that property. please write new test that is explicitly focused on the properties we're testing.

decentralion (commented):
Looks good overall, and I really appreciate the extended description for the PR -- made it a lot easier to orient on. However, there is one tricky issue, which is the use of the credTimestamp as the distribution timestamp. See discussion in comments.

src/core/ledger/ledger.js
decentralion: Idea: the `_processAction` handler could get the action as the first argument, and the event as an optional second argument. (technically this is redundant because the action is contained within the event, but it would be very convenient in the context of the typed arguments to the action handling functions)

src/core/ledger/ledger.js
```diff
@@ -33,7 +33,20 @@ import {
...
+  type GrainReceipt,
+} from "./grainAllocation";
+
+/**
+ * Timestamped record of a grain payment made to an Identity from a specific Allocation.
+ */
+type AllocationReceipt = {|
+  allocationId: AllocationId,
+  grainReceipt: GrainReceipt,
+  timestampMs: TimestampMs,
```
decentralion, line 48: Consider changing this from `timestampMs` to `credTimestampMs` so that later we can add `ledgerTimestampMs` without undue confusion.

decentralion (approved):
Switching from comment to "approve with nits" because in offline discussion, @blueridger and I agreed that in most cases, the cred timestamp is good enough to use as a timestamp for the grain.

src/core/ledger/ledger.js
blueridger: accepted

src/core/ledger/ledger.js
blueridger: accepted

src/core/ledger/ledger.js
blueridger: accepted

src/core/ledger/ledger.test.js
blueridger: accepted

src/core/ledger/ledger.js
blueridger: accepted

blueridger:
all nits accepted and changed. merging with existing approval.

PR #2535, opened 2020-11-21
Discord: Support "props channels"

decentralion:
On our Discord instance, we have a "props" channel where people are
encouraged to give props to other members of the community. Ideally, the
person getting props should receive a very large fraction of the Cred
flow from the props. However, with our current Cred semantics, the
person making the props splits Cred evenly with the person receiving
props.

This commit fixes that by adding a special, optional field to the
discord config where you can specify which channel(s) are props
channels. When a user is mentioned in one of those channels, they get a
"props" edge instead of a "mentions" edge. By default, the props edge
has a weight of 19x, so a message with a single props will flow 95% of
the Cred to the person getting props, and 5% to the person creating the
props as a small incentive for recognizing others.

In principle, we might not want to add this special case logic directly
into the Discord plugin; we might instead want to support adding
"micro plugins" to cover this kind of use case.

In practice, this doesn't add much complexity and is something we really
want, so I think hacking it in is justified.

Test plan: There isn't any automated testing covering the Discord graph
creation, but I manually tested this and am confident it works.
Specifically, I re-ran our own cred scores using this change and am
seeing reasonable diffs in Cred (see below). Also, inspecting the Cred
explorer shows that people who authored props messages are now getting
less Cred, and people who received props are now getting more Cred.

| Name | Before | After | Change % |
| ---- | ------ | ----- | -------- |
| decentralion | 24756 | 24102 | −3% |
| wchargin | 11403 | 11416 | 0% |
| beanow | 5692 | 5649 | −1% |
| s-ben | 5101 | 5015 | −2% |
| lbstrobbe | 4856 | 4841 | 0% |
| hammad | 3642 | 3702 | 2% |
| topocount | 1790 | 1799 | 1% |
| burrrata | 1735 | 1737 | 0% |
| KuraFire | 1649 | 1731 | 5% |
| bex | 1586 | 1616 | 2% |
| sourcecred | 1396 | 1396 | 0% |
| mzargham | 1261 | 1280 | 2% |
| sandpiper | 1113 | 1262 | 13% |
| dependabot | 1255 | 1257 | 0% |
| panchomiguel | 904 | 970 | 7% |
| vsoch | 910 | 913 | 0% |
| brianlitwin | 889 | 895 | 1% |
| evan | 642 | 703 | 10% |
| greenkeeper | 667 | 666 | 0% |
| joiecousins | 552 | 630 | 14% |
| yalor | 481 | 561 | 16% |
| Thena | 426 | 466 | 9% |
| youngkidwarrior | 463 | 437 | −6% |
| benoxmo | 443 | 432 | −2% |
| flyingzumwalt | 362 | 398 | 10% |
| amico | 259 | 273 | 5% |
| rpp | 239 | 251 | 5% |
| doctorrobinson | 290 | 233 | −19% |
| github-actions | 223 | 223 | 0% |
| owocki | 210 | 208 | −1% |
| deltafreq | 205 | 204 | −1% |
| ian | 129 | 199 | 54% |
| cortanav | 105 | 148 | 41% |
| protocol | 140 | 140 | 0% |
| Louisgrx | 139 | 139 | 0% |
| samkuhlmann | 132 | 132 | 0% |
| sriche | 129 | 128 | −1% |
| peth | 166 | 125 | −25% |
| coopahtroopa | 125 | 123 | −1% |
| JavierCanovas | 109 | 121 | 11% |

### DIFF
--- a/src/plugins/experimental-discord/config.js
+++ b/src/plugins/experimental-discord/config.js
@@ -58,6 +58,12 @@ export type DiscordConfigJson = {|
   // Note that channels use a snowflake id only.
   // defaultWeight is used to set weights for channels that don't have a specified weight
   +channelWeightConfig?: ChannelWeightConfig,
+  // List of channels which are considered "props channels".
+  // In a props channel, we have an extra rule: if someone is mentioned in a message,
+  // we create a "props" edge to the mentioned user instead of a regular "mentions" edge.
+  // We can set a higher weight on these props edges, which allows us to flow Cred in a props
+  // mostly to the people getting propsed, rather than to the author of the props message.
+  +propsChannels?: $ReadOnlyArray<Model.Snowflake>,
 |};
 
 const parserJson: C.Parser<DiscordConfigJson> = C.object(
@@ -74,12 +80,14 @@ const parserJson: C.Parser<DiscordConfigJson> = C.object(
       defaultWeight: C.number,
       weights: C.dict(C.number),
     }),
+    propsChannels: C.array(C.string),
   }
 );
 
 export type DiscordConfig = {|
   +guildId: Model.Snowflake,
   +weights: WeightConfig,
+  +propsChannels: $ReadOnlyArray<Model.Snowflake>,
 |};
 
 /**
@@ -106,6 +114,7 @@ export function _upgrade(json: DiscordConfigJson): DiscordConfig {
         defaultWeight: 1,
       },
     },
+    propsChannels: json.propsChannels || [],
   };
 }
 
--- a/src/plugins/experimental-discord/config.js
+++ b/src/plugins/experimental-discord/config.js
@@ -62,7 +62,7 @@ export type DiscordConfigJson = {|
   // In a props channel, we have an extra rule: if someone is mentioned in a message,
   // we create a "props" edge to the mentioned user instead of a regular "mentions" edge.
   // We can set a higher weight on these props edges, which allows us to flow Cred in a props
-  // mostly to the people getting propsed, rather than to the author of the props message.
+  // mostly to the people receiving props, rather than to the author of the props message.
   +propsChannels?: $ReadOnlyArray<Model.Snowflake>,
 |};
 
--- a/src/plugins/experimental-discord/createGraph.js
+++ b/src/plugins/experimental-discord/createGraph.js
@@ -21,10 +21,12 @@ import {
   addsReactionEdgeType,
   reactsToEdgeType,
   mentionsEdgeType,
+  propsEdgeType,
 } from "./declaration";
 import * as Model from "./models";
 
-import {type WeightConfig, reactionWeight} from "./reactionWeights";
+import {type DiscordConfig} from "./config";
+import {reactionWeight} from "./reactionWeights";
 
 // Display this many characters in description.
 const MESSAGE_LENGTH = 30;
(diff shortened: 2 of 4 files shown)


src/plugins/experimental-discord/createGraph.js
```diff
@@ -202,6 +221,9 @@ export type GraphMessage = {|
   +author: Model.GuildMember | null,
   +reactions: $ReadOnlyArray<GraphReaction>,
   +mentions: $ReadOnlyArray<Model.GuildMember>,
+  // Included so we can apply any channel-based rules (e.g. creating props
+  // edges intsead of mentions edges) at graph construction time.
```
topocount, line 225: nit: spelling
```suggestion
  // edges instead of mentions edges) at graph construction time.
```

src/plugins/experimental-discord/config.js
```diff
@@ -58,6 +58,12 @@ export type DiscordConfigJson = {|
   // Note that channels use a snowflake id only.
   // defaultWeight is used to set weights for channels that don't have a specified weight
   +channelWeightConfig?: ChannelWeightConfig,
+  // List of channels which are considered "props channels".
+  // In a props channel, we have an extra rule: if someone is mentioned in a message,
+  // we create a "props" edge to the mentioned user instead of a regular "mentions" edge.
+  // We can set a higher weight on these props edges, which allows us to flow Cred in a props
+  // mostly to the people getting propsed, rather than to the author of the props message.
```
topocount, line 65: nit: picky grammar. This is a subjective suggestion, left to your discretion.
```suggestion
  // mostly to the people receiving props, rather than to the author of the props message.
```

topocount (approved):
I agree that the need for this configuration outweighs concerns about this specific implementation. 

I'm curious what a "micro-plugin" system for discord might look like, rather than the ability to simply configure the mentions weights. I think I see it: the role, channel, and hypothetical mentions weight configs can be utilized as lower-level APIs for these mini-plugins to provide specific types of channels that are easier for users to understand and implement for themselves (as opposed to having to understand the concept of a weight before they do anything else)

hammadj (commented):
Seems legit enough, agree that this hacky fix is worthwhile until we get the proper micro-plugins setup.

However, the actual changes in cred don't seem to align with my expectations. LB didnt have any change in cred, and Bex had the same increase in cred as me, which doesn't seem right since I'm pretty confident that they received a lot more props than me. Wondering if something weird is happening where people are losing cred by giving props or something?

One explanation I can think of is that LB was equally dishing out props as they were receiving props, and that those actions cancelled out. If that's the case, then it's not behaving as intended because it shouldn't be a zero-sum thing.

hammadj (commented):
Seems legit enough, agree that this hacky fix is worthwhile until we get the proper micro-plugins setup.

However, the actual changes in cred don't seem to align with my expectations. LB didnt have any change in cred, and Bex had the same increase in cred as me, which doesn't seem right since I'm pretty confident that they received a lot more props than me. Wondering if something weird is happening where people are losing cred by giving props or something?

One explanation I can think of is that LB was equally dishing out props as they were receiving props, and that those actions cancelled out. If that's the case, then it's not behaving as intended because it shouldn't be a zero-sum thing.

blueridger (approved):

blueridger:
> Seems legit enough, agree that this hacky fix is worthwhile until we get the proper micro-plugins setup.
> 
> However, the actual changes in cred don't seem to align with my expectations. LB didnt have any change in cred, and Bex had the same increase in cred as me, which doesn't seem right since I'm pretty confident that they received a lot more props than me. Wondering if something weird is happening where people are losing cred by giving props or something?
> 
> One explanation I can think of is that LB was equally dishing out props as they were receiving props, and that those actions cancelled out. If that's the case, then it's not behaving as intended because it shouldn't be a zero-sum thing.

I think people giving props getting a decrease in cred is exactly why LB's and Bex's scores are lower than you'd expect. Both LB and Bex are frequent propsers. There is a zero-sum thing happening with this change, because the amount of minted cred is not changing, it is just being redistributed.

src/plugins/experimental-discord/declaration.js
```diff
@@ -75,6 +75,14 @@ export const mentionsEdgeType: EdgeType = deepFreeze({
   description: "Connects a message to the member being mentioned.",
 });
 
+export const propsEdgeType: EdgeType = deepFreeze({
+  forwardName: "gives props to",
+  backwardName: "recevies props from",
+  prefix: EdgeAddress.append(edgePrefix, "PROPS"),
+  defaultWeight: {forwards: 19, backwards: 1 / 16},
```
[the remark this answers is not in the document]
decentralion: If props flow with 19x weight, and authorship flows with 1x weight, then a message that props a single person will flow 95% of Cred to the props-ee, 5% to the author.

PR #2545, opened 2020-11-22
grain cli: add simulation flag

elihanover:
This adds an optional `--sim` flag which allows users to run a simulation of the upcoming grain distribution without writing to the ledger or accounts.  This will be more useful as we allow more advanced simulation configurations, and should not be considered final.

Ultimately, we will want to create simulated ledger and accounts data which can be analyzed more easily than copying and pasting a markdown table.  This also could serve as the foundation for simulating multiple distributions that a community could vote on.

__Test Plan:__
Trivial change and sanity checked against the similar code in `src/cli/go.js`.  Tested myself using `yarn grain` and `yarn grain --sim`, checking my local cred instance's ledger for changes.

### DIFF
--- a/src/cli/grain.js
+++ b/src/cli/grain.js
@@ -21,8 +21,11 @@ function die(std, message) {
 }
 
 const grainCommand: Command = async (args, std) => {
-  if (args.length !== 0) {
-    return die(std, "usage: sourcecred grain");
+  let simulation = false;
+  if (args.length === 1 && (args[0] === "--simulation" || args[0] === "-s")) {
+    simulation = true;
+  } else if (args.length !== 0) {
+    return die(std, "usage: sourcecred grain [--simulation]");
   }
 
   const baseDir = process.cwd();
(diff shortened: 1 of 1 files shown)


src/cli/grain.js
```diff
@@ -21,7 +21,10 @@ function die(std, message) {
 }
 
 const grainCommand: Command = async (args, std) => {
-  if (args.length !== 0) {
+  let simulation = false;
+  if (args.length === 1 && args[0] === "--sim") {
```
decentralion, line 25: Recommend making it more descriptive with an optional abbreviation, e.g. let the flag be `--simulate` or `-s` for short.

decentralion (commented):
- the help text should get updated to mention this flag
- when the simulate flag is provided, the output message should make it clear the result is a simulation, vs the current one which says "Distributed ...", which can be misleading since no distribution occured.

topocount (commented):
@elihanover It'd be awesome to lay out the larger effort in more detail, listing each potential PR, so it's clear where this one fits into the larger feature.

src/cli/grain.js
```diff
@@ -59,28 +62,34 @@ const grainCommand: Command = async (args, std) => {
...
   std.out(
     dedent`\
-      usage: sourcecred grain
+      usage: sourcecred grain [--simulation || -s]
 
       Distribute Grain (or whatever currency this Cred instance is tracking)
       for Cred intervals in which Grain was not already distributed.
 
+      When the '--simulation' (-s) flag is provided, the instance's ledger and accounts
+      data will not be written to, allowing for testing the output of various configurations.
```
decentralion, line 91: suggest saying "no Grain will actually be distributed" rather than "the instance's ledger and accounts data will not be written to". It's clearer to explain the behavior rather than explaining the implementation. :)

decentralion (approved):
LGTM. In a separate PR, I'd like to see a flag that forces it to do exactly one distribution regardless of many it's supposed to create. This would allow you to simulate a distribution even if there isn't one due until the next interval, or to make a special distribution out-of-order (e.g. for a bonus).

PR #2546, opened 2020-11-23
immediate policy: optionally include past N weeks

elihanover:
Right now, the immediate policy uses cred from the most recent interval.  This generalizes
it such that users can create policies based on the past N weeks, without the discounting
of the Recent policy.

This includes an optional `numPeriodsLookback` parameter in the `ImmediatePolicy` type, which
defaults to 1 in config until #2600 gets merged, at which point we'll allow users to specify
alternative lookback lengths.

Note, this raises a question around naming, as Recent is actually a better name here, and we
could use something for the current Recent policies that highlights the discounting more.

__Test Plan__
Unit tests provided check that the `immediatePolicy` works for valid params, and also checks
for invalid `credIntervalCount` parameters.

This also updates the snapshot in `core/ledger/ledger.test` to include this extra paramter
on the `ImmediatePolicy`.


src/core/ledger/policies/immediate.js
```diff
@@ -22,16 +22,19 @@ export type ImmediatePolicy = {|
...
  * Split a grain budget in proportion to the cred scores in
- * the most recent time interval.
+ * the most recent time interval, with the option to extend the interval
+ * to include the last {history} weeks.
  */
 export function immediateReceipts(
   budget: G.Grain,
-  identities: ProcessedIdentities
+  identities: ProcessedIdentities,
+  history: number = 1
```
decentralion, line 31: confusing parameter name.

decentralion (commented):
Lacks testing, both for the basic case, and for edge cases (what if history is 0, or is longer than the actual history)?

out of curiosity, what's motivating this change?

src/core/ledger/policies/immediate.js
elihanover: Writing tests now.  For parameter name, how does `credIntervalCount` feel?  It's a bit wordy, but avoids the confusion around "credIntervals" which is used to describe an `IntervalSequence`.

Regarding the motivation, I thought this was a natural and useful generalization such that users could add a non-discounted "recent" policy (let's work out naming clashes).  For instance, I might want to reward contributors for the past 3 months, but not use the progressive discounting within that range.

src/core/ledger/policies/immediate.js
```diff
@@ -22,16 +22,33 @@ export type ImmediatePolicy = {|
...
  * Split a grain budget in proportion to the cred scores in
- * the most recent time interval.
+ * the most recent time interval, with the option to extend the interval
+ * to include the last {credIntervalCount} weeks.
  */
 export function immediateReceipts(
   budget: G.Grain,
-  identities: ProcessedIdentities
+  identities: ProcessedIdentities,
+  credIntervalCount: number = 1
```
decentralion, line 31: Why isn't the `credIntervalCount` part of the policy? Seems like good information to include there e.g. for future audit-ability. In whihc case this function can just take the `ImmediatePolicy` object, in addition to the identities.

decentralion, line 31: For names, I think `numPeriodsLookback` (or similar) could work as well. I don't object to `credIntervalCount`

src/core/ledger/policies/immediate.js
```diff
@@ -22,16 +22,33 @@ export type ImmediatePolicy = {|
...
 ): $ReadOnlyArray<GrainReceipt> {
-  const amounts = G.splitBudget(
-    budget,
-    identities.map((i) => i.mostRecentCred)
+  const totalIntervals = identities[0].cred.length;
+  if (credIntervalCount < 1 || credIntervalCount > totalIntervals) {
+    throw new Error(
+      `credIntervalCount must be in range [1, totalIntervals], got ${credIntervalCount}`
```
decentralion, line 36: I think failing when `intervalCount > totalIntervals` is annoying, b.c. it means a config that is valid for one project might be invalid for another. E.g. I decide to fork off of MetaGame's config, but they have a policy that distributes for 12 weeks lookback, but my project is only 4 weeks old. I recommend soft capping the interval count instead.

src/core/ledger/policies/immediate.js
elihanover: Used `numPeriodsLookback`.  It might be confusing, since we are now using intervals and periods interchangeably, however, I don't have any better ideas atm so let's roll with that.

src/core/ledger/policies/immediate.js
elihanover: Good point.

src/api/grainConfig.js
```diff
@@ -11,6 +11,7 @@ export type GrainConfig = {|
   +balancedPerWeek?: number,
   +recentPerWeek?: number,
   +recentWeeklyDecayRate?: number,
+  +numPeriodsLookback?: number,
```
decentralion, line 14: if we add it in the config, it should probably say `immediateNumPeriodsLookback` so that it's clear which policy it applies to.

That said, we should really come up with a backwards compatible way to add arbitrary policies, rather than continuing to add extra fields here. Could we do that instead of adding another field? I'm thinking there would be a field like `policies` which is an optional array of explicit allocation policies, and the hardcoded per-policy fields just implicitly add extra policies. But then we could migrate everyone to just use the policies field. (This would be a separate PR.)

src/core/ledger/policies/immediate.test.js
```diff
@@ -0,0 +1,132 @@
...
+      ];
+      const policy = {
+        policyType: "IMMEDIATE",
+        budget: G.ONE,
+        numPeriodsLookback: 4,
+      };
+      expect(immediateReceipts(policy, processedIdentities)).toEqual(expected);
+    });
+
+    it("returns a valid $ReadOnlyArray of GrainReceipt's at min interval length", () => {
```
decentralion, line 99: Technically speaking, $ReadOnlyArray is a type-level designation, not a runtime designation. So at runtime, there is no read-only array, just an actual array. 

Also, does "min interval length" mean "1 period lookback"? What does "valid" mean in this case, exactly? Do we just mean it "correctly computes receipts when numPeriodsLookback is 1"?

(same questions about the test case above)

decentralion (commented):
Looks pretty good! Nits mostly, and that I'd rather not add more one-off fields to the grain config. I'd be down with merging this one basically as-is if we remove the change to the grainConfig api, then we could follow up with work to allow arbitrary policies there.

src/api/grainConfig.js
elihanover: I agree.  I can open #2513 back up.  Should have made a clearer case/vision when I opened that one.  That also lets us take multiple of the same policy type, which is nice.

src/core/ledger/policies/immediate.test.js
elihanover: Done.

src/api/grainConfig.js
elihanover: Done.

src/api/grainConfig.js
```diff
@@ -11,6 +11,7 @@ export type GrainConfig = {|
   +balancedPerWeek?: number,
   +recentPerWeek?: number,
   +recentWeeklyDecayRate?: number,
+  +immediateNumPeriodsLookback?: number,
```
decentralion, line 14: I don't want any more fields added here because we would deprecate it immediately (per #2600). Please remove any changes to this file from this PR.

src/core/ledger/policies/immediate.js
```diff
@@ -18,20 +19,38 @@ export type Immediate = "IMMEDIATE";
 export type ImmediatePolicy = {|
   +policyType: Immediate,
   +budget: G.Grain,
+  +numPeriodsLookback?: number,
```
decentralion, line 22: Policies should only have required fields; optionality gets delegated to the config. See https://github.com/sourcecred/sourcecred/pull/2600#issuecomment-755839137

src/api/grainConfig.js
elihanover: Done.  This default to 1 for now.  We can add customization in the config after #2600 as suggested.

src/core/ledger/policies/immediate.js
elihanover: Done.

src/api/grainConfig.test.js
```diff
@@ -94,6 +95,38 @@ describe("api/grainConfig", () => {
       expect(() => toDistributionPolicy(x)).not.toThrow();
     });
 
+    it("default immediate's numPeriodsLookback to 1", () => {
```
decentralion, line 98: nit: this test would be cleaner to set everything except `immediatePerWeek` to 0, so there'd be less noise in the `expectedDistributionPolicy`.

(no change required)

decentralion (approved):

src/api/grainConfig.test.js
elihanover: Done, this makes a lot of sense.

topocount:
@elihanover I've been burned by this kind of issue in other projects, and compulsively rebase my topic branches now to avoid it.

PR #2550, opened 2020-11-24
Ledger: add APIs for getting allocation and distribution by ID

blueridger:
# Description
This adds APIs for looking up an allocation or a distribution by ID by adding lookup maps as properties on the Ledger and filing those maps during the parsing of the DISTRIBUTE_GRAIN actions in the ledger log. Access to a distribution lets us compare all related allocations and see the timestamp. Access to an allocation gives us information about policy type and lets us compare all related receipts (identity-level payouts).

This is a co-requisite for https://github.com/sourcecred/sourcecred/pull/2534 to surface the allocationId and distributionId, which will allow future development of features that provide transparent reporting on how grain is being distributed, such as the new Explorer Home page.

# Test Plan
`yarn test` passes, unit tests added for full coverage

### DIFF
--- a/src/core/ledger/ledger.js
+++ b/src/core/ledger/ledger.js
@@ -25,10 +25,15 @@ import {type NodeAddressT, NodeAddress} from "../graph";
 import {type TimestampMs} from "../../util/timestamp";
 import * as NullUtil from "../../util/null";
 import * as uuid from "../../util/uuid";
-import {type Distribution, parser as distributionParser} from "./distribution";
+import {
+  type Distribution,
+  type DistributionId,
+  parser as distributionParser,
+} from "./distribution";
 import * as G from "./grain";
 import {JsonLog} from "../../util/jsonLog";
 import * as C from "../../util/combo";
+import {type Allocation, type AllocationId} from "./grainAllocation";
 
 /**
  * Every Identity in the ledger has an Account.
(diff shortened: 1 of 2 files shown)


src/core/ledger/ledger.js
```diff
@@ -159,6 +165,47 @@ export class Ledger {
...
+  distributionByAllocationId(allocationId: UuidT): Distribution {
+    for (const distribution of this._distributions.values()) {
+      if (
+        distribution.allocations.find(
+          (allocation) => allocation.id === allocationId
+        )
+      )
+        return distribution;
+    }
+    throw new Error(`no Distribution for allocation id: ${allocationId}`);
```
decentralion, line 206: I think we should invest upfront in making this performant, by adding a `_allocationIdToDistribution: Map<AllocationId, DistributionId>` field to the ledger.

decentralion (commented):
Looks pretty good. In addition to the performance improvement suggested, we should add an API method which returns an iterator over all the allocations and another which returns an iterator over the distributions. We should iterate over them in logical order (i.e. same order as in the event log)

src/core/ledger/ledger.test.js
```diff
@@ -847,21 +879,118 @@ describe("core/ledger/ledger", () => {
...
+            paid: g("17"),
+            active: true,
+          };
+          expect(ledger.accounts()).toEqual([ac1, ac2]);
+        });
+
+        it("should record the allocations and distributions", () => {
+          expect(ledger.allocation(allocation1.id)).toEqual(allocation1);
+          expect(ledger.allocation(allocation2.id)).toEqual(allocation2);
+          expect(ledger.distribution(distribution1.id)).toEqual(distribution1);
```
decentralion, line 976: nit: may as well check for distribution 2.

decentralion (approved):
looks good. per `yarn coverage`, ledger continues to have 100% test coverage after this commit.

src/core/ledger/ledger.test.js
blueridger: adding and merging

PR #2552, opened 2020-11-25
Refactor Credrank Testing & Test CredGraph

decentralion:
This commit sets up testing for the CredGraph module, which was
previously entirely untested. Getting an example CredGraph is a bit of a
pain, because it requires constructing a MarkovProcessGraph, which is
itself fairly involved.

We already had logic for constructing an example MarkovProcessGraph, for
MPG's own testing, and I didn't want to duplicate it. Instead, I
factored out all the example logic from `markovProcessGraph.test.js`
into `credrank/testUtils.js`. This is similar to the pattern we have for
testing regular Graphs, see `core/graphTestUtil.js`.

I then set up some basic sanity-check testing in `credGraph.test.js`,
mostly to show that it's now possible.

Test plan: This is a test-only change. `yarn test` passes.

### DIFF
--- a/src/core/credrank/testUtils.js
+++ b/src/core/credrank/testUtils.js
@@ -1,0 +1,157 @@
+// @flow
+
+import deepFreeze from "deep-freeze";
+import {
+  NodeAddress as NA,
+  EdgeAddress as EA,
+  type Node as GraphNode,
+  type Edge as GraphEdge,
+  Graph,
+} from "../graph";
+import {type Weights as WeightsT} from "../weights";
+import {type WeightedGraph} from "../weightedGraph";
+import {
+  type NodeWeightEvaluator,
+  type EdgeWeightEvaluator,
+  nodeWeightEvaluator,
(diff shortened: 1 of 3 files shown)


src/core/credrank/testUtils.js
```diff
@@ -0,0 +1,157 @@
...
+
+// Connects contribution 1 to participant 1
+export const e1: GraphEdge = {
+  address: ea("e1"),
+  src: c1.address,
+  dst: participantNode1.address,
+  timestampMs: 3,
+};
+
+// Connects c0 to c1.
```
blueridger, line 93: nit: If you're writing the comments anyways, doc comments instead of // comments enable sweet IDE visibility that improves developer speed, even (and maybe especially) on silly test utils like this.

blueridger (approved):

src/core/credrank/testUtils.js
decentralion: Ack, will do in future.

PR #2554, opened 2020-11-25
Add more detail in LedgerViewer for ADD_ALIAS and TRANSFER_GRAIN

BrianBelhumeur:
# Description
Currently, the ledger viewer lists the ADD ALIAS action, but does not provide any additional information. To make the viewer more useful, I've added the details of the add alias action. The description string in the ledger comes in the format:
`channel/[@handle](url)`
For the purposes of the UI, I've cut out the `(url)` part of the description because it's noisy and adds little (if any) UI value. The final value of the details line ends up as:
`account ⇐ channel/@handle (user / bot / etc)`

Also, for the TRANSFER GRAIN action, there is no detail. I have added detail in the format:
`sender (user / bot / etc) ⇒ amount ⇒ recipient (user / bot / etc)`

Examples:
Screen Shot 2020-11-27 at 23 40 15

# Test Plan
Tested manually. Much of the code is copied from other parts of the LedgerViewer.js file, so it's already well-tested.


KuraFire:
Requesting a syntax change to the ALIAS line output: `account ⇐ channel/@handle`

It would make it look something like this:
```
ADD ALIAS           KuraFire ⇐ discord/@Faruk
ADD ALIAS           KuraFire ⇐ discourse/@kurafire
ADD ALIAS           KuraFire ⇐ github/@kurafire
CREATE IDENTITY     KuraFire
```

The type (user, bot, project) is something we may want to express in a different way, or at least much more subtly than the big pill button-style labels. Or, we could invert the coloring of those labels, and make it an embossed imprint rather than something standing out as an attention grabber. How easy is it to tweak its colors, @BrianBelhumeur ?

src/ui/components/LedgerViewer.js
```diff
@@ -222,13 +222,22 @@ const LedgerEventRow = React.memo(
         case "ADD_ALIAS":
           try {
             const account = ledger.account(action.identityId);
-            // description has format: channel/[@handle](url). url has little value in the UI, so cut it
-            const aliasHandle = action.alias.description.replace(/\(.*\)$/, "");
+            let alias = action.alias.description;
+            // description has format: channel/[@handle](url). Parse out alias, account, url
+            const descriptionParts = alias.match(
+              /^([^/]*)\/\[(@[^\]]*)\]\(([^)]*)\)$/
```
KuraFire, line 228: What a regex we need for this 😂 

(In my head I thought of just stripping `[` and `]` from the output since we don't allow them as characters in identity, but this is obviously much more robust)

KuraFire (approved):
Dope, thanks @BrianBelhumeur !

BrianBelhumeur:
> Requesting a syntax change to the ALIAS line output: `account ⇐ channel/@handle`
> 
> It would make it look something like this:
> 
> ```
> ADD ALIAS           KuraFire ⇐ discord/@Faruk
> ADD ALIAS           KuraFire ⇐ discourse/@kurafire
> ADD ALIAS           KuraFire ⇐ github/@kurafire
> CREATE IDENTITY     KuraFire
> ```
> 
> The type (user, bot, project) is something we may want to express in a different way, or at least much more subtly than the big pill button-style labels. Or, we could invert the coloring of those labels, and make it an embossed imprint rather than something standing out as an attention grabber. How easy is it to tweak its colors, @BrianBelhumeur ?

Tweaking colors is pretty easy. What color(s) do you have in mind? We can do solid bubble (as it is now) or outline, and whatever colors for text / background / outline.

KuraFire:
@BrianBelhumeur Ideally, I would want a new CSS class to be added just for the user types for now. This isn't really MUI convention afaik, but it could be something like a class called `.MuiChip-tiny` that would produce overrides like this:
```
.MuiChip-root.MuiChip-tiny {
    color: #828282;
    background-color: #383838;
    text-shadow: 0 1px 0 rgba(0,0,0, 0.4);
}
.MuiChip-tiny .MuiChip-labelSmall {
    font-size: 0.55rem;
}
```

(I don't want it to affect the Grain distribution buttons)

KuraFire:
lgtm 👍🏼

src/ui/components/LedgerViewer.js
```diff
@@ -48,6 +48,12 @@ const useStyles = makeStyles((theme) => {
       paddingLeft: theme.spacing(2),
       paddingRight: theme.spacing(1),
     },
+    chip: {
+      fontSize: "0.55rem",
+      color: "#828282",
+      backgroundColor: "#383838",
```
hammadj, line 54: Should use values from the theme instead of hardcoded ones

src/ui/components/LedgerViewer.js
```diff
@@ -216,9 +223,47 @@ const LedgerEventRow = React.memo(
...
+            );
+          } catch (e) {
+            console.warn("Unable to find account for action: ", action);
+            return (
+              <IdentityDetails
+                id={action.identityId}
+                name="[Unknown Account]"
+              />
+            );
+          }
```
hammadj, line 266: Should extract this logic into its own component since there's a decent amount of logic here.

src/ui/components/LedgerViewer.js
```diff
@@ -254,6 +299,69 @@ const LedgerEventRow = React.memo(
...
+
+          return (
+            <>
+              {sender}
+              &nbsp; &rArr; &nbsp;
+              {amount}
+              &nbsp; &rArr; &nbsp;
+              {recipient}
+            </>
+          );
```
hammadj, line 364: Should extract this logic into its own component since there's a decent amount of logic here.

src/ui/components/LedgerViewer.js
BrianBelhumeur: These are one-off colors. Do they still belong in the theme? If so, where do they go?

KuraFire (approved):
lgtm 👍🏼

src/ui/components/LedgerViewer.js
BrianBelhumeur: Done

PR #2563, opened 2020-11-26
CLI: Add -d diff flag to credrank command

blueridger:
# Description
Adds a `-d` flag to the `credrank` command that outputs a table showing the diff between the existing credgraph as read from json and the new credgraph. This will allow developers to easily see if/how their code changes affect credrank scores.

I chose to use `console.table` instead of the existing line-by-line method because the table is the primary product of the command when the flag is included, and readability (specifically, scanning the % Change column) is most important. Although this makes the table wider, it still seems like it should fit fine for 80-char terminal users in most cases.

# Test Plan
Not tested by automated tests.

To create cred diffs, I added `+1` to the end of line 107 or 104 in `core/credrank/compute.js`

Used `scdev credrank -d`

### Negative change:
Screen Shot 2020-11-26 at 11 13 13 AM

### Positive change:
Screen Shot 2020-11-26 at 11 14 24 AM

### When there is no existing json
Screen Shot 2020-11-26 at 11 26 34 AM

### With a few lines of context:
Screen Shot 2020-11-26 at 11 00 45 AM

### 80-char terminal with long cred scores:
Screen Shot 2020-11-26 at 11 43 03 AM

### Also copies and pastes well:

│    0    │       'hammadj'       │   '55.4'   │  '73.0'  │  '+31.8%'  │
│    1    │       'Beanow'        │   '48.7'   │  '66.3'  │  '+36.2%'  │
│    2    │ 'decentralion-github' │   '23.7'   │  '41.3'  │  '+74.3%'  │
│    3    │        'Thena'        │   '13.8'   │  '31.4'  │ '+127.8%'  │
│    4    │  'topocount-github'   │   '12.2'   │  '29.8'  │ '+144.8%'  │
```
│    0    │       'hammadj'       │   '55.4'   │  '73.0'  │  '+31.8%'  │
│    1    │       'Beanow'        │   '48.7'   │  '66.3'  │  '+36.2%'  │
│    2    │ 'decentralion-github' │   '23.7'   │  '41.3'  │  '+74.3%'  │
│    3    │        'Thena'        │   '13.8'   │  '31.4'  │ '+127.8%'  │
│    4    │  'topocount-github'   │   '12.2'   │  '29.8'  │ '+144.8%'  │
```

### DIFF
--- a/src/cli/credrank.js
+++ b/src/cli/credrank.js
@@ -4,15 +4,17 @@ import fs from "fs-extra";
 import stringify from "json-stable-stringify";
 import {join as pathJoin} from "path";
 import {sum} from "d3-array";
+import {format} from "d3-format";
 
 import sortBy from "../util/sortBy";
 import {credrank} from "../core/credrank/compute";
-import {CredGraph} from "../core/credrank/credGraph";
+import {CredGraph, type Participant} from "../core/credrank/credGraph";
 import {LoggingTaskReporter} from "../util/taskReporter";
 import {computeBonusMinting, createBonusGraph} from "../core/bonusMinting";
 import type {Command} from "./command";
-import {loadInstanceConfig, prepareCredData} from "./common";
+import {loadInstanceConfig, prepareCredData, loadCredGraph} from "./common";
 import {merge} from "../core/weightedGraph";
+import {type Uuid} from "../util/uuid";
 
 function die(std, message) {
   std.err("fatal: " + message);
@@ -20,8 +22,19 @@ function die(std, message) {
 }
 
 const credrankCommand: Command = async (args, std) => {
-  if (args.length !== 0) {
-    return die(std, "usage: sourcecred credrank");
+  let shouldIncludeDiff = false;
+  const processedArgs = args.filter((arg) => {
+    switch (arg) {
+      case "-d":
+        shouldIncludeDiff = true;
+        return false;
+      default:
+        return true;
+    }
+  });
+
+  if (processedArgs.length !== 0) {
+    return die(std, "usage: sourcecred credrank [-d]");
   }
   const taskReporter = new LoggingTaskReporter();
   taskReporter.start("credrank");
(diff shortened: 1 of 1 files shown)


src/cli/credrank.js
```diff
@@ -73,4 +97,38 @@ function printCredSummaryTable(credGraph: CredGraph) {
...
+  const priorParticipants: Map<Uuid, Participant> = new Map();
+  for (const participant of priorCredGraph.participants())
+    priorParticipants.set(participant.id, participant);
+
+  const credParticipants = Array.from(credGraph.participants());
+  const sortedParticipants = sortBy(credParticipants, (p) => -p.cred);
+  function row({cred, description, id}) {
+    const prior = priorParticipants.get(id);
+    if (!prior)
+      throw `Participant [${description}, ${id}] exists in the new scores but not in the old. Rerun without -d to refresh.`;
```
decentralion, line 111: I note that if there's a new participant, it will error, but if a participant disappeared, it will silently elide their removal. Seems inconsistent and like it may result in confusion.

src/cli/credrank.js
```diff
@@ -73,4 +97,38 @@ function printCredSummaryTable(credGraph: CredGraph) {
...
+    const percentageChange = (100 * (cred - prior.cred)) / prior.cred;
+    const percentageChangeStr =
+      percentageChange > 10000
+        ? ">10,000%"
+        : `${percentageChange > 0 ? "+" : ""}${percentageChange.toFixed(1)}%`;
```
decentralion, line 116: I recommend using d3.format rather than hand rolling. (We already have this dep.)
https://github.com/d3/d3-format

src/cli/credrank.js
```diff
@@ -73,4 +97,38 @@ function printCredSummaryTable(credGraph: CredGraph) {
...
+      "% Change": percentageChangeStr,
+    };
+  }
+  type Row = {
+    "Name": string,
+    "Prior Cred": string,
+    "New Cred": string,
+    "% Change": string,
+  };
+  const rows: Array<Row> = sortedParticipants.slice(0, 20).map((n) => row(n));
```
decentralion, line 130: nit: this type seems a little unnecessary since it's basically just a re-declaration of the return type which is just one line above, and never leaves the scope of this function. no change requested though. also technically $ReadOnlyArray would be more accurate than Array.

decentralion (approved):
Currently it shows the change in absolute Cred but not change in relative Cred. often times though, the relative Cred score are what matters most. might consider showing change in relative Cred as well.

that said, this seems fine and useful as is, so approved. feel free to merge as is, or to make modifications and then merge on this approval.

src/cli/credrank.js
blueridger: flow was being weird about console.table's expectations without typing `rows` like this. Playing with it now, turns out I can put the slice/map inline in the console.table without complaint, which is much nicer.

src/cli/credrank.js
blueridger: accepted

src/cli/credrank.js
blueridger: added a check for participant count equality

PR #2570, opened 2020-12-03
cli/common: loadCurrencyDetails

elihanover:
### Description
Loads custom currency config, falling back on defaults if necessary.

__Related__
- Used by #2571
- Uses #2572 

### Test Plan
Unit test to check fallback to defaults on bad path to currency config.  Manually tested on 1Hive's instance to check success case with custom configs.

### DIFF
--- a/src/cli/common.js
+++ b/src/cli/common.js
@@ -29,6 +29,11 @@ import {CredGraph, parser as credGraphParser} from "../core/credrank/credGraph";
 import {loadFileWithDefault, loadJsonWithDefault} from "../util/disk";
 import {parser as pluginBudgetParser} from "../api/pluginBudgetConfig";
 import {applyBudget, type Budget} from "../core/mintBudget";
+import {
+  parser as currencyConfigParser,
+  type CurrencyDetails,
+} from "../api/currencyConfig";
+import {defaultCurrencyConfig} from "../api/currencyConfig";
 
 import * as Weights from "../core/weights";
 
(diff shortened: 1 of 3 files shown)


src/cli/common.js
```diff
@@ -197,3 +202,22 @@ function loadPluginBudget(baseDir: string): Promise<Budget | null> {
...
+export async function loadCurrencyDetails(
+  currencyDetailsPath: string
+): Promise<CurrencyDetails> {
+  const {name, suffix} = await loadJsonWithDefault(
+    currencyDetailsPath,
+    currencyConfigParser,
+    () => ({
+      name: DEFAULT_NAME,
+      suffix: DEFAULT_SUFFIX,
+    })
```
decentralion, line 219: Please export a  `defaultCurrencyConfig()` function from `currencyConfig.js`. That way it's possible to add a new field and update the default all in one place.

elihanover: Done, thanks for suggestion, much cleaner.

src/cli/common.js
```diff
@@ -197,3 +202,19 @@ function loadPluginBudget(baseDir: string): Promise<Budget | null> {
...
+export async function loadCurrencyDetails(
+  currencyDetailsPath: string
+): Promise<CurrencyDetails> {
+  const {name, suffix} = await loadJsonWithDefault(
+    currencyDetailsPath,
+    currencyConfigParser,
+    defaultCurrencyConfig
+  );
+
+  return {name, suffix};
```
decentralion, line 219: The unnecessary destructuring into `{name, suffix}` makes this code brittle--now someone adding a new field would need to make a boilerplate change to this function too. Just do `return await loadJsonWithDefault(...)`

elihanover: Agreed, done.

decentralion (approved):

PR #2577, opened 2020-12-05
Remove Recharts Dependency

mkusold:
# Description
I went to work on the mobile responsitivity of the sparkline chart in the Explorer in the FE. I assumed it was going to be in Recharts after seeing it listed in the project's dependencies; however it appears D3 is being used for charts. I looked through the repo and it appears Recharts is completely unused. Let's remove it so that it doesn't bloat our build, introduce potential security risks, and just for the sake of Clean Code.

# Test Plan
To test out, just run `yarn` and `yarn test` to ensure that removing Recharts has not affected the project.
PS. This is my first PR. Please let me know if I'm not adhering to contribution requirements in any way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "react-markdown": "^4.3.1",
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
-    "recharts": "^1.8.5",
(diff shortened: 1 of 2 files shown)


wchargin (approved):
Thanks!

PR #2578, opened 2020-12-05
Bug/explorer mobile responsive fix

mkusold:
# Description
Currently, in the explorer view, the D3 sparkline charts break screen width responsitivity. This simple fix allows the SVG chart to scale and thus allows this page to be more responsive.

# Test
Run the site locally (`yarn start`) and go to the Explorer view. (which is also the default route).
Resize your screen or check it out in the chrome inspector and note that scrolling right in the mobile view no longer has the whitespace it did before.

Current:
explorer-mobile-before

Proposed PR:
explorer-mobile-after

### DIFF
--- a/src/ui/components/AdminApp.js
+++ b/src/ui/components/AdminApp.js
@@ -5,6 +5,7 @@ import {Redirect, Route, useHistory} from "react-router-dom";
 import {Admin, Resource, Layout, Loading} from "react-admin";
 import {createMuiTheme} from "@material-ui/core/styles";
 import pink from "@material-ui/core/colors/pink";
+import {makeStyles} from "@material-ui/core/styles";
 import fakeDataProvider from "ra-data-fakerest";
 import {Explorer} from "./Explorer/Explorer";
 import {ExplorerHome} from "./ExplorerHome/ExplorerHome";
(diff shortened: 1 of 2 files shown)


mkusold:
I also noticed we have a bit of WET code regarding the CredTimeline component. I propose consolidating the two near duplicates into one reusable component; however, I'm new to the project and don't want to overreach. I would also love more context on the top-level routing. In AdminApp.js there are 'routes' and 'backendroutes'. The Ledger and the Explorer, in particular, seem to overlap; however, how do I manually view the 'backendroutes' and what user role / purpose do these routes fulfill?

decentralion (commented):
I tested and couldn't repro the bug in Chrome, but did repro in safari, and verified that the change does what it says on the tin. I recommend specifying which browsers you've tested on for issues like this.

Before merging, this need CI to pass -- you can run `yarn test` locally, in this case lint and prettier are failing. See contributing.md for more info.

I'm not super familiar with css best practices in react and material-ui, going to add hammad/brian as reviewers since I think they're better suited.

mkusold:
@decentralion 
I was able to reproduce the bug in Chrome, Safari, and Firefox. 
Steps:
1. Run the site
2. Go to the Explorer view
3. Minimize the screen to around 900px width or less
4. Scroll to the right
The scrolling is what exposes the chart overflow with the white background.

hammadj (approved):
Looks good to me, feel free to merge after fixing the failing CI

topocount:
`yarn sharness-full` is passing locally for me after rebasing, so I'm going to go ahead and merge this.

PR #2579, opened 2020-12-08
Weights type: uniformly import as WeightT

elihanover:
__Context__
The `Weights` type is currently imported as either `Weights` or `WeightsT`.  The latter case is
used when the type conflicts with the `core/weights.js` module which is imported as `Weights` as
well.

__Change__
This renames the `Weights` type to `WeightsT` and changes all imports/instances of the type formerly
known as `Weights`.  

__Advantages__
This gives us consistency, which has the following advantage:

(1) Searching for instances of the `Weights` (now `WeightsT`) type is much simpler as we don't need
to also sort through instances of importing the `core/weights.js` module (and vice versa).

__Test Plan__
Manually checked import instances of the `core/weights.js` module for imports, and used Flow to check
for the remaining cases within those files.

Can also search for import instances using `git grep "import \(type \)\?{.*Weights.*}"`.

### DIFF
--- a/src/analysis/credView.js
+++ b/src/analysis/credView.js
@@ -1,7 +1,7 @@
 // @flow
 
 import sortedIndex from "lodash.sortedindex";
-import {type Weights, type EdgeWeight} from "../core/weights";
+import {type WeightsT, type EdgeWeight} from "../core/weights";
 import {type CredResult, compute} from "./credResult";
 import {type TimelineCredParameters} from "./timeline/params";
 import {type PluginDeclarations} from "./pluginDeclaration";
@@ -149,7 +149,7 @@ export class CredView {
     return this._credResult.weightedGraph.graph;
   }
 
-  weights(): Weights {
+  weights(): WeightsT {
     return this._credResult.weightedGraph.weights;
   }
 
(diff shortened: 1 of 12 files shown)


decentralion (commented):
For changes like this, I generally like the test plan to include a test for verifying that we didn't miss any. In this case, I'd suggest something like `git grep "import \(type \)\?{.*Weights.*}"`. That returns 19 hits, which is amenable to manual review; such manual review surfaces:
`src/analysis/weightsToEdgeEvaluator.js:import type {Weights} from "../core/weights";`

Overall, I like consistency so I'm open to this change. But if we're gonna do this, we should just rename `type Weights` to `type WeightsT` directly, which ensures both that it'll be consistent at time of change, but also that future usage will also be consistent with the principle.

elihanover:
> For changes like this, I generally like the test plan to include a test for verifying that we didn't miss any. In this case, I'd suggest something like `git grep "import \(type \)\?{.*Weights.*}"`. That returns 19 hits, which is amenable to manual review; such manual review surfaces:
> `src/analysis/weightsToEdgeEvaluator.js:import type {Weights} from "../core/weights";`

Done, added as suggestion.

> Overall, I like consistency so I'm open to this change. But if we're gonna do this, we should just rename `type Weights` to `type WeightsT` directly, which ensures both that it'll be consistent at time of change, but also that future usage will also be consistent with the principle.

Agreed, done.

decentralion (approved):
Opens up a space in the namespace for `class Weights`, which would be the classy version of this data type. :) 
With a bijection to go from Weights <-> WeightsT, we could then replace all the WeightsT usage one at a time, and then remove WeightsT.

elihanover:
@decentralion bijection ready 🍻, I'll submit a PR once this gets merged.

PR #2588, opened 2020-12-12
MPG: Support retrieving the intervals

decentralion:
This commit adds an `intervals` accessor to the MarkovProcessGraph which
allows retrieving the intervals. In order to get the end time of the
last interval, we add an extra numeric field to the JSON serialized form
(and the in memory representation).

I also added a new explicit failure case when there are no intervals.

Test plan: Unit tests added, `yarn test` passes.

### DIFF
--- a/src/core/credrank/markovProcessGraph.js
+++ b/src/core/credrank/markovProcessGraph.js
@@ -72,7 +72,7 @@ import * as NullUtil from "../../util/null";
 import * as MapUtil from "../../util/map";
 import type {TimestampMs} from "../../util/timestamp";
 import {type SparseMarkovChain} from "../algorithm/markovChain";
-import {type IntervalSequence} from "../interval";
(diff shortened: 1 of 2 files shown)


src/core/credrank/markovProcessGraph.js
```diff
@@ -165,6 +165,7 @@ export type MarkovProcessGraphJSON = {|
   +indexedEdges: $ReadOnlyArray<IndexedMarkovEdge>,
   +participants: $ReadOnlyArray<Participant>,
   +epochStarts: $ReadOnlyArray<number>,
+  +lastEpochEnd: number,
```
blueridger, line 168: nit: why not TimestampMs?

blueridger (approved):

src/core/credrank/markovProcessGraph.js
decentralion: Changed to `lastEpochEndMs`

PR #2584, opened 2020-12-11
grain allocation: Underpaid Policy

elihanover:
## Description
The current `Balanced` policy is prone to prioritizing paying users with
large underpayments at the expense of others.  This `Underpaid` policy
attempts to provide another option that communities can use alongside the
`Balanced` policy to dampen the impact of this undesirable property.

In English: "`Underpaid` tries to get everyone below some threshold of
underpayment by delaying the speed at which the most underpaid
people are paid out."

__How It Works__
The `Underpaid` policy pays contributors who are owed at least N grain as
specified in the policy's config, quadratically weighing the qualifying
underpayment amounts, using a configurable exponent in range (0, 1].

## __Test Plan__
- Unit tests provided for `Underpaid` in `core/ledger/grainAllocation` tests.
- Grain config unit tests provided in `api/grainConfig` tests.

__Other Considerations__
- In practice, `Underpaid` is a simple fork of `Balanced`.  Next, I plan
on refactoring the policy modules so that we can share component functions
across them.
- As followup, I plan on splitting policy tests into their own test modules.

PR #2600, opened 2020-12-14

elihanover:
__Context__
After implementing the `Recent` policy and a variant of `Balanced` (#2584), we can see that adding policy parameters for new policies is both tedious on from our perspective as well as for the user.

More specifically, on the user side, their config is a flat object with parameters from different policies sitting side by side.  Here, we implement more structure by taking an array of policies.

Right now, an example `config/grain.json` might look as follows:
```js
{
  "balancedPerWeek": 30
  "immediatePerWeek": 50,
  "recentPerWeek": 20,
  "maxSimultaneousDistributions": 2,
  "discount": 0.5
}
```

This also forced us to check each policy explicitly downstream in `toDistributionPolicy`, which could be a lot cleaner if we could leverage the `AllocationPolicy` type we already have.  Now, we only need to check that at least one policy is provided and budgets are positive.

```js
{
  allocationPolicies: [
    {
      policyType: "BALANCED",
      budget: "100"
    },
    {
      policyType: "RECENT",
      budget: "50",
      discount: 0.1
    },
    {
      policyType: "RECENT",
      budget: "100",
      discount: 0.5
    }
}
```

This also lets us take multiple versions of the same policies.  This allows us to create richer ensembles of policies.  For instance, I might want to run a `Recent` policy with a 90% discount, and another one with a 5% discount.

__Test Plan__
Unit tests are provided here for both the parser, and well as `toDistributionPolicy`, which serves to check the config.

### DIFF
--- a/src/api/grainConfig.js
+++ b/src/api/grainConfig.js
@@ -1,33 +1,44 @@
 // @flow
 
 import {type DistributionPolicy} from "../core/ledger/applyDistributions";
-import * as G from "../core/ledger/grain";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
+import {
+  type AllocationPolicy,
+  policyConfigParser,
+} from "../core/ledger/policies";
+import {fromInteger as toNonnegativeGrain} from "../core/ledger/policies/nonnegativeGrain";
 import {toDiscount} from "../core/ledger/policies/recent";
 
 export type GrainConfig = {|
-  +immediatePerWeek?: number,
-  +balancedPerWeek?: number,
-  +recentPerWeek?: number,
-  +recentWeeklyDecayRate?: number,
+  +immediatePerWeek?: number, // (deprecated)
+  +balancedPerWeek?: number, // (deprecated)
+  +recentPerWeek?: number, // (deprecated)
+  +recentWeeklyDecayRate?: number, // (deprecated)
+  +allocationPolicies: $ReadOnlyArray<AllocationPolicy>,
   +maxSimultaneousDistributions?: number,
 |};
 
 export const parser: C.Parser<GrainConfig> = C.object(
-  {},
   {
+    allocationPolicies: C.array<AllocationPolicy>(policyConfigParser),
+  },
+  {
+    maxSimultaneousDistributions: C.number,
     immediatePerWeek: C.number,
     balancedPerWeek: C.number,
     recentPerWeek: C.number,
     recentWeeklyDecayRate: C.number,
-    maxSimultaneousDistributions: C.number,
   }
 );
 
 /**
  * Create a DistributionPolicy from GrainConfig, checking that config
  * fields can form valid policies.
+ *
+ * Moving forward, policies will need to be passed in the `allocationPolicies`
+ * parameter; however to avoid backcompatability issues, we optionally allow
+ * the deprecated fields for the time being.
  */
 export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
   const immediatePerWeek = NullUtil.orElse(x.immediatePerWeek, 0);
@@ -49,10 +60,11 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
       `balanced budget must be nonnegative integer, got ${balancedPerWeek}`
     );
   }
-  const allocationPolicies = [];
+
+  const allocationPoliciesDeprecated: Array<AllocationPolicy> = [];
   if (immediatePerWeek > 0) {
-    allocationPolicies.push({
-      budget: G.fromInteger(immediatePerWeek),
+    allocationPoliciesDeprecated.push({
+      budget: toNonnegativeGrain(immediatePerWeek),
       policyType: "IMMEDIATE",
     });
   }
@@ -61,23 +73,29 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
     if (recentWeeklyDecayRate == null) {
       throw new Error(`no recentWeeklyDecayRate specified for recent policy`);
     }
-    allocationPolicies.push({
-      budget: G.fromInteger(recentPerWeek),
+    allocationPoliciesDeprecated.push({
+      budget: toNonnegativeGrain(recentPerWeek),
       policyType: "RECENT",
       discount: toDiscount(recentWeeklyDecayRate),
     });
   }
   if (balancedPerWeek > 0) {
-    allocationPolicies.push({
-      budget: G.fromInteger(balancedPerWeek),
+    allocationPoliciesDeprecated.push({
+      budget: toNonnegativeGrain(balancedPerWeek),
       policyType: "BALANCED",
     });
   }
+
   const maxSimultaneousDistributions = NullUtil.orElse(
     x.maxSimultaneousDistributions,
     Infinity
   );
-  return {allocationPolicies, maxSimultaneousDistributions};
+  return {
+    allocationPolicies: x.allocationPolicies.concat(
+      allocationPoliciesDeprecated
+    ),
+    maxSimultaneousDistributions,
+  };
 }
 
 function isNonnegativeInteger(x: number): boolean {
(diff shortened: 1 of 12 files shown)


src/api/grainConfig.js
```diff
@@ -4,23 +4,21 @@ import {type DistributionPolicy} from "../core/ledger/applyDistributions";
...
+} from "../core/ledger/policies";
 
 export type GrainConfig = {|
-  +immediatePerWeek?: number,
-  +balancedPerWeek?: number,
-  +recentPerWeek?: number,
-  +recentWeeklyDecayRate?: number,
+  +allocationPolicies: $ReadOnlyArray<AllocationPolicy>,
   +maxSimultaneousDistributions?: number,
 |};
```
decentralion, line 15: This breaks backcompat and forces an immediate upgrade for every user. Let's not do this.

Instead, please mark (thru comments) that all of the existing fields are deprecated, and leave them as optional fields.
Add the new allocationPolicies field.
Then we can upgrade all of the examples and our instance to use allocationPolicies.
Then we can remove it (ideally with a deprecation warning printed to console for the next few releases)

src/api/grainConfig.js
```diff
@@ -30,56 +28,24 @@ export const parser: C.Parser<GrainConfig> = C.object(
...
-    throw new Error(
-      `immediate budget must be nonnegative integer, got ${immediatePerWeek}`
-    );
-  }
-  if (!isNonnegativeInteger(recentPerWeek)) {
-    throw new Error(
-      `recent budget must be nonnegative integer, got ${recentPerWeek}`
-    );
+  if (!x.allocationPolicies.length) {
+    throw new Error(`Please provide at least one allocation policy, got zero.`);
```
decentralion, line 32: Zero is a valid input with an unambiguous interpretation. Throwing errors on such inputs just makes the system harder to reason about. Please support zero.

src/api/grainConfig.js
```diff
@@ -30,56 +28,24 @@ export const parser: C.Parser<GrainConfig> = C.object(
...
-  if (recentPerWeek > 0) {
-    const {recentWeeklyDecayRate} = x;
-    if (recentWeeklyDecayRate == null) {
-      throw new Error(`no recentWeeklyDecayRate specified for recent policy`);
+
+  x.allocationPolicies.map((policy) => {
+    if (!G.gt(policy.budget, G.ZERO)) {
+      throw new Error(
+        `${policy.policyType} budget must be positive integer, got ${policy.budget}`
+      );
```
decentralion, line 39: As above: a zero budget is totally fine.

src/api/grainConfig.js
```diff
@@ -4,23 +4,21 @@ import {type DistributionPolicy} from "../core/ledger/applyDistributions";
...
+  type AllocationPolicy,
+  allocationPolicyParser,
+} from "../core/ledger/policies";
 
 export type GrainConfig = {|
-  +immediatePerWeek?: number,
-  +balancedPerWeek?: number,
-  +recentPerWeek?: number,
-  +recentWeeklyDecayRate?: number,
+  +allocationPolicies: $ReadOnlyArray<AllocationPolicy>,
```
decentralion, line 13: Every AllocationPolicy has a budget field which is a raw attoGrain value, e.g. if they want to distribute 10000 Grain per period, they would need to write `10000000000000000000000`. This is not a user friendly UX.

As a proof-of-viability, in the next iteration please include in the PR description an example of what SC's own grain config file will look like once updated.

decentralion:
The big issue with this PR is that AllocationPolicies do not serialize in a way that's convenient for hand-editing, because Grain does not serialize in a convenient fashion.
I recommend solving this by splitting into two types: a `Policy` and `PolicyConfig`.
The `Policy` is the raw data needed to compute the policy.
The `PolicyConfig` is a serializable type that affords a convenient way to hand-edit the policy.

Then you can write parsers for the `PolicyConfig` (which is great for all user-provided input; ensures clear error messages when configs are wrong) and functions to transform Policy <-> PolicyConfig

Heuristic: PolicyConfig has many optional fields, Policy has only required fields.

src/api/grainConfig.js
elihanover: Done, I agree with this approach.

src/api/grainConfig.js
elihanover: Done.

src/api/grainConfig.js
elihanover: Done.

src/api/grainConfig.js
elihanover: Done, added parser to take number and transform it to `NonnegativeGrain` so users can write amounts as before.

elihanover:
- I've taken your advice and allowed for optional properties, making sure to test they're both compatible.
- Added config parsers for each allocation policy which compose into a `policyConfigParser`, which allow us to take grain amounts instead of attoGrain amounts, without the need for an additional type.
- Added a `NonnegativeGrain` type which is used by each allocation policy as the type for their budget.

All in all, once we remove the deprecated config, we'll be able to replace `toDistributionPolicy` with a simple parser combinator.

**Question:** what dir should `NonnegativeGrain` sit in?

src/api/grainConfig.js
```diff
@@ -1,33 +1,44 @@
...
+  +recentWeeklyDecayRate?: number, // (deprecated)
+  +allocationPolicies: $ReadOnlyArray<AllocationPolicy>,
   +maxSimultaneousDistributions?: number,
 |};
 
 export const parser: C.Parser<GrainConfig> = C.object(
-  {},
   {
+    allocationPolicies: C.array<AllocationPolicy>(policyConfigParser),
+  },
```
decentralion, line 25: By making this required, we break every existing GrainConfig. Let's make it optional too (default to []).

src/api/grainConfig.test.js
```diff
@@ -3,122 +3,240 @@
...
+  recipient,
+});
 
 describe("api/grainConfig", () => {
   describe("parser", () => {
-    it("does not throw with no params", () => {
-      expect(parser.parseOrThrow({})).toEqual({});
+    it("errors if missing allocationPolicies key", () => {
+      expect(() => parser.parseOrThrow({})).toThrowError(`missing key`);
     });
```
decentralion, line 41: as noted above, we should handle this case for backcompat

src/api/grainConfig.js
```diff
@@ -49,10 +60,11 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
       `balanced budget must be nonnegative integer, got ${balancedPerWeek}`
     );
   }
-  const allocationPolicies = [];
+
+  const allocationPoliciesDeprecated: Array<AllocationPolicy> = [];
```
decentralion, line 64: Why not just push them directly into the allocationPolicies array?

decentralion (commented):
Looks pretty good now. For clarity of review I'd prefer this be split into three PRs:

1. Add the nonnegative grain type into `core/ledger/grain` or `core/ledger/nonnegativeGrain`. My 2c is that putting it into `core/ledger/grain` is fine.
2. Add the new serializedAllocationPolicy parser. One thing I'm a little concerned with is: where are we documenting the serializedAllocationPolicy type, so users will know how to write them? I mostly want this to be in its own PR so it can be clear where/how this is documented.
3. update the GrainConfig to use the optional allocation policies, and mark the existing fields as deprecated

elihanover:
Closing in favor of #2666.  Splitting this into 3 PRs.

src/api/grainConfig.test.js
elihanover: Done

src/api/grainConfig.js
elihanover: Done

src/api/grainConfig.js
elihanover: Ack, but I like thinking of these processes as a tree.  One branch handles `x.allocationPolicies` and the other handles the deprecated policies.

PR #2608, opened 2020-12-21
Spec out WeightedGraph2 external API

topocount:
# Description

WeightedGraph2 is the successor to WeightedGraph. It is
a class rather than a type, and as such, it abstracts away the
requirement of assigning every single node and edge (entity) each their own
weight values. With this class, arbitrary sets of nodes and edges can
all be assigned a weight with one assignment, by assigning an Address
prefix a weight.

An Address prefix is a subset of address parts that always contains
the address' root part as the first entry. The prefix may also include
an entity's entire address in this case, if a plugin wishes to update
an entity's weight after calling `addNode` or `addWeight`.

Entities may also have their weights assigned directly. If an
entity is passed in without a weight, it is assigned the default weight
of 1 (or 1 in both directions for edges).

This class will also be extensible to work include tags in the future.

test plan: none; No logic is implemented. This is solely for the purpose
of API design review.

# Test Plan
none; No logic is implemented. This is solely for the purpose
of API design review.


src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+  +children: Map<AddressPart, NodeWeightBranch>,
+};
+
+/**
+ * the core structure in the Edge weight-resolution tree
+ */
+export type EdgeWeightBranch = {
+  +weight?: EdgeWeight,
+  +children: Map<AddressPart, EdgeWeightBranch>,
+};
```
decentralion, line 34: Have you taken a look at the [Trie](https://github.com/sourcecred/sourcecred/blob/master/src/core/trie.js) module? We should be able to re-use a Trie rather than implementing any tree logic here.

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+ * The HeavyGraph class is a replacement for the WeightedGraph type. This class
+ * eliminates the need give every node and edge that exists a weight. Rather,
+ * address prefixes can be assigned a weight and when a node or edge weight is
+ * queried, the longest prefix with a weight assigned will have its value
+ * returned.
+ *
+ * Over the longer-term, WeightedGraph should be
+ * refactored out of existence HeavyGraph should completely replace it.
+ */
+export class HeavyGraph {
```
decentralion, line 46: Let's call it `WeightedGraph2`, with an intent to rename it to `WeightedGraph` once we've updated every user. I think the name "WeightedGraph" is much more descriptive and intuitive than "HeavyGraph".

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+   * Set the weight for a given "class" of nodes. Each node that contains the
+   * supplied prefix in its address will potentially return this weight if no
+   * weight is assigned for a larger prefix that the node's address contains.
+   */
+  setNodePrefixWeight(
+    _unused_prefix: NodeAddressT,
+    _unused_w: NodeWeight
+  ): this {
+    throw new Error("method not implemented");
+  }
```
decentralion, line 61: In the future, I think we should use tags to accomplish this goal. But for now, agree that we need this API so we can be compatible with the existing codebase.

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Add a node and optionally assign its weight. If a prefix of the node's
+   * address doesn't have a weight assigned, a weight must be included or an
+   * error will throw.
+   */
```
decentralion, line 79: "If a prefix doesn't have a weight assigned, a weight must be included" -- this is a departure from our existing semantics, which treats the `1` as the default node weight and `{1,1}` as the default edge weight. Having it throw an error makes the behavior inconsistent, consider the following two examples:
```js
let wg = new WeightedGraph2();
let node = exampleNode();
// fine
wg.addNode(node);
wg.setNodePrefixWeight(NodeAddress.empty, 2);


wg = new WeightedGraph2();
wg.setNodePrefixWeight(NodeAddress.empty, 2)
// this will unexpectedly throw an error
wg.addNode(node)
```

We switched the order of operations, and the result should be equivalent, but now it errors instead of succeeding.

decentralion, line 79: What happens if you add a node that already exists, but are re-assigning the weight?

decentralion, line 79: What happens if there's already a prefix weight assigned for that node's exact address?

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+  addNode(_unused_n: Node, _unused_w?: NodeWeight): this {
+    throw new Error("method not implemented");
+  }
+
```
decentralion, line 83: Missing: An iterator for the weighted nodes

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+  /**
```
decentralion, line 84: Missing: A method for retrieving a particular weighted node

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+   * Return the weight of the node, or the weight of the longest prefix of the
+   * node address if the node doesn't have its own weight.
+   */
+  getNodeWeight(_unused_a: NodeAddressT): NodeWeight {
+    throw new Error("method not implemented");
+  }
```
decentralion, line 99: Assuming we keep the default weights of `1` and `{1,1}` (with multiplicative semantics), as discussed in a comment above, then we can realize that weights are really associated with addresses and not with nodes. Since every node has an address, whenever we are retrieving a particular node we can return it with the weight associated with its address. 

But every address has a weight, and not every address has a node. So for the docstring to say "Return the weight of the node, or the weight of the address if the node doesn't have a weight" makes a confusing API. I think we should just create a clear and locally consistent API, where we are returning the weight associated with the address directly.

src/core/heavyGraph.js
```diff
@@ -0,0 +1,122 @@
...
+  +weight?: EdgeWeight,
+  +children: Map<AddressPart, EdgeWeightBranch>,
+};
+
+/**
+ * The HeavyGraph class is a replacement for the WeightedGraph type. This class
+ * eliminates the need give every node and edge that exists a weight. Rather,
+ * address prefixes can be assigned a weight and when a node or edge weight is
+ * queried, the longest prefix with a weight assigned will have its value
+ * returned.
```
decentralion, line 41: This is the wrong semantics for graph weights; we return the product of all of the weights along the address path chain. See:

https://github.com/sourcecred/sourcecred/blob/651db6de017cc74dfc85f16960144d0849fab105/src/core/algorithm/weightEvaluator.js#L10-L63

decentralion (commented):
Looks like a nice start! See inline comments. All the issues I raised for nodes also apply to edges.

decentralion (approved):
Should depend on @elihanover's work in https://github.com/sourcecred/sourcecred/pull/2617

src/core/heavyGraph.js
```diff
@@ -0,0 +1,127 @@
+// @flow
```
hammadj, line 1: Looks like a great start, only nit is that the filename should also be updated to `weightedGraph2.js` instead of `heavyGraph.js`

PR #2610, opened 2020-12-22
cli: Add simulation flag to graph cmd

blueridger:
# Description

This adds a simulation flag to the `sourcecred graph` command, which simply skips writing changes to disk when added. This allows faster iteration during development, especially when using the `-d` flag, because it is most likely that a developer will want to repeatedly diff changes against a single "original" graph, rather than comparing each tweak to the last tweak. This change will allow a developer to execute such an iterative workflow by simply repeating `sourcecred graph -d -s` instead of constantly switching between branches to reset to the original graph.

I considered combining these semantics into the existing `-d` command since I would generally expect the two flags to be used together, but I opted to give the user finer grained control just in case.

# Test Plan
- confirmed that the order of params like `sourcecred/github --simulation -d` doesn't matter
- confirmed that both `-s` and `--simulation` work
- confirmed that ledger is changed iff `-s` is absent
- confirmed that the taskReporter output is as expected for each combination of flags:

## both -s and -d
```
  GO   sourcecred/discord: generating graph
 DONE  sourcecred/discord: generating graph: 17ms
  GO   sourcecred/discord: diffing with existing graph
  sourcecred/discord - Unchanged
 DONE  sourcecred/discord: diffing with existing graph: 13ms
```

## -d only
```
  GO   sourcecred/discord: generating graph
 DONE  sourcecred/discord: generating graph: 23ms
  GO   sourcecred/discord: diffing with existing graph
  sourcecred/discord - Unchanged
 DONE  sourcecred/discord: diffing with existing graph: 13ms
  GO   sourcecred/discord: writing graph
 DONE  sourcecred/discord: writing graph: 4ms
```

## -s only
```
  GO   sourcecred/discord: generating graph
 DONE  sourcecred/discord: generating graph: 16ms
```

## neither flag
```
  GO   sourcecred/discord: generating graph
 DONE  sourcecred/discord: generating graph: 17ms
  GO   sourcecred/discord: writing graph
 DONE  sourcecred/discord: writing graph: 4ms
```

### DIFF
--- a/src/cli/graph.js
+++ b/src/cli/graph.js
@@ -43,10 +43,15 @@ function die(std, message) {
 
 const graphCommand: Command = async (args, std) => {
   let shouldIncludeDiff = false;
+  let isSimulation = false;
   const processedArgs = args.filter((arg) => {
     switch (arg) {
       case "-d":
         shouldIncludeDiff = true;
+        return false;
+      case "--simulation":
+      case "-s":
+        isSimulation = true;
         return false;
       default:
         return true;
@@ -86,11 +91,17 @@ const graphCommand: Command = async (args, std) => {
     ledger
   );
   for (const name of pluginsToLoad) {
+    const generateGraphTask = `${name}: generating graph`;
+    taskReporter.start(generateGraphTask);
     const plugin = NullUtil.get(config.bundledPlugins.get(name));
-    const task = `${name}: generating graph`;
-    taskReporter.start(task);
     const dirContext = pluginDirectoryContext(baseDir, name);
     const weightedGraph = await plugin.graph(dirContext, rd, taskReporter);
+
+    const identities = await plugin.identities(dirContext, taskReporter);
+    for (const identityProposal of identities) {
+      ensureIdentityExists(ledger, identityProposal);
+    }
+    taskReporter.finish(generateGraphTask);
 
     if (shouldIncludeDiff) {
       const diffTask = `${name}: diffing with existing graph`;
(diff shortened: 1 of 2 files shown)


src/cli/graph.js
```diff
@@ -92,6 +97,12 @@ const graphCommand: Command = async (args, std) => {
     const dirContext = pluginDirectoryContext(baseDir, name);
     const weightedGraph = await plugin.graph(dirContext, rd, taskReporter);
 
+    const identities = await plugin.identities(dirContext, taskReporter);
+    for (const identityProposal of identities) {
+      ensureIdentityExists(ledger, identityProposal);
+    }
+    taskReporter.finish(task);
```
decentralion, line 104: I notice we change when `task` (which should really be called `generateGraphTask`-- mind renaming?) gets resolved. I think the new semantics might be better? but I want to confirm this is intentional.

decentralion (commented):
Seems like a really nice change, well motivated by the PR description, thank you. I def agree with leaving the ability to pass `-d` without `-s`. However, I am struggling to think of a reason you would want `-s` and not `-sd` because just `-s` is a no-op (useful only for timing measurements I guess?). So I think it would be reasonable to make `-s` also imply `-d`.

src/cli/graph.js
blueridger: Correct, this is intentional. I think it's better this way too. More atomic and less nesting. I'll update the variable name.

blueridger:
> Seems like a really nice change, well motivated by the PR description, thank you. I def agree with leaving the ability to pass `-d` without `-s`. However, I am struggling to think of a reason you would want `-s` and not `-sd` because just `-s` is a no-op (useful only for timing measurements I guess?). So I think it would be reasonable to make `-s` also imply `-d`.

Agreed it's a little weird in this case. Timing measurements is almost enough motivation to justify the no-op simulation, and designing for future cases contributes too, but I think my main reasoning is simply consistent semantics with the existing `grain -s` and my plans to add `credrank -s`. I think `-s` should consistently do exactly what the command would otherwise do except write to disk.

decentralion (approved):

decentralion:
Kind of wish we had unit tests here, specifically for the "if simulation, do not re-write the ledger" invariant.

PR #2611, opened 2020-12-23
cli: Add simulation flag to credrank cmd

blueridger:
# Description
This adds a simulation flag to the `sourcecred credrank` command, which simply skips writing changes to disk when added. This allows faster iteration during development, especially when using the `-d` flag, because it is most likely that a developer will want to repeatedly diff changes against a single "original" credGraph, rather than comparing each tweak to the last tweak. This change will allow a developer to execute such an iterative workflow by simply repeating `sourcecred credrank -d -s` instead of constantly switching between branches to reset to the original credGraph.

# Test Plan
`scdev credrank` outputs `  GO   write cred graph
 DONE  write cred graph: 8ms` whereas `scdev credrank -s` does not.

### DIFF
--- a/src/cli/credrank.js
+++ b/src/cli/credrank.js
@@ -23,10 +23,15 @@ function die(std, message) {
 
 const credrankCommand: Command = async (args, std) => {
   let shouldIncludeDiff = false;
+  let isSimulation = false;
   const processedArgs = args.filter((arg) => {
     switch (arg) {
       case "-d":
         shouldIncludeDiff = true;
+        return false;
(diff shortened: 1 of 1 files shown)


decentralion (approved):
From an end user perspective, I'd probably want to simulate an end-to-end run of CredRank w/ different parameters, i.e. using a different graph. Like when I change the Discord weights and want to see the E2E Cred changes. This won't be useful for that end user need, but it does still seem useful for debugging the credrank command in particular.

PR #2616, opened 2020-12-28
Ledger table

blueridger:
# Description
This implements the TableState into the LedgerViewer. It migrates existing pagination to the TableState implementation, and adds sort-by-timestamp functionality.

# Test Plan
React nodes are not covered by unit tests.
## Manual Testing
I tested:
- extensive combinations of pagination, sort, and filtering
- multiple filters applied together
- making a change to the ledger (creating an identity or transferring grain) and verifying that the change appears in the table
- calling useTableState with and without optional parameters

You can see some manual testing on [this demo video](https://photos.app.goo.gl/6jsXkXaH33ZgoBpu5). The demo video includes event name filtering, but that is not included in this PR, since it needs some more UX thought.

To test for yourself, check out this branch and run `yarn start --instance /path/to/instance`

### DIFF
--- a/src/ui/components/LedgerViewer.js
+++ b/src/ui/components/LedgerViewer.js
@@ -7,6 +7,7 @@ import React, {
   useMemo,
   useState,
 } from "react";
+import deepFreeze from "deep-freeze";
 import Table from "@material-ui/core/Table";
 import Typography from "@material-ui/core/Typography";
 import TablePagination from "@material-ui/core/TablePagination";
@@ -18,6 +19,7 @@ import Toolbar from "@material-ui/core/Toolbar";
 import DialogContent from "@material-ui/core/DialogContent";
 import Tooltip from "@material-ui/core/Tooltip";
 import Chip from "@material-ui/core/Chip";
+import TableSortLabel from "@material-ui/core/TableSortLabel";
 import TableCell from "@material-ui/core/TableCell";
 import TableContainer from "@material-ui/core/TableContainer";
 import TableHead from "@material-ui/core/TableHead";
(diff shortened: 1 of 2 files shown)


src/ui/components/LedgerViewer.js
```diff
@@ -99,21 +105,29 @@ export const LedgerViewer = ({
             <TableRow>
               <TableCell>Event</TableCell>
               <TableCell>Details</TableCell>
-              <TableCell align="right">Date</TableCell>
+              <TableCell align="right">
+                <TableSortLabel
+                  active={ts.sortName === DATE_SORT.name}
+                  direction={
+                    ts.sortName === DATE_SORT.name ? ts.sortOrder : DEFAULT_SORT
```
hammadj, line 112: Why do we only use the sortOrder from TableState when its Date and not any other filter?

src/ui/components/LedgerViewer.js
```diff
@@ -51,6 +58,13 @@ const useStyles = makeStyles((theme) => {
   };
 });
 
+const DATE_SORT = deepFreeze({
+  name: Symbol("Date"),
+  fn: (e: LedgerEvent) => e.ledgerTimestamp,
+});
+
+const PAGINATION_OPTIONS = deepFreeze([25, 50, 100]);
```
hammadj, line 66: Curious to hear why we are deepFreezing these consts.

hammadj (approved):
few questions, looks good overall!

src/ui/components/LedgerViewer.js
blueridger: Each TableSortLabel corresponds to 1 column. The sortOrder applies only to the current sortName. Although there is only 1 sortable column in this PR, there could be others later. If this column is not the actively sorted column, then the button should show the default direction (the direction that the column will sort to upon becoming the actively sorted column).

The easiest way to understand is probably trying out the UX for yourself.

src/ui/components/LedgerViewer.js
blueridger: Dandelion recommended it. It seems to align with our pattern of having the most restrictive writability possible for stored values.

PR #2617, opened 2020-12-31
core/weights: Weights class

elihanover:
This commit adds a `Weights` class which is part of an effort to
shape a cleaner experience handling node and edge weights, particularly
when using a `NodeWeightEvaluator` or `EdgeWeightEvaluator`.

Currently, we cannot get the node or edge weight directly as it requires transforming
a Map into a Trie (which is performed by the `Node/EdgeWeightEvaluator`).
This class creates an interface for getting node and edge weights directly
from the class, letting it lazily handle contructing the trie.

```js
// Before
// weights: WeightsT
import { nodeWeightEvaluator, edgeWeightEvaluator } from "../***/weightEvaluator"
const nwe = nodeWeightEvaluator(weights);
const ewe = edgeWeightEvaluator(weights);
const nodeWeight = nwe(nAddress);
const edgeWeight = ewe(eAddress);

// After
// weights: Weights (class)
const nodeWeight = weights.nodeWeight(nAddress);
const edgeWeight = weights.edgeWeight(eAddress);
```

__Test Plan__
Unit tests provided for the class methods that test along the same conditions
that the `WeightsT` functions are tested.  For the `node/edgeWeight` methods,
tests are a superset of `weightEvaluator` tests, also checking for the lazy
creation of a `node/edgeWeightEvaluator` on the class.


decentralion (commented):
This API is very inconsistent with adjacent APIs, e.g. the Graph API.

Consider the flow with Graph:

```
const g = new Graph()
g.addNode(srcNode)
g.addNode(dstNode)
g.addEdge(theEdge)
```

We would logically expect Weights to have a similar API

```
const w = new Weights()
weights.setNodeWeight(srcNode.address, 7)
weights.setEdgeWeight(myEdge.address, {to: 5, fro: 3})
```

With this API, the only way to construct a `Weights` is to construct an old, weird `WeightsT`, so it does not accomplish the intent of the refactor.

decentralion:
Also -- take a look at https://github.com/sourcecred/sourcecred/pull/2608, @topocount is working on a class that will want to consume the weights class.

elihanover:
> This API is very inconsistent with adjacent APIs, e.g. the Graph API.
> 
> Consider the flow with Graph:
> 
> ```
> const g = new Graph()
> g.addNode(srcNode)
> g.addNode(dstNode)
> g.addEdge(theEdge)
> ```
> 
> We would logically expect Weights to have a similar API
> 
> ```
> const w = new Weights()
> weights.setNodeWeight(srcNode.address, 7)
> weights.setEdgeWeight(myEdge.address, {to: 5, fro: 3})
> ```
> 
> With this API, the only way to construct a `Weights` is to construct an old, weird `WeightsT`, so it does not accomplish the intent of the refactor.

Done.  Check out the `WeightsI` type in the PR message.

elihanover:
@decentralion I'm having trouble with `sharness` test headers.  For instance on "core/weights/weights.test.js" I have the (failing) following header:
```js
describe("core/weights/weightsT", () => {
```

src/core/weights/nodeWeights/nodeWeightsT.test.js
```diff
@@ -0,0 +1,90 @@
...
+  nodeWeightsT: [string, number][]
+): NodeWeightsT {
+  const nw = empty();
+  for (const [addrPart, weight] of nodeWeightsT) {
+    nw.set(NodeAddress.fromParts([addrPart]), weight);
+  }
+  return nw;
+}
+
+describe("core/weights/edgeWeights/edgeWeightsT", () => {
```
topocount, line 16: I think this might be at least part of the issue with your tests

```suggestion
describe("core/weights/nodeWeights/nodeWeightsT", () => {
```

topocount (commented):
This seems sensible to me.

optional nit: I'd prefer to see the implementation itself split into a separate PR from the refactor that utilizes this new type. It'll get you more cred if you decide to do that

topocount (commented):
I just realized a few other details, based on the design of WeightedGraph2. We should be using tries instead of Maps to store Node and Edge weights, no?

@elihanover do you want me to refactor this a bit further to favor tries instead of maps? Not sure if you're working on other stuff, and I'd be happy to help get this in since I need it for WG2.

src/core/weights/edgeWeights/edgeWeights.js
```diff
@@ -0,0 +1,53 @@
...
+import {orElse as either} from "../../../util/null";
+import {
+  type EdgeWeightsT,
+  type EdgeOperator,
+  type EdgeWeight,
+  empty,
+  copy,
+  merge,
+} from "./edgeWeightsT";
+
```
topocount, line 14:
```suggestion
/**
 * A higher level interface on EdgeWeightsT that includes helper functions
 */
```

src/core/weights/nodeWeights/nodeWeights.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const NodeWeights = (
+  nodeWeightsT: NodeWeightsT = empty()
+): NodeWeightsI => {
+  let nwe;
+  return {
+    fold: () => nodeWeightsT,
+    getNodeWeight: (address: NodeAddressT): NodeWeight => {
+      nwe = either(nwe, nodeWeightEvaluator(nodeWeightsT));
+      return nwe(address);
```
topocount, line 43: Why are we lazily defining the node and edge weight evaluators on the fly? If this is provided by the algorithm module and only used by the algorithm module why not just definitively assign it?

topocount (commented):
What's the reason for not turning all these objects into classes? It'd be nice to properly call these abstractions interfacess and classes and actually mean the javascript class and flow interface

src/core/weights/edgeWeights/edgeWeights.js
topocount: Let's add a similar comment for NodeWeights too

src/core/weights/weights.js
```diff
@@ -0,0 +1,137 @@
...
+ * order object, allowing us to interface with the underlying
+ * Node/EdgeWeights while holding the aggregated Weights here.
+ */
+export const Weights = (
+  nodeWeights: NodeWeightsI = NodeWeights(),
+  edgeWeights: EdgeWeightsI = EdgeWeights()
+): WeightsI => {
+  const {getNodeWeight, setNodeWeight, fold: nodeWeightsT} = nodeWeights;
+  const {getEdgeWeight, setEdgeWeight, fold: edgeWeightsT} = edgeWeights;
+  const fold = () => toWeightsT(nodeWeights.fold(), edgeWeights.fold());
```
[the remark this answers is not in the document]
elihanover: 👍

src/core/weights/edgeWeights/edgeWeights.js
elihanover: ✅

src/core/weights/nodeWeights/nodeWeights.js
elihanover: It's a sort of a gorilla/banana dynamic here.  If we're just using Weights to modify data or merge or copy Weights data, we don't need to carry along this trie with us.

elihanover: I would just load it if the lazily creation was complicated, but I think the lazy version is simple enough.

elihanover:
Regarding classes, I try to avoid them unless they're needed.  All we're doing here is composing some functions and data, and I think the code reflects that simple process.

Also creating these with functions over traditional constructors makes composing them cleaner.  For instance, here we'd have to add `new`, bulking up any transformation using these.  Maybe this is minor, but I see classes as doing more to get more bulk.
```js
/**
 * Lifts Node/EdgeWeightsT instead of Node/EdgeWeights by abstracting
 * the intermediate transformatino to Node/EdgeWeights.
 */
export function weightsTToWeights(
  nodeWeightsT: NodeWeightsT = emptyNodeWeightsT(),
  edgeWeightsT: EdgeWeightsT = emptyEdgeWeightsT()
): WeightsI {
  return Weights(NodeWeights(nodeWeightsT), EdgeWeights(edgeWeightsT));
  // return Weights(new NodeWeights(nodeWeightsT), new EdgeWeights(edgeWeightsT));
}
```

Could be interesting to see this code using classes, but my sense is that it would make it a little more complicated / bulky.

topocount:
@decentralion I think this is good to merge, pending a pass from you

topocount:
@elihanover it looks like there are some merge conflicts that need to be resolved now

decentralion (commented):
This PR is huge -- looks like it's adding new logic and also doing an across-the-codebase find+replace. If we're moving forward w/ this, it needs to get split up into atomic pieces, e.g. add new logic first, then do any renames.

However I feel like we should hold off until we have the refactoring crew / graph squad with a clearer plan.

PR #2622, opened 2021-01-04
Testing: Remove Snapshot from createGraph tests 

topocount:
[This file](https://github.com/sourcecred/sourcecred/blob/5a6eaccd305642ee0be444344d4fb772199cecf5/src/plugins/discourse/createGraph.test.js) fails the sharness test for file names in the describe string
when grep is aliased without the `-a` option. This option forces grep to
attempt to parse files, even if they are recognized as binary type.

The reason for this seems to be that in POSIX systems, any file
containing a null literal is recognized by default as a binary file.

Rather than utilizing a snapshot, the test now manually transforms the
edge properties into console and source-safe parts for comparison.
Additionally, a length check is added to ensure that likedPosts are
indeed tested and empty arrays aren't compared.

test plan: yarn test

### DIFF
--- a/src/plugins/discourse/createGraph.test.js
+++ b/src/plugins/discourse/createGraph.test.js
@@ -2,9 +2,16 @@
 
 import sortBy from "../../util/sortBy";
 import * as NullUtil from "../../util/null";
+import {type TimestampMs} from "../../util/timestamp";
 import type {ReadRepository} from "./mirrorRepository";
 import type {Topic, Post, PostId, TopicId, LikeAction, User} from "./fetch";
-import {EdgeAddress, type Node, type Edge, Graph} from "../../core/graph";
+import {
+  EdgeAddress,
+  NodeAddress,
+  type Node,
+  type Edge,
+  Graph,
+} from "../../core/graph";
 import {
   _createReferenceEdges,
   weightForTrustLevel,
@@ -489,6 +496,20 @@ describe("plugins/discourse/createGraph", () => {
       expect(seenTrustLevels).toEqual(new Set([null, 0, 2, 3]));
     });
     it("creates hasLikedPost edges from topics to posts", () => {
+      type SafeEdge = {|
+        +address: string[],
+        +dst: string[],
+        +src: string[],
+        +timestampMs: TimestampMs,
+      |};
+      function getSafeEdge(e: Edge): SafeEdge {
+        return {
+          address: EdgeAddress.toParts(e.address),
+          dst: NodeAddress.toParts(e.dst),
+          src: NodeAddress.toParts(e.src),
+          timestampMs: e.timestampMs,
+        };
+      }
       const {repo, data, likes, url} = example();
       const postLikeWeight = {};
       for (const post of repo.posts()) {
@@ -504,24 +525,18 @@ describe("plugins/discourse/createGraph", () => {
       for (const post of repo.posts()) {
         if (postLikeWeight[post.id] > 0) {
           const edge = NE.topicHasLikedPostEdge(url, post);
+          const textSafeEdge = getSafeEdge(edge);
           const weight = postLikeWeight[post.id];
-          expectedTopicHasLikedPosts.push({edge, weight});
+          expectedTopicHasLikedPosts.push({textSafeEdge, weight});
         }
       }
-      expect(data.topicHasLikedPosts).toEqual(expectedTopicHasLikedPosts);
-      expect(expectedTopicHasLikedPosts).toMatchInlineSnapshot(`
-        Array [
-          Object {
-            "edge": Object {
-              "address": "E sourcecred discourse topicHasLikedPost https://url.com 1 3 ",
-              "dst": "N sourcecred discourse post https://url.com 3 ",
-              "src": "N sourcecred discourse topic https://url.com 1 ",
-              "timestampMs": 1,
-            },
-            "weight": 2.25,
-          },
-        ]
-      `);
+      const likedPostsFromTopic = [];
+      for (const {edge, weight} of data.topicHasLikedPosts) {
+        const textSafeEdge = getSafeEdge(edge);
+        likedPostsFromTopic.push({textSafeEdge, weight});
+      }
+      expect(likedPostsFromTopic).toEqual(expectedTopicHasLikedPosts);
+      expect(likedPostsFromTopic.length).toBe(1);
     });
   });
 


topocount:
It looks like the null literals are breaking the diff in github, interestingly.

decentralion (commented):
Since these lines are autogenerated by jest (it's part of `toMatchinlineSnapshot`) this feels like a bandaid fix. If we regenerate with `yarn unit -u` does it re-break?

topocount:
> Since these lines are autogenerated by jest (it's part of `toMatchinlineSnapshot`) this feels like a bandaid fix. If we regenerate with `yarn unit -u` does it re-break?

@decentralion just tested this. No git diffs are generated from this branch when `yarn unit -u` is run.

wchargin (commented):
Thanks for fixing this!

Please remove `toMatchInlineSnapshot` here. It’s good to check that
there are no diffs when you run `yarn unit -u`, but I don’t trust that
to continue being the case. Source files must absolutely not have null
bytes in them.

This kind of thing is why I don’t like `toMatchInlineSnapshot`.

wchargin:
@topocount and I looked into why this was failing on his machine but
working in CI (and on my machine and others’). The test looks like this:

```shell
        grep "^describe\(\.skip\)\?(" -- "${root}/src/${filename}" >describes &&
        grep -F -- "${filename%.test.js}" describes
```

In the usual case, the first `grep` writes a single line to `describes`,
something like:
The first line is supposed to write a single line to `describes`, like:

```
describe("plugins/discourse/createIdentities", () => {
```

But since this file contained binary data, my `grep` would instead
write:

```
Binary file /path/to/sourcecred/src/plugins/discourse/createGraph.test.js matches
```

Now, this happens to match the pattern in the second `grep`, so the test
still silently passed. But @topocount has GNU grep 3.5 (arch btw), which
now [emits that message to stderr instead][cl]:

> ** Changes in behavior
>
> The message that a binary file matches is now sent to standard error
> and the message has been reworded from "Binary file FOO matches" to
> "grep: FOO: binary file matches", to avoid confusion with ordinary
> output or when file names contain spaces and the like, and to be
> more consistent with other diagnostics.

So, for @topocount, `describes` was empty and the second `grep` failed.

The fact that this passed for me (et al.) is really just an accident,
that the error happened to be emitted to stdout and contain the
filename. The right fix is to remove the null bytes so that the test
behaves, and (separately) add a test that our source files are well
formed text files with no null bytes. @topocount wants to do this, so
I look forward to receiving a separate PR. That change doesn’t block
this one.

[cl]: https://savannah.gnu.org/forum/forum.php?forum_id=9820

wchargin (commented):
This is better, but I would prefer that there not be null bytes in the
snapshot file, either. It is still JavaScript source, we still do want
to see diffs for it and view the data with proper delimiters, and it
still shouldn’t be marked as binary.

Glancing at the test, it looks like it’s already far more complicated
than I would expect. It… programmatically constructs the expected value…
and then asserts that the expected value is equal to the snapshot? Why
not just construct the expected value manually? You know, like a normal
test. (You may be picking up on what I don’t like: snapshot tests
incentivize writing bad tests that are hard to read.)

If, for some reason that I’m missing, we really do want the snapshot,
you can emit a snapshotted structure in a form that doesn’t have null
bytes. Something like:

```javascript
const edge = ...;
const textSafeEdge = {
  address: EdgeAddress.toParts(edge.address),
  dst: NodeAddress.toParts(node.dst),
  src: NodeAddress.toParts(node.src),
  timestampMs: node.timestampMs,
};
expectedTopicHasLikedPosts.push({textSafeEdge, weight});
```

…at which point I don’t especially mind the `toMatchInlineSnapshot`.

topocount:
> Glancing at the test, it looks like it’s already far more complicated
> than I would expect. It… programmatically constructs the expected value…
> and then asserts that the expected value is equal to the snapshot? Why
> not just construct the expected value manually? You know, like a normal
> test. (You may be picking up on what I don’t like: snapshot tests
> incentivize writing bad tests that are hard to read.)

@wchargin I considered manually constructing the data to test against, but then I ran into the appeal of snapshots: They make for faster test writing, and I carried on with using a snapshot because I considered the (relative) speed of writing the test to be the appeal, and intent of how the test was constructed. 

That being said, I'm internalizing that in a maintenance lifecycle, reading code is as important (probably more important) as writing code, and if it takes more time to write something that is botjh 1) more maintainable over the longer term (as I believe your example is, since those APIs should be updated if the the Address encoding ever changes) and 2) more readable, then it is worth spending an extra few/several minutes writing the test code.

wchargin:
I mean, look. It seems clear that this particular test *was never read
by either the author of the code or the reviewer*, since the inline
snapshot that it’s supposed to match is garbage. A proper editor renders
the null bytes, whereas the GitHub UI simply omits them entirely (even
separate from the GitHub *diff* view corrupting the patch entirely):

![Screenshot of the test in question in Vim and in GitHub. The Vim view
has addresses like`E^@sourcecred^@discourse^@...`, whereas the GitHub
view has addresses like `Esourcecreddiscourse...`.][ss]

So I would contend that snapshots have made it faster to write
*something*, but certainly not a good test.

> That being said, I'm internalizing that in a maintenance lifecycle,
> reading code is as important (probably more important) as writing
> code, and if it takes more time to write something that is both 1)
> more maintainable over the longer term (as I believe your example is,
> since those APIs should be updated if the the Address encoding ever
> changes) and 2) more readable, then it is worth spending an extra
> few/several minutes writing the test code.

Yes, this is certainly part of it, and is why we explicitly prioritize
ease of review over ease of slamming out PRs.

[ss]: https://user-images.githubusercontent.com/4317806/103677975-d3caa200-4f37-11eb-89fe-9bcdcf162e1a.png

decentralion:
Removed myself as a reviewer, with the theory "most PRs only need one reviewer" and it seems like @wchargin has it covered.

src/plugins/discourse/createGraph.test.js
```diff
@@ -489,6 +496,20 @@ describe("plugins/discourse/createGraph", () => {
       expect(seenTrustLevels).toEqual(new Set([null, 0, 2, 3]));
     });
     it("creates hasLikedPost edges from topics to posts", () => {
+      declare type SafeEdge = {|
```
wchargin, line 499: nit: Can this just be `type` rather than `declare type`?

wchargin (approved):
Lovely! Thanks.

PR #2625, opened 2021-01-06
cli/common: rename loadWeightedGraph to mergePluginWeightedGraphs

elihanover:
This is a simple renaming to make clear which weighted graph we're
loading, and also the fact that we're merging the plugin weighted
graphs into *the* graph.

__Test Plan__
Trivial rename, unit tests pass, manually checked for all potential
import instances.

### DIFF
--- a/src/cli/common.js
+++ b/src/cli/common.js
@@ -91,7 +91,7 @@ export async function prepareCredData(
   dependencies: $ReadOnlyArray<BonusPolicy>,
 |}> {
   const [weightedGraph, ledger] = await Promise.all([
-    await loadWeightedGraph(baseDir, config),
+    await loadAndMergePluginWeigtedGraphs(baseDir, config),
     await loadLedger(baseDir),
   ]);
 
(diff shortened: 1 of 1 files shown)


decentralion (commented):
The name `mergePluginWeightedGraphs` makes me think this is a function that will take as arguments a set of plugin graphs, and return a merged one (which would be consistent w/ all the other merge functions in the codebase). When we are loading data from disk, we consistently call that function a load function.

If you think this function name is unclear, maybe rename it to `loadMergedWeightedGraph` or something, I def think the verb should be "load" and not "merge" though.

elihanover:
Good point.  I renamed to `loadMergedPluginWeigtedGraphs` from the original `loadWeightedGraph`.

Wordy, yes, but it's important to indicate that merging is occurring, and that the plugin graphs are the graphs being merged.

elihanover:
@decentralion any idea why full ci tests are failing.  Not quite sure how to interpret log.

decentralion (approved):
Sharness failure is due to some unrelated Discourse stuff--I suspect it may have been unlucky rebase timing, I think we fixed that up for master. Here's the relevant log:

```
    expecting success: 
        diff -qr . "$snapshot_directory"
    Files ./output/accounts.json and /home/circleci/repo/sharness/__snapshots__/test-instance/output/accounts.json differ
    Files ./output/credResult.json and /home/circleci/repo/sharness/__snapshots__/test-instance/output/credResult.json differ
    Files ./output/graphs/sourcecred/discourse/graph.json and /home/circleci/repo/sharness/__snapshots__/test-instance/output/graphs/sourcecred/discourse/graph.json differ
    not ok 4 - should be identical to the snapshot
    #	
    #	    diff -qr . "$snapshot_directory"
    #	
    # failed 1 among 4 test(s)
```

PR #2627, opened 2021-01-06
credrank: Add intervals getter to CredGraph, rename credPerEpoch to c…

blueridger:
# Description
This adds a passthrough `intervals()` getter on the CredGraph, which will allow consumers like the CredGrainView to access the intervals, which is important for contextualizing the Participant credPerInterval array in time.

It also renames the credPerEpoch array to credPerInterval. They are ultimately the same, but I think credPerInterval is more semantically intuitive, especially now that we are exposing the intervals and not the epochs.
"How much cred is associated with a given time period" instead of "how much cred is associated with a startTime timestamp"
In my opinion, the idea of epochs should not be exposed outside of the CredGraph internal logic, in preference of intervals.

# Test Plan
Added a basic unit test.
CI passes.

### DIFF
--- a/src/core/credrank/credGraph.js
+++ b/src/core/credrank/credGraph.js
@@ -11,6 +11,7 @@ import {
 } from "./markovProcessGraph";
 import {type MarkovEdge, type TransitionProbability} from "./markovEdge";
 import {payoutGadget} from "./edgeGadgets";
+import {type IntervalSequence} from "../interval";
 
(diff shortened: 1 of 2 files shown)


decentralion (approved):
> In my opinion, the idea of epochs should not be exposed outside of the CredGraph internal logic, in preference of intervals.

Agreed -- feel free to refactor/rename APIs liberally.

PR #2628, opened 2021-01-07
Allow front end to function without Cred

blueridger:
# Description
This change allows the front end to load successfully even if there is no `output` directory. In such a case, it simply displays the following message on the Explorer page while still allowing all other pages to be used functionally, and it uses the Accounts page as the root page instead of the Explorer page.
```
This page is unavailable because Cred information was unable to load. Calculate cred through the CLI in order to use this page.
```
By not requiring cred calculation for most of the pages, instance maintainers can use the admin portal on a fresh clone of an instance, without having to worry about doing a mock `go` or copying in files from the gh-pages branch. This will make it MUCH easier to make make identity changes and transfer grain, as well as verify such changes during PR review.

# Test Plan
## Manual Testing:
1. Checkout `master` branch in `cred` package
2. From sourcecred package, `yarn start --instance /Path/to/instance/cred`
3. Verify that the Accounts page loads by default.
4. Verify that the Explorer page shows only the new info message.
5. Verify that other pages are viewable/functional.
6. Verify Transfer Grain, Merge Identity, Activate Identity actions work fully.
## Manual Regression Testing:
1. Checkout `gh-pages` branch in `cred` package
2. From sourcecred package, `yarn start --instance /Path/to/instance/cred`
4. Verify that the Explorer page is loads by default.
5. Verify weight configuration can be recalculated.
6. Verify other pages work.

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -91,10 +91,11 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 type ExplorerHomeProps = {|
-  +initialView: CredView,
+  +initialView: CredView | null,
 |};
 
 export const ExplorerHome = ({initialView}: ExplorerHomeProps): ReactNode => {
+  if (!initialView) return null;
   const classes = useStyles();
   const [tab, setTab] = useState<number>(1);
   const [checkboxes, setCheckboxes] = useState({
--- a/src/ui/load.js
+++ b/src/ui/load.js
@@ -11,7 +11,7 @@ import {
 export type LoadResult = LoadSuccess | LoadFailure;
 export type LoadSuccess = {|
   +type: "SUCCESS",
-  +credView: CredView,
+  +credView: CredView | null,
   +ledger: Ledger,
   +bundledPlugins: $ReadOnlyArray<pluginId.PluginId>,
   +hasBackend: Boolean,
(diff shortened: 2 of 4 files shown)


src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -91,10 +91,11 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 type ExplorerHomeProps = {|
-  +initialView: CredView,
+  +initialView: CredView | null,
 |};
 
 export const ExplorerHome = ({initialView}: ExplorerHomeProps): ReactNode => {
+  if (!initialView) return null;
```
hammadj, line 98: You can also render a Redirect component here in case someone ends up on this page with no cred view

hammadj (approved):

src/ui/components/ExplorerHome/ExplorerHome.js
blueridger: This is a backend WIP page, and this line will be changed before the launch since the page will not use a CredView object.

src/ui/load.js
```diff
@@ -33,16 +33,19 @@ export async function load(): Promise<LoadResult> {
...
   try {
-    const json = await responses[0].json();
-    const credResult = credResultFromJSON(json);
-    const credView = new CredView(credResult);
+    let credView = null;
+    if (responses[0].ok) {
+      const json = await responses[0].json();
+      const credResult = credResultFromJSON(json);
+      credView = new CredView(credResult);
+    }
```
topocount, line 48: Definitely out of scope for this PR, but I'm looking forward to refactoring response parsing sequence into something more readable.

topocount (commented):
Thanks @blueridger ! Super useful patch

PR #2631, opened 2021-01-07
Write a design doc for the migration to Credrank

decentralion:
This is a design doc describing how we'll migrate to CredRank.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038637804811) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/design-docs/2021/q1/credrank-migration.md
+++ b/design-docs/2021/q1/credrank-migration.md
@@ -2,11 +2,12 @@
 
 ## Created: 2021.01.07
 
-## Last Updated: 2021.01.08
+## Last Updated: 2021.01.25
 
 ## Authors:
 
 - @decentralion
+- @blueridger
 
 ## Overview
 
@@ -17,9 +18,9 @@ and to remove [Timeline Cred].
 [credrank]: https://discourse.sourcecred.io/t/credrank-scalable-interpretable-flexible-attribution/654
 [timeline cred]: https://github.com/sourcecred/sourcecred/pull/1212
 
-## Steps
+## Implementation Steps
 
-### Finish Implementing CredGrainView
+### 1) Finish Implementing CredGrainView [DONE]
 
 The `CredGrainView` (see src/core/credGrainView.js) is a class that makes it
 convenient to retrieve Cred (and Grain) data about any participant. It will act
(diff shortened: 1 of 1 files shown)


design-docs/2021/q1/credrank-migration.md
```diff
@@ -0,0 +1,97 @@
...
+Currently, CGV exists as an interface, but most of the methods are not
+implemented and just throw errors. It needs to be implemented and tested.
+
+### Create a CredRank-compatible Cred Explorer
+
+Presently, the Explorer (src/ui/components/Explorer/Explorer.js) depends on the
+`CredView` class, which means it uses TimelineCred scores. We should create a
+new overview page that depends on `CredGrainView` instead.
+
+Less ambitiously, we could just port the existing UI over to use `CredView`.
```
blueridger, line 37: Worth noting that this will likely regress the top contributions dropdown feature.

design-docs/2021/q1/credrank-migration.md
```diff
@@ -0,0 +1,97 @@
...
+Once the new Grain command is available, it should be wired in to the CLI for
+testing, but not actively used until the migration.
+
+### Migrate
+
+Once we have a replacement UI and Grain command that use the `CredGrainView`,
+we will be ready to atomically migrate over to the new algorithm. This
+migration will have three steps:
+
+1. In src/cli/go.js, replace the `score` command with the `credrank` command in
```
blueridger, line 67: nit, I would probably choose to rename `credrank` to `score` and rename `score` to `timeline-cred` (or remove the timeline-cred CLI) to avoid UX churn.

blueridger (approved):
Approved with comments

design-docs/2021/q1/credrank-migration.md
topocount: I think it might be even better to just expose `credrank` and `timeline-cred` (or `timelineCred`?) in the lower-level CLI explicitly.

This wouldn't necessarily induce much churn for users at least, since these commands aren't exposed by name in the instance scripts. It might be a bit of change for developers but we could alias `score` to `credrank` if that's indeed a concern.

design-docs/2021/q1/credrank-migration.md
```diff
@@ -0,0 +1,97 @@
+# CredRank Migration Design Doc
+
+## Last Updated: 2021.01.07
```
topocount, line 3: nit: worth adding a "Created" field?

topocount (approved):

design-docs/2021/q1/credrank-migration.md
decentralion: Done. @topocount I think when we do the migration, we will regress the top contributions dropdown until we do a UI redesign based on @KuraFire's work. And I think that's OK b.c. the current feature is nearly useless due to its lack of context.

design-docs/2021/q1/credrank-migration.md
decentralion: I like credrank better than score. Mostly because I think the CredRank algorithm is really cool, so I'd rather surface that cool name than the more generic and boring "score".

As @topocount pointed out, this won't induce much churn for users because we're telling users to use `sourcecred go` rather than run lower-level commands directly; the lower-level commands become more of a core dev affordance.
 However, there are some users out there that still have pointers to `score`, e.g. [1hive/pollen](https://github.com/1Hive/pollen). So we should either help them upgrade or leave score as an alias.

decentralion: Oh, and there's no reason to keep `timeline-cred` as a command; we're killing timeline cred with fire 🔥

design-docs/2021/q1/credrank-migration.md
decentralion: Done.

design-docs/2021/q1/credrank-migration.md
KuraFire: +1, I would even be fine if that feature is simply taken out until the new UI redesign for Participants view is done. 😂

KuraFire (approved):

PR #2634, opened 2021-01-08
Utilize ES6 concat syntax instead of ES2019 flat method

topocount:
Node V10 does not support ES2019 Array.prototype.flat() syntax. This
causes issues with unit tests. Since @wchargin and sometimes I utilize
Node V10 I've created this PR.

While the build succeeds without this change still, unit tests with jest fail.

test plan: nvm use 10 && yarn unit # assumes the use of nvm to access node V10


decentralion (commented):
I don't object to this change, but I would also support dropping node v10. It's EOL in less than 5 months.
https://nodejs.org/en/about/releases/

wchargin (approved):
I also don’t feel super strongly since it’s near EOL, but I like
supporting Node 10 while it’s supported. I personally use Node 10.

PR #2636, opened 2021-01-08
util: Add findLast and findLastIndex util functions

blueridger:
# Description
I found it annoying that JS doesn't provide reverse-order versions of Array.find and Array.findIndex so I wrote my own. Without this method, there is a trade off between readability (slice.reverse.find) and efficiency (index-based loop). Now, these can be used to provide the efficiency of an index loop with better readability.

These do not technically have all of the functionality of Array.find and Array.findIndex (namely the optional callback params), but they can be updated to include that later if needed. Most cases do not need it.

# Test Plan
Unit tests added for the util methods.
The `_findCurrentBudgetValue` example is well-covered by unit tests.

### DIFF
--- a/src/util/findLast.js
+++ b/src/util/findLast.js
@@ -1,0 +1,25 @@
+// @flow
+
+// This is simply a reverse-order version of Array.prototype.find()
+// Returns the last element in the array that satisfies the provided
+// testing function.
+// Otherwise, it returns undefined, indicating that no element passed the test.
+export function findLast<T>(
+  array: $ReadOnlyArray<T>,
+  testingFunction: (obj: T) => boolean
+): ?T {
+  return array[findLastIndex(array, testingFunction)];
+}
+
+// This is simply a reverse-order version of Array.prototype.findIndex()
+// Returns the index of the last element in the array that satisfies the
+// provided testing function.
+// Otherwise, it returns -1, indicating that no element passed the test.
+export function findLastIndex<T>(
+  array: $ReadOnlyArray<T>,
+  testingFunction: (obj: T) => boolean
+): number {
+  let index = array.length - 1;
+  while (index >= 0 && !testingFunction(array[index])) index--;
+  return index;
+}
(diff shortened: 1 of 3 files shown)


src/util/findLast.js
```diff
@@ -0,0 +1,28 @@
...
+// Returns the last element in the array that satisfies the provided
+// testing function.
+// Otherwise, it returns undefined, indicating that no element passed the test.
+export function findLast<T>(
+  array: $ReadOnlyArray<T>,
+  testingFunction: (obj: T) => boolean
+): ?T {
+  let index = array.length - 1;
+  while (index >= 0 && !testingFunction(array[index])) index--;
+  if (index >= 0) return array[index];
```
hammadj, line 13: Would be nice if findLast composed findLastIndex

src/util/findLast.test.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+
+import {findLast, findLastIndex} from "./findLast";
+
+describe("util/findLast", () => {
```
hammadj, line 5: Would be nice to test out the empty case here as well (empty array)

hammadj (approved):
Looks good, couple minor suggestions/nits

src/util/findLast.js
BrianBelhumeur: Concurred

blueridger: Accepted and updated.

src/util/findLast.test.js
blueridger: Accepted and added.

wchargin:
Is there a reason that we don’t like Lodash’s functions of the same names?

  - <https://lodash.com/docs/4.17.15#findLast>
  - <https://lodash.com/docs/4.17.15#findLastIndex>

blueridger:
Oh shoot... should I revert this?

wchargin:
No strong preference, since we depend on the Lodash modules individually
(and thus this would be a dep update), and Lodash functions often have
some kinds of magical edge cases that can be confusing (haven’t looked
into these two closely). Was mostly just wondering if you’d considered
using them, since I didn’t see it mentioned in the PR description.

PR #2637, opened 2021-01-09
Implement CredGrainView

blueridger:
# Description
This is an implementation for the CredGrainView, which is responsible for aligning cred and grain data in time using intervals. It will be used initially by the Explorer UI and the credrank-based Grain distributions.

# Design decisions
I decided to do most of the processing in the CredGrainView instead of the TimeScopedCredGrainView. This should be more efficient for high-frequency time scoping, as long as time scoping the processed data is faster than time scoping graph/ledger data, which it is for now. This may become more controversial if we add a bunch more data caches to this class, but I hope we will not in favor of creating more atomic classes.

I decided to remove the filter and sort options from the originally sketched interface, as most of that is probably better done downstream. If there is sorting or filtering that we need in the future that would be more appropriate to have in this module (such as filtering by allocation type), we can add that in later as needed.

# Test Plan
I added super thorough unit tests. They cover the grain logic extremely well, however the cred logic is tested only somewhat well. It only tests against a credGraph in which both participants have cred in both intervals. I accept this because the cred logic is far less complex and more expensive to create test data for.

No manual testing. (manual testing will come in future work using this module)

### DIFF
--- a/src/core/credGrainView.js
+++ b/src/core/credGrainView.js
@@ -7,40 +7,40 @@
  * It is useful for cases where you want to view a participant's Cred and Grain
  * data simultaneously, for example for creating summary dashboards.
  */
-import {CredGraph} from "./credrank/credGraph";
-import {type Identity, type IdentityType, type IdentityId} from "./identity";
+import deepFreeze from "deep-freeze";
+import {
+  CredGraph,
+  type Participant as GraphParticipant,
+} from "./credrank/credGraph";
+import {type Identity, type IdentityId} from "./identity";
+import * as G from "./ledger/grain";
 import {type Grain} from "./ledger/grain";
-import {Ledger} from "./ledger/ledger";
+import {Ledger, type Account} from "./ledger/ledger";
 import {type TimestampMs} from "../util/timestamp";
-import {type IntervalSequence} from "./interval";
+import findLastIndex from "lodash.findlastindex";
+import {type IntervalSequence, intervalSequence} from "./interval";
 
 /**
  * Cred and Grain data for a given participant.
  *
  * Implicitly has an associated time scope, which will
- * be the time scope of the TimeScopedCredGrainView that generated
- * this.
+ * be the time scope of the CredGrainView or TimeScopedCredGrainView that
+ * generated this.
+ *
+ * The indices of credPerInterval/grainEarnedPerInterval correspond to the
+ * same indices in the IntervalSequence of the CredGrainView or
+ * TimeScopedCredGrainView that generated this.
  */
 export type ParticipantCredGrain = {|
   +identity: Identity,
   // Total Cred earned during the time scope.
   +cred: number,
   // Cred earned in each interval within the time scope.
-  +credOverTime: $ReadOnlyArray<number>,
+  +credPerInterval: $ReadOnlyArray<number>,
   // Total Grain earned during the time scope.
   +grainEarned: Grain,
   // Grain earned in each interval within the time scope.
-  +grainEarnedOverTime: $ReadOnlyArray<Grain>,
-|};
-
-export type ParticipantSort = "BY_CRED" | "BY_GRAIN";
-export type ParticipantsOptions = {|
-  // How to sort. Always sorts from highest to lowest.
-  // Defaults to BY_CRED
-  +sort?: ParticipantSort,
-  // Which identity types should be included.
-  // Defaults to including all types.
-  +includedTypes?: Set<IdentityType>,
+  +grainEarnedPerInterval: $ReadOnlyArray<?Grain>,
 |};
 
 /**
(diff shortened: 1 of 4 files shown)


src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+
+    const graphParticipants = new Map<IdentityId, GraphParticipant>();
+    for (const participant of credGraph.participants()) {
+      graphParticipants.set(participant.id, participant);
+    }
+    for (const account of ledger.accounts()) {
+      const graphParticipant = graphParticipants.get(account.identity.id);
+      if (!graphParticipant)
+        throw new Error(
+          "The graph is missing an account that exists in the ledger."
```
decentralion, line 80: would be helpful to include the id and identity name for the missing identity.

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+        );
+
+      this._participants.push({
+        identity: account.identity,
+        cred: graphParticipant.cred,
+        credPerInterval: graphParticipant.credPerInterval,
+        grainEarned: account.paid,
+        grainEarnedPerInterval: this._calculateGrainEarnedPerInterval(account),
+      });
```
decentralion, line 89: rather than a push-based approach, how about a map?

`this._participants = ledger.accounts().map(...`

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+    }
+  }
+
+  _calculateGrainEarnedPerInterval(account: Account): $ReadOnlyArray<Grain> {
+    const grainEarnedPerInterval = [];
+    for (const interval of this._intervals) {
+      const grainEarnedThisInterval =
+        account.allocationHistory.reduce((grain, allocationReceipt) => {
+          if (!grain) grain = G.ZERO;
```
decentralion, line 98: I prefer providing an initial value of G.ZERO to the reduce fn, rather than implicitly giving an initial value of undefined and then checking here. It's cleaner.

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+          if (
+            interval.startTimeMs < allocationReceipt.credTimestampMs &&
+            allocationReceipt.credTimestampMs <= interval.endTimeMs
+          )
+            grain = G.add(grain, allocationReceipt.grainReceipt.amount);
+        }) || G.ZERO;
```
decentralion, line 104: If you provide an initial value to reduce, you won't need `|| G.ZERO` here.

Per [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce):
`arr.reduce(callback( accumulator, currentValue, [, index[, array]] )[, initialValue])`

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+      grainEarnedPerInterval.push(grainEarnedThisInterval);
```
decentralion, line 105: Rather than using push, I recommend just setting `grainEarnedPerInterval = this._intervals.map(...`

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+      this._participants,
+      this._intervals,
       startTimeMs,
       endTimeMs
     );
   }
 
   intervals(): IntervalSequence {
-    return this._defaultTimeScope.intervals();
+    return cloneDeep(this._intervals);
```
decentralion, line 123: Interesting that we clone the IntervalSequence but not the participants. If we're worried about a bad caller mutating things, wouldn't we also be worried that they may mutate the participant array?

I would probably prefer deepFreeze to cloneDeep but don't feel strongly.

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+        inclusiveStartIndex,
+        exclusiveEndIndex
+      );
+      const grainEarnedPerInterval = participant.grainEarnedPerInterval.slice(
+        inclusiveStartIndex,
+        exclusiveEndIndex
+      );
+      return {
+        identity: participant.identity,
+        cred: credPerInterval.reduce((a, b) => a + b),
```
decentralion, line 172: Reduce should have an initial value of `0`. Otherwise it will error when the array is empty:

```js
> [].reduce((a,b) => a + b)
Uncaught TypeError: Reduce of empty array with no initial value
```

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+        credPerInterval,
+        grainEarned: grainEarnedPerInterval.reduce((a, b) => G.add(a, b)),
```
decentralion, line 174: Reduce should have an initial value of `G.ZERO`

src/core/credGrainView.js
```diff
@@ -54,77 +60,133 @@ export type ParticipantsOptions = {|
...
+    });
   }
 
-  participants(
-    options: ParticipantsOptions
-  ): $ReadOnlyArray<ParticipantCredGrain> {
-    const _ = options;
-    throw new Error("not yet implemented");
+  intervals(): IntervalSequence {
+    return cloneDeep(this._intervals);
```
decentralion, line 181: Same question as before for why we clone the intervals but not the participants.
Note if we were to be consistent (clone the participants too) that would make these APIs un-performant, which suggests to me we should take a different approach, e.g. using freezes.

decentralion (commented):
Seems reasonable overall.

src/core/credGrainView.js
blueridger: I can add that, but flow will still want this line regardless.

src/core/credGrainView.js
blueridger: Flow won't let me remove it.

src/core/credGrainView.js
blueridger: sure

src/core/credGrainView.js
blueridger: accepted

src/core/credGrainView.js
blueridger: yeah, probably overkill

src/core/credGrainView.js
blueridger: accepted

src/core/credGrainView.js
blueridger: accepted

src/core/credGrainView.js
blueridger: sure, will use freeze

src/core/credGrainView.js
blueridger: accepted

src/core/credGrainView.js
```diff
@@ -54,77 +54,137 @@ export type ParticipantsOptions = {|
...
-      -Infinity,
-      Infinity
+    this._intervals = deepFreeze(credGraph.intervals());
+
+    const graphParticipants = new Map<IdentityId, GraphParticipant>();
+    for (const participant of credGraph.participants()) {
+      graphParticipants.set(participant.id, participant);
+    }
+
+    this._participants = deepFreeze(
```
decentralion, line 70: hmmm. this technically mutates the input, since it deepFreezes otherwise-mutable objects like `graphParticipant.credPerInterval`. I think it's fine because those are read-only anyway, but just flagging it.

src/core/credGrainView.js
```diff
@@ -54,77 +54,137 @@ export type ParticipantsOptions = {|
...
+      let grain = G.ZERO;
+      account.allocationHistory.forEach((allocationReceipt) => {
+        if (
+          interval.startTimeMs < allocationReceipt.credTimestampMs &&
+          allocationReceipt.credTimestampMs <= interval.endTimeMs
+        )
+          grain = G.add(grain, allocationReceipt.grainReceipt.amount);
+      });
+      return grain;
+    });
```
decentralion, line 102: Let `A` be number of allocations and `I` be number of intervals. Then this runs in `O(A*I)` time but could run in `O(I+A)` time if we leverage the fact that allocation history is in time-sorted order. Ie., we don't actually need to iterate over every allocation for every interval.

Since we run this for every participant (`P`), we wind up with something like `O(P*I*A)` instead of `O(P * (I+A))`, inside the constructor. Having expensive constructors is kind of an anti-pattern already, so I think it might be worth pre-emptively optimizing this. It should be pretty straightforward to do so.

I'm going to approve this because I think the code is correct as written and we can optimize it in another PR (and maybe add some safety checks, like runtime verifying that the allocations are, in fact, in chronological order).

src/core/credGrainView.js
```diff
@@ -54,77 +54,137 @@ export type ParticipantsOptions = {|
...
+      findLastIndex(
+        originalIntervals,
+        (interval) => interval.endTimeMs <= endTimeMs
+      ) + 1;
 
-  participant(id: IdentityId): ParticipantCredGrain | null {
-    const _ = id;
-    throw new Error("not yet implemented");
+    this._intervals = deepFreeze(
+      intervalSequence(
```
decentralion, line 155: Nit: we already froze the original interval sequence in the outer constructor, so we could just do `Object.freeze` and get the same level of safety here. Likely too clever since it might confuse future readers, and the performance implications are negligible. :)

src/core/credGrainView.test.js
```diff
@@ -0,0 +1,405 @@
+// @flow
+
+import {CredGrainView} from "./credGrainView";
+import * as GraphUtil from "./credrank/testUtils";
+import * as LedgerUtil from "./ledger/testUtils";
+import {g} from "./ledger/testUtils";
+import * as uuid from "../util/uuid";
+
+describe("core/credGrainView", () => {
```
decentralion, line 9: please add a test for the case where the graph has a participant that's missing in the ledger

decentralion (approved):
approved-with-nits (mostly, please add a missing test)

src/core/credGrainView.js
blueridger: I thought about that too, but wasn't sure if it was a worthwhile optimization. Sounds like yes, as a fast follow.

src/core/credGrainView.test.js
blueridger: accepted

Issue #2455, opened 2020-11-04
Discourse: Category / tag based minting


decentralion:
Request from Maker community: they'd like to be able to set custom mint rules based on Discourse categories and tags:

- Amount to mint per topic in (category or tag)
- Mint multiplier per like in (category or tag)

Let's do this by EOY.

PR #2642, opened 2021-01-09
Discourse: Refactor like weight logic

topocount:
In preparation for adding more weight calculations for likes, weight
calculations are moved to their own file. This builds towards the
completion of #2455.

n.b. This is similar to how the experimental discord reaction weight
calculations are organized.

test plan: yarn unit

### DIFF
--- a/src/plugins/discourse/weights.js
+++ b/src/plugins/discourse/weights.js
@@ -1,0 +1,30 @@
+// @flow
+
+import {type NodeWeight} from "../../core/weights";
+import {DEFAULT_TRUST_LEVEL_TO_WEIGHT} from "./createGraph";
+import {type User} from "./fetch";
+
+export function likeWeight(user: ?User): NodeWeight {
+  if (user == null) {
+    return 0;
+  }
+  return _trustLevelWeight(user.trustLevel);
+}
+
+export function _trustLevelWeight(trustLevel: number | null): NodeWeight {
+  if (trustLevel == null) {
+    // The null trust level shouldn't happen in practice, right now users who
+    // only like but never post will have null trust level (will be fixed by #2045).
+    // This means they could have trust level 1. But to be conservative, we treat anyone
+    // with a null trust level as if they have trust level 0.
+    // Possibly this could come up with deleted users too.
+    return DEFAULT_TRUST_LEVEL_TO_WEIGHT["0"];
+  }
+
+  const key = String(trustLevel);
+  const weight = DEFAULT_TRUST_LEVEL_TO_WEIGHT[key];
+  if (weight == null) {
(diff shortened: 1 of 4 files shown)


src/plugins/discourse/weights.js
```diff
@@ -0,0 +1,30 @@
...
+import {type User} from "./fetch";
+
+export function likeWeight(user: ?User): NodeWeight {
+  if (user == null) {
+    return 0;
+  }
+  return _trustLevelWeight(user);
+}
+
+export function _trustLevelWeight({trustLevel}: User): NodeWeight {
```
decentralion, line 14: I think it'd be cleaner to just pass in the trust level directly, rather than the user object. It would match the function name too, as in `return _trustLevelWeight(user.trustLevel)`. If it's a weight for a trust level, shouldn't it take a trust level?

src/plugins/discourse/weights.js
```diff
@@ -0,0 +1,30 @@
...
+  if (trustLevel == null) {
+    // The null trust level shouldn't happen in practice, right now users who
+    // only like but never post will have null trust level (will be fixed by #2045).
+    // This means they could have trust level 1. But to be conservative, we treat anyone
+    // with a null trust level as if they have trust level 0.
+    // Possibly this could come up with deleted users too.
+    return 0;
```
decentralion, line 21: Hmm, the comment and impl are inconsistent; it says we treat someone with null trust level as though they have TL0, but actually we return hardcoded 0, even if the weight for TL0 is something else (e.g. 0.1).

I know it's not a new bug in this PR, but maybe fix while you're touching this code? Or just add a comment, it would be hard to test w/o making the trust level dict customizable.

src/plugins/discourse/weights.test.js
```diff
@@ -0,0 +1,34 @@
...
+    it("throws an error for an invalid trustLevel", () => {
+      const thunk = () => _trustLevelWeight(getMockUser(-1));
+      expect(thunk).toThrowError("invalid trust level");
+    });
+    it("works as expected for a regular user", () => {
+      expect(_trustLevelWeight(getMockUser(0))).toEqual(0);
+      expect(_trustLevelWeight(getMockUser(1))).toEqual(0.1);
+      expect(_trustLevelWeight(getMockUser(2))).toEqual(1);
+      expect(_trustLevelWeight(getMockUser(3))).toEqual(1.25);
+      expect(_trustLevelWeight(getMockUser(4))).toEqual(1.5);
```
decentralion, line 31: We should test for the case where the trust level is null, too.

decentralion (approved):
I don't think my review feedback got addressed. I'm going to approve-with-nits since it's a minor change that I"m suggesting.

PR #2643, opened 2021-01-10
Add tag and category parsing to DiscourseConfig

topocount:
These configs are unopinionated, in that they may be partially
shaped (only containing a default weight, or configured weights), and
the parser will resolve to the defaults if anything is missing from the 
config file.

builds towards #2455
paired with @decentralion 
test plan: yarn unit

### DIFF
--- a/src/plugins/discourse/config.js
+++ b/src/plugins/discourse/config.js
@@ -3,54 +3,21 @@
 import * as Combo from "../../util/combo";
 import {optionsShapeParser, type MirrorOptions} from "./mirror";
 import {
-  type TagConfig,
-  type CategoryConfig,
-  tagConfigParser,
-  categoryConfigParser,
+  type WeightsConfig,
+  weightsConfigParser,
+  type SerializedWeightsConfig,
 } from "./weights";
+
+export type SerializedDiscourseConfig = {|
+  +serverUrl: string,
+  +mirrorOptions?: $Shape<MirrorOptions>,
+  +weights: SerializedWeightsConfig,
+|};
 
 export type DiscourseConfig = {|
   +serverUrl: string,
   +mirrorOptions?: $Shape<MirrorOptions>,
-  /**
-   * Categories can be configured to confer specific like-weight multipliers
-   * when added to a Topic.
-   * If a category does not have a configured weight, the deafaultWeight is applied.
-   * An example configuration might look like:
-   * "categoryWeights": {
-   *   "defaultWeight": 1,
-   *   "weights": {
-   *     "5": 0,
-   *     "36": 1.25
-   *   }
-   * }
-   * where "5" and "36" are the categoryIds in discourse.
-   *
-   * An easy way to find the categoryId for a given category is to browse to the
-   * categories section in discourse
-   * (e.g. https://discourse.sourcecred.io/categories).
-   * Then mousing over or clicking on a category will bring you to a url that
-   * has the shape https://exampleUrl.com/c/<category name>/<categoryId>
-   * Clicking on the community category in sourcecred navigates to
-   * https://discourse.sourcecred.io/c/community/26 for example, where the
-   * categoryId is 26
-   */
-  +categoryWeights?: CategoryConfig,
-  /**
-   * Tags can be configured to confer specific like-weight multipliers when
-   * added to a Topic.
-   * If a tag does not have a configured weight, the defaultWeight is applied.
-   * An example configuration might look like:
-   * "tagWeights": {
-   *   "defaultWeight": 1,
-   *   "weights": {
-   *     "foo": 0,
-   *     "bar": 1.25
-   *   }
-   * }
-   * where foo and bar are the names of tags used in discourse.
-   */
-  +tagWeights?: TagConfig,
+  +weights: WeightsConfig,
 |};
 
 export const parser: Combo.Parser<DiscourseConfig> = (() => {
@@ -69,8 +36,7 @@ export const parser: Combo.Parser<DiscourseConfig> = (() => {
     },
     {
       mirrorOptions: optionsShapeParser,
-      tagWeights: tagConfigParser,
-      categoryWeights: categoryConfigParser,
+      weights: weightsConfigParser,
     }
   );
 })();
--- a/src/plugins/discourse/config.js
+++ b/src/plugins/discourse/config.js
@@ -2,10 +2,55 @@
 
 import * as Combo from "../../util/combo";
 import {optionsShapeParser, type MirrorOptions} from "./mirror";
+import {
+  type TagConfig,
+  type CategoryConfig,
+  tagConfigParser,
+  categoryConfigParser,
+} from "./weights";
 
 export type DiscourseConfig = {|
   +serverUrl: string,
   +mirrorOptions?: $Shape<MirrorOptions>,
+  /**
+   * Categories can be configured to confer specific like-weight multipliers
+   * when added to a Topic.
+   * If a category does not have a configured weight, the deafaultWeight is applied.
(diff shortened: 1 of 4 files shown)


src/plugins/discourse/weights.js
```diff
@@ -1,9 +1,75 @@
...
+      weights: C.dict(C.number),
+    }
+  ),
+  upgrade
+);
+
+// categoryIDs should all be numbers, so the parser verifies that it is
+// indeed a number, then return a string since it is a dict key
+export function parseCategoryId(id: string): string {
+  const result = parseInt(id);
```
wchargin, line 55: `parseInt(id, 10)`, else strings like `"0x123"` are accepted.

src/plugins/discourse/weights.js
```diff
@@ -1,9 +1,75 @@
...
+      defaultWeight: C.number,
+      weights: C.dict(C.number),
+    }
+  ),
+  upgrade
+);
+
+// categoryIDs should all be numbers, so the parser verifies that it is
+// indeed a number, then return a string since it is a dict key
+export function parseCategoryId(id: string): string {
```
wchargin, line 54: Why does this need to be exported as part of the public API?

src/plugins/discourse/weights.js
```diff
@@ -1,9 +1,75 @@
...
+  +defaultWeight: NodeWeight,
+  +weights: {
+    [string]: NodeWeight,
+  },
+|};
+
+export type CategoryConfig = Config;
+export type TagConfig = Config;
+
+export const DEFAULT_CONFIG: Config = {
```
wchargin, line 26: Why are `DEFAULT_CONFIG`, `upgrade`, and `parseCategoryId` exported?
They all seem like implementation details, and `DEFAULT_CONFIG` could be
mutated by poorly behaved clients.

src/plugins/discourse/weights.js
```diff
@@ -1,9 +1,75 @@
...
+};
+
+export function upgrade(c: SerializedConfig): Config {
+  return {
+    defaultWeight: NullUtil.orElse(
+      c.defaultWeight,
+      DEFAULT_CONFIG.defaultWeight
+    ),
+    weights: NullUtil.orElse(c.weights, DEFAULT_CONFIG.weights),
+  };
```
wchargin, line 38: Perhaps just

```javascript
function upgrade(c: SerializedConfig): Config {
  return {
    defaultWeight: 1,
    weights: {},
    ...config,
  }
}
```

? More readable since it’s shorter, only cares about presence of fields
rather than nullity/undefinedness, and makes it easy to inline the
`DEFAULT_CONFIG` object entirely.

src/plugins/discourse/weights.js
```diff
@@ -1,9 +1,75 @@
...
+|};
+
+export type Config = {|
+  +defaultWeight: NodeWeight,
+  +weights: {
+    [string]: NodeWeight,
+  },
+|};
+
+export type CategoryConfig = Config;
```
wchargin, line 23: The `categoryConfigParser` ensures that the parsed keys represent
numbers, but that’s not documented on `CategoryConfig` at all. Is that
meant to be a type-level invariant (it’s not, currently), a property of
the parser, or something else?

topocount: @wchargin It's just meant to be a property of the parser. I'd like it to be a type-level invariant but I ran into a conflict with the `dict` parser where all keys must ultimately be strings, which makes sense.

I realize these types completely lack documentation right now. I was going to go back and document the types after I got everything working end-to-end, but I appreciate your pre-review nonetheless.

src/plugins/discourse/weights.js
topocount: I was expecting to use them in tests in later PRs.

src/plugins/discourse/weights.js
```diff
@@ -1,9 +1,75 @@
...
+  +weights?: {
+    [string]: NodeWeight,
+  },
+|};
+
+export type Config = {|
+  +defaultWeight: NodeWeight,
+  +weights: {
+    [string]: NodeWeight,
+  },
```
decentralion, line 20: I like the distinction between serialized and non-serialized configs. For the non-serialized configs, I recommend ES6 maps over object dicts. Pretty much the only reason we ever use dicts is because maps don't serialize.

src/plugins/discourse/weights.js
decentralion: If we use maps for the configs per my above suggestion, I believe this stops being a problem. The parser can correctly parse the dict keys as numbers where appropriate.

topocount:
I'd love to deduplicate my code in this most recent push more, but I found it really hard to work with the the functional object.entries array in a type-safe way. This is what I was trying to implement to some degree:

```js
function categoryMapFromDict(w: WeightConfig): CategoryWeights {
  const mapWeights = Object.entries(w).map(([key, weight]) => [
    parseInt(key, 10),
    // needed to satisfy flow here, or it defaults to "mixed" type
    parseInt(weight, 10),
  ]);
  return new Map(mapWeights);
}

function tagMapFromDict(w: WeightConfig): TagWeights {
  const mapWeights = Object.entries(w).map(([key, weight]) => [
    key,
    // needed to satisfy flow here, or it defaults to "mixed" type
    parseInt(weight, 10),
  ]);
  return new Map(mapWeights);
}

function upgrade(isCategory: boolean) {
  return (c: SerializedConfig): Config => {
    if (isCategory) {
      return ({
        defaultWeight: NullUtil.orElse(c.defaultWeight, 1),
        weights: c.weights ? categoryMapFromDict(c.weights) : new Map(),
      }: CategoryConfig);
    }
    return ({
      defaultWeight: NullUtil.orElse(c.defaultWeight, 1),
      weights: c.weights ? tagMapFromDict(c.weights) : new Map(),
    }: TagConfig);
  };
}
```

wchargin:
I’m not sure what exactly you’re saying the problem is (are you running
into a Flow error? what does it say?), but at a glance:

  - type-annotating your tuples may help, since often they can be
    inferred as arrays when that’s not what you want;
  - `src/util/map.js` exists and may help;
  - every function has a single return type (this isn’t Go, e.g.), so
    I’m not sure what you mean by your comment about `fmap`.

src/plugins/discourse/config.js
```diff
@@ -1,29 +1,39 @@
...
+export type SerializedDiscourseConfig = {|
+  +serverUrl: string,
+  +mirrorOptions?: $Shape<MirrorOptions>,
+  +weights: SerializedWeightsConfig,
+|};
 
 export type DiscourseConfig = {|
   +serverUrl: string,
   +mirrorOptions?: $Shape<MirrorOptions>,
+  +weights?: WeightsConfig,
```
decentralion, line 20: Weights should be mandatory in the DiscourseConfig (its optional in the SerializedConfig, fill in with defaults when not present)

src/plugins/discourse/weights.test.js
```diff
@@ -31,4 +31,14 @@ describe("plugins/discourse/weights", () => {
...
   });
+
+  describe("parseCategoryId", () => {
+    it("rejects keys that arent numbers", () => {
+      const thunk = () => parseCategoryId("bad");
+      expect(thunk).toThrow(`CategoryId should be a string integer; got bad`);
+    });
+    it("accepts keys that are numbers", () => {
+      expect(parseCategoryId("5")).toEqual("5");
+    });
```
decentralion, line 42: Let's check that it rejects non-integer numbers like "5.2"

decentralion (approved):
approved-with-nits

PR #2644, opened 2021-01-10
Discourse: implement configurable tag and category weights

topocount:
Tags and categories on a Topic can now be configured with individual weights,
which currently are multiplied together to yield the total weight each like
on a Post in a Topic  will contain in a graph.

Tags and Topic Category can be used in combination to create like-weight
 multipliers on Topics. This means that if any assigned tag or topic category
 is set to 0, all likes on that topic will mint zero cred.

# Tags
Tags can be configured to confer specific like-weight multipliers when
added to a Topic.
If a tag does not have a configured weight, the defaultWeight is applied.
An example configuration might look like:
```json
"weights": {
  "defaultTagWeight": 1,
  "tagWeights": {
    "foo": 0,
    "bar": 1.25,
    "baz": 2
  }
}
```
where foo and bar are the names of tags used in discourse.
  
When multiple tags are assigned to a topic, their weights are multiplied
together to yield a total tag Weight multiplier. In our example configuration,
if both foo and bar are added to a topic, likes on posts in the topic will
have a weight of 0, (0 * 1.25 = 0), which means that no cred will be minted
by those likes.
  
If "bar" and "baz" are both added to another topic, the likes on all posts
in that topic will carry a weight of 2.5 (1.25 * 2 = 2.5), which means that
2.5x as much cred will be minted by those likes.

# Categories
Categories can be configured to confer a specific like-weight multiplier
when added to a Topic.
If a category does not have a configured weight, the defaultWeight is applied.
An example configuration might look like:
```json
weights: {
 "defaultCategoryWeight": 1,
 "categoryWeights": {
   "5": 0,
   "36": 1.25
 }
}
```
where "5" and "36" are the categoryIds in discourse.
  
An easy way to find the categoryId for a given category is to browse to the
categories section in discourse
(e.g. https://discourse.sourcecred.io/categories).
Then mousing over or clicking on a category will bring you to a url that
has the shape https://exampleUrl.com/c/ \<category name\>/\<categoryId\>
Clicking on the community category in sourcecred navigates to
https://discourse.sourcecred.io/c/community/26 for example, where the
categoryId is 26

builds towards #2455

test plan: yarn unit

### DIFF
--- a/src/plugins/discourse/createGraph.js
+++ b/src/plugins/discourse/createGraph.js
@@ -4,7 +4,13 @@ import {Graph, type Node, type Edge} from "../../core/graph";
 import {type NodeWeight} from "../../core/weights";
 import {weightsForDeclaration} from "../../analysis/pluginDeclaration";
 import {type WeightedGraph} from "../../core/weightedGraph";
-import {type PostId, type TopicId, type Post} from "./fetch";
+import {
+  type PostId,
+  type TopicId,
+  type CategoryId,
+  type Tag,
+  type Post,
+} from "./fetch";
 import {type ReadRepository} from "./mirrorRepository";
 import {declaration} from "./declaration";
 import {
@@ -13,7 +19,7 @@ import {
   linksToReferences,
 } from "./references";
 import * as NE from "./nodesAndEdges";
-
+import {type DiscourseConfig} from "./config";
 import {likeWeight} from "./weights";
 
 export type GraphUser = {|
@@ -80,13 +86,18 @@ export const DEFAULT_TRUST_LEVEL_TO_WEIGHT = Object.freeze({
 });
 
 export function _createGraphData(
-  serverUrl: string,
+  config: DiscourseConfig,
   repo: ReadRepository
 ): GraphData {
+  const {serverUrl} = config;
   const users = repo.users().map((u) => NE.userNode(serverUrl, u.username));
 
+  const topicIdToCategory = new Map<TopicId, CategoryId>();
+  const topicIdToTags = new Map<TopicId, $ReadOnlyArray<Tag>>();
   const topics: $ReadOnlyArray<GraphTopic> = repo.topics().map((topic) => {
     const node = NE.topicNode(serverUrl, topic);
+    topicIdToCategory.set(topic.id, topic.categoryId);
+    topicIdToTags.set(topic.id, topic.tags);
     const hasAuthor = NE.authorsTopicEdge(serverUrl, topic);
     return {node, hasAuthor};
   });
@@ -131,8 +142,16 @@ export function _createGraphData(
     const createsLike = NE.createsLikeEdge(serverUrl, like);
     const likes = NE.likesEdge(serverUrl, like);
     const user = repo.findUser(like.username);
-    const weight = likeWeight(user);
-
+    const post = repo.postById(like.postId);
+    const topicId = post ? post.topicId : null;
+    let weight = 0;
+    if (topicId) {
+      const categoryId = topicIdToCategory.get(topicId);
+      const tags = topicIdToTags.get(topicId);
+      weight = likeWeight(config.weights, user, categoryId, tags);
+    } else {
+      weight = likeWeight(config.weights, user);
+    }
     // Update how much total like weight this post has, so that we can
     // set up a hasLikedPost edge flowing cred from the topic
     const existingWeight = postIdToLikeWeight.get(like.postId) || 0;
@@ -201,10 +220,10 @@ export function _graphFromData({
 }
 
 export function createGraph(
-  serverUrl: string,
+  config: DiscourseConfig,
   repo: ReadRepository
 ): WeightedGraph {
-  const data = _createGraphData(serverUrl, repo);
+  const data = _createGraphData(config, repo);
   return _graphFromData(data);
 }
 
--- a/src/plugins/discourse/weights.js
+++ b/src/plugins/discourse/weights.js
@@ -6,6 +6,7 @@ import * as MapUtil from "../../util/map";
 import {orElse as either} from "../../util/null";
 import {DEFAULT_TRUST_LEVEL_TO_WEIGHT} from "./createGraph";
 import {type User} from "./fetch";
+import {type CategoryId as MirrorCategoryId, type Tag} from "./fetch";
 
 // Expected to be an integer string, like "1" or "123"
 export opaque type CategoryId: string = string;
@@ -26,10 +27,69 @@ export function parseTagId(id: string): TagId {
 }
 
 export type SerializedWeightsConfig = {|
+  /**
+   * Tags can be configured to confer specific like-weight multipliers when
+   * added to a Topic.
+   * If a tag does not have a configured weight, the defaultWeight is applied.
+   * An example configuration might look like:
+   * ```
+   * "weights": {
+   *   "defaultTagWeight": 1,
+   *   "tagWeights": {
+   *     "foo": 0,
+   *     "bar": 1.25,
+   *     "baz": 2
+   *   }
+   *   // categoryWeight configs...
+   * }
+   * ```
+   * where foo and bar are the names of tags used in discourse.
+   *
+   * When multiple tags are assigned to a topic, their weights are multiplied
+   * together to yield a total tag Weight multiplier. In our example configuration,
+   * if both foo and bar are added to a topic, likes on posts in the topic will
+   * have a weight of 0, (0 * 1.25 = 0), which means that no cred will be minted
+   * by those likes.
+   *
+   * If "bar" and "baz" are both added to another topic, the likes on all posts
+   * in that topic will carry a weight of 2.5 (1.25 * 2 = 2.5), which means that
+   * 2.5x as much cred will be minted by those likes.
+   */
   +defaultTagWeight?: NodeWeight,
   +tagWeights?: {|[TagId]: NodeWeight|},
+  /**
+   * Categories can be configured to confer a specific like-weight multiplier
+   * when added to a Topic.
+   * If a category does not have a configured weight, the defaultWeight is applied.
+   * An example configuration might look like:
+   * ```
+   * weights: {
+   *  "defaultCategoryWeight": 1,
+   *  "categoryWeights": {
+   *    "5": 0,
+   *    "36": 1.25
+   *  }
+   *  // tagWeight configs...
+   * }
+   * ```
+   * where "5" and "36" are the categoryIds in discourse.
+   *
+   * An easy way to find the categoryId for a given category is to browse to the
+   * categories section in discourse
+   * (e.g. https://discourse.sourcecred.io/categories).
+   * Then mousing over or clicking on a category will bring you to a url that
+   * has the shape https://exampleUrl.com/c/<category name>/<categoryId>
+   * Clicking on the community category in sourcecred navigates to
+   * https://discourse.sourcecred.io/c/community/26 for example, where the
+   * categoryId is 26
+   */
   +defaultCategoryWeight?: NodeWeight,
   +categoryWeights?: {|[CategoryId]: NodeWeight|},
+  /*
+   * Tags and Topic Category can be used in combination to create like-weight
+   * multipliers on Topics. This means that if any assigned tag or topic category
+   * is set to 0, all likes on that topic will mint zero cred.
+   */
 |};
 
 export type WeightsConfig = {|
@@ -63,11 +123,17 @@ export const weightsConfigParser: C.Parser<WeightsConfig> = C.fmap(
   upgrade
 );
 
-export function likeWeight(user: ?User): NodeWeight {
-  if (user == null) {
-    return 0;
-  }
-  return _trustLevelWeight(user.trustLevel);
+export function likeWeight(
+  weights: WeightsConfig,
+  user: ?User,
+  category: ?MirrorCategoryId,
+  tags: ?$ReadOnlyArray<Tag>
+): NodeWeight {
+  const trustLevel = user == null ? 0 : _trustLevelWeight(user.trustLevel);
+  const categoryWeight =
+    category == null ? 1 : _categoryWeight(category, weights);
+  const tagWeight = tags == null ? 1 : _weightFromTags(tags, weights);
+  return trustLevel * categoryWeight * tagWeight;
 }
 
 export function _trustLevelWeight(trustLevel: number | null): NodeWeight {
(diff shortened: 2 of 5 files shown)


s-ben:
Just looking at the commit message, seeing this logic:
>Each tag attached to a topic also acts as a multiplier as well.

So, if I have two tags (tag A and tag B), with weights of 10%. Does that mean the resulting weight of a topic is 120% of a normal topic? Or does the multiplication happen in series (i.e. topic A adds 10% (110%), then topic B is applied and adds (110 x 0.10 = 121%?). Apologies if getting semantics of weight multipliers. Would it possible for these to be additive? Just trying to understand logic.

topocount:
> Just looking at the commit message, seeing this logic:
> 
> > Each tag attached to a topic also acts as a multiplier as well.
> 
> So, if I have two tags (tag A and tag B), with weights of 10%. Does that mean the resulting weight of a topic is 120% of a normal topic? Or does the multiplication happen in series (i.e. topic A adds 10% (110%), then topic B is applied and adds (110 x 0.10 = 121%?). Apologies if getting semantics of weight multipliers. Would it possible for these to be additive? Just trying to understand logic.

@s-ben If you want a tag to add 10%, and another tag to add 20% on top of that, you'd weight them as 1.1 and 1.2, respectively. @decentralion would be curious to wee your answer to this

s-ben:
>@s-ben If you want a tag to add 10%, and another tag to add 20% on top of that, you'd weight them as 1.1 and 1.2, respectively. and they'd total to more than 130% (132%) @decentralion would be curious to see your answer to this.

>As far as I understand it, node weights are always calculated via multiplication Thanks @topocount, makes sense. 

Thanks @topocount. LFW was asking in chat if it worked that way and I say I believe so, but wanted to make sure.

src/plugins/discourse/createGraph.js
```diff
@@ -131,7 +142,13 @@ export function _createGraphData(
     const createsLike = NE.createsLikeEdge(serverUrl, like);
     const likes = NE.likesEdge(serverUrl, like);
     const user = repo.findUser(like.username);
-    const weight = likeWeight(user);
+    const post = repo.postById(like.postId) || {};
```
decentralion, line 145: So if we were to type this out we would write `const post: Post | {} = ...`. That feels much less natural than `const post: ?Post` or `const post: Post | null`. Fix?

src/plugins/discourse/weights.js
```diff
@@ -63,11 +65,26 @@ export const weightsConfigParser: C.Parser<WeightsConfig> = C.fmap(
   upgrade
 );
 
-export function likeWeight(user: ?User): NodeWeight {
+export function likeWeight(
+  {weights = {}}: DiscourseConfig,
```
decentralion, line 69: I think it'll be cleaner to make weights mandatory (see: https://github.com/sourcecred/sourcecred/pull/2643#discussion_r556213209) and just take a definite non-null DiscourseConfig

src/plugins/discourse/weights.js
```diff
@@ -63,11 +65,26 @@ export const weightsConfigParser: C.Parser<WeightsConfig> = C.fmap(
...
+  category: ?DiscourseCategoryId,
+  tags: ?$ReadOnlyArray<Tag>
+): NodeWeight {
   if (user == null) {
     return 0;
   }
-  return _trustLevelWeight(user);
+  if (category == null) {
+    category = -Infinity;
+  }
```
decentralion, line 79: Shouldn't this be a type error? CategoryId is a string not a number. Why is Flow OK with this?

src/plugins/discourse/weights.js
```diff
@@ -63,11 +65,26 @@ export const weightsConfigParser: C.Parser<WeightsConfig> = C.fmap(
...
+  if (tags == null) {
+    tags = [];
+  }
+  return (
+    _trustLevelWeight(user) *
+    _categoryWeight(category, weights) *
+    _weightFromTags(tags, weights)
```
decentralion, line 86: I recommend switching to the approach we looked at while pair programming, like:

```js
const categoryMult = categoryId == null ? 1 : _categoryWeight(categoryId, weights);
const tagMult = ...
```

src/plugins/discourse/weights.js
```diff
@@ -87,3 +104,30 @@ export function _trustLevelWeight({trustLevel}: User): NodeWeight {
   }
   return weight;
 }
+
+export function _categoryWeight(
+  category: DiscourseCategoryId,
+  {categoryWeights = new Map(), defaultCategoryWeight = 1}: WeightsConfig
+): NodeWeight {
+  const categoryString = category.toString();
```
decentralion, line 112: category is already a string

src/plugins/discourse/weights.js
```diff
@@ -87,3 +104,30 @@ export function _trustLevelWeight({trustLevel}: User): NodeWeight {
   }
   return weight;
 }
+
+export function _categoryWeight(
+  category: DiscourseCategoryId,
+  {categoryWeights = new Map(), defaultCategoryWeight = 1}: WeightsConfig
```
decentralion, line 110: Let's just take a WeightsConfig without any default initialization, let the DiscourseConfig have responsibility for always having a sensible default.

src/plugins/discourse/weights.js
```diff
@@ -87,3 +104,30 @@ export function _trustLevelWeight({trustLevel}: User): NodeWeight {
...
+): NodeWeight {
+  let totalWeight = 1;
+  for (const tag of tags) {
+    const weight = tagWeights.get(tag);
+    if (weight != null) {
+      totalWeight *= weight;
+    } else {
+      totalWeight *= defaultTagWeight;
+    }
+  }
```
decentralion, line 131: This is incorrect. Consider the case where we have defaultTagWeight as 2, no weights on any particular tags, and a like is associated with 2 tags. Then we will incorrectly get a weight of `4` because we multiply by the default weight twice.

I suggest the following code:

```js
const weightForTag = (tagId: TagId) => tagWeights.get(tagId) || 1;
return tags.reduce((acc, tag) => acc * weightForTag(tag), defaultTagWeight)
```
```

File: src/plugins/discourse/weights.test.js [hunk 8]
```diff
@@ -32,6 +60,46 @@ describe("plugins/discourse/weights", () => {
...
+      });
+      expect(_weightFromTags(["you're"], getWeights(config))).toBe(5);
+    });
+    it("can return a weight set to 0", () => {
+      const config = getConfig({defaultTagWeight: 1, tagWeights: {"it": 0}});
+      expect(_weightFromTags(["it"], getWeights(config))).toBe(0);
+    });
+    it("returns 1 when no tags are assigned", () => {
+      expect(_weightFromTags([], getWeights(getConfig()))).toBe(1);
+    });
```
decentralion, line 100: No, it should return the defaultTagWeight, not 1

src/plugins/discourse/weights.test.js
```diff
@@ -32,6 +60,46 @@ describe("plugins/discourse/weights", () => {
...
+  describe("_tagWeight", () => {
+    it("returns the default weight when the category doesn't have a configured weight", () => {
+      expect(_weightFromTags(["tag"], getWeights(getConfig()))).toBe(1);
+    });
+    it("returns the default weight when the category doesn't have a configured weight", () => {
+      const config = getConfig({
+        defaultTagWeight: 1,
+        tagWeights: {"you're": 5},
+      });
+      expect(_weightFromTags(["you're"], getWeights(config))).toBe(5);
```
decentralion, line 92: Please set the default weight to something other than 1 so this test can catch if we just are returning hardwired 1

src/plugins/discourse/weights.test.js
```diff
@@ -32,6 +60,46 @@ describe("plugins/discourse/weights", () => {
...
+      });
+      expect(_categoryWeight(1, getWeights(config))).toBe(0);
+    });
+  });
+
+  describe("_tagWeight", () => {
+    it("returns the default weight when the category doesn't have a configured weight", () => {
+      expect(_weightFromTags(["tag"], getWeights(getConfig()))).toBe(1);
+    });
+    it("returns the default weight when the category doesn't have a configured weight", () => {
```
decentralion, line 87: s/category/tag/

src/plugins/discourse/weights.test.js
```diff
@@ -32,6 +60,46 @@ describe("plugins/discourse/weights", () => {
...
+        categoryWeights: {"1": 0},
+      });
+      expect(_categoryWeight(1, getWeights(config))).toBe(0);
+    });
+  });
+
+  describe("_tagWeight", () => {
+    it("returns the default weight when the category doesn't have a configured weight", () => {
+      expect(_weightFromTags(["tag"], getWeights(getConfig()))).toBe(1);
+    });
```
decentralion, line 86: s/category/tag, also the test says it checks that we return the default weight, but it actually checks if it returns 1 -- ie it doesn't test what happens if the default weight is not 1. please fix

src/plugins/discourse/weights.test.js
```diff
@@ -32,6 +60,46 @@ describe("plugins/discourse/weights", () => {
     });
   });
 
+  describe("_categoryWeight", () => {
+    it("returns the default weight when the category doesn't have a configured weight", () => {
```
decentralion, line 64: should test for a defaultCategoryWeight that is not 1

src/plugins/discourse/weights.test.js
```diff
@@ -32,6 +60,46 @@ describe("plugins/discourse/weights", () => {
...
+      expect(_weightFromTags(["you're"], getWeights(config))).toBe(5);
+    });
+    it("can return a weight set to 0", () => {
+      const config = getConfig({defaultTagWeight: 1, tagWeights: {"it": 0}});
+      expect(_weightFromTags(["it"], getWeights(config))).toBe(0);
+    });
+    it("returns 1 when no tags are assigned", () => {
+      expect(_weightFromTags([], getWeights(getConfig()))).toBe(1);
+    });
+  });
```
decentralion, line 101: consider adding a test that would catch the existing bug in the implementaiton, i.e. a case where defaultTagWeight is not 1, and there are multiple non-matching tags. (not required b.c. i think the exact same bug is unlikely to re-occur, but it's good habit to write test cases for any previously-failing implementation to prevent regressions)

src/plugins/discourse/weights.test.js
```diff
@@ -10,11 +18,31 @@ function getMockUser(t: number): User {
...
+    },
+  });
+}
+
+function getWeights(c: DiscourseConfig): WeightsConfig {
+  if (!c.weights) {
+    throw new Error("missing weights");
+  }
+  return c.weights;
+}
```
decentralion, line 39: as discussed elsewhere: make the WeightsConfig mandatory and we don't need this function at all.

src/plugins/discourse/weights.test.js
```diff
@@ -10,11 +18,31 @@ function getMockUser(t: number): User {
   };
 }
 
+function getConfig(s: ?Object): DiscourseConfig {
```
decentralion, line 21: do we actually use the `s` argument? Also it's type should be `$Shape<SerializedDiscourseConfig>`

topocount: There's no good way to pass in `categoryId`s to a function like this though, since it's opaque and an object key. this parser also parses `SerializedWeightConfigs` before upgrading them to `WeightConfigs` so passing in a simple object should be safe, right?

src/plugins/discourse/weights.js
topocount: Your suggestion is backwards, I think.

It should be:
```js
const weightForTag = (tagId: TagId) => either(tagWeights.get(tagId), defaultTagWeight);
  return tags.reduce((acc, tag) => acc * weightForTag(tag), 1);
```

wchargin: > This is incorrect. Consider the case where we have defaultTagWeight as 2, no weights on any particular tags, and a like is associated with 2 tags. Then we will incorrectly get a weight of `4` because we multiply by the default weight twice.

…why? @topocount and I discussed this, and we don’t see why this should
be the case. We interpret `defaultTagWeight` as meaning “any tag whose
weight is not given in `tagWeights` is taken to have this weight”.
Logically, this should mean that you multiply by the default weight
multiple times if there are multiple unweighted tags.

So, if your config has `defaultTagWeight: 3` and explicit tag weights
`A: 2` and `B: 5`, then:

  - a like with no tags has weight 1;
  - a like with tag `A` has weight 2;
  - a like with tags `A` and `B` has weight 10;
  - a like with tags `A`, `B`, and `C` has weight 30;
  - a like with tags `A`, `B`, `C`, and `D` has weight 90.

This is just the natural product semantics.

Is this not how you interpret `defaultTagWeight`? I can’t think of a
sensible interpretation that leads to the semantics that you describe.
If we pick a different interpretation, we should add comments to
`discourse/weights.js` to document the semantics.

src/plugins/discourse/weights.js
```diff
@@ -63,11 +64,23 @@ export const weightsConfigParser: C.Parser<WeightsConfig> = C.fmap(
...
+  category: ?MirrorCategoryId,
+  tags: ?$ReadOnlyArray<Tag>
+): NodeWeight {
   if (user == null) {
     return 0;
   }
-  return _trustLevelWeight(user.trustLevel);
+  if (tags == null) {
+    tags = [];
+  }
```
decentralion, line 78: redundant with the `tags == null` check a few lines down.

src/plugins/discourse/weights.js
```diff
@@ -87,3 +100,22 @@ export function _trustLevelWeight(trustLevel: number | null): NodeWeight {
   }
   return weight;
 }
+
+export function _categoryWeight(
+  category: MirrorCategoryId,
+  {categoryWeights, defaultCategoryWeight}: WeightsConfig
+): NodeWeight {
+  const weight = categoryWeights.get(category);
+  if (weight != null) return weight;
```
decentralion, line 109: nit (no change required): i think its cleaner as `return weight == null ? defaultCategoryWeight : weight`

src/plugins/discourse/weights.js
```diff
@@ -87,3 +100,22 @@ export function _trustLevelWeight(trustLevel: number | null): NodeWeight {
...
+  return defaultCategoryWeight;
+}
+
+export function _weightFromTags(
+  tags: $ReadOnlyArray<Tag>,
+  {tagWeights, defaultTagWeight}: WeightsConfig
+): NodeWeight {
+  const weightForTag = (tagId: TagId) =>
+    either(tagWeights.get(tagId), defaultTagWeight);
+  return tags.reduce((acc, tag) => acc * weightForTag(tag), 1);
```
decentralion, line 120: i cant reply to the old review, but re discussion [here](https://github.com/sourcecred/sourcecred/pull/2644#discussion_r557017642) from @wchargin:

yeah, i see how this is actually the logical semantics for defaultTagWeight -- good point.

decentralion (approved):

decentralion:
Thanks for making this happen @topocount! I think this should be good to cut a release from, then let's help @s-ben get this set up for Maker

s-ben:
> Thanks for making this happen @topocount! I think this should be good to cut a release from, then let's help @s-ben get this set up for Maker

Woot! Ready to help test as soon as it's ready.

PR #2645, opened 2021-01-11
Add identity filtering to Admin interface

### LINKED ISSUE #2490: Create a filtering interface for The Identities admin view
Currently, there is no easy way to find identities you want to modify, except for scrolling. We need a filter box to make navigating this page simpler.

BrianBelhumeur:
# Description
Fixes #2490 

Currently, the admin interface does not have any filtering functionality. This means for any non-trivial number of identities, the admin must either scroll manually and look for the requested identity, or (more frequently) use the browser's built-in search function. Either way makes for a terrible UX.

This change adds identity filtering. Characters are typed into the new filter box and filtering of identities is done via name, "in order, but not necessarily sequentially". Meaning the letters in the filter box must all appear in the identity name in that order, but can have other letters in between.
E.g.:
>The matching should be fuzzy for now, meaning `an` will match any identity name that contains both an `a` and an `n` in that order, but not necessarily sequentially. So, it'll match `sandpiper` and `decentralion` but it will not match `Thena`.

I also made a few unrequested improvements.
1. The `Create Identity` button is disabled when the text input is blank. It's more clear that it's actionable without the required info (namely, and identity to create)
2. The button that takes the UX from "Identity edit mode" to "Identity creation mode" has been renamed and restyled. Previously, it was named "New Identity", and attached to and styled the same as "Create Identity" and "Save Ledger to Disk" buttons, giving me the impression that it performed an action of making a new identity. I changed it to be detached and styled differently, and labeled "Back" to more clearly indicate it's a UI view changer, not an identity creation action.

# Test Plan
I tested this change manually. Filtering is done via tableState, which is well-tested.

Flows manually tested:
- Creating a new user
- Updating a user name
- Adding an alias to a new user
- Adding an alias to an existing user
- Filtering existing user list
- Creating a new user, then immediately filtering list
- Selecting a user to edit, then filtering list
- Editing username, then filtering list
- Saving ledger to disk

# Screenshots of changes 
## Old view, identity creation mode:
Screen Shot 2021-01-13 at 00 48 25

## New view, identity creation mode:
Screen Shot 2021-01-12 at 23 53 22

## Old view, identity update mode:
Screen Shot 2021-01-13 at 00 01 53

## New view, identity update mode:
Screen Shot 2021-02-12 at 18 06 51

## Filtering view, with results:
Screen Shot 2021-01-13 at 01 18 05

## Filtering view, no results:
Screen Shot 2021-01-13 at 01 19 19


src/ui/components/LedgerAdmin.js
```diff
@@ -56,6 +73,9 @@ export const LedgerAdmin = (): ReactNode => {
   const [selectedId, setSelectedId] = useState<IdentityId | null>(null);
   const [promptString, setPromptString] = useState<string>("Add Identity:");
   const [checkboxSelected, setCheckBoxSelected] = useState<boolean>(false);
+  const filteredLedger = useTableState(
+    ledger.accounts().map((a) => a.identity)
```
blueridger, line 77: You should probably put this in a memo, as TableState does not do that for you. You can do this inline if you want.
```
useTableState(
    useMemo(ledger.accounts().map((a) => a.identity), [ledger])
)
```

src/ui/components/LedgerAdmin.js
```diff
@@ -14,6 +14,7 @@ import {
   TextField,
 } from "@material-ui/core";
 import {useLedger} from "../utils/LedgerContext";
+import {useTableState} from "../../webutil/tableState";
```
blueridger, line 17: This makes me happy :)

src/ui/components/LedgerAdmin.js
BrianBelhumeur: Got it, thanks!

src/ui/components/LedgerAdmin.js
BrianBelhumeur: Makes me happy to use it!

blueridger (commented):
The test plan section is thin. Could you provide the test cases you verified, and screenshots of the UI changes (maybe even before/after)?

Also, what's up with the cancel button rename? I don't see that in the mocks, and it's not intuitively clear to me what that change is about.

BrianBelhumeur:
@blueridger I added my testing procedures and several screenshots into the description, as well as addressed your curiosity about the unrequested change of the "New Identity" button. Thank you for the feedback, and please let me know if you have more Qs or comments.

blueridger (approved):
Looks GREAT! Thanks for the extra info. Made it super easy to approve.

src/ui/components/LedgerAdmin.js
```diff
@@ -169,14 +209,33 @@ export const LedgerAdmin = (): ReactNode => {
...
         <>
           <AliasView selectedId={selectedId} />
           <IdentityMerger selectedId={selectedId} />
         </>
-      )}
+      )}{" "}
+      <div className={classes.spreadRow}>
+        <h3>
+          Identities{" "}
+          {ledger.accounts().length > 0 && (
```
hammadj, line 227: Instead of calling ledger.accounts() again, you can use the memoized `accounts` variable defined in the comment I made above on Line 76

hammadj (changes requested):
Looks good, just need to adjust how we memoize the ledger account state + some naming nits (filteredLedger > accountTableState)

src/ui/components/LedgerAdmin.js
BrianBelhumeur: Good catch, fixed.

src/ui/components/LedgerAdmin.js
```diff
@@ -96,35 +116,52 @@ export const LedgerAdmin = (): ReactNode => {
...
-          .map((a) => a.identity)
-          .map((identity, index) =>
-            renderIdentity(identity, index < numAccounts - 1)
-          )}
-      </>
+  };
+
+  const renderIdentities = () => {
+    const renderIdentity = (i: Identity) => (
+      <ListItem button onClick={() => setActiveIdentity(i)} key={i.id}>
```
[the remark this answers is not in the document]
BrianBelhumeur: Ah yes I was debugging the `must have key` warning in the console and forgot to remove these. Thanks!

src/ui/components/LedgerAdmin.js
```diff
@@ -96,35 +116,52 @@ export const LedgerAdmin = (): ReactNode => {
...
-        {ledger
-          .accounts()
-          .map((a) => a.identity)
-          .map((identity, index) =>
-            renderIdentity(identity, index < numAccounts - 1)
-          )}
-      </>
+  };
+
+  const renderIdentities = () => {
```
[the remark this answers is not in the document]
BrianBelhumeur: Done and done.

hammadj (approved):

KuraFire:
@BrianBelhumeur can we make the blue of the Aliases name a styled color please? I don't know what the existing MUI  available color scheme is, but let's not leave it as a completely unstyled link.

src/ui/components/LedgerAdmin.js
```diff
@@ -1,22 +1,22 @@
...
   FormControlLabel,
   List,
-  ListItem,
   TextField,
 } from "@material-ui/core";
 import {useLedger} from "../utils/LedgerContext";
+import {useTableState} from "../../webutil/tableState";
 import {IdentityMerger} from "./IdentityMerger";
 import {type Identity, type IdentityId} from "../../core/identity";
 import {AliasView} from "./AliasView";
```
KuraFire, line 18: Is the AliasView component where the unstyled Aliases links are?

KuraFire (changes requested):
Requested change: link style / color.

src/ui/components/LedgerAdmin.js
BrianBelhumeur: Why yes, yes it is.

KuraFire (approved):
Thanks @BrianBelhumeur !

PR #2649, opened 2021-01-12
Ledger: move reuseable test code into testUtils

blueridger:
# Description
Simple refactor to move some test code into the ledger/testUtils module because I want to reuse some of it for the CredGrainView tests.

# Test Plan
yarn test

### DIFF
--- a/src/core/ledger/testUtils.js
+++ b/src/core/ledger/testUtils.js
@@ -1,4 +1,9 @@
 // @flow
+
+import cloneDeep from "lodash.clonedeep";
+import {Ledger} from "./ledger";
+import {newIdentity, type Identity, type IdentityId} from "../identity";
+import * as G from "./grain";
 import * as uuid from "../../util/uuid";
 
 export interface UuidMock {
@@ -6,7 +11,19 @@ export interface UuidMock {
   setNextUuid(id: uuid.Uuid): void;
 }
 
-export function createUuidMock(): UuidMock {
+export interface DateMock {
+  resetFakeDate(): void;
+  setFakeDate(id: number): void;
+}
+
+export interface LedgerMock {
+  identity1(): Identity;
+  identity2(): Identity;
+  ledgerWithIdentities(): Ledger;
+  ledgerWithActiveIdentities(): Ledger;
+}
+
+export const createUuidMock = (): UuidMock => {
   const randomMock = jest.spyOn(uuid, "random");
 
   let nextFakeUuidIndex = 0;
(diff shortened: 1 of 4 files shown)


src/core/ledger/testUtils.js
```diff
@@ -29,3 +34,63 @@ export function createUuidMock(): UuidMock {
...
+
+const {resetFakeUuid, setNextUuid} = uuidMock;
+
+// Helper for constructing Grain values.
+export const g = (s: string): G.Grain => G.fromString(s);
+
+let nextFakeDate = 0;
+export function resetFakeDate() {
+  nextFakeDate = 0;
+}
```
decentralion, line 48: So there's an issue here where we have a piece of module state that's mutating across tests. Thus, we could get nondeterministic output depending on the order that the different test files execute in.

Basically, setting `nextFakeDate = 0` in the outer scope was kinda sketchy when this was inside a test describe block, but it's even sketchier when we factor it out into a module.

Can we wrap this state up in a closure like with `createUuidMock`

src/core/ledger/testUtils.js
```diff
@@ -1,12 +1,23 @@
...
   setNextUuid(id: uuid.Uuid): void;
 }
 
-export function createUuidMock(): UuidMock {
+export interface DateMock {
+  resetFakeDate(): void;
+  setFakeDate(id: number): void;
+}
+
+let uuidMock = _createUuidMock();
```
decentralion, line 19: Seems like we're still leaking test-mutable state into the module scope here.

blueridger: fair enough... Looks like to do it right I'll need to move the identity and ledger builders into the mock state. Shouldn't be too hard.

src/core/ledger/testUtils.js
```diff
@@ -26,6 +43,78 @@ export function createUuidMock(): UuidMock {
...
+    // Use this when you want specific timestamps, rather than just
+    // auto-incrementing
+    jest.spyOn(global.Date, "now").mockImplementationOnce(() => ts);
+  }
+  jest.spyOn(global.Date, "now").mockImplementation(() => nextFakeDate++);
+
+  return {setFakeDate, resetFakeDate};
+};
+
+export const createLedgerMock = (
```
decentralion, line 65: IMO this isn't really a mock--more like a fixture (`testLedgerFixture`?).

decentralion (approved):

src/core/ledger/testUtils.js
blueridger: also felt sus to me. will change

PR #2435, opened 2020-10-28
Generate base docs

youngkidwarrior:
This PR adds jsdoc to to the repo. Docs are saved locally in the /docs folder. 

Use the `yarn docs` command to generate docs programmatically seperated by class. Using the already available documentation.

Changes are not made to docusaurus, following PRs will handle this.

Ignoring flow typing in script file as well. I'm unfamiliar with flow, so I can handle this in another PR if it is necessary.

Here's an example of the docs running on my local machine.

Cred_Docs

PR #2650, opened 2021-01-12
Flow typed docs script

youngkidwarrior:
# Description
Dependant on #2435 
Flow typing the local generated docs script. Using the flow comment type structure because we need it to work as plain javascript.

# Test Plan
- Run `yarn docs` to generate the docs folder
- Ensure it logs `success` for all generated docs

### DIFF
--- a/scripts/build-docs.js
+++ b/scripts/build-docs.js
@@ -5,12 +5,12 @@ const path = require("path");
 const jsdoc2md = require("jsdoc-to-markdown");
 const chalk = require("chalk");
 
-function success(message) {
+function success(message /*: string */) {
   const label = chalk.bgGreen.bold.white(" SUCCESS ");
   console.log(`${label} ${message}`);
 }
 
-function error(message) {
+function error(message /*: string */) {
   const label = chalk.bgRed.bold.white(" FAIL ");
   console.log(`${label} ${message}`);
 }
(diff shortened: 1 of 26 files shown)


BrianBelhumeur:
Looks like you accidentally committed your `.DS_Store` file

decentralion (commented):
The PR description and test plan need to communicate what the change does, and how the reviewer can be confident that the change is correct. Right now, that's not really possible--reading the description I don't actually know what this PR does or how to test it. Does it add a command that generates docs? How can I view those docs? How will those docs get consumed or integrated elsewhere?

Please re-write this so there's a test plan. It could be something like:

- Run the following command to generate docs
- Run the following command to view the docs
- Explore the docs and confirm they are correctly generated from the code comments

It would also be nice to have a screenshot or such of how the docs look once generated.

See: [the relevant section of CONTRIBUTING.md](https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#when-writing-commit-messages)

youngkidwarrior:
@decentralion my bad on the test plan! 

This PR is a dependent on #2435 which has all the stuff you asked for (description of changes, video example, etc)

hammadj (commented):
Changed the base of this PR to point at the branch for #2435 so that both the changes dont show in the PR. Looks mostly good, but there's an update to yarn.lock that doesn't belong here since there wasn't change to package.json.

PR #2651, opened 2021-01-14
Discourse: Convert CategoryId to string

topocount:
Even though categoryId is an integer, it makes little sense to
treat it as an such, since it's an identifier, and it has no 
relationship to to other identifiers.

test plan: yarn unit


decentralion (approved):
Agreed that it's more convenient to treat these as strings.

PR #2655, opened 2021-01-14
Use LedgerManager to control loading ledger state on frontend

hammadj:
# Description

Replaces the manual loading method used previously to using the LedgerManager to abstract
the loading logic. Can be used to enable different ledger storage backends in the future.

# Test Plan

- Ensure ledger still loads correctly and UI updates when making ledger changes
- Edit the ledger.json to include a "bad" event (e.g. double spending grain in a grain transfer) and ensure that loading the ledger fails on the front end

image

### DIFF
--- a/src/ui/utils/LedgerContext.js
+++ b/src/ui/utils/LedgerContext.js
@@ -2,6 +2,7 @@
 
 import * as React from "react";
 import {Ledger} from "../../core/ledger/ledger";
+import {LedgerManager} from "../../api/ledgerManager";
 
 const LedgerContext = React.createContext<LedgerContextValue>({
   ledger: new Ledger(),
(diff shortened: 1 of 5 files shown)


src/ui/utils/LedgerContext.js
```diff
@@ -14,17 +15,21 @@ type LedgerContextValue = {|
...
+  ledgerManager,
 }: LedgerProviderProps): React.Node => {
-  const [{ledger}, setLedgerState] = React.useState({ledger: initialLedger});
+  const [{ledger}, setLedgerState] = React.useState({
+    ledger: ledgerManager.ledger,
+  });
 
-  const updateLedger = (ledger: Ledger) => setLedgerState({ledger});
+  // Workaround to trigger UI update in React since ledger object is mutable
+  const updateLedger = (_?: Ledger) =>
```
blueridger, line 31: Why keep the param?

blueridger (commented):
Since you are adding error case logic, could you please add the error case to your test plan? Perhaps by manually injecting bad data into the ledger.

src/ui/utils/LedgerContext.js
blueridger: Ah I think I understand. Because this is only one implementation of theoretically many, and others may need the param.

hammadj: Yea exactly, didn't want to break existing code that was passing in the ledger since we might need it in future implementations.

blueridger (approved):

KuraFire:
I'm confused here, @hammadj : There are three reviewers that haven't yet reviewed, but the PR was merged anyway.

This touches upon UI, but there is no illustration of what the UI behavior will look like, or how it will behave. How can I approve this in my review?

The way this PR went through doesn't strike me as the process we're supposed to have?

PR #2656, opened 2021-01-15
Update Save Ledger logic to use LedgerManager

hammadj:
# Description
Replaced all the instances of the manually written POST requests with a saveLedger function
so that the LedgerManager logic is used to save to disk and abstracted from the UI layer. Can
be used in the future to persist the ledger to various backends.

Also added an alert so the user gets a visual confirmation that the Ledger has been
saved (or if there was an error)

# Test Plan
Update the ledger from the Identities / Transfer / Special distribution pages
and save it. Ensure the ledger file is updated and the alert confirms the ledger was saved.

### DIFF
--- a/src/ui/utils/LedgerContext.js
+++ b/src/ui/utils/LedgerContext.js
@@ -7,11 +7,13 @@ import {LedgerManager} from "../../api/ledgerManager";
 const LedgerContext = React.createContext<LedgerContextValue>({
   ledger: new Ledger(),
   updateLedger: () => {},
+  saveToDisk: () => Promise.resolve(),
 });
 
 type LedgerContextValue = {|
   +ledger: Ledger,
   updateLedger(ledger: Ledger): void,
+  saveToDisk(): Promise<void>,
 |};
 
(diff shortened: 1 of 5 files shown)


blueridger (approved):

src/ui/utils/LedgerContext.js
```diff
@@ -31,8 +33,21 @@ export const LedgerProvider = ({
...
     setLedgerState({ledger: ledgerManager.ledger});
 
+  const saveToDisk = async () => {
+    try {
+      const res = await ledgerManager.persist();
+      if (res.error) {
+        alert(`Failed to save ledger: ${res.error}`);
+        return;
+      }
+      alert("Ledger saved to disk");
```
blueridger, line 43: Have you run this by Faruk?

KuraFire: No but I'm not going to focus design work on the outdated explorer / admin UI anymore; instead I'm focusing on ensuring the UI and IXD is designed and UX-optimized for the new Explorer. (Thanks Thena! <3 )

PR #2660, opened 2021-01-16
immediate policy: add lookback parameter

elihanover:
Right now, the immediate policy uses cred from the most recent interval.  This generalizes
it such that users can create policies based on the past N weeks, without the discounting
of the Recent policy.

This includes an optional `numIntervalsLookback` parameter in the `ImmediatePolicy` type, which
defaults to 1 in config until #2600 gets merged, at which point we'll allow users to specify
alternative lookback lengths.

Note, this raises a question around naming, as Recent is actually a better name here, and we
could use something for the current Recent policies that highlights the discounting more.

__Test Plan__
Unit tests provided check that the `immediatePolicy` works for valid params, and also checks
for invalid `numIntervalsLookback` parameters.

This also updates the snapshot in `core/ledger/ledger.test` to include this extra parameter
on the `ImmediatePolicy`.

__Note__
This is a fix of #2546, which passed CI tests but broke on changes in master.


elihanover:
This should be good to go.  Just had to rebase on master and fix a new test file that caused the flow error, as it was creating `ImmediatePolicy`s without a lookback parameter.

src/core/ledger/policies/immediate.test.js
```diff
@@ -0,0 +1,118 @@
...
+        budget: fromGrain(G.ONE),
+        numPeriodsLookback: 4,
+      };
+      const policy2 = {
+        policyType: "IMMEDIATE",
+        budget: fromGrain(G.ONE),
+        numPeriodsLookback: 50,
+      };
+      const expected1 = immediateReceipts(policy1, processedIdentities);
+      const expected2 = immediateReceipts(policy2, processedIdentities);
```
blueridger, line 61: nit technically this is `actual`

src/api/grainConfig.js
```diff
@@ -67,6 +67,7 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
     allocationPoliciesDeprecated.push({
       budget: toNonnegativeGrain(immediatePerWeek),
       policyType: "IMMEDIATE",
+      numPeriodsLookback: 1, // TODO(eli): no customization until after #2600.
```
blueridger, line 70: this should probably be renamed `numIntervalsLookback ` for consistency.

elihanover: Agreed, done.

src/core/ledger/policies/immediate.test.js
elihanover: Done, I wrote as follows:
```js
const expected = immediateReceipts(policy1, processedIdentities);
const actual = immediateReceipts(policy2, processedIdentities);
```
Of course both are actual, but since the test looks for the second to match the first, the first is essentially the "expected".

src/core/ledger/policies/immediate.test.js
```diff
@@ -0,0 +1,118 @@
...
+        numIntervalsLookback: 4,
+      };
+      const policy2 = {
+        policyType: "IMMEDIATE",
+        budget: fromGrain(G.ONE),
+        numIntervalsLookback: 50, // 50 > number of intervals (4)
+      };
+      const expected = immediateReceipts(policy1, processedIdentities);
+      const actual = immediateReceipts(policy2, processedIdentities);
+      expect(expected).toEqual(actual);
```
blueridger, line 62: should be reversed: `expect(actual).toEqual(expected);` this is slightly important because it affects the test fail output syntax. Feel free to fix and merge on this approval.

blueridger (approved):

Issue #2464, opened 2020-11-06
[ExplorerHome] Use paginated CredGrainView


topocount:
## AC (Acceptance Criteria)
- ExplorerHome uses CredGrainView and NOT CredView
- ExplorerHome displays real cred data loaded from json
- CredGrainView is optional similar to CredView (I can load the admin view without a credGraph.json in my instance)
- TableState+MaterialUI pagination

## Not AC
- filtering/sorting

BrianBelhumeur:
Taking with permission from @befitsandpiper

BrianBelhumeur:
For clarity and to keep the PRs more atomic, I'm going to split this into two PRs.

PR #2665, opened 2021-01-18
Refactor Explorer admin UI components into their own files

BrianBelhumeur:
# Description
In the course of prepping to do issue #2464, I saw some refactoring that was due for the Explorer page in the admin UI. This is a pure refactoring PR with no functional changes.

- Factor out FilterSelect and WeightsConfigSection components
- Convert from aphrodite to material-ui style creation

# Test Plan
Tested manually. Here are some before and after screenshots (yes they are supposed to be identical)
## Before
Explorer list before
Explorer weights before

## After
Explorer list after
Explorer weights after

### DIFF
--- a/src/ui/components/Explorer/Explorer.js
+++ b/src/ui/components/Explorer/Explorer.js
@@ -8,44 +8,33 @@ import {
   TableCell,
   TableHead,
   TableRow,
-  Menu,
-  MenuItem,
-  ListItem,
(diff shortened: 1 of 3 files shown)


blueridger (approved):
FYI it's probably not worthwhile to invest in this page because work to deprecate it is currently underway.

hammadj (approved):

PR #2671, opened 2021-01-18

elihanover:
__Context__
After implementing the Recent policy and a variant of Balanced (#2584), we can see that adding policy parameters for new policies is both tedious on from our perspective as well as for the user.

More specifically, on the user side, their config is a flat object with parameters from different policies sitting side by side. Here, we implement more structure by taking an array of policies.

Right now, an example config/grain.json might look as follows:

```js
{
  "balancedPerWeek": 30
  "immediatePerWeek": 50,
  "recentPerWeek": 20,
  "maxSimultaneousDistributions": 2,
  "discount": 0.5
}
```

This also forced us to check each policy explicitly downstream in toDistributionPolicy, which could be a lot cleaner if we could leverage the AllocationPolicy type we already have. Now, we have a parser which will eliminate the need for the
function entirely after we remove the deprecated config.

```js
{
  allocationPolicies: [
    {
      policyType: "BALANCED",
      budget: 100
    },
    {
      policyType: "RECENT",
      budget: 50,
      discount: 0.1
    },
    {
      policyType: "RECENT",
      budget: 100,
      discount: 0.5
    }]
}
```

Note that for the time being, we support the deprecated config.  For example:
```js
{
  balancedPerWeek: 10,
  immediatePerWeek: 10,
  allocationPolicies: [
    {
      policyType: "BALANCED",
      budget: 100
    },
    {
      policyType: "RECENT",
      budget: 50,
      discount: 0.1
    },
    {
      policyType: "RECENT",
      budget: 100,
      discount: 0.5
    }]
}
```

This also lets us take multiple versions of the same policies. This allows us to create richer ensembles of policies. For instance, I might want to run a Recent policy with a 90% discount, and another one with a 5% discount.

__Test Plan__
Unit tests are provided here for both the parser, and well as toDistributionPolicy, which serves to check the config.

### DIFF
--- a/src/api/grainConfig.js
+++ b/src/api/grainConfig.js
@@ -3,33 +3,45 @@
 import {type DistributionPolicy} from "../core/ledger/applyDistributions";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
+import {
+  type AllocationPolicy,
+  policyConfigParser,
+} from "../core/ledger/policies";
 import {fromInteger as toNonnegativeGrain} from "../core/ledger/nonnegativeGrain";
 import {toDiscount} from "../core/ledger/policies/recent";
 
 export type GrainConfig = {|
-  +immediatePerWeek?: number,
-  +balancedPerWeek?: number,
-  +recentPerWeek?: number,
-  +recentWeeklyDecayRate?: number,
+  +immediatePerWeek?: number, // (deprecated)
+  +balancedPerWeek?: number, // (deprecated)
+  +recentPerWeek?: number, // (deprecated)
+  +recentWeeklyDecayRate?: number, // (deprecated)
+  +allocationPolicies?: $ReadOnlyArray<AllocationPolicy>,
   +maxSimultaneousDistributions?: number,
 |};
 
 export const parser: C.Parser<GrainConfig> = C.object(
   {},
   {
+    allocationPolicies: C.array<AllocationPolicy>(policyConfigParser),
+    maxSimultaneousDistributions: C.number,
     immediatePerWeek: C.number,
     balancedPerWeek: C.number,
     recentPerWeek: C.number,
     recentWeeklyDecayRate: C.number,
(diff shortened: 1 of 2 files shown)


src/api/grainConfig.js
```diff
@@ -50,9 +62,9 @@ export function toDistributionPolicy(x: GrainConfig): DistributionPolicy {
     );
   }
 
-  const allocationPolicies = [];
+  const allocationPoliciesDeprecated: Array<AllocationPolicy> = [];
```
decentralion, line 65: Nit: I don't think there's any need for this variable--you could just push them directly into `allocationPolicies`

src/api/grainConfig.test.js
```diff
@@ -3,122 +3,237 @@
...
+  budget: toNonnegativeGrain(budget),
+  memo,
+  recipient,
+});
 
 describe("api/grainConfig", () => {
   describe("parser", () => {
-    it("does not throw with no params", () => {
-      expect(parser.parseOrThrow({})).toEqual({});
-    });
```
decentralion, line 12: Let's keep this test? It should still be fine with no params.

decentralion (approved):
approve-with-nits

Also, the PR description says there needs to be at least 1 allocation policy. 
> Now, we only need to check that at least one policy is provided and budgets are positive.


These semantics would be bad -- an empty array of policies is totally fine, and may be reasonable in some circumstances. Happily, the implementation does not discriminate against empty policy arrays. Let's keep it that way.

src/api/grainConfig.test.js
elihanover: Done

src/api/grainConfig.js
elihanover: Ack, I left as is, as I like thinking of these processes as a tree. One branch handles x.allocationPolicies and the other handles the deprecated policies.

PR #2672, opened 2021-01-18
cli/grain2: grain2 command using CredGrainView

elihanover:
__Context__
    This commit implements a `grain2` command, which is a simple fork of `grain` that
    moves away from using the soon to be deprecated `CredView` and instead uses `CredGraph`.
    This is part of a larger effort from @blueridger to replace `CredView` with `CredGrainView`.

Note that `grain2` now loads the `CredGraph` instead of `CredResults`.

This commit also implements forks of `applyDistributions` and `computeCredAccounts` to
    take a `CredGraph` instead of a `CredView`.  We can phase out the originals when we
    deprecate `CredView` officially.

__Assumptions__
    - Original applyDistributions and computeCredAccounts works.  I forked this assuming
      it works, but there's no test for it.

__Test Plan__
    Ran `grain2` and `grain` on the same sample instance:
    - Distribution output in ledger matches using JSON.stringify
    - -s flag does not write to ledger in grain2 as expected

### DIFF
--- a/src/cli/grain2.js
+++ b/src/cli/grain2.js
@@ -1,0 +1,114 @@
+// @flow
+
+import fs from "fs-extra";
+import {join} from "path";
+import {loadFileWithDefault, loadJson} from "../util/disk";
+import {Ledger} from "../core/ledger/ledger";
+import {applyDistributions2} from "../core/ledger/applyDistributions";
+import {computeCredAccounts2} from "../core/ledger/credAccounts";
+import stringify from "json-stable-stringify";
+import * as G from "../core/ledger/grain";
+import dedent from "../util/dedent";
+import {loadCredGraph} from "./common";
+
+import * as GrainConfig from "../api/grainConfig";
+import type {Command} from "./command";
+
+function die(std, message) {
+  std.err("fatal: " + message);
+  return 1;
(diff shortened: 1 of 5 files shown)


blueridger:
Re: test plan. What did you verify? What was the expected similarity or difference between the command outputs that you looked for?

decentralion (approved):
Agree w/ @blueridger that it'd be nice to see a more specific test plan. Approving since the implementation looks good.

elihanover:
Update: I realized that the `CredGraph` interface was sufficient for the `grain2` command without lifting it into a `CredGrainView`, which is reflected in this amended commit.

Regarding tests: I manually compared the ledger output using `JSON.stringify` and also ran the -s flag to ensure it works.

blueridger (approved):
So this is fine for now. What we really ought to do is replace the CredAccountData, AllocationIdentity, and ProcessedIdentities structures with CredGrainView. However that would involve too much forking if we tried to do that in parallel with grainV1, so let's do the migration using what you have now, and worry about the refactor after.

elihanover:
Alright, I'll merge.  Let me know when you think it's a good time to refactor.

Issue #2667, opened 2021-01-18
Outside-in Refactoring Battle Plan


hammadj:
In an effort to refactor / cleanup the core codebase before our push for new features like the creditor and [new plugin system](https://github.com/sourcecred/sourcecred/issues/2409), I plan to tackle the "outside-in" refactor. In other words, improving out external facing APIs and code / repo organization.

Main Tasks:
- [ ] Improve out NPM package build output to include flow type information so that its easier for people to use the NPM package without having to refer to the source code. Stretch goal: figure out if we can have TS types included as well with minimal maintenance overhead.
- [ ] Update codebase to use [Lerna](https://github.com/lerna/lerna) + [Yarn Workspaces](https://classic.yarnpkg.com/en/docs/workspaces/). Initially with just a single package that contains our current codebase, and then moving out pieces like the UI / Frontend and eventually plugins
- [ ] Prototype a "micro plugin" as a separate package and spec out what changes are needed to the current plugin system to accomplish that

PR #2674, opened 2021-01-20
Add source files to published NPM package

hammadj:
Related: #2667 

# Description

This allows editors to show code completion, highlighting, and makes it easy for users to know what options / params
are available when consuming the sourcecred API from the NPM package. This is accomplished by including the src/ folder
in the list of files included in the NPM package.

Adding the ["module" field](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md) in the package.json also enables users to use the `import sourcecred from 'sourcecred';` syntax and get the same code completion / highlighting benefits as using `const sourcecred = require('sourcecred');`.

Lastly, setting `libraryExport: "default"` in webpack.config.api.js makes it so users dont need to write
`const sourcecred = require('sourcecred').default;` when importing the package.

# Test Plan

Install the test package that's bundled using this new setup and see if code completion / highlighting is working
correctly: `yarn add sourcecred-publish-test`

Here's some gifs of me testing this out:

Before:

After:

### DIFF
--- a/config/webpack.config.api.js
+++ b/config/webpack.config.api.js
@@ -33,6 +33,7 @@ module.exports = ({
     chunkFilename: "[name].[chunkhash:8].chunk.js",
     libraryTarget: "umd",
     library: "sourcecred",
+    libraryExport: "default",
     // Use `this` for compatibility with both Node and browser.
     globalObject: "this",
   },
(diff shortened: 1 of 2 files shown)


blueridger (approved):

topocount (approved):

wchargin (commented):
I see that @blueridger and @topocount have approved; anything in
particular you want my review on?

hammadj:
@wchargin Just tagged you since DL mentioned to get your sanity check for webpack stuff / if you're cool with this approach. Will merge, thanks!

PR #2679, opened 2021-01-22
Add Ethereum plugin to public API

hammadj:
# Description

Allows programmatic creation / linking of identities from the SourceCred NPM package

# Test Plan

Ensure CI tests pass and Ethereum plugin is available in packaged output (`npm pack`)

e.g. 
```js
import sourcecred from 'sourcecred';

const identities = sourcecred.plugins.ethereum.utils.identity.createIdentities([...])
```

### DIFF
--- a/src/plugins/ethereum/utils.js
+++ b/src/plugins/ethereum/utils.js
@@ -1,0 +1,16 @@
+// @flow
+
+import {createIdentity, createIdentities} from "./createIdentities";
+import {parseAddress, truncateEthAddress} from "./ethAddress";
+import {nodeAddressForEthAddress} from "./ethAddressNode";
+
+export const address = {
+  parseAddress,
+  truncateEthAddress,
+  nodeAddressForEthAddress,
+};
+
+export const identity = {
+  createIdentities,
+  createIdentity,
+};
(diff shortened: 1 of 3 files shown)


src/plugins/ethereum/utils.js
```diff
@@ -0,0 +1,16 @@
+// @flow
```
blueridger, line 1: same as https://github.com/sourcecred/sourcecred/pull/2682#discussion_r562988134

src/plugins/ethereum/utils.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const address = {
+  parseAddress,
+  truncateEthAddress,
+  nodeAddressForEthAddress,
+};
+
+export const identity = {
+  createIdentities,
+  _createIdentity,
```
blueridger, line 15: the `_` indicates to me an attribute intended to be private. Should it be renamed?

src/plugins/ethereum/utils.js
hammadj: Same answer https://github.com/sourcecred/sourcecred/pull/2682#discussion_r562990467

src/plugins/ethereum/utils.js
hammadj: Yea I see no reason for it to be private, its already being exported in the file its in

blueridger (approved):

PR #2682, opened 2021-01-22
Add Ledger utils to public API

hammadj:
# Description

Allows users of the NPM package to use helpful functions like diffLedger, computeDistribution,
ensureIdentityExists, etc.

# Test Plan

Ensure CI tests pass and Ledger utils are available in packaged output (npm pack)

### DIFF
--- a/src/core/ledger/utils.js
+++ b/src/core/ledger/utils.js
@@ -1,0 +1,14 @@
+// @flow
+
+import {applyDistributions} from "./applyDistributions";
+import {computeDistribution} from "./computeDistribution";
+import {computeCredAccounts} from "./credAccounts";
+
+export {diffLedger} from "./diffLedger";
+export {ensureIdentityExists} from "./identityProposal";
+
+export const distributions = {
+  applyDistributions,
+  computeDistribution,
+  computeCredAccounts,
+};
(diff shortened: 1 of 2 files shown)


blueridger (approved):

src/core/ledger/utils.js
```diff
@@ -0,0 +1,14 @@
...
+import {computeCredAccounts} from "./credAccounts";
+
+export {diffLedger} from "./diffLedger";
+export {ensureIdentityExists} from "./identityProposal";
+
+export const distributions = {
+  applyDistributions,
+  computeDistribution,
+  computeCredAccounts,
+};
```
blueridger, line 14: This seems like a strictly API-related file, and it looks to be a deviation from our current pattern of directly importing the modules into the api/index.js file and then composing them into a structure there. What's the justification for creating this file?

hammadj: I wanted these "utils" files to be the single place where a certain module can selectively export the various useful helper functions that may exist (i.e. the "external API") and have it colocated in the actual module folder. Otherwise the `api/index.js` file gets polluted with context specific knowledge of what functions should be exported from each of the different modules.

A cleaner way to do it would be to take it another level up where each module just has an index.js file that acts as the external facing API and the api/index.js file becomes even more generalized, however I didn't want to break any existing code / scripts that people might be using since we will likely be doing a more drastic restructure of our APIs once we move to the monorepo structure, so I wanted to wait for that before doing a big uplift.

yay runon sentences

blueridger (approved):

PR #2686, opened 2021-01-23
cli/grain(2): add deprecation info for grain and grain2 commands

elihanover:
This commit adds context around the deprecation of the grain
command in favor of "grain2", which does not rely on the soon
to be deprecated `CredView`.

### DIFF
--- a/src/cli/grain.js
+++ b/src/cli/grain.js
@@ -23,6 +23,13 @@ function die(std, message) {
   return 1;
 }
 
+/**
+ * The grain command is soon to be deprecated, as part of a transition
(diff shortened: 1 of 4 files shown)


blueridger (commented):
Could you add comments to the other 2 forked methods too? They could be hanging around for up to a couple months.

hammadj (approved):
Should also add comment to `computeCredAccounts2` and `applyDistributions2`, otherwise looks good

elihanover:
@blueridger you want a last look before I merge?

blueridger (approved):
good enough

PR #2693, opened 2021-01-24
Make numIntervalsLookback field in Immediate policy config optional

hammadj:
# Description
Ledgers that had Immediate policy distributions before the numIntervalsLookback feature was added
don't have this field in their logs, which results in a runtime error that breaks backwards compatibility.
This fixes it by making this field optional and defaulting it to 1 when it's not defined.

# Test Plan

Ensure CI tests pass and ledgers with old Immediate distributions load without errors


src/core/ledger/policies/immediate.js
```diff
@@ -65,11 +66,13 @@ export const immediateConfigParser: P.Parser<ImmediatePolicy> = P.object({
...
-  budget: grainParser,
-  numIntervalsLookback: P.number,
-});
+export const immediatePolicyParser: P.Parser<ImmediatePolicy> = P.object(
+  {
+    policyType: P.exactly(["IMMEDIATE"]),
+    budget: grainParser,
+  },
+  {numIntervalsLookback: P.number}
+);
```
blueridger, line 75: Why did this change? What are the semantics going on here?

src/core/ledger/policies/immediate.js
```diff
@@ -23,7 +23,7 @@ export type Immediate = "IMMEDIATE";
 export type ImmediatePolicy = {|
   +policyType: Immediate,
   +budget: NonnegativeGrain,
-  +numIntervalsLookback: number,
+  +numIntervalsLookback?: number,
```
elihanover, line 26: I had this optional, but DL wanted policy fields to be required.  Maybe just have the parser set this as one if missing.

src/core/ledger/policies/immediate.js
elihanover: Maybe have the `numIntervalsLookback` parser default to 1 if missing.

elihanover (commented):
Makes sense to me.  I think @decentralion had strong opinion about the field being required in the policy, but shouldn't be a problem with the right parser.

decentralion (commented):
Good catch and fix @hammadj -- we really need to incorporate loading real instances into our CI so we can start catching these compat breaks.

I agree with @elihanover's comment though, let's keep it required on the policy type, but make it optional in the parser (with a default to 1).

This should also allow a much smaller change overall.

src/core/ledger/policies/immediate.js
hammadj: Moving numIntervalsLookback to the second param makes it optional instead of required, which prevents breaking old ledger entries that didn't have it defined

blueridger (approved):

src/core/ledger/policies/immediate.js
```diff
@@ -65,11 +67,22 @@ export const immediateConfigParser: P.Parser<ImmediatePolicy> = P.object({
...
+    },
+    {
+      numIntervalsLookback: P.number,
+    }
+  ),
+  (policy) => ({
+    ...policy,
+    numIntervalsLookback:
+      policy.numIntervalsLookback != null ? policy.numIntervalsLookback : 1,
+  })
```
elihanover, line 84: LGTM.  Nit: maybe kick this out into its own function so the parser's fmap can be more descriptive calling something like `upgradeLookback`.

elihanover (approved):

PR #2695, opened 2021-01-25
Core: Implement WeightedGraph2

topocount:
WeightedGraph2 is a class that encapsulates `WeightsI` and `Graph`. It
provides an API that is similar to Graph in both form and expected
behavior, except it returns a calculated weight along with each node
and edge.

It also allows address prefixes to have weights assigned, which allows
for weight values to compound beyond an addresses' configured weight.
This behavior is actually provided via the algorithm's `WeightsEvaluator`,
but this API allows for more intuitive configuration

test plan: yarn unit

Tests cover the entire class and verify the functionality described here
and in the docstrings

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038310768785) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/core/weightedGraph2.js
+++ b/src/core/weightedGraph2.js
@@ -2,20 +2,40 @@
 
 import {
   type Edge,
+  type EdgesOptions,
   type Node,
   type EdgeAddressT,
   type NodeAddressT,
+  type GraphComparison,
+  type GraphJSON,
+  compareGraphs,
   Graph,
 } from "./graph";
 import {type NodeWeight} from "./weights/nodeWeights";
 import {type EdgeWeight} from "./weights/edgeWeights";
-import {type WeightsComparison} from "./weights/weightsT";
-import {NodeTrie, EdgeTrie} from "./trie";
+import {
+  type WeightsComparison,
+  type WeightsJSON,
+  compareWeightsT,
+} from "./weights/weightsT";
+import {
+  Weights,
+  type WeightsI,
+  fromJSON as fromWeightsJSON,
+} from "./weights/weights";
+import {toCompat, fromCompat, type Compatible} from "../util/compat";
 
 /**
  * An entry in the array returned by AddressModule.toParts()
  */
 export type AddressPart = string;
+
+export type WeightedGraphJSON = {|
+  graph: GraphJSON,
+  weights: WeightsJSON,
+|};
+
+export type CompatibleWeightedGraphJSON = Compatible<WeightedGraphJSON>;
 
 export type WeightedNode = {|
   +node: Node,
@@ -27,6 +47,14 @@ export type WeightedEdge = {|
   +weight: EdgeWeight,
 |};
 
+export type WeightedGraphComparison = {|
+  graphComparison: GraphComparison,
+  weightsComparison: WeightsComparison,
+  weightedGraphsAreEqual: boolean,
+|};
+
+const COMPAT_INFO = {type: "sourcecred/weightedgraph", version: "0.1.0"};
+
 /**
  * The WeightedGraph2 class is a replacement for the WeightedGraph type. This class
  * eliminates the need give every node and edge that exists a weight. Rather,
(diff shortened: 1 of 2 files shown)


src/core/weightedGraph2.js
```diff
@@ -37,89 +65,252 @@ export type WeightedEdge = {|
  * WeightedGraph2 should completely replace it.
  */
 export class WeightedGraph2 {
-  _nodeWeights: NodeTrie<NodeWeight>;
-  _edgeWeights: EdgeTrie<EdgeWeight>;
-  graph: Graph;
+  _weights: WeightsI;
+  _graph: Graph;
+
+  constructor(graph?: Graph, weights?: WeightsI) {
```
elihanover, line 71: It makes sense to take `WeightsI` as this is one layer of abstraction higher, but it might be useful to provide some transformation from `NodeWeightsT` and `EdgeWeightsT` (or `WeightsT`). Those types are all data types rather than interfaces, so you can think of this as lifting data into `WeightedGraph2` context, rather than some other context.

```js
(Graph, NodeWeightsT, EdgeWeightsT)  => WeightedGraph2
// or
(Graph, WeightsT)  => WeightedGraph2
```

src/core/weightedGraph2.js
```diff
@@ -37,89 +65,252 @@ export type WeightedEdge = {|
  * WeightedGraph2 should completely replace it.
  */
 export class WeightedGraph2 {
```
elihanover, line 67: It would be nice to have some interface that describes what `WeightedGraph2` does, similar to `WeightsI` or `NodeWeightsI`.

elihanover, line 67: From our first refactoring call, I get the sense that composability will matter a lot with these types.  I find that it's easier / more concise to compose objects than classes (in that classes are just objects with more complications), but that's just me.  If there's a benefit of using a class here, fine by me.  If not, I lean towards the style used in `Weights` and `Node/EdgeWeights`.

src/core/weightedGraph2.js
```diff
@@ -37,89 +65,252 @@ export type WeightedEdge = {|
...
-  _nodeWeights: NodeTrie<NodeWeight>;
-  _edgeWeights: EdgeTrie<EdgeWeight>;
-  graph: Graph;
+  _weights: WeightsI;
+  _graph: Graph;
```
elihanover, line 69: My sense is that these should be immutable.  Another advantage of using an object is that it would properly shield this data in its closure.

elihanover (commented):
We had some back and forth about using classes in the `Weights` PR, and my sense is that we should get on the same page for consistency.  My general m.o. is that if you don't need 'em, don't use em, but I'd like to get your perspective before I conclude anything.

src/core/weightedGraph2.js
```diff
@@ -37,89 +65,252 @@ export type WeightedEdge = {|
...
 
   /**
    * Set the weight for a node address prefix.
    */
-  setNodePrefixWeight(
-    _unused_prefix: NodeAddressT,
-    _unused_w: NodeWeight
-  ): this {
-    throw new Error("method not implemented");
+  setNodePrefixWeight(prefix: NodeAddressT, w: NodeWeight): this {
```
elihanover, line 79: `setNodePrefixWeight` _is_ `setNodeWeight` on `WeightsI`.  It seems redundant to rewrite the function here, when we can simply set it:
```js
// Even with classes, you can just set this in the constructor
constructor(graph?: Graph, weights?: WeightsI) {
  ...
  setNodePrefixWeight = weights.setNodeWeight;
  setEdgePrefixWeight = weights.setEdgeWeight;
}
```

src/core/weightedGraph2.js
```diff
@@ -37,89 +65,252 @@ export type WeightedEdge = {|
...
+   */
+  static merge(wgs: Iterable<WeightedGraph2>): WeightedGraph2 {
+    const graphs: Array<Graph> = [];
+    const weights: Array<WeightsI> = [];
+    for (const wg of wgs) {
+      graphs.push(wg._graph);
+      weights.push(wg._weights);
+    }
+    return new WeightedGraph2(Graph.merge(graphs), Weights().merge(weights));
   }
```
elihanover, line 276:
```js
// This seems cleaner to me
const graphs = wgs.map({_graph} => _graph)
const weights = wgs.map({_weights} => _weights)
```
```js
// If you're feeling frisky
const graphs = wgs.map(prop("_graph"))
const weights = wgs.map(prop("_weights"))
```

src/core/weightedGraph2.js
topocount: @elihanover these can't be immutable because WG2 operates on them. WG2 is meant to be make it easy to build a WeightedGraph and not just read a WeightedGraph.

topocount: One thing I'm not sure about: Do we want to utilize a `copy` of each of these instead of the original reference?

src/core/weightedGraph2.js
topocount: My intuition was only to spec out an interface if it is meant to be utilized by other classes, but I like this idea,

elihanover: I find it helpful to understand the code on a high level quickly, even if it's just for that

src/core/weightedGraph2.js
elihanover: That's how I did `Weights`.  You know you're working with a copy each time, so you don't run into any issues of handling/sharing state in multiple places.  Seems more pure.

elihanover: That's how I did `Weights`.  You know you're working with a copy each time, so you don't run into any issues of handling/sharing state in multiple places.  Seems more pure.

src/core/weightedGraph2.js
topocount: @elihanover  I considered this but at the risk of being pedantic, this syntax is not valid because `Iterable`s do not have access to array methods. 

Moreover, this suggestion requires two full iterations over the `Iterable`, instead of one, and I wanted to optimize this a bit.

elihanover: Gotcha, makes sense to me.

src/core/weightedGraph2.js
topocount: @decentralion would be curious to get your opinion on this.

I was trying to follow the structure of the `Graph` class moreso than `WeightsI`, unless `Graph` is scheduled to be refactored to follow `WeightsI`'s object-literal return instead of an ES6 class

decentralion (commented):
Looks pretty good overall. Most of my comments have to do with the comments rather than the implementation. However, there is one serious issue; I believe that weighted graphs produced by this module are not compatible with weighted graphs from the existing module. Please write a test that serializes a WG1 and tries to deserialize with WG2 logic, and vice versa.

While you're at it--let's please add a method that converts WG1 to WG2 and vice versa (both can live in this module), and let's check that the round trip is equality.

PR #2696, opened 2021-01-25
Convert Explorer-Home to use CredGrainView

BrianBelhumeur:
# Description
Per issue #2464, we're migrating the new Explorer Admin UI view from using CredView to using CredGrainView. This change will allow easy time period selection (this week, last week, last month, all time, etc), however this PR does not address the UI functionality of doing so. That will be in a future PR.

# Test Plan
I tested this change using the template-instance backend that I had generated a credGraph.json file (via `yarn credrank`).

This is the screenshot of the output. The graph at the top is the summary of all cred over time.
Screen Shot 2021-01-25 at 11 54 34

I also tested running the admin UI without a credGraph.json file (via the normal backed on this repo) and it correctly displays a blank explorer-home page, while allowing navigation to the other parts of the admin UI. We'll need a proper "nothing to see here, cuz there's no credGraph.json file" type message eventually.


KuraFire:
> I also tested running the admin UI without a credGraph.json file (via the normal backed on this repo) and it correctly displays a blank explorer-home page, while allowing navigation to the other parts of the admin UI. We'll need a proper "nothing to see here, cuz there's no credGraph.json file" type message eventually.

What are the circumstances that there is no `credGraph.json` file?

KuraFire:
Oh, and as for the UI in the test output: the Cred by Plugin section can be removed, we're pushing back on that for now.

blueridger:
> What are the circumstances where there is no `credGraph.json` file?

When serving a freshly cloned instance. This allows maintainers to carry out a lot of their admin tasks without having to somehow compute/load scores.

blueridger (approved):
This is lacking a lot of the React state management that I'd expect is needed to reduce re-renders, but that can be added in your next PR with TableState.

src/ui/components/AdminApp.js
```diff
@@ -67,8 +70,14 @@ const createAppLayout = ({hasBackend, currency}: LoadSuccess) => {
...
   credView: CredView | null,
   hasBackend: Boolean,
-  currency: CurrencyDetails
+  currency: CurrencyDetails,
+  credGraph: CredGraph | null,
+  ledgerManager: LedgerManager
 ) => {
+  const credGrainView = credGraph
+    ? new CredGrainView(credGraph, ledgerManager.ledger)
+    : null;
```
blueridger, line 79: Is this something that should be memoized as well @hammadj ?

KuraFire:
> > What are the circumstances where there is no `credGraph.json` file?
> 
> When serving a freshly cloned instance. This allows maintainers to carry out a lot of their admin tasks without having to somehow compute/load scores.

Cool, makes sense. LGTM 👍🏼

KuraFire (approved):

BrianBelhumeur:
@blueridger yes that's the plan. This is the refactor PR that's necessary to put the TableState in place. When I do that PR is when I'll do memoization and optimization. Thanks for keeping attention on performance concerns.

Issue #2646, opened 2021-01-11
Make the Transfer view mobile-friendly


topocount:
the transfer view works on desktop but doesn't fit into a portrait view or mobile view

felixbbertrand:
Can you assign me this ? @topocount

felixbbertrand:
@KuraFire

KuraFire:
Thanks @felixbbertrand & @thr3m. I left my comments in the other PR:

https://github.com/sourcecred/sourcecred/pull/2699#issuecomment-767988087

PR #2699, opened 2021-01-26
making the view of transfer mobile friendly

felixbbertrand:
# Description

  Making the transfer view  mobile friendly.

# Test Plan

  No test has been added, since its basically a css change
  
  
  Before changes: 
image

  
  After changes :
image

  paired with @thr3m

### DIFF
--- a/src/ui/components/Transfer.js
+++ b/src/ui/components/Transfer.js
@@ -81,12 +81,11 @@ export const Transfer = ({
   const [memo, setMemo] = useState<string>("");
   const isXSmall = useMediaQuery((theme) => theme.breakpoints.down("xs"));
 
-  const isDisabled = (
-      !Number(amount) ||
-      !(sender && receiver) ||
-      gt(fromFloatString(amount), sender.balance) ||
-      lt(fromFloatString(amount), fromFloatString("0"))
-    );
+  const isDisabled =
+    !Number(amount) ||
+    !(sender && receiver) ||
+    gt(fromFloatString(amount), sender.balance) ||
+    lt(fromFloatString(amount), fromFloatString("0"));
 
   const submitTransfer = () => {
     if (sender && receiver) {
--- a/src/ui/components/Transfer.js
+++ b/src/ui/components/Transfer.js
@@ -22,14 +22,27 @@ const useStyles = makeStyles((theme) => ({
     padding: "5px 20px",
     display: "flex",
     alignItems: "center",
+    width: "200px",
   },
   triangle: {
     width: 0,
     height: 0,
+    top: "100%",
     background: theme.palette.background,
     borderTop: "30px solid transparent",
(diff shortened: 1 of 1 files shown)


felixbbertrand:
#2646

KuraFire:
Awesome work, both of you! From the screenshot it looks like some of the button state interactions have already been improved, too. 🥳  

Can you provide a screen capture of the interactions we discussed on a call yesterday? Or verify for me that the Transfer Grain button remains disabled until all three of From, To, and Amount have a valid value?

To recap, my requested UX improvements (since you're adjusting the Transfer Grain screen):

1. Add a visual 'arrow' pointing down (could be just a small CSS arrow using an `:after` pseudo-element ([example for this technique](http://html-tuts.com/css-arrows-up-down-left-right-triangle/)) to the Amount field. In other words, a triangle pointing down below the Amount gray area would suffice :)
2. Have the "Save Ledger" button in `:disabled` state until a Grain transfer has been submitted, at which point there is also…
3. A message above the "Save Ledger" button in a bold white color, reading: "Changes not yet saved to ledger!"

KuraFire (changes requested):
See my previous comment for requested changes. Beyond those, looks great! 😎 👍🏼

blueridger:
@KuraFire i had suggested they separate your point 2 and 3 into a separate PR, since those improvements aren't mobile-specific and really represent a distinct feature increment from what this PR accomplishes.

felixbbertrand:
@KuraFire yeah @blueridger told us that points 2-3 should be done into another PR which makes sense.

felixbbertrand:
pr

src/ui/components/Transfer.js
```diff
@@ -57,6 +79,16 @@ export const Transfer = ({
...
+  const isXSmall = useMediaQuery((theme) => theme.breakpoints.down("xs"));
+
+  const disabled = (amount) => {
+    return (
+      !Number(amount) ||
+      !(sender && receiver) ||
+      gt(fromFloatString(amount), sender.balance) ||
+      lt(fromFloatString(amount), fromFloatString("0"))
+    );
+  };
```
hammadj, line 91: Why does this need to be a function if it's only used once? Can just do:

```suggestion
  const isDisabled = (
      !Number(amount) ||
      !(sender && receiver) ||
      gt(fromFloatString(amount), sender.balance) ||
      lt(fromFloatString(amount), fromFloatString("0"))
    );
```

hammadj (approved):
Looks good, minor nit

src/ui/components/Transfer.js
felixbbertrand: Unsure, is there a benefit between making it a function rather than a const?

felixbbertrand (commented):
@KuraFire  I'm stuck with requested changes even tho @hammadj approved these changes

PR #2700, opened 2021-01-26
server: provide CORS headers

decentralion:
This modifies the sourcecred serve command so that it provides CORS headers.

This way it will be possible to directly depend on SC data from within observable notebooks.

Test plan: Prior to this change, attempting to load JSON in Observable from the server fails to fetch with a CORS issue. After this change, it works.

You can use this notebook to test locally: https://observablehq.com/@decentralion/localhost-cors-fetch-test

### DIFF
--- a/src/cli/serve.js
+++ b/src/cli/serve.js
@@ -1,6 +1,7 @@
 // @flow
 
 import type {$Response as ExpressResponse} from "express";
+import cors from "cors";
 
 import type {Command} from "./command";
 import {loadInstanceConfig} from "./common";
@@ -29,6 +30,7 @@ const serveCommand: Command = async (args, std) => {
   }
 
   const server = express();
+  server.use(cors());
 
   // override static config to enable ledger updates
   server.get(
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "bottleneck": "^2.19.5",
     "chalk": "^4.0.0",
     "commonmark": "0.29.1",
+    "cors": "^2.8.5",
     "d3-array": "^2.4.0",
     "d3-format": "^2.0.0",
     "d3-scale": "^3.2.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3338,6 +3338,14 @@ core-util-is@1.0.2, core-util-is@~1.0.0:
   resolved "https://registry.yarnpkg.com/core-util-is/-/core-util-is-1.0.2.tgz#b5fd54220aa2bc5ab57aab7140c940754503c1a7"
   integrity sha1-tf1UIgqivFq1eqtxQMlAdUUDwac=
 
+cors@^2.8.5:
+  version "2.8.5"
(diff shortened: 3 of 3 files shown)


hammadj (commented):
image

Still getting a CORS error

topocount (approved):
This actually works for me by doing the following:
```bash
# in the sourcecred repo
git checkout 2021-01-25-CORS-server
yarn && yarn build
# move to a cred instance
cd ../sc-cred
path/to/sourcecred.js serve
# navigate to the observableHQ notebook linked above
```

blueridger (approved):

src/cli/serve.js
```diff
@@ -29,6 +30,7 @@ const serveCommand: Command = async (args, std) => {
   }
 
   const server = express();
+  server.use(cors());
```
wchargin, line 33: This introduces a security vulnerability.

Your blanket `cors()` applies to all routes, including the `POST` route
for `/data/ledger.json`. This means that if you have an instance admin
server running, and you visit *any website*, then that website can
overwrite your ledger with arbitrary data.

Repro: run the `cd sharness/__snapshots__/test-instance` instance and
open both <http://localhost:6006> and <https://example.com>. Execute the
following JavaScript statement in your console:

```javascript
await fetch("http://localhost:6006/data/ledger.json", {method: "POST", body: "test"})
```

Before this patch, this works from localhost (same-origin) but fails
from example.com. By “fails”, I mean both that the promise rejects and
also that the file is not overwritten. After this patch, example.com is
able to overwrite the data.

The same-origin policy is a security mechanism. Please be careful and
thoughtful when working around it.

wchargin (changes requested):
Blocking until security bug fixed.

topocount:
@wchargin at the minimum I suppose the solution for this would be to only allow `GET` requests. However, to further refine this, I believe we should explicitly whitelist observablehq.com, no?

wchargin:
I don’t think that we should whitelist Observable or give it any special
treatment. In case of security bug, that just lowers the scope to
“anyone who wants to pwn you just has to create an Observable and link
you to it”. And baking knowledge about random third-party software
platforms into our API server feels pretty icky to me.

Only allowing GET (and HEAD, etc.) requests is fine if we’re comfortable
exposing all the data to the network, since these methods are required
by standard to be [safe].

This would still permit any website to _read_ all SourceCred data from
any instance serving on your computer, even though it may contain
sensitive or private information. That’s kind of required for the
purpose of this change, so I’m not opposed to it, but I think that it’s
worth at least considering whether to guard this behavior behind an
off-by-default flag. That is, run `sourcecred serve` and your data is
still safe; run `sourcecred server --public` or something and anyone can
read it.

(I’m not sure what the best flag name is. We’d want to avoid confusion
with the distinction between binding loopback vs. all addresses; we
already bind to `::0`, so it’s already public to LAN. And `--public`
alone might not be scary enough, since it has a neutral-to-positive
connotation in general. Maybe `--expose-to-internet`?)

[safe]: https://tools.ietf.org/html/rfc7231#section-4.2.1

PR #2711, opened 2021-01-31
Setup separate API builds for web and node environments

hammadj:
# Description

The current API build is designed to only work in web browsers and not a NodeJS environment in the backend.
In order to allow API access to SourceCred from any environment, this fix creates two separate webpack
builds for web and node environments so that things like `fetch` are correctly polyfilled for the intended
environment. Previously trying to import and use the SourceCred API in a node environment would cause a
fatal error.

# Test Plan

Ensure `yarn build:api` runs correctly and two separate builds are created in dist/server/api.js
and dist/client/api.js.

### DIFF
--- a/config/webpack.config.api.js
+++ b/config/webpack.config.api.js
@@ -12,7 +12,7 @@ const getClientEnvironment = require("./env");
 
 const env = getClientEnvironment(null);
 
-module.exports = ({
+const baseConfig = {
   // Don't attempt to continue if there are any errors.
   bail: true,
   node: {
@@ -73,4 +73,26 @@ module.exports = ({
     new webpack.DefinePlugin(env.individuallyStringified),
   ],
   mode: process.env.NODE_ENV,
-} /*: any */);
+};
+
+const client = {
+  ...baseConfig,
+  target: "web",
+  output: {
+    ...baseConfig.output,
+    path: `${paths.apiBuild}/client`,
+    libraryTarget: "umd",
+  },
+};
+
+const server = {
+  ...baseConfig,
+  target: "node",
+  output: {
+    ...baseConfig.output,
+    path: `${paths.apiBuild}/server`,
+    libraryTarget: "commonjs",
+  },
+};
+
+module.exports = ([client, server] /*: Array<any>*/);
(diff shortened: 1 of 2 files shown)


config/webpack.config.api.js
```diff
@@ -73,4 +73,26 @@ module.exports = ({
...
+  ...baseConfig,
+  target: "node",
+  output: {
+    ...baseConfig.output,
+    path: `${paths.apiBuild}/server`,
+    libraryTarget: "commonjs",
+  },
+};
+
+module.exports = [client, server];
```
topocount, line 98: This needs an explicit type annotation or flow is not happy on my machine: 
```suggestion
module.exports = ([client, server] /*: Array<any>*/);
```

topocount (commented):
This is a relatively simple change, but it impacts at least a couple environments. Could you outline (and/or better yet link to) some tests in both Observable and Node. It'd be good to document the reproductions line by line in case other bugs come up and to ensure the steps are clear.

config/webpack.config.api.js
hammadj: Flow passes in CI, wondering why it's failing on your machine. Reluctant to cast to any if it's not absolutely necessary. Do you maybe have an old version of flow or something?

hammadj:
> This is a relatively simple change, but it impacts at least a couple environments. Could you outline (and/or better yet link to) some tests in both Observable and Node. It'd be good to document the reproductions line by line in case other bugs come up and to ensure the steps are clear.

Here's an Observable notebook: https://observablehq.com/@hammadj/sourcecred-ledger-viewer

Note that there is a breaking change where the import needs to be changed from `sc = (await require("sourcecred@latest")).default` to `sc = (await require("sourcecred@latest"))`, but the actual bundle loads and works fine.

And for node, here's a repl.it: https://repl.it/@hammadj1/SourceCred-Test

config/webpack.config.api.js
topocount: Now that [types-first](https://flow.org/en/docs/lang/types-first/) is the default (as of version 0.134), file exports cannot be implicit `any`. 
```
❯ yarn flow --version
yarn run v1.22.10
$ flow --version
Flow, a static type checker for JavaScript, version 0.135.0
Done in 0.19s.
```

topocount (approved):

PR #2712, opened 2021-01-31
Fix createIdentity in discord plugin and add tests

hammadj:
# Description

Previously if the `member.nick` field was undefined or an empty string, the createIdentity function
would throw a "Cannot read properly 'slice' of undefined" error instead of falling back to the username.

Additionally, the code was previously untested so I added tests to ensure correctness in the future.

# Test Plan

Ensure CI tests pass

### DIFF
--- a/src/plugins/discord/models.js
+++ b/src/plugins/discord/models.js
@@ -64,7 +64,7 @@ export type User = {|
 
 export type GuildMember = {|
   +user: User,
-  +nick: string | null,
+  +nick: ?string,
   +roles: $ReadOnlyArray<Snowflake>,
 |};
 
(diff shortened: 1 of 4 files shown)


src/plugins/discord/models.js
```diff
@@ -64,7 +64,7 @@ export type User = {|
 
 export type GuildMember = {|
   +user: User,
-  +nick: string | null,
+  +nick: string | ?null,
```
decentralion, line 67: `string | ?null` is just a more confusing way of saying `?string`

decentralion (approved):

src/plugins/discord/models.js
hammadj: Ah yea, I forget in Flow that ? includes null and undefined. In TypeScript null and undefined have to be specified separately. Will fix

Issue #2237, opened 2020-09-10
New SC Explorer Dashboard: Stats Circles


KuraFire:
Atomic-style ticket for #2227 — Stats Circles component:

![SourceCred Explorer Dashboard - 3  Stats Circles@2x](https://user-images.githubusercontent.com/33087/92780941-88710000-f358-11ea-807a-642ca21227e4.png)

### Specifications:

- Circles are 128px × 128px to start
- Circle colors are:
   - `#6174CC` Blue for **Cred**
   - `#FFAA3D` Yellow for **Grain**
   - ~~`#FDBBD1` Pink for **Participants**~~
   - `#4BD76D` Green for **Grain Per Cred**
- Text inside the Circles should resize to fit on one line; suggested font size ranges:
   - min. size: 21px
   - max. size: 34px
- Text labels below the Circles are in 21px to start. Upon implementation we may adjust. 
- Exact label names for the Circles (note capitalization):
   - Cred This Week
   - Grain Harvested
   - ~~Active Participants~~
   - Grain per Cred
- Spacing between the Circles:
   - `margin-left: 0` (left-aligned to the same container that holds all other content)
   - `margin-right: 0` (right-aligned to the same container)
   - flexible spacing in-between the circles, with a **minimum gutter width of 20px**
- Spacing below Circles (above labels): **13px**

**Update 2021-02-05:** The Active Participants circle is to be taken out for now, pending future design changes.

blueridger:
Shouldn't have closed. New AC:
- top stats are NOT affected by filtering/pagination

Not AC:
- top stats will be affected by Timeframe Control only

KuraFire:
Addendum:

The Active Participants circle is to be taken out for now, pending future design changes.

blueridger:
V1 should not include grain per cred

BrianBelhumeur:
To clarify:
- We will only have two circles, "Cred This Week" and "Grain Harvested"?
- Since the stats will be affected by the timeframe control, I assume the label on "Cred This Week" should update to reflect the interval being shown

KuraFire:
@BrianBelhumeur both of these assumptions are correct. The Stats circles should reflect the Timeframe Control's / top chart's selected period of time. And to start, we'll only have two Stats circles for now.

PR #2715, opened 2021-02-02
Pagination for explorer home

BrianBelhumeur:
# Description
This PR adds pagination, filtering, and sorting for the Activity Table on Explorer Home via the TableState util. Sorting can be done via cred and grain, and is done by clicking on their header labels on the table. Filtering can be done by name, entering text into the box at the top right, or by type of participant (participant, organization, bot, project) with the checkboxes at the bottom. Also, the Explorer Home page now uses the proper currency name and suffix, loaded from config (defaults to "Grain" and "g", respectively).

Closes issue #2464 

# Test Plan
Tested manually with the following workflows
- Load page and confirm that default sort is by cred, descending
- Click on cred header and confirm that sort is now by cred, ascending
- Click on grain header and confirm that sort is by grain, ascending (assumed since grain is 0 for all users in data set)
- Click on grain header again and confirm that sort is by grain, descending (assumed since grain is 0 for all users in data set)
- Choose different pagination options and confirm that the proper number of rows are shown
- Click forward and back on the pages, and confirm that the proper page of results is shown
- Type letters into the name filter at top and confirm that only names that match filter string are shown (same letters, in order, but not necessarily sequential)
- Confirm that averages and totals for cred and grain are correct for visible list entries filtered by name
- Check various checkboxes for participant type at bottom of list, and confirm that only participants of said type are visible
- Confirm that averages and totals for cred and grain are correct for visible entries filtered by type


src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -88,77 +98,164 @@ const useStyles = makeStyles((theme) => ({
...
+    {
+      title: `${currencyName}`,
+      value: `6,765${currencySuffix}`,
+      className: classes.grainCircle,
+    },
+    {
+      title: "Active Participants",
+      value: allParticipants.length,
+      className: classes.participantCircle,
+    },
```
blueridger, line 199: Active participants should not be included for now.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -88,77 +98,164 @@ const useStyles = makeStyles((theme) => ({
...
-  let credTimelineSummary = initialView.intervals().map(() => 0);
+  if (tsParticipants.currentPage.length > 0) {
+    credAndGrainSummary.avgCred =
+      credAndGrainSummary.totalCred / tsParticipants.currentPage.length;
+    credAndGrainSummary.avgGrain =
+      credAndGrainSummary.totalGrain / tsParticipants.currentPage.length;
+  } else {
+    credAndGrainSummary.avgCred = 0;
+    credAndGrainSummary.avgGrain = 0;
+  }
```
blueridger, line 186: I'm pretty sure you could put this in a `useEffect` react wrapper that watches `tsParticipants.currentPage` to prevent it from re-aggregating needlessly.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -88,77 +98,164 @@ const useStyles = makeStyles((theme) => ({
...
+      .split("")
+      .join("+.*");
+    const regex = new RegExp(filterString);
+
+    tsParticipants.createOrUpdateFilterFn("filterIdentities", (participant) =>
+      regex.test(participant.identity.name.toLowerCase())
     );
+  };
+
+  const handleCheckboxFilter = (event) => {
```
blueridger, line 221: These should be wrapped in a `useCallback` I believe

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -225,50 +325,108 @@ export const ExplorerHome = ({initialView}: ExplorerHomeProps): ReactNode => {
...
-                      <CredTimeline data={row.chart} />
+                {tsParticipants.currentPage.length > 0 ? (
+                  tsParticipants.currentPage.map((row) => (
+                    <TableRow key={row.identity.name}>
+                      <TableCell component="th" scope="row">
+                        {row.identity.name}
+                      </TableCell>
+                      <TableCell>{Math.round(row.cred)}</TableCell>
+                      <TableCell>
+                        {Number(row.grainEarned).toFixed(2) + currencySuffix}
```
blueridger, line 371: You should use the `format` function provided in the `core/ledger/grain.js` module instead. (this is because grain is a BigInt, so I believe casting to number can fail for large values that may be stored)

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -88,77 +98,164 @@ const useStyles = makeStyles((theme) => ({
...
+    );
 
-  const createData = (username, cred, grain, chart) => ({
-    username,
-    cred,
-    grain,
-    chart,
-  });
+    credAndGrainSummary.totalCred += participant.cred;
+    credAndGrainSummary.totalGrain += Number(participant.grainEarned);
```
blueridger, line 175: Grain should not be cast as a number. It needs to be handled by the BigInt library used in `core/ledger/grain.js`

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -88,77 +98,164 @@ const useStyles = makeStyles((theme) => ({
...
+  }
 
-  // sort by cred amount, highest to lowest
-  const nodes = sortBy(initialView.participants(), (n) => -n.cred);
-  // create an array of 0s for the cred summary graph at the top of the page
-  let credTimelineSummary = initialView.intervals().map(() => 0);
+  if (tsParticipants.currentPage.length > 0) {
+    credAndGrainSummary.avgCred =
+      credAndGrainSummary.totalCred / tsParticipants.currentPage.length;
```
blueridger, line 180: This should be `tsParticipants.length` instead of `tsParticipants.currentPage.length`

blueridger (commented):
This PR is way too big. I could see it broken up into 2-5 PRs. I won't ask that you break this one up this time, but in the future each PR should correspond to an atomic increment in functionality. This makes it easier to review and easier to revert. A break down for this could have been:
1. Add TableState
2. Add Pagination
3. Add filtering
4. Add sorting
5. update CSS/Theme/circles

https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#semantically-atomic-commits

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: Ok, removed.

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: I ended up using `useMemo` because these calculations need to be included in the render, and `useMemo` runs before render, whereas `useEffect` runs after render.

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: Done, thank you. I completely forgot about the special handling of grain calculations.

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: Done.

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: In this case, I don't think `useCallback` would buy us anything. It's more beneficial for eliminating needless rerenders of child components. Re-defining these functions every cycle is very low cost, and wrapping them in `useCallback` mostly, if not entirely, offsets the cost benefit.

If you're interested, here are some decent write-ups:
https://kentcdodds.com/blog/usememo-and-usecallback
https://blog.hackages.io/react-hooks-usecallback-and-usememo-8d5bb2b67231

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: The totals and averages are calculated for the visible rows of the page. So if you filter down to, say, three names, the average and totals will be for just those three participants. I implemented it this way because that's what makes the most sense to me, but it's fully open for discussion.

BrianBelhumeur:
> This PR is way too big. I could see it broken up into 2-5 PRs. I won't ask that you break this one up this time, but in the future each PR should correspond to an atomic increment in functionality. This makes it easier to review and easier to revert. A break down for this could have been:
> 
> 1. Add TableState
> 2. Add Pagination
> 3. Add filtering
> 4. Add sorting
> 5. update CSS/Theme/circles
> 
> https://github.com/sourcecred/sourcecred/blob/master/CONTRIBUTING.md#semantically-atomic-commits

Ah yes, I see, and apologize for the extra overhead. I'll make sure to do smaller, more atomic commits in the future.

src/ui/components/ExplorerHome/ExplorerHome.js
blueridger: Thanks!

src/ui/components/ExplorerHome/ExplorerHome.js
blueridger: The filtered length is represented by `tsParticipants.length` as seen [here](https://github.com/sourcecred/sourcecred/blob/master/src/webutil/tableState.js#L259).
`tsParticipants.currentPage.length` will be whatever the pagination is limiting to (50, 100, 150).
So the question is should the aggregates include only what is on the currently paginated page, or an aggregate of what is on all pages? The latter seems like the clear choice to me.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -88,77 +99,167 @@ const useStyles = makeStyles((theme) => ({
...
+      totalCred: 0,
+      totalGrain: fromInteger(0),
+      avgCred: 0,
+      avgGrain: fromInteger(0),
+    };
 
-  const rows = nodes.map((node) => {
-    const {credPerInterval} = node;
+    if (tsParticipants.currentPage.length > 0) {
+      for (const participant of tsParticipants.currentPage) {
```
blueridger, line 170: (we will have to add a new `allPages` output to TableState to implement my aggregation across all pages, it appears)

blueridger:
From @KuraFire 
> From all pages, yes. They update to reflect the time control (All time, Last Week, Last Month etc) but not to changes to the activity table list below

BUT you know what I just remembered? That's all out of scope for your task! I reopened https://github.com/sourcecred/sourcecred/issues/2237 and the needed changes can be done as a part of that.

blueridger (approved):

Issue #2231, opened 2020-09-10
New SC Explorer Dashboard: Timeframe Control


KuraFire:
Atomic-style ticket for #2227 — Timeframe Control component:

![SourceCred Explorer Dashboard - 2  Timeframe Control@2x](https://user-images.githubusercontent.com/33087/92668745-88272500-f2c4-11ea-9568-07277f36bad9.png)

### Specifications:

This control adjusts the following components’ data fields:
- Top Chart (can be punted, see below)
- Stats Circles (all four, see ticket # )
- Activity Table, _first three columns only_:
   - Participant
   - Cred
   - Grain
- Plugin Bar Chart

The four options correspond to data ranges as follows:
- **This Week:** from the most recent Sunday, midnight UTC time to `current time`
- **Last Week:** the full week before the most recent Sunday, midnight UTC time **(Default view upon page load)**
- **This Month:** the four full weeks before the most recent Sunday, midnight UTC time
- **All Time:** entire available timeframe for the project, from when the SourceCred instance first started tracking to `current time`

#### Top Chart interaction
In the future, we plan on having a time chart segment selection feature, where you can click and drag across the Top Chart to select an arbitrary time period to show data for. That is dependent on a future change to Cred computation and tracking (to make it daily). That said, it would be a nice **OPTIONAL BONUS** if an overlay of some kind could be rendered on the Top Chart as you change the control.

This concept's implementation is blocked by: #2232 Visual UI and functional prototype demonstration of Top Chart timeframe representation

PR #2737, opened 2021-02-09
ExplorerHome UI: timeframe control

blueridger:
# Description
This adds basic Timeframe Control (https://github.com/sourcecred/sourcecred/issues/2231) to the new ExplorerHome using the 4 preset time windows.

Some things that will need to be worked out after this PR:
- This week an last week show no line graphs because there is only 1 data point. Kinda awk.
- Activity Table header needs styled and timestamps need improved according to spec

# Test Plan
Screen Shot 2021-02-08 at 3 51 43 PM
Screen Shot 2021-02-08 at 3 51 54 PM
Screen Shot 2021-02-08 at 3 52 03 PM
Screen Shot 2021-02-08 at 3 52 11 PM

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -1,5 +1,10 @@
 // @flow
-import React, {useState, useMemo, type Node as ReactNode} from "react";
+import React, {
+  useState,
+  useEffect,
+  useMemo,
+  type Node as ReactNode,
+} from "react";
 import {
   Checkbox,
   Container,
@@ -33,6 +38,8 @@ import type {CurrencyDetails} from "../../../api/currencyConfig";
 import {format, add, div, fromInteger} from "../../../core/ledger/grain";
 import CredTimeline from "./CredTimeline";
 import {IdentityTypes} from "../../../core/identity/identityType";
+import {type Interval, type IntervalSequence} from "../../../core/interval";
+import {formatTimestamp} from "../../utils/dateHelpers";
 
 const useStyles = makeStyles((theme) => ({
   root: {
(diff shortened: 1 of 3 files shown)


topocount (approved):
The code looks good, aside from the small TableState refactor that got snuck in there :wink: 

tests work as expected. The only thing that's a bit confusing is the KPI circles don't update at all, and I would intuitively expect one or two of them to do so. Would be curious to get @KuraFire 's opinion.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -314,7 +393,10 @@ export const ExplorerHome = ({
               marginBottom: "20px",
             }}
           >
-            <span style={{fontSize: "24px"}}>Last Week&apos;s Activity</span>
+            <span style={{fontSize: "24px"}}>
+              {TIMEFRAME_OPTIONS[tab].tableLabel}
+            </span>
+            <span style={{fontSize: "16px"}}>{formatInterval(interval)}</span>
```
BrianBelhumeur, line 399: Nit: whenever we touch font sizing styles, they should be converted to rem (or em) vs px in order to support accessibility: https://www.24a11y.com/2019/pixels-vs-relative-units-in-css-why-its-still-a-big-deal/

src/webutil/tableState.js
```diff
@@ -159,6 +159,7 @@ export function useTableState<T>(
     },
   }
 ): TableState<T> {
+  const data = useMemo(() => dataWrapper.data, [dataWrapper.data]);
```
BrianBelhumeur, line 162: I like that we're memoizing inside TableState... does this obviate the need to do so outside before passing data in?

BrianBelhumeur (approved):
Approved with comments

src/webutil/tableState.js
blueridger: Maybe but I think not. If I understand correctly, if dataWrapper.data is not memoized and is rebuilt every render, then everything downstream of it will also be rebuilt every render.

src/ui/components/ExplorerHome/ExplorerHome.js
blueridger: accepted

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -135,7 +184,19 @@ export const ExplorerHome = ({
   if (!initialView) return null;
 
   const classes = useStyles();
-  const [tab, setTab] = useState<number>(1);
+  const [tab, setTab] = useState<number>(TIMEFRAME_OPTIONS.length - 1);
+  const [interval, setInterval] = useState<Interval>(
```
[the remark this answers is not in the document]
blueridger: accepted

blueridger:
> The only thing that's a bit confusing is the KPI circles don't update at all, and I would intuitively expect one or two of them to do so.

That will be handled by a different issue.

Issue #2260, opened 2020-09-14
New SC Explorer Dashboard: Activity Table


KuraFire:
Atomic-style ticket for #2227 — Activity Table component. See UX wireframe image below.

Note ahead: since this is a big ticket just for the Activity Table component, it may make sense to further break down (atomize?) this into separate tickets for each part.

#### Important changes from current table
- The first 3 columns in the table will show content based on the **Timeframe Control** (see: #2231)
- The last column shows Contribution Charts for Participants across **All Time**
- This table will initially **NOT** support exploration of a Participant's contributions (no + button to explore detailed breakdowns).
   - This feature will return with the upcoming Participant View, which may include key details we want shown on the Activity Table (via a + button or similar)

#### Colors of texts in Activity Table in `DARK MODE`
- `#FAFBFD` for Participants and Cred scores, and all table header + footer rows’ texts
- `#FFDDC6` for Projects and Organizations
- `#C5A2C5` for Bots
- `#6174CC` for Cred values
- `#FFAA3D` for Grain values

**Note:** See second image below for a visual example of the table in `DARK MODE` with the above colors.

#### Variable Text Labels
- The "Last Week's Activity" header should read as follows, based on the visitor's selected Timeframe Control (see: #2231 ). Note capitalization and proper apostrophes (`’`, not `'` marks).
   - This Week’s Activity (so far)
   - **Last Week’s Activity** (default view)
   - Past Month’s Activity
   - All Time Activity
- Date label should be formatted as follows under the varying conditions, as set by the Timeframe Control:
   - This Week: `Sep. 6 to Today`  
     (past however many days since the last Harvest)
   - **Last Week:** `Aug. 30 to Sep. 5, 2020`  
     (past 1 week that ended with the last Harvest)
   - This Month: `Aug. 16 to Sep. 5, 2020`  
     (past 4 weeks that ended with the last Harvest)
   - All Time: `Aug. 13, 2018 to Today`  
     (project tracking start date to Today) 

If the start date in this display is of a different year than the end date, add it as specified for All Time (`, YYYY`). 

#### Filter input field
This field filters down the list of Participants (incl. bots, orgs, projects) to items that match whatever is typed. E.g. typing `kura` would filter it down to show just the `kurafire` entry and any others that happen to have `kura` in their name. 

**Filtering behavior:** 
[...]

PR #2741, opened 2021-02-10
Activity Table date label formatting

BrianBelhumeur:
# Description
This PR builds upon @blueridger's work implementing the date interval selection. These changes include:
- Date string formatting as per issue #2260
-- Show start year only when different from end year
-- Show "Today" instead of current or future date
- Default to "Last Week's Activity" time interval on page load
- Eliminated a visible repaint by refactoring interval selection out of `useEffect` and into a small helper function

# Test Plan
Tested manually, viewing date formats when switching between each date interval

## Before
https://www.loom.com/share/4d5b8af1b2c34bb7824e4f3ef8841345

## After
https://www.loom.com/share/e5e20f03e95d400ca846d08dde893ced

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -197,12 +197,10 @@ export const ExplorerHome = ({
   if (!initialView) return null;
 
   const classes = useStyles();
-  // default view is Last Week's Activity
+  // default view is Last Week's Activity (array index 1)
   const [tab, setTab] = useState<number>(1);
   const [selectedInterval, setSelectedInterval] = useState<Interval>(
-    TIMEFRAME_OPTIONS[TIMEFRAME_OPTIONS.length - 1].selector(
-      initialView.intervals()
-    )
+    TIMEFRAME_OPTIONS[1].selector(initialView.intervals())
   );
   const updateTimeframe = (index) => {
     setTab(index);
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -1,10 +1,5 @@
 // @flow
-import React, {
-  useState,
-  useEffect,
-  useMemo,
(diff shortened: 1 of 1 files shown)


src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -354,22 +349,51 @@ export const ExplorerHome = ({
...
+
+      return someDate >= today;
+    };
+
+    const currentYear = new Date().getFullYear();
+    const startDate = new Date(interval.startTimeMs);
+    const endDate = new Date(interval.endTimeMs);
+
+    const startTime =
+      startDate.getFullYear() === currentYear
```
blueridger, line 386: Isnt this supposed to compare the startDate year to the _endDate year_, not the current year.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -354,22 +349,51 @@ export const ExplorerHome = ({
...
+        minute: undefined,
+        year: undefined,
+      });
+
+    const isTodayOrLater = (someDate) => {
+      const today = new Date();
+      today.setHours(0);
+      today.setMinutes(0);
+      today.setSeconds(0);
+      today.setMilliseconds(0);
```
blueridger, line 376: Fun fact, these four lines can be replaced with `today.setHours(0, 0, 0, 0)`

BrianBelhumeur: Oooh that is fun, thank you!

src/ui/components/ExplorerHome/ExplorerHome.js
BrianBelhumeur: Ah good point. In my head I was assuming end date year would always be the current year, but I can see how that would not necessarily be the case. Good catch, and thank you.

KuraFire (approved):
Ya lgtm! 👍🏼

blueridger (approved):

PR #2742, opened 2021-02-10
Sort grain numerically, not alphabetically in Activity Table

BrianBelhumeur:
# Description
I got the cred repo going as my test instance and realized that the grain sort on the activity table is sorting string-wise rather than numerically, because grain is represented as a string. I updated the sorting function to wrap the grain string in a bigInt() so it now sorts correctly.

# Test Plan
Tested manually with `cred` repo data (which has real grain values) with the following steps:

- Load page
- Click "Grain" header for ascending sort. Confirm sort is ascending
- Click "Grain" header again for descending sort. Confirm sort is descending

## Before
Screen Shot 2021-02-10 at 22 46 43

## After
Screen Shot 2021-02-10 at 22 46 00

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -23,6 +23,7 @@ import {
 import {makeStyles} from "@material-ui/core/styles";
 import TableSortLabel from "@material-ui/core/TableSortLabel";
 import deepFreeze from "deep-freeze";
+import bigInt from "big-integer";
 import {CredGrainView} from "../../../core/credGrainView";
 import {
   useTableState,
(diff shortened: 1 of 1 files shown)


topocount (approved):
good catch @BrianBelhumeur !

topocount:
@BrianBelhumeur one bit of feedback: Test plans are most helpful when they're instructions for verifying the correctness of your work, so a reviewer (who might not be familiar with sourcecred CLI workflows) can easily the instructions and see for themselves how they should inspect the result of the change.

KuraFire (approved):
Wonderful!

blueridger (approved):

BrianBelhumeur:
> @BrianBelhumeur one bit of feedback: Test plans are most helpful when they're instructions for verifying the correctness of your work, so a reviewer (who might not be familiar with sourcecred CLI workflows) can easily the instructions and see for themselves how they should inspect the result of the change.

Notre! Updated the comment with the test plan and will do so in the future. Thank you!

Issue #2744, opened 2021-02-11
Ledger.lastDistributionTimestamp edge case refactor


decentralion:
In ledger, the lastDistributionTimestamp method returns -Infinity if there has never been any distribution.
I think it would be cleaner if it returns `null` in that case, as it would be clearer in the type signature that callers need to consider the edge case where there haven't been any distributions. A `-Infinity` is more likely to sneak through unexpectedly and cause a user-facing bug.

https://github.com/sourcecred/sourcecred/blob/703e8b529883c8b923fba2a178d561f23da05f9b/src/core/ledger/ledger.js#L702-L710

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038215716709) by [Unito](https://www.unito.io/learn-more)

PR #2745, opened 2021-02-11
update docs for ledger.lastDistributionTimestamp

decentralion:
The API for ledger.lastDistributionTimestamp doesn't specify what
happens when there haven't been any distributions. Counter-intuitively,
we return -Infinity, and we actually depend on that behavior elsewhere,
so it can't easily be changed. This commit just updates the
documentation so that we track this accordingly, and also I filed a new
issue to consider changing this.

Test plan: `yarn test` passes, since it's only an update to docs/test
code, this definitely suffices.

There's a potential improvement refactor here, see https://github.com/sourcecred/sourcecred/issues/2744

### DIFF
--- a/src/core/ledger/ledger.js
+++ b/src/core/ledger/ledger.js
@@ -704,8 +704,15 @@ export class Ledger {
    *
    * We provide this because we may want to have a policy that issues one
    * distribution for each interval in the history of the project.
+   *
+   * If there were never any distributions, then -Infinity will be returned.
    */
   lastDistributionTimestamp(): TimestampMs {
+    // TODO(#2744): Amend interface to return null instead of -Infinity
+    // This will be a better interface primarily because in flow checking, it's
+    // easy for the client to not realize that -Infinity could get returned, and
+    // making an explicit return `TimestampMs | null` type will encourage callers
+    // to handle the edge case explicitly.
     return this._lastDistributionTimestamp;
   }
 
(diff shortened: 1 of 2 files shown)


src/core/ledger/ledger.test.js
```diff
@@ -1416,9 +1416,12 @@ describe("core/ledger/ledger", () => {
         failsWithoutMutation(ledger, thunk, "invalid timestamp");
       }
     });
-  });
-
-  describe("timestamps", () => {
```
blueridger, line 1421: nit: Looks like this was misnamed, not duplicate. Instead of bringing the non-timestamp-related `it` block under the `timestamps` describe block, maybe this describe block should be renamed `uuids` or something.

src/core/ledger/ledger.js
```diff
@@ -778,6 +786,10 @@ export class Ledger {
     account.paid = G.add(params.grainReceipt.amount, account.paid);
     account.allocationHistory.push(params);
     account.balance = G.add(params.grainReceipt.amount, account.balance);
+    this._totalCirculatingGrain = G.add(
+      this._totalCirculatingGrain,
+      params.grainReceipt.amount
+    );
```
blueridger, line 792: Did you mean to bake this into this PR? If so, please update the description to acknowledge it.

decentralion: Nope, thanks!

src/core/ledger/ledger.test.js
decentralion: Good catch!

decentralion (commented):
Updated, thanks!

blueridger (approved):

PR #2746, opened 2021-02-11
Refactor AccountOverview to use Ledger APIs

decentralion:
This commit does a slight refactor of the AccountOverview so that rather
than manually scanning for the last distribution timestamp by iterating
over every ledger event, it uses the existing
`Ledger.lastDistributionTimestamp` API. This has two benefits:

1. It's much more performant, as now a frontend UI load doesn't need to
do another scan over every event in the ledger history, instead using a
cached property that's present on the ledger
1. It avoids re-implementing ledger specific scanning logic, even though
it does use the public API it still feels like a leakage of ledger
implementation details

Test plan: Load the UI and verify that the display is correct, both in
the case with/without any distributions

Note: This makes a slight semantic change to the last distribution timestamp indicator. Previously, we displayed the ledger timestamp (i.e. when was the distribution event created). Now we show the cred timestamp (i.e. when did the Cred period associated with the last distribution end). I think this is fine.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038305658509) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/ui/components/AccountOverview.js
+++ b/src/ui/components/AccountOverview.js
@@ -12,7 +12,6 @@ import {type Account} from "../../core/ledger/ledger";
 import {type CurrencyDetails} from "../../api/currencyConfig";
 import * as G from "../../core/ledger/grain";
 import {useLedger} from "../utils/LedgerContext";
-import findLast from "lodash.findlast";
 import {formatTimestamp} from "../utils/dateHelpers";
 import {makeStyles} from "@material-ui/core/styles";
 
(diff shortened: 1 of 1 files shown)


KuraFire (approved):
I am all in favor of performance improvements to the UI 👍🏼 👍🏼

blueridger (approved):
needs rebased and lint needs fixed, but can merge on this approval after those two things are fixed.

decentralion:
Rebased and should be technically correct. However, I want to note that this does make a slight semantic change. Previously, we displayed the "ledger timestamp" of the distribution, now we are showing the "cred timestamp". For example, suppose that the Cred interval ends at midnight on Sunday, but it takes 3 hours before we run the job to create a Grain distribution. Previously, we would see the Distribution timestamp as 3am Monday, but now it will show midnight Sunday. I think this is fine; so I'm going to merge it, but I'll update the PR description to record the change.

PR #2749, opened 2021-02-11
CLI: Run 'credrank' quietly during 'go'

blueridger:
# Description
I am making this change in time for the upcoming version release, which will not include the credrank migration.
This adds `credrank` to the `go` command, which will allow us to forcefully browse to the unfinished `explorer-home` page and compare the results on live instances. This will be nice to preview what changes to cred are coming in the following version release.

I also added a `--stealth` flag so that the `go` command doesn't print credrank scores that are not actually being used, so as not to confuse our instance maintainers. We can remove this flag after the migration.

# Test Plan
```
scdev go && scdev serve
```
Verified that both the `explorer` (from the menu) and `explorer-home` (from forceful browsing) pages work.

### DIFF
--- a/src/cli/credrank.js
+++ b/src/cli/credrank.js
@@ -24,10 +24,14 @@ function die(std, message) {
 const credrankCommand: Command = async (args, std) => {
   let shouldIncludeDiff = false;
   let isSimulation = false;
+  let shouldRunStealth = false;
   const processedArgs = args.filter((arg) => {
     switch (arg) {
       case "-d":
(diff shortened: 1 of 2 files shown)


topocount (approved):
nit: While we're here, do you mind adding the `--stealth` and `--simulate` flags to the credrank command usage docstring?

I'm comfortable with not adding them, but I favor documenting them at this point.

topocount:
I've also noticed that no `credrankHelp` command exists like it does for other functions, but that's probably for another pull request.

PR #2764, opened 2021-02-17
Create design pattern for Instance and Api and use for credrank

blueridger:
# Description
WHEW. So I sat down to implement a new feature, but I quickly realized I could not, in good conscience, do so without first creating a refactor, else my feature would only be compounding some ugly tech debt. I hope this doesn't step on refactor team toes, but I felt pretty confident in my ability to set these patterns using my relevant industry experience, and I wanted it done.

This PR does not add any new functionality. It just refactors existing functionality into new design patterns.

# New Design Patterns
This PR sets 2 new design patterns and implements them the minimum amount needed to work with the Credrank CLI.
## 1. Instances
Instances really did not have any cohesive representation in code. So I've created a minimal Instance interface as well as a local disk implementation. I just copied the existing I/O code from `cli/common.js` and `cli/credrank.js` and organized/consolidated it into the instance implementation.
## 2. Primary API's
The CLI commands were acting as the primary API's for SourceCred, which meant that we weren't really supporting programatic use of our library's main functions. So I separated the core API logic out of `cli/common.js` and `cli/credrank.js` and added it to a new `api/credrank.js` module, and made the cli depend on that module. This allows us to define our highest-level entry points with clear inputs and output, much like a service endpoint would have.
## Implications for CLI
The above 2 patterns mean that the CLI has a new pattern: It makes use of the Instance and Api features described above, but otherwise it is only responsible for command line I/O -- AS IT SHOULD BE.
See https://github.com/sourcecred/sourcecred/pull/2764

# Benefits
- More maintainable! For example, this will let me add a feature that requires a new config like cred tithing easily and cleanly.
- Aligns our stack with industry patterns and best-practices, making it easier to onboard.
- We could add additional instance implementations like a github-based one (the LedgerStorage work could play very nicely here). This could make programatic use of our library in environments like Observable notebooks SUPER EASY.
- Once we migrate the `graph` and `load` commands to use the new design patterns, we'll be able to delete most if not all of `cli/common.js` (YAY).

# Test Plan
This is NOT covered by automated tests. I find this ok for now, because none of this code was covered prior to this refactor, and it's close enough to the CLI that manual testing should work fine.

Manually tested `scdev credrank` with and without `-d -s` and `--stealth` to make sure no behavior has changed. Tested on top of https://github.com/sourcecred/sourcecred/pull/2764
## Before
```
scdev credrank 
  GO   credrank
  GO   load data
 DONE  load data: 2774ms
  GO   run CredRank
 DONE  run CredRank: 13s
# Top Participants By Cred

| Description | Cred | % |
| --- | --- | --- |
| decentralion | 36725.8 | 19.3% |
| lbstrobbe | 14312.3 | 7.5% |
| wchargin | 12604.8 | 6.6% |
| s-ben | 11458.6 | 6.0% |
| KuraFire | 10812.0 | 5.7% |
| hammad | 10800.1 | 5.7% |
| Thena | 8551.9 | 4.5% |
| topocount | 7874.8 | 4.1% |
| panchomiguel | 7253.0 | 3.8% |
| joiecousins | 7126.9 | 3.7% |
| beanow | 6595.3 | 3.5% |
| sandpiper | 6491.0 | 3.4% |
| bex | 6367.6 | 3.3% |
| Jojo | 2849.1 | 1.5% |
| benoxmo | 2376.4 | 1.2% |
| dependabot | 2142.2 | 1.1% |
| burrrata | 2016.7 | 1.1% |
| eeli | 1883.9 | 1.0% |
| mzargham | 1718.6 | 0.9% |
| youngkidwarrior | 1498.3 | 0.8% |
  GO   write cred graph
 DONE  write cred graph: 1482ms
 DONE  credrank: 18s
```
## After
```
scdev credrank
  GO   credrank
  GO   load data
 DONE  load data: 757ms
  GO   run CredRank
 DONE  run CredRank: 15s
# Top Participants By Cred

| Description | Cred | % |
| --- | --- | --- |
| decentralion | 36725.8 | 19.3% |
| lbstrobbe | 14312.3 | 7.5% |
| wchargin | 12604.8 | 6.6% |
| s-ben | 11458.6 | 6.0% |
| KuraFire | 10812.0 | 5.7% |
| hammad | 10800.1 | 5.7% |
| Thena | 8551.9 | 4.5% |
| topocount | 7874.8 | 4.1% |
| panchomiguel | 7253.0 | 3.8% |
| joiecousins | 7126.9 | 3.7% |
| beanow | 6595.3 | 3.5% |
| sandpiper | 6491.0 | 3.4% |
| bex | 6367.6 | 3.3% |
| Jojo | 2849.1 | 1.5% |
| benoxmo | 2376.4 | 1.2% |
| dependabot | 2142.2 | 1.1% |
| burrrata | 2016.7 | 1.1% |
| eeli | 1883.9 | 1.0% |
| mzargham | 1718.6 | 0.9% |
| youngkidwarrior | 1498.3 | 0.8% |
  GO   writing changes
 DONE  writing changes: 1439ms
 DONE  credrank: 18s
```


topocount (commented):
I think this looks amazing. My only initial feedback is that this probably makes more sense as two PRs: 
- One with the new design pattern
- One where you actually utilize it in the credrank CLI command.

src/api/instance/localInstanceImpl.js
```diff
@@ -0,0 +1,184 @@
...
+const CREDGRAPH_PATH = ["output", "credGraph.json"];
+const GRAPHS_DIRECTORY = ["output", "graphs"];
+const GRAPHS_PATH = ["graph.json"];
+const LEDGER_PATH = ["data", "ledger.json"];
+
+/**
+This is an Instance implementation that reads and writes using relative paths
+on the local disk.
+ */
+export class LocalInstance implements Instance {
```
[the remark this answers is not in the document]
blueridger: Not attached. I actually generally see the class name include the suffix too. Or sometimes instead of the "Impl" suffix, the interface gets an "I" suffix. Makes it a little easier to delineate, but I don't really care.

src/api/instance/instance.js
```diff
@@ -0,0 +1,23 @@
...
+import {CredGraph} from "../../core/credrank/credGraph";
+import {Ledger} from "../../core/ledger/ledger";
+import {type DependenciesConfig} from "../dependenciesConfig";
+
+/**
+This interface represents a SourceCred Instance, which is simply a read/write
+interface for configurations and outputs of the main SourceCred API.
+ */
+export interface Instance {
+  loadCredrankInput(): Promise<CredrankInput>;
```
wchargin, line 12: Naming: might I suggest that the dual to `write` be called `read` rather
than `load`, just for general consistency? (e.g., with `writeFile` and
`readFile` from Node’s `fs` API)

src/api/instance/instance.js
```diff
@@ -0,0 +1,23 @@
...
+import {type CredrankInput} from "../credrank";
+import {CredGraph} from "../../core/credrank/credGraph";
+import {Ledger} from "../../core/ledger/ledger";
+import {type DependenciesConfig} from "../dependenciesConfig";
+
+/**
+This interface represents a SourceCred Instance, which is simply a read/write
+interface for configurations and outputs of the main SourceCred API.
+ */
+export interface Instance {
```
wchargin, line 11: At a glance, it’s not clear to me why this API has the shape that it
does. It’s not very orthogonal: the CredRank input is read only, the
ledger and dependencies config are write-only, and the cred graph is
read-write. As a reader, I’m left wondering why.

Diving in a bit deeper, we can see that a ledger and a dependencies
config are part of the CredRank input, so that explains why they don’t
need dedicated read methods, but the factoring still feels lopsided. And
the cred graph I/O methods are actually a bit different than the rest,
since CredRank never needs to read the cred graph; presumably this is
intended for downstream consumers, but this isn’t clear.

What do you think about an interface that more closely mirrors the
structure of CredRank, and clarifies how `readCredGraph` is different
from the rest? Something like:

```javascript
interface Instance {
  /** Reads inputs required to run CredRank. */
  readCredrankInput(): Promise<CredrankInput>;
  /** Writes output after running CredRank. */
  writeCredrankOutput(o: CredrankOutput): Promise<void>;

  /** Reads a cred graph generated by a previous run of CredRank. */
  readCredGraph(): Promise<CredGraph>;
}
```

src/api/instance/localInstanceImpl.js
```diff
@@ -0,0 +1,184 @@
...
+} from "../../core/credrank/credGraph";
+
+const DEPENDENCIES_PATH = ["config", "dependencies.json"];
+const WEIGHT_OVERRIDES_PATH = ["config", "weights.json"];
+const BUDGET_PATH = ["config", "pluginBudgets.json"];
+const INSTANCE_CONFIG_PATH = ["sourcecred.json"];
+const CREDGRAPH_PATH = ["output", "credGraph.json"];
+const GRAPHS_DIRECTORY = ["output", "graphs"];
+const GRAPHS_PATH = ["graph.json"];
+const LEDGER_PATH = ["data", "ledger.json"];
```
wchargin, line 43: These are global mutable state; since they’re not exported, can we at
least type them as `$ReadOnlyArray<string>` to prevent accidental
mutation?

src/api/instance/localInstanceImpl.js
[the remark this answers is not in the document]
wchargin: My thoughts here:

  - Interfaces are the most important, so they should always get the
    good names. Thus, `List`, not `IList`. So, I approve your choice of
    interface name `Instance`.

  - For concrete implementations, the “distinguishing characteristic
    followed by interface name” pattern works well: `ArrayList`,
    `BrowserLocalStore`, etc. So, I approve your choice of class name
    `LocalInstance`.

  - If the only thing you can think of to name your concrete
    implementation is `FooImpl`, then you *probably* don’t need an
    interface at all, and can just replace the interface with the
    concrete class and give it a good name. (I won’t present this as an
    absolute, but I’ve never yet seen `FooImpl` actually be the right
    choice.)

  - In general, the file name should follow its primary export, so I
    agree that `localInstance.js` is better. The `Impl.js` is just as
    superfluous in the filename; eschew noise.

src/api/instance/localInstanceImpl.js
```diff
@@ -0,0 +1,184 @@
...
+      () => []
+    );
+  }
+
+  async loadPluginsBudget(): Promise<Budget | null> {
+    const budgetPath = pathJoin(this._baseDirectory, ...BUDGET_PATH);
+    return loadJsonWithDefault(budgetPath, pluginBudgetParser, () => null);
+  }
+
+  writePluginDirectory(pluginId: string): string {
```
wchargin, line 170: This reads a bit confusingly to me. From the name, I’d expect it to
write a bunch of files under a plugin directory, but instead it just
creates the directory itself. Can we call it `createPluginDir`?

src/cli/credrank.js
```diff
@@ -1,20 +1,16 @@
...
 
-import fs from "fs-extra";
-import stringify from "json-stable-stringify";
-import {join as pathJoin} from "path";
 import {sum} from "d3-array";
 import {format} from "d3-format";
 
 import sortBy from "../util/sortBy";
-import {credrank} from "../core/credrank/compute";
+import {credrank as callCredrank} from "../api/credrank";
```
wchargin, line 7: nit: Why `callCredrank` rather than just `credrank`? All function calls
are calls; this just confuses me when I read it because I have to figure
out what it does that’s special (answer: nothing).

src/api/instance/localInstanceImpl.js
```diff
@@ -0,0 +1,184 @@
...
+  async loadCredrankInput(): Promise<CredrankInput> {
+    const [
+      pluginGraphs,
+      ledger,
+      weightOverrides,
+      dependencies,
+      pluginsBudget,
+    ] = await Promise.all([
+      await this.loadPluginGraphs(),
+      await this.loadLedger(),
```
wchargin, line 71: This does not do what you want it to: you are actually awaiting all the
results in sequence to construct the array, then passing an array of
already resolved values to `Promise.all`, which returns immediately.
Drop inner `await`s to recover concurrency.

src/api/instance/localInstanceImpl.js
```diff
@@ -0,0 +1,184 @@
...
+
+  async writeLedger(ledger: Ledger): Promise<void> {
+    const ledgerPath = pathJoin(this._baseDirectory, ...LEDGER_PATH);
+    return fs.writeFile(ledgerPath, ledger.serialize());
+  }
+
+  async writeCredGraph(credGraph: CredGraph): Promise<void> {
+    const cgJson = stringify(credGraph.toJSON());
+    const outputPath = pathJoin(this._baseDirectory, ...CREDGRAPH_PATH);
+    await fs.writeFile(outputPath, cgJson);
```
wchargin, line 98: nit: Might as well be consistent about whether we `return` or `await` a
promise in tail position? (I generally like returning them, but at the
behest of others I’ve been convinced to await them instead.)

src/api/credrank.js
```diff
@@ -0,0 +1,76 @@
...
+import {contractions as identityContractions} from "../core/identity";
+import {credrank as computeCredrank} from "../core/credrank/compute";
+
+export type CredrankInput = {|
+  pluginGraphs: $ReadOnlyArray<WeightedGraph>,
+  ledger: Ledger,
+  dependencies: DependenciesConfig,
+  weightOverrides: WeightsT,
+  pluginsBudget: Budget | null,
+|};
```
wchargin, line 27: Is there a reason that these fields are mutable rather than read-only
(e.g., `+pluginGraphs: ...`)? We use read-only struct fields by default,
and I don’t see a need for these to be mutable.

src/api/credrank.js
```diff
@@ -0,0 +1,76 @@
...
+
+export type CredrankOutput = {|
+  credGraph: CredGraph,
+  updatedLedger: Ledger,
+  updatedDependencies: DependenciesConfig,
```
wchargin, line 32: Maybe just `ledger` and `dependencies`? Since they’re fields on a type
with `Output` in its name, it stands to reason that they’ve been
updated.

src/api/instance/localInstanceImpl.js
```diff
@@ -0,0 +1,184 @@
...
+  constructor(baseDirectory: string) {
+    this._baseDirectory = baseDirectory;
+  }
+
+  //////////////////////////////
+  //
+  //  Interface Functions
+  //
+  //////////////////////////////
+
```
wchargin, line 61: nit: We don’t typically use banner comments; this block takes up 6 lines
that, imho, don’t convey much useful information. Suggest removing, but
if you want to keep them, maybe tone down to just `// Line comments`?

src/api/instance/instance.js
wchargin, line 11: Optional, and can do later: any thought for separating out the writable
part of the interface? Following up on my suggestion above:

```javascript
interface Instance {
  /** Reads inputs required to run CredRank. */
  readCredrankInput(): Promise<CredrankInput>;
  /** Writes output after running CredRank. */
  writeCredrankOutput(o: CredrankOutput): Promise<void>;

  /** Reads a cred graph generated by a previous run of CredRank. */
  readCredGraph(): Promise<CredGraph>;
}
interface WritableInstance extends Instance {
  /** Writes output after running CredRank. */
  writeCredrankOutput(o: CredrankOutput): Promise<void>;
}
```

I suspect that this might be useful when you want to write an `Instance`
implementation that reads from GitHub without authentication, at which
point you won’t be able to implement the `write*` methods.

src/api/instance/instance.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import {type CredrankInput} from "../credrank";
+import {CredGraph} from "../../core/credrank/credGraph";
+import {Ledger} from "../../core/ledger/ledger";
+import {type DependenciesConfig} from "../dependenciesConfig";
+
+/**
+This interface represents a SourceCred Instance, which is simply a read/write
+interface for configurations and outputs of the main SourceCred API.
```
wchargin, line 9: nits on effective documentation:

> This interface represents a SourceCred Instance, which is simply a read/write
> interface for configurations and outputs of the main SourceCred API.

“This interface” is superfluous; the doc comment for this interface
clearly describes this interface.

“represents a SourceCred Instance, which is” is superfluous; it’s in the
SourceCred repository and is called `Instance`.

“simply a read/write interface for […]” is meaningful!

“configurations and outputs” makes me ask, “the natural dual of
*outputs* is *inputs*; in what sense are *configurations* different?”.
Since the method does in fact return just a `CredrankInput`, perhaps
“inputs” is simpler.

“of the main SourceCred API” feels helpful to me, because it specifies
the level of abstraction that I should expect. But it looks like this
really is all about CredRank. Maybe we could say that more precisely.

So, how about:

```javascript
/**
 * Simple read-write interface for CredRank inputs and outputs.
 */
```

?

Exercise for the reader: apply this same philosophy to `LocalInstance`.

src/api/credrank.js
```diff
@@ -0,0 +1,76 @@
...
+|};
+
+/**
+  This is a primary SourceCred API that combines the given inputs into a single
+  WeightedGraph and then runs the Credrank algorithm on it to create a CredGraph
```
wchargin, line 37: nit: CredRank (caps) in prose. In code identifiers, we camel-case things
like `loadGithubRepo` and `runCredrank` because it’s more consistent
(otherwise you run into junk like `XMLHttpRequest`), but we use the
proper English spellings when spelling English.

wchargin (commented):
I support this general idea. Some comments inline.

src/api/instance/instance.js
blueridger: accepted

src/api/instance/instance.js
blueridger: The reason it's lopsided for now is that I only added the minimum needed for credrank, but plan to expand it for other commands/usage. I'm down to combine the writes like you suggest right now.

src/api/instance/localInstanceImpl.js
blueridger: accepted

src/api/instance/localInstanceImpl.js
blueridger: accepted

src/cli/credrank.js
blueridger: It was causing a namespace conflict with the credrank cli function that it's being called from.

src/api/instance/localInstanceImpl.js
blueridger: accepted

src/api/credrank.js
blueridger: accepted

src/api/credrank.js
blueridger: accepted

src/api/instance/localInstanceImpl.js
blueridger: I personally appreciate them for the purpose of easily navigating/scrolling through the file, and also for encouraging future functions to be placed appropriately.

src/api/instance/instance.js
blueridger: Yeah, I also thought this would be useful in the future. Sure, it's pretty simple to add in.

src/api/instance/instance.js
blueridger: I'll apply most this feedback, though I don't want to narrow the scope to just credrank, since the plan is to incrementally add the inputs and outputs for other APIs. (I could stub them, but would rather leave them out for now.)

src/api/credrank.js
blueridger: accepted

src/api/instance/localInstanceImpl.js
blueridger: accepted `await`

blueridger:
Applied feedback and broke it up into 2 PRs. https://github.com/sourcecred/sourcecred/pull/2767/

src/api/credrank.js
```diff
@@ -0,0 +1,76 @@
...
+
+  Might mutate the ledger that is passed in.
+ */
+export async function credrank(input: CredrankInput): Promise<CredrankOutput> {
+  let weightedGraph = overrideWeights(
+    merge(input.pluginGraphs),
+    input.weightOverrides
+  );
+  const dependenciesWithIds = input.dependencies.map((d) =>
+    // This mutates the ledger, adding new identites when needed.
```
wchargin, line 48: sp.: *identities*

src/api/instance/instance.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import {type CredrankInput, type CredrankOutput} from "../credrank";
+import {CredGraph} from "../../core/credrank/credGraph";
+
+/**
+  Simple read/write interface for inputs and outputs of the main SourceCred API.
```
wchargin, line 6: The newly factored `ReadOnlyInstance` isn’t read-write; fix docs?

wchargin (commented):
I can’t reply on-threads due to force-push, so responses here:

> > nit: Why `callCredrank` rather than just `credrank`? All function calls
> > are calls; this just confuses me when I read it because I have to figure
> > out what it does that’s special (answer: nothing).
>
> It was causing a namespace conflict with the credrank cli function that it's being called from.

It doesn’t seem to be, though? There is no other symbol `credrank` in
scope (the function that it’s called from is `credrankCommand`), and
renaming it back to `credrank` doesn’t introduce any Flow errors.

> The reason it's lopsided for now is that I only added the minimum needed for credrank, but plan to expand it for other commands/usage. I'm down to combine the writes like you suggest right now.

Even if we add other commands, we could keep the `readFooInput` and
`writeFooOutput` structure rather than requiring clients to piece
together the puzzle from components. (If there’s concern around reading
some sub-inputs multiple times, the `Instance` implementation can cache
actual disk I/O.)

I also think that it’s worth considering whether you’d want separate
commands to have their own interfaces: having targeted, thin interfaces
as opposed to “god objects” can be nice. Rust, Swift, and Go do this
extensively, and it works well.

In any case, happy to revisit once there are more commands and something
concrete to discuss.

> broke it up into 2 PRs

Including the use case in this PR made it easier to see how the API is
used and how well it fits clients’ needs. In general, we say that each
commit should include “the minimal amount of code required to implement,
test, and document that change”, and for new APIs, the actual clients
are a kind of test for the design. Especially since the other PR is
small (20 LOC), would you mind recombining?

blueridger:
feedback accepted.

decentralion (approved):
I really like this general approach and have been wanting to see something like this! However, I think the interface is too tightly coupled to the immediate needs of implementing the CredRank CLI command. I encourage thinking about a second use case, e.g. doing a Cred analysis workflow in Observable. In that case, we want to be able to:

- instantiate an instance by pointing to the branch url on GitHub
- make changes to the instance's configuration, e.g. changing plugin weights
- recompute the Cred scores in browser
- retrieve scores etc for analysis

Here's an idea for an interface that could fit these use cases better:

```js

export interface ReadOnlyInstanceConfig {
  activePlugins(): Promise<$ReadOnlyArray<PluginName>>;
  ledger(): Promise<ReadOnlyLedger>; // nb. really should be ReadOnlyLedger, altho that type doesn't actually exist
  // Make a new mutable instance with the same settings
  fork(): Promise<InstanceConfig>;
  // etc
  // doesn't have CredGraph
}

export interface InstanceConfig extends ReadOnlyInstanceConfig {
  setActivePlugins($ReadOnlyArray<PluginName>);
  setLedger(Ledger);
  // etc
}

// A read-only type that combines a CredGraph with a ReadOnlyInstanceConfig.
// Guarantees that the CredGraph is sync'd with the config.
export interface Instance {
  config(): Promise<ReadOnlyInstanceConfig>;
  credGraph(): Promise<ReadOnlyCredGraph>;
  // maybe put "peek-thru" helper methods for getting stuff out of the config conveniently

  static compute(config: ReadOnlyInstanceConfig): Promise<Instance>;
}
```

Then, we can also implement an interface like

```js
export interface InstanceSerializer {
  writeConfig(InstanceConfig)
  writeInstance(Instance)
  readConfig(): Promise<instanceConfig>
  readInstance(): Promise<Instance>
}
```

And implement `DiskInstanceSerializer`, `NetworkInstanceSerializer`, etc.

This PR is a really big improvement as is, so we don't need to implement these refactors here, I'm happy for this to be a followon that's owned by refactoring team (e.g. me).

P.S. I approve this PR, but I'm kind of a drive-by review (havne't been reading the rest of the discussion) so please consider William's approval canoncial. :)

sharness/__snapshots__/test-instance/config/dependencies.json
```diff
@@ -4,7 +4,11 @@
...
+        "periods": [
+            {
+                "startTime": "2019-01-01",
+                "weight": 0.05
+            }
+        ]
     }
-]
-
+]
```
wchargin, line 14: Why these changes, especially the missing newline at end of file? They
look unrelated to this change, and `yarn test` passes for me without
them. Please fix?

wchargin (approved):
To [@decentralion’s review][1]: I’d suggest not tackling that now. It’s
more complicated, and I would have some questions about your
formulation.

[1]: https://github.com/sourcecred/sourcecred/pull/2764#pullrequestreview-595858813

sharness/__snapshots__/test-instance/config/dependencies.json
blueridger: Its required for yarn test-full. It was slightly inaccurate for me to say this PR doesn't change behavior. Prior to this change, writes to the configuration would only happen if the configuration had been changed by the code, but now the configuration always is written. I chose this because it made the code simpler and we can put that optimization back in later if we want. But this means that the expected snapshot needs to match the computer-generated json. This file was edited for this PR by running the update-snapshots script.

wchargin: Ack; thanks.

PR #2767, opened 2021-02-19
Use new Instance and CredRank API in the CLI

blueridger:
# Description
This uses the Instance interface and CredRank API in the CredRank CLI. More details and test plan can be found in https://github.com/sourcecred/sourcecred/pull/2764

### DIFF
--- a/sharness/__snapshots__/test-instance/config/dependencies.json
+++ b/sharness/__snapshots__/test-instance/config/dependencies.json
@@ -4,7 +4,11 @@
         "autoInjectStartingPeriodWeight": 0.05,
         "id": "2zWwkGHnBGXmEcYneeWAzw",
         "name": "SourceCred",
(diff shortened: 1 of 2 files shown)


wchargin (commented):
Copy-pasting [comment from #2764][1]:

> Including the use case in this PR made it easier to see how the API is
> used and how well it fits clients’ needs. In general, we say that each
> commit should include “the minimal amount of code required to implement,
> test, and document that change”, and for new APIs, the actual clients
> are a kind of test for the design. Especially since the other PR is
> small (20 LOC), would you mind recombining?

Further evidence: the fact that this PR has to refer to the other for
its test plan suggests that it’s not really self-contained.

[1]: https://github.com/sourcecred/sourcecred/pull/2764#pullrequestreview-594900672

Issue #2768, opened 2021-02-19
Compressed storage migration path


wchargin:
(Paired with: @topocount, @hammadj, @wchargin.)

## Motivation
Users are hitting GitHub file size limits when their `credResult.json` file is larger than 100 MB, the hard cap for GitHub's API.

Goal: unblock users from deploying to GitHub Pages even if files like `credResult.json` exceed 100 MB in size.

## Plan
We'll compress data within physical files to honor GitHub's per-file size limit. This should be a fairly surgical change, without requiring any new technologies or patterns.

Define two interfaces:

```javascript
/*
 * A byte string represented as a JavaScript string with
 * code points between 0 and 255, inclusive.
 */
type ByteString = string;

interface DataStorage {
  get(key: string): Promise<ByteString>;
  // may also want "list"
}

interface WritableDataStorage extends DataStorage {
  set(key: string): Promise<ByteString>;
  // may also want "delete", "delete by prefix", ...
}
```

* Implement DiskStorage (node only) that implements the `WriteableDataStorage` interface
* Update CLI to use `DiskStorage` instead of using `fs` directly
* Rewrite `ledgerDiskStorage` to `NetworkStorage` (implementing `WriteableDataStorage`, for arbitrary files rather than just the ledger)
* Change `LedgerManager` usage on frontend to use `NetworkStorage`
* Implement `ZipStorage`, which can wrap any `DataStorage` implementation to shard logical files across physical files.
* Replace usages of `DiskStorage` and `NetworkStorage` in the CLI and frontend with sharded equivalents.

## Non-goals, future work
We considered using Git LFS. It may well be a better long-term solution. For now, we don't want to incur the additional complexity for either us or our users.

As we shard files, we still write them to disk, and require manual pushes to GitHub by instance maintainers. Direct-to-remote writes are a whole separate can of worms.

The fact that we have to resort to sharding is a clear signal that we should actively investigate more sustainable storage solutions for the long term. Happily, we're already doing that.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038250955742) by [Unito](https://www.unito.io/learn-more)

wchargin:
cc @decentralion; please review and LGTM.

cc @blueridger: advisory re: your API/IO changes; please feel free to
offer any feedback.

topocount:
[...]

PR #2776, opened 2021-02-22
Implement DiskStorage

topocount:
DiskStorage is a low-level class intended for I/O with a local disk. It
implements the WritableDataStorage interface, which is meant to allow
for uniform abstractions for any potential I/O the codebase interacts
with.

Payloads are specified as byte strings. This type is meant to enforce the
abstract representation of blobs. Uint8Arrays are chosen as the
underlying type because they are supported in both the browser and
NodeJS. Additionally, they cannot be confused for builtin JavaScript
strings.

builds towards #2768
merge next: #2777 

test plan: unit tests are provided

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038249657730) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/core/storage/disk.js
+++ b/src/core/storage/disk.js
@@ -1,0 +1,57 @@
+// @flow
+
+import type {WritableDataStorage} from "./index";
+import {join as pathJoin} from "path";
+import normalize from "../../util/pathNormalize";
+import fs from "fs-extra";
+
+/**
+ * Disk Storage abstracts away low-level file I/O operations.
+ */
+export class DiskStorage implements WritableDataStorage {
+  +_basePath: string;
+
+  /**
+   * basePath will resolve to '.' if an empty string is passed in
+   */
+  constructor(basePath: string) {
+    this._basePath = normalize(basePath);
+  }
+
+  /**
+   * The `path` parameter must be relative to the `basePath` set at
+   * construction. Any I/O errors when attempting to read contents at the path
+   * will cause an error to be thrown.
+   */
+  async get(path: string): Promise<Uint8Array> {
+    const validPath = this._checkPathPrefix(path);
+    return fs.readFile(validPath);
+  }
+
+  /**
+   * The `path` parameter must be relative to the `basePath` set at
+   * construction. Any I/O errors when attempting to write contents at the path
+   * will cause an error to be thrown.
+   */
+  async set(path: string, contents: Uint8Array): Promise<void> {
+    const validPath = this._checkPathPrefix(path);
+    return fs.writeFile(validPath, contents);
+  }
+
+  /**
+   * `path` is a relative file path to the `basePath`. The normalized
+   * result cannot be above the instance's base path. This is ensures that any
+   * attempts at accessing the file system outside of the instance
+   * will cause an error to throw.
+   */
+  _checkPathPrefix(path: string): string {
+    const fullPath = normalize(pathJoin(this._basePath, path));
+    // Ensure normalized path isn't outside basePath
+    if (!(fullPath.startsWith(this._basePath) || this._basePath === ".")) {
+      throw new Error(
+        "Path construction error; possible path traversal attack"
+      );
+    }
+    return fullPath;
+  }
+}
--- a/src/core/storage/disk.test.js
+++ b/src/core/storage/disk.test.js
@@ -1,0 +1,104 @@
+// @flow
+import {DiskStorage} from "./disk";
+import tmp from "tmp";
+import fs from "fs-extra";
+import {join as pathJoin} from "path";
+// $FlowIgnore[missing-export]
+import {TextEncoder, TextDecoder} from "util";
+
+describe("core/storage/disk", () => {
+  function tmpWithContents(contents: mixed, dir: string = basedir.name) {
+    const name: string = tmp.tmpNameSync({dir});
+    fs.writeFileSync(name, JSON.stringify(contents));
+    const fname = name.split("/").pop();
+    return fname;
+  }
+  const badPath = () => pathJoin(tmp.dirSync().name, "not-a-real-path");
+  const fooDefault = () => ({foo: 1337});
+  const basedir = tmp.dirSync();
+  const encoder = new TextEncoder();
+  // $FlowIgnore[incompatible-call]
+  const decoder = new TextDecoder();
+
+  describe("get", () => {
+    it("returns an existing file", async () => {
+      const instance = new DiskStorage(basedir.name);
+      const fname = tmpWithContents(fooDefault());
+
+      const result = await instance.get(fname);
+      expect(JSON.parse(decoder.decode(result))).toEqual(fooDefault());
+    });
+    it("throws if file doesn't exist", async () => {
+      const instance = new DiskStorage(basedir.name);
+      const thunk = () => instance.get(badPath());
+      await expect(thunk).rejects.toThrow("ENOENT");
+    });
+    it("throws if a normalized path is outside the base path", async () => {
+      expect.hasAssertions();
+      const instance = new DiskStorage(basedir.name);
+      const thunk = () => instance.get("../falseFile");
+      expect(thunk).rejects.toThrow("Path construction error");
+    });
+  });
+  describe("set", () => {
+    // needed because Uint8Arrays are not deeply equal to buffers according to
+    // jest, so we explicitly compare the values
+    function compareUint8Arrays(array1: Uint8Array, array2: Uint8Array) {
+      array2.forEach((value, i) => {
+        expect(value).toBe(array1[i]);
+      });
+      expect(array1.length).toBe(array1.length);
+    }
+    const contents = [
+      ["test.txt", encoder.encode("hello")],
+      ["testagain.txt", new Uint8Array([0xab, 0xcd])],
+    ];
+    it("invalid UTF-8 cannot survive a roundtrip through text decoder/encoder", () => {
+      // This is a sanity check to show that data that doesn't map to a native
+      // type can still be handled by DiskStorage
+      const roundtrip = (a) =>
+        // $FlowIssue[incompatible-call]
+        new TextEncoder().encode(new TextDecoder().decode(a));
+
+      const [_, invalidText] = contents[1];
+      expect(roundtrip(invalidText).length).not.toEqual(invalidText.length);
+    });
+    it("can write a file to disk", async () => {
+      expect.assertions(9);
+      const instance = new DiskStorage(basedir.name);
+      for (const [fileName, content] of contents) {
+        await instance.set(fileName, content);
+        const result = await instance.get(fileName);
+        compareUint8Arrays(result, content);
+      }
+    });
+    it("overwrites existing content", async () => {
+      expect.assertions(9);
+      const instance = new DiskStorage(basedir.name);
+      const [fileName] = contents[0];
+      for (const [_, content] of contents) {
+        await instance.set(fileName, content);
+        const result = await instance.get(fileName);
+        compareUint8Arrays(result, content);
+      }
+    });
+    it("throws if a normalized path is outside the base path", async () => {
+      expect.hasAssertions();
+      const instance = new DiskStorage(basedir.name);
+      const thunk = () =>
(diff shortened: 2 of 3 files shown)


topocount:
@wchargin for these class implementations, would it make sense to add higher-level helper functions that ease the conversion from some builtin type to a `ByteString`? I have these utility functions:

https://github.com/sourcecred/sourcecred/blob/9e7be522a184bbdd24cc02e3718958736466f701/src/core/storage/index.js#L29-L46

But I feel inclined to internalize them to the `DiskStorage` class or at least write helper methods that utilize them and abstract away the `set ` and `get` calls

src/core/storage/disk.js
```diff
@@ -0,0 +1,53 @@
+// @flow
+
+import type {ByteString, WritableDataStorage} from "./index";
+import {join as pathJoin, normalize} from "path";
+import fs from "fs-extra";
+
+/**
+ * Disk Storage abstracts away low-level file I/O operations, implementing the
+ * Writable Data Storage interface. In this implementation, `path` aliases to
+ * a key, and the `contents` parameter maps to the value.
```
wchargin, line 10: > In this implementation, `path` aliases to a key, and the `contents`
> parameter maps to the value.

What does this mean? What is aliasing, what does it mean to “alias to a
key”, and what does it mean to “map to the value”? Are these doc
comments better suited for specific methods? (The `DiskStorage` type
itself does not have `path` or `contents` parameters.) I’m not sure what
you’re trying to convey.

src/core/storage/disk.js
```diff
@@ -0,0 +1,53 @@
...
+export class DiskStorage implements WritableDataStorage {
+  +_basePath: string;
+
+  constructor(basePath?: string) {
+    this._basePath = basePath != null ? basePath : process.cwd();
+  }
+
+  /**
+   * Get receives a file path relative to the base path set at construction
+   * and returns the file contents encoded as a byte string.
```
wchargin, line 21: Please don’t do the golang doc thing where we pretend that symbols are
case-insensitive proper nouns. This is hard to read because it’s
ungrammatical—“Get receives” isn’t English. A simple transliteration
would be “The `get` method receives […]” (with backticks), but since
this is documentation on the `get` method, that’s superfluous.

Which parts of what you’re documenting are not obvious from the method
signature? Specify those, and leave the rest out.

src/core/storage/disk.js
```diff
@@ -0,0 +1,53 @@
...
+   */
+  async get(path: string): Promise<ByteString> {
+    const validPath = this._checkPathPrefix(path);
+    return fs.readFile(validPath);
+  }
+
+  /**
+   * Set accepts the a file path relative to the base path along with the file
```
wchargin, line 29: nit: Duplicate determiner “*the a*”, plus advise above.

src/core/storage/disk.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /**
+   * _checkPathPrefix accepts a relative file path then verifies that absolute
+   * file is not outside the instance's base path. This is ensures that any
+   * attempts at accessing the file system outside of the instance
+   * will cause an error to throw.
+   */
+  _checkPathPrefix(path: string): string {
+    const fullPath = normalize(pathJoin(this._basePath, path));
```
wchargin, line 44: This `normalize` is platform-dependent. Can we use `util/pathNormalize`
so that it’s (a) portable and (b) equivalent to what we’d use on the
frontend? It’s a drop-in replacement.

src/core/storage/disk.js
```diff
@@ -0,0 +1,53 @@
...
+    // Ensure normalized path isn't outside basePath
+    if (!fullPath.startsWith(this._basePath)) {
```
wchargin, line 46: This still leaks if `_basePath === ""`, no?

src/core/storage/disk.test.js
```diff
@@ -0,0 +1,69 @@
...
+  const fooDefault = () => ({foo: 1337});
+  const getInstance = (baseDir?: string) => new DiskStorage(baseDir || "");
+  const basedir = tmp.dirSync();
+  const encoder = new TextEncoder();
+  // $FlowIgnore[incompatible-call]
+  const decoder = new TextDecoder();
+
+  it("DiskStorage accepts a basepath", () => {
+    const instance = new DiskStorage(tmpdir());
+    expect(instance._basePath).toBe(tmpdir());
```
wchargin, line 26: Please don’t poke into the private implementation details for tests.
What behavior are you trying to test here? Can it be observed through
the public interface? If so, observe it; if not, by definition, it
doesn’t matter.

src/core/storage/disk.test.js
```diff
@@ -0,0 +1,69 @@
...
+    it("returns an existing file", async () => {
+      const instance = getInstance("");
+      const fname = tmpWithContents(fooDefault());
+
+      const result = await instance.get(fname);
+      expect(JSON.parse(decoder.decode(result))).toEqual(fooDefault());
+    });
+    it("throws if file doesn't exist", async () => {
+      const instance = getInstance();
+      const result = async () => await instance.get(badPath());
```
wchargin, line 39: We usually call this `thunk` or something, since it’s not actually a
result at all; rename for clarity?

wchargin, line 39: This `async() => await ...` is superfluous; can just be `() => ...`.

src/core/storage/disk.test.js
```diff
@@ -0,0 +1,69 @@
...
+    it("overwrites existing content", async () => {
+      const instance = getInstance(basedir.name);
+      const newContents = encoder.encode("goodBye");
+      await instance.set(fileName, contents);
+      await instance.set(fileName, newContents);
+      const result = await instance.get(fileName);
+      expect(decoder.decode(result)).toBe(decoder.decode(newContents));
+    });
+  });
+  describe("_checkPathPrefix", () => {
```
wchargin, line 62: As before, this is testing private implementation details. It’s also
_not_ testing the important thing, which is that actually calling `get`
or `set` with a traversing path is caught. (What if the calls to the
validation function get dropped in a botched refactoring?) Please do
test this, but via the public API.

src/core/storage/disk.test.js
```diff
@@ -0,0 +1,69 @@
...
+      const result = await instance.get(fname);
+      expect(JSON.parse(decoder.decode(result))).toEqual(fooDefault());
+    });
+    it("throws if file doesn't exist", async () => {
+      const instance = getInstance();
+      const result = async () => await instance.get(badPath());
+      await expect(result).rejects.toThrow("ENOENT");
+    });
+  });
+  describe("set", () => {
```
wchargin, line 43: It’s okay to be using `TextEncoder`/`TextDecoder` for test data, but
this module really operates on bytes (as it must). Please include some
tests where the data is not valid JSON and not valid UTF-8. I want these
tests to catch an error given by, say, an accidental roundtrip through
encoding. If you don’t know where to start, take this building block:

```javascript
> var b1 = new Uint8Array([0x11]), b2 = new Uint8Array([0x99]);
> var roundtrip = (b) => new TextEncoder().encode(new TextDecoder().decode(b));
> [b1, roundtrip(b1)];
[ Uint8Array(1) [ 17 ], Uint8Array(1) [ 17 ] ]
> [b1, roundtrip(b2)];
[ Uint8Array(1) [ 17 ], Uint8Array(3) [ 239, 191, 189 ] ]
```

src/core/storage/index.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+// $FlowIgnore[missing-export]
+import {TextEncoder, TextDecoder} from "util";
+
+/**
+ * A byte string represented as a Uint8Array. Each item in a byteString array
+ * is a Uint8 representing a byte value.
+ */
+export type ByteString = Uint8Array;
```
wchargin, line 9: Now that a `ByteString` is just a `Uint8Array` rather than a type with
side conditions, I think that we can just drop this type alias entirely.
What do you think?

src/core/storage/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Data Storage is meant as a low-level interface to data that is not
+ * already in process memory. If the value for the key entered cannot be found,
+ * an error should throw.
```
wchargin, line 14: documentation:

  - Again, “Data Storage” isn’t anything; `DataStorage` is the name of
    this interface.
  - Don’t tell me that it’s meant as something. You are the author and
    may be authoritative. If it’s meant as a low-level interface to
    data, say “A low-level interface to data […]”.
  - There’s no _requirement_ that the data not be in process memory: a
    `MemoryDataStorage` is in fact a useful testing utility that we may
    want to define. Rephrase to make it clear that implementations are
    merely *permitted* to perform I/O?
  - Errors do not throw; errors are thrown.
  - “the value for the key”—what key? `DataStorage` does not have keys.
    This comment is appropriate for the `get` method, and should live
    there instead.

wchargin, line 14: Can we at least mention something about keys being filesystem-friendly?
We can flesh out the details later.

src/core/storage/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * fromByteString exists as a helper for converting byte strings read out of a
+ * DataStorage implementation into regular javascript strings.
+ */
+export function fromByteString(b: ByteString): string {
+  // $FlowIgnore[incompatible-call]
+  const decoder = new TextDecoder();
+  return decoder.decode(b);
+}
```
wchargin, line 46: I vote to drop these. The Node-specific `util` dependency complicates
the situation, and these are inlinable as `new TextEncoder().encode` and
`new TextDecoder().decode`, which are clearer re: encoding, anyway.

wchargin (commented):
Cool; I like this overall. Some comments inline.

> would it make sense to add higher-level methods that ease the conversion from some builtin type to a `ByteString`?

Let’s leave them out for now. We will want something to that effect
later, but I think that the final form will end up different. Add them
when you have a use for them, and let that use shape them.

src/core/storage/disk.js
topocount: Using the portable `normalize` will make this impossible. Users can still pass in a root `/` at construction, but an empty string now resolves to `'.'`.

src/core/storage/disk.test.js
topocount: @wchargin Thanks for this tip. I set one up as a sanity check, though I didn't throw an error. I just used it to demonstrate that some data that works with this implementation cannot be processed with Text encoder

src/core/storage/disk.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+describe("core/storage/disk", () => {
+  function tmpWithContents(contents: mixed, dir: string = basedir.name) {
+    const name: string = tmp.tmpNameSync({dir});
+    fs.writeFileSync(name, JSON.stringify(contents));
+    const fname = name.split("/").pop();
+    return fname;
+  }
+  const badPath = () => pathJoin(tmp.dirSync().name, "not-a-real-path");
+  const fooDefault = () => ({foo: 1337});
```
blueridger, line 17: Curious why these are methods. Maybe to ensure test data isn't crossing over from test to test? The way I'm used to seeing this done is:
```
  let badPath, fooDefault;
  beforeEach(() => {
    badPath = pathJoin(tmp.dirSync().name, "not-a-real-path");
    foodDefault = {foo: 1337};
  });
```
I see advantages to both.

blueridger (approved):

src/core/storage/disk.test.js
topocount: I think we've been avoiding using test hooks like this because it makes refactoring more difficult. Debugging test hook behavior can be quite painful sometimes.

src/core/storage/disk.test.js
```diff
@@ -0,0 +1,105 @@
...
+  const getInstance = (baseDir: string) => new DiskStorage(baseDir);
```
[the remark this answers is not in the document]
topocount: This was vestigially useful. Thanks for the catch

PR #2785, opened 2021-02-24
Improve save ledger interaction

thr3m:
# Description
Wasnt able to implement the popup when the form state is "dirty". Tried every single stackoverflow issues. Might be a problem with React Hooks.

Paired with @felixbbertrand 

# Test Plan

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038210152930) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/ui/components/Transfer.js
+++ b/src/ui/components/Transfer.js
@@ -124,7 +124,7 @@ export const Transfer = ({
 
   const handleSaveToLedgerWarning = (_) => {
     if (
-      isSavedToLedger || 
+      isSavedToLedger ||
       sender !== null ||
       receiver !== null ||
       amount !== ""
--- a/src/ui/components/Transfer.js
+++ b/src/ui/components/Transfer.js
@@ -113,17 +113,21 @@ export const Transfer = ({
     }
   };
 
-  const handleSaveToDisk = (_) => {
+  const handleSaveToDisk = () => {
     setIsSavedToLedger(!isSavedToLedger);
     saveToDisk();
+    setMemo("");
+    setReceiver(null);
+    setSender(null);
+    setAmount("");
   };
 
   const handleSaveToLedgerWarning = (_) => {
     if (
-      isSavedToLedger ||
+      isSavedToLedger || 
       sender !== null ||
       receiver !== null ||
-      amount !== null
+      amount !== ""
     ) {
       return (
         <div className={classes.saveToLedgerAlert}>
--- a/src/ui/components/Transfer.js
+++ b/src/ui/components/Transfer.js
@@ -1,5 +1,5 @@
 // @flow
-import React, {useState, type Node as ReactNode} from "react";
+import React, {useEffect, useState, type Node as ReactNode} from "react";
 import {Button, Container, TextField, useMediaQuery} from "@material-ui/core";
 import {makeStyles} from "@material-ui/core/styles";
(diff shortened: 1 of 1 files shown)


KuraFire:
> Wasnt able to implement the popup when the form state is "dirty".

@thr3m Do you mean the alert popup when navigating to a different page when the Ledger hasn't been saved to disk yet?

I'm curious what might be the cause there, because this is implemented on Facebook very aggressively, i.e. if you start writing a post and then try to navigate away, it'll prompt you beforehand about your "unfinished post" — and it does so even under circumstances where it no longer makes sense sometimes (like an emptied field leaving nothing as "unfinished post").

If this cannot easily be implemented, it's fine to skip the alert popup for now, but I'd like to see what the UI looks like when the ledger is in a dirty state before I approve this with a UI review.

thr3m:
Screen Recording 2021-02-24 at 7 25 15 PM

thr3m:
https://user-images.githubusercontent.com/31906840/109085153-2114f580-76d7-11eb-91c0-c06441e42ee7.mov

KuraFire (approved):
Solid =)

src/ui/components/Transfer.js
```diff
@@ -97,12 +106,37 @@ export const Transfer = ({
       });
       updateLedger(nextLedger);
       setAmount("");
+      setIsSavedToLedger(!isSavedToLedger);
```
blueridger, line 109: This is a bug. If you do 2 transfers, then isSavedToLedger would end up being false. This should be `setHasUnsavedChanges(true);`

src/ui/components/Transfer.js
```diff
@@ -97,12 +106,37 @@ export const Transfer = ({
...
       setSender(nextLedger.account(sender.identity.id));
       setReceiver(nextLedger.account(receiver.identity.id));
       setMemo("");
     }
   };
 
+  const handleSaveToDisk = () => {
+    setIsSavedToLedger(!isSavedToLedger);
```
blueridger, line 117: `setHasUnsavedChanges(false);`

src/ui/components/Transfer.js
```diff
@@ -79,6 +87,7 @@ export const Transfer = ({
   const [receiver, setReceiver] = useState<Account | null>(null);
   const [amount, setAmount] = useState<string>("");
   const [memo, setMemo] = useState<string>("");
+  const [isSavedToLedger, setIsSavedToLedger] = useState<boolean>(false);
```
blueridger, line 90: This name is a little confusing, because I expect it to be `false` in dirty, but it's actually the opposite. A clearer name would be `hasUnsavedChanges`

blueridger (changes requested):
Please make these changes, and test the case where a user makes 2 transfers before saving to disk.

blueridger:
The diff includes a bunch of stuff it shouldn't. Needs rebased.

src/ui/components/Transfer.js
```diff
@@ -97,12 +106,37 @@ export const Transfer = ({
       });
       updateLedger(nextLedger);
       setAmount("");
+      setHasUnsavedChanges(!hasUnsavedChanges);
```
blueridger, line 109: This still needs changed to `setHasUnsavedChanges(true);`

Remember to test the case where 2 transfers are done before saving to disk.

blueridger (approved):

PR #2786, opened 2021-02-25
PersonalAttributions: Add capability in MarkovProcessGraph

blueridger:
paired with @topocount 

### Naming justification
**"Personal"** differentiates this type of attribution/tithing from the somewhat similar "dependency" attribution/tithing.

**"Attribution"** was chosen over words like "tithe" and "tribute" because it doesn't come with the hierarchical baggage and it is simply a very accurate matching definition:
_the action of regarding something as being caused by a person or thing_
_the action of ascribing a work or remark to a particular author, artist, or person_

# Description

This Pull request implements a method for participants flowing a proportion of their cred to other participants during specified epochs. 

This is made possible through the creation of edge gadgets that connect participant epoch nodes. The weights are the product of the configured proportion and the Markov Process Graph's `beta` value`
# Test Plan
Unit tests have been provided for the Markov Process Graph.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038300240132) by [Unito](https://www.unito.io/learn-more)


src/core/credrank/edgeGadgets.js
```diff
@@ -212,3 +213,45 @@ export const backwardWebbingGadget: EdgeGadget<WebbingAddress> = (() => {
   };
   return {prefix, toRaw, fromRaw, markovEdge};
 })();
+
+export type PersonalAttributionAddress = {|
+  +epochStart: TimestampMs,
+  +fromParticipantId: IdentityId,
+  +toParticipantId: IdentityId,
+|};
```
decentralion, line 221: Seems right. 👍

src/core/credrank/personalAttribution.js
```diff
@@ -0,0 +1,150 @@
...
+    epochStarts: $ReadOnlyArray<TimestampMs>
+  ) {
+    // Validate that:
+    // 1. There is only 1 entry per [fromParticipantId, toParticipantId] pair.
+    // 2. Proportions are in chronological order.
+    // 3. Proportions are a number between 0 and 1.
+    const identityPairs = new Set();
+    for (const pa of personalAttributions) {
+      if (identityPairs.has([pa.fromParticipantId, pa.toParticipantId]))
+        throw `More than one attribution found from [${pa.fromParticipantId}] to [${pa.toParticipantId}]`;
```
decentralion, line 43: Hmm I think it's really important that we allow this value to change over time--e.g. Garnet gives Amethyst 20% in 2020 but only 10% in 2021 -- as I understand the type, that would be implemented by having multiple entries with teh same from->to but different proportions and timestamps.

src/core/credrank/personalAttribution.js
```diff
@@ -0,0 +1,150 @@
+// @flow
+
+import findLast from "lodash.findlast";
+import {type IdentityId, identityIdParser} from "../identity/id";
+import type {TimestampMs} from "../../util/timestamp";
+import * as C from "../../util/combo";
+
+export type PersonalAttributionProportion = {|
+  +timestampMs: TimestampMs,
```
decentralion, line 9: How do we express that Cred is flowing for a fixed period, e.g. the month of april 2021? It feels like having separate `startTimestampMs` and `endTimestampMs` would be nice, although that invites the question of what to do when the data has specified overlapping periods (do they sum? replace each other somehow?). We could also have "latest timestamp wins" semantics.

decentralion (commented):
Overall, looks reasonable; in particular the changes to MarkovProcessGraph seem right. I think we can make the logic in `personalAttribution.js` easier to follow if we orient it around intermediate data types that make our queries natural. I'm also unclear on what the intended semantics are around updating attribution weights over time, or how to specify fixed-duration personal attributions.

blueridger:
So to address your comments
1. The semantics of the PersonalAttribution are very similar to the [BonusPolicy](https://github.com/sourcecred/sourcecred/blob/master/src/core/bonusMinting.js#L42-L59) semantics that you authored. Note that each PersonalAttribution has an _array_ of proportions. So to create a fixed-period proportion, you would add a beginning proportion with the desired weight and then a second proportion with a value of 0 to end it.
2. I intentionally chose to use a class as the intermediate representation as an opinionated statement, contrasting the pattern of free-floating types and functions that you recommend. Instead of importing a bunch of utility functions that exist at an equal level of composition, grouping utility functions in a class allows fewer imports and a namespace hierarchy that makes the usage in MPG much more legible. This could also be done with `import * as moduleName` however I think classes have the added benefit of creating such component hierarchy within the module file itself.
3. I chose `$ReadOnlyMap<IdentityId, GroupedPersonalAttributions>` as the intermediate because it is memory efficient, it serializes back into PersonalAttributions easily, and it is sufficiently computationally efficient even though it requires a small amount of nested traversal to extract epoch-level information. This can be iterated later without too much trouble, if we decide we need to make it more efficient.
4. I am happy to create some private validation functions to slim down the constructor to 3 lines.

src/core/credrank/markovProcessGraph.test.js
```diff
@@ -154,6 +157,112 @@ describe("core/credrank/markovProcessGraph", () => {
...
       });
     });
+    describe("personalAttribution", () => {
+      it("can create a virtual markov edge for each entry", () => {
+        // this tests all attribution entries, regardless of configured
+        // intervals
+        const mpg = markovProcessGraph(attributions);
+        for (const attr of attributions) {
+          for (const recipient of attr.recipients) {
+            for (const proportion of recipient.proportions) {
```
blueridger, line 167: I don't think I can approve most of these tests. Way too much logic. This is problematic for a number of reasons:
1. Who will test the tests?
2. When reading this code, the expected output is obscured from me.
3. f(x) == f(x) isn't a very strong test (duplicating the logic and then saying "it's the same whether I run it here or there" tests the environment, not the logic)

Ideally, you would define the expected edges explicitly, rather than using for loops and arithmetic. An easy hack for this is to assert the actual against an empty array, and then verify and copy the actual from the failed test output.

src/core/credrank/markovProcessGraph.js
```diff
@@ -958,6 +958,7 @@ function virtualizedMarkovEdge(
         personalAttributionAddress.fromParticipantId,
         personalAttributionAddress.toParticipantId
       ) || 0;
+    if (proportionValue === 0) return null;
```
blueridger, line 961: Is this needed? is this purely for robustness, since it isn't really reachable due to address filtering?

src/core/credrank/testUtils.js
```diff
@@ -158,15 +158,43 @@ export const weightedGraph: () => WeightedGraph = () => ({
...
-export const markovProcessGraph: () => MarkovProcessGraph = () =>
-  MarkovProcessGraph.new(args());
+const attribution1 = {
+  fromParticipantId: participant1.id,
+  recipients: [
+    {
+      toParticipantId: participant2.id,
+      proportions: [
+        {timestampMs: -2, proportionValue: 0.2},
+        {timestampMs: -1, proportionValue: 0.1},
```
blueridger, line 177: Seems a little strange to test with negative timestamps.

src/core/credrank/markovProcessGraph.test.js
```diff
@@ -154,6 +157,112 @@ describe("core/credrank/markovProcessGraph", () => {
...
+          epochStart: interval.startTimeMs,
+          fromParticipantId: attr.fromParticipantId,
+          toParticipantId: recipient.toParticipantId,
+        };
+        const containsEdge = Array.from(mpg.edges()).includes(
+          personalAttributionGadget.markovEdge(details, 0)
+        );
+        expect(containsEdge).toBe(false);
+      });
+      it("filters out 0-weighted edges", () => {
```
blueridger, line 249: Seems like a strange test case. When would we ever encounter such an edge address, considering the above test case?

src/core/credrank/markovProcessGraph.test.js
```diff
@@ -154,6 +157,112 @@ describe("core/credrank/markovProcessGraph", () => {
...
+                .proportionValue
+            : 0;
+          const edge = personalAttributionGadget.markovEdge(
+            details,
+            parameters.beta * proportion
+          );
+          checkMarkovEdge(mpg, edge);
+        }
+      });
+      it("propagates the latest proportion into subsequent epochs", () => {
```
blueridger, line 208: I like the test cases you arranged for the most part.

src/core/credrank/markovProcessGraph.test.js
```diff
@@ -154,6 +157,112 @@ describe("core/credrank/markovProcessGraph", () => {
...
+        const attr = attributions[1];
+        const recipient = attr.recipients[0];
+        const details = {
+          epochStart: interval.startTimeMs,
+          fromParticipantId: attr.fromParticipantId,
+          toParticipantId: recipient.toParticipantId,
+        };
+        const containsEdge = Array.from(mpg.edges()).includes(
+          personalAttributionGadget.markovEdge(details, 0)
+        );
```
blueridger, line 246: This is probably the maximum amount of logic that would fit within best practice.

src/core/credrank/testUtils.js
topocount: Negative timestamps just represent time before January 1, 1970. They should be valid.

src/core/credrank/markovProcessGraph.test.js
topocount: My goal here was to make lookups using the `mpg.edge` query consistent with the `mpg.edges()` calls. Even though we're working with virtualized addresses, I figured this behavior should be consistent. Should this not be the case?

src/core/credrank/markovProcessGraph.js
topocount: This _is_ needed though, since it will yield 0-weighted edges when called from `mpg.edge()`. Do we want virtualized edges with 0-weight to return from this query?

src/core/credrank/markovProcessGraph.test.js
topocount: I was just following the test patterns outlined in this file, I typically hardcode the expected returns but I wanted the tests to be consistent with what already exists here. Will make these changes.

src/core/credrank/markovProcessGraph.js
blueridger: I mean that mpg.edge() call should never happen because where would someone get the address except by some sort of forceful construction? But it's fine, makes it more robust.

blueridger (commented):
Approved. Also added doc strings to personalAttribution.js

Since I was the original PR author, @topocount will need to approve the code as well before merging.

topocount (approved):

Issue #2229, opened 2020-09-09
New SC Explorer Dashboard: Top Chart


KuraFire:
Atomic-style ticket for #2227 — Top Chart component:

![SourceCred Explorer Dashboard - 1  Top Chart@2x](https://user-images.githubusercontent.com/33087/92655394-2c4da380-f2a6-11ea-8f8b-5d92d075fc5e.png)

**Specifications:**

* Chart is flexible width, 100% width of explorer container (meaning: the width of the `<table>` in the current explorer)
* Chart height: aim for 100px, check in with KuraFire if that doesn't render the lines well
* The X-axis of the Top Chart should have markers at the beginning (left), and end (present day, right). 
  * FUTURE EXPANSION: labels and more markers, but some data gathering is needed first.
  * FUTURE EXPANSION: hover mini-popovers on the chart lines that show number + date
* Chart has **3** lines (colors listed are for **DARK THEME**):
  * `#6174CC` Blue for **Cred**
  * `#FFAA3D` Yellow for **Grain**
  * ~~`#FDBBD1` Pink for **Participants** — see CAVEAT on **"Active Participants"** below~~
* The Legend in the bottom left matches the colors as small circular swatches without a border, unlike the image suggests (wait for visual UI comps for details; in the interim, use 1em height & width).

**Update 2021-02-05:** the Participants elements of the top chart (line, legend) & stats circle are taken out for the time being, pending future design.

#### ~~CAVEAT on **"Active Participants"**~~
~~The attached image of the Top Chart wireframe mentions "active participants"; this term is measured as follows:~~

> ~~Participants who made **a contribution that earned Cred**~~

~~The Top Chart and the "Contribution Chart" column in the Activity Table always show data for All Time. In the rest of the Explorer wireframe, it means:~~

> ~~Participants who made **a contribution that earned Cred** _during_ the selected Timeframe~~

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038574980018) by [Unito](https://www.unito.io/learn-more)

KuraFire:
@befitsandpiper Just a heads up, the Explorer UI is initially going to be designed in Dark mode, so I've updated the colors in this ticket to reflect the ones I'll be using.

blueridger:
V1 will not have active participants.

blueridger:
Acceptance Criteria:
- Cred and Grain aggregates are both shown as overlapping lines
- Cred and Grain are in the legend
[...]

PR #2787, opened 2021-02-26
Yaz/feb2021/top chart

saintmedusa:
# Description
This is a PR for the changes proposed by the [Top Chart story](https://github.com/sourcecred/sourcecred/issues/2229) in the [Feature Team Board](https://github.com/orgs/sourcecred/projects/6).  

In short, a Grain line has been added to the Top Chart line graph to accompany the existing Cred line. Within the `ExplorerHome` component, we added a timeline calculation for both grain and cred, grabbing data from the state `allParticipants`.   

We renamed `CredTimeline` to `ExplorerTimeline` because there is an existing component called `CredTimeline` and our changes allow it to display Grain as well. Within `ExplorerTimeline`, we updated the d3 code to include grain and tweaked how the x-axis is calculated to cover the entirety of the graph (before, the line would stop short to varying degrees, depending on the current interval, and it was not even visible for the week view). Now the line length is consistent across intervals. Our fix is a stop gap, by duplicating the last data point in the given interval. We suggest a more comprehensive change: the intervals defined in`TIMEFRAME_OPTIONS` in `ExplorerHome` should include enough data points to have a meaningful line. @magwalk will open up an issue for this and will drop a link in the comments of this PR. 

# Test Plan

This was tested manually on our webpack servers. All intervals are showing our changes correctly.
Screen Shot 2021-02-25 at 5 32 48 PM
Screen Shot 2021-02-25 at 5 32 34 PM
Screen Shot 2021-02-25 at 5 32 24 PM

# A Note on Coding Style
Maggie and I wrote this code with a high-priority intention for readability and clarity, which means there are more lines of code than other styles. While we respect the desire for upmost-efficiency, please consider this as you review.


magwalk:
This is the issue for doing a more meaningful fix for the top chart data: https://github.com/sourcecred/sourcecred/issues/2788

src/ui/components/ExplorerHome/ExplorerTimeline.js
```diff
@@ -0,0 +1,113 @@
...
+  const viewBox = `0 0 ${width} ${height}`;
+  const intervals = props.timelines.cred.length;
+  let grainAsNumber, range, grainValues;
+  if (grainExists) {
+    const grain = props.timelines.grain || [];
+    grainAsNumber = grain.map((g) => {
+      return Number(g);
+    });
+    const credAndGrain = props.timelines.cred.concat(grainAsNumber);
+    range = extent(credAndGrain);
```
blueridger, line 40: Is it possible to have 2 different ranges for cred and grain? We can't really expect that one won't be super inflated compared to the other, which would cause the less inflated one to appear like a straight line at the bottom if they share a range.

src/ui/components/ExplorerHome/ExplorerTimeline.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+  const grainExists = props.timelines.grain && true;
+  const width = props.width || 300;
+  const height = props.height || 25;
+  const viewBox = `0 0 ${width} ${height}`;
+  const intervals = props.timelines.cred.length;
+  let grainAsNumber, range, grainValues;
+  if (grainExists) {
+    const grain = props.timelines.grain || [];
+    grainAsNumber = grain.map((g) => {
```
blueridger, line 36: This would be better as a `const`

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -237,6 +242,43 @@ export const ExplorerHome = ({
...
+        return participant.grainEarnedPerInterval.slice();
+      }
+    );
+    const credAccumulator = (
+      credTimelineAccumalator: Array<number>,
+      participantCred
+    ) =>
+      credTimelineAccumalator.map(
+        (intervalTotal, i) => intervalTotal + participantCred[i]
+      );
```
blueridger, line 262:
```suggestion
    const credAccumulator = (
      credTimelineAccumulator: Array<number>,
      participant
    ) =>
      credTimelineAccumulator.map(
        (intervalTotal, i) => intervalTotal + participant.credPerInterval[i]
      );
```
I did read your note about using more variables to increase legibility, however I believe the code would be more legible if you removed lines 245-255 and changed your accumulators like this. And it would be more efficient.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -237,6 +242,43 @@ export const ExplorerHome = ({
...
+    const grainAccumulator = (
+      grainTimelineAccumalator: Array<Grain>,
```
blueridger, line 264: typo: `accumalator` -> `accumulator`

src/ui/components/ExplorerHome/ExplorerTimeline.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+    const credAndGrain = props.timelines.cred.concat(grainAsNumber);
+    range = extent(credAndGrain);
+    // This is a temporary/quick fix to ensure that the timeline lines fill the entire x-axis
+    // and ensures that a line is displayed when there is only one interval.
+    grainValues = grainAsNumber.concat(grainAsNumber[intervals - 1]);
+  } else {
+    range = extent(props.timelines.cred);
+  }
+  // This is a temporary/quick fix to ensure that the timeline lines fill the entire x-axis
```
blueridger, line 47: Kind of confused how added another element to the array makes it fill the entire x-axis. Shouldn't the x-axis expand with the array if you do this?

blueridger (approved):
So there are some things I'd like addressed, but since @BrianBelhumeur is waiting on this PR, I will merge this and you can make changes in a follow up PR.

src/ui/components/ExplorerHome/ExplorerTimeline.js
saintmedusa: That was our first approach and we ran into some poblems with that approach - because cred's range did not start at 0, it put the grain line below the x-axis. I thought about just simply setting the lower limit of the range to 0 but we opted for this approach instead. We'll give it another try, though!

src/ui/components/ExplorerHome/ExplorerTimeline.js
magwalk: You are correct! I think there is something weird happening with how we're using D3 here, because the line should take up the entire width of the x-axis, but for some reason it wasn't.

src/ui/components/ExplorerHome/ExplorerTimeline.js
saintmedusa: looks like we need to leave `grainValues` in a `let` statement on line 33 but everything else works as a `const`

src/ui/components/ExplorerHome/ExplorerHome.js
saintmedusa: Yes, that makes sense, but the current code is the result of lots of fenangling due to the fact that flow hates `.reduce`. We tried your changes and it made the flow monster angryyyyyyyyy - bascially, even though we are using `reduce` to construct a new array, flow complains that the array type of `allParticipants` (`Array<ParticipantCredGrain>`) is incompatible with the desired type of our new array (`Array<number>`).

saintmedusa: When you are back from your much needed break we could look at this code together to make it more pretty but we are going to leave it for now

src/ui/components/ExplorerHome/ExplorerTimeline.js
saintmedusa: the issue we opened should resolve this

PR #2802, opened 2021-03-01
Proposed Fixes to ExplorerHome/ExplorerTimeline from PR #2787

magwalk:
# Description
topchart-2

This PR is in response to some of the feedback we got from Thena on our last PR https://github.com/sourcecred/sourcecred/pull/2787

These changes include:
- Changing `grainAsNumber` to a const instead of a let
- Fixing typos in 2 variables
- Changing the `ExplorerTimeline` so there is a separate range for each line that is displayed

We do have some concerns about having the separate ranges for each line:
- For the single-week view, the lines overlap with each other.
- The lines don't make sense in relation to each other, in the future there needs to be an axis label that can communicate to the user what the line values are.  

Paired with @saintmedusa

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerTimeline.js
+++ b/src/ui/components/ExplorerHome/ExplorerTimeline.js
@@ -30,20 +30,19 @@ const ExplorerTimeline = (props: ExplorerTimelineProps): ReactNode => {
   const height = props.height || 25;
   const viewBox = `0 0 ${width} ${height}`;
   const intervals = props.timelines.cred.length;
-  let range, grainValues;
+
+  let grainValues;
   if (grainExists) {
     const grain = props.timelines.grain || [];
     const grainAsNumber = grain.map((g) => {
       return Number(g);
     });
-    const credAndGrain = props.timelines.cred.concat(grainAsNumber);
-    range = extent(credAndGrain);
+
     // This is a temporary/quick fix to ensure that the timeline lines fill the entire x-axis
     // and ensures that a line is displayed when there is only one interval.
     grainValues = grainAsNumber.concat(grainAsNumber[intervals - 1]);
-  } else {
-    range = extent(props.timelines.cred);
   }
+
   // This is a temporary/quick fix to ensure that the timeline lines fill the entire x-axis
   // and ensures that a line is displayed when there is only one interval.
   const credValues = props.timelines.cred.concat(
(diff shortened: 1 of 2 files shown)


src/ui/components/ExplorerHome/ExplorerTimeline.js
```diff
@@ -30,20 +30,19 @@ const ExplorerTimeline = (props: ExplorerTimelineProps): ReactNode => {
...
   const intervals = props.timelines.cred.length;
-  let grainAsNumber, range, grainValues;
+
+  let grainValues;
   if (grainExists) {
     const grain = props.timelines.grain || [];
-    grainAsNumber = grain.map((g) => {
+    const grainAsNumber = grain.map((g) => {
       return Number(g);
     });
```
BrianBelhumeur, line 39:
```suggestion
    const grainAsNumber = grain.map(Number);
```
because `Number()` is a function, you can actually shorten the map to this!

BrianBelhumeur (approved):

blueridger (approved):

KuraFire:
From a general UI perspective this looks fine, but I'm curious about the (sample?) data used in the instance that the video animation is made with, as it shows "This month" to have a Cred rise, but "All time" only shows a Cred bump (incl. back down) in the past but no Cred activity in the most recent end of the graph.

KuraFire:
@blueridger What's the point of having UI reviews set up in the structure if PRs constantly get merged before UI has a chance to review them?

blueridger:
Good point. For the explorer home, I'm trying to bias towards fast merges due to how much parallel development has been happening in these files. I've enabled a setting that will now require a codeowner approval before merging, which should help. There is a question here around whether design needs to sign off on every PR on a beta feature or if maybe there are broader checkpoints where we should wait for design.

blueridger:
> I'm curious about the (sample?) data used in the instance that the video animation is made with, as it shows "This month" to have a Cred rise, but "All time" only shows a Cred bump (incl. back down) in the past but no Cred activity in the most recent end of the graph
@KuraFire this is because in that sample data the change this month is negligibly small compared to the spike in all time, so it appears flat on all time.

KuraFire:
@blueridger That's a fair point! I think that if devs can get into the habit of doing UI screenshots or screen capture videos for the interaction(s) per the pull request, it's easy for the UI team to do a quick review & approval (or request changes/clarifications if needed), so that we don't have to do PRs going back in to address issues that got merged in a quick process and then caught after the fact. And if there IS a video/image, I'm fine with being tagged on Discord for a review request :)

I'm happy to do a quick "here's how to capture screenshots or video real easily" demo to the dev team to encourage this, but mine would be for Mac users, and I know at least some devs are on Windows and Linux. Thoughts?

PR #2803, opened 2021-03-02
Redirect target IdentityIds to map to base Account

### LINKED ISSUE #2753: Merging active, existing Identities Breaks Ledger History distribution popover
# Description
After merging myself and Beanow in the "Identities" View on SourceCred's production cred instance, I went back to the "Ledger History" view and received a React Admin Error when attempting to click into a distribution popover, I received this error:
Screenshot from 2021-02-12 11-17-39

I initially believed this is a symptom of a larger problem involving ledger consistency across all components, but I've found that saving the ledger to disk and reloading the page does not fix this issue, so it's an issue with the ledger not providing this account after it's been merged, or more correctly, not omitting this account from the popover table at all and foregoing this query.

# Detailed Replication Steps
1. In the `sourcecred/sourcecred` repo, run `git checkout master && yarn && yarn start`
2. navigate to the identites view ([http://localhost:8080/#/admin](http://localhost:8080/#/admin)) and merge "dandelion" into "william":
Peek 2021-02-12 11-32
3. navigate to the "Ledger History" View and click on a distribution pill: 
Peek 2021-02-12 11-35

topocount:
Because there is little chance of collision as future identities are
created, it is safe to redirect target IdentityIds to point to the current base
account in the `Ledger.prototype._accounts` Map.

A unit test has been added to ensure that old IdentityIds don't lose
their connection to latest baseAccount.

It'd be preferable to track identity merge history at some point, but
for now, this lightweight implementation will allow for sucessfull
account lookups when rendering older grain distributions and other events
that preceded a merge.

fixes #2753

test plan:
- Unit tests updated
- Follow the replication steps in #2753 to ensure no error is thrown in
react admin or the browser console


src/core/ledger/ledger.js
```diff
@@ -77,6 +77,8 @@ type MutableAccount = {|
   active: boolean,
   // key-value store of blockchain addresses the account receives grain to
   payoutAddresses: PayableAddressStore,
+  // allows for reverse-lookups of IdentityIds
```
blueridger, line 80: This comment could be better. What does the array contain? Would be worth noting that it contains the account's main identity id in addition to merged ids, since that isn't fully intuitive.

blueridger (approved):

PR #2804, opened 2021-03-02
ExplorerHome: fix duplicate timeline value

blueridger:
# Description
Currently, the ExplorerHome timeline feature duplicates the last value in the array no matter what. This changes it so that it only duplicates the value for arrays of size 1.

# Test Plan
Verified that "this week" and "last week" still show a horizontal line. Verified that "this month" and "all time" no longer have a horizontal line at the end.
chrome-capture


magwalk:
Solid fix.

KuraFire (approved):
lgtm 👍🏼

PR #2805, opened 2021-03-05
Add Isomorphic Text Encoder/Decoder import

topocount:
This is needed for node V10. Once it is deprecated,, these classes are globally
available in Observable, node V12+ and the browser.

test plan:
1. add this module to the api.js file.
2. `yarn build`

Node:

1. `node`
2. `const sc = require("./dist/server/api.js")`
3. `let t = sc.sourcecred.core.textEncoding`
4. `t.decode(t.encode("a"))` // should output "a"

Observable: https://observablehq.com/@topocount/encoder-test

browser: imported into ui/index.js an completed a round-trip test on
load

builds towards #2768
merge after: #2776 
merge next: #2777 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038246141847) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/core/storage/textEncoding.js
+++ b/src/core/storage/textEncoding.js
@@ -1,0 +1,31 @@
+// @flow
+
+import {
+  // $FlowIssue[missing-export]
+  TextEncoder as importedTextEncoder,
+  TextDecoder as importedTextDecoder,
+} from "util";
+
+let encoder, decoder;
+// ensure jest skips this condition
+if (typeof window !== "undefined" && typeof process === "undefined") {
+  if (!(window.TextEncoder && window.TextDecoder)) {
+    throw new Error("No Encoder classes available.");
+  }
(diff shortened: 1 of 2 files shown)


src/api/index.js
```diff
@@ -42,6 +43,7 @@ const api = {
       graphToMarkovChain,
     },
     graph,
+    textEncoding,
```
wchargin, line 46: Why expose these directly? Can we remove them?

wchargin (commented):
General approach seems good to me, and your set of test cases seems
good—thanks! Could you please add a unit test to act as a smoke test?

src/api/index.js
topocount: I had exposed them for testing purposes, but they can be removed

wchargin (approved):

PR #2817, opened 2021-03-09
Convert Explorer Home timeline summaries to for loop for greater efficiency

BrianBelhumeur:
# Description
In tidying up the Explorer Home page for launch, I noticed a bit of lag upon switching the timeline control, especially for longer periods (e.g. All Time). I profiled the code and saw that compiling the timeline data is fairly intensive. This is understandable given that we're looping through the array of all participants, and within that looping though all time periods and summing the info.

In order to speed this up, I refactored the maps and reduces into one pass for both cred and grain summaries. This sped things up around 20% for All Time calculations. Below are timings in milliseconds for This Week, Last Week, This Month, All Time in order.

### Original timings in ms
Screen Shot 2021-03-02 at 22 58 44

### Refactored timings in ms
Screen Shot 2021-03-02 at 23 36 05

# Test Plan
Tested manually with the following plan:

* Load page
* Click on different time periods
* Ensure that the top graph changes accordingly

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200040151419673) by [Unito](https://www.unito.io/learn-more)


saintmedusa:
So glad for the fix. Was cringing while writing it.

src/ui/components/ExplorerHome/ExplorerHome.js
```diff
@@ -244,40 +238,37 @@ export const ExplorerHome = ({
...
+
+        for (let i = 0; i < length; i++) {
+          credTotalsTimeline[i] =
+            credPerInterval[i] + credTotalsTimeline[i] || 0;
+
+          grainTotalsTimeline[i] = add(
+            grainEarnedPerInterval[i],
+            grainTotalsTimeline[i]
+          );
+        }
```
blueridger, line 263: This confuses me. I believe it's a true assumption that the length of the arrays should always be equal and the elements should always be present as a number/Grain value

BrianBelhumeur: > This confuses me. I believe it's a true assumption that the length of the arrays should always be equal and the elements should always be present as a number/Grain value

@blueridger I also believe the arrays being the same length and the values present is a valid assumption. My intention in including the `Math.max()` is defensive programming. If there were to be a case in which, say, one array were populated and the other zero-length, it wouldn't cause issues of valid data going missing. Do you think this scenario is too far-fetched to bother putting in a guard for? Or perhaps a better comment would make it less confusing?

blueridger: I think we can trust that assumption and remove the guard rail. Also, why doesn't 261 have an or zero like 257? I'm guessing Flow is not considering the index out of bounds possibility and bigInt is simply casting undefined as 0. Just surprising and may be worth making the zero conversion explicit.

BrianBelhumeur: Ok sounds good. 261 doesn't have an `|| 0` because, as you rightly surmised, it casts undefined as 0. However, this is only the behavior of the bigInt library, and not `BigInt` native implementation, which throws on `BigInt(undefined)`. I discovered this today in digging into PR # 2821. For forward compatibility, an `|| 0` guard is a good idea and I'll add it. I think flow didn't complain because the bigInt library isn't typed.

KuraFire (approved):
Always excited to see performance improvements! 😍

Issue #2298, opened 2020-09-28
Explorer Home frontend cleanup/final pass


befitsandpiper:
This ticket is to cleanup any minor styling/design spec issues remaining.

catch-all ticket can be found here: https://github.com/sourcecred/sourcecred/issues/2227

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038482971852) by [Unito](https://www.unito.io/learn-more)

blueridger:
Launch blocking feedback from @KuraFire 

- [ ] Add a second pagination utility row above the table header
- [ ] Remove the background fill color from the pagination utility rows (top and bottom)
- [ ] Checkboxes at the bottom should start out checked
- [ ] when no checkboxes are checked the table should be empty
- [ ] Add this help text row above the average row:
```
<TableRow key="help-text">
                  <TableCell colSpan={4}>
                    Average and Total numbers represent the list shown above only.
                  </TableCell>
                </TableRow>
```

BrianBelhumeur:
All set!

PR #2819, opened 2021-03-10
Explorer Home Activity table participant type checkbox updates

BrianBelhumeur:
# Description
Per issue #2298, checkboxes that filter the activity table by type of participant should start off checked by default, and the table should be empty if none of them are checked. This is a change from the previous behavior where boxes were unchecked by default, and only filtered the list if one or more were checked (all checked and all unchecked were the same results).

# Test Plan
* Load page
* Verify that all checkboxes are checked
* Verify that table contains all expected participants
* Uncheck a checkbox and verify that participant type disappears from the list
* Repeat previous step until all checkboxes are unchecked
* Verify that list is empty

# Demo
https://www.loom.com/share/f3ab05d534c04e0196d562a1bb6fc9ce

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200043970364608) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -35,7 +35,6 @@ import {
   add,
   div,
   fromInteger,
-  fromString,
   type Grain,
 } from "../../../core/ledger/grain";
 import ExplorerTimeline from "./ExplorerTimeline";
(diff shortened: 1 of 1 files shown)


blueridger (approved):
Design review still required.

KuraFire (approved):
Lgtm! 👍🏼

PR #2820, opened 2021-03-10
Bugfix: remove duplicate accounts from ledger.accounts

blueridger:
# Description
https://github.com/sourcecred/sourcecred/pull/2803 introduced a bug because it caused there to be multiple entries per account in the `_accounts` map in the ledger, causing the `accounts()` getter to be return duplicates. This PR is a super simple solution.

# Alternatives considered
We could also add a new array/set property that stores all of accounts without duplicates in order to make the getter faster, but that would result in more code complexity and longer ledger construction time that would probably negate the latency savings in most cases.

# Test Plan
- Go to the `explorer-home` page and confirm that there are no longer duplicates.
- Confirm that new unit test fails before ledger change and succeeds after.

### DIFF
--- a/src/core/ledger/ledger.js
+++ b/src/core/ledger/ledger.js
@@ -159,7 +159,7 @@ export class Ledger {
    * Return all the Accounts in the ledger.
    */
   accounts(): $ReadOnlyArray<Account> {
-    return Array.from(this._accounts.values());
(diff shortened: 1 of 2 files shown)


topocount (approved):
Thanks!

PR #2821, opened 2021-03-10
Explorer Home: Fix display of Average Grain in Activity table

BrianBelhumeur:
# Description
In fixing other issues, I noticed that the calculation of Average Grain in the Activity Table on Explorer Home was incorrect (always 0). The TLDR is that Grain is saved as a very large, high-precision string and when dividing the number of grain / the number of users, the number of users was being converted into a very large integer as well, which made the precision drop by 18 decimal places and therefore appear as 0. (E.g. what should have been, say 450 was .000000000000000450, which displayed as 0).

The fix is to simply allow the `div()` function to accept a number as the divisor since that's converted directly into a BigInt. In fact, I'd argue that the divisor should **only** be a number, and not a Grain, because the conceptual idea is that you're dividing a quantity of grain into a certain number of parts, and if the number of parts is also padded by 18 decimal places, you'll lose those places in the division such that the quotient becomes stripped of its 18 decimal places of precision (as happened above).

One drawback of this fix is that not **any** number is acceptable, as BigInt only accepts integers. Any non-integer number passed in would throw an error. I looked into Flow typing to handle this, but it does not seem to be able to specify "integers only".

An alternative fix would be to convert the number of users to a string, then create a grain value from that string using `fromString`, which does not add the 18-places of precision like `fromInteger` does. In fact, the test for the `div()` function does just that. However, `fromString(String(someNumber))` is messy and very unintuitive when a `fromInteger()` function exists, even though that's not the right choice in this case either.

Also, all of the above arguments apply equally to the `mul()` function as well, with the exception that multiplication is associative so it technically shouldn't matter which factor is the Grain and which is the integer, but it would make sense to keep the same signature as `div()` for consistency.

Feedback and improved solutions are welcome.

# Test Plan
* Load page
* Verify Average Grain is indeed the average for the participants shown
* Filter list down to only a few entries
* Verify Average Grain is updated to reflect the average for only the participants shown

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200046196345722) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/core/credGrainView.js
+++ b/src/core/credGrainView.js
@@ -14,7 +14,7 @@ import {
 } from "./credrank/credGraph";
 import {type Identity, type IdentityId} from "./identity";
 import * as G from "./ledger/grain";
-import {type Grain} from "./ledger/grain";
+import {type Grain, add, ZERO} from "./ledger/grain";
 import {Ledger, type Account} from "./ledger/ledger";
 import {type TimestampMs} from "../util/timestamp";
 import findLastIndex from "lodash.findlastindex";
@@ -56,11 +56,15 @@ export class CredGrainView {
   _ledger: Ledger;
   _participants: $ReadOnlyArray<ParticipantCredGrain>;
   _intervals: IntervalSequence;
+  _credTotals: Array<number>;
+  _grainTotals: Array<Grain>;
 
   constructor(credGraph: CredGraph, ledger: Ledger) {
     this._credGraph = credGraph;
     this._ledger = ledger;
     this._intervals = deepFreeze(credGraph.intervals());
+    this._credTotals = [];
+    this._grainTotals = [];
 
     const graphParticipants = new Map<IdentityId, GraphParticipant>();
     for (const participant of credGraph.participants()) {
@@ -75,14 +79,25 @@ export class CredGrainView {
             `The graph is missing account [${account.identity.name}: ${account.identity.id}] that exists in the ledger.`
           );
 
+        const grainEarnedPerInterval = this._calculateGrainEarnedPerInterval(
+          account
+        );
+        for (let i = 0; i < this._intervals.length; i++) {
+          this._credTotals[i] =
+            graphParticipant.credPerInterval[i] + this._credTotals[i] || 0;
+
+          this._grainTotals[i] = add(
+            grainEarnedPerInterval[i],
+            this._grainTotals[i] || ZERO
+          );
+        }
+
         return {
           identity: account.identity,
           cred: graphParticipant.cred,
           credPerInterval: graphParticipant.credPerInterval,
           grainEarned: account.paid,
-          grainEarnedPerInterval: this._calculateGrainEarnedPerInterval(
-            account
-          ),
+          grainEarnedPerInterval,
         };
       })
     );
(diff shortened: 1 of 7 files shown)


blueridger:
You should use [multiplyFloat](https://github.com/sourcecred/sourcecred/blob/master/src/core/ledger/grain.js#L192) instead. `mul` and `div` are intended to be robust and precise, whereas `multiplyFloat` is intended to be more mess-tolerant. The idea of dividing Grain by another Grain does have semantic value, when you want to find the ratio between two grain values.

BrianBelhumeur:
> You should use [multiplyFloat](https://github.com/sourcecred/sourcecred/blob/master/src/core/ledger/grain.js#L192) instead. The grain.js module probably doesn't need any edits for now. `mul` and `div` are intended to be robust and precise, whereas `multiplyFloat` is intended to be more mess-tolerant. The idea of dividing Grain by another Grain does have semantic value, when you want to find the ratio between two grain values.

@blueridger I saw `multiplyFloat` and as I understood it, the intention behind it was to be able to handle multiplication with floating-point numbers, which `BigInt` isn't equipped to do. Thus the conversation of the `BigInt` to a `Number` and the associated lossiness. In this case, I'm dividing by an integer and thus shouldn't have to incur any conversion loss because the divisor can become a BigInt instead of the dividend becoming a Number. Perhaps this is a purely academic exercise in this case because it's calculating an average, which by nature is imprecise, but this question seems like it could have implications in the future. Looking again at `multiplyFloat` I see guard code to ensure that the number provided is indeed an integer which I think would be a good addition if this edit is kept, but I'm also happy to punt on this and simply create a `divideFloat = (dividend, divisor) => mulitiplyFloat(dividend, 1/divisor)` wrapper function and move on.

Good point on dividing Grain/Grain in order to find a ratio, I didn't think of that.

KuraFire (approved):
From a UI Review position this looks good, but consider this UI approval dependent on @blueridger's dev review feedback.

decentralion (commented):
You're right that the `mul` and `div` methods don't really make any sense as written -- indeed calling `mul(Grain.ONE, Grain.ONE)` would return 10^18 Grain, which isn't sensical.

However, simply allowing regular numbers to be put in, without actually fixing the API contract or adding any unit tests, isn't a good solution. Here is what I suggest:

1. Remove the existing `mul` and `div` methods since they make no sense. Update the ExplorerHome to just display a `-` for the average grain, pending a fix.

2. Add a well-thought-out API for doing Grain division that addresses our actual needs. Should have a docstring that explains why/how to use it. Write some unit tests to verify the precision loss is acceptable. See https://github.com/sourcecred/sourcecred/pull/1959 for an example

3. Update ExplorerHome to use the new API

If you do this, I'd be particularly interested in reviewing step 2.


Edit: Actually, reading the review discussion thread, I kind of agree with @blueridger that we should just use multiplyFloat. I don't feel the precision really matters much here, since it's just a UI value.

BrianBelhumeur:
@decentralion I went with a new function called `divideFloat` which is just a wrapper for `multiplyFloat(grain, 1/num)`. That will get this issue out the door without scope creep. I'll also see to it that your comment is converted into a proper issue that can be slated as a standalone piece of work.

blueridger (approved):

BrianBelhumeur:
> @decentralion I went with a new function called `divideFloat` which is just a wrapper for `multiplyFloat(grain, 1/num)`. That will get this issue out the door without scope creep. I'll also see to it that your comment is converted into a proper issue that can be slated as a standalone piece of work.

Drew up issue #2843 for this

PR #2816, opened 2021-03-09
Implement ZipStorage

topocount:
ZipStorage composes with other DataStorage implementations. It compresses
`value`s before passing them into the underlying `baseStorage` class.

The longer reaching purpose of this class is to allow SourceCred
instances to operate further under platform file size limits. If instances
continue to expand and bump up against these limits, solutions such as a
ShardedStorage composable class can be implemented as well.

merge next: #2858 

test plan: unit tests have been included

builds towards #2768

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038204718973) by [Unito](https://www.unito.io/learn-more)

PR #2844, opened 2021-03-16
Implement and utilize NetworkStorage

topocount:
This class is a DataStorage wrapper for basic fetch GET calls.
It throws with the HTTP error when the response is not "OK". This allows
for I/O helpers to handle different error codes accordingly.

test plan:
unit test are provided. the ui/load.js file has also been updated and
works with the snapshotted instance and a local SourceCred instance.

builds towards: #2768
merge after: #2840
merge next: #2816

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200067172804864) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/src/core/storage/network.test.js
+++ b/src/core/storage/network.test.js
@@ -1,0 +1,68 @@
+// @flow
+
+import {NetworkStorage} from "./network";
+
+jest.mock("cross-fetch", () => ({
+  // needed to utilize fetch as a default export.
+  __esModule: true,
+  default: (path) => {
+    switch (path) {
+      case "base/validPath":
+        return Promise.resolve({
+          arrayBuffer: () => new Uint8Array([1, 2]).buffer,
+          ok: true,
+          status: 200,
+          statusText: "OK",
+        });
+      case "base/serverError":
+        return Promise.resolve({
+          arrayBuffer: () => new ArrayBuffer(0),
+          ok: false,
+          status: 500,
+          statusText: "INTERNAL ERROR",
+        });
+      default:
+        return Promise.resolve({
+          arrayBuffer: () => new ArrayBuffer(0),
+          ok: false,
+          status: 404,
+          statusText: "NOT FOUND",
+        });
+    }
+  },
+}));
+
+describe("core/storage/network", () => {
+  describe("NetworkStorage", () => {
+    const value = new Uint8Array([1, 2]);
+    it("returns a payload", async () => {
+      expect.hasAssertions();
+
+      const storage = new NetworkStorage("");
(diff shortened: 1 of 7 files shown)


src/ui/load.js
```diff
@@ -26,44 +30,45 @@ export async function load(): Promise<LoadResult> {
...
+      currencyParser,
+      defaultCurrencyConfig
+    ),
+    loadJsonWithDefault(
+      networkStorage,
+      // TODO: Write Parser for CredGraph
+      "output/credGraph.json",
+      Combo.fmap(((Combo.raw: any): Combo.Parser<CredGraphJSON>), (graphJson) =>
+        CredGraph.fromJSON(graphJson)
+      ),
```
blueridger, line 53: why not use the parser from the CredGraph module? https://github.com/sourcecred/sourcecred/blob/master/src/core/credrank/credGraph.js#L128-L136

src/ui/load.js
```diff
@@ -26,44 +30,45 @@ export async function load(): Promise<LoadResult> {
...
   const ledgerManager = new LedgerManager({
     storage: diskStorage,
   });
 
   const queries = [
-    fetch("sourcecred.json"),
-    fetch("static/server-info.json"),
-    fetch("config/currencyDetails.json"),
-    fetch("output/credGraph.json"),
+    loadJson(networkStorage, "sourcecred.json", (Combo.raw: any)),
```
blueridger, line 39: https://github.com/sourcecred/sourcecred/blob/master/src/api/instanceConfig.js#L37

src/core/storage/network.test.js
```diff
@@ -0,0 +1,71 @@
...
+      expect.hasAssertions();
+
+      const storage = getNetworkStorage();
+      const result = await storage.get("base/validPath");
+      await expect(result).toEqual(value);
+    });
+    it("respects URL bases", async () => {
+      expect.hasAssertions();
+      const storage = getNetworkStorage("base");
+      const thunk = async () => storage.get("base/validPath");
```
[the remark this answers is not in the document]
topocount: This extra assertion is just meant to illustrate that a path that might be valid in some contexts is not valid when `base` is set at construction. It's more of an assertion of specific functionality.

We would expect an invalid path to error in any context, while `validPath` is only accessible under specific contexts

src/ui/load.js
topocount: I had to split out the raw parser from the bundledplugins parser, since the bundledplugins parser actually reaches into node fs modules.

blueridger (approved):

PR #2858, opened 2021-03-17
Utilize ZipStorage

topocount:
## Description
For now, this class will be used to compress weighted graphs and cred
graphs. these artifacts aren't especially readable on disk and also take
up a 10s of MBs. This will allow larger instances to continue utilizing
Github for the foreseeable future. If further optimization is needed, a
higher-order ShardedStorage class can be implemented and used.

merge after #2816 
builds towards #2768 

## Test Plan
`yarn build`

in a local instance:

1. `scDev go`
1. `scDev grain`
1. `scDev serve`

open http://localhost:6006 and verify that the frontends load as
expected.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200077407311401) by [Unito](https://www.unito.io/learn-more)


topocount:
I'm really not a fan the fact that git doesn't even attempt to diff gzip files

blueridger (commented):
Doesn't this need wrapped too? https://github.com/sourcecred/sourcecred/blob/master/src/ui/load.js#L57

You should be sure to clean your workspace as a part of the test steps. I'm guessing that's why this didn't get caught.

blueridger (approved):

PR #2861, opened 2021-03-18
Bugfix: Reinstantiate CredValue for mutation

topocount:
Description
-
In explorer timeline component, we attempt to mutate a prop.
Reinstantiation of the value within the component allows the mutation to
occur, and the render succeeds.

The issue occurs specifically when The credValue array is of length 1,
    and we attempt to add a second entry in the array, presumably to
    form a line.

Test Plan
-
Find some data that that has CredView array lengths(s) of 1. I used the
user-provided data here:
https://discord.com/channels/453243919774253079/718263631158050896/822135748014112798

paste the output and data folders into your instance of choice, and
serve it. Ensure explorer home loads.

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerTimeline.js
+++ b/src/ui/components/ExplorerHome/ExplorerTimeline.js
@@ -29,7 +29,6 @@ const ExplorerTimeline = (props: ExplorerTimelineProps): ReactNode => {
   const width = props.width || 300;
   const height = props.height || 25;
   const viewBox = `0 0 ${width} ${height}`;
-
   let grainValues;
   if (grainExists) {
     const grain = props.timelines.grain || [];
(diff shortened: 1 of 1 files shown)


src/ui/components/ExplorerHome/ExplorerTimeline.js
```diff
@@ -29,7 +29,6 @@ const ExplorerTimeline = (props: ExplorerTimelineProps): ReactNode => {
   const width = props.width || 300;
   const height = props.height || 25;
   const viewBox = `0 0 ${width} ${height}`;
-
   let grainValues;
   if (grainExists) {
     const grain = props.timelines.grain || [];
```
blueridger, line 34: this too?

topocount: `grainValues` is actually what potentially would need this re-instantiation, but it doesn't since it is derived from a map operation on the `grain` prop.

blueridger (approved):

PR #2865, opened 2021-03-20
Cease Support for Node v10

topocount:
Node V10 is end-of-life in just over a month. Rather than fixing anymore
bugs related to version compatibility, we've decided to enforce this
end of support in package.json

test plan:
0. install nvm or use some other node version manager
1. `nvm use 10 && yarn` // should error on the yarn installation
2. `nvm use 12 && yarn` // should succeed

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,9 @@
     "prepublishOnly": "yarn build",
     "docs": "node ./scripts/build-docs.js"
   },
+  "engines": {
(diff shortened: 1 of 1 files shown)


wchargin (approved):
OK with me; looks like `better-sqlite3` has prebuilt binaries for both v12 and v14.

blueridger (approved):

PR #2894, opened 2021-03-30
Docs: Update plugin design doc to include instance-plugin Contract

topocount:
- Add sections that provide context
- Specify the new instance-plugin contract
- Add example process for Maker's Discourse plugin

### DIFF
--- a/design-docs/2021/q1/plugin-redesign.md
+++ b/design-docs/2021/q1/plugin-redesign.md
@@ -120,18 +120,17 @@ not modifiable from within an instance. Within an instance, plugins are
 specified as active via the root [sourcecred.json] file by the instance
 maintainer.
 
-[bundledplugins.js]: https://github.com/sourcecred/sourcecred/blob/3e77f40a2335c597d18f192e36e756f6c452fc7b/src/api/bundledPlugins.js
 [sourcecred.json]: https://github.com/sourcecred/cred/blob/887584b602eee1ad8e68d88b2733070d1d6732fa/sourcecred.json
 
 ### Data Model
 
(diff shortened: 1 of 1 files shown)


blueridger:
white-list -> allow-list

Don't see why the Instance should define a plugin name for the plugin (how will that name be used?). Plugin declarations already provide a name property. Probably better for an instance to just create an array of plugins.

It doesn't acknowledge that the new plugin structure will make instance setup much more complicated, because creating an instance will mean creating all the data provider services the instance will use. I think we should consider continuing to support "lite" plugins like our currently plugins that don't require additional infrastructure set up. And I think this should be considered in the doc.

blueridger:
Also, how exactly does sourcecred cli or an observable query for the plugins or their declarations with this new contract? It'd be nice to understand. Is sourcecred core going to somehow import the config file from its dynamic location? Is the CLI entrypoint going to become the instance's responsibility?

topocount:
> It doesn't acknowledge that the new plugin structure will make instance setup much more complicated, because creating an instance will mean creating all the data provider services the instance will use. I think we should consider continuing to support "lite" plugins like our currently plugins that don't require additional infrastructure set up. And I think this should be considered in the doc. Or, could we integrate the new cron job with the existing instance's cron job, so that an instance and all its dataproviders can exist as a monorepo/monoservice?

The below assumes that all logic and mirrored data exists in Github.

A monorepo definitely helps with managing shared dependencies, but these repos will be serving as a backend, so actually having different github repos is actually probably easier, since it makes things more modular. It makes sense to have all the actual plugin logic exist in a monorepo with the instance, but the actual data should exist in a separate repo.

blueridger (approved):

PR #2896, opened 2021-03-31
Write pluginDeclarations to disk during graph CLI

blueridger:
# Description
This writes plugin declarations to disk during the graph CLI. This will be needed by the weight configuration page. We can't load the plugins themselves in browser because those modules depend on `fs` so we need to write the declarations so they can be loaded in browser.

# Test Plan
1. scdev graph
2. verify the pluginDeclarations.json is generated


topocount (commented):
A few questions:
If these are needed for weights config, can we just add endpoints to the local express service that support these? Or is it necessary to make the weights UI available in the statically-served site too? I'd like to avoid outputting what is effectively duplicated data when we can just serve it via expressJS.

How was this metadata loaded into the frontend before? Was it included in an old timeline cred artifact?

blueridger:
They were included in the CredResult.json before. Since plugins are currently tightly coupled with disk writing, it's necessary to have plugin declarations accessible for browser use outside of the BundledPlugins codepath. Observables may find use for this json as well. The long term solution is programatic access to the declarations of plugins without having to go through the disk-reliant Plugin interface implementations.

blueridger:
closing in favor of a lighter weight solution: https://github.com/sourcecred/sourcecred/pull/2898

PR #2898, opened 2021-04-02
Add MVP weights configuration page

blueridger:
# Description
When we launched 0.8, we removed the WeightConfigSection, even though it's a pretty important feature. This PR adds an MVP back in. Differences from the 0.7 version:
- It does not yet offer any cred score simulation functionality.
- It has it's own page instead of being a modal on the explorer page.
- It does not include the alpha/beta/gamma weight configuration.

Screen Shot 2021-04-02 at 5 34 49 PM

# Design choices
I didn't fret too much about the cleanliness in AdminApp.js, since it will change soon as we create a parent Simulator component.

The `bundledDeclarations` module is just a browser-friendly simplification of the `bundledPlugins` module.
# Test Plan
simple case:
1. Opened the new page, changed weights, downloaded the file and visually skimmed.

with upload:
1. Uploaded the weights.json from our prod instance
2. changed 1 weight
3. ran a formatter on the old file and committed changes
4. downloaded the new file, replaced the old file with it, and ran a formatter on it.
5. `git diff`
```
thena@thenas-mbp cred % git diff
diff --git a/config/weights.json b/config/weights.json
index 988971c..c14e155 100644
--- a/config/weights.json
+++ b/config/weights.json
@@ -3235,7 +3235,7 @@
       "N\u0000sourcecred\u0000discourse\u0000like\u0000https://discourse.sourcecred.io\u0000youngkidwarrior\u00003069\u0000": 0.1,
       "N\u0000sourcecred\u0000discourse\u0000like\u0000https://discourse.sourcecred.io\u0000youngkidwarrior\u00003071\u0000": 0.1,
       "N\u0000sourcecred\u0000discourse\u0000post\u0000": 0,
-      "N\u0000sourcecred\u0000discourse\u0000topic\u0000": 0,
+      "N\u0000sourcecred\u0000discourse\u0000topic\u0000": 1,
       "N\u0000sourcecred\u0000discourse\u0000user\u0000": 0,
       "N\u0000sourcecred\u0000github\u0000COMMENT\u0000": 0,
       "N\u0000sourcecred\u0000github\u0000COMMIT\u0000": 0,
```

### DIFF
--- a/src/api/bundledDeclarations.js
+++ b/src/api/bundledDeclarations.js
@@ -1,0 +1,35 @@
+// @flow
+
+import type {PluginDeclaration} from "../analysis/pluginDeclaration";
+import {declaration as githubDeclaration} from "../plugins/github/declaration";
+import {declaration as discourseDeclaration} from "../plugins/discourse/declaration";
+import {declaration as discordDeclaration} from "../plugins/discord/declaration";
+import {declaration as initiativesDeclaration} from "../plugins/initiatives/declaration";
+import {declaration as ethereumDeclaration} from "../plugins/ethereum/declaration";
+import {type RawInstanceConfig} from "./rawInstanceConfig";
+import {type PluginId} from "./pluginId";
+
+export function bundledDeclarations(): {[pluginId: string]: PluginDeclaration} {
+  return {
+    "sourcecred/github": githubDeclaration,
+    "sourcecred/discourse": discourseDeclaration,
+    "sourcecred/discord": discordDeclaration,
+    "sourcecred/initiatives": initiativesDeclaration,
+    "sourcecred/ethereum": ethereumDeclaration,
+  };
+}
+
+export function upgradeRawInstanceConfig(
+  raw: RawInstanceConfig
+): $ReadOnlyMap<PluginId, PluginDeclaration> {
+  const allBundledPlugins = bundledDeclarations();
+  const bundledPlugins = new Map();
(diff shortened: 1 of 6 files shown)


KuraFire:
@blueridger This is great! Awesome to have a starting point to get going with while I design the additional features. One requested change that should be easy to do right now:

Instead of "Upload/download weights: {two icons}", let's make it cleaner: two buttons, with the following labels:

`Download weights` and `Upload weights`

Two buttons that are far more intuitive and usable, without the need of a description line.

blueridger:
@KuraFire done!

src/ui/components/ExplorerHome/ExplorerSimulator.js
```diff
@@ -0,0 +1,773 @@
+// @flow
```
topocount, line 1: I don't see where this simulation view is utilized yet. Can we defer pushing it until it's actually reviewable?

topocount (approved):
I think we should remove the simulator until I can actually test it in a typical user flow, or you add it as a hidden route. Otherwise it's not clear to me what should be reviewed in that file.

Otherwise I think this is a well-optimized MVP.

src/ui/components/ExplorerHome/ExplorerSimulator.js
blueridger: oops, that's supposed to be deleted!

Issue #2815, opened 2021-03-08
ExplorerHome: fix width on small screens


blueridger:
On my mac, when the menu is open, the explorer home overflows the width of the screen. This should be changed so that the width responsively adjusts and does not overflow.
![](https://user-images.githubusercontent.com/11550396/110390214-7c6dae80-801a-11eb-824a-6f28c19e9ef7.png)

# Tech notes
The changes that need made are in [this directory](https://github.com/sourcecred/sourcecred/blob/master/src/ui/components/ExplorerHome).

Look at the [readme](https://github.com/sourcecred/sourcecred) and particularly the ["using a modified frontend"](https://github.com/sourcecred/sourcecred#using-a-modified-frontend) section to get your environment set up.

# PM Notes
[This is non-blocking for launch.]

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038243155129) by [Unito](https://www.unito.io/learn-more)

blueridger:
From @KuraFire : Top Chart (incl.) X axis line should be exact same page width as the Activity Table, or "100% width of the page container" that holds the contents.

PR #2899, opened 2021-04-02
Responsive Improvements on Explorer Home

magwalk:
[Re. ExplorerHome: fix width on small screens #2815](https://github.com/sourcecred/sourcecred/issues/2815)

Improves the responsiveness of Explorer Home. Previously the content
overflowed below 1200px. This fixes the page so it flows a bit better between 
sizes, though there is still some screen overflow between 600px and 740px when
the menu is open (the table might be resisting being scaled down, sometimes tables
get stuck while scaling down if the contents of the table cells are too large).

I did take a few design liberties with this code change - i.e. changing the table
so it takes up the same width as the top chart. The layout/alignment felt a little confusing
with the table narrower than the page width. 

I do think we want to give some thought to how we'd like this design to work
at lower sizes (i.e. smaller fonts for the headings, flowing the filter input below
the "Last Week's Activity" heading, a different design for the table or
hiding the Cred Timeline on mobile) so it can work well on phones.

Screen capture of what the page looks like on resize: https://www.loom.com/share/f4a46960f95547599e8513e00de81e51

# Test Plan
Open the Explorer Home and resize the browser.

### DIFF
--- a/src/ui/components/ExplorerHome/ExplorerHome.js
+++ b/src/ui/components/ExplorerHome/ExplorerHome.js
@@ -6,7 +6,6 @@ import {
   FormControlLabel,
   FormGroup,
   FormLabel,
-  Hidden,
   Paper,
   Tab,
   Tabs,
@@ -23,6 +22,7 @@ import {
 import {makeStyles} from "@material-ui/core/styles";
 import TableSortLabel from "@material-ui/core/TableSortLabel";
 import grey from "@material-ui/core/colors/grey";
+import useMediaQuery from "@material-ui/core/useMediaQuery";
 import deepFreeze from "deep-freeze";
 import bigInt from "big-integer";
 import {CredGrainView} from "../../../core/credGrainView";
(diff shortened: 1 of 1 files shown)


KuraFire:
@magwalk @blueridger Would it be possible to include one additional change to this PR? Okay if you prefer it as a future separate PR. The design spec is:

- Hide the "Contributions charts" column below the 600px breakpoint.

magwalk:
> @magwalk @blueridger Would it be possible to include one additional change to this PR? Okay if you prefer it as a future separate PR. The design spec is:
> 
> * Hide the "Contributions charts" column below the 600px breakpoint.

I think that would make a ton of sense for this PR. Happy to include.

magwalk:
Updated per @KuraFire request. Here's a new loom: https://www.loom.com/share/96f8dfa6d93f4edfbd9dbcb4b1de69ac

KuraFire:
> Updated per @KuraFire request. Here's a new loom: https://www.loom.com/share/96f8dfa6d93f4edfbd9dbcb4b1de69ac

Oh wow, my estimate of the widths I was looking at was off, can we set that breakpoint at `768px` instead, @magwalk ?

magwalk:
New loom for latest changes: https://www.loom.com/share/b38db55f2cc748aeb1047a6b4be38ea4

BrianBelhumeur (approved):

Issue #2778, opened 2021-02-22
Allow grain UI to show more than 2 digits past the decimal


befitsandpiper:
1hive is starting to move minimum distribution sizes close to the lower end of hat the UI shows bc of the recent HNY price increases. It would be nice if the number of decimals shown was configurable.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038301242225) by [Unito](https://www.unito.io/learn-more)

topocount:
I think this might be an issue with the grain format function, but it is a pretty subjective concern:
https://github.com/sourcecred/sourcecred/blob/aa1a46943d83da3f61bbfae797021976e08d1240/src/core/ledger/grain.js#L146-L147

The question is, what is the expected behavior when the total amount to render is smaller than the decimal total? At what point do we switch to scientific notation? should we introduce some notion of significant figures into this function to ensure we don't have absurdly long number strings showing in our frontend?

ETH has special names for different orders of magnitude of ETH: https://github.com/ethereumjs/ethereumjs-units. Would you want to develop unit names for different orders of magnitude of Honey?

I think a more sustainable solution is [SI unit prefixes](https://en.wikipedia.org/wiki/Metric_prefix). Would that work for the folks at 1Hive? It'd also probably help to have a tooltip that shows the raw decimal as well.

decentralion:
Note: The format function already supports customizing the number of decimals:

https://github.com/sourcecred/sourcecred/blob/aa1a46943d83da3f61bbfae797021976e08d1240/src/core/ledger/grain.js#L103-L107

We could just update the currencyConfig file to admit a custom decimal precision:
https://github.com/sourcecred/sourcecred/blob/aa1a46943d83da3f61bbfae797021976e08d1240/src/api/currencyConfig.js

PR #2912, opened 2021-04-06
Allow grain UI to show more than 2 digits past the decimal

magwalk:
Allows instance maintainers to customize the number of decimals
that are displayed with their currency (fix for ticket #2778) 

# Description
Adds a new property `decimalsToDisplay` to the `currencyDetails.json` config. The default value is 2. 

Sample `currencyDetails.json`:
```
{
  "decimalsToDisplay": 4
}
```

Adding this property results in this change to the UI: 
Screen Shot 2021-04-05 at 6 01 45 PM

# Test Plan
- load the UI without a currencyDetails.json file
- load the UI with a currencyDetails.json file with a custom value for `decimalsToDisplay`: 
```
{
  "decimalsToDisplay": 4
}
```
- load the UI with a currencyDetails.json file that has other values but not `decimalsToDisplay`, like so: 
```
{
  "currencyName": "honey",
  "currencySuffix": "h" 
}
```

### DIFF
--- a/src/api/currencyConfig.js
+++ b/src/api/currencyConfig.js
@@ -4,11 +4,12 @@ import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
 
 /**
- * Shape of concurrencyDetails.json on disk
+ * Shape of currencyDetails.json on disk
  */
 type SerializedCurrencyDetails = {|
   +currencyName?: string,
   +currencySuffix?: string,
+  +decimalsToDisplay?: number,
 |};
 
 /**
@@ -17,10 +18,12 @@ type SerializedCurrencyDetails = {|
 export type CurrencyDetails = {|
   +name: string,
   +suffix: string,
+  +decimals: number,
 |};
 
 export const DEFAULT_NAME = "Grain";
 export const DEFAULT_SUFFIX = "g";
+export const DEFAULT_DECIMALS = 2;
 
 /**
  * Utilized by combo.fmap to enforce default currency values
@@ -32,6 +35,7 @@ function upgrade(c: SerializedCurrencyDetails): CurrencyDetails {
   return {
     name: NullUtil.orElse(c.currencyName, DEFAULT_NAME),
     suffix: NullUtil.orElse(c.currencySuffix, DEFAULT_SUFFIX),
+    decimals: NullUtil.orElse(c.decimalsToDisplay, DEFAULT_DECIMALS),
   };
 }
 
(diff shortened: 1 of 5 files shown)


src/api/currencyConfig.js
```diff
@@ -4,11 +4,12 @@ import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
 
 /**
- * Shape of concurrencyDetails.json on disk
+ * Shape of currencyDetails.json on disk
```
magwalk, line 7: typo fix

magwalk:
I think I got all the places where currency is displayed and should show decimal points, let me know if there are others I didn't catch.

blueridger:
We should probably apply this to the stats circle on the explorer home as well. And also the GrainReceiptTable component of the LedgerViewer.

KuraFire:
@magwalk YAY! Thank you for this much-wanted improvement :)

**Design thought:**
Correct me if I'm wrong, but this seems to create a (brand?) new config file, `currencyDetails.json`? Can we make configuration values go in a central configuration location? 

(I know Dandelion had hesitations about that, but I'm totally unclear on why we would want there to be 5 places in your setup to configure one single instance to your needs.)

**For my UI review:**
Does this come with an upper limit? We have no need for letting people enter "144" and break the UI. I don't know if we need to support more than 8 digits?

blueridger:
@KuraFire currencyDetails.json is an existing file, this just adds a new attribute to that file.

We only support up to 18 digits. Adding a conditional throw in the upgrade function could be nice to enforce the value into 0-18.

magwalk:
So this brings up an interesting breaking point for the stats circles UI. With the current text size on the stats circles, there is a limit of 9 digits that can be displayed before the text breaks out of the circles. A question I have for @KuraFire - how should the UI respond when there are more than 9 digits? 

Here's how it breaks (this example shows 10 digits): 
Screen Shot 2021-04-07 at 8 07 24 AM

magwalk:
> So this brings up an interesting breaking point for the stats circles UI. With the current text size on the stats circles, there is a limit of 9 digits that can be displayed before the text breaks out of the circles. A question I have for @KuraFire - how should the UI respond when there are more than 9 digits?
> 

Talked it over with a couple devs and we are going to push off this work because this seems to be an edge case that isn't likely to be broken by any of our users. I will create a bug issue to track the work to fix this.

BrianBelhumeur (approved):

src/api/currencyConfig.js
```diff
@@ -32,17 +35,26 @@ function upgrade(c: SerializedCurrencyDetails): CurrencyDetails {
   return {
```
blueridger, line 35:
```suggestion
  if (c.decimalsToDisplay > 18 || c.decimalsToDisplay < 0) throw new Error("currencyConfig: decimalsToDisplay must be between 0-18")
  return {
```

magwalk:
New loom demonstrating the new stats circles values with compact formatting https://www.loom.com/share/26c674259ab94adf9970bb19220221e5

KuraFire (approved):
Awesome work! ❤️

BrianBelhumeur (approved):

PR #2915, opened 2021-04-07
Discord: do not throw for unknown channel type

blueridger:
# Description
The discord plugin has a fragility that throws on unknown channel types. This is a problem because when new channel types are added by discord, such as their new Stage channel type, our plugin breaks. This is currently happening to 1Hive.

We don't really care about channel type. We only care whether it is a text channel or not, and we ignore other channels. So there's really no reason to be picky about non-text channels. This PR creates a default "UKNOWN" channel type that gracefully handles unknown channel types.

# Test Plan
@befitsandpiper is going to test this change on the 1Hive instance. Results TBA.

### DIFF
--- a/src/plugins/discord/models.js
+++ b/src/plugins/discord/models.js
@@ -20,7 +20,8 @@ export type ChannelType =
   | "GROUP_DM"
   | "GUILD_CATEGORY"
   | "GUILD_NEWS"
-  | "GUILD_STORE";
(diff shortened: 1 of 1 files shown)


topocount (approved):

blueridger:
@befitsandpiper confirmed it works.

topocount:
Replicated and tested this bug and tested the patch on our test server

PR #2916, opened 2021-04-07
Re-add /explorer as redirect to /explorer-home

blueridger:
# Description
We broke a URL (/explorer) for no real reason as a part of 0.8.0. Since root redirects to the explorer url, users had the explorer url bookmarked/hyperlinked, and those broke when we changed the url to be /explorer-home. This makes /explorer the primary URL again, and creates a redirect from /explorer-home to /explorer. In case anyone hasn't updated their hyperlinks/bookmarks yet, this will avoid some minor confusion and also be a reminder not to break stuff unnecessarily.

# Test Plan
1. scdev serve
1. verify that /explorer loads by default
2. forceful browse to /explorer-home and verify redirect to /explorer

### DIFF
--- a/src/ui/components/AdminApp.js
+++ b/src/ui/components/AdminApp.js
@@ -104,11 +104,14 @@ const customRoutes = (
     weights: emptyWeights(),
   });
   const routes = [
-    <Route key="explorer-home" exact path="/explorer-home">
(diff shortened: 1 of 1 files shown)


topocount (approved):

saintmedusa (approved):
LGTM

PR #2935, opened 2021-04-09
Grain: add --force flag to redo last distribution

blueridger:
# Description
I've been running into an annoying reoccurring manual process, so I automated it! As a dev or instance maintainer, sometimes I want to simulate a grain distribution but the past week already has a distribution. So I would manually delete the distribution from the ledger.json to redo it for analytics. With this PR, I don't need to manually edit the JSON. Instead, I simply use the `-f` / `--force` flag to allow the grain command to append another distribution even if there is already an existing distribution for the last interval!

# Test Plan
1. checkout sourcecred/cred gh-pages branch
2. hand delete last distribution from ledger.json
3. `scdev grain -s` and verify 1 expected distribution simulated
5. `scdev grain -f` and verify 1 expected distribution written to ledger
6. `scdev grain -s` and verify 0 expected distributions
7. `scdev grain -f` and verify 1 new expected distribution written to ledger, and verify console output shows the same timestamp as previous distribution
8. `scdev grain --simulation --force` and verify 1 expected distribution simulated

### DIFF
--- a/src/cli/grain.js
+++ b/src/cli/grain.js
@@ -14,24 +14,34 @@ function die(std, message) {
   return 1;
 }
 
-/**
- * The grain command is soon to be deprecated, as part of a transition
- * to @blueridger's `CredGrainView`.  This original grain command uses
- * `CredView`, which will be deprecated.
- *
- * grain2 forks this command and eliminates the dependence on `CredView`
- */
 const grainCommand: Command = async (args, std) => {
   let simulation = false;
-  if (args.length === 1 && (args[0] === "--simulation" || args[0] === "-s")) {
-    simulation = true;
-  } else if (args.length !== 0) {
-    return die(std, "usage: sourcecred grain [--simulation]");
+  let allowLastDistributionOverwrite = false;
+  const processedArgs = args.filter((arg) => {
+    switch (arg) {
+      case "-f":
+      case "--force":
+        allowLastDistributionOverwrite = true;
+        return false;
+      case "-s":
+      case "--simulation":
+        simulation = true;
+        return false;
+      default:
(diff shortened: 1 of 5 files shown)


topocount (commented):
I'm not sure how I feel about this existing in userland, as users are supposed to treat the ledger as append-only. Would we leave it undocumented?

blueridger:
I get that. However. I'm sure I'm not the only one hand-editing. I bet our users are hand editing, and so this is an improvement over that. One thing I could do is require -f to be used with -s in the CLI. Ultimately though, it seems a small sacrifice of principals for a large improvement to how our product is actually getting used.

Another thought is that I could make the -f flag enable multiple distributions on a single interval, which is currently not allowed but I can't think of a reason why interval:distribution must be a strict 1:1. Yeah that sounds way better. Imma try that.

src/cli/grain.js
```diff
@@ -71,14 +81,17 @@ const grainCommand: Command = async (args, std) => {
...
-      usage: sourcecred grain [--simulation || -s]
+      usage: sourcecred grain [--simulation | -s] [--force | -f]
 
       Distribute Grain (or whatever currency this Cred instance is tracking)
       for Cred intervals in which Grain was not already distributed.
 
       When the '--simulation' (-s) flag is provided, no grain will actually be distributed,
       allowing for testing the output of various configurations.
 
+      When the '--force' (-f) flas is provided, it will overwrite the last distribution
```
topocount, line 92:
```suggestion
      When the '--force' (-f) flag is provided, it will overwrite the last distribution
```

topocount (approved):

PR #2950, opened 2021-04-15
Ledger: lastDistributionTimestamp edge case refactor

hozzjss:
- now ledger would return null as the default value

# Description
If there has never been any distribution, it would be clearer for the lastDistributionTimestamp to be null, so that handling based on that might be easier than the currenct `lastDistributionTimestamp === -Infinity`

Solves https://github.com/sourcecred/sourcecred/issues/2744
The change is a breaking change in that all the code before that used the comparison to -Infinity would be changed to a null check, I might suggest we use 0 instead as it's the only falsy number in JS and it wouldn't change the interface as much but that's just an idea, null is much much better if it's okay to change that interface

# Test Plan
nothing special just run `yarn test`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "javascript.validate.enable": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "javascript.validate.enable": false
-}
--- a/src/core/ledger/applyDistributions.js
+++ b/src/core/ledger/applyDistributions.js
@@ -47,7 +47,9 @@ export function applyDistributions(
   const credIntervals = credGraph.intervals();
   const distributionIntervals = _chooseDistributionIntervals(
     credIntervals,
-    ledger.lastDistributionTimestamp(),
+    // Distributions after -Infinity are all the current
+    // distributions
+    ledger.lastDistributionTimestamp() || -Infinity,
     currentTimestamp,
     policy.maxSimultaneousDistributions
   );
--- a/src/core/ledger/applyDistributions.js
+++ b/src/core/ledger/applyDistributions.js
@@ -1,6 +1,7 @@
 // @flow
 
 import {type TimestampMs} from "../../util/timestamp";
+import * as NullUtil from "../../util/null";
 import {type IntervalSequence, intervalSequence} from "../interval";
 import {Ledger} from "./ledger";
 import {type AllocationPolicy} from "./policies";
@@ -49,7 +50,7 @@ export function applyDistributions(
     credIntervals,
     // Distributions after -Infinity are all the current
     // distributions
-    ledger.lastDistributionTimestamp() || -Infinity,
+    NullUtil.orElse(ledger.lastDistributionTimestamp(), -Infinity),
     currentTimestamp,
     policy.maxSimultaneousDistributions
   );
--- a/src/core/ledger/applyDistributions.js
+++ b/src/core/ledger/applyDistributions.js
@@ -48,8 +48,6 @@ export function applyDistributions(
   const credIntervals = credGraph.intervals();
   const distributionIntervals = _chooseDistributionIntervals(
     credIntervals,
-    // Distributions after -Infinity are all the current
-    // distributions
     NullUtil.orElse(ledger.lastDistributionTimestamp(), -Infinity),
     currentTimestamp,
     policy.maxSimultaneousDistributions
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "javascript.validate.enable": false
+}
```
decentralion, line 3: should be in your personal (global) `.gitignore` if you use vscode

src/core/ledger/applyDistributions.js
```diff
@@ -47,7 +47,9 @@ export function applyDistributions(
   const credIntervals = credGraph.intervals();
   const distributionIntervals = _chooseDistributionIntervals(
     credIntervals,
-    ledger.lastDistributionTimestamp(),
+    // Distributions after -Infinity are all the current
+    // distributions
+    ledger.lastDistributionTimestamp() || -Infinity,
```
decentralion, line 52: nit: this is _technically_ incorrect, since `0` is a valid timestamp corresponding to 1970, and this code would set `-Infinity` for timestamp 0. in practice, we won't have to worry about distributions from 1970, but it's good to be in a habit of not assuming that `x == null` implies `x === null` when `x` is a number. similarly when `x` is a string, since `""` is a valid string that would evaluate as false in this case.

this is what we have `NullUtil.orElse` for, recommend using that!

decentralion (commented):
A bit of drive by commentary, hope you don't mind. :)

Code looks fine overall, will let others approve.

.vscode/settings.json
hozzjss: That makes sense, I just pushed a commit that should resolve that

src/core/ledger/applyDistributions.js
hozzjss: That makes sense, I actually think it's really nice if there were like time travelers that used sourcecred, unironically that is lol, pushed a commit that should resolve that as well

hozzjss:
Thanks @decentralion for the review really appreciate your input

src/core/ledger/applyDistributions.js
```diff
@@ -47,7 +48,9 @@ export function applyDistributions(
   const credIntervals = credGraph.intervals();
   const distributionIntervals = _chooseDistributionIntervals(
     credIntervals,
-    ledger.lastDistributionTimestamp(),
+    // Distributions after -Infinity are all the current
+    // distributions
```
blueridger, line 52: We tend to be conservative about non-documentation, inline comments. Often they add more confusion than clarity, and not leaning on comments forces us to write code that is readable and understandable on its own. I'd remove this comment.

src/core/ledger/ledger.test.js
```diff
@@ -1592,11 +1592,11 @@ describe("core/ledger/ledger", () => {
         failsWithoutMutation(ledger, thunk, "invalid timestamp");
       }
     });
-    it("lastDistributionTimestamp returns -Infinity if there have not been any distributions", () => {
+    it("lastDistributionTimestamp returns null if there have not been any distributions", () => {
       // consider changing this behavior; see
       // https://github.com/sourcecred/sourcecred/issues/2744
```
blueridger, line 1597: This comment should be removed.

blueridger (commented):
Some minor feedback.

Also it'd be nice to have a manual test included that tests against a fresh instance. Good learning opportunity too. Steps:
1. Make sure you've set up the ["using a modified backend" section](https://github.com/sourcecred/sourcecred#using-a-modified-backend)
2. run `yarn build`
1. Checkout https://github.com/sourcecred/template-instance
2. Initialize the configs
   1. change https://github.com/sourcecred/template-instance/blob/master/config/grain.json#L2 to be non-zero
   2. change https://github.com/sourcecred/template-instance/blob/master/sourcecred.json#L2 to be ["sourcecred/discourse"]
3. from your local template-instance clone, run `scdev go`
4. run `scdev serve` go to the url it outputs, and navigate to the grain accounts page to make sure it works and that there is no "last distribution" message at the bottom right
5. ctrl-c and run `scdev grain` and verify that the output shows grain distribution
6. repeat step 6, but verify that there IS a "last distribution" message at the bottom right

src/core/ledger/applyDistributions.js
hozzjss: That makes sense, I'm also on the side of preferring readable code than overly hand-holdy comment lol, I applied these changes

src/core/ledger/ledger.test.js
hozzjss: Done and done!

hozzjss:
Took me some time to wrap my head around setting everything up, but completed my manual test and it passed yaaaayyyy!!

AL0YSI0US (approved):
I saw this and it looks good to me, Thank you @hozzjss for talking me through your approach to this bug and the codebase!

blueridger (approved):

Issue #2949, opened 2021-04-15
Cred data for analysis in output file


s-ben:
**Problem**
In TimelineCred, basic graph data was output to `credResult.json`. I was using data in this file to do analysis in [Python](https://github.com/s-ben/sourcecred-analysis), for instance to calculate Cred stats for updates for the Maker community. In CredRank, due to virtualization of some of the data, we now have `credGraph.json.gzip`, which contains some data, but not enough to do meaningful analysis. Myself -and likely others- are going to need more data in order to perform meaningful analysis in a number of use cases, such as:

- General data viz of graph
- Analysis of weight changes
- Gaming detection
- Research

We have some observable notebooks that pull data from the graph into consumable formats. E.g. the [Cred and Grain CSV Downloader](https://observablehq.com/@sourcecred/cred-and-grain-csv-downloader?branchUrl=https://github.com/sourcecred/cred/tree/gh-pages). This is great. But I imagine a demand from myself and others for lower-level access to more data, which does not require knowledge of Javascript or the relevant APIs. Already, other communities have created their own Python scripts for doing data visualization and automation, which have likely broken with CredRank. 

**Describe the solution you'd like**

I would like lower-level graph data in a more easily consumable format, such as CSV or JSON file. Perhaps a command on the CLI that optionally creates the file, to avoid bloat? No ideas on the data schema, as don't have experience there.. 

![image](https://user-images.githubusercontent.com/1634777/114824649-104d3b80-9d7a-11eb-8349-d21c9cb27dec.png)

Jk. Kinda....I can imagine people (myself included) wanting granular Cred data for analysis. Below I give the MVP (basic data that would unblock me) and additional data we'll have demand for eventually. 

*Basic data*

- Data for each node (users and contributions), including: Cred score, Cred minted, author, platform, node type, timestamp, description.
- Cred data over time, including user Cred scores and contribution Cred scores (if relevant in CredRank). I.e. I want to be able to create plots of a user or contribution's Cred score over time. 
- Data for each edge, such as edge type (what's in `address` might do?), reversed, source, destination, transition probability, Cred flow. 

*Nice to have (but will need in future)*

[...]

PR #2967, opened 2021-04-27
CLI: Add 'analysis' command for generating auxilary data

blueridger:
# Description
I previously removed `accounts.json` because it sounded like people weren't really using it. Turns out they were. Story of 0.8's life. This PR re-adds it, except instead of being an output of the `grain` command as before, a new `analysis` command has been created to generate it. Other data structures will be created in the analysis API in the future.

I chose to create a new command so that our main commands are faster and more semantically atomic.

# Test Plan
1. in sourcecred repo: yarn build
2. in gh-pages branch of cred repo:
  - `scdev analysis` and verified output/accounts.json now exists and looks correct at a glance
  - `scdev help` and `scdev help analysis`


src/cli/help.js
```diff
@@ -55,6 +57,7 @@ function usage(print: (string) => void): void {
       graph         build Cred graph from cached plugin data
       site          update your cred site with the latest changes
       credrank      calculate cred scores from existing graph
+      analysis      generates useful data analysis structures
```
topocount, line 60: nit
```suggestion
      analysis      outputs useful cred and grain data for analysis
```

topocount (commented):
would it be too much work to inflate all active plugin graphs with this command as well?

topocount (approved):

src/cli/help.js
blueridger: `outputs` seems a little misleading as a CLI help tip, since nothing is outputted to the console.

I kept it vague instead of saying cred and grain to leave the door open for any sort of data we want to generate. 

\***slaps roof of analysis command*** "You can fit so many auxilary data structures in this bad boy."

blueridger:
> inflate all active plugin graphs

Don't really know what you mean by this. We definitely have plans to load more stuff in here, see https://github.com/sourcecred/sourcecred/issues/2949

PR #3003, opened 2021-05-03
Set up grain distribution tracking for grain integrations

topocount:
# Semantics

A "tracked distribution" is a distribution that has been created when
distribution tracking is enabled

Tracked distributions are catalogued as "not executed" upon creation, and
only when `markDistributionExecuted` is called, will they be marked as "executed".
Note that `markDistributionExecuted` does not itself executed integration
functionality. This exists outside of the ledger context for now.

# Further Details

Grain integration tracking will allow us to track which distributions are
eligible to be processed by a grain integration. This will be needed for
situations where a grain integration is actually transacting funds, such as
on a blockchain. Grain integrations that generate some form of output that
does not directly relate to a transaction can still execute, but they
may be executed multiple times to generate the report in different
contexts, so there is no need to track these executions in the ledger.

# merge plan

Merge before #3007

# Test Plan
unit tests are provided.

### DIFF
--- a/src/core/ledger/ledger.js
+++ b/src/core/ledger/ledger.js
@@ -143,8 +143,8 @@ export class Ledger {
   _allocationsToDistributions: Map<AllocationId, DistributionId>;
   _latestTimestamp: TimestampMs = -Infinity;
   _lastDistributionTimestamp: TimestampMs | null = null;
-  _trackGrainIntegration: boolean = false;
-  _grainIntegrationDistributions: Map<DistributionId, boolean>;
+  _shouldTrackGrainIntegration: boolean = false;
+  _grainIntegrationStatuses: Map<DistributionId, boolean>;
 
   constructor() {
     this._ledgerEventLog = new JsonLog();
@@ -155,7 +155,7 @@ export class Ledger {
     this._allocations = new Map();
     this._distributions = new Map();
     this._allocationsToDistributions = new Map();
-    this._grainIntegrationDistributions = new Map();
+    this._grainIntegrationStatuses = new Map();
   }
 
   /**
@@ -616,8 +616,8 @@ export class Ledger {
     ) {
       this._lastDistributionTimestamp = distribution.credTimestamp;
     }
-    if (this._trackGrainIntegration) {
-      this._grainIntegrationDistributions.set(distribution.id, false);
+    if (this._shouldTrackGrainIntegration) {
+      this._grainIntegrationStatuses.set(distribution.id, false);
     }
   }
 
@@ -778,7 +778,7 @@ export class Ledger {
    * no-ops if integration tracking is already enabled
    */
   enableIntegrationTracking(): Ledger {
-    if (!this._trackGrainIntegration) {
+    if (!this._shouldTrackGrainIntegration) {
       this._createAndProcessEvent({
         type: "ENABLE_GRAIN_INTEGRATION",
       });
@@ -790,17 +790,21 @@ export class Ledger {
    * no-ops if integration tracking is already disabled
    */
   disableIntegrationTracking(): Ledger {
-    if (this._trackGrainIntegration) {
+    if (this._shouldTrackGrainIntegration) {
       this._createAndProcessEvent({
         type: "DISABLE_GRAIN_INTEGRATION",
       });
     }
     return this;
   }
-  _toggleGrainIntegrationTracking(_: ToggleGrainIntegration) {
-    this._trackGrainIntegration = !this._trackGrainIntegration;
-    if (!this._trackGrainIntegration)
-      this._grainIntegrationDistributions.clear();
+
+  _enableGrainIntegrationTracking(_: EnableGrainIntegration) {
+    this._shouldTrackGrainIntegration = true;
+  }
+
+  _disableGrainIntegrationTracking(_: DisableGrainIntegration) {
+    this._shouldTrackGrainIntegration = false;
+    this._grainIntegrationStatuses.clear();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/core/ledger/ledger.js
```diff
@@ -768,6 +774,79 @@ export class Ledger {
...
+  }
+
+  /**
+   * Returns the status of a tracked distribution
+   * If the distribution has not been executed by the integration, `false` is
+   * returned.
+   * If the distribution has been executed, `true` is returned.
+   * If the distribution is untracked, `undefined` is returned.
+   */
+  integrationExecuted(id: DistributionId): ?boolean {
```
blueridger, line 846: `isIntegrationExecuted` - boolean prefixes like "is" or "has" go a long way for code readability.

src/core/ledger/ledger.js
```diff
@@ -768,6 +774,79 @@ export class Ledger {
...
+  }
+
+  /**
+   * Mark integrations as executed.
+   * Validates that:
+   * a) executions are tracked
+   * b) executions are not yet marked as executed
+   * c) distribution tracking is enabled
+   */
+  runIntegration(id: DistributionId): Ledger {
```
blueridger, line 812: `markDistributionExecuted` / `MARK_DISTRIBUTION_EXECUTED` - I don't imagine integrations should be called by the ledger in the future. Ideally there is an orchestrator at the API level running integrations and then writing them to ledger. So the execution will never be the ledger event; the marking of the execution is the ledger event. And the integration isn't being marked, then distribution is.

src/core/ledger/ledger.js
```diff
@@ -768,6 +774,79 @@ export class Ledger {
     account.payoutAddresses.delete(JSON.stringify(currencyResult.value));
   }
 
+  /**
+   * no-ops if integration tracking is already enabled
+   */
+  enableIntegrationTracking(): Ledger {
+    if (!this._trackIntegration) {
+      this._createAndProcessEvent({
+        type: "TOGGLE_INTEGRATION",
```
blueridger, line 783: We've kinda regretted using TOGGLE for identity activations, because it makes it really hard to reason about whether it's a toggle on or off. Would probably be better to learn from that regret and differentiate between ENABLE_ and DISABLE_ ledger events. Also, "integration" is kinda vague, and it'd be nice to be more specific like ENABLE_PAYOUT_INTEGRATION or ENABLE_DISTRIBUTION_INTEGRATION or ENABLE_GRAIN_INTEGRATION and then use the same terminology for the method name like "enablePayoutTracking" or whatever.

src/core/ledger/ledger.js
```diff
@@ -1016,6 +1103,19 @@ const transferGrainParser: C.Parser<TransferGrain> = C.object({
...
 
+type ToggleIntegration = {|
+  +type: "TOGGLE_INTEGRATION",
+|};
+
+const integrationParser: C.Parser<ToggleIntegration> = C.object({
+  type: C.exactly(["TOGGLE_INTEGRATION"]),
+});
+
+type RunIntegration = {|
```
blueridger, line 1114: Don't you need to create a parser for this too?

src/core/ledger/ledger.test.js
```diff
@@ -1598,6 +1598,154 @@ describe("core/ledger/ledger", () => {
...
+    describe("Toggle Integrations", () => {
+      it("can enable integrations", () => {
+        const ledger = new Ledger();
+        ledger.enableIntegrationTracking();
+        expect(ledger.eventLog()).toEqual([
+          {
+            "action": {
+              type: "TOGGLE_INTEGRATION",
+            },
+            ledgerTimestamp: 4,
```
blueridger, line 1616: ??? is this relying on a stable test order to output stable auto-incremented timestamps? Seems flimsy. Would be better to reset the clock in a "beforeEach" or something.

src/core/ledger/ledger.test.js
```diff
@@ -1598,6 +1598,154 @@ describe("core/ledger/ledger", () => {
...
+              type: "TOGGLE_INTEGRATION",
+            },
+            ledgerTimestamp: 6,
+            uuid: "000000000000000000019A",
+            version: "1",
+          },
+        ]);
+        expect(ledger._trackIntegration).toBe(false);
+      });
+      it("is a no-op when repeatedly disabling integrations", () => {
```
blueridger, line 1647: this it block's message is swapped with the it block below it.

src/core/ledger/ledger.test.js
```diff
@@ -1598,6 +1598,154 @@ describe("core/ledger/ledger", () => {
...
+              "version": "1",
+            },
+          ]);
+          expect(Array.from(ledger.trackedDistributions())).toEqual([
+            distribution.id,
+          ]);
+          expect(ledger.integrationExecuted(distribution.id)).toBe(true);
+        });
+      });
+      describe("trackDistributions", () => {
```
blueridger, line 1721: "tracked"

src/core/ledger/ledger.js
```diff
@@ -768,6 +774,80 @@ export class Ledger {
...
+        type: "DISABLE_GRAIN_INTEGRATION",
+      });
+    }
+    return this;
+  }
+  _toggleGrainIntegrationTracking(_: ToggleGrainIntegration) {
+    this._trackGrainIntegration = !this._trackGrainIntegration;
+    if (!this._trackGrainIntegration)
+      this._grainIntegrationDistributions.clear();
+  }
```
blueridger, line 804: nit: Not totally sold on this. Seems like an unnecessary addition of complexity / point of failure / deviation from the pattern.

PR #3007, opened 2021-05-03
Specify a Grain Integration Contract

topocount:
This PR contains two artifacts that should be implemented by Grain
Integrations.

The first is the `IntegrationConfig`. The integration provides the
expected config values. These config values should then be checked
against the current ledger and instance config to ensure the integration
can execute without error.

The second is the `grainIntegration` function interface. This is
interface accepts a map of payout addresses and grain amounts, and a
Currency type, which contains either a protocol indicator or token
address if the protocol is the EVM-based. If the function returns an
`IntegrationResult` and accounting is enabled, the ledger is updated with
the new balances.

Note that the interface shape will change when the needed updates are made to the
needed configs. This is mostly to show the constituent parts needed to make this
work.

Functions that wrap the grain integration are implemented to show how it
is meant to be used, and are meant to be helpers once this feature is in
production.

# Merge Plan

merge after #3003 

# Test Plan

Unit tests are provided for the helper functions.

The helper functions are meant to illustrate how grain integrations will be
used.

### DIFF
--- a/src/core/ledger/grainIntegration.js
+++ b/src/core/ledger/grainIntegration.js
@@ -5,6 +5,7 @@ import {type Distribution} from "./distribution";
 import {getDistributionBalances} from "./distributionSummary/distributionSummary.js";
 import {type Currency} from "./currency.js";
 import {type IdentityId} from "../identity";
+import * as NullUtil from "../../util/null";
 import * as G from "./grain";
 
 type Transfer = {|
@@ -12,14 +13,14 @@ type Transfer = {|
   memo: string,
 |};
 
-export type AccountDistributions = Map<PayoutAddress, G.Grain>;
+export type PayoutDistributions = Map<PayoutAddress, G.Grain>;
 export type PayoutAddressToId = Map<PayoutAddress, IdentityId>;
 export type TransferredGrain = Map<PayoutAddress, Transfer>;
 
-export type IntegrationResult = {|
+export type PayoutResult = {|
   // Amounts that are actually distributed by the integration.
-  // If Grain is still not tracked offchain, these can be recorded as transfers
-  // in the ledger to the "sink" Identity.
+  // If Grain balances are tracked in the ledger, these will be recorded as
+  // transfers in the ledger to the "sink" Identity.
   transferredGrain: TransferredGrain,
 |};
 
@@ -27,7 +28,7 @@ export type IntegrationConfig = {|
   // This determines whether contemporary grain balances are tracked by the ledger.
   // `true` is effectively the current state of the ledger.
   // `false` sets all grain balances to zero and disables transfers. This
-  // is utilized to enforce the exististence of token balances outside of the
+  // is utilized to enforce the existence of token balances outside of the
   // ledger. Importantly, Grain Receipts from allocations are still tracked,
   // because some grain distribution strategies rely on this information.
   accountingEnabled: boolean,
@@ -47,10 +48,7 @@ export type IntegrationConfig = {|
  * if accounting is enabled. Otherwise, grain balances will be tracked
  * elsewhere.
  */
-export type grainIntegration = (
-  AccountDistributions,
-  Currency
-) => ?IntegrationResult;
+export type GrainIntegration = (PayoutDistributions, Currency) => ?PayoutResult;
 
 ///////////////////
 // Helper functions
@@ -61,14 +59,14 @@ export type grainIntegration = (
 // sink identity
 export function executeGrainIntegration(
   ledger: Ledger,
-  integration: grainIntegration,
+  integration: GrainIntegration,
   distribution: Distribution,
   currency: Currency,
   accountingEnabled: boolean,
   processDistributions: boolean,
   sink?: IdentityId
 ): Ledger {
-  const {distributions, payoutAddressToId} = buildDistributionIndexes(
+  const {payoutDistributions, payoutAddressToId} = buildDistributionIndexes(
     ledger,
     distribution,
     JSON.stringify(currency)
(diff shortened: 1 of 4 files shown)


src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+  memo: string,
+|};
+
+export type AccountDistributions = Map<PayoutAddress, G.Grain>;
+export type PayoutAddressToId = Map<PayoutAddress, IdentityId>;
+export type TransferredGrain = Map<PayoutAddress, Transfer>;
+
+export type IntegrationResult = {|
+  // Amounts that are actually distributed by the integration.
+  // If Grain is still not tracked offchain, these can be recorded as transfers
```
blueridger, line 21: "still not tracked offchain" I feel like this is the opposite of what is meant here. -> "still tracked offchain" or "still not tracked onchain"

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+type Transfer = {|
+  amount: G.Grain,
+  memo: string,
+|};
+
+export type AccountDistributions = Map<PayoutAddress, G.Grain>;
+export type PayoutAddressToId = Map<PayoutAddress, IdentityId>;
+export type TransferredGrain = Map<PayoutAddress, Transfer>;
+
+export type IntegrationResult = {|
```
blueridger, line 19: Integration still feels too vague to me. "Payout" is a lot more descriptive, and could probably be substituted in most places.

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+import * as G from "./grain";
+
+type Transfer = {|
+  amount: G.Grain,
+  memo: string,
+|};
+
+export type AccountDistributions = Map<PayoutAddress, G.Grain>;
+export type PayoutAddressToId = Map<PayoutAddress, IdentityId>;
+export type TransferredGrain = Map<PayoutAddress, Transfer>;
```
blueridger, line 17: nit: Is a map really what you want here? I don't see individual look ups happening, and this seems like something that would want to be serialization-friendly and array.map-friendly (optimize for transformability over lookup).

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * This function definition is implemented by Grain Integrations.
+ * Grain integrations allow distributions to be executed programmatically
+ * beyond the ledger. However, an integration might have some side-effects
+ * that require the ledger to be updated, and it therefore has the option of
+ * returning a list of of ledger operations. The ledger will update the ledger
+ * if accounting is enabled. Otherwise, grain balances will be tracked
+ * elsewhere.
+ */
+export type grainIntegration = (
```
blueridger, line 50: nit: GrainIntegration

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+import {type Currency} from "./currency.js";
+import {type IdentityId} from "../identity";
+import * as G from "./grain";
+
+type Transfer = {|
+  amount: G.Grain,
+  memo: string,
+|};
+
+export type AccountDistributions = Map<PayoutAddress, G.Grain>;
```
blueridger, line 15: nit: PayoutDistributions

Also same question, should this maybe be an array instead of a map?

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+  const payoutAddressToId = new Map();
+  const balances = getDistributionBalances(distribution);
+  for (const [id, amount] of balances.entries()) {
+    const {payoutAddresses, identity} = ledger.account(id);
+    const address = payoutAddresses.get(currencyId);
+    if (!address) continue;
+    // need to allow for identities that have since been merged to still claim
+    // funds if accounts are merged between a grain distribution a
+    // grainIntegration call.
+    const total = distributions.get(address) ?? G.ZERO;
```
blueridger, line 118: I think this is the first time we've used ?? which is an ES2020 operator. Just wondering what the compatibility implications of this are.

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,124 @@
...
+export function executeGrainIntegration(
+  ledger: Ledger,
+  integration: grainIntegration,
+  distribution: Distribution,
+  currency: Currency,
+  accountingEnabled: boolean,
+  processDistributions: boolean,
+  sink?: IdentityId
+): Ledger {
+  const {distributions, payoutAddressToId} = buildDistributionIndexes(
```
blueridger, line 71: distributions -> payoutDistributions

making variable names correspond closely to their type is a readability blessing.

src/core/ledger/grainIntegration.js
topocount: We'll have integrations that don't actually pay out though, for like logging or record keeping purposes

src/core/ledger/grainIntegration.js
topocount: Maps are versatile, and simple enough to iterate over. If an array is needed, using `Array.from()` will suffice on the grain integration side. 

This data structure will strictly be accessed by the Grain Integration in memory,, so a Map is okay. If we start passing it through an RPC or some other transport that requires serilaization for some reason, we'll have to revisit it, course.

src/core/ledger/grainIntegration.js
```diff
@@ -0,0 +1,127 @@
...
+import * as G from "./grain";
+
+type Transfer = {|
+  amount: G.Grain,
+  memo: string,
+|};
+
+export type PayoutDistributions = Array<[PayoutAddress, G.Grain]>;
+export type PayoutAddressToId = Map<PayoutAddress, IdentityId>;
+export type TransferredGrain = Array<[PayoutAddress, Transfer]>;
```
blueridger, line 18: I expected you to add payoutAddress as a field to the Transfer type. Why not do that instead? It would make Transfer semantically more whole and simplify derived types.

(also TransferredGrain -> Transfers)

blueridger (approved):

PR #3012, opened 2021-05-06
Monorepo init

### LINKED ISSUE #2992: Create monorepo with one package (the existing code)
Convert sourcecred/sourcecred into a monorepo, but don't split out the codebase into separate packages. Just refactor the folder structure and install and configure lerna.

hozzjss:
# Description

This is as breaking as it gets the repo was moved from the root folder to packages/sourcecred
I made a few decisions regarding this move that are questionable and in need of review,  
Should resolve #2992
# Test Plan

I have ran both yarn test and yarn test --full
I have checked CircleCI and made sure that the changes are working
I have also followed the [manual test guide](https://github.com/AL0YSI0US/trusty-rabbit-hole/blob/main/setting-up-a-sourcecred-development-env.md) after `cd`ing into packages/sourcecred, and everything seemed to be working fine, but I am very very skeptical, this pull request should be high priority on @topocount's recommendation to minimize overhead and merge conflicts


topocount (commented):
@hozzjss  This is good progress. We can approach many of the smaller things incrementally, but we should be adding scripts in the root `package.json` file via lerna to expose scripts in the package manifests.

For example, we should be able to run `yarn test` and `yarn test --full` from the monorepo root and see the appropriate tests run in all packages. This entry should look something like this
```json
{
  "scripts": {
    "test": "lerna run test"
  }
}
```
This will run the test script in all packages. If you want to filter on a specific package, you  run
```bash
$ yarn lerna run test --scope=sourcecred
```
**Note** this filters on the `name` field in package.json, not the package folder name.
Here are more details: https://github.com/lerna/lerna/tree/main/core/filter-options

In summary we should never run scripts from non-root directories. Lerna will do that for us.

packages/sourcecred/config/webpack.config.web.js
```diff
@@ -181,9 +181,11 @@ function makeConfig(mode /*: "production" | "development" */) /*: Object */ {
...
       strictExportPresence: true,
       rules: [
-        // TODO: Disable require.ensure as it's not a standard language feature.
-        // We are waiting for https://github.com/facebookincubator/create-react-app/issues/2176.
-        // { parser: { requireEnsure: false } },
+        {
+          parser: {
+            requireEnsure: false,
+          },
+        },
```
topocount, line 188: I appreciate you catching this, but it'd be cleaner to make this its own PR

packages/sourcecred/src/plugins/discord/update_discord_api_snapshots.sh
```diff
@@ -15,7 +15,7 @@ if [ -z "${SOURCECRED_TEST_SERVER_TOKEN:-}" ]; then
   exit 1
 fi
 
-toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)/packages/sourcecred"
```
topocount, line 18: Can you capture this as an environment variable at the top of the script? It's much more readable to hoist this sort of detail to the top of the file, and then invoke it where needed, even if it;s just once. 

We might have to reconsider how we organize our bash scripts so we can centralize details like this and make debugging, updates and refactors easier. 

How do you feel about storing this value at repo root and then appending it everywhere it's needed, rather than appending this raw string everywhere? Basically just create a `CORE_SUBPATH` variable with this string in a monorepo_vars.sh file that gets invoked everywhere.

packages/sourcecred/src/plugins/discourse/update_discourse_api_snapshots.sh
```diff
@@ -10,7 +10,7 @@ if [ ! "$(jq --version)" ]; then
   exit 1
 fi
 
-toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)/packages/sourcecred"
```
topocount, line 13: same as above

packages/sourcecred/src/plugins/git/example/synchronizeToGithub.sh
```diff
@@ -6,7 +6,7 @@ main() {
     if [ "${1:-}" = "-n" ] || [ "${1:-}" = "--dry-run" ]; then
         DRY_RUN=1
     fi
-    cd "$(git rev-parse --show-toplevel)"
+    cd "$(git rev-parse --show-toplevel)/packages/sourcecred"
```
topocount, line 9: same as above

packages/sourcecred/src/plugins/github/fetchGithubOrgTest.sh
```diff
@@ -60,7 +60,7 @@ main() {
   if [ -n "${SOURCECRED_BIN:-}" ]; then
     BIN="$(readlink -f "${SOURCECRED_BIN}")"
   fi
-  cd "$(git rev-parse --show-toplevel)"
+  cd "$(git rev-parse --show-toplevel)/packages/sourcecred"
```
topocount, line 63: same as above

packages/sourcecred/src/plugins/wip-discord/update_discord_api_snapshots.sh
```diff
@@ -15,7 +15,7 @@ if [ -z "${SOURCECRED_DISCORD_BOT_TOKEN:-}" ]; then
   exit 1
 fi
 
-toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)/packages/sourcecred"
```
topocount, line 18: same as above

scripts/flow-mono.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+
+for package in $(ls ${toplevel}/packages); do
+  cd $toplevel
+  if [ -f "$toplevel/packages/$package/.flowconfig" ]; then
+    yarn flow-mono create-symlinks $toplevel/packages/$package/.flowconfig
+    yarn flow-mono create-symlinks $toplevel/packages/$package/.flowconfig-ci
+  fi;
+done
```
topocount, line 10: ballin'

topocount (commented):
@hozzjss This is sick! One thing to noodle on, but otherwise this came together really nicely.

package.json
```diff
@@ -1,179 +1,19 @@
...
-      "json",
-      "web.jsx",
-      "jsx",
-      "node",
-      "mjs"
-    ]
+    "test": "lerna run test --",
+    "build": "lerna run build",
+    "install": "lerna bootstrap ; ./scripts/flow-mono.sh",
+    "start": "lerna run start"
```
topocount, line 14: Lerna doesn't print package process outputs until the jobs finish. These flags will allow packages to print to the console in real time.

```suggestion
    "test": "lerna run test --stream --",
    "build": "lerna run build --stream",
    "install": "lerna bootstrap ; ./scripts/flow-mono.sh",
    "start": "lerna run start --stream"
```

packages/sourcecred/config/webpack.config.web.js
hozzjss: That makes sense, I fixed that

packages/sourcecred/src/plugins/discord/update_discord_api_snapshots.sh
hozzjss: That sounds like a great idea I'll look into how to define reusable variables in bash and do that

package.json
hozzjss: That is awesome thanks for that!

scripts/flow-mono.sh
hozzjss: Awesome!

scripts/monorepo_vars.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+export CORE_SUBPATH="$(git rev-parse --show-toplevel)/packages/sourcecred"
```
topocount, line 2: Let's derive the core path programmatically using the sh and lerna:
```suggestion
CORE_PACKAGE_SUBPATH=$(lerna ll --loglevel=silent | grep ^sourcecred' ' | cut --delimiter=' ' --fields=3)
export CORE_PATH="$(git rev-parse --show-toplevel)/$CORE_PACKAGE_SUBPATH"
```

packages/sourcecred/scripts/deploy.sh
```diff
@@ -7,15 +7,17 @@ set -eu
...
 
 export GIT_CONFIG_NOSYSTEM=1
 export GIT_ATTR_NOSYSTEM=1
 
 main() {
     parse_args "$@"
 
-    toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)/packages/sourcecred"
-    cd "${toplevel}"
+    cd "$CORE_SUBPATH"
```
topocount, line 20: This is technically `CORE_PATH`

scripts/monorepo_vars.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/sh
-export CORE_SUBPATH="$(git rev-parse --show-toplevel)/packages/sourcecred"
\ No newline at end of file
+CORE_PACKAGE_SUBPATH=$(lerna ll --loglevel=silent | grep ^sourcecred' ' | cut --delimiter=' ' --fields=3)
+export CORE_PATH="$(git rev-parse --show-toplevel)/$CORE_PACKAGE_SUBPATH"
```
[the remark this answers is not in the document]
hozzjss: Oh so that's what broke it, thanks for the tip dude

packages/sourcecred/scripts/deploy.sh
hozzjss: Oh now I get it

hozzjss:
During CI for some reason the pacakges/sourcecred/sharness/load_test_instance.t file it won't resolve `CORE_PATH`
in this line
```
    snapshot_directory="${CORE_PATH}/sharness/__snapshots__/test-instance/" &&
```
and would always resolve it to be `/home/circleci/repo//sharness/__snapshots__/test-instance/` which is weird
tried to debug but couldn't find a suitable solution

packages/sourcecred/sharness/load_test_instance.t
```diff
@@ -19,15 +19,18 @@ export GIT_ATTR_NOSYSTEM=1
 # shellcheck disable=SC1091
 . ./sharness.sh
 
+toplevel="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+
+. $toplevel/scripts/monorepo_vars.sh
+
```
topocount, line 25: This is a nuance with Lerna that I honestly don't completely understand. Inside a package, lerna will not return its
subpath.
Here's a fix:
```suggestion
cd "${toplevel}" || exit 1
. $toplevel/scripts/monorepo_vars.sh
cd "${CORE_PATH}" || exit 1

```

topocount (commented):
@hozzjss this should do it

topocount:
@hozzjss sorry I thought cd'ing around was a bit hacky so I just made the Path construction from lerna more robust

topocount:
@hozzjss if you can get this rebased, we can merge it today

topocount (approved):

hozzjss:
> @hozzjss if you can get this rebased, we can merge it today

Will do!!

Issue #2779, opened 2021-02-22
Grain distribution quantity input should be string that allows a decimal point


befitsandpiper:
# User Story
As an instance maintainer in a community with a very expensive token and a small instance budget, I want to be able to distribute a fraction of a token so that I can distribute smaller/more precise amounts than increments of a single token.

Example: In 1hive, the $HNY token price is ~900 USD, so it is impossible to distribution $500 per week, since the minimum allowed is 1 grain = 1 $HNY = $900.

# Tech Notes
Some files of interest:
https://github.com/sourcecred/sourcecred/blob/master/src/api/grainConfig.js
https://github.com/sourcecred/sourcecred/tree/master/src/core/ledger/policies
https://github.com/sourcecred/sourcecred/blob/master/src/core/ledger/grain.js

The allocation configs currently parse a json number and validate that it is an integer. This should be changed to a `string` representation of a number that may have a decimal point and that is [parseable to `Grain`](https://github.com/sourcecred/sourcecred/blob/master/src/core/ledger/grain.js#L233).

# Alternatives considered
An alternative solution would be to make your instance's grain not 1:1 with the token, for example 10 grain could be equal to 1 $HNY. This is difficult to enact retroactively though, because it would mean re-writing the entire ledger to create a consistent grain evaluation. And also, this is difficult because it adds a new currency value / conversion, which would probably mean more confusion and cognitive load for the entire community.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200032326400385/1200038210995889) by [Unito](https://www.unito.io/learn-more)

topocount:
@befitsandpiper can you lay out this user story a bit more? I think this just might be an issue of configuration.

Floats should not be used with cryptocurrencies because they are non-deterministic. Grain in sourceCred is typed as a BigInt but by default the first 18 of those digits are reserved as decimals by default: https://github.com/sourcecred/sourcecred/blob/aa1a46943d83da3f61bbfae797021976e08d1240/src/core/ledger/grain.js#L46-L50

Likewise, ERC20 tokens in Ethereum are similar, in that they are typed as uint256 (256 bits, or 32 bytes, are reserved for the value), but the first 18 digits are typically reserved for decimals.

blueridger:
We should be able to take a float in the configuration and immediately convert it to a grain string without risk of non-deterministic behavior.

wchargin:
Floats aren’t “nondeterministic”, and neither are the common arithmetic
[...]

PR #3057, opened 2021-05-18
Fractional grain

BrianBelhumeur:
Closes issue #2779 

Paired with @hozzjss on this (thank you!)

# Summary
Currently, grain distributions can only be made in whole integer amounts. For certain projects (in this case, 1Hive) this can cause a problem because they map grain:TOKEN (whatever token they use) 1:1. This means they can only distribute in whole-token quantities. For tokens with high values, this can be undesirable. E.g. the project may want to award half a token's worth of value in a given period.

This PR changes grain budget parsing to allow for, in addition to integer values, string-defined floating point values. The reason we use a string for floating-point values is to avoid floating-point math (rounding) errors.

I also considered allowing floating-point numbers in the budget JSON files, converting them to strings right away, and then converting those to Grain, but I am not convinced that conversion from floating-point number to string representation of floating-point number is always lossless. It may be, and if it is, a future update should allow for that possibility as well. Or I can update this PR with that functionality. I tried to research it, but I could not find any definitive info on it.

# Testing
I modified and added unit tests, as well as tested budgets by hand using `scdev grain -s` on the `cred` repo with floating-point string and integer budgets
## Screenshot
Screen Shot 2021-05-19 at 00 21 22

### DIFF
--- a/packages/sourcecred/src/api/grainConfig.js
+++ b/packages/sourcecred/src/api/grainConfig.js
@@ -3,22 +3,25 @@
 import {type DistributionPolicy} from "../core/ledger/applyDistributions";
 import * as C from "../util/combo";
 import * as NullUtil from "../util/null";
+import * as G from "../core/ledger/grain";
 import {
   type AllocationPolicy,
   policyConfigParser,
 } from "../core/ledger/policies";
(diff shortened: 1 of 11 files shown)


AL0YSI0US (approved):
The "howdy" lines in the codebase got me. 🤣
Great Job Y'all.

PR #3059, opened 2021-05-19
flow mono is unnecesary when package is deployed

### LINKED ISSUE #3058: 0.8.6 breaks github action during package install
The flow mono script that exists within the monorepo root directory is being run even though the npm package has no access to it
Log:
```
error /home/runner/work/sc-stacks-test/sc-stacks-test/node_modules/sourcecred: Command failed.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Exit code: 127
Command: ../../scripts/flow-mono.sh
Arguments: 
Directory: /home/runner/work/sc-stacks-test/sc-stacks-test/node_modules/sourcecred
Output:
/bin/sh: 1: ../../scripts/flow-mono.sh: not found
Error: Process completed with exit code 127.
```

hozzjss:
This should resolve #3058 
But it needs a few tweaks still one of these is to run flow-mono.sh during development or only during ci which I think already runs because flow tests would fail without it, let me know if you think this necessary
Also one caveat is that this needs to be released as a hotfix as the current version breaks all instances

### DIFF
--- a/packages/sourcecred/package.json
+++ b/packages/sourcecred/package.json
@@ -130,8 +130,7 @@
     "flow": "flow",
     "lint": "eslint src config --max-warnings 0",
     "prepublishOnly": "yarn build",
-    "docs": "node ./scripts/build-docs.js",
-    "postinstall": "../../scripts/flow-mono.sh"
+    "docs": "node ./scripts/build-docs.js"
   },
   "engines": {
     "node": ">=12"
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "test": "lerna run test --stream --",
     "build": "lerna run build --stream",
-    "install": "lerna bootstrap ; ./scripts/flow-mono.sh",
+    "install": "lerna bootstrap && ./scripts/flow-mono.sh",
     "start": "lerna run start --stream"
   },
   "workspaces": [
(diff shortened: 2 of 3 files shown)


packages/sourcecred/package.json
```diff
@@ -130,8 +130,7 @@
     "flow": "flow",
     "lint": "eslint src config --max-warnings 0",
     "prepublishOnly": "yarn build",
-    "docs": "node ./scripts/build-docs.js",
-    "postinstall": "../../scripts/flow-mono.sh"
```
topocount, line 134: oof my bad. this shouldn't have gotten merged

topocount (commented):
@hozzjss I've fixed the duplicate execution of flow-mono.

packages/sourcecred/package.json
hozzjss: What's the game plan here

topocount (commented):
@hozzjss please record a test plan in the PR description. I have a bit of a hacky one, but I'm curious to hear your idea.

We'll push out a release as soon as we get this merged.

topocount (approved):
This test plan doesn't fail in the main branch, since the code in the development repo is successully installed already.

My suggestion: 
1. temporarily delete the flow-mono.sh script
2. cd into the sourcecred package and attempt to install. ensure nothing outside of the package folder is executed, and the install succeeds. note: tests will not pass when invoked from here though.

PR #3063, opened 2021-05-22
Discord: show a friendlier message on 403s

blueridger:
403's are normal when the bot encounters a private channel it doesn't have perms to see, so we should not output a scary stacktrace.

# Test plan
`scdev load`

Before:
```
  GO   sourcecred/discord: discord/SourceCred Test Server/#moderator-only
Error: 403 Forbidden: bad API username or key?
https://discordapp.com/api/channels/830088231538130965/messages?after=0&limit=100
    at failIfMissing (webpack:///./src/plugins/discord/fetcher.js?:12:2767)
    at Fetcher._fetchJson (webpack:///./src/plugins/discord/fetcher.js?:12:143)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Fetcher.messages (webpack:///./src/plugins/discord/fetcher.js?:12:1480)
    at async Mirror.addMessages (webpack:///./src/plugins/discord/mirror.js?:9:88)
    at async Mirror.update (webpack:///./src/plugins/discord/mirror.js?:7:623)
    at async DiscordPlugin.load (webpack:///./src/plugins/discord/plugin.js?:22:1331)
    at async Promise.all (index 0)
    at async loadCommand (webpack:///./src/cli/load.js?:16:277)
 DONE  sourcecred/discord: discord/SourceCred Test Server/#moderator-only: 160ms
```

After:
```
  GO   sourcecred/discord: discord/SourceCred Test Server/#moderator-only
Skipping private channel.
 DONE  sourcecred/discord: discord/SourceCred Test Server/#moderator-only: 158ms
```

### DIFF
--- a/packages/sourcecred/src/plugins/discord/mirror.js
+++ b/packages/sourcecred/src/plugins/discord/mirror.js
@@ -36,7 +36,10 @@ export class Mirror {
       try {
         await this.addMessages(channel.id);
       } catch (e) {
-        console.warn(e);
+        const warn = e?.message?.includes("403")
+          ? "Skipping private channel."
+          : e;
+        console.warn(warn);
       }
       reporter.finish(`discord/${guild.name}/#${channel.name}`);
     }


packages/sourcecred/src/plugins/discord/mirror.js
```diff
@@ -36,7 +36,10 @@ export class Mirror {
       try {
         await this.addMessages(channel.id);
       } catch (e) {
-        console.warn(e);
+        const warn = e.message.includes("403")
```
hozzjss, line 39: I might not be particularly satisified with this check and also skeptical about the message being part of the error, do we have any other object that we can be sure of, like status, or any check as to not break the execution if the message was an optional at times, I imagine a scenario where either the api changes for these messages, or a time where the internet cuts off and the execution is halted because of "Cannot read property includes of null"

hozzjss (commented):
Gosh I hate these lol, great work @blueridger love it, had just a small comment you might have a better sense, let me know what you think!

packages/sourcecred/src/plugins/discord/mirror.js
blueridger: we throw this error ourselves, so it should be ok
https://github.com/sourcecred/sourcecred/blob/main/packages/sourcecred/src/plugins/discord/fetcher.js#L229

I will add null safety though, since who knows what could find it's way up the promise chain.

hozzjss: Awesome!

hozzjss (approved):
All good to go!

Issue #2975, opened 2021-04-29
[Identity] As a user, I can auth with an Eth wallet.


blueridger:
# AC
- Authing with web3modal and web3 libraries

# Not AC
- signatures
- IDX integrations
- Doing anything with the provider once its connected

PR #3068, opened 2021-05-24
Setup Web3Context with web3modal and ethers.js

hammadj:
Sets up the logic and components for connecting to Ethereum wallets using Web3Modal with support for both MetaMask and WalletConnect.
UI to actually login will be implemented in upcoming PRs.

Related: #2975

Test Plan: Ensure app runs without crashing


packages/sourcecred/src/ui/utils/Web3Context.js
```diff
@@ -0,0 +1,106 @@
+// @flow
+
+import WalletConnectProvider from "@walletconnect/web3-provider";
+import {providers} from "ethers";
+import * as React from "react";
+import Web3Modal from "web3modal";
+
+const INFURA_ID = "a60f8c4d3d4a40a49c4568570a7546b7";
```
hammadj, line 8: Created a SourceCred project on Infura under my account. Can replace this with an official sourcecred account if/when we need but the free tier should suffice for a long time

blueridger: Can you provide a little more context on this?

hammadj: Infura is used to connect users to an existing Ethereum node without needing to run a node / browser client themselves. Its pretty much used by every dapp out there, and required by WalletConnect. If we get big enough for us to start hitting the limits, we can change it to require instance maintainers to provide their own Infura ID in config somewhere.

blueridger (approved):

PR #3069, opened 2021-05-24
Update README.md

blueridger:
The root should just be for summarizing and navigating to individual packages.


AL0YSI0US (approved):
> "The root should just be for summarizing and navigating to individual packages." 
> -@blueridger

@hozzjss ☝🏾 Looks like what we made was a reference document. 
Something a lil different is needed for the front-facing repo.

Maybe we can nail down all the packages and begin drafting the README.
I'd also like to pick your brain about the current refactoring (monorepo)

hozzjss:
Glad that wasn't a breaking change

PR #3070, opened 2021-05-24
Add web3 login button in AppBar

hammadj:
Created LoginButton component to integrate logic from Web3Context and added it to the AppBar

Related: #2975

Test Plan: Run app, click Login in the app bar, sign in with MetaMask / WalletConnect and ensure your address is shown.


packages/sourcecred/src/ui/components/AppBar.js
```diff
@@ -137,6 +138,7 @@ const AppBar = (props: Props): React.Node => {
             children
           )}
           <LoadingIndicator />
+          <LoginButton />
```
blueridger, line 141: You'll find an `isDev` flag in AdminApp.js

Would you pass it to this module and && it with the button? You can test by adding the "?dev=true" URL param

hammadj: Addressed

blueridger (approved):

PR #3075, opened 2021-05-27
Implement readWeightedGraphForPlugin() in readInstance.js

magwalk:
Following the model of localInstance.js, this implements readWeightedGraphForPlugin() in readInstance.js and its corresponding private function in readInstance.js so in a browser-compatible way. 

# Test Plan
## Testing in node 
```yarn build
node
sc = require("./packages/sourcecred/dist/server/api.js")
i = sc.sourcecred.instance.readInstance.getNetworkReadInstance("https://raw.githubusercontent.com/sourcecred/cred/gh-pages/")
i.readWeightedGraphForPlugin("sourcecred/discourse").then(r => console.log(r.graph._nodes))
```
The output should be a large object and should not error.

### DIFF
--- a/packages/sourcecred/src/api/instance/readInstance.js
+++ b/packages/sourcecred/src/api/instance/readInstance.js
@@ -154,8 +154,17 @@ export class ReadInstance implements ReadOnlyInstance {
     throw "not yet implemented";
   }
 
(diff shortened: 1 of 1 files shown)


hozzjss (approved):
This rocks!!

PR #3077, opened 2021-05-31
Update Grain Integration to Accept file output

topocount:
This metadata will be useful for storing outputs, reports, and other
metadata one might want to access from a grain distribution.


blueridger (approved):
This code is getting kinda gnarly... I forget, is there a design doc for any of this I could look at?

PR #3078, opened 2021-05-31
Implement CSV Grain Integration Plugin

topocount:
# Description
This plugin outputs a csv file. Each row contains a payout address in
the first column and the raw amount to be paid in the second column.

The output is optimized for https://disperse.app

# Merge Plan
Merge after #3077 

# Test Plan
unit tests are provided.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 
 # dependencies
 /node_modules
-
+**/node_modules
 # testing
 /coverage
 
@@ -27,5 +27,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+lerna-debug.log
 .DS_Store
-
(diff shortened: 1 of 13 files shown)


blueridger (approved):
Can you comment on why the grainIntegrations folder is being put in the root directory? These PR descriptions feel slim for how many design decisions are being made.

topocount:
> Can you comment on why the grainIntegrations folder is being put in the root directory? These PR descriptions feel slim for how many design decisions are being made.

Sure, I think making them separate packages now will save some overhead later. I see them as optional extensions of SourceCred and not necessarily part of the SourceCred core product, so putting each of them individually in the packages folder felt like an indication of a tighter coupling than we desire. I'm happy to make grainIntetegrations a subfolder of packages though

blueridger:
> make grainIntegrations a subfolder of packages though

This would feel more intuitive to me.

PR #3079, opened 2021-05-31
Set up Grain Integration support in Config API

topocount:
# Description
This PR sets up a grain config for grain integrations. The config is validated against a `bundledGrainIntegrations` allow-list. and the mapped integration function is then set as a part of the config.

The ledger currency parser is now utilized to allow for token and some other crypto configs to be set in an instance and utilized by the ledger when setting payout addresses. NOTE: this config is not directly utilized by the ledger yet

Finally, The grain integration interface has been updated to accept the entire config. This is done to allow for verification that the instance is configured correctly by the intance, and potentially error if a config is incompatible with the invoke grain integration.

# Merge Strategy
Merge after #3078 

# Test Plan
unit tests have been updated

### DIFF
--- a/packages/sourcecred/src/api/bundledGrainIntegrations.js
+++ b/packages/sourcecred/src/api/bundledGrainIntegrations.js
@@ -1,0 +1,36 @@
+// @flow
+
+import type {GrainIntegrationFunction} from "../core/ledger/grainIntegration";
+import * as C from "../util/combo";
+
+import {csvIntegration} from "sc-grainIntegration-csv";
+
+export type GrainIntegration = {|
+  name: string,
+  function: GrainIntegrationFunction,
+|};
+
+type AllowedDeclarations = {[pluginKey: string]: GrainIntegrationFunction};
+
+const allowedDeclarations: AllowedDeclarations = {
+  "csv": csvIntegration,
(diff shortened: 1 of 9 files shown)


packages/sourcecred/src/api/bundledGrainIntegrations.js
```diff
@@ -0,0 +1,32 @@
...
+): GrainIntegrationFunction {
+  switch (integration) {
+    case "csv":
+      return csvIntegration;
+    default:
+      throw new Error(
+        "grain integration not found; enter a valid `integration` value in config/grain.json"
+      );
+  }
+}
```
blueridger, line 24: This diverges from the pattern used by bundledPlguins and bundledDeclarations. Why?

topocount: Updated to better approximate the pattern in other bundle files.

I really wanted to have an error message thrown if the config was invalid, since an invalid config would lead to undefined behavior. I thought a switch/case control flow best suited that but I don't mind conforming to the norm. It's also much easier to read all available plugins without reading through all `case`s, I suppose.

blueridger (approved):

PR #3082, opened 2021-05-31
Bensodenkamp clear error issue761

bensodenkamp:


packages/sourcecred/src/plugins/github/fetchGithubRepo.js
```diff
@@ -329,8 +328,8 @@ export async function postQuery(
           break;
         case "GRAPHQL_ERROR":
           console.error(
-            "Unexpected GraphQL error; this may be a bug in SourceCred: ",
-            JSON.stringify({postBody: postBody, error: error.error})
+            "Unexpected GraphQL error: " + error.error.errors[0].message
+            /* JSON.stringify({postBody: postBody, error: error.error} */
```
blueridger, line 332: Ideally we don't regress other error cases just to make 1 known error case cleaner. A better solution would be to search error.error.errors for an object with type === "NOT_FOUND" and print a nice message (something like "X repo was not found. Please check for typos and ensure that your API key has permission to read the repo.") if found, but keep the original behavior if not found. This way we are precise when we know what the issue is, and verbose when we don't.

blueridger (commented):
You have write perms on this repo, can you please change the PR to pull from a branch instead of a fork? That way the full CI suite can run.

PR #3083, opened 2021-06-01
Better error for wrong repository name

bensodenkamp:
Redoing this PR on non-forked branch. Graphql returns an object with a list of errors, so iterating over them. Giving a nice message if an error is 'NOT_FOUND' and keeping previous behavior for any error that is not 'NOT_FOUND'. In the latter case, only sowing the errors inside the graphql error now.

Test Plan:

From the packages/sourcecred directory of the local sourcecred/sourcecred repo:
yarn build
chmod +x bin/fetchAndPrintGithubRepo.js
bin/fetchAndPrintGithubRepo.js [a github user] [a nonexistant repository] [a valid github api token for the specified user]

Confirm the follwing error:
Unable to find the specified repository. Please check for typos in the repository owner and name and confirm that you are using the correct token in $SOURCECRED_GITHUB_TOKEN

bin/fetchAntPrintGithubRepo.js [a github user] [the name of one of that user's repos] [a valid github api token for the specified user]

Confirm a non-error object starting with:

{
    "__typename": "Repository",

### DIFF
--- a/packages/sourcecred/src/plugins/github/fetchGithubRepo.js
+++ b/packages/sourcecred/src/plugins/github/fetchGithubRepo.js
@@ -327,7 +327,6 @@ export async function postQuery(
           console.error(error.error);
           break;
         case "GRAPHQL_ERROR":
-
           error.error.errors.forEach((error) => {
             if (error.type === "NOT_FOUND") {
               console.error(
--- a/packages/sourcecred/src/plugins/github/fetchGithubRepo.js
+++ b/packages/sourcecred/src/plugins/github/fetchGithubRepo.js
@@ -329,7 +329,7 @@ export async function postQuery(
         case "GRAPHQL_ERROR":
 
           error.error.errors.forEach((error) => {
-            if (error.type == "NOT_FOUND") {
+            if (error.type === "NOT_FOUND") {
               console.error(
                 "Unable to find the specified repository. Please check for typos " +
                   "in the repository owner and name and confirm that you are using the correct token in " +
--- a/packages/sourcecred/src/plugins/github/fetchGithubRepo.js
+++ b/packages/sourcecred/src/plugins/github/fetchGithubRepo.js
@@ -133,7 +133,7 @@ const GITHUB_GRAPHQL_SERVER = "https://api.github.com/graphql";
 
 type GithubResponseError =
   | {|+type: "FETCH_ERROR", error: Error|}
-  | {|+type: "GRAPHQL_ERROR", error: mixed|}
+  | {|+type: "GRAPHQL_ERROR", error: {errors: [{|type: string|}]}|}
   | {|+type: "RATE_LIMIT_EXCEEDED", error: mixed|}
   | {|+type: "GITHUB_INTERNAL_EXECUTION_ERROR", error: mixed|}
   | {|+type: "BAD_CREDENTIALS", error: mixed|}
@@ -313,8 +313,7 @@ export async function postQuery(
   });
   return retryGithubFetch(postBody, token).catch(
     (error: GithubResponseError) => {
-      const type = error.type;
-      switch (type) {
+      switch (error.type) {
         case "GITHUB_INTERNAL_EXECUTION_ERROR":
         case "NO_DATA":
           console.error(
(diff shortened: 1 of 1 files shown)


bensodenkamp:
Test Plan:


From the packages/sourcecred directory of the local sourcecred/sourcecred repo:
yarn build
chmod +x bin/fetchAndPrintGithubRepo.js
bin/fetchAndPrintGithubRepo.js [a github user] [a nonexistant repository] [a valid github api token for the specified user]
* Confirm the follwing error:

Unable to find the specified repository. Please check for typos in the repository owner and name and confirm that you are using the correct token in $SOURCECRED_GITHUB_TOKEN

packages/sourcecred/src/plugins/github/fetchGithubRepo.js
```diff
@@ -328,10 +327,20 @@ export async function postQuery(
           console.error(error.error);
           break;
         case "GRAPHQL_ERROR":
-          console.error(
-            "Unexpected GraphQL error; this may be a bug in SourceCred: ",
-            JSON.stringify({postBody: postBody, error: error.error})
-          );
+          error.error.errors.forEach((error) => {
```
blueridger, line 330: This introduces a new assumption that error.error will be defined and will have an errors property, and this code will now crash if it doesn't--it's not null-safe. Looking at the [graphql spec](http://spec.graphql.org/June2018/#sec-Errors) it seems like this is probably an ok assumption to make, but it's still a little nervous-making and it's the kind of thing that'd be worth noting and justifying in the PR description in the future. I don't know enough about graphql and github's wrapping of it to feel comfortable counting on much of anything. I think I will ask that you add some sort of fallback strategy to make this nullish safe in the case that the API returns something unexpected.

packages/sourcecred/src/plugins/github/fetchGithubRepo.js
```diff
@@ -313,8 +313,7 @@ export async function postQuery(
   });
   return retryGithubFetch(postBody, token).catch(
     (error: GithubResponseError) => {
-      const type = error.type;
-      switch (type) {
+      switch (error.type) {
```
blueridger, line 316: I appreciate the little refactor-as-you-go changes you baked in.

blueridger (commented):
Would you also add a test plan to the description that describes how you tested it and what test cases you tested?

packages/sourcecred/src/plugins/github/fetchGithubRepo.js
bensodenkamp: So I'm still learning about using javascript in a node context, and what flow type checking does and does not protect me from, so I did some reading and experimentation. I may be totally wrong, but here's what I seem to have figured out:

the 'error' that is being evaluated is generated between lines 156 and 193:

        if (x.errors) {
          if (
            x.errors.length === 1 &&
            x.errors[0].message.includes("it could be a GitHub bug")
          ) {
            return Promise.reject(
              ({
                type: "GITHUB_INTERNAL_EXECUTION_ERROR",
                error: x,
              }: GithubResponseError)
            );
          } else if (
            x.errors.length === 1 &&
            x.errors[0].type === "RATE_LIMITED"
          ) {
            return Promise.reject(
              ({type: "RATE_LIMIT_EXCEEDED", error: x}: GithubResponseError)
            );
          } else {
            return Promise.reject(
              ({type: "GRAPHQL_ERROR", error: x}: GithubResponseError)
            );
          }
        }
        if (x.data === undefined) {
          if (x.message && x.message.includes("Bad credentials")) {
            return Promise.reject(
              ({type: "BAD_CREDENTIALS", error: x}: GithubResponseError)
            );
          } else {
            // See https://github.com/sourcecred/sourcecred/issues/350
            return Promise.reject(
              ({type: "NO_DATA", error: x}: GithubResponseError)
            );
          }
        }
        return Promise.resolve(x.data);
      }),

It is a wrapper for the response error, and it checks to make sure the response error has an 'errors' before returning the wrapper. So we already know that if something has a type of 'GRAPHQL_ERROR', the error.error in that object has an errors. I suppose we don't know if error.error.errors is actually a list, but this would already have puked if it is not:

if (x.errors) {
          if (
            x.errors.length === 1 &&
            x.errors[0].message.includes("it could be a GitHub bug")

We also don't know if the list that is error.error.errors is a list of things that all have a 'type' attribute, but based on some experiments I've done, this code will not puke if 'type' is simply missing. It will just evaluate false on the ===:

if (error.type === "NOT_FOUND") {
              console.error(
                "Unable to find the specified repository. Please check for typos " +
                  "in the repository owner and name and confirm that you are using the correct token in " +
                  "$SOURCECRED_GITHUB_TOKEN"
              );
            } else {
              console.error(
                "Unexpected GraphQL error: " +
                  JSON.stringify({postBody: postBody, error: error})
              );
            }
          });

Does this evaluation seem correct? Do I have a bad assumption somewhere obvious(ish) here?

packages/sourcecred/src/plugins/github/fetchGithubRepo.js
bensodenkamp: heh, flow actually forced me to do this

bensodenkamp:
> Would you also add a test plan to the description that describes how you tested it and what test cases you tested?

So I put a basic smoke test in the description, but it does occur to me that we don't seem to have a clean way to mock responses so that we could actually do unit tests against malformed responses. Think this is worth looking into? Might be a significant-ish refactor.

packages/sourcecred/src/plugins/github/fetchGithubRepo.js
blueridger: Yeah, good research! The only assumption being made is that a known-truthy errors attribute is an array, which seems safe enough to assume. No change needed.

blueridger (approved):

PR #3084, opened 2021-06-01
Cosmetic updates to ledger events

bensodenkamp:
fixed: Update design of Ledger EVENT types in UI

### DIFF
--- a/packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
+++ b/packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
@@ -63,21 +63,14 @@ type LedgerEventRowProps = {|
 type EventRow = (LedgerEventRowProps) => ReactNode;
 
 function _friendlyFormatActionType(type: string): string {
-  //Memoize computed label values. Small set of expected labels,
-  //potentially large number of times displayed. Best to save the values.
-
-  const label = useMemo(() => {
-    //Split words by _, capitalize them, and join them with " "
-    var words = type.split("_");
-    words.forEach((word, idx) => {
-      word = word.toLowerCase();
-      word = word.charAt(0).toUpperCase() + word.slice(1);
-      words[idx] = word;
-    });
-    return words.join(" ");
-  }, [type]);
-
-  return label;
+  //Split words by _, capitalize them, and join them with " "
+  var words = type.split("_");
+  words.forEach((word, idx) => {
+    word = word.toLowerCase();
+    word = word.charAt(0).toUpperCase() + word.slice(1);
+    words[idx] = word;
+  });
+  return words.join(" ");
 }
 
 const LedgerEventRow = (props: LedgerEventRowProps): ReactNode => {
@@ -149,13 +142,13 @@ const LedgerEventRow = (props: LedgerEventRowProps): ReactNode => {
     }
   };
 
+  const type = props.event.action.type;
+  const label = useMemo(() => _friendlyFormatActionType(type), [type]);
   return (
     <TableRow>
       <TableCell component="th" scope="row">
         <ThemeProvider theme={actionTheme}>
-          <Typography color="primary">
-            {_friendlyFormatActionType(props.event.action.type)}
-          </Typography>
+          <Typography color="primary">{label}</Typography>
         </ThemeProvider>
       </TableCell>
       <TableCell>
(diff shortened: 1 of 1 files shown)


packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
```diff
@@ -122,9 +145,11 @@ const LedgerEventRow = (props: LedgerEventRowProps): ReactNode => {
   return (
     <TableRow>
       <TableCell component="th" scope="row">
-        <Typography variant="button">
-          {props.event.action.type.replace("_", " ")}
-        </Typography>
+        <ThemeProvider theme={actionTheme}>
+          <Typography color="primary">
+            {_friendlyFormatActionType(props.event.action.type)}
```
hozzjss, line 150: I would put this value within a `useMemo`

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
```diff
@@ -50,6 +62,17 @@ type LedgerEventRowProps = {|
...
+  //Split words by _, capitalize them, and join them with " "
+  var words = type.split("_");
+  words.forEach((word, idx) => {
+    word = word.toLowerCase();
+    word = word.charAt(0).toUpperCase() + word.slice(1);
+    words[idx] = word;
+  });
+  return words.join(" ");
+}
+
```
hozzjss, line 75: I would put this in a `useCallback`

hozzjss (commented):
Great work, just some small suggestions

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
bensodenkamp: Is the result of this purely a difference in performance and an observation of performance best practices or are there other differences in the results?

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
bensodenkamp: Same question as above

bensodenkamp:
Test Plan:

in local sourcecred/sourcecred repository on main branch, from the root of the repository issue:

1. yarn build
2. yarn start
3. go to: http://localhost:8080/webpack-dev-server/ in a browser
5. Confirm that

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
hozzjss: Just a way to avoid memory clutter and unnecessary computations between re-renders not to recreate the function or recalculate the values especially if this is a list item where this would happen an n amount of times

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
hozzjss: Actually this is fine it’s outside the component

hozzjss: My bad

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
bensodenkamp: Updated.. I'm not sure what the right way to apply useMemo would have been within the content of Typography, but I did implement it in the function. It was easier for me to reason about what was being memoized and why.

packages/sourcecred/src/ui/components/LedgerViewer/LedgerEventRow.js
```diff
@@ -63,14 +63,21 @@ type LedgerEventRowProps = {|
...
-  return words.join(" ");
+  //Memoize computed label values. Small set of expected labels,
+  //potentially large number of times displayed. Best to save the values.
+
+  const label = useMemo(() => {
+    //Split words by _, capitalize them, and join them with " "
+    var words = type.split("_");
+    words.forEach((word, idx) => {
+      word = word.toLowerCase();
+      word = word.charAt(0).toUpperCase() + word.slice(1);
```
hozzjss, line 74: Almost there, use memo doesn't work outside a component this can be fixed by reverting the function like it was before and instead within the component at the top use it like this
```js
const type = props.event.action.type
const label = useMemo(() => _friendlyFormatActionType(type), [type])
```

and then evaluate the label inside JSX like this
```js
<Typography color="primary">
            {label}
 </Typography>
```

hozzjss (approved):
Great work!!

PR #3085, opened 2021-06-01
Implement Grain Integration functionality in the CLI

topocount:
# Details
This officially hooks up grain integrations into something accessible via the CLI!

I'm open to setting up a CLI flag in addition to allowing the invocation of 
grain integrations at the config level as well.

These changes also contain a semantic change for the grain integration
typing. The function exported by a grain integration is now named
`GrainIntegrationFunction`
The original `GrainIntegration` type which used to represent the
function is now as follows:
```js
type GrainIntegration = {|
  function: GrainIntegrationFunction,
  name: string
|}
```
    
# Merge Plan
Merge after #3079

# Test Plan
0. `yarn build`
1. from repo root: `cd
   packages/sourcecred/sharness/__snapshots__/test-instance`
2. `scdev grain -f`
3. open the new file in the output/grainIntegration and observe 2 lines
   of csv output. The first column containing payout addresses, the
   second column containing payout amounts in BigInt form.

# Next Steps
All we need to use it is a way to easily set payout addresses for
participants. This was done by hand in a NodeJS repl in order to
enable the above test case.


packages/sourcecred/src/cli/grain.js
```diff
@@ -45,6 +46,25 @@ const grainCommand: Command = async (args, std) => {
...
+        integrationCurrency,
+        false
+      );
+      if (result.grainIntegrationOutput)
+        instance.writeGrainIntegrationOutput(
+          result.grainIntegrationOutput,
+          distribution
+        );
+    });
+  }
```
blueridger, line 66: Shouldn't this go in the grain API instead of the grain CLI since it is business logic?

packages/sourcecred/src/api/instance/localInstance.js
```diff
@@ -200,6 +206,23 @@ export class LocalInstance implements Instance {
...
+  async writeGrainIntegrationOutput(
+    output: GrainIntegrationOutput,
+    distribution: Distribution
+  ): Promise<void> {
+    mkdirx(pathJoin(...GRAIN_INTEGRATION_DIRECTORY));
+    const credDateString = new Date(
+      distribution.credTimestamp
+      // utilise the `SE` datestring format since it appears like `YYYY-MM-DD`
+      // with numbers and is therefore easy to sort on and human-readable.
+    ).toLocaleDateString("en-SE");
```
blueridger, line 218: we have a util/timestamp.js module that has a toISO function that does this. Probably better to use it.

packages/sourcecred/src/cli/grain.js
topocount: good point. I'll separate the business logic from the I/O

packages/sourcecred/src/api/main/grain.js
```diff
@@ -37,3 +41,30 @@ export async function grain(input: GrainInput): Promise<GrainOutput> {
...
+        grainIntegration.function,
+        distribution,
+        integrationCurrency,
+        false
+      );
+      results.push(result);
+    });
+  }
+  return results;
+}
```
blueridger, line 70: Seems strange for this to return an array of objects that each contain the same ledger, and that in the CLI we encode the assumption that the ledger we passed in is going to be mutated and so we can ignore the ledgers that come out. What about a refactor that looks more like:
```suggestion
  if (integrationCurrency && grainIntegration) {
  let ledgerResult = ledger;
    distributions.forEach((distribution) => {
      const {ledger: newLedger, output, distributionCredTimestamp} = executeGrainIntegration(
        ledger,
        grainIntegration.function,
        distribution,
        integrationCurrency,
        false
      );
      results.push({output, distributionCredTimestamp});
      ledgerResult = newLedger;
    });
  }
  return {ledger: ledgerResult, results};
}
```

packages/sourcecred/src/cli/grain.js
```diff
@@ -45,6 +45,16 @@ const grainCommand: Command = async (args, std) => {
 
   const {distributions, ledger} = await grain(grainInput);
 
+  const grainIntegrationResults = executeGrainIntegrationsFromGrainInput(
+    grainInput,
+    ledger,
+    distributions
+  );
```
blueridger, line 52: This is related to my other comment.
```suggestion
  const {distributions, ledger: ledgerBeforeIntegrations} = await grain(grainInput);

  const {ledger, results: grainIntegrationResults} = executeGrainIntegrationsFromGrainInput(
    grainInput,
    ledgerBeforeIntegrations,
    distributions
  );
```

packages/sourcecred/src/api/main/grain.js
topocount: yeah good catch. We don't ever want to assume it's safe to mutate the same ledger reference.

packages/sourcecred/src/api/main/grain.js
```diff
@@ -37,3 +47,36 @@ export async function grain(input: GrainInput): Promise<GrainOutput> {
...
+        grainIntegration.function,
+        distribution,
+        integrationCurrency,
+        false
+      );
+      ledgerResult = result.ledger;
+      results.push(result);
+    });
+  }
+  return {ledger: ledgerResult, results};
```
blueridger, line 81: Still seems weird / too heavy that we are returning a bunch of ledgers in results right here. Is there a reason for it, or can we slim down and toss the extra references to garbage collection?

packages/sourcecred/src/api/main/grain.js
```diff
@@ -24,8 +25,15 @@ export type GrainOutput = {|
   +ledger: Ledger,
 |};
 
+// Similar to GrainIntegrationResult but excludes the ledger
+// since only the most recent ledger is relevant
+export type GrainIntegrationMultiResult = {|
```
blueridger, line 30: nit: Why multi? Makes me think it's going to be an array. Not sure I have a better suggestion, I'm all turned around on the namings in this epic lol. timeStampedGrainIntegrationOutput? idk. you can keep it.

packages/sourcecred/src/api/instance/instance.js
```diff
@@ -47,5 +47,7 @@ export interface Instance extends ReadOnlyInstance {
   writeLedger(ledger: Ledger): Promise<void>;
 
   /** writes grain integration output */
-  writeGrainIntegrationOutput(result: GrainIntegrationResult): Promise<void>;
+  writeGrainIntegrationOutput(
+    result: $Shape<GrainIntegrationMultiResult>
```
blueridger, line 51: why shape? seems unnecessary since the GrainIntegrationOutput attribute is optional already

blueridger (approved):

packages/sourcecred/src/api/instance/instance.js
topocount: I wanted to be unopinionated enough to accept the `GrainIntegrationResult`type.

PR #3086, opened 2021-06-01
Cache CredGrainView for improved UI load times

blueridger:
# Description
the UI doesn't need the whole credgraph, just a subset of that data represented as a CredGrainView. This moves the credGraph->credGrainView transformation to the API run in the github action so that the client side gets a latency improvement.

# Design choices
I chose not to use a file extension, because our zip library (pako) doesn't generate a file that corresponds to any file extension. The gzip file extension we have previously used is misleading because our files cannot be unzipped using gzip.

# Test plan
1. `yarn build` in sourcecred/sourcecred
2. In sourcecred/cred/gh-pages:
```
scdev credrank
yarn serve    # 5.3 second load time in browser
scdev serve    # 1.8 second load time in browser
```


hozzjss (approved):
Went through this and joyously tested it out, it's blazing fast, and definitely a low risk high impact feature that I'd release as soon as we can, really great work @blueridger !!! 🎉🎉🎉

PR #3088, opened 2021-06-03
localinstance extends readinstance (code dedup)

bensodenkamp:
Made localinstance extend readinstance, and removed duplicate code.

Test plan:

from test instance root (This should use the local instance for api interaction): 
scdev go


blueridger (approved):
🔥

PR #3094, opened 2021-07-19
Monorepo fix: put github templates in root

blueridger:
Issues templates wasn't working because it needs to be in root.


hozzjss (approved):
Amazing observation, approved!!

PR #3099, opened 2021-07-30
UI: Add identity id tooltip to Grain Accounts page

blueridger:
# Description
Sometimes you just want to look up someone's identity ID, and the easiest way before this was through the tooltips on the Ledger History page, which is a pain. This adds the same tooltip on the grain accounts page so it's less annoying to look up someone's ID.

# Test Plan
chrome-capture (11)


saintmedusa (approved):
Lgtm

Issue #2878, opened 2021-03-25
UI: Add Date filtering to LedgerView


KuraFire:
This ticket is quite similar to #2630, but covers filtering for Dates (via a range) in the Ledger instead of filtering by Events.

## Design Spec
Here is an updated interface & specifications for updates to the Ledger page:

![Ledger - new default](https://user-images.githubusercontent.com/33087/112197658-8c17f600-8bc9-11eb-991a-39d75313da66.png)

## Context Matters (Disclaimer)
The screenshot above shows changes which will be represented in at least three different tickets. This ticket only addresses Date filtering.

- #2870 addresses the changes in color and capitalization of the Event Types in the left column.
- #2630 covers filtering by Event types in the Ledger

There are two new filtering types being introduced to the Ledger: Filter Events, and Filter Date (select a range). Here is a close-up view:
![Ledger - new filters row](https://user-images.githubusercontent.com/33087/112198842-c3d36d80-8bca-11eb-8285-bc55126557c7.png)

## Filter by Date range: UI spec
- The two date pickers for this filter use the [standard Material UI date picker component](https://material-ui.com/components/pickers/#material-ui-pickers), and **not** the native date picker. _(Chrome made a **much** less accessible native picker that's awful to use.)_
- For mobile, let's test Material UI's default handling of this before exploring specifics. This can be a separate ticket in the future.

![Ledger - date picker](https://user-images.githubusercontent.com/33087/112199513-7a375280-8bcb-11eb-921e-1ca2edd484ce.png)

## Filter by Date range: Interaction Design spec
The filter works reductively:

- By default (initial page load), no dates are selected
- Selecting a Start Date (but no end date) filters as expected: from selected Start Date to Today
- Selecting an End Date (but no start date) filters as expected: from start of project to selected End Date
- Selected both Start _and_ End Date filters as expected: within that range, inclusively (i.e. both selected dates are included in the range)

Resetting the field once a Date is selected (either/both):
- a small "Clear ×" link-style `<button>` is added above the right edge of the date input field;
- activating this button sets the value back to `null` to clear the field, and updates the filtering of the ledger table accordingly;
- Clearing the field should **not** return the focus on the date input field

[...]

PR #3122, opened 2021-08-13
UI: Add Filtering to to Ledger View

m-aboelenein:
closes https://github.com/sourcecred/sourcecred/issues/2878


packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
```diff
@@ -82,7 +86,17 @@ export const LedgerViewer = ({
...
+    setStartDateFilter(date);
+  };
+  const handleEndDateChange = (date) => {
+    setEndDateFilter(date);
+  };
+
   const eventLog = useMemo(() => [...ledger.eventLog()], [ledger]);
+  const [startDateFilter, setStartDateFilter] = useState(new Date());
+  const [endDateFilter, setEndDateFilter] = useState(new Date());
+
```
hozzjss, line 99: Could you use the new Date placeholder outside the component so that it's not evaluated at every re-render? a single `new Date()`

packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
```diff
@@ -82,7 +86,17 @@ export const LedgerViewer = ({
     setAllocation(null);
   }, []);
 
+  const handleStartDateChange = (date) => {
+    setStartDateFilter(date);
+  };
+  const handleEndDateChange = (date) => {
+    setEndDateFilter(date);
+  };
+
```
hozzjss, line 95: Could you use `useCallback` here?

hozzjss (changes requested):
Great work dude, just a few comments

packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
m-aboelenein: sure that sounds reasonable however the initial state will be changed later to be empty this was just for testing

packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
m-aboelenein: will do

packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
hozzjss: sounds reasonable

packages/sourcecred/package.json
```diff
@@ -28,6 +30,7 @@
     "d3-shape": "^2.0.0",
     "d3-time": "^2.0.0",
     "d3-time-format": "^3.0.0",
+    "date-fns": "2.23.0",
```
blueridger, line 33: is this being used? I see you using `@date-io/date-fns` instead.

blueridger:
Way to knock out your first task!

Since this is adding new UI, would you add screenshots?

packages/sourcecred/package.json
m-aboelenein: installed it for handling the date formats in the filter, later on, will remove it if I don't use it

m-aboelenein: the `@date-io/date-fns` library xposes an abstraction interface over date-fns for the provider to consume the actual functionality exists in date-fns

m-aboelenein:
https://www.loom.com/share/f7aab08536a34b0b8c95c8f1f0ac976c

packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
```diff
@@ -95,12 +102,63 @@ export const LedgerViewer = ({
     }
   );
 
+  const handleChangeStartDate = useCallback((date) => {
+    setStartDateFilter(date);
+  }, []);
+
+  const handleChangeEndDate = useCallback((date) => {
+    setEndDateFilter(date);
+  }, []);
```
hozzjss, line 111: after thinking more about this, why not just pass the function as is, if it is for future additions to logic that can def be added later but for now, why not just pass `setStartDateFilter` and `setEndDateFilter` as is?

hozzjss (commented):
just one last thing

packages/sourcecred/src/ui/components/LedgerViewer/LedgerViewer.js
m-aboelenein: done

hozzjss (approved):
Awesome work!!!

PR #3146, opened 2021-08-19
Add ignoreActiveStatus option to Special Distributions

blueridger:
# Description
Before this change: Special Distributions can be added in the grain.json config but will fail if the recipient is not active (activated to receive grain).

After this change: An optional `ignoreActiveStatus` attribute can be added to a Special Distribution and set to true to allow that distribution to succeed whether or not the recipient is activated. (see test plan for examples)

# Design Decisions
- The special distribution now takes a credGrainView. This is aligned with a larger effort of deprecating allocationIdentities and processedIdentities data structures in favor of using the credGrainView as input for all policies.
- Ledger handling of DistributeGrain events has changed to allow distribution to inactive accounts when the policy has a truthy `ignoreActiveStatus` attribute. This will apply to all grain policies, so the door is open for other policies to implement such configuration in the future. I believe this is a reasonable change to the Ledger's assumptions since the policy data structure is stored in the LedgerLog and thus the condition can be checked and enforced during regeneration. One reason not to distribute to inactive participants is consent to income, however a Special Distribution is semantically unlikely to be sent to someone without their involvement.

# Justification
The Special Distribution type hasn't been publicly endorsed because it breaks "Grain comes from Cred." However, now that it is possible to include Special Distributions in the grain config, it may actually be a very nice solution to the question of "how does sourcecred work when some people are getting fixed salaries?" This is a reality of most organizations and probably will be for some time, and with good reason--it prioritizes meeting human needs, unlike our purely meritocratic paradigm.

The target use case for this PR: People with fixed salaries can be paid through special distributions instead of being paid outside of the sourcecred system, so that if they ever stop being salaried and want to earn normal grain, the Balanced policy will be aware of their previous pay and will not try to "catch them up" unless they actually were underpaid.

This is something the upcoming Product circle will likely dogfood.

# Test Plan
- regression tested the Special Distribution UI page at `/special-distribution`
- Tested `grain.json` changes on sourcecred/cred/gh-pages with `scdev grain -s -f`
### Regression test grain config with active account
Input:
```
{
  "allocationPolicies": [
    {
      "policyType": "SPECIAL",
      "memo": "test",
      "budget": 1000,
      "recipient": "6wyPOVWLmZ2b35eCmOrz1w"
    }
  ],
}

```
Output:
```
SPECIAL Policy
Budget 1,000.000g
Memo: test
Recepient: 6wyPOVWLmZ2b35eCmOrz1w
|          name          |    total    |     %     |
| ---------------------- | ----------- | --------- |
|         Thena          |  1,000.000  |  100.00%  |
```
### Test inactive account without ignore flag
input config:
```
{
  "allocationPolicies": [
    {
      "policyType": "SPECIAL",
      "memo": "test",
      "budget": 1000,
      "recipient": "6XoP0YuaULs3BWR6QaKmpg"
    }
  ],
}
```
Output:
```
Error: no active grain account for identity: 6XoP0YuaULs3BWR6QaKmpg
    at specialReceipts (webpack:///./src/core/ledger/policies/special.js?:18:281)
    at receipts (webpack:///./src/core/ledger/grainAllocation.js?:16:830)
    at computeAllocation (webpack:///./src/core/ledger/grainAllocation.js?:15:296)
```
### Test inactive with ignore flag
input config:
```
{
  "allocationPolicies": [
    {
      "policyType": "SPECIAL",
      "memo": "test",
      "budget": 1000,
      "recipient": "6XoP0YuaULs3BWR6QaKmpg",
      "ignoreActiveStatus": true
    }
  ],
}

```
Output:
```
SPECIAL Policy
Budget 1,000.000g
Memo: test
Recepient: 6XoP0YuaULs3BWR6QaKmpg
IgnoreActiveStatus: true
|          name          |    total    |     %     |
| ---------------------- | ----------- | --------- |
|     buffalonetwork     |  1,000.000  |  100.00%  |
```

### DIFF
--- a/packages/sourcecred/src/core/ledger/grainAllocation.js
+++ b/packages/sourcecred/src/core/ledger/grainAllocation.js
@@ -68,17 +68,15 @@ export function computeAllocation(
   });
 }
 
-/* This is a simplified case that should not require a credGrainView */
 export function computeAllocationSpecial(
   policy: AllocationPolicy,
-  identities: $ReadOnlyArray<AllocationIdentity>
+  credGrainView: CredGrainView
 ): Allocation {
   const validatedPolicy = _validatePolicy(policy);
-  const processedIdentities = processIdentities(identities);
   if (validatedPolicy.policyType === "SPECIAL") {
     return _validateAllocationBudget({
       policy,
-      receipts: specialReceipts(validatedPolicy, processedIdentities),
+      receipts: specialReceipts(validatedPolicy, credGrainView),
       id: randomUuid(),
     });
   } else {
@@ -121,7 +119,7 @@ function receipts(
     case "BALANCED":
       return balancedReceipts(policy, credGrainView, effectiveTimestamp);
     case "SPECIAL":
-      return specialReceipts(policy, identities);
+      return specialReceipts(policy, credGrainView);
     // istanbul ignore next: unreachable per Flow
     default:
       throw new Error(`Unknown policyType: ${(policy.policyType: empty)}`);
--- a/packages/sourcecred/src/core/ledger/grainAllocation.test.js
+++ b/packages/sourcecred/src/core/ledger/grainAllocation.test.js
@@ -441,16 +441,22 @@ describe("core/ledger/grainAllocation", () => {
         done();
       });
 
-      it("distributes the budget to the stated recipient", () => {
+      it("computeAllocation distributes the budget to the stated recipient", async () => {
         const i1 = aid(0, [1]);
+        const credGraph2 = await GraphUtil.credGraph2();
+        const ledger = ledgerWithActiveIdentities(id3, id4);
+        const credGrainView = CredGrainView.fromCredGraphAndLedger(
+          credGraph2,
+          ledger
+        );
         const policy = {
           policyType: "SPECIAL",
           budget: nng(100),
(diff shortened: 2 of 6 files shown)


decentralion (commented):
The PR says it's adding the `ignoreActiveStatus` option, but it doesn't describe what this option does or why it's needed to accomplish the stated goals. Please update the description so that as a reviewer, I can understand what's happening technically.

I'm not really watching GH notifications, so please ping on Discord when this is ready for re-review.

decentralion (changes requested):
The active/inactive distinction was added on the basis that receiving Grain can trigger legal/tax complications, and so participants cannot recieve Grain unless they've actively consented to that by marking themselves as active. This change would circumvent user consent and enable the project to pay people regardless of their wishes. I see it as misaligned with our values so it's a block from me. I have no objection to paying people through the special policy, but they should have marked themselves active first.

blueridger:
Discussed offline. @decentralion suggested the alternative of adding an "exclude" ID array field to the recent policy, since that covers the use case more accurately, and doesn't corrupt the invariant that "inactive == blocked from any grain activity"

blueridger:
Created issue for exclusion field: https://github.com/sourcecred/sourcecred/issues/3185

PR #3154, opened 2021-08-21
Add ConstructorPlugin for ExternalPlugin testing

blueridger:
# Description
Since `identityProposals` are used during the `graph` api, it would be useful to test run the `graph` api when developing an [ExternalPlugin](https://github.com/sourcecred/sourcecred/pull/3114). The `graph` api is already exposed in our library, the difficult bit is building a `GraphInput` structure which requires Plugin objects. This PR adds and exposes a ConstructorPlugin which can be used by 3rd party plugins to easily do such testing.

# Test Plan
`yarn build`
`node`
`sc = require("./dist/server/api.js").sourcecred`
```
> p = new sc.plugins.ConstructorPlugin({pluginId: "external/test"})
ConstructorPlugin {
  _weightedGraph: {
    graph: yt {
      _nodes: Map(0) {},
      _edges: Map(0) {},
      _incidentEdges: Map(0) {},
      _modificationCount: 0,
      _cachedOrder: [Object],
      _invariantsLastChecked: [Object]
    },
    weights: { nodeWeights: Map(0) {}, edgeWeights: Map(0) {} }
  },
  _identityProposals: [],
  _declaration: {
    name: 'test',
    nodePrefix: 'N\x00external\x00test\x00',
    edgePrefix: 'E\x00external\x00test\x00',
    nodeTypes: [ [Object], [Object] ],
    edgeTypes: [ [Object] ],
    userTypes: [ [Object] ]
  }
}
> p.declaration()
Promise {
  {
    name: 'test',
    nodePrefix: 'N\x00external\x00test\x00',
    edgePrefix: 'E\x00external\x00test\x00',
    nodeTypes: [ [Object], [Object] ],
    edgeTypes: [ [Object] ],
    userTypes: [ [Object] ]
  }
}
> p.identities()
Promise { [] }
```
```
> sc.api.graph.graph({plugins: [{plugin: p, directoryContext: null}], ledger: new sc.ledger.ledger.Ledger()}).then(r => console.log(r))
Promise { <pending> }
> {
  pluginGraphs: [ { pluginId: undefined, weightedGraph: [Object] } ],
  ledger: Ga {
    _ledgerEventLog: Yr { _items: [] },
    _nameToId: Map(0) {},
    _lowercaseNames: Set(0) {},
    _aliasAddressToIdentity: Map(0) {},
    _accounts: Map(0) {},
    _allocations: Map(0) {},
    _distributions: Map(0) {},
    _allocationsToDistributions: Map(0) {},
    _latestTimestamp: -Infinity,
    _lastDistributionTimestamp: null,
    _shouldTrackGrainIntegration: false,
    _grainIntegrationStatuses: Map(0) {}
  }
}
```


packages/sourcecred/src/api/main/graph.js
```diff
@@ -50,7 +50,7 @@ export async function graph(
   // Build graphs
   const pluginGraphs = [];
   for (const {plugin, directoryContext} of input.plugins) {
-    if (scope.includes(plugin.id)) {
+    if (!scope || scope.includes(plugin.id)) {
```
decentralion, line 53: when I see that scope is an optional array, I intuitively expect that if scope is unset, it will default to an empty value, i.e. an empty array. however, the semantics are that if scope is unset, then every plugin is implicitly in scope, which is counter-intuitive. can you please add a docstring explaining the semantics here, and what happens if scope is unset?

also, from the PR description it's not clear why this change was needed.

decentralion (commented):
What's the relationship between ConstructorPlugin and ExternalPlugin? The docs say that ConstructorPlugin allows testing ExternalPlugin but I don't see any example to that effect.

Also, re naming: the intention was for names to follow the `organization/plugin` pattern, as in `sourcecred/github` or `gitcoin/ethereum` or such. Using `external` as a catch-all organizatinon name undermines the intent here, which was that two separate organizations might have their own versions of (say) a discourse plugin, without creating conflicts.

PR #3159, opened 2021-08-23
Add Cache version suffix in circleCI

topocount:
This allows us to manually flush the build cache when needed.

test plan: The merkle Grain Integration needs the build cache cleared to
begin running successfully and this should do it. Unfortunately this can
really only be tested in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,9 @@ commands:
           keys:
             - v1-node<< parameters.node_version >>-dependencies-{{ checksum "yarn.lock" }}
             # fall back to using the latest cache if no exact match is found
-            - v1-node<< parameters.node_version >>-dependencies-
+            # added a versioning suffix to allow for manual cache clearing:
+            # https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache
+            - v1-node<< parameters.node_version >>-dependencies-v0
       - run: yarn install
       - save_cache:
           paths:


.circleci/config.yml
```diff
@@ -24,7 +24,9 @@ commands:
           keys:
             - v1-node<< parameters.node_version >>-dependencies-{{ checksum "yarn.lock" }}
             # fall back to using the latest cache if no exact match is found
-            - v1-node<< parameters.node_version >>-dependencies-
+            # added a versioning suffix to allow for manual cache clearing:
+            # https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache
+            - v1-node<< parameters.node_version >>-dependencies-v0
```
blueridger, line 29: This doesn't match the article linked, which uses an environment variable. Could you explain a little more how this is going to work?

topocount: I'm on the fence about using an environment variable to flush the cache, because it seems like an unnecessarily high-permissioned, centralized locale. By keeping it in the file, we can track cache clearing in a single place as a part of a change set.

Additionally, if we use an environment variable, this affects all branches. I think managing cache flushes on a per-branch basis is preferable, personally.

blueridger: OH I see, the other PR is pointing to this branch. Cool I get it now. Yeah that seems better.

blueridger (approved):

PR #3164, opened 2021-08-24
CredGrainView: Add advice to common error message

blueridger:
People get this error frequently and ask about it in tech support. The solution is usually just to rerun `yarn start` since all ledger accounts are given a graph node during `credrank`. I decided not to mention `yarn start` explicitly so that we don't have inconsistency if we ever change the default instance script commands.


hozzjss (approved):

blueridger:
approved on the condition that if anyone asks how to recalculate scores, then we'll consider adding an actual command to the advice

PR #3177, opened 2021-08-30
Prepare to publish initial csv grain integration

topocount:
> NOTICE
This also bumps `sourcecred` to v0.9.3 to minimize churn

# Details

In order to prevent the issue with v0.9.2 from occurring again, the release process
will be modified to test an experimental build from npm before tagging it as "latest",
which makes it the default version that is downloaded.

# Test plan:
This test plan relies on a global installation of lerna:
```sh
yarn global add lerna
```

1. `yarn clean && yarn && lerna run build` Create a fresh local build to
   be used with `scDev`
2. `yarn clean` remove all dependencies, we need to replicate a prod
   install next
3. `lerna bootstrap --no-hoist '*' -- --production` Install dependencies
   like they would be installed in an instance. the flow-mono.sh script
   execution should error. This is okay.
4. In a local sourcecred/cred instance, run  through the manual test
   script here: https://roamresearch.com/#/app/SourceCred/page/nHxYhbtca

All invocations with `scdev` should execute without error.


blueridger (approved):

blueridger (approved):

blueridger:
In retrospect, would've been nice to put v0.9.3 somewhere in the commit title...

PR #3179, opened 2021-08-30
Expose coerce helper for names in API

blueridger:
# Description
In 3rd-party plugin development, we enforce that identityProposal names only contain letters, numbers, and dashes. This forces 3rd party developers to clean their platform usernames to meet our requirement. Internally, we have a helper function to do this cleaning, so this PR exposes that helper function for 3rd party use.

# Test Plan
yarn build


hozzjss (approved):
thanks for the exposé

PR #3181, opened 2021-08-31
Transfer UI: make max a button and don't clear accounts after save

blueridger:
# Description
This PR includes 3 minor changes on the Transfer Grain page:
1. Makes the "max: XXX.XXg" field a button that fills the amount text field with the max
2. Pipes the current from and to accounts into the Autocomplete fields so that if we set them to null, the fields will actually clear. (tested before I implemented point 3)
3. Change the save-to-disk behavior to no longer clear the from and to fields. (note: the transfer grain button already does not clear the from and to fields). It seems a little nicer since often people want to reuse one of the accounts in multiple transfers. Would like Design's perspective though.

# Test Plan
## Before:
chrome-capture (12)
## After:
chrome-capture (13)

### DIFF
--- a/packages/sourcecred/src/ui/components/AccountSelector.js
+++ b/packages/sourcecred/src/ui/components/AccountSelector.js
@@ -9,6 +9,7 @@ type DropdownProps = {|
   +ledger: Ledger,
   +setCurrentAccount: (Account | null) => void,
   +placeholder?: string,
+  +currentAccount: Account | null,
 |};
 
 const useStyles = makeStyles({combobox: {margin: "0px 32px 16px"}});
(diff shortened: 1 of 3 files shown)


magwalk:
I have 2 small edits to make the thing a little more responsive-friendly. On my computer on screen sizes below 1482px wide, the `input` was getting a little narrow and cutting off the "amount" placeholder, and the button text was breaking onto two lines which made the arrow not so much an arrow anymore:

Screen Shot 2021-08-31 at 11 49 10 AM

I have the CSS changes ready locally and would be happy to push them up if you'd like.

topocount (approved):
This is approved pending the removal of the "Changes not saved to ledger" notice upon a successful transfer and save

topocount:
perhaps the save should clear the inputs?

magwalk:
> perhaps the save should clear the inputs? @magwalk?

Yep, ideally hitting "Save ledger to disk" would revert the form to its original state with the "From" and "To" dropdowns empty. 

Also this is a separate issue but IMO the "Changes not saved to ledger" shows up a little too early because it's not actionable information and therefore adds an unnecessary distraction. It might make sense to have it show up after the user hits "Transfer grain". I'd be happy to make a separate issue for making this UX a bit smoother.

blueridger:
Ok I reverted back to ALL fields getting cleared after save ledger to disk. And now "Changes not saved to ledger" only shows after you make a transfer, and not after you hit save changes.

Issue #3139, opened 2021-08-18
Merkle Tree gas-efficient Grain Integration


blueridger:

topocount:
# Current Merge Plan

1. #3192 This is an update to core code and is mostly reviewed.
2. #3271 This is an update to the grain integration interface and needs a fresh review
3. #3148 This is basically ready to merge, but I'm putting it behind the code closer to core

PR #3192, opened 2021-09-21
Feature: Disable balance accounting in the ledger

topocount:
# Description

When utilizing an on chain grain integration, tracking account balances
in the SourceCred ledger makes little sense. These changes disable
balance accounting, while preserving the functionality key to
SourceCred, which is tracking distributions across identities. While
balances are no longer tracked, paid totals will continue to be. This
feature, combined with Distribution execution tracking, narrows the
scope of the ledgers focus to the state of distributions, and whether or
not they have been executed outside of SourceCred.

When account balance tracking is disabled, the following changes are
made to ledger behavior:
- all grain balances are zeroed out in mutable accounts tracked by the
  ledger
- grain transfers are disabled
- The ledger stores the currently configured `integrationCurrency` at the instant
  accounting is disabled as `externalCurrency`. This allows for the configuration
  to be historically tracked in case the configuration is accidentally changed, and
  allows for more predictable behavior in the ledger in this mode. If the `integrationCurrency`
  needs to be updated, the `disableAccounting` method can just be re-run with the
  new currency to update the configuration within the ledger.
- distributed grain is not added to a user's balance. the paid total is
  updated, however. Undistributed balances are tracked as a whole through
  the distribution tracking functionality added earlier.
- identities without a relevant payout address will be deactivated
- identities that have no relevant payout addresses configured cannot be
  activated. relevant means the external currency configured in the
  ledger's state.

Now that all integration functionality has settled in the ledger, the `executeGrainIntegration`
interface can be simplified and cleaned up. No information has been
lost; Config details can now simply be queried from the ledger directly.

Config inconsistencies are now tracked in the main/grain.js API file.
The main api will be tracking changes to the external currency config and
update the ledger dynamically before any integrations run. This seemed like
the best compromise between "magic" and usability. I think it's important that
the ledger tracks all integration currency changes for the purpose of transparency,
but also using a config file seems to be the best input method for now. This
tracking can be deprecated when we migrate to a webUI for config management.

# Test Plan

Unit tests included

# Merge Plan
[merge plan can be found here](https://github.com/sourcecred/sourcecred/issues/3139#issuecomment-978073574)
# Future Work

~~The current entry point is integration currency config in the `grain.json` config file. However, I'd really like to set up a CLI dialog for setting this and other integration config in the near-term. This is a hard requirement for the smart contract deployment for the merkle distribution integration, so I might as well include these few extra dialogs as a part of that story.~~

Update: I ended up just creating config entries in config/grain.json that are checked against the ledger every time grain distribution runs via the CLI.

### DIFF
--- a/packages/sourcecred/src/api/grainConfig.js
+++ b/packages/sourcecred/src/api/grainConfig.js
@@ -14,10 +14,9 @@ import {
   parser as bundledGrainIntegrationParser,
 } from "./bundledGrainIntegrations";
 
-// NOTE: This type is deprecated since v0.10.0 and should not be modified
-// further
 export type RawGrainConfig = {|
   +allocationPolicies: $ReadOnlyArray<AllocationConfig>,
+  +accountingEnabled?: boolean,
   +maxSimultaneousDistributions?: number,
   +sinkIdentity?: Name,
   +processDistributions?: boolean,
@@ -41,6 +40,7 @@ export const rawParser: C.Parser<RawGrainConfig> = C.object(
     maxSimultaneousDistributions: C.number,
     sinkIdentity: nameParser,
     processDistributions: C.boolean,
+    accountingEnabled: C.boolean,
     integration: bundledGrainIntegrationParser,
   }
 );
@@ -67,3 +67,5 @@ export const parser: C.Parser<GrainConfig> = C.fmap(
     accountingEnabled: NullUtil.orElse(config.accountingEnabled, true),
   })
 );
+
+// const allocationToPolicyParser<Allocation
--- a/packages/sourcecred/src/api/grainConfig.js
+++ b/packages/sourcecred/src/api/grainConfig.js
@@ -9,10 +9,8 @@ import {
   allocationConfigParser,
 } from "../core/ledger/policies";
 import {type Name, parser as nameParser} from "../core/identity/name";
-import {
-  type GrainIntegration,
-  parser as bundledGrainIntegrationParser,
-} from "./bundledGrainIntegrations";
+import type {GrainIntegration} from "../core/ledger/grainIntegration";
+import {parser as bundledGrainIntegrationParser} from "./bundledGrainIntegrations";
 
 export type RawGrainConfig = {|
   +allocationPolicies: $ReadOnlyArray<AllocationConfig>,
--- a/packages/sourcecred/src/api/grainConfig.js
+++ b/packages/sourcecred/src/api/grainConfig.js
@@ -14,6 +14,8 @@ import {
   parser as bundledGrainIntegrationParser,
 } from "./bundledGrainIntegrations";
 
+// NOTE: This type is deprecated since v0.10.0 and should not be modified
+// further
 export type RawGrainConfig = {|
   +allocationPolicies: $ReadOnlyArray<AllocationConfig>,
   +maxSimultaneousDistributions?: number,
@@ -25,6 +27,7 @@ export type RawGrainConfig = {|
 export type GrainConfig = {|
   +allocationPolicies: $ReadOnlyArray<AllocationPolicy>,
   +maxSimultaneousDistributions: number,
+  +accountingEnabled: boolean,
   +sinkIdentity?: Name,
   +processDistributions?: boolean,
   +integration?: GrainIntegration,
@@ -51,6 +54,7 @@ export const parser: C.Parser<GrainConfig> = C.fmap(
       maxSimultaneousDistributions: C.number,
       sinkIdentity: nameParser,
       processDistributions: C.boolean,
+      accountingEnabled: C.boolean,
       integration: bundledGrainIntegrationParser,
     }
   ),
@@ -60,5 +64,6 @@ export const parser: C.Parser<GrainConfig> = C.fmap(
       config.maxSimultaneousDistributions,
       Infinity
     ),
+    accountingEnabled: NullUtil.orElse(config.accountingEnabled, true),
   })
 );
--- a/packages/sourcecred/src/api/main/grain.js
+++ b/packages/sourcecred/src/api/main/grain.js
@@ -4,6 +4,7 @@ import {CredGraph} from "../../core/credrank/credGraph";
 import {Ledger} from "../../core/ledger/ledger";
 import type {CurrencyDetails} from "../currencyConfig";
 import {type GrainConfig} from "../grainConfig";
+import {type Currency as IntegrationCurrency} from "../../core/ledger/currency";
 import type {Distribution} from "../../core/ledger/distribution";
 import {applyDistributions} from "../../core/ledger/applyDistributions";
 import {type TimestampMs} from "../../util/timestamp";
@@ -11,7 +12,6 @@ import {
   executeGrainIntegration,
   type GrainIntegrationOutput,
 } from "../../core/ledger/grainIntegration";
-import deepEqual from "lodash.isequal";
 
 export type GrainInput = {|
   +credGraph: CredGraph,
@@ -68,19 +68,7 @@ export function executeGrainIntegrationsFromGrainInput(
   const integrationCurrency = grainInput.currencyDetails.integrationCurrency;
   const grainIntegration = grainInput.grainConfig.integration;
   const results = [];
-  if (!ledger.accounting().enabled) {
-    const ledgerCurrency = ledger.accounting().currency;
-    if (!deepEqual(ledgerCurrency, integrationCurrency)) {
-      integrationCurrency
-        ? ledger.setExternalCurrency(
-            integrationCurrency.chainId,
-            integrationCurrency.tokenAddress
-              ? integrationCurrency.tokenAddress
-              : undefined
-          )
-        : ledger.removeExternalCurrency();
-    }
-  }
+  ledger = configureLedger(ledger, grainInput);
   // track the latest ledger in the for-loop for the purposes of returning it
   // at the top level, observing that any function may deep-copy
   // the ledger (thus creating a new reference we'll need to track) and also
@@ -99,4 +87,53 @@ export function executeGrainIntegrationsFromGrainInput(
     });
   }
   return {ledger: ledgerResult, results};
+}
+
+function configureLedger(ledger: Ledger, input: GrainInput): Ledger {
+  const {grainConfig} = input;
+  ledger = configureLedgerAccounting(ledger, grainConfig.accountingEnabled);
+  ledger = configureDistributionProcessing(
+    ledger,
+    grainConfig.processDistributions
+  );
+  ledger = configureIntegrationCurrency(
+    ledger,
+    input.currencyDetails.integrationCurrency
+  );
+  return ledger;
+}
+
+function configureLedgerAccounting(
+  ledger: Ledger,
+  accountingEnabled: boolean
+): Ledger {
+  accountingEnabled ? ledger.enableAccounting() : ledger.disableAccounting();
+  return ledger;
+}
+
+function configureDistributionProcessing(
+  ledger: Ledger,
+  processDistributions: ?boolean
+): Ledger {
+  processDistributions
+    ? ledger.enableIntegrationTracking
+    : ledger.disableIntegrationTracking;
+  return ledger;
+}
+
+function configureIntegrationCurrency(
+  ledger: Ledger,
+  integrationCurrency: ?IntegrationCurrency
+): Ledger {
+  if (!ledger.accounting().enabled) {
+    integrationCurrency
+      ? ledger.setExternalCurrency(
+          integrationCurrency.chainId,
+          integrationCurrency.tokenAddress
+            ? integrationCurrency.tokenAddress
+            : undefined
+        )
+      : ledger.removeExternalCurrency();
+  }
+  return ledger;
 }
--- a/packages/sourcecred/src/api/main/grain.js
+++ b/packages/sourcecred/src/api/main/grain.js
@@ -11,6 +11,7 @@ import {
   executeGrainIntegration,
   type GrainIntegrationOutput,
 } from "../../core/ledger/grainIntegration";
+import deepEqual from "lodash.isequal";
 
 export type GrainInput = {|
   +credGraph: CredGraph,
@@ -67,6 +68,19 @@ export function executeGrainIntegrationsFromGrainInput(
   const integrationCurrency = grainInput.currencyDetails.integrationCurrency;
   const grainIntegration = grainInput.grainConfig.integration;
   const results = [];
+  if (!ledger.accounting().enabled) {
+    const ledgerCurrency = ledger.accounting().currency;
+    if (!deepEqual(ledgerCurrency, integrationCurrency)) {
+      integrationCurrency
+        ? ledger.setExternalCurrency(
+            integrationCurrency.chainId,
+            integrationCurrency.tokenAddress
+              ? integrationCurrency.tokenAddress
+              : undefined
+          )
+        : ledger.removeExternalCurrency();
+    }
+  }
   // track the latest ledger in the for-loop for the purposes of returning it
   // at the top level, observing that any function may deep-copy
   // the ledger (thus creating a new reference we'll need to track) and also
@@ -77,9 +91,7 @@ export function executeGrainIntegrationsFromGrainInput(
       const result = executeGrainIntegration(
         ledgerResult,
         grainIntegration.function,
-        distribution,
-        integrationCurrency,
-        false
+        distribution
       );
       const {output, distributionCredTimestamp, ledger: nextLedger} = result;
       ledgerResult = nextLedger;
--- a/packages/sourcecred/src/api/main/grain.js
+++ b/packages/sourcecred/src/api/main/grain.js
@@ -2,11 +2,11 @@
 
 import {CredGraph} from "../../core/credrank/credGraph";
 import {Ledger} from "../../core/ledger/ledger";
-import type {CurrencyDetails} from "../currencyConfig";
-import {type GrainConfig} from "../grainConfig";
-import {type Currency as IntegrationCurrency} from "../../core/ledger/currency";
-import type {Distribution} from "../../core/ledger/distribution";
 import {applyDistributions} from "../../core/ledger/applyDistributions";
+import type {CurrencyDetails} from "../currencyConfig";
+import type {GrainConfig} from "../grainConfig";
+import type {Currency as IntegrationCurrency} from "../../core/ledger/currency";
+import type {Distribution} from "../../core/ledger/distribution";
 import type {TimestampMs} from "../../util/timestamp";
 import {
   executeGrainIntegration,
@@ -30,6 +30,7 @@ export type GrainOutput = {|
 // since only the most recent ledger is relevant
 export type GrainIntegrationMultiResult = {|
   output?: GrainIntegrationOutput,
+  configUpdate: Object,
   distributionCredTimestamp: TimestampMs,
 |};
 
@@ -62,11 +63,11 @@ export async function grain(input: GrainInput): Promise<GrainOutput> {
  * Marshall grainInput from a Grain Configuration file for use with
  * executeGrainIntegration function
  */
-export function executeGrainIntegrationsFromGrainInput(
+export async function executeGrainIntegrationsFromGrainInput(
   grainInput: GrainInput,
   ledger: Ledger,
   distributions: $ReadOnlyArray<Distribution>
-): GrainIntegrationResults {
+): Promise<GrainIntegrationResults> {
   const integrationCurrency = grainInput.currencyDetails.integrationCurrency;
   const grainIntegration = grainInput.grainConfig.integration;
   const results = [];
@@ -76,16 +77,21 @@ export function executeGrainIntegrationsFromGrainInput(
   // supporting the case where the distributions parameter is an empty array
   let ledgerResult = ledger;
   if (integrationCurrency && grainIntegration) {
-    distributions.forEach((distribution) => {
-      const result = executeGrainIntegration(
+    for (const distribution of distributions) {
+      const result = await executeGrainIntegration(
         ledgerResult,
-        grainIntegration.function,
+        grainIntegration,
         distribution
       );
-      const {output, distributionCredTimestamp, ledger: nextLedger} = result;
+      const {
+        output,
+        distributionCredTimestamp,
+        ledger: nextLedger,
+        configUpdate,
+      } = result;
       ledgerResult = nextLedger;
-      results.push({output, distributionCredTimestamp});
-    });
+      results.push({output, distributionCredTimestamp, configUpdate});
+    }
   }
   return {ledger: ledgerResult, results};
 }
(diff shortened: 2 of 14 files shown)


packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +869,153 @@ export class Ledger {
...
+    }
+  }
+
+  /**
+   * Disabling accounting has the following effects on the ledger:
+   * - disables grain transfers
+   * - zeroes out all balances
+   * - deactivates accounts that don't have a PayoutAddress set for the
+   *   current instance configuration, and prevents them from activating until
+   *   they do.
```
blueridger, line 944: This belongs somewhere else or should be removed. It doesn't help me understand setExternalCurrency

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +869,153 @@ export class Ledger {
...
+   *
+   *   This function can be called while accounting is disabled to update
+   *   the ledger's external currency configuration. Note that this will also
+   *   deactivate identities missing a payout address that matches the new
+   *   configuration.
```
blueridger, line 949: \+ prevent activate of identities missing matching address...
\+ disables accounting if it is not already disabled

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +869,153 @@ export class Ledger {
...
+   */
+  setExternalCurrency(chainId: ChainId, tokenAddress?: EthAddress): Ledger {
+    // ensure accounting isn't already disabled
+    const currency = buildCurrency(chainId, tokenAddress);
```
blueridger, line 953: Unclear how this comment relates to this statement. Swap order?

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +869,153 @@ export class Ledger {
...
+        currency,
+      });
+
+      if (!this._balanceAccountingEnabled) {
+        // Deactivate accounts if they don't have an address stored for
+        // the newly configured currency.
+        this._createAndProcessEvent({
+          type: "DISABLE_ACCOUNTING",
+        });
+      }
```
blueridger, line 967: This should go in _setExternalCurrency

Non-underscored: utility wrapper for a single action (absolutely no mutation beyond a single _createAndProcessEvent, virtually no business logic, maybe small efficiency no-op logic)
Underscored: strongly coded and semantically whole action (self-sufficient business logic and validation)

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +869,153 @@ export class Ledger {
...
+  _setExternalCurrency({currency}: SetExternalCurrency) {
+    this._externalCurrency = currency;
+  }
+
+  removeExternalCurrency(): Ledger {
+    if (!this._balanceAccountingEnabled) {
+      throw new Error(
+        "Cannot remove the external currency while Accounting is Disabled"
+      );
+    }
```
blueridger, line 983: Belongs in _removeExternalCurrency

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -1137,6 +1323,42 @@ const executeDistributionParser: C.Parser<MarkDistributionExecuted> = C.object({
...
+const enableExternalCurrencyParser: C.Parser<SetExternalCurrency> = C.object({
+  type: C.exactly(["SET_EXTERNAL_CURRENCY"]),
+  currency: currencyParser,
+});
+
+type RemoveExternalCurrency = {|
+  +type: "REMOVE_EXTERNAL_CURRENCY",
+|};
+
+const disableExternalCurrencyParser: C.Parser<RemoveExternalCurrency> = C.object(
```
blueridger, line 1356:
```suggestion
const setExternalCurrencyParser: C.Parser<SetExternalCurrency> = C.object({
  type: C.exactly(["SET_EXTERNAL_CURRENCY"]),
  currency: currencyParser,
});

type RemoveExternalCurrency = {|
  +type: "REMOVE_EXTERNAL_CURRENCY",
|};

const removeExternalCurrencyParser: C.Parser<RemoveExternalCurrency> = C.object(
```

packages/sourcecred/src/api/main/grain.js
```diff
@@ -67,6 +68,19 @@ export function executeGrainIntegrationsFromGrainInput(
   const integrationCurrency = grainInput.currencyDetails.integrationCurrency;
   const grainIntegration = grainInput.grainConfig.integration;
   const results = [];
+  if (!ledger.accounting().enabled) {
+    const ledgerCurrency = ledger.accounting().currency;
+    if (!deepEqual(ledgerCurrency, integrationCurrency)) {
```
blueridger, line 73: nit: potentially a little redundant since the ledger has no-op logic.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -518,7 +531,13 @@ export class Ledger {
...
       throw new Error(`identity ${id} not found`);
     }
-    const {active} = this.account(id);
+    const {active, payoutAddresses} = this.account(id);
+    if (this._externalCurrency) {
+      const currencyKey = getCurrencyKey(this._externalCurrency);
+      if (!payoutAddresses.get(currencyKey)) {
+        throw new Error("Cannot activate user: No payout address set");
+      }
+    }
```
blueridger, line 540: Would you put this in _activate?

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -518,7 +531,13 @@ export class Ledger {
     if (!this._accounts.has(id)) {
       throw new Error(`identity ${id} not found`);
     }
-    const {active} = this.account(id);
+    const {active, payoutAddresses} = this.account(id);
+    if (this._externalCurrency) {
```
blueridger, line 535: It seems like there is a possible path where you can disableAccounting while still having an externalCurrency defined. This seems undesireable for this particular line of code. It seems to me like you want tight coupling between disabling accounting and having an external currency. If this is true, I'm wondering if you should maybe combined the actions together, specifically merge disable/enableAccounting into set/removeExternalCurrency. If there is a reason to have them separate, please elaborate.

blueridger (commented):
I'm a little confused about what the entry point for this is going to be. Can you add some insight to the description? Like is there going to be a UI button that calls setExternalCurrency? What entry points will there be for disableAccounting?

I have some pretty structural comments in this review, so I will review unit tests once those are settle.

packages/sourcecred/src/core/ledger/ledger.js
topocount: @blueridger I originally had a tight coupling, but realized, that an external currency might be desirable for reporting or other reasons. The external currency is needed to find the correct payout address for any distribution. The upshot of this is that even something as simple as the CSV integration depends upon the external currency being set.

I also intentionally am allowing for our extant use case (the one we've been running up until this point in  theSourceCred community) where the ledger serves as an off-chain tracker for IOUs (accounting enabled) while allowing for payouts onchain that can then be processed in the ledger automatically. This basically means that Seth's/the multisig's current workflow would be simplified because the ledger updates would be automated if we decided to conitnue to payout in USDC. This story basically is ideal for smaller communities that want to pay folks in crypto but don't yet want to handle the overhead of deploying a smart contract and/or setting up their own token.

topocount:
> I'm a little confused about what the entry point for this is going to be. Can you add some insight to the description? Like is there going to be a UI button that calls setExternalCurrency? What entry points will there be for disableAccounting?

@blueridger thanks for requesting this. Just added the following to the description:

he current entry point is integration currency config in the grain.json config file. However, I'd really like to set up a CLI dialog for setting this and other integration config in the near-term. This is a hard requirement for the smart contract deployment for the merkle distribution integration, so I might as well include these few extra dialogs as a part of that story.

packages/sourcecred/src/api/main/grain.js
```diff
@@ -67,6 +68,19 @@ export function executeGrainIntegrationsFromGrainInput(
...
+      integrationCurrency
+        ? ledger.setExternalCurrency(
+            integrationCurrency.chainId,
+            integrationCurrency.tokenAddress
+              ? integrationCurrency.tokenAddress
+              : undefined
+          )
+        : ledger.removeExternalCurrency();
```
topocount, line 81: I realized this evening that I need to do this for `enable/disableAccounting` and `enable/disableDistributionTracking` as well.

packages/sourcecred/src/core/ledger/ledger.js
topocount: @blueridger I'm not sure I understand why. We only need to check this during the initial activation attempt. Thereafter, can't we rest assured that the prior state leading up to the activation will always be the same? Isn't that kinda the point of this data structure?

packages/sourcecred/src/core/ledger/ledger.js
topocount: @blueridger Good catch! I've now completely decoupled external currency config settings from accounting settings. Sorry, I think I bit off a bit more than I should have for a single PR in this refactor and missed a bunch of details. I appreciate your patience on this review pass.

packages/sourcecred/src/api/grainConfig.js
```diff
@@ -14,6 +14,8 @@ import {
   parser as bundledGrainIntegrationParser,
 } from "./bundledGrainIntegrations";
 
+// NOTE: This type is deprecated since v0.10.0 and should not be modified
```
blueridger, line 17: Didn't I introduce this in 0.10.0 for future use? You should add accountingEnabled to this as well and remove the comment.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -61,12 +63,21 @@ type AllocationReceipt = {|
...
+  +trackDistributions: boolean,
+  +currency: ?Currency,
+|};
 /**
  * Every Identity in the ledger has an Account.
  */
 type MutableAccount = {|
   identity: Identity,
-  // The current Grain balance of this account
+  // The current Grain balance of this account. All balances are zero when
```
blueridger, line 79: Reminder that we should start thinking about using and substituting /** */ for the docs, including attributes I believe.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -646,6 +667,9 @@ export class Ledger {
     return this;
   }
   _transferGrain({from, to, amount}: TransferGrain) {
+    if (!this._balanceAccountingEnabled) {
+      throw new Error("Transfers are disabled when Accounting is is disabled");
```
blueridger, line 671: typo

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -769,6 +793,55 @@ export class Ledger {
...
+  /**
+   * Disabling accounting has the following effects on the ledger:
+   * - disables grain transfers
+   * - zeroes out all balances
+   * - deactivates accounts that don't have a PayoutAddress set for the
+   *   current instance configuration, and prevents them from activating until
+   *   they do.
+   */
+  disableAccounting(): Ledger {
+    // ensure accounting isn't already disabled
```
blueridger, line 825: redundant comment, the code is plenty readable on its own + already have a no-op comment below

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -769,6 +793,55 @@ export class Ledger {
...
+    }
+    // no-op if already disabled
+    return this;
+  }
+
+  _disableAccounting(_: DisableAccounting) {
+    // ensure an external currency is set
+    this.externalCurrency();
+
+    this._balanceAccountingEnabled = false;
```
blueridger, line 839: This should go below the mutations comment

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +869,153 @@ export class Ledger {
     return this._grainIntegrationStatuses.get(id);
   }
 
+  /**
```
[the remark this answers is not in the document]
blueridger: Unaddressed

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +921,80 @@ export class Ledger {
...
+   *   This function can be called while accounting is disabled to update
+   *   the ledger's external currency configuration. Note that this will also
+   *   deactivate identities missing a payout address that matches the new
+   *   configuration when accounting is disabled.
```
blueridger, line 928: Would be good to add the behavior of this when accounting is enable. Throws? No effect?

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +921,80 @@ export class Ledger {
...
+
+  /**
+   * Returns the state of the ledger's accounting configuration. If accounting
+   * is enabled, currency will be undefined. if accounting is disabled, a
+   * Currency will be set.
+   */
+  accounting(): AccountingStatus {
+    return {
+      enabled: this._balanceAccountingEnabled,
+      trackDistributions: this._shouldTrackGrainIntegration,
```
blueridger, line 984: nit: I think something about the naming is throwing me off here. It always tracks distributions, right? It's more about tracking the distribution's integration status.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -847,6 +921,80 @@ export class Ledger {
...
+      currency: this._externalCurrency,
+    };
+  }
+
+  /**
+   * Expectably returns the currency from the
+   * externalCurrency prop. It can only be invoked when a currency is set.
```
blueridger, line 991: It can be invoked either way.
```suggestion
   * externalCurrency prop. Throws if a currency is not set.
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -976,7 +1136,28 @@ export class Ledger {
...
-    account.balance = G.add(params.grainReceipt.amount, account.balance);
+    if (this._balanceAccountingEnabled)
+      account.balance = G.add(params.grainReceipt.amount, account.balance);
+  }
+
+  /**
+   * Helper method for deactivating accounts that don't have a payout address
+   * set for the configured `currencyKey`
+   */
+  _deactivateAccounts(): void {
```
blueridger, line 1147:
```suggestion
  _deactivateAccountsWithoutPayoutAddress(): void {
```

blueridger (commented):
For best clarity, I recommend that you add a consistent comment structure for all 4 of the new actions (both enable/set and disable/remove) of the form:
```
When <the other> is enabled: effect/behavior
When <the other> is disabled: effect/behavior
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -770,7 +796,70 @@ export class Ledger {
...
+   *    created since tracking was enabled. Their statuses can be queried using
+   *    `isGrainIntegrationExecuted`.
+   *
+   * Distribution Tracking can be enabled and disabled at any time.
+   * Some integrations might error if tracking is not enabled, to ensure
+   * their status is tracked in the ledger. This allows for the decoupling of
+   * (1) calculating a grain distribution and then (2) actually distributing
+   * it via the grain integration.
+   *
+   * No-ops if integration tracking is already enabled
```
blueridger, line 862: I like this being in the docstring instead of inline.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -770,7 +796,70 @@ export class Ledger {
...
+    // ensure an external currency is set
+    this.externalCurrency();
+
+    // Warning! Mutations below
+    this._balanceAccountingEnabled = false;
+    this._eraseLedgerBalances();
+  }
+
+  /**
+   * Enabling Distribution Tracking has the following effects:
```
blueridger, line 849:
```suggestion
   * Enabling Integration Tracking has the following effects:
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -770,7 +796,70 @@ export class Ledger {
...
+   *  - Subsequent grain distributions are left as unexecuted until
+   *    `markDistributionAsExecuted` is called on it.
+   *  - The `trackedDistributions` getter will return an array of distributionIds
+   *    created since tracking was enabled. Their statuses can be queried using
+   *    `isGrainIntegrationExecuted`.
+   *
+   * Distribution Tracking can be enabled and disabled at any time.
```
blueridger, line 856:
```suggestion
   * Integration Tracking can be enabled and disabled at any time.
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -781,8 +870,16 @@ export class Ledger {
     return this;
   }
 
+  _enableGrainIntegrationTracking(_: EnableGrainIntegration) {
```
blueridger, line 873: Breaks the pattern of the public and private function names matching
```suggestion
  _enableIntegrationTracking(_: EnableGrainIntegration) {
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -793,10 +890,6 @@ export class Ledger {
     return this;
   }
 
-  _enableGrainIntegrationTracking(_: EnableGrainIntegration) {
-    this._shouldTrackGrainIntegration = true;
-  }
-
   _disableGrainIntegrationTracking(_: DisableGrainIntegration) {
```
blueridger, line 893:
```suggestion
  _disableIntegrationTracking(_: DisableGrainIntegration) {
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -819,23 +920,99 @@ export class Ledger {
...
-      type: "MARK_DISTRIBUTION_EXECUTED",
-      id,
-    });
+    this._grainIntegrationStatuses.set(id, true);
+  }
+
+  /**
+   *   Setting a currency is a prerequisite to executing a grain integration.
+   *
+   *   When accounting is enabled, this function will set an external currency
```
blueridger, line 929: nice

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -819,23 +920,99 @@ export class Ledger {
...
+  setExternalCurrency(chainId: ChainId, tokenAddress?: EthAddress): Ledger {
+    const currency = buildCurrency(chainId, tokenAddress);
+
+    if (!deepEqual(currency, this._externalCurrency)) {
+      this._createAndProcessEvent({
+        type: "SET_EXTERNAL_CURRENCY",
+        currency,
+      });
+      // Deactivate accounts if they don't have an address stored for
+      // the newly configured currency.
```
blueridger, line 946: nit: the code is sufficiently readable (yay verbose function naming!) so this comment seems redundant.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -819,23 +920,99 @@ export class Ledger {
...
+      if (!this._balanceAccountingEnabled) {
+        this._deactivateAccountsWithoutPayoutAddress();
+      }
+    }
+
+    // no-op if already disabled
```
blueridger, line 952: copy pasta. Move to docstring and correct it to something like "if the currency params match the existing currency configuration"

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -819,23 +920,99 @@ export class Ledger {
...
     return this;
   }
-  _markDistributionExecuted({id}: MarkDistributionExecuted) {
-    this._grainIntegrationStatuses.set(id, true);
+
+  _setExternalCurrency({currency}: SetExternalCurrency) {
+    this._externalCurrency = currency;
```
blueridger, line 957: This is a mutation, should go below the mutation line.

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -819,23 +920,99 @@ export class Ledger {
...
+    if (!deepEqual(currency, this._externalCurrency)) {
+      this._createAndProcessEvent({
+        type: "SET_EXTERNAL_CURRENCY",
+        currency,
+      });
+      // Deactivate accounts if they don't have an address stored for
+      // the newly configured currency.
+      if (!this._balanceAccountingEnabled) {
+        this._deactivateAccountsWithoutPayoutAddress();
+      }
```
blueridger, line 949: This should go in the private method under the mutations line, right?

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -819,23 +920,99 @@ export class Ledger {
...
+      currency: this._externalCurrency,
+    };
   }
 
   /**
-   * Is cleared each time `disableIntegrationTracking` is called
+   *
+   *
+   * this._grainIntegrationStatuses Is cleared each time
+   * `disableIntegrationTracking` is called.
```
blueridger, line 1001:
```suggestion
   * When integration tracking is disabled, returns...
   * When integration tracking is enabled, returns...
```

packages/sourcecred/src/core/ledger/ledger.js
```diff
@@ -976,7 +1165,31 @@ export class Ledger {
...
+      if (!payoutAddress) this.deactivate(account.identity.id);
+    }
+  }
+
+  _eraseLedgerBalances(): void {
+    const accountsIterator = this._accounts.values();
+    for (const account of accountsIterator) {
+      // zero out all balances, so there is no confusion about them.
+      // By zeroing them out, they cannot be compared to external balances
+      // in any meaningful way.
```
blueridger, line 1190: Would be better as a docstring.

blueridger (commented):
It looks like you are using `git please` even though you are committing new changes as new commits. This makes it a little harder to track which commits are new. I'd recommend just using `git push` if you are not using `git commit --amend`

topocount:
> It looks like you are using `git please` even though you are committing new changes as new commits. This makes it a little harder to track which commits are new. I'd recommend just using `git push` if you are not using `git commit --amend`

@blueridger this PR has been open for a while, and I rebased to keep the changes up to date with the most recent updates in `main`. I've been appending changes via new commits, since rebasing doesn't change the date of a commit, it should be possbile to page into the relevant changes by referencing dates. I'll post deep links containing the relevant commits going forward.

packages/sourcecred/src/core/ledger/ledger.js
topocount: it can't. Calls to public methods from private methods result in out-of-order timestamps.

This is an acceptable location because it only calls public public methods under the hood. I'd like to keep this as a private method even though it only invokes a public method though because I don't want it to be a documented part of the API right now.

blueridger:
Ahhh I see. No need to deep link, I can look for the timestamps. Thanks.

packages/sourcecred/src/core/ledger/ledger.js
blueridger: Ah cool

blueridger (approved):

Issue #3178, opened 2021-08-30
Add delimiter to guildId in Discord config


magwalk:
Currently, we allow users to add a delimiter with a label so the strings of numbers on config files are much easier to grok for channel weights and role weights. It would be great if we could add the delimiter to the guildId so instances that use multiple discord servers can easily distinguish which one is which. 

See the delimited function in action here: https://github.com/sourcecred/sourcecred/blob/main/packages/sourcecred/src/plugins/discord/config.js#L89

To update this code here: https://github.com/sourcecred/sourcecred/blob/main/packages/sourcecred/src/plugins/discord/config.js#L74

magwalk:
Closed as of: https://github.com/sourcecred/sourcecred/pull/3194

PR #3194, opened 2021-09-21
support delimiter for Discord's guildid

amrro:
# Description

Enable to delimiter to `guildId` in order to recognize numbers and id just like comments
as described in #3178

# Test Plan

Unit tests included with an example of delimiter.

### DIFF
--- a/packages/sourcecred/src/plugins/discord/config.test.js
+++ b/packages/sourcecred/src/plugins/discord/config.test.js
@@ -6,7 +6,7 @@ describe("plugins/discord/config", () => {
   it("can load a basic config", () => {
     const raw = [
       {
-        guildId: "453243919774253079//sourcecred",
+        guildId: "453243919774253079",
         reactionWeightConfig: {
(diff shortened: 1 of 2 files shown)


amrro:
Fixes #3178

packages/sourcecred/src/plugins/discord/config.test.js
```diff
@@ -6,7 +6,7 @@ describe("plugins/discord/config", () => {
   it("can load a basic config", () => {
     const raw = [
       {
-        guildId: "453243919774253079",
+        guildId: "453243919774253079//sourcecred",
```
hozzjss, line 9: Could you add a test for when there's no delimiter since this should be optional and backwards compatible and since i fear the future updates to this

hozzjss (changes requested):
Nice dude, just a small thing

hozzjss (approved):
Okay looks good to me but I want someone else to take a look as well

Issue #3188, opened 2021-09-20
As an instance maintainer, I can configure discord channels by category instead of just by single channel. 


magwalk:
Per this discussion: https://discord.com/channels/453243919774253079/718263631158050896/888764649770741770

This would allow users to set their discord configuration to go by channel category (GUILD_CATEGORY) instead of just by channel name (GUILD_TEXT). For instances that have a very high number of channels, this would make setup a lot easier so they don't have to have every single channel signified in the config.

thohoh:
In one of our chat categories we have dynamic channels being added to /moved to archive category. Yet being able to track those channels for contribution is also important. If, by specifying just channels category I can make sure all the new channels will be scanned for contributions - that would be awesome.

PR #3195, opened 2021-09-22
Discord: add support for Channel Category weights

blueridger:
# Description
closes https://github.com/sourcecred/sourcecred/issues/3188

This adds support for Category-wide weights on Discord. You can add channel category IDs into the same channelWeightConfig.weights array, and they will be applied. The semantics are:
- **If no weight is set for channel or category,** use defaultWeight^2
- **If a weight is set for channel but not category, or vice versa,** use specified weight * (defaultWeight || 1)
- **If both a channel weight and a category weight is set,** multiply them together

# Design decisions
The defaultWeight^2 looks a little odd, but considering the defaultWeight is generally either 0 or 1, this won't have much effect. This is important for relative weight consistency between various use cases in the case that someone does set a default besides 0 or 1.

I chose to include the category ids in the same configuration array as channel ids because when we build support for threads, the parent_id will be polymorphic to either a category or a channel. We will also have to consider channelGrandparents at that point... wee!

I chose not to follow the existing pattern of attaching the channel metadata to the reactions in the DB because it will be much more efficient to just pass the channel metadata down during createGraph, as I have done. It may be worth changing the channelId to work similarly in the future.

# Test Plan
`yarn test` - updated unit tests thoroughly and also updated snapshot tests

## Manual Testing
1. `yarn clean-all && scdev load` on our prod instance
2. `scdev graph && scdev credrank` to set the baseline and do some basic regression testing
3. Added `"678302244524785699": 99` (PRODUCT category in our server) to the channelWeightConfig.weights array in the discord config file
4. `scdev graph -d` and made sure the node weight changes looked right.
```
NodeAddress["sourcecred","discord","REACTION","670490733077332000","🥳","778446953004007475","875846029881278534"]
Old:    297
New:    3

NodeAddress["sourcecred","discord","REACTION","670490733077332000","🪲","258786206387666944","804454196656209982"]
Old:    297
New:    3

  sourcecred/discord - Summary of Changes
Node Diffs: 0
Node Weight Diffs: 1009
Edge Diffs: 0
Edge Weight Diffs: 0
```

### DIFF
--- a/packages/sourcecred/src/plugins/discord/__snapshots__/fetcher.test.js.snap
+++ b/packages/sourcecred/src/plugins/discord/__snapshots__/fetcher.test.js.snap
@@ -6,60 +6,70 @@ Array [
     "id": "678348980756938813",
     "name": "Text Channels",
     "nsfw": false,
+    "parentId": null,
     "type": "GUILD_CATEGORY",
   },
   Object {
     "id": "678348980798881844",
     "name": "Voice Channels",
     "nsfw": false,
+    "parentId": null,
     "type": "GUILD_CATEGORY",
   },
   Object {
     "id": "678348980849213472",
     "name": "general",
     "nsfw": false,
+    "parentId": "678348980756938813",
     "type": "GUILD_TEXT",
   },
   Object {
     "id": "678348980878573661",
     "name": "General",
     "nsfw": false,
+    "parentId": "678348980798881844",
     "type": "GUILD_VOICE",
   },
   Object {
     "id": "678394406507905129",
     "name": "pagination",
     "nsfw": false,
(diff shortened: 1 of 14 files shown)


AL0YSI0US:
Yay...Hacking. 💻

hozzjss:
> The defaultWeight^2 looks a little odd, but considering the defaultWeight is generally either 0 or 1, this won't have much effect. This is important for relative weight consistency between various use cases in the case that someone does set a default besides 0 or 1.

I'm not 100% with this decision tbh and I don't see the use case i would have just used default weight since it is the default but you might have a better opinion on this

hozzjss (approved):
All good tests pass, manual tests are alright

PR #3203, opened 2021-09-24
Discord: Simplify default graph + simplifyGraph config

blueridger:
# Description
This makes two changes:
1. omits messages (and connected reactions and members) when the sum of the reaction weights for that message is 0
2. creates a `simplifyGraph` configuration option that omits reaction nodes (and connected members) and instead adds up the reaction weights and directly sets the sum as the message node weight. This sacrifices analytic possibilities for permance. To use this feature, add the config attribute `"simplifyGraph": true`

Reduced graph size means reduced memory usage during credrank, which is one of our biggest current bottlenecks.

# Test Plan

Set up a baseline graph pre-change (one of our prod channels) and then manually verified with `scdev graph sourcecred/discord -d -s >> results.txt`

1. Changes to default graph: [results-false.txt](https://github.com/sourcecred/sourcecred/files/7227758/results-false.txt)

2. Changes to graph when `"simplifyGraph": true` [results-true.txt](https://github.com/sourcecred/sourcecred/files/7227757/results-true.txt)

### DIFF
--- a/packages/sourcecred/src/plugins/discord/createGraph.js
+++ b/packages/sourcecred/src/plugins/discord/createGraph.js
@@ -296,7 +296,9 @@ export function createGraph(
   repo: SqliteMirrorRepository
 ): WeightedGraphT {
   const graphMessages = findGraphMessages(repo);
-  return _createGraphFromMessages(config, graphMessages);
+  return config.simplifyGraph
+    ? _createSimpleGraphFromMessages(config, graphMessages)
+    : _createGraphFromMessages(config, graphMessages);
 }
 
 export function _createGraphFromMessages(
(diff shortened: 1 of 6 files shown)


packages/sourcecred/src/plugins/discord/createGraph.js
```diff
@@ -320,48 +322,116 @@ export function _createGraphFromMessages(
...
-          reactions,
-          channelParentId
-        )
-      );
-      wg.graph.addNode(node);
-      wg.graph.addNode(memberNode(reactingMember));
-      wg.graph.addEdge(reactsToEdge(reaction, message));
-      wg.graph.addEdge(
-        addsReactionEdge(reaction, reactingMember, message.timestampMs)
+      const weight = reactionWeight(
```
rjharmon, line 328: lots of args : /  ... maybe reactionWeight should just take {weights, graphMessage}

I see you're just refactoring what already existed, so grain of salt.

packages/sourcecred/src/plugins/discord/createGraph.js
```diff
@@ -320,48 +322,116 @@ export function _createGraphFromMessages(
...
+        message,
+        reaction,
+        reactingMember,
+        propsChannels,
+        reactions,
+        channelParentId
+      );
+    }
+    if (messageWeight) {
+      if (author) {
```
rjharmon, line 414: Looks like some copy/paste here.  Maybe you considered extracting  a `addMessageToGraphWithWeight` function already

blueridger: accepted

packages/sourcecred/src/plugins/discord/createGraph.js
blueridger: Passing graphMessage would take the params down from 7 to 5... Usually what we do is turn a long list of params into an object, so then they at least have names identifying them, instead of mere param order.

rjharmon: +1 options object, non-blocking

packages/sourcecred/src/plugins/discord/createGraph.js
```diff
@@ -320,48 +320,53 @@ export function _createGraphFromMessages(
...
     }
-
-    for (const {member, count} of mentions) {
-      wg.graph.addNode(memberNode(member));
-      let edge;
-      if (propsChannels.has(channelId)) {
-        edge = propsEdge(message, member);
-      } else {
-        edge = mentionsEdge(message, member);
+    if (messageWeight) {
```
rjharmon, line 347: It's unclear to me why this is added here.  Is there a higher-level meaning we can assign to messageWeight, like `const hasNoWeight=!messageWeight`.  non-blocking

rjharmon: ...assuming that's the reasoning for the conditional

rjharmon: ... and if we can stand the incremental cost of another variable.  

else maybe `// don't bother adding weights to a message having no value` ??

packages/sourcecred/src/plugins/discord/createGraph.js
```diff
@@ -320,48 +320,53 @@ export function _createGraphFromMessages(
...
+        weights,
+        message,
+        reaction,
+        reactingMember,
+        propsChannels,
+        reactions,
+        channelParentId,
+      });
+      messageWeight += weight;
+      if (weight && !config.simplifyGraph) {
```
rjharmon, line 336: OK I think I'm getting the hang of this change now. 

I would probably put the simple case first here: if (simplifyGraph) { ... } else { ...complicated stuff ... }

I might even use `if (!messageWeight) continue` here near the top of the loop, because it would seem more obvious.  I don't know if you'd treat `continue` like a smell, though.  WDYT?

rjharmon (approved):
Looks good, just a few "thinking out loud" comments, some of which might help increase code clarity

packages/sourcecred/src/plugins/discord/createGraph.js
blueridger: I like it

rjharmon (approved):
sweet

rjharmon (approved):

PR #3010, opened 2021-05-05
PersonalAttributions: add config and add to API to complete feature

blueridger:
# Description
This adds a config to the instance for the Personal Attribution feature. It also uses the config in the credrank API. This completes the Personal Attribution feature V1.

# Test Plan
Edited the `config/personalAttributions.json` and then ran `scdev credrank -d` and verified console output as well as json reformatting.
## No config file
Console output:
```

# Top Participants By New Cred
┌─────────┬──────────────────┬────────────┬───────────┬──────────┐
│ (index) │       Name       │ Prior Cred │ New Cred  │ % Change │
├─────────┼──────────────────┼────────────┼───────────┼──────────┤
│    0    │  'decentralion'  │ '37739.3'  │ '37739.3' │  '0.0%'  │
│    1    │   'lbstrobbe'    │ '17159.3'  │ '17159.3' │  '0.0%'  │
│    2    │     'Thena'      │ '16911.7'  │ '16911.7' │  '0.0%'  │
│    3    │     's-ben'      │ '14595.6'  │ '14595.6' │  '0.0%'  │
│    4    │    'KuraFire'    │ '14573.2'  │ '14573.2' │  '0.0%'  │
│    5    │    'wchargin'    │ '13000.1'  │ '13000.1' │  '0.0%'  │
│    6    │   'topocount'    │ '12720.9'  │ '12720.9' │  '0.0%'  │
│    7    │  'panchomiguel'  │ '12141.9'  │ '12141.9' │  '0.0%'  │
│    8    │     'hammad'     │ '11310.1'  │ '11310.1' │  '0.0%'  │
│    9    │  'joiecousins'   │  '9122.3'  │ '9122.3'  │  '0.0%'  │
│   10    │      'Jojo'      │  '8023.5'  │ '8023.5'  │  '0.0%'  │
│   11    │      'bex'       │  '7134.5'  │ '7134.5'  │  '0.0%'  │
│   12    │   'sandpiper'    │  '6903.1'  │ '6903.1'  │  '0.0%'  │
│   13    │     'beanow'     │  '6524.2'  │ '6524.2'  │  '0.0%'  │
│   14    │   'echojuliet'   │  '5063.7'  │ '5063.7'  │  '0.0%'  │
│   15    │   'dependabot'   │  '2811.3'  │ '2811.3'  │  '0.0%'  │
│   16    │    'benoxmo'     │  '2515.5'  │ '2515.5'  │  '0.0%'  │
[...]

PR #3206, opened 2021-09-25
Add raw GrainConfig parsers

blueridger:
# Description
This adds the idea of grain Config types that represent the raw grain JSON. It also adds parsers for these types.

This is part of a larger trend of decoupling validation of JSON and augmentation of JSON. These will be used in upcoming changes, and I anticipate using them more broadly over time.

# Test Plan
I've done some testing already in a follow up change that relies on this. None included in this PR though.

### DIFF
--- a/packages/sourcecred/src/core/ledger/policies/balanced.js
+++ b/packages/sourcecred/src/core/ledger/policies/balanced.js
@@ -28,6 +28,12 @@ export type BalancedPolicy = {|
   +numIntervalsLookback: number,
 |};
 
+export type BalancedConfig = {|
+  +policyType: Balanced,
+  +budget: string | number,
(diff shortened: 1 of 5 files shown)


topocount (approved):
My only request: can you link to some other PRs (open or closed) that build on this larger trend of decoupling the JSON parsing and augmenting work?

blueridger:
Examples of separated validation and augmentation:
https://github.com/sourcecred/sourcecred/pull/3010
https://github.com/sourcecred/sourcecred/pull/2844

One reason for this trend is it allows 3rd party developers to have a similar interface as end users when configs are taken as API inputs. It also allows better config editing/handling like in https://github.com/sourcecred/sourcecred/pull/3210

PR #3207, opened 2021-09-25
Combo: Add fmap as a parser class method

blueridger:
# Description

BEHOLD. I'm always so annoyed working with fmap because in order to add or remove it, you have to make two edits: one on either side of the base parser. This adds a class method on the Parser class that wraps fmap, making it syntactically similar to Array.prototype.map() which makes it much nicer to work with in my opinion.

```
// Before

C.fmap(baseParser, (base) => base.toString())

// After

baseParser.fmap((base) => base.toString())
```

I included some example usages in this PR. I'll be baking such migrations into other PRs as I go along.

# Test Plan
sourcecred/cred/gh-pages
1. `scdev credrank` and verify no errors
2. `scdev serve` open site and verify no errors

### DIFF
--- a/packages/sourcecred/src/util/combo.js
+++ b/packages/sourcecred/src/util/combo.js
@@ -36,6 +36,9 @@ export class Parser<+T> {
       throw new Error(result.err);
     }
   }
+  fmap<U>(f: (T) => U): Parser<U> {
+    return fmap(this, f);
+  }
 }
 
 // Helper type to extract the underlying type of a parser: for instance,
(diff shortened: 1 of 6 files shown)


packages/sourcecred/src/util/combo.js
```diff
@@ -36,6 +36,9 @@ export class Parser<+T> {
       throw new Error(result.err);
     }
   }
+  fmap<U>(f: (T) => U): Parser<U> {
```
rjharmon, line 39: +1 for impatience

hozzjss (approved):

wchargin (commented):
FWIW, I'm not enthused about this change because the `Parser` class is
meant to be a thin type that's orthogonal to the actual parsing function
itself, not a grab bag home for combinators. But I also don't have
enough skin in the game to really object.

(Obviously, in a language with a good type system, all the combinators
could be made available with method syntax via traits or UFCS or
similar: Rust, Haskell, D can all do this. But we have no such luxuries
here in JS-land.)

PR #3209, opened 2021-09-28
GitHub storage conforms data storage

### LINKED ISSUE #2871: Refactor the Ledger's GithubStorage to comform to the DataStorage interface
Now that we have `DataStorage` and `WritableDataStorage`, we should refactor [`GithubStorage`](https://github.com/sourcecred/sourcecred/blob/master/src/api/ledgerStorage/github.js) to utilize these interfaces.

This is also an excellent chance to utilize the `raw.githubusercontent.com` api instead of graphQL for `read`/`get` calls, since it's unauthenticated and free.

amrro:
# Description
#2871 & #2768 describe the motivation and issue exactly.

# Test Plan
Will write testing to test the functionality of it. 
Also I've tried it on a real-world example and it worked only on `get()` method. 

closes #2871 , #3240

paired with: @topocount

### DIFF
--- a/packages/sourcecred/src/core/storage/github.js
+++ b/packages/sourcecred/src/core/storage/github.js
@@ -6,7 +6,7 @@ import type {GithubToken} from "../../plugins/github/token";
 import type {RepoId, RepoIdString} from "../../plugins/github/repoId";
 import fetch from "cross-fetch";
 import {decode as base64Decode} from "base-64";
-import {decode, encode} from "./textEncoding";
+import {encode} from "./textEncoding";
 import {toBase64} from "@aws-sdk/util-base64-browser";
 
 
--- a/packages/sourcecred/src/core/storage/github.js
+++ b/packages/sourcecred/src/core/storage/github.js
@@ -1,0 +1,168 @@
+// @flow
+
+import type {WritableDataStorage} from "./index";
+import {stringToRepoId} from "../../plugins/github/repoId";
+import type {GithubToken} from "../../plugins/github/token";
+import type {RepoId, RepoIdString} from "../../plugins/github/repoId";
+import fetch from "cross-fetch";
+import {decode as base64Decode} from "base-64";
+import {Ledger} from "../ledger/ledger";
+import {decode, encode} from "./textEncoding";
+
+const GET_LEDGER_QUERY = `
+query getLedger($owner: String!, $repo: String!, $expression:String!) {
+  repository(owner: $owner, name: $repo) {
+    object(expression: $expression) {
+      ... on Blob {
+        oid
+      }
+    }
+  }
+}
+`;
+
+type CreateBlobRes = {
+  sha: string,
+  url: string,
+};
+
+type Opts = {
+  apiToken: GithubToken,
+  repo: RepoIdString,
+  branch: string,
+};
+
+export class GithubStorage implements WritableDataStorage {
+  static ENDPOINT: string = "https://api.github.com";
+
+  constructor(opts: Opts) {
+    this._token = opts.apiToken;
+    this._repoId = stringToRepoId(opts.repo);
+    this._branch = opts.branch;
+  }
+
+  _token: GithubToken;
+  _repoId: RepoId;
+  _branch: string;
+
+  /**
+   * data/ledger.json
+   * @param path data/ledger.json
+   * @returns {Promise<Ledger>}
+   */
+  async get(path: string): Promise<Uint8Array> {
+    const options = {
+      method: "POST",
+      headers: {
+        Authorization: `Bearer ${this._token}`,
+      },
+      body: JSON.stringify({
+        query: GET_LEDGER_QUERY,
+        variables: {
+          owner: this._repoId.owner,
+          repo: this._repoId.name,
+          expression: `${this._branch}:${path}`,
+        },
+      }),
+    };
+
+    const result = await fetch(`${GithubStorage.ENDPOINT}/graphql`, options);
+
+    const {data} = await result.json();
+    const blobSha = data.repository.object.oid;
+
+    const ledgerBlobRes = await fetch(
+      `${this._getRepoEndpoint()}/git/blobs/${blobSha}`,
+      {
+        method: "GET",
+        headers: {
+          Authorization: `Bearer ${this._token}`,
+        },
+      }
+    );
+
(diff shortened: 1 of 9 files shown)


amrro:
I've copied the exact logic from the old `GithubStorage` that conforms to `LedgerStorage` and moved, as [suggested](https://github.com/sourcecred/sourcecred/issues/2871#issuecomment-851645451), and update it to conform to `DataStorage`/`writableDataStorage`. 

I've tried to make it work on real world example to `get()` and `set()` for `sourcecred/cred` & `amrro/sourcecred`. 
It was hard to test and find where the issues are which lead to the next suggestions or motivation. 



### Problem
Each method from `GithubStorage` is mutli-request to the GitHub's API. That inherently makes the method complex and error-prone, Hence: 
- Methods are long with convoluted logic. 
- As a result, It's hard to read, took me a lot of time to understand what is going on.
- We need to validate the Rest Request itself, to handle if there are any errors. 
- We need to validate the data passed and returned from the Github Itself. 
- We need to validate the data passed by the user.
- A lot of logic and hard to test.

I guess this all makes it so complex to work with. Also, it's hard to show the right error to the user in order to recover from this error. Also, **I guess there's a lot of edge cases, that we might not notice**. 
#### Example. 
> As a user, I want to load the config file for the first time, which might not exist yet in the GitHub’s Repo.

In order to load any file, Github is using `git lfs` to store large files away form the repo itself. Apparently we need `oid`, or unique object identifier, for this file in order to locate where it's stored and load it the next request. 

If the request fails to find this file, or the `oid`, it just returns `null`. Consequently, we need to recover from this and notify user that the file doesn't exist on the specified path. 

### Proposals

We need to abstract these Github specific operations from using it as a storage for two important reasons: 
- Sourcecred work as plugin-based system; In the future, we might need to support more custom usages like self-hosted git repos.
- Maintain both, Github operations & using Github as Storage, separately  in order to make it easier to work with them, reduce the cost of maintain them.

#### Suggestion: 
I found this extensible, official API library for Github: [`ocatkit/core`](https://github.com/octokit/core.js). I found it as a great library it reduces the required knowledge to work with GitHub in a declarative way. 
here a grabbed example from the official [docs](https://github.com/octokit/core.js#rest-api-example): 
```js
// Create a personal access token at https://github.com/settings/tokens/new?scopes=repo
const octokit = new Octokit({ auth: `personal-access-token123` });

const response = await octokit.request("GET /orgs/{org}/repos", {
  org: "octokit",
  type: "private",
});

```

It's declarative and easier to read. Also, it supports plugins.

amrro:
This is an example I found in the project on how long validation it needs in order to validate a single request. 
https://github.com/sourcecred/sourcecred/blob/6fc431c2211f07b11c7497b8ad9d5655080c0e1a/packages/sourcecred/src/plugins/github/fetchGithubRepo.js#L153-L197

packages/sourcecred/src/core/storage/github.test.js
```diff
@@ -0,0 +1,91 @@
...
+    it("make sure everything is working.", async () => {
+      // expect.hasAssertions();
+      const storage = new GithubStorage({
+        apiToken: `GithubToken`,
+        repo: `${owner}/${name}`,
+        branch: "string",
+      });
+
+      const result = await storage.get("data/ledger.json");
+      await expect(result).toEqual("value");
```
hozzjss: this test fails

blueridger:
I support the use of ocatkit/core

packages/sourcecred/src/plugins/github/repoId.test.js
```diff
@@ -25,7 +25,6 @@ describe("plugins/github/repoId", () => {
   });
   describe("RepoIdString type", () => {
     it("manually constructing a RepoIdString is illegal", () => {
-      // $FlowExpectedError[incompatible-type]
       const _unused_repoIdString: RepoIdString = "foobar";
```
topocount, line 28: Why shouldn't this error anymore? I think it should

packages/sourcecred/src/core/storage/github.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  _getRepoEndpoint(): string {
+    const {owner, name} = this._repoId;
+    return `${GithubStorage.ENDPOINT}/repos/${owner}/${name}`;
+  }
+}
+
+export class WritableGithubStorage extends GithubStorage implements  WritableDataStorage {
+  async set(path: string, ledger: Uint8Array, message?: string): Promise<void> {
```
topocount: The code is already generalized, so as discussed, let's generalized the payload param (and its handlers) from `ledger` to `content`.

packages/sourcecred/src/core/storage/github.test.js
```diff
@@ -0,0 +1,120 @@
...
+      const storage = new GithubStorage({
+        apiToken: `GithubToken`,
+        repo: `${mockOwner}/${mockName}`,
+        branch: mockBranch,
+      });
+
+      const result = await storage.get("data/ledger.json");
+      await expect(result).toEqual(encode(""));
+    });
+    it("throws on upward traversal when the base is a url", async () => {
```
topocount, line 103: I think this test title is incorrect

packages/sourcecred/src/core/storage/github.test.js
```diff
@@ -0,0 +1,120 @@
...
+        branch: mockBranch,
+      });
+
+      const path = "data/file.json";
+      const value = "hello";
+      const encodedValue = encode(value);
+      await storage.set(path, encodedValue, "update text");
+
+      const result = await storage.get(path);
+      await expect(result).toEqual(encode(value));
```
topocount, line 117: why aren't we decoding the result here? Is this a bug in the test fixtures or the Github Class?

amrro: just passing type array mimicking the real-world example

packages/sourcecred/src/core/storage/github.test.js
```diff
@@ -0,0 +1,120 @@
...
```
topocount, line 117: ahh I see. This is super pedantic, but this is clearer to me: 
```suggestion
      await expect(decode(result)).toEqual(value);
```

topocount (approved):
one last nit!

PR #3210, opened 2021-09-29
Deprecate old GrainConfig attributes + add updater

blueridger:
# Description

# Test Plan


packages/sourcecred/src/core/ledger/grain.js
```diff
@@ -161,11 +161,11 @@ export function format(
...
- * format(100000000000000) === "0.0001g"
- * format(150000000000000000000) === "150g"
- * format(15000000000000000000000) === "15,000g"
- * format(15000000000000000000000, "seeds") === "15,000seeds"
- * format(15000000000000000000000, "") === "15,000"
+ * formatAndTrim(100000000000000) === "0.0001g"
+ * formatAndTrim(150000000000000000000) === "150g"
+ * formatAndTrim(15000000000000000000000) === "15,000g"
+ * formatAndTrim(15000000000000000000000, "seeds") === "15,000seeds"
+ * formatAndTrim(15000000000000000000000, "") === "15,000"
```
topocount, line 168: :star_struck:

packages/sourcecred/src/cli/update/utils.js
```diff
@@ -0,0 +1,24 @@
...
+// Attributes from the `optionals` parameter will only be added if they
+// are not undefined.
+// It is recommended to define T explicitly when using this.
+// Example: `const object = buildObject<MyType>(...);`
+export function buildObject<T: {}>(
+  required: T,
+  optionals: $Shape<T>,
+  exclusions: $ReadOnlyArray<$Values<T> | typeof undefined> = [undefined]
+): T {
+  let config: T = cloneDeep(required);
```
topocount, line 15: Nit: drop this extra dependency ref and just do this:
```suggestion
  let config: T = {...required};
```

packages/sourcecred/src/cli/update/utils.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+import cloneDeep from "lodash.clonedeep";
+
+// This is a type-friendly way to add optional attributes to a config.
+// Attributes from the `optionals` parameter will only be added if they
+// are not undefined.
+// It is recommended to define T explicitly when using this.
+// Example: `const object = buildObject<MyType>(...);`
```
topocount, line 9: I think some example invocations using the options or even just exclusions would be super useful here. The API is not exactly self-explanatory. Also not entirely sure why this needs to be in the cli/update path. This seems to have some broader application. Could see it in a path having to do with config definitions, honestly.

packages/sourcecred/src/cli/update/utils.js
```diff
@@ -0,0 +1,24 @@
...
+export function buildObject<T: {}>(
+  required: T,
+  optionals: $Shape<T>,
+  exclusions: $ReadOnlyArray<$Values<T> | typeof undefined> = [undefined]
```
topocount, line 13: What's wrong with this?
```suggestion
  exclusions: $ReadOnlyArray<$Values<T>> = []
```
Is it some idiosyncrasy in the flow parser, or are you actually expecting some array entries to be undefined?

topocount (commented):
This looks pretty good so far. It'll be easier to do a full pass after it's rebased on the approved branch that is passing tests.

packages/sourcecred/src/cli/update/utils.js
blueridger: But if there are objects nested within `required` then those objects will still be passed by reference. I think it's better to have a clean deep clone.

packages/sourcecred/src/cli/update/utils.js
blueridger: Actually expecting undefined values in `optionals` as the `exclusions` are the values that should be not be passed from `optionals` into the result.

packages/sourcecred/src/cli/update/utils.js
blueridger: I'll look for another place to put it and will add more documentation.

PR #3251, opened 2021-11-02
Add auto-generated API documentation

blueridger:
# Description
The goal of this PR is to create a solution for computer-generated documentation. It includes these pieces:
- Added `documentation` dependency to the mono-repo root
- Added a yarn script and bash script for generating the docs. Usage: `yarn docs` at either the root or package level
- Ran the script to generate docs for sourcecred 0.10.0
   - **Note: I put this in a separate commit to make reviewing easier**

The docs aren't super great, but they are a decent start (better than nothing) and they are at least highly searchable.

What this PR means for maintenance is that when someone is doing a version release, they can also update the docs with `yarn docs`

This also means that we will need to be more on top of using `/**   */` docstrings on everything we want in the API docs.

# Design choices
I don't know how to design for mono-repos very well yet, so if something I've done is super tacky, let me know what a better way to do it would be. It seems simple and maintainable enough as I've done it.

Docs are put into a /docs folder for github pages compatibility. I figure we can just serve gh-pages out of `main`.

Each package will have its own folder in the site, and the script will set the currently generated docs as the default root "latest" (replacing the existing latest) and then also create an archival duplicate under a version-named folder.

# Test Plan
## Docs viewing
Open the html file you want to look at in a browser

## Docs generation script
- `yarn docs` at root level
- `yarn docs` at package level
- edit a docstring and rerun script, verify docs have been over-written
- change version number and rerun script, verify "latest" default docs have been over-written and new version archive added


hozzjss (approved):
looks good lfg

PR #3258, opened 2021-11-10
Analysis API: Neo4j csv files

blueridger:
# Description
closes #3254 
This PR adds end-to-end Neo4j-formatted CSV file generation to the Analysis API/CLI. Once generated, the CSV files can be imported into Neo4j using neo4j-admin.
Usage: `yarn sourcecred analysis -n` or `yarn sourcecred analysis --neo4j`
Effect: CSV files are written in the directory `output/neo4j` and some Neo4j CLI help is printed in the console.

This is useful because Neo4j is a graph database that has a query language and visualization tools, allowing technical users to do powerful graph analytics very efficiently.

I also baked in adding the analysis API to our library.

# Design choices
I put it under a feature flag because right now, some instances run `analysis` in their github actions and commit the results, so it's better for the default behavior to skip the neo4j feature to reduce clutter. It might be worth it at some point to make all analysis submodules opt-in, though this would be a breaking change.

I'm sure we could think of improvements that could be made to this sort of code, but I didn't nitpick too much since sloppier non-critical code is kind of the whole point of the `analysis` API.

Making Flow like hand-made Iterators is weird. Don't look at me. (Flow has some open tickets for bug fixes in their Iterator typing.)

# Test Plan
1. `scdev analysis` and verify neo4j files are not generated.
2. `scdev analysis -n` and then follow the instructions in the console to import into Neo4j
3. Open Neo4j DB and verify:
Screen Shot 2021-11-09 at 9 35 32 PM

PR #3259, opened 2021-11-10
Util: add batchArray and batchIterator methods

blueridger:
# Description
These methods help batch Arrays and Iterators (split the contents up into chunks).

The `batchIterator` method is used in the follow-up PR: https://github.com/sourcecred/sourcecred/pull/3258

The `batchArray` just because it might be useful at some point soon as we run into more scalability concerns in places we don't have iterators in use yet.

# Test Plan
unit tests + https://github.com/sourcecred/sourcecred/pull/3258

### DIFF
--- a/packages/sourcecred/src/util/batch.js
+++ b/packages/sourcecred/src/util/batch.js
@@ -10,6 +10,7 @@ export function batchArray<T>(
   array: $ReadOnlyArray<T>,
   batchSize: number
 ): $ReadOnlyArray<$ReadOnlyArray<T>> {
+  if (batchSize < 1) throw new Error("BatchSize must be 1 or more.");
   const result = [];
   let backlog = array;
   while (backlog.length) {
@@ -41,6 +42,7 @@ export function batchIterator<T>(
   iterator: Iterator<T> | Generator<T, void, void>,
   batchSize: number
 ): BatchIterator<T> {
+  if (batchSize < 1) throw new Error("BatchSize must be 1 or more.");
   let itemsCompletedInCurrentBatch = 0;
   let queue = iterator.next();
   let hasNext = !queue.done;
--- a/packages/sourcecred/src/util/batch.js
+++ b/packages/sourcecred/src/util/batch.js
@@ -1,0 +1,74 @@
+// @flow
+
+/**
+Returns an array of arrays that contains all of the items in the original
+array parameter, but batched into arrays no larger than the batchSize.
+Example:
+  batch([1,2,3,4,5], 2) = [[1,2], [3,4], [5]]
+*/
+export function batchArray<T>(
+  array: $ReadOnlyArray<T>,
+  batchSize: number
+): $ReadOnlyArray<$ReadOnlyArray<T>> {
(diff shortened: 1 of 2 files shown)


packages/sourcecred/src/util/batch.js
```diff
@@ -0,0 +1,74 @@
...
+Example:
+  batch([1,2,3,4,5], 2) = [[1,2], [3,4], [5]]
+*/
+export function batchArray<T>(
+  array: $ReadOnlyArray<T>,
+  batchSize: number
+): $ReadOnlyArray<$ReadOnlyArray<T>> {
+  const result = [];
+  let backlog = array;
+  while (backlog.length) {
```
amrro, line 15: nit: I've tried to come up with a functional way without worrying about length that satisfies most of corner cases, then found `reduce` function: it's not well-documented on flow.
```js
function batchArray<T>(array: $ReadOnlyArray<T>, batchSize: number): $ReadOnlyArray<$ReadOnlyArray<T>> {
  return array.reduce(
    (previousValue, currentItem, idx) =>
      (idx % batchSize === 0 // simply, should we start a new array item or not.
        ? previousValue.push([currentItem]) // yes? okay, wrap the current item into array.
        : previousValue[previousValue.length - 1].push(currentItem)) &&
      previousValue,
    <T>[] // initial value incorrect syntax
  );
}
``` 

consider it as an alternative, but _personally_ I think loops are source of evil.

packages/sourcecred/src/util/batch.js
```diff
@@ -0,0 +1,74 @@
...
+    result.push(backlog.slice(0, batchSize));
```
amrro, line 16: pedantically speaking: what if the `batchSize` is `0` or with minus value? 
I think, the `while` loop will run forever, if the `batchSize` is smaller than `1`, for sure if `0`.

packages/sourcecred/src/util/batch.test.js
```diff
@@ -0,0 +1,80 @@
+// @flow
+
+import {batchArray, batchIterator} from "./batch";
+
+describe("src/util/batch", () => {
+  describe("batch", () => {
+    it("batches an array bigger than the batch size", () => {
```
amrro, line 7: Suggestion: add tests for corner cases of `batchSize`.

amrro (changes requested):

packages/sourcecred/src/util/batch.js
blueridger: Clever, I think that reduces readability significantly. It's more understandable as written--a batching method working at the granularity of batches.

packages/sourcecred/src/util/batch.test.js
blueridger: accepted

packages/sourcecred/src/util/batch.js
blueridger: Good point. I'll add a conditional throw.

amrro (approved):

PR #3262, opened 2021-11-12
Fix sorting in accounts preview

amrro:
# Description
sorting wasn't working properly due to a bug with `useMemo`, it affected the sorting when you change the sorting order for the same column. It won't work.

# Test Plan
Manually.


topocount (approved):
works for me!

PR #3263, opened 2021-11-12
Add support for sorting by Active accounts.

### LINKED ISSUE #3237: Sort by active status in grain accounts page
As an instance admin, I want to quickly be able to view who is active, but deactivations can cause mixing of statuses in the grain-sorted list. Add a sort by active to this file using the TableState, as used by the other sort options: https://github.com/sourcecred/sourcecred/blob/main/packages/sourcecred/src/ui/components/AccountOverview.js

amrro:
# Description
Accounts can be sorted by active or not accounts.  

# Test Plan
Manually.

closes #3237

### DIFF
--- a/packages/sourcecred/src/ui/components/AccountOverview.js
+++ b/packages/sourcecred/src/ui/components/AccountOverview.js
@@ -54,6 +54,14 @@ export const AccountOverview = ({
   );
 
   const accounts = useMemo(() => Array.from(ledger.accounts()), []);
+
+  const ACTIVE_SORT = useMemo(() =>
+    deepFreeze({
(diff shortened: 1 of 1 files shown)


topocount (approved):

packages/sourcecred/src/ui/components/AccountOverview.js
```diff
@@ -54,7 +54,7 @@ export const AccountOverview = ({
   );
 
   const accounts = useMemo(() => Array.from(ledger.accounts()), []);
-
+
```
topocount, line 57: this might not pass lint

PR #3264, opened 2021-11-13
Feat/import plugins

zcstarr:
# Description
This PR is to enable sourcecred to be able to import plugins from npm or local js files. The motivation behind this is to improve the ability for 3rd party developers to extend sourcecred using the existing Plugin interface. 

The approach here was to be a non breaking change that would allow developers to specify plugins as they already due, but with the additional feature set of  "mynpmpackage" or "/some/path/my-local-dev-package.js" as the source for the plugin data. 

I believe this will lessen the pressure for 3rd party integrations to be pushed into sourcecred.
-->

# Test Plan
The test plan is to build a dummy package, that can be locally imported in tests, as well as installed in the local environment to replicate the real world usecase. The local replication can be achieved by using npm pack for the test plugin and then installing that test plugin package into the node environment. 

This will allow us to test the plugin end to end with relying on a network connection or packages outside of sourcecred to verify correctness.

### DIFF
--- a/packages/sourcecred/.flowconfig
+++ b/packages/sourcecred/.flowconfig
@@ -12,5 +12,6 @@ flow-typed
 
 [options]
 include_warnings=true
+module.ignore_non_literal_requires=true
 
 [strict]
--- a/packages/sourcecred/src/api/bundledPlugins.js
+++ b/packages/sourcecred/src/api/bundledPlugins.js
@@ -11,11 +11,19 @@ import {
   ExternalPlugin,
   ExternalPluginIdOwner,
 } from "../plugins/external/plugin";
+import {PackagePlugin} from "../plugins/package/plugin";
 import {DiskStorage} from "../core/storage/disk";
 
 /**
  * Returns an object mapping owner-name pairs to CLI plugin
  * declarations; keys are like `sourcecred/github`.
+ *
+ * External plugins use the `external/xxxx` prefix.
+ *
+ * Packaged plugins maybe specified directly as the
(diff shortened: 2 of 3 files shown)


zcstarr:
note this is a wip, not quite ready for a PTAL. But this is the general idea will discuss with @topocount  in the upcoming week.

topocount:
@zcstarr thanks for submitting this. I've looked it over and will be glad to discuss this on Tuesday or Wedensday

zcstarr:
@topocount  with the latest commits, there are probably a few talking points. 

PluginId has a very specific format, which makes it hard to overload as the source for what plugin to load. 

I'm uncertain if playing around with the PluginId fromString regex, makes sense or what the restrictions/concerns are with PluginId. I kind of bypassed it with writing fromStringRaw, but seems like graph and some other places might be using pluginId for other purposes, that using the extension might break.

The other thing here is that for tests I've uploaded a package that is the same package that I transpile and use for tests. Ideally I'd like to have the test harness produce the asset the "packaged dummy plugin", I had this almost working which is the installPlugin Task, however I ran into some issues with execution context. 

The alternative is to just have a package that is specifically for testing the package plugin, I've done this just to prove out it does indeed work. 

Last bit is that the way things worked previously is that if you don't specify a plugin correctly you get undefined, from calling getPlugin().  With the package plugin extension, you'll instead get back a package plugin that doesn't load or work, and will throw an error when used. This is a change from the existing behavior worth noting. I didn't change the signature of the method, because I wanted to confirm that this wouldn't be problematic.

PR #3271, opened 2021-11-23
Allow Grain Integrations to update their own configs

topocount:
A grain integration might want to store some details within an instance,
  such as a smart contract address. This update allows arbitrary updates
  to a scoped object within the grain config file.

[Merge Plan](https://github.com/sourcecred/sourcecred/issues/3139#issuecomment-978073574)

Test Plan:

Unit tests are provided at the grain integration interface. Integration
testing will be demonstrated in the test instance using the merkle
integration.

### DIFF
--- a/packages/sourcecred/src/core/ledger/grainIntegration.js
+++ b/packages/sourcecred/src/core/ledger/grainIntegration.js
@@ -28,6 +28,9 @@ export type PayoutResult = {|
   // If Grain balances are tracked in the ledger, these will be recorded as
   // transfers in the ledger to the "sink" Identity.
   transferredGrain: TransferredGrain,
+  // integration-specific config changes that need to be persisted are
+  // returned in the configUpdate object.
+  configUpdate: Object,
   // output files and content
   outputFile?: GrainIntegrationOutput,
 |};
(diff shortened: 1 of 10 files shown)


packages/sourcecred/src/core/ledger/grainIntegration.js
```diff
@@ -46,6 +49,7 @@ export type IntegrationConfig = {|
   // distributed funds via a configured integration
   processDistributions: boolean,
   currency: Currency,
+  integration: ?Object,
```
blueridger, line 52: This is a confusing line. Maybe it wants to be renamed `config` or something?

blueridger (approved):
1 change requested, but otherwise approved

packages/sourcecred/src/core/ledger/grainIntegration.js
topocount: This is confusing in the context of the named `IntegrationConfig` type, but I don't think its confusing once within the context of an integration. Within an integration, its pretty clearly just reading its own configuration.

From within the context of a grain integration, we have 
```
{
  accounting,
  currency,
  processDisitributions,
  integration
}
```

I'd be open to renaming it `self` but I think naming it `config` would more confusing and less descriptive in the alternative context, which needs to be much more approachable for devs looking to build their own integrations.

topocount: I've added a comment to clarify what's going on instead.

PR #3279, opened 2021-12-06
Convert snapshot file name serialization to hashes

topocount:
Snapshot file names were originally encoded in base64. This caused a
problem for longer request URIs because the file name would grow with
the size of the request.

by switching to sha256 message digests for the file names, we avoid this
issue of file name size growth.

This problem affects file systems that have a smaller maximum file name
lengths. The max allowable is 255 chars. The user reporting an issue had
theirs set to 143. sha256 message digests are all 64 chars in length.

These files are also now excluded from the npm releases as well, for good
measure and should no longer trip up any users looking setup or maintain
a SourceCred instance.

test plan:

unit tests should all still pass.


packages/sourcecred/package.json
```diff
@@ -186,7 +186,8 @@
     "/dist",
     "/src",
     "!src/**/*.test.js?(.snap)",
-    "!src/ui"
+    "!src/ui",
+    "!src/**/snapshots"
```
blueridger, line 190: do we need to include `!src/**/__snapshots__` too? Or are the underscores some weird path symbol I don't understand

packages/sourcecred/src/plugins/discord/mockSnapshotFetcher.js
```diff
@@ -1,13 +1,16 @@
...
 import {Fetcher} from "./fetcher";
+import {createHash} from "crypto";
 
 async function snapshotFetch(url: string | Request | URL): Promise<Response> {
   const snapshotDir = "src/plugins/discord/snapshots";
-  const filename = base64url(url);
+  const shasum = createHash("sha256");
+  shasum.update(Buffer.from(((url: any): string), "utf8"));
+  const filename = shasum.digest("hex");
+  //console.log({filename});
```
blueridger, line 13: debug comment might want removed

blueridger (approved):
a couple nits

packages/sourcecred/package.json
topocount: The optional .snap trailer excludes the files in the __snapshots__ folders

blueridger: ah i see

PR #3280, opened 2021-12-07
combo bugfix: `spacer` option renamed to `space`

topocount:
`spacer` is not an option in `json-stable-stringify`
and therefore has no effect on the formatting of strings passed in. As
demonstrated in the updated unit tests, spacing is now applied to the
expected strings.

Additionally instead of finding a way to correctly render an empty
array, a more verbose, human readable error messsage is supplied when
the empty case throws. I also added a contextual prefix to improve the
readability of both the empty and non-empty case.

paired with: @amrro

test plan:
Unit tests updated to demonstrate the correct indentation.

### DIFF
--- a/packages/sourcecred/src/util/combo.js
+++ b/packages/sourcecred/src/util/combo.js
@@ -205,19 +205,6 @@ export function fmap<T, U>(p: Parser<T>, f: (T) => U): Parser<U> {
   });
 }
 
-// replacer for stringify.JSON, in order to replace empty array with just `[]`.
-//
-// this replacer function can be extended and universally be used with other
-// aspect like empty objects.
-function emptyArrayReplacer(key: string, value: any) {
-  // check if the array is empty.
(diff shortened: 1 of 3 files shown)


packages/sourcecred/src/util/combo.js
```diff
@@ -248,10 +235,11 @@ function emptyArrayReplacer(key: string, value: any) {
 export function orElse<T: $ReadOnlyArray<Parser<mixed>>>(
   parsers: T,
   errorFn?: (string[]) => string = (errors) =>
-    `no parse matched: ${stringify(errors, {
-      replacer: emptyArrayReplacer,
-      spacer: 4,
-    })}`
+    errors.length === 0
+      ? "orElse Parser: no parsers entered"
```
amrro, line 239: Empty errors array always avoided 😃 👍🏻.

amrro (approved):

Issue #1539, opened 2020-01-10
Remove relationalView.addData


decentralion:
RelationalView has an `addRepository` method, whose presence suggests it's safe to combine multiple repositories' data in a single RelationalView. however, the comments reveal this is not the case:

https://github.com/sourcecred/sourcecred/blob/02c59d9967661a64ef9b3aad31198a2d265123ec/src/plugins/github/relationalView.js#L59-L67

we once had a reason for needing this function; see #711. however, now that we have proper organization support, we should refactor RelationalView so that it takes the data in the constructor, and doesn't expose the `addRepository` method. this way, no-one will be led into using the RelationalView in a way that is actually unsupported.

black32167:
Hi @decentralion, could you please have a look at my PR when have time and let me know if that's  make sense?

PR #3292, opened 2022-01-12
Removing RelationaView.addRepository

black32167:
this is  a  very minimalistic change to  address [#1539](https://github.com/sourcecred/sourcecred/issues/1539)(Remove relationalView.addRepository)

### DIFF
--- a/packages/sourcecred/src/plugins/github/relationalView.js
+++ b/packages/sourcecred/src/plugins/github/relationalView.js
@@ -47,7 +47,7 @@ export class RelationalView {
   _mapReferences: Map<N.RawAddress, N.ReferentAddress[]>;
   _mapReferencedBy: Map<N.RawAddress, N.TextContentAddress[]>;
 
-  constructor(): void {
+  constructor(repository?: T.Repository): void {
     this._repos = new Map();
     this._issues = new Map();
     this._pulls = new Map();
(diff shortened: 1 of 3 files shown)


packages/sourcecred/src/plugins/github/relationalView.test.js
```diff
@@ -444,16 +442,6 @@ describe("plugins/github/relationalView", () => {
     });
   });
 
-  it("addRepository is idempotent", () => {
```
black32167, line 447: This test seems redundant given we're removed 'addRepository'

packages/sourcecred/src/plugins/github/relationalView.js
```diff
@@ -47,7 +47,7 @@ export class RelationalView {
   _mapReferences: Map<N.RawAddress, N.ReferentAddress[]>;
   _mapReferencedBy: Map<N.RawAddress, N.TextContentAddress[]>;
 
-  constructor(): void {
+  constructor(repository?: T.Repository): void {
     this._repos = new Map();
```
black32167, line 51: Technically we do not need  `_repos` map as now only one repository, but moving to singular repository field  would require more significant changes (since `RelationalView` API will be altered). Maybe worthwhile to create a follow-up task to clean this up?

packages/sourcecred/src/plugins/github/relationalView.js
```diff
@@ -47,7 +47,7 @@ export class RelationalView {
   _mapReferences: Map<N.RawAddress, N.ReferentAddress[]>;
   _mapReferencedBy: Map<N.RawAddress, N.TextContentAddress[]>;
 
-  constructor(): void {
+  constructor(repository?: T.Repository): void {
```
blueridger, line 50: Why is this optional? Would it be easy to make required?

blueridger (commented):
Looks great, welcome to the codebase! Just one small question comment.

packages/sourcecred/src/plugins/github/relationalView.js
black32167: @blueridger Yeah this is because deserialization mechanism reads `RepoEntry` (internal state) but constructor accepts `T.Repository` (and then converts it to `RepoEntry` internally). I will try to turn it into factory/static method so  construction from  `T.Repository` will look like:
```
var rv = RelationalView.fromRepository(repo);
```
Unless you suggest anything else :man_shrugging:

packages/sourcecred/src/plugins/github/__snapshots__/relationalView.test.js.snap
```diff
@@ -326,14 +326,6 @@ Array [
 ]
 `;
 
-exports[`plugins/github/relationalView RelationalView entity: repos has expected number of them 1`] = `1`;
```
black32167, line 329: These tests do not seem relevant after moving to singular 'repo'

black32167:
**Update**
* The `repo` constructor argument has made mandatory.
* Separate `fromGraphQL` factory method introduced to create `RelationalView` from external representation
* Plural `RelationalView.repos()` method is removed  as well as an argument in the `RelationalView.repo()`.

PR #3309, opened 2022-01-28
Instructional sharness snapshot error message

blueridger:
# Description
When sharness-full fails, it's usually just because a code change was made that requires the snapshots to be updated, but currently that's only really passed on person to person as institutional knowledge. This PR adds some comments in the test file that are output in the console log so that any future maintainer or contributor can be pointed in the direction of the update_snapshots.sh script.

# Test Plan
I had CI run with this on code that needed a snapshot update. Open this run and browser search "not ok" to see the helpful tip displayed.
https://app.circleci.com/pipelines/github/sourcecred/sourcecred/7447/workflows/e66948ff-2814-4c11-a2c2-d77813ccd700/jobs/19653


amrro (approved):
Thanks, This is much helpful.

PR #3318, opened 2022-02-04
Explorer Home: Make Grain column name match configuration

blueridger:
# Description
The grain column name in the explorer page didn't dynamically match currency configuration (it always said Grain). This fixes that.

# Test Plan
yarn start --instance path/to/instance/with/currency/config

also regression test with instance without currency config

Screen Shot 2022-02-04 at 1 20 23 PM


amrro (approved):
Nit: I suggest to add `USDC (grains)`

In order to link the docs/philosophy within the actual data. Like hey, don't forget what grains are.

PR #3345, opened 2022-03-04
CredRank: add timestamp to MarkovProcessGraph nodes

blueridger:
# Description
This PR adds timestamps to the CredGraph by making a few small changes:
- makes the timestampMs value in the WeightedGraph nodes propagate into the MarkovProcessGraph nodes
- infers the timestamp on epoch-related nodes from the epoch start time
- propagates timestampMs into neo4j analysis output

This is really just for our end users to do easier analytics, though I might use it for 1st party analytics too.

I think timestamps were omitted in the first place in order to keep the MarkovProcessGraph as slim as possible. I think the analytics benefit for this is worth the increased memory usage.

# Test Plan
regression test: scdev credrank

also ran `scdev analysis -n` and used the neo4j import commands to verify nodes have timestamps. This query shows only the seed node does not get a timestamp:
`MATCH (n) WHERE not EXISTS(n.timestampMs) RETURN n LIMIT 25 `


s-ben:
Ty!

hozzjss (approved):
LFG

PR #3348, opened 2022-03-23
Bugfix: explicit directory creation for plugin outputs

blueridger:
# Description
Users are getting blocked by the error `[Error: ENOENT: no such file or directory, open '/home/runner/work/makerdao-cred/makerdao-cred/output/graphs/sourcecred/discourse/graph']` because our code is not running mkdir for plugin output directories. This bug didn't show in testing because on our dev laptops we were just overwriting within existing directories from a gh-pages clone rather than a fresh checkout of the main branch. This PR fixes the bug, making it so that the mkdir method is called down the directory chain before interacting with all plugin output directories.

# Test Plan
1. checkout sourcecred/cred/gh-pages
2. delete the ./output folder
3. scdev go --no-load


hozzjss (approved):
Looks good lfg

PR #3360, opened 2022-07-29
Dw 354/discord messages api issue

johnmark13:
Discord message IDs have passed the quintillion mark and this has led to an infinite loop in the codebase due to a string comparison on numeric IDs. This PR proposes a simple fix of casting the IDs to numbers before doing the comparison.

Manually tested and working, thanks to @hozzjss for the work.

### DIFF
--- a/packages/sourcecred/package.json
+++ b/packages/sourcecred/package.json
@@ -3,7 +3,7 @@
   "description": "a tool for communities to measure and reward value creation",
   "homepage": "https://sourcecred.io",
   "repository": "github:sourcecred/sourcecred",
-  "version": "0.11.2",
+  "version": "0.11.1",
   "private": false,
   "dependencies": {
     "@aws-sdk/util-base64-browser": "^3.37.0",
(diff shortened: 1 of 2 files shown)


packages/sourcecred/package.json
```diff
@@ -3,7 +3,7 @@
   "description": "a tool for communities to measure and reward value creation",
   "homepage": "https://sourcecred.io",
   "repository": "github:sourcecred/sourcecred",
-  "version": "0.11.1",
+  "version": "0.11.2",
```
hozzjss, line 6: no need to include the version upgrade

.github/workflows/yarn.yml
```diff
@@ -0,0 +1,30 @@
+name: Yarn CI
```
hozzjss, line 1: this whole file might be a different pr

hozzjss (changes requested):

packages/sourcecred/package.json
johnmark13: Agree.

